Issue #51, opened 2020-10-29
crashes observed while running soak-test


jgraettinger:
```
2020-10-29T18:50:36.935Z INFO  estuary::flow-consumer<11062> > FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
 2020-10-29T18:50:36.950Z INFO  estuary::flow-consumer<11062> >  1: 0xa093f0 node::Abort() [node]
 2020-10-29T18:50:36.951Z INFO  estuary::flow-consumer<11062> >  2: 0xa097fc node::OnFatalError(char const*, char const*) [node]
 2020-10-29T18:50:36.952Z INFO  estuary::flow-consumer<11062> >  3: 0xb8431e v8::Utils::ReportOOMFailure(v8::internal::Isolate*, char const*, bool) [node]
 2020-10-29T18:50:36.952Z INFO  estuary::flow-consumer<11062> >  4: 0xb84699 v8::internal::V8::FatalProcessOutOfMemory(v8::internal::Isolate*, char const*, bool) [node]
 2020-10-29T18:50:36.952Z INFO  estuary::flow-consumer<11062> >  5: 0xd31055  [node]
 2020-10-29T18:50:36.953Z INFO  estuary::flow-consumer<11062> >  6: 0xd316e6 v8::internal::Heap::RecomputeLimits(v8::internal::GarbageCollector) [node]
 2020-10-29T18:50:36.953Z INFO  estuary::flow-consumer<11062> >  7: 0xd3df65 v8::internal::Heap::PerformGarbageCollection(v8::internal::GarbageCollector, v8::GCCallbackFlags) [node]
 2020-10-29T18:50:36.954Z INFO  estuary::flow-consumer<11062> >  8: 0xd3ee15 v8::internal::Heap::CollectGarbage(v8::internal::AllocationSpace, v8::internal::GarbageCollectionReason, v8::GCCallbackFlags) [node]
 2020-10-29T18:50:36.954Z INFO  estuary::flow-consumer<11062> >  9: 0xd404bf v8::internal::Heap::HandleGCRequest() [node]
 2020-10-29T18:50:36.955Z INFO  estuary::flow-consumer<11062> > 10: 0xceefa5 v8::internal::StackGuard::HandleInterrupts() [node]
[...]

PR #59, opened 2020-11-09
moar testing

jgraettinger:
This is a grab-bag of unrelated PRs which are squashing small and medium issues I've encountered.
It also extends the set-based soak test substantially, by adding logical partitions and doing the same reduction, but in a register this time (which should give identical results).

Still working on re-producing the weird gRPC flow-control issue spotted the other day...


jgraettinger:
Fixes #51

psFried (approved):
LGTM

go/shuffle/read.go
```diff
@@ -89,14 +91,48 @@ func NewReadBuilder(
...
+			// We must receive from the stream.
+			if err = r.onRead(r.doRead()); err == nil && r.resp.TerminalError == "" {
+				continue
+			} else if err == io.EOF {
+				return message.Envelope{}, err
+			} else if err == nil {
+				return message.Envelope{}, errors.New(r.resp.TerminalError)
+			}
+
+			// Stream is broken, but may be retried.
```
psFried, line 120: Is there a case where we should stop re-trying and just return the error? I feel like maybe we should at some point just show the shard status as failed instead of continuing to re-try indefinitely. But I'd need some learning on the possible error conditions at this point and error handling for failed shards before I form a strong opinion.

psFried (approved):
Looks good % one question about error handling.

go/shuffle/read.go
jgraettinger: In general, we continue to retry so long as we can expect that circumstance might eventually change such that we can make meaningful progress.

That's because we already have a shard assignment who's lifetime scopes how long we want to retry, and should that assignment be removed, it'll cancel the context (aborting further retries). So long as the assignment context is valid (meaning we have the assignment and continue to hold its associated etcd lease), we keep trying.

We're also separately plumbing through terminal errors (things like schema validations), which do abort processing.
For all other retry-able errors, we _do_ want to pick up processing again after eg a multi-hour network partition is healed.

jgraettinger:
Thanks!

PR #60, opened 2020-11-10
Refactor CSV ingestion value parsing

psFried:
The [docs PR](https://github.com/estuary/docs/pull/4) should have a good description of the behavior. The behavior changes in this PR are actually fairly minimal, but I added some tests for existing behaviors as well.


examples/weird-types.flow.yaml
```diff
@@ -0,0 +1,92 @@
+collections:
+  - name: weird-types/optionals
+    schema:
+      type: object
+      properties:
+        theKey: { type: string }
+        string:
```
jgraettinger, line 7: nit: mix of ``property:\ntype: string`` here vs ``property: { type: string }`` below.

I've been using inline {} whenever it's a single-property object, if that sounds good to you for standardization.

go/ingest/ws_csv.go
```diff
@@ -0,0 +1,233 @@
...
+type wsCsvIngester struct {
+	buffer             *bytes.Buffer
+	csvReader          *csv.Reader
+	projections        []fieldMapping
+	pointers           []flow.Pointer
+	lastMustExistIndex int
+}
+
+// First frame is headers, subsequent frames are documents.
+func (self *wsCsvIngester) onHeader(collection *pf.CollectionSpec) error {
```
jgraettinger, line 101: can you run a linter over this? It will tell you `self` is an anti-pattern, for example.

VSCode's Go tooling should make this easy.

jgraettinger (approved):
LGTM, but please run a lint check per below

PR #61, opened 2020-11-11
standardize shard names

psFried:
Also includes an unrelated fix for catalog tests to compare floats using an epsilon value.


jgraettinger (approved):
LGTM, thanks for fixing

PR #62, opened 2020-11-13
Validate that test verify step docs are ordered

psFried:
The documents in a test verify step must be in lexicographical order,
according to the collection's key. This commit adds a verification
during catalog builds to verify that the documents are in the proper
order, so that we can fail fast when the user provides them out of
order.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/62)

### DIFF
--- a/crates/catalog/src/collection.rs
+++ b/crates/catalog/src/collection.rs
@@ -116,6 +116,17 @@ impl Collection {
             |row| row.get(0),
         )?)
     }
+
+    /// Returns the collection's key
+    pub fn key(&self, db: &DB) -> Result<Vec<String>> {
+        let json = db.query_row(
+            "SELECT key_json FROM collections WHERE collection_id = ?;",
+            rusqlite::params![self.id],
+            |row| row.get::<usize, String>(0),
+        )?;
+        let key = serde_json::from_str(&json)?;
+        Ok(key)
+    }
 }
 
 #[cfg(test)]
@@ -128,8 +139,8 @@ mod test {
     use serde_json::json;
 
     #[test]
-    fn test_register() -> Result<()> {
-        let db = create(":memory:")?;
+    fn test_register() {
+        let db = create(":memory:").unwrap();
 
         let schema = json!({
             "$anchor": "foobar",
@@ -165,13 +176,15 @@ mod test {
                     (1, 'application/vnd.estuary.dev-catalog-spec+yaml', X'1234', FALSE),
                     (10, 'application/schema+yaml', CAST(? AS BLOB), FALSE);",
             sql_params![schema],
-        )?;
+        )
+        .unwrap();
         db.execute(
             "INSERT INTO resource_urls (resource_id, url, is_primary) VALUES
                     (1, 'test://example/spec', TRUE),
                     (10, 'test://example/schema.json', TRUE)",
             sql_params![],
-        )?;
+        )
+        .unwrap();
 
         let spec: specs::Collection = serde_json::from_value(json!({
             "name": "test/collection",
@@ -181,15 +194,16 @@ mod test {
                 "field_a": {"location": "/a/a", "partition": true},
                 "field_b": {"location": "/b/b", "partition": false},
             }
-        }))?;
+        }))
+        .unwrap();
 
-        Scope::empty(&db)
+        let collection = Scope::empty(&db)
             .push_resource(Resource { id: 1 })
             .then(|scope| Collection::register(scope, &spec))
             .expect("failed to register collection");
 
         // Expect that the schema was processed.
-        assert!(Resource { id: 10 }.is_processed(&db)?);
+        assert!(Resource { id: 10 }.is_processed(&db).unwrap());
 
         // Expect the collection records the absolute schema URI, with fragment component.
         let dump = dump_tables(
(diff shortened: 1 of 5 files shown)


src/catalog/collection.rs
```diff
@@ -127,6 +138,45 @@ mod test {
     use rusqlite::params as sql_params;
     use serde_json::json;
 
+    #[test]
```
jgraettinger, line 141: A lot of repetitive fixture creation here. Really this is just about exercising these getters against _some_ fixture. How about folding the assertions themselves into `test_register` ?

src/catalog/error.rs
```diff
@@ -106,6 +107,9 @@ pub enum Error {
 
     #[error("The --catalog was not built by this version of flowctl. Catalog version: '{0}'")]
     CatalogVersionMismatch(String),
+
+    #[error(transparent)]
```
jgraettinger, line 111: nit: sigh this was already a dumping ground, but it's getting worse... 

No action for this PR

src/catalog/specs.rs
```diff
@@ -11,6 +11,15 @@ use std::time::Duration;
 #[schemars(example = "CollectionName::example")]
 pub struct CollectionName(#[schemars(schema_with = "CollectionName::schema")] String);
 
+impl<T> From<T> for CollectionName
```
jgraettinger, line 14: I don't think this should be an auto-magic conversion. If you want a CollectionName, you have to explicitly ask for one.

Otherwise it defeats the purpose of the newtype pattern, if I can convert (say) a RelativeUrl to a CollectionName via , eg, `.as_ref().clone().into()`. I can't read that and immediately know that I've subverted the type system.

src/catalog/test_case.rs
```diff
@@ -119,14 +124,121 @@ impl TestCase {
...
+
+#[derive(Debug)]
+pub struct TestExpectationsOutOfOrder {
+    collection_key: Vec<String>,
+    reordered_documents: Vec<Value>,
+}
+
+impl Display for TestExpectationsOutOfOrder {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        writeln!(f, "The expected documents must be provided in lixicographic order according to the collection key: [{}]",
```
jgraettinger, line 172: lexicographic. Also, "Documents to *verify* must... ?"

"expected" isn't a formal concept here; it's "verify" (unless we want to change it wholesale). Also please change TestExpectationsOutOfOrder to TestVerifyOutOfOrder (or similar) ? thx

src/catalog/error.rs
psFried: yeah I feel the same. I think it'll make more sense to revisit this as we break things up into separate crates. Specifically, I'm thinking about a possible factoring that separates the code for reading vs building a catalog, since most downstream crates will only need the read functionality.

jgraettinger (approved):
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/62)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @psFried)

___
*[src/catalog/error.rs, line 111 at r1](https://reviewable.io/reviews/estuary/flow/62#-MMG_wECtOswqAp2vUoX-r1-111:-MMHq-JB3kSa2MDvafc0:b4c9lyr) ([raw file](https://github.com/estuary/flow/blob/41beea3058c0ed0ed3c46f38eee1ea76214e35e7/src/catalog/error.rs#L111)):*

yeah I feel the same. I think it'll make more sense to revisit this as we break things up into separate crates. Specifically, I'm thinking about a possible factoring that separates the code for reading vs building a catalog, since most downstream crates will only need the read functionality.
</blockquote></details>

:thumbsup: 
___
*[src/catalog/test_case.rs, line 148 at r2](https://reviewable.io/reviews/estuary/flow/62#-MMHv0SS4e4Lg8_CgApM:-MMHv0SS4e4Lg8_CgApN:b935g66) ([raw file](https://github.com/estuary/flow/blob/bb77c042114383baeeb0de36596b43400d3a6331/src/catalog/test_case.rs#L148)):*
> ```Rust
> }
> 
> fn is_sorted_by<T, F>(seq: &[T], cmp_fun: F) -> bool
> ```

haha was about to ask "why are you implementing this?", and then saw that it's nightly only. Gross. Maybe just comment that we should be able to remove this when https://github.com/rust-lang/rust/issues/53485 resolves?

jgraettinger (commented):
> Reviewed 5 of 5 files at r2.
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/62)* status: all files reviewed, 1 unresolved discussion (waiting on @psFried)

psFried (commented):
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/62)* status: all files reviewed, 1 unresolved discussion (waiting on @jgraettinger)

___
*[src/catalog/test_case.rs, line 148 at r2](https://reviewable.io/reviews/estuary/flow/62#-MMHv0SS4e4Lg8_CgApM:-MMJGpBHE74r1B2NP6Vr:b-896fix) ([raw file](https://github.com/estuary/flow/blob/bb77c042114383baeeb0de36596b43400d3a6331/src/catalog/test_case.rs#L148)):*

haha was about to ask "why are you implementing this?", and then saw that it's nightly only. Gross. Maybe just comment that we should be able to remove this when https://github.com/rust-lang/rust/issues/53485 resolves?
</blockquote></details>

Done.

jgraettinger (approved):
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/62)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @psFried)

Issue #36, opened 2020-10-19
don't allow projections of integer-like object `properties` in schemas


jgraettinger:
We try to allow most anything, but explicit integer-like objection properties really mess with dynamic object construction via JSON Pointer. Specifically, we should prohibit the creation of projections `/1` or `/2` from this:

```
type: object
properties:
  "1": {type: string}
  "2": {type: integer}
```

etc. This is _not_ a prohibition against integer-like keys in objects across the board, just a restriction against giving them static projections.

The rationale is this restriction means we can unambiguously take a projection pointer like `/foo/1/bar` and know how to dynamically construct it. Otherwise, it's technically ambiguous whether `/foo` is an object or array.

PR #63, opened 2020-11-16
Fix several outstanding issues & chores

jgraettinger:
This PR is a grab-bag. Fixed chores:

 * Prohibits integer properties #36 
 * Adds `note` to diff output
 * Verifies that derivation transforms agree on shuffle key types
 * Generates '-' inference items for documentation of variable arrays, as discussed on Slack.
 * Fixes TypeScript type generation when objects with properties are mixed with `additionalProperties` or `patternProperties`.

I also updated inference to always generate child property / index locations, even if the parent could be another type.
This is both for documentation, and also to generate a more meaningful "this location may not always exist" rather than
"no inference at this location" if you forget to restrict the parent to `type: object`.

Finally, I re-worked types::Set to model FRACTIONAL vs INTEGER, and have NUMBER be sugar for both. FRACTIONAL / "fractional" is a compatible extension of the JSON schema spec which lets us round-trip this case. I've raised it on the JSON schema slack for consideration, but doubt that will go anywhere.

_Definitively_ recommend reviewing this commit-by-commit. I've wired up Reviewable to the Estuary org (which we can do now, since it's public) and recommend using it rather than GitHub.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/63)


psFried (approved):
> Reviewed 5 of 5 files at r1, 7 of 7 files at r2, 3 of 3 files at r3, 5 of 5 files at r4, 5 of 5 files at r5, 1 of 1 files at r6, 4 of 4 files at r7.
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/63)* status: all files reviewed, 2 unresolved discussions (waiting on @jgraettinger)

___
*[examples/stock-stats/schemas/L1-tick.schema.yaml, line 31 at r7](https://reviewable.io/reviews/estuary/flow/63#-MMHjdA23vemd16N1qNR:-MMHjdA23vemd16N1qNS:b-u50o5) ([raw file](https://github.com/estuary/flow/blob/1c53ce809f09a666cb2a7dc39b8b84a2bb3506c8/examples/stock-stats/schemas/L1-tick.schema.yaml#L31)):*
> ```YAML
> 
>   # Must be here because incompatible additionalProperties is used above.
>   _meta: { type: object }
> ```

Oh I hadn't thought about this. I feel like we should eventually either add this property automatically, or else error if it's not present and `additionalProperties` doesn't allow it. Fine to just deal with that later, though.
___
*[json/schema/types.rs, line 18 at r2](https://reviewable.io/reviews/estuary/flow/63#-MMHOk-20Wh9O-qXsFRC:-MMHOk-35HjGQpPAZhLs:b20tdd1) ([raw file](https://github.com/estuary/flow/blob/ab2d1cbfb9a6e45b2fb6c6cb50fc25d270a4b680/json/schema/types.rs#L18)):*
> ```Rust
> // INT_OR_FRACT is a composite for "number". It's not called NUMBER to avoid
> // giving the impression that this is a fundamental type.
> pub const INT_OR_FRAC: Set = Set(INTEGER.0 | FRACTIONAL.0);
> ```

I'm honestly not sure how I feel about this... My personal instinct would be to have a special case for interesctions/unions of the `NUMBER` and `INTEGER` types rather than introduce a new fundamental type. I'm hesitant to expose the "fractional" type, since it's not something we could undo without breaking compatibility. That said, the only actual downsides I see are kinda minor:

- We might spit out a "fractional" inference (e.g. in generated docs), which might be surprising/confusing to users
- If a user used "fractional" in their own schema, it would no longer be re-usable in other contexts

Those really aren't huge deals, IMO, but I'm also a little skeptical of the problem it's solving. The schema `allOf: [{type: number}, {type: integer}]` seems like the kind of thing that ought to be exceedingly rare in practice. I'm struggling to think of a valid reason for doing that, even if you're composing schemas maintained by someone else. I think it might still be worth solving at some point, given that type inferences are required in order for CSV ingestion, but it doesn't really seem worth the addition of a non-standard primitive type.

Another thought that just occurred to me is that we maybe don't _need_ to expose `"fractional"` as a type that users could use in their schemas, which could sidestep the compatibility issue.

jgraettinger (commented):
Thanks

> *[Reviewable](https://reviewable.io/reviews/estuary/flow/63#-:-MMHn6Cj4wtdR9j8qv5x:b-zcjvte)* status: all files reviewed, 2 unresolved discussions (waiting on @psFried)

___
*[examples/stock-stats/schemas/L1-tick.schema.yaml, line 31 at r7](https://reviewable.io/reviews/estuary/flow/63#-MMHjdA23vemd16N1qNR:-MMHmizb1bz9HfwL7tyy:brpyfee) ([raw file](https://github.com/estuary/flow/blob/1c53ce809f09a666cb2a7dc39b8b84a2bb3506c8/examples/stock-stats/schemas/L1-tick.schema.yaml#L31)):*

Oh I hadn't thought about this. I feel like we should eventually either add this property automatically, or else error if it's not present and `additionalProperties` doesn't allow it. Fine to just deal with that later, though.
</blockquote></details>

Erroring faster if it's not allowed seems reasonable (otherwise it errors when running tests, which also isn't a big deal). For later. Doing this well also depends on having a good way of synthesizing example documents, using defaults and examples of the annotations.
___
*[json/schema/types.rs, line 18 at r2](https://reviewable.io/reviews/estuary/flow/63#-MMHOk-20Wh9O-qXsFRC:-MMHn89o0WBn-1xv_r-G:brmxdwg) ([raw file](https://github.com/estuary/flow/blob/ab2d1cbfb9a6e45b2fb6c6cb50fc25d270a4b680/json/schema/types.rs#L18)):*
> My personal instinct would be to have a special case for interesctions/unions of the NUMBER and INTEGER types rather than introduce a new fundamental type.

Yea, that was my first approach too. I worked up a whole implementation that I ended up junking, in favor of this.

The unresolvable issue is that you can't represent the "difference" operation (or equivalently, "not"): what is `NUMBER minus INTEGER` ? Or equivalently, `!INTEGER` ?

Once I saw that, and realized there was otherwise going to be this unresolvable hole in the type algebra, I got motivated that 'fractional' was probably the best way forward.

I hit this in unit testing something else (I forget what). It's pretty easy to hit if you, say, are `$ref`'ing in another schema that uses a `number`, and overlay it with an `integer` restriction without realizing it's already specified in the base, or because you want the more specific type. The result is pretty confusing, because the location then has INVALID type.

I don't much care if we _ever_ advertise that `fractional` exists.

PR #76, opened 2020-12-03
bindings: add crate and Go module with Service inter-op mechanism

jgraettinger:
A Service looks like a bidirectional channel, but is one which is
synchronously polled to send and receive framed message code and data
payloads. It's intended as a fundamental building block for building
bidirectional in-process streaming services that interoperate between
Go and Rust -- an interface which is a heck of a lot faster than gRPC.

It achieves zero-copy semantics by pushing Go []byte buffers down
into CGO invocations, and surfacing Rust-owned []byte arenas and
Frame payloads. There are some rules callers need to adhere to,
such as not scribbling "sent" memory until Poll() is called, and
not referencing Rust-owned memory returned by Poll() after _another_
Poll() call is made.

It amortizes the CGO overhead by vectorizing dispatch. On the Rust
side, some common functions are provided to monomorphize appropriate
bindings for an implementation of a Service trait.

Lies, damn lies, and benchmarks:

This shows a variety of stride patterns (number of sent messages
per Poll() call), some of which are deliberately worst case.
A comparative benchmark with a "naieve" CGO service is included,
which is also zero-copy (via some unsafe hacks) and approximates an
optimized non-Service implementation.

    $ go test ./go/bindings/ -v -bench='.' -benchtime 3s
    === RUN   TestUpperService
    --- PASS: TestUpperService (0.00s)
    === RUN   TestUpperServiceNaieve
    --- PASS: TestUpperServiceNaieve (0.00s)
    goos: linux
    goarch: amd64
    pkg: github.com/estuary/flow/go/bindings
    BenchmarkUpperService
    BenchmarkUpperService/1
    BenchmarkUpperService/1-24              43287075                81.8
    ns/op
    BenchmarkUpperService/3
    BenchmarkUpperService/3-24              48669957                71.3
    ns/op
    BenchmarkUpperService/4
    BenchmarkUpperService/4-24              100000000               33.1
    ns/op
    BenchmarkUpperService/11
    BenchmarkUpperService/11-24             82098810                43.3
    ns/op
    BenchmarkUpperService/15
    BenchmarkUpperService/15-24             88228374                40.6
    ns/op
    BenchmarkUpperService/17
    BenchmarkUpperService/17-24             126318896               28.7
    ns/op
    BenchmarkUpperService/31
    BenchmarkUpperService/31-24             100000000               33.5
    ns/op
    BenchmarkUpperService/32
    BenchmarkUpperService/32-24             149790940               24.0
    ns/op
    BenchmarkUpperService/63
    BenchmarkUpperService/63-24             127205176               28.3
    ns/op
    BenchmarkUpperService/137
    BenchmarkUpperService/137-24            150098632               24.1
    ns/op
    BenchmarkUpperService/426
    BenchmarkUpperService/426-24            157566507               22.9
    ns/op
    BenchmarkUpperServiceNaieve
    BenchmarkUpperServiceNaieve-24          38690248                91.2
    ns/op
    PASS
    ok      github.com/estuary/flow/go/bindings     55.728s

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/76)

### DIFF
--- a/crates/bindings/src/service.rs
+++ b/crates/bindings/src/service.rs
@@ -1,30 +1,4 @@
-/// Service is a trait implemented by services which may be called from Go.
-pub trait Service {
-    /// Error type returned by Service invocations.
-    type Error: std::error::Error;
-    /// Create a new instance of the Service.
-    fn create() -> Self;
-    /// Invoke with the given op-code & |data| payload.
-    /// It extends |arena| with any returned []byte data, and pushes output messages onto |out|.
-    fn invoke(
-        &mut self,
-        code: u32,
-        data: &[u8],
-        arena: &mut Vec<u8>,
-        out: &mut Vec<Out>,
-    ) -> Result<(), Self::Error>;
-}
-
-/// Output frame produced by a Service.
-#[repr(C)]
-pub struct Out {
-    /// Service-defined response code.
-    pub code: u32,
-    /// Begin data offset into the Channel arena.
-    pub begin: u32,
-    /// End data offset into the Channel arena.
-    pub end: u32,
-}
+pub use protocol::cgo::{Out, Service};
 
 /// InN is a variadic input which invokes itself against a Service.
 pub trait InN {
--- a/crates/bindings/src/service.rs
+++ b/crates/bindings/src/service.rs
@@ -1,0 +1,233 @@
+/// Service is a trait implemented by services which may be called from Go.
+pub trait Service {
+    /// Error type returned by Service invocations.
+    type Error: std::error::Error;
+    /// Create a new instance of the Service.
+    fn create() -> Self;
+    /// Invoke with the given op-code & |data| payload.
+    /// It extends |arena| with any returned []byte data, and pushes output messages onto |out|.
+    fn invoke(
+        &mut self,
+        code: u32,
+        data: &[u8],
+        arena: &mut Vec<u8>,
+        out: &mut Vec<Out>,
+    ) -> Result<(), Self::Error>;
+}
+
+/// Output frame produced by a Service.
+#[repr(C)]
+pub struct Out {
+    /// Service-defined response code.
+    pub code: u32,
+    /// Begin data offset into the Channel arena.
+    pub begin: u32,
+    /// End data offset into the Channel arena.
+    pub end: u32,
+}
+
+/// InN is a variadic input which invokes itself against a Service.
+pub trait InN {
+    fn invoke<S: Service>(
+        self: &Self,
+        svc: &mut S,
+        arena: &mut Vec<u8>,
+        out: &mut Vec<Out>,
+    ) -> Result<(), S::Error>;
+}
+
+/// Input frame produced from CGO, which is a single service invocation.
+/// 16 bytes, or 1/4 of a typical cache line.
+#[repr(C)]
+pub struct In1 {
+    data_ptr: *const u8,
+    data_len: u32,
+    code: u32,
+}
+
+impl InN for In1 {
+    #[inline]
+    fn invoke<S: Service>(
+        self: &Self,
+        svc: &mut S,
+        arena: &mut Vec<u8>,
+        out: &mut Vec<Out>,
+    ) -> Result<(), S::Error> {
+        svc.invoke(
+            self.code,
+            unsafe { std::slice::from_raw_parts(self.data_ptr, self.data_len as usize) },
+            arena,
+            out,
+        )
+    }
+}
+
+/// Four invocations, composed into one struct.
+/// 64 bytes, or one typical cache line.
+#[repr(C)]
+pub struct In4 {
+    in0: In1,
+    in1: In1,
+    in2: In1,
+    in3: In1,
+}
+
+impl InN for In4 {
+    #[inline]
+    fn invoke<S: Service>(
+        self: &Self,
+        svc: &mut S,
+        arena: &mut Vec<u8>,
+        out: &mut Vec<Out>,
+    ) -> Result<(), S::Error> {
+        self.in0.invoke(svc, arena, out)?;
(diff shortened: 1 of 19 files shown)


crates/bindings/src/service.rs
```diff
@@ -0,0 +1,230 @@
...
+    out_cap: usize,
+
+    // Opaque service pointer.
+    svc_impl: *mut ServiceImpl,
+}
+
+/// Create a new Service instance, wrapped in an owning Channel.
+/// This is intended to be monomorphized by each Service implementation,
+/// and exposed via cbindgen.  See the UpperCase service for an example.
+#[inline]
```
psFried, line 89: What's the purpose of the `#[inline]`? I wouldn't expect calls to these to ever be inlined in practice since it's called from go.

crates/bindings/src/service.rs
```diff
@@ -0,0 +1,230 @@
...
+    in0: In1,
+    in1: In1,
+    in2: In1,
+    in3: In1,
+}
+
+/// Sixteen invocations, composed into one struct.
+/// 256 bytes, or four typical cache lines.
+#[repr(C)]
+pub struct In16 {
```
psFried, line 32: I assume there's something preventing you from using `[In1; 16]` instead of the struct. Curious what prevents that from working. Not a big deal, but there's a few places where it'd be cleaner to loop over an array. Might also be able to just use some `unsafe` voodoo to `transmute` this into an array on the rust side, if there's a reason to keep the array out of the C header.

crates/bindings/src/upper_case.rs
```diff
@@ -0,0 +1,82 @@
...
+            end: arena.len() as u32,
+        });
+
+        self.sum_length += data.len() as u32;
+    }
+}
+
+// Define cbindgen <=> CGO hooks for driving the UpperCase service.
+
+#[no_mangle]
```
psFried, line 37: Might be more trouble than it's worth, but there's the [paste crate](https://github.com/dtolnay/paste) that allows you to template sets of functions like this using `macro_rules` macros. So you'd be able to do something like `expose_c_functions!{"upper_case", UpperCase}` to generate these. Macros can be tough to read/write, though, so fine if we'd rather just type it out the old fashioned way.

psFried (commented):
Just had a look and left a couple of thoughts inline. Looks like it's going well!

crates/bindings/src/service.rs
jgraettinger: These aren't called from go, they're called from monomorphized one-liners for each specific service.

crates/bindings/src/service.rs
jgraettinger: Yea, Go needs to be able to access each of these fields. They're individually set by matching implementations on the CGO side.

This song and dance is largely driven by hard requirements of CGO and its needs re: tracking and understanding where every Go pointer is. This forces the ABI to use explicit fields and basically rules out arrays of any kind. [gory details](https://golang.org/cmd/cgo/#hdr-Passing_pointers)

I'd love to, for example, have the Go side pass an array pointer & size of `In1` structs into a single invoke. But, that violates the rule "Go code may pass a Go pointer to C provided the Go memory to which it points does not contain any Go pointers". AKA you can't nest.

crates/bindings/src/upper_case.rs
jgraettinger: I'd like to keep it simple for now. Can revisit later.

jgraettinger:
Think this is ready to land. I have one more planned tweak (making `code` an i32 instead of a u32), but otherwise I've transitioned to using it rather than implementing it. Thx

psFried (changes requested):
> Reviewed 5 of 11 files at r1, 4 of 6 files at r2.
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/76)* status: 2 of 19 files reviewed, 1 unresolved discussion (waiting on @jgraettinger and @psFried)

___
*[go/bindings/service_test.go, line 1 at r2](https://reviewable.io/reviews/estuary/flow/76#-MOILBxAE5IKjwx-6Zg-:-MOILBxB0W9phLdQAe9y:b1szkos) ([raw file](https://github.com/estuary/flow/blob/9ab08e2796694f2788455be8bb2c63f118e8a855/go/bindings/service_test.go#L1)):*
> ```Go
> package bindings
> ```

It would be nice to have a test case that has a Service returning an error, unless I just missed that.

jgraettinger (commented):
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/76)* status: 2 of 19 files reviewed, 1 unresolved discussion (waiting on @psFried)

___
*[go/bindings/service_test.go, line 1 at r2](https://reviewable.io/reviews/estuary/flow/76#-MOILBxAE5IKjwx-6Zg-:-MOIMJ_vAs_wZqu0eJNE:b-htmrkw) ([raw file](https://github.com/estuary/flow/blob/9ab08e2796694f2788455be8bb2c63f118e8a855/go/bindings/service_test.go#L1)):*

It would be nice to have a test case that has a Service returning an error, unless I just missed that.
</blockquote></details>

it's part of several test cases / search for "whoops!"

psFried (approved):
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/76)* status: 2 of 19 files reviewed, all discussions resolved (waiting on @psFried)

___
*[go/bindings/service_test.go, line 1 at r2](https://reviewable.io/reviews/estuary/flow/76#-MOILBxAE5IKjwx-6Zg-:-MOIO5hu0ytOTR_k_aKo:b-l9sfo) ([raw file](https://github.com/estuary/flow/blob/9ab08e2796694f2788455be8bb2c63f118e8a855/go/bindings/service_test.go#L1)):*

it's part of several test cases / search for "whoops!"
</blockquote></details>

lol whoops indeed!

PR #77, opened 2020-12-07
doc: relax `merge` and `append` rules, allowing for LHS to be null

jgraettinger:
When null, these reduction annotations behave as no-ops.

This behavior allows the current register state to condition whether
an `merge` or `append` reduction should be done, which otherwise isn't
possible as the "update" lambda doesn't have access to the register.

It's more efficient than reducing, and then pruning (using prune
annotations that don't exist yet), because we can track and optimize for
the fact that the register value hasn't change from its loaded RocksDB
state.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/77)

### DIFF
--- a/crates/doc/src/reduce/strategy.rs
+++ b/crates/doc/src/reduce/strategy.rs
@@ -9,7 +9,9 @@ use std::convert::TryFrom;
 #[derive(Serialize, Deserialize, Debug)]
 #[serde(tag = "strategy", deny_unknown_fields, rename_all = "camelCase")]
 pub enum Strategy {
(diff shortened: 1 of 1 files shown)


psFried (approved):
> Reviewed 1 of 1 files at r1.
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/77)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jgraettinger)

psFried:
I think this looks great. Didn't have any comments on the code, since it looks so straightforward.

Issue #68, opened 2020-11-16
catalog specs redux


jgraettinger:
Tracking issue for planned changes to catalog specifications:

  * [x] `materializationTarget` will be reworked to `endpoint`, which is meaningful both for captures and materializations. It has a *name* and *configuration* (parameters specific to the driver).
  * [x] `materializations` will be a top-level entity, and binds a *target* name, *collection*, and *endpoint* with *configuration*. The (target, collection, endpoint) tuple must be unique. Target is intended to be meaningful to the endpoint (eg, a topic or table name). 
  * [x] `captures` will be a top-level entity, and binds a *source* name, *collection*, and *endpoint* with *configuration* (like materializations). The (source, collection, endpoint) tuple must be unique. Source is meaningful to the endpoint (eg, a source table for CDC).
  * [x] Allow cycles in derivation dependencies (allow both A -> B -> A and A -> A) during catalog builds.
  * [x] Change representation of collections in flow.yaml to be an object instead of an array. The keys will be the collection name.
  * [x] Allow cyclic imports between catalog resources (implied requirement of derivation cycles).
  * [x] Verify referential integrity at the _end_ of build stages, so we can present a list of errors rather than fail immediately with just one (c.f. [pragma integrity_check](https://www.sqlite.org/pragma.html#pragma_integrity_check) and [foreign_key_check](https://www.sqlite.org/pragma.html#pragma_foreign_key_check))
  * [x] Similarly, lift current fail-fast triggers to inspection passes which collect and report multiple errors.
  * [x] Eliminate internal DB identifiers (ex resource_id, collection_id). Instead, use content-addressed primary keys (collection_name, canonical resource uri). Goal is to make database tables an assemblage of pure "facts" learned from specs, which can be independently manipulated during processing without regard to cross-table DB states (like rowids).
  * [x] Similarly, directly embed JSON specs which don't _need_ to be bespoke tables (lambdas, bootstraps, partition selectors, etc). Only named resources should be table entities.
  * [x] Enable modular SQL test (ideally using insta, and without using sqlite CLI).
  * [x] Separate current `catalog` crate into `catalog-models` (concerned with representation only), vs `catalog-build` (concerned with build passes).

Guiding principles:
[...]

PR #78, opened 2020-12-08
Data bindings

psFried:
WIP on catalog-specs-redux

This is a work in progress on #68 that implements the YAML specification. The `materializations` and `captures` here are non-functional, so the old materialization machinery was left in place for now. The idea is for the two to live side-by-side until the new materialization and captures are fully functional.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/78)

### DIFF
--- a/crates/catalog/src/capture.rs
+++ b/crates/catalog/src/capture.rs
@@ -1,0 +1,66 @@
+use crate::{specs, Collection, Endpoint, Error, Resource, Result, Scope, DB};
+
+#[derive(Debug, Copy, Clone)]
+pub struct Capture {
+    id: i64,
+    resource: Resource,
+}
+
+impl Capture {
+    pub fn register(scope: Scope, spec: &specs::Capture) -> Result<Capture> {
+        let (endpoint_id, source_entity): (Option<i64>, &str) = match &spec.source {
+            specs::CaptureSource::External { endpoint, target } => {
+                let resolved = scope
+                    .push_prop("endpoint")
+                    .then(|scope| Endpoint::get_by_name(scope, endpoint.as_str()))?;
+                (Some(resolved.id), target.as_str())
+            }
+            specs::CaptureSource::Builtin(builtin_source) => (None, builtin_source.type_name()),
+        };
+
+        let collection = scope
+            .push_prop("target")
+            .then(|scope| Collection::get_imported_by_name(scope, spec.target.name.as_ref()))?;
+        let resource = scope.resource();
+
+        let mut stmt = scope.db.prepare_cached("INSERT INTO captures
+                                               (resource_id, endpoint_id, source_entity, target_collection_id)
+                                               VALUES (?, ?, ?, ?);")?;
+
+        stmt.execute(rusqlite::params![
+            resource.id,
+            endpoint_id,
+            source_entity,
+            collection.id
+        ])?;
+        let id = scope.db.last_insert_rowid();
+
+        Ok(Capture { id, resource })
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use crate::{dump_tables, test_register};
+
+    #[test]
+    fn capture_is_registered_for_flow_ingester() {
+        let yaml = r##"
+            collections:
+              - name: foo
+                key: [/id]
+                schema:
+                  type: object
+                  properties:
+                    id: { type: integer }
+                  required: [id]
+
+            captures:
+              - source: flow-ingester
+                target: { name: foo }
+            "##;
+        let db = test_register(yaml).expect("failed to register");
+        let results = dump_tables(&db, &["endpoints", "captures"]).unwrap();
+        insta::assert_yaml_snapshot!(results);
+    }
+}
--- a/crates/catalog/src/catalog.sql
+++ b/crates/catalog/src/catalog.sql
@@ -224,8 +224,12 @@ CREATE TABLE materialization_targets (
 );
 
 
--- Endpoints
+-- Endpoints each represent connection information for an external system, which may be used for
+-- either a materialization or a capture.
 --
+-- :resource_id:
+--     Optional reference to the resource that defined this endpoint. This will be NULL only for
+--     the "builtin" endpoints provided by flow-ingester.
 -- :endpoint_name:
 --     Human-readable identifier of this endpoint, as provided in the catalog spec.
 -- :endpoint_type:
@@ -235,7 +239,7 @@ CREATE TABLE materialization_targets (
 --     The URI to use to connect to the target system.
 CREATE TABLE endpoints (
     endpoint_id INTEGER PRIMARY KEY NOT NULL,
-    resource_id INTEGER NOT NULL REFERENCES resources (resource_id),
+    resource_id INTEGER REFERENCES resources (resource_id),
     endpoint_name TEXT NOT NULL,
     endpoint_type TEXT NOT NULL CONSTRAINT 'endpoint_type must be a recognized type' CHECK(endpoint_type IN ('postgres', 'sqlite')),
     endpoint_uri TEXT NOT NULL,
@@ -243,6 +247,7 @@ CREATE TABLE endpoints (
     UNIQUE (endpoint_name COLLATE NOCASE),
     CONSTRAINT "Endpoint name isn't valid (may include Unicode letters, numbers, -, _, ., or /)" CHECK (endpoint_name REGEXP '^[\pL\pN\-_./]+$')
 );
+
 
 -- Materializations bind a source collection, endpoint, and target.
 --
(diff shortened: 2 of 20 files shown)


jgraettinger (changes requested):
I got through most of this, up until materialization.rs, and appreciate more why you're focused on the postgres driver implementation / that there are details still to figure out w/ respect to the driver.

Come to think of it, if we are going to standardize on open API, I don't know why the Driver interface wouldn't be open API also...

> *[Reviewable](https://reviewable.io/reviews/estuary/flow/78#-:-MOHxHl7BF5MUWQYPZle:bimtd4t)* status: 0 of 20 files reviewed, 7 unresolved discussions (waiting on @jgraettinger and @psFried)

___
*[crates/catalog/src/catalog.sql, line 262 at r4](https://reviewable.io/reviews/estuary/flow/78#-MOHlrR-8of_z0Jfq16O:-MOHlrR-8of_z0Jfq16P:b-t380o6) ([raw file](https://github.com/estuary/flow/blob/be892a19627142a2ebc90a5309093eab0eb55e8e/crates/catalog/src/catalog.sql#L262)):*
> ```TSQL
> --     database endpoints, this will be a table name. But this could also represent an S3 object key
> --     or prefix, the name of a kinesis stream, kafka topic, etc.
> -- :source_collection_id:
> ```

This should just be collection_id, I think.

"source collection" is too conceptually bound to derivations / transformations.
___
*[crates/catalog/src/catalog.sql, line 270 at r4](https://reviewable.io/reviews/estuary/flow/78#-MOHsKXuEYheur1i9vKt:-MOHsKXv9bD3PfBEJtb9:b-bgl6cq) ([raw file](https://github.com/estuary/flow/blob/be892a19627142a2ebc90a5309093eab0eb55e8e/crates/catalog/src/catalog.sql#L270)):*
> ```TSQL
>     resource_id INTEGER NOT NULL REFERENCES resources (resource_id),
>     endpoint_id INTEGER NOT NULL REFERENCES endpoints (endpoint_id),
>     target_entity TEXT,
> ```

What do you think about `binding` instead of `target_entity` & `source_entity` ? The latter feel like two words for the same thing. Target & source are already over-used and perhaps too vague as well.

Dunno, reaching here, but seeking to bring in some different, descriptive & succinct nomenclature.
___
*[crates/catalog/src/catalog.sql, line 295 at r4](https://reviewable.io/reviews/estuary/flow/78#-MOHunfc9pDyx7glOqCH:-MOHunfc9pDyx7glOqCI:b-ioifg5) ([raw file](https://github.com/estuary/flow/blob/be892a19627142a2ebc90a5309093eab0eb55e8e/crates/catalog/src/catalog.sql#L295)):*
> ```TSQL
> 
> -- View of materializations with projections that are invalid.
> -- :materialization_id:
> ```

nit: comment and auto-format this? (use the vscode sql extension?).
___
*[crates/catalog/src/catalog.sql, line 317 at r4](https://reviewable.io/reviews/estuary/flow/78#-MOHvjZ09FOJcxUZtCDw:-MOHvjZ09FOJcxUZtCDx:b-pxexa8) ([raw file](https://github.com/estuary/flow/blob/be892a19627142a2ebc90a5309093eab0eb55e8e/crates/catalog/src/catalog.sql#L317)):*
> ```TSQL
> -- View of each materialization as a complete JSON representation of a CollectionSpec from
> -- flow-protocol.
> CREATE VIEW materializations_json AS
> ```

i'm not sure what this really _is_  -- materializations joined with collections specs? but it should be using the `collections_json` view which already produces JSON CollectionSpecs.
___
*[crates/catalog/src/endpoint.rs, line 10 at r4](https://reviewable.io/reviews/estuary/flow/78#-MOHwCo10MAzhnJ01W3K:-MOHwCo10MAzhnJ01W3L:bxsrt24) ([raw file](https://github.com/estuary/flow/blob/be892a19627142a2ebc90a5309093eab0eb55e8e/crates/catalog/src/endpoint.rs#L10)):*
> ```Rust
> 
> impl Endpoint {
>     pub fn register(scope: Scope, endpoint_name: &str, spec: &specs::Endpoint) -> Result<Endpoint> {
> ```

nit: comments on public fn's
___
*[crates/catalog/src/endpoint.rs, line 11 at r4](https://reviewable.io/reviews/estuary/flow/78#-MOHw5JQExU4ZWnSZzIg:-MOHw5JQExU4ZWnSZzIh:b8dlat7) ([raw file](https://github.com/estuary/flow/blob/be892a19627142a2ebc90a5309093eab0eb55e8e/crates/catalog/src/endpoint.rs#L11)):*
> ```Rust
> impl Endpoint {
>     pub fn register(scope: Scope, endpoint_name: &str, spec: &specs::Endpoint) -> Result<Endpoint> {
>         let endpoint_type = spec.type_name();
> ```

nit: inline into params! block below ? 
___
*[crates/catalog/src/materialization.rs, line 57 at r4](https://reviewable.io/reviews/estuary/flow/78#-MOHx4y--u6oOCooensm:-MOHx4y--u6oOCooensn:b5ycmcw) ([raw file](https://github.com/estuary/flow/blob/be892a19627142a2ebc90a5309093eab0eb55e8e/crates/catalog/src/materialization.rs#L57)):*
> ```Rust
> // Queries the materialization_invalid_projections view and returns any errors found there.
> // This must only be called after the materialization has been inserted.
> fn validate_materialization_fields(db: &DB, materialization_id: i64) -> Result<()> {
> ```

nit: why not an impl Materialization ?

jgraettinger (changes requested):
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/78)* status: 0 of 20 files reviewed, 7 unresolved discussions (waiting on @psFried)

___
*[crates/catalog/src/catalog.sql, line 270 at r4](https://reviewable.io/reviews/estuary/flow/78#-MOHsKXuEYheur1i9vKt:-MOS-46g-uPxYoTg-2dG:bgxxm8t) ([raw file](https://github.com/estuary/flow/blob/be892a19627142a2ebc90a5309093eab0eb55e8e/crates/catalog/src/catalog.sql#L270)):*

What do you think about `binding` instead of `target_entity` & `source_entity` ? The latter feel like two words for the same thing. Target & source are already over-used and perhaps too vague as well.

Dunno, reaching here, but seeking to bring in some different, descriptive & succinct nomenclature.
</blockquote></details>

One other thought is that, while `captures` and `materializations` deserve to be different specification concepts, IMO they shouldn't be different catalog database tables. The direction (capture vs binding) would instead be a bool column on a unified table.

psFried (commented):
Yeah it would be really nice if we could use a single interface definition language for all services that a user might implement. I think it bears a bit of research. I'm working on writing up a separate GH issue where we can talk about that.

There's a lot of good suggestions here. Where I didn't respond, my intent is just to leave it unresolved until I actually make the changes.

> *[Reviewable](https://reviewable.io/reviews/estuary/flow/78#-:-MOS94Mz-yDYDmukM_pv:b-4w38ds)* status: 0 of 20 files reviewed, 7 unresolved discussions (waiting on @jgraettinger and @psFried)

___
*[crates/catalog/src/catalog.sql, line 270 at r4](https://reviewable.io/reviews/estuary/flow/78#-MOHsKXuEYheur1i9vKt:-MOS4_OF2G1arBa6Dm_N:b-kex5us) ([raw file](https://github.com/estuary/flow/blob/be892a19627142a2ebc90a5309093eab0eb55e8e/crates/catalog/src/catalog.sql#L270)):*

One other thought is that, while `captures` and `materializations` deserve to be different specification concepts, IMO they shouldn't be different catalog database tables. The direction (capture vs binding) would instead be a bool column on a unified table.
</blockquote></details>

Yeah I'm not super keen on `target_entity`, but `binding` doesn't seem any less confusing to me, honestly. The things it may refer to are database tables, broker topics, maybe file globs, etc. Basically "a thing within the given endpoint". In the past I've used the term `path` for this same concept, which I'm also not super fond of. Though, now that I think of it, maybe `path` is less confusing than `binding` or `_entity`, and it also has the benefit of being pretty clear in the case that the endpoint is anything like a filesystem.

Regarding combining the two tables, I don't think I see the benefit there. Am I missing something? Materializations need to have an associated set of selected fields, which captures don't need. And the whole captures protocol is yet to be designed, so there may also be additional data that we need to store for captures but not materializations, too. The burning question there (which I'm trying not to get too sidetracked on) is how we should handle converting the multitude of source formats into JSON documents (i.e. is each driver responsible for converting source formats into json, or do we handle that generically).
___
*[crates/catalog/src/catalog.sql, line 317 at r4](https://reviewable.io/reviews/estuary/flow/78#-MOHvjZ09FOJcxUZtCDw:-MOS1J9H37HGk3tw_mNs:bz95yf6) ([raw file](https://github.com/estuary/flow/blob/be892a19627142a2ebc90a5309093eab0eb55e8e/crates/catalog/src/catalog.sql#L317)):*

i'm not sure what this really _is_  -- materializations joined with collections specs? but it should be using the `collections_json` view which already produces JSON CollectionSpecs.
</blockquote></details>

This re-uses the same shape as a CatalogSpec, but differs in that materializations_json only includes the projections that were resolved for each materialization. The intent was for this view to be used to get the payloads for Validate and Apply RPCs. The new protocol does things a little differently, so this is probably not even needed anymore.

psFried:
Closing this PR, as we've gone with a different approach.

PR #79, opened 2020-12-08
initial design of materialization grpc protocol

psFried:
Here's a walkthrough of two different scenarios for materializations. The first describes how it would work when we're doing a new materialization, and the second is for a subsequent build/apply of the catalog with the same materialization.

First time materialization to a database:

- At Build time:
    - Flowctl build sends MaterializationSpec with all projections to ValidateMaterializationBuild
    - MaterializationBuildResponse status will be either `ERROR` or `NOT_READY`.
        - Error is considered terminal and the error message is displayed to user as a build error.
    - The constraints from the response are used to select/validate the fields for the materialization
      before persisting them to the catalog. Build error if no valid solution is possible given the
      `include` or `exclude` from the spec.
- At Apply time:
    - ValidateMaterializationBuild is called again with the final set of fields. An `ERROR` status fails
      the build.
    - If response status is `NOT_READY`, and user has given consent (`initialization_payload` from last
      response may be displayed), then invoke the `ApplyMaterialization` RPC and assert we get a `READY`
      status in the response, or bail if anything else.
    - If user has not given consent to apply the `initialization_payload`, then we continue on our happy
      way and assume that they'll take care of it themselves. 
- At runtime:
    - If the materialization starts before they apply the DDL, the `StartMaterializationShard` rpc 
      will not return a `READY` status, which will be considered a shard failure.
    - If the DDL is initalized, then `StartMaterializationShard` returns a `READY` status, which
      allows the materialization to proceed.

Subsequent builds of a catalog with the same materialization to Postgres:

- At Build time:
    - Flowctl build still sends MaterializationSpec with all projections to ValidateMaterializationBuild
    - MaterializationBuildResponse status will be either `ERROR` or `READY`.
        - Error is considered terminal and the error message is displayed to user as a build error.
        - READY means that the materialization has been initialized, even though the fields may be
          different.
    - For systems where we don't handle schema migrations (basically all of them), constraints from the
      response will indicate that the specific fields that were already materialized are all
      `FIELD_REQUIRED`, and all others are `FIELD_FORBIDDEN`. If the set of projections has changed, it
      will trigger a build error with the specific constraint violations, so we'll be able to say
      exactly which fields are missing or extra.
    - If someone did want to handle schema migrations, then they'd simply be more permissive in the
      projection constraints that get returned.
- At Apply time:
    - ValidateMaterializationBuild is called again with the final set of fields. 
    - The response should indicate a `READY` status, since the table is already created with the
      expected columns.
    - An `ERROR` status fails the build.
- At runtime:
    - `StartMaterializationShard` will still be called and the implementation is expected to return
      an `ERROR` if the fields are not compatible, though in practice this condition would be caught
      earlier during the build process. But it's important that this is the final check of what's
      OK to proceed, since there could be manual processes for schema migration that we might want to
      make room for in the future.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/79)

### DIFF
--- a/crates/protocol/src/flow.rs
+++ b/crates/protocol/src/flow.rs
@@ -674,6 +674,212 @@ pub struct ClearRegistersResponse {
     #[prost(message, optional, tag = "2")]
     pub header: ::std::option::Option<super::protocol::Header>,
 }
+/// Represents a materialization of the given collection to an external system with the given
+/// endpoint.
+#[derive(Clone, PartialEq, ::prost::Message)]
+pub struct MaterializationSpec {
+    #[prost(string, tag = "1")]
+    pub materialization_id: std::string::String,
+    #[prost(string, tag = "2")]
+    pub endpoint_type: std::string::String,
+    #[prost(string, tag = "3")]
+    pub endpoint: std::string::String,
+    #[prost(string, tag = "4")]
+    pub target_entity: std::string::String,
+    #[prost(message, optional, tag = "5")]
+    pub collection: ::std::option::Option<CollectionSpec>,
+}
+/// Result of validating a specific projection.
+#[derive(Clone, PartialEq, ::prost::Message)]
+pub struct ProjectionResult {
+    /// The field name of the projection.
+    #[prost(string, tag = "1")]
+    pub field: std::string::String,
+    /// The validation result.
+    #[prost(enumeration = "ProjectionConstraint", tag = "2")]
+    pub constraint: i32,
+    /// Optional human readable reason for the given constraint. Implementations are strongly
+    /// encouraged to supply a good descriptive message here, especially if the constraint is
+    /// anything other than LOCATION_RECOMMENDED or LOCATION_OPTIONAL.
+    #[prost(string, tag = "3")]
+    pub reason: std::string::String,
+}
+/// Describes the current state of a Materialization as stored in the remote system.
+#[derive(Clone, PartialEq, ::prost::Message)]
+pub struct MaterializationStatus {
+    #[prost(enumeration = "materialization_status::StatusCode", tag = "1")]
+    pub status_code: i32,
+    /// Human readable message which will be shown to the user.
+    #[prost(string, tag = "2")]
+    pub status_message: std::string::String,
+}
+pub mod materialization_status {
+    #[derive(Clone, Copy, Debug, PartialEq, Eq, Hash, PartialOrd, Ord, ::prost::Enumeration)]
+    #[repr(i32)]
+    pub enum StatusCode {
+        /// The materialization has not been initialized yet.
+        NotReady = 0,
+        /// The materialization has been initialized.
+        Ready = 1,
+        /// There's been an error connecting to the remote system, or the remote system is in an invalid
+        /// state. The status_message should have details.
+        Error = 2,
+    }
+}
+/// Sent in response to validating or applying a materialization.
+#[derive(Clone, PartialEq, ::prost::Message)]
+pub struct MaterializationBuildResponse {
+    /// The status of the materialization, which may account for the current state of the remote
+    /// system, in addition to the spec provided in a request.
+    #[prost(message, optional, tag = "1")]
+    pub status: ::std::option::Option<MaterializationStatus>,
+    /// The initialization payload is a system-specific text that can be used to initialize the
+    /// materialization in the target system. For example, the SQL CREATE TABLE statements for a
+    /// database. Not all systems will have this. It's provided here so that the user can choose to
+    /// initialize the target system themselves rather than having it done automatically.
+    #[prost(string, tag = "2")]
+    pub initialization_payload: std::string::String,
+    /// The type of this materialization. This is always chosen by the target system, as most systems
+    /// can only support one type.
+    #[prost(enumeration = "MaterializationType", tag = "3")]
+    pub r#type: i32,
+    /// The results of validating the projections of the materialization. A connector must include a
+    /// ProjectionResult for every projection in the materialization from the request. It is an error
+    /// to omit one.
+    /// TODO: we could just say that there's a default constraint (FIELD_OPTIONAL?) if one is ommitted.
+    #[prost(message, repeated, tag = "4")]
+    pub projection_results: ::std::vec::Vec<ProjectionResult>,
+}
+// Below TransactionalStore and StreamStore are speculative sketches of what the runtime
+// implementations of the materialization protocol might look like. These were done just to have a
+// better idea of what the Connector protocol might look like more holistically. A few specific
+// things that are missing or TBD:
+// - Standardized encoding of values as tuples
+// - Accounting for partition keys of individual documents, and the key ranges for the current
+//   shard.
+// - Defining transactional semantics for StreamStore
+
+/// Returned in response to a StartMaterializationShard request to indicate whether it's ok for the
+/// materialization to start.
+#[derive(Clone, PartialEq, ::prost::Message)]
+pub struct MaterializationStartResponse {
+    /// Indicates whether we're ok to proceed. Any status besides READY will prevent the
+    /// materialization from starting.
+    #[prost(message, optional, tag = "1")]
+    pub status: ::std::option::Option<MaterializationStatus>,
+    /// The checkpoint that's been committed from a previous instance. This will be null if this is the
+    /// first time that the materialization is running, or if the remote system is not a transactional
+    /// store.
+    #[prost(message, optional, tag = "2")]
+    pub checkpoint: ::std::option::Option<super::consumer::Checkpoint>,
+}
+/// Status returned by all responses that read/write to the target system
+#[derive(Clone, PartialEq, ::prost::Message)]
+pub struct StoreStatus {
+    #[prost(enumeration = "store_status::Code", tag = "1")]
+    pub code: i32,
+    #[prost(string, tag = "2")]
+    pub error_message: std::string::String,
+}
+pub mod store_status {
+    #[derive(Clone, Copy, Debug, PartialEq, Eq, Hash, PartialOrd, Ord, ::prost::Enumeration)]
+    #[repr(i32)]
+    pub enum Code {
+        Ok = 0,
+        Error = 1,
+    }
+}
+/// Request to retrieve a set of documents from a transactional store.
+#[derive(Clone, PartialEq, ::prost::Message)]
+pub struct GetDocumentsRequest {
+    #[prost(bytes, tag = "1")]
+    pub arena: std::vec::Vec<u8>,
+    #[prost(message, repeated, tag = "2")]
+    pub keys: ::std::vec::Vec<get_documents_request::Key>,
+}
+pub mod get_documents_request {
+    #[derive(Clone, PartialEq, ::prost::Message)]
+    pub struct Key {
+        #[prost(message, repeated, tag = "1")]
+        pub key: ::std::vec::Vec<super::Field>,
+    }
+}
+/// TODO: think of a better way to represent missing documents?
+/// Response for retrieving documents from a transacional store. Each key from the request
+/// must have a corresponding document here and the results must be in the same order as they were in
+/// the request. Documents that do not exist must be represented here as Fields that point to a "null" json
+/// value within the arena.
+#[derive(Clone, PartialEq, ::prost::Message)]
+pub struct GetDocumentsResponse {
+    #[prost(message, optional, tag = "1")]
+    pub status: ::std::option::Option<StoreStatus>,
+    #[prost(bytes, tag = "2")]
+    pub arena: std::vec::Vec<u8>,
+    #[prost(message, repeated, tag = "3")]
+    pub documents: ::std::vec::Vec<Field>,
+}
+/// Request to either insert or update documents in a remote system. Each Document contains the
+/// complete set of extracted fields corresponding to the projections from the MaterializationSpec.
+/// This message is used for both inserts and updates, and the sets of fields will be exactly the
+/// same in either case.
+#[derive(Clone, PartialEq, ::prost::Message)]
+pub struct UpdateDocumentsRequest {
+    #[prost(bytes, tag = "1")]
+    pub arena: std::vec::Vec<u8>,
+    #[prost(message, repeated, tag = "2")]
+    pub documents: ::std::vec::Vec<update_documents_request::Document>,
+}
+pub mod update_documents_request {
+    #[derive(Clone, PartialEq, ::prost::Message)]
+    pub struct Document {
+        #[prost(message, repeated, tag = "1")]
+        pub fields: ::std::vec::Vec<super::Field>,
+    }
+}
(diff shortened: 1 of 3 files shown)


jgraettinger (changes requested):
> Reviewed 1 of 1 files at r1.
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/79)* status: 1 of 2 files reviewed, 7 unresolved discussions (waiting on @jgraettinger and @psFried)

___
*[go/protocol/flow.proto, line 592 at r1](https://reviewable.io/reviews/estuary/flow/79#-MO2SV4O9HRuaKTq0HK1:-MO2SV4O9HRuaKTq0HK2:b-qxkoxu) ([raw file](https://github.com/estuary/flow/blob/d21a96aa4ca5acb2c55b3868b29bf8c9b67fd526/go/protocol/flow.proto#L592)):*
> ```protobuf
>   rpc ClearRegisters(ClearRegistersRequest) returns (ClearRegistersResponse);
> }
> ```

There's enough here that I think we need to lift this out of `flow.proto` and into it's own protobuf `materialize` package, that depends on `flow.proto`.

It's motivated to do this now, rather than latter, because it deeply affects the naming of many of these messages -- e.x.`materialize.Status` rather than `flow.MaterializationStatus`, and `materialize.Constraint` rather than `flow.ProjectionConstraint`.

We can (and should!) use tighter, more focused names if we're already scoped to the domain at hand.
___
*[go/protocol/flow.proto, line 624 at r1](https://reviewable.io/reviews/estuary/flow/79#-MO2TlfIBTY9nfpoKUvw:-MO2TlfIBTY9nfpoKUvx:b-9423g3) ([raw file](https://github.com/estuary/flow/blob/d21a96aa4ca5acb2c55b3868b29bf8c9b67fd526/go/protocol/flow.proto#L624)):*
> ```protobuf
> 
> // Result of validating a specific projection.
> message ProjectionResult {
> ```

Wouldn't this be a ValidationResult ? Or better, seems like _this_ the `Constraint` and current `ProjectionConstraint` should be a sub-enum `Constraint.Code` ?
___
*[go/protocol/flow.proto, line 637 at r1](https://reviewable.io/reviews/estuary/flow/79#-MO2R7CO4Twd0W1STFAq:-MO2R7CO4Twd0W1STFAr:b-j1j1ik) ([raw file](https://github.com/estuary/flow/blob/d21a96aa4ca5acb2c55b3868b29bf8c9b67fd526/go/protocol/flow.proto#L637)):*
> ```protobuf
> // Describes the current state of a Materialization as stored in the remote system.
> message MaterializationStatus {
>   enum StatusCode {
> ```

nit: just `enum Code`, as this will stutter otherwise (MaterializationStatus.StatusCode).
___
*[go/protocol/flow.proto, line 654 at r1](https://reviewable.io/reviews/estuary/flow/79#-MO2RVEL2Czr3RMDpD5S:-MO2RVEL2Czr3RMDpD5T:b-2hrw2h) ([raw file](https://github.com/estuary/flow/blob/d21a96aa4ca5acb2c55b3868b29bf8c9b67fd526/go/protocol/flow.proto#L654)):*
> ```protobuf
> // a given type of target system. For example, most databases can only be a TRANSACTIONAL_STORE, and
> // most pub-sub systems can only be a STREAM.
> enum MaterializationType {
> ```

I think we should side-step attempting to define this concept "grain". I don't think we can realistically pin down the varieties of types that may exist, and _we don't need to_.

The salient detail now is merely "should I bother to call fetch() or not?" -- let's just plumb that through, rather than forcing materializations to define themselves.

I'm thinking a no-argument, unary RPC called at runtime that returns a Settings message, or something.
___
*[go/protocol/flow.proto, line 672 at r1](https://reviewable.io/reviews/estuary/flow/79#-MO2VO7A3vBEBHCR6_QS:-MO2VO7A3vBEBHCR6_QT:bs89zc1) ([raw file](https://github.com/estuary/flow/blob/d21a96aa4ca5acb2c55b3868b29bf8c9b67fd526/go/protocol/flow.proto#L672)):*
> ```protobuf
>   // system, in addition to the spec provided in a request.
>   MaterializationStatus status = 1 [ (gogoproto.nullable) = false ];
>   // The initialization payload is a system-specific text that can be used to initialize the
> ```

There isn't any action that Flow can take with this, aside from presenting it to the user. While it may hold an initialization string, what this really _is_ is a driver-generated, user-presentable message. We should define it here as such, and think through how it fits into the overall build & apply workflow. That it happens to hold a CREATE TABLE statement they can copy-paste is an implementation choice of the driver.
___
*[go/protocol/flow.proto, line 688 at r1](https://reviewable.io/reviews/estuary/flow/79#-MO2U9Ax0Dhdpdbu83MP:-MO2U9Ax0Dhdpdbu83MQ:bswokr2) ([raw file](https://github.com/estuary/flow/blob/d21a96aa4ca5acb2c55b3868b29bf8c9b67fd526/go/protocol/flow.proto#L688)):*
> ```protobuf
> }
> 
> service Connector {
> ```

I'd rather not introduce a formal "Connector" concept. There may happen to be binaries that service both ingestion and materialization (though I'm skeptical), but they would provide both ends as separate service implementations over the same gRPC connection.

And again, as part of a materialization-scoped service definition, these method names can be simply `Validate` and `Apply`.
___
*[go/protocol/flow.proto, line 722 at r1](https://reviewable.io/reviews/estuary/flow/79#-MO2WQR7AY6xYtz7UZV2:-MO2WQR877JBYX8ovqSf:big2u37) ([raw file](https://github.com/estuary/flow/blob/d21a96aa4ca5acb2c55b3868b29bf8c9b67fd526/go/protocol/flow.proto#L722)):*
> ```protobuf
> 
> }
> ```

As per above, I don't think these are separate services but rather one service that's able to provide details to the runtime about how to optimally drive it

psFried (commented):
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/79)* status: 1 of 2 files reviewed, 7 unresolved discussions (waiting on @jgraettinger)

___
*[go/protocol/flow.proto, line 592 at r1](https://reviewable.io/reviews/estuary/flow/79#-MO2SV4O9HRuaKTq0HK1:-MO2bwUv-h8aqcjVJ_yP:bo6x7f9) ([raw file](https://github.com/estuary/flow/blob/d21a96aa4ca5acb2c55b3868b29bf8c9b67fd526/go/protocol/flow.proto#L592)):*

There's enough here that I think we need to lift this out of `flow.proto` and into it's own protobuf `materialize` package, that depends on `flow.proto`.

It's motivated to do this now, rather than latter, because it deeply affects the naming of many of these messages -- e.x.`materialize.Status` rather than `flow.MaterializationStatus`, and `materialize.Constraint` rather than `flow.ProjectionConstraint`.

We can (and should!) use tighter, more focused names if we're already scoped to the domain at hand.
</blockquote></details>

agreed regarding names. No need to repeat `Materialization` everywhere if you're within a `materialize` package.
___
*[go/protocol/flow.proto, line 624 at r1](https://reviewable.io/reviews/estuary/flow/79#-MO2TlfIBTY9nfpoKUvw:-MO2l-Wb7K_r11G_zI57:b8pfjbr) ([raw file](https://github.com/estuary/flow/blob/d21a96aa4ca5acb2c55b3868b29bf8c9b67fd526/go/protocol/flow.proto#L624)):*

Wouldn't this be a ValidationResult ? Or better, seems like _this_ the `Constraint` and current `ProjectionConstraint` should be a sub-enum `Constraint.Code` ?
</blockquote></details>

Good suggestions :+1: 
___
*[go/protocol/flow.proto, line 637 at r1](https://reviewable.io/reviews/estuary/flow/79#-MO2R7CO4Twd0W1STFAq:-MO2l9o767fSg92EdBE7:b-hu3qx) ([raw file](https://github.com/estuary/flow/blob/d21a96aa4ca5acb2c55b3868b29bf8c9b67fd526/go/protocol/flow.proto#L637)):*

nit: just `enum Code`, as this will stutter otherwise (MaterializationStatus.StatusCode).
</blockquote></details>

:+1:  will do!
___
*[go/protocol/flow.proto, line 654 at r1](https://reviewable.io/reviews/estuary/flow/79#-MO2RVEL2Czr3RMDpD5S:-MO2Z4Lz1rPybysXk2KG:b-728bj6) ([raw file](https://github.com/estuary/flow/blob/d21a96aa4ca5acb2c55b3868b29bf8c9b67fd526/go/protocol/flow.proto#L654)):*

I think we should side-step attempting to define this concept "grain". I don't think we can realistically pin down the varieties of types that may exist, and _we don't need to_.

The salient detail now is merely "should I bother to call fetch() or not?" -- let's just plumb that through, rather than forcing materializations to define themselves.

I'm thinking a no-argument, unary RPC called at runtime that returns a Settings message, or something.
</blockquote></details>

I think I disagree on this one...
We don't need to pin down the varieties that may exist, but we absolutely need to pin down the APIs that we use and how we'll use them. This Type is merely meant to encapsulate that. Having separate Service definitions per type is meant to make it clear and unambiguous which functions you need to implement and how they'll be called. I think a Settings message is basically just another way of representing the same thing. In any case we have some value that's returned from the service itself that tells us how we can interact with that API at runtime. The problem I see with Settings is that they can become confusing and error prone as they evolve, since only certain combinations of settings are valid. Having a one-to-one mapping between Type and Service makes it obvious which functions you need to implement for each type. It also allows us to define future services that may be totally different from those we have today.

The other implication currently is that a transactional_store is expected to also store and retrieve the consumer offsets. The motivation for hoisting this to a MaterializationType was to have a single value that determines how flow will use the service (whether it will retrieve documents and whether it will commit checkpoints). I'm thinking of a possible third type of system that is a "store" in the sense that it will hold fully reduced values by key, but that lacks the transactional support for storing the Checkpoint. In such a system, we could have flow manage the storage of the fully-reduced flow documents and checkpoints, and only store the other projected fields in the external system. My intuition is that most REST-based services would fall into that category.

More generally, I think there's also a benefit to having the service classify itself because it allows us to evolve the API in a forward compatible way. Though I could see an argument for up-leveling this particular concern, it seems like something we might want to know at build/apply time, since that's likely when we'd need to deal with any compatibility concerns.

Basically, the goal here is to plan for the ability to evolve the API over time. But let me know if that's unconvincing or if there's another problem with it, and we can maybe talk through some other possible solutions.
___
*[go/protocol/flow.proto, line 672 at r1](https://reviewable.io/reviews/estuary/flow/79#-MO2VO7A3vBEBHCR6_QS:-MO2iXp9BbYRKTPYfqgr:boz9vkx) ([raw file](https://github.com/estuary/flow/blob/d21a96aa4ca5acb2c55b3868b29bf8c9b67fd526/go/protocol/flow.proto#L672)):*

There isn't any action that Flow can take with this, aside from presenting it to the user. While it may hold an initialization string, what this really _is_ is a driver-generated, user-presentable message. We should define it here as such, and think through how it fits into the overall build & apply workflow. That it happens to hold a CREATE TABLE statement they can copy-paste is an implementation choice of the driver.
</blockquote></details>

Yeah, I agree with all that. The name of `initialization_payload` was chosen merely to convey the _intent_ for how we'll use it, but ultimately the driver can put whatever it wants in there. Happy to change the name if you have a suggestion, or I can just update docs to clarify all that. In terms of what we do with it apart from just showing it, I was thinking we might want to persist it in the catalog, so that we could support workflows where people need to manually apply this stuff without making them select/copy from their terminal. But I think that's probably about as far as we'd need to go, and either way definitely treating this as an opaque blob from Flow's perspective.
___
*[go/protocol/flow.proto, line 688 at r1](https://reviewable.io/reviews/estuary/flow/79#-MO2U9Ax0Dhdpdbu83MP:-MO2kQN1DGA1kTj0-MKK:b-gjxrpu) ([raw file](https://github.com/estuary/flow/blob/d21a96aa4ca5acb2c55b3868b29bf8c9b67fd526/go/protocol/flow.proto#L688)):*

I'd rather not introduce a formal "Connector" concept. There may happen to be binaries that service both ingestion and materialization (though I'm skeptical), but they would provide both ends as separate service implementations over the same gRPC connection.

And again, as part of a materialization-scoped service definition, these method names can be simply `Validate` and `Apply`.
</blockquote></details>

Fair. Maybe MaterializationBuild would be a better service name?
___
*[go/protocol/flow.proto, line 722 at r1](https://reviewable.io/reviews/estuary/flow/79#-MO2WQR7AY6xYtz7UZV2:-MO2kdQu58b4iAiYeC0B:bhtdy2z) ([raw file](https://github.com/estuary/flow/blob/d21a96aa4ca5acb2c55b3868b29bf8c9b67fd526/go/protocol/flow.proto#L722)):*

As per above, I don't think these are separate services but rather one service that's able to provide details to the runtime about how to optimally drive it
</blockquote></details>

Yeah let's hash that out in the other comment thread

jgraettinger (changes requested):
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/79)* status: 1 of 2 files reviewed, 7 unresolved discussions (waiting on @jgraettinger and @psFried)

___
*[go/protocol/flow.proto, line 654 at r1](https://reviewable.io/reviews/estuary/flow/79#-MO2RVEL2Czr3RMDpD5S:-MO2pZtQAExZoMuxnHGb:b-o6h0sf) ([raw file](https://github.com/estuary/flow/blob/d21a96aa4ca5acb2c55b3868b29bf8c9b67fd526/go/protocol/flow.proto#L654)):*

I think I disagree on this one...
We don't need to pin down the varieties that may exist, but we absolutely need to pin down the APIs that we use and how we'll use them. This Type is merely meant to encapsulate that. Having separate Service definitions per type is meant to make it clear and unambiguous which functions you need to implement and how they'll be called. I think a Settings message is basically just another way of representing the same thing. In any case we have some value that's returned from the service itself that tells us how we can interact with that API at runtime. The problem I see with Settings is that they can become confusing and error prone as they evolve, since only certain combinations of settings are valid. Having a one-to-one mapping between Type and Service makes it obvious which functions you need to implement for each type. It also allows us to define future services that may be totally different from those we have today.

The other implication currently is that a transactional_store is expected to also store and retrieve the consumer offsets. The motivation for hoisting this to a MaterializationType was to have a single value that determines how flow will use the service (whether it will retrieve documents and whether it will commit checkpoints). I'm thinking of a possible third type of system that is a "store" in the sense that it will hold fully reduced values by key, but that lacks the transactional support for storing the Checkpoint. In such a system, we could have flow manage the storage of the fully-reduced flow documents and checkpoints, and only store the other projected fields in the external system. My intuition is that most REST-based services would fall into that category.

More generally, I think there's also a benefit to having the service classify itself because it allows us to evolve the API in a forward compatible way. Though I could see an argument for up-leveling this particular concern, it seems like something we might want to know at build/apply time, since that's likely when we'd need to deal with any compatibility concerns.

Basically, the goal here is to plan for the ability to evolve the API over time. But let me know if that's unconvincing or if there's another problem with it, and we can maybe talk through some other possible solutions.
</blockquote></details>

Let me put pen to paper and work up the design I have in mind, and we can talk about them comparatively.

psFried:
Closing this in favor of #82

PR #80, opened 2020-12-08
Switch over to CGO extractor & FDB tuples

jgraettinger:
(This builds on the cgo PR, and isn't appropriate for review until that lands)

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/80)


jgraettinger:
Next up for review. This has been rebased on the just-merged cgo work.

jgraettinger:
I've rebased to fix the current `master` build break.

psFried (commented):
> Reviewed 7 of 7 files at r1, 5 of 5 files at r2.
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/80)* status: 11 of 34 files reviewed, 1 unresolved discussion (waiting on @jgraettinger and @psFried)

___
*[crates/tuple/src/value.rs, line 19 at r1](https://reviewable.io/reviews/estuary/flow/80#-MOXGH---f_raBS88wRv:-MOXGH-0CrCPx72jouEL:btj6ayp) ([raw file](https://github.com/estuary/flow/blob/00e60916a3c2778557a2b27a9b7b7541c34e7565/crates/tuple/src/value.rs#L19)):*
> ```Rust
>                 } else if let Some(n) = n.as_i64() {
>                     n.pack(w, tuple_depth)
>                 } else {
> ```

Wondering if we should handle the case where the number isn't representable as an f64, or if we should just punt on that until we think about numeric precision more holistically. Either way is fine by me.

psFried (approved):
> Reviewed 9 of 9 files at r3, 11 of 11 files at r4, 3 of 3 files at r5.
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/80)* status: all files reviewed, 4 unresolved discussions (waiting on @jgraettinger)

___
*[crates/protocol/src/cgo.rs, line 63 at r3](https://reviewable.io/reviews/estuary/flow/80#-MOXIHeSDhuBuCG9c2Ul:-MOXIHeSDhuBuCG9c2Um:b-z0of8u) ([raw file](https://github.com/estuary/flow/blob/ff8f43a193664448d458664ba27d773c84766f52/crates/protocol/src/cgo.rs#L63)):*
> ```Rust
>     /// |arena| beginning at arena offset |begin|. An Out frame that references
>     /// the arena span (through the current arena length) is pushed.
>     fn send_bytes(code: u32, begin: usize, arena: &mut Vec<u8>, out: &mut Vec<Out>) {
> ```

nit:  To me `send` implies that this is actually being written somewhere. Maybe `append_bytes`?
___
*[crates/protocol/src/flow.rs, line 382 at r4](https://reviewable.io/reviews/estuary/flow/80#-MOXJ0c4-R_ez_QWlQat:-MOXJ0c52Ak3PMrstb7J:b-x4dn0l) ([raw file](https://github.com/estuary/flow/blob/e4c74aa5f6c10615e753d8b11fbc3b909ec14581/crates/protocol/src/flow.rs#L382)):*
> ```Rust
>     pub packed_key: ::std::vec::Vec<Slice>,
>     /// Extracted shuffle key of each document, with one Field for each
>     /// component of the composite shuffle key.
> ```

Not sure what this change is about. Do we even need this file anymore?
___
*[go/runtime/materialize.go, line 246 at r5](https://reviewable.io/reviews/estuary/flow/80#-MOXKF6pAm6lyGDwYu3t:-MOXKF6pAm6lyGDwYu3u:b8x8nm9) ([raw file](https://github.com/estuary/flow/blob/392df3de360a6a01e9dc2f875f1c2e0fdb02b28d/go/runtime/materialize.go#L246)):*
> ```Go
> 	// database) and add that to the Combine. The "packed" shuffle key is used to key the cache
> 	// and map of observed documents.
> 	// NOTE: use string(key) to avoid allocation if the map key already exists.
> ```

I'm curious about this, and my google fu was unsuccessful. Does go have some special optimization for this case?

jgraettinger (commented):
Thanks!

> *[Reviewable](https://reviewable.io/reviews/estuary/flow/80#-:-MOmaEOE4oJyMrdU8Dj2:b-ucfknh)* status: all files reviewed, 4 unresolved discussions (waiting on @psFried)

___
*[crates/protocol/src/cgo.rs, line 63 at r3](https://reviewable.io/reviews/estuary/flow/80#-MOXIHeSDhuBuCG9c2Ul:-MOm_NV57_6SEQUARSt4:b8erasl) ([raw file](https://github.com/estuary/flow/blob/ff8f43a193664448d458664ba27d773c84766f52/crates/protocol/src/cgo.rs#L63)):*

nit:  To me `send` implies that this is actually being written somewhere. Maybe `append_bytes`?
</blockquote></details>

well, this is intended to be highly evocative of a bidirectional, streaming interaction. Users of this are thinking in terms of sending messages, not appending to buffers.
___
*[crates/protocol/src/flow.rs, line 382 at r4](https://reviewable.io/reviews/estuary/flow/80#-MOXJ0c4-R_ez_QWlQat:-MOm_bSV3Q7GoDtAHM5E:b4vcm11) ([raw file](https://github.com/estuary/flow/blob/e4c74aa5f6c10615e753d8b11fbc3b909ec14581/crates/protocol/src/flow.rs#L382)):*

Not sure what this change is about. Do we even need this file anymore?
</blockquote></details>

yes, we do. Basically we're no longer sending the extracted Field components in the shuffle response, because that's duplicative with the packed tuples.
___
*[crates/tuple/src/value.rs, line 19 at r1](https://reviewable.io/reviews/estuary/flow/80#-MOXGH---f_raBS88wRv:-MOm_nda7uCOFPqQ2b2-:ba0yng4) ([raw file](https://github.com/estuary/flow/blob/00e60916a3c2778557a2b27a9b7b7541c34e7565/crates/tuple/src/value.rs#L19)):*

Wondering if we should handle the case where the number isn't representable as an f64, or if we should just punt on that until we think about numeric precision more holistically. Either way is fine by me.
</blockquote></details>

Well, f64 is the only remaining possibility here. serde_json::Value is one of u64, i64, or f64.

Representing integer types > 64 bits requires falling back to string encoding.
___
*[go/runtime/materialize.go, line 246 at r5](https://reviewable.io/reviews/estuary/flow/80#-MOXKF6pAm6lyGDwYu3t:-MOma3nh477hKuH923Uq:bvxze5m) ([raw file](https://github.com/estuary/flow/blob/392df3de360a6a01e9dc2f875f1c2e0fdb02b28d/go/runtime/materialize.go#L246)):*

I'm curious about this, and my google fu was unsuccessful. Does go have some special optimization for this case?
</blockquote></details>

Yep, sure does. "Go escape analysis" is the google term to use. There are some build time flags you can pass to get it to spit out decisions it's making about when things escape to heap, etc.

PR #85, opened 2020-12-24
cgo bridge all the things (remove flow-worker)

jgraettinger:
* Switch over combine & derive API's to use the CGO bridge.
* Integrate register RocksDB recovery log recording.
* Remove lots of stuff.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/85)


psFried (changes requested):
> Reviewed 3 of 3 files at r1, 7 of 7 files at r2, 16 of 16 files at r3, 8 of 8 files at r4, 3 of 3 files at r5, 8 of 8 files at r6, 14 of 14 files at r7, 7 of 7 files at r8, 6 of 6 files at r9.
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/85)* status: all files reviewed, 5 unresolved discussions (waiting on @jgraettinger)

___
*[Makefile, line 168 at r9](https://reviewable.io/reviews/estuary/flow/85#-MPkaclt3cWlHWBf_CNE:-MPkaclt3cWlHWBf_CNF:b-x4d1k9) ([raw file](https://github.com/estuary/flow/blob/0651fb482411de118601367191520d1e26a74633/Makefile#L168)):*
> ```Makefile
> 	tar xzf ${WORKDIR}/tmp.tgz -C ${WORKDIR}/rocksdb-v$* --strip-components=1
> 	rm ${WORKDIR}/tmp.tgz
> 	# TODO(johnny): We should remove PORTABLE, and instead restrict CI to compatible hardware.
> ```

I looked into that at one point and found that github actions don't give you that kind of control unless you use self-hosted runners, which is problematic for public repositories.
___
*[Makefile, line 200 at r9](https://reviewable.io/reviews/estuary/flow/85#-MPkbAgZDM0C6h0Izs-A:-MPkbAgZDM0C6h0Izs-B:b7arimt) ([raw file](https://github.com/estuary/flow/blob/0651fb482411de118601367191520d1e26a74633/Makefile#L200)):*
> ```Makefile
> 
> ${RUSTBIN}: ${ROCKSDIR}/${LIBROCKS}
> 	FLOW_VERSION=${VERSION} cargo build --release --locked
> ```

good idea
___
*[crates/bindings/src/service.rs, line 160 at r4](https://reviewable.io/reviews/estuary/flow/85#-MPkRg-r-2rvRTTlf8LV:-MPkRg-r-2rvRTTlf8LW:b-uem4qs) ([raw file](https://github.com/estuary/flow/blob/e3655435ac3a926451c5d19e5798c5d29c1e2b5b/crates/bindings/src/service.rs#L160)):*
> ```Rust
>     if let Err(err) = r {
>         // Set terminal error string.
>         err_str = format!("{:?}", anyhow::Error::new(err));
> ```

Does this just provide a better formatting?
___
*[crates/derive/src/combine_api.rs, line 119 at r3](https://reviewable.io/reviews/estuary/flow/85#-MPelUin0pfYpqDK3bjd:-MPelUin0pfYpqDK3bje:b-y8ffcy) ([raw file](https://github.com/estuary/flow/blob/48a8436358f78b2ef62c4bb03a62431de27507a5/crates/derive/src/combine_api.rs#L119)):*
> ```Rust
>         // Send serialized document.
>         let begin = arena.len();
>         serde_json::to_writer(&mut *arena, &doc).expect("encoding cannot fail");
> ```

Nit: this "re-borrow" pattern is ambiguous here because it could either result in a mutable slice or a mutable reference to a vec (which `arena` already is). Not sure what the rules are for deciding which one wins, but I think we definitely want just the mutable reference to the vec since it'll allow it to grow if needed. So IMO just get rid of the `&mut *` and pass the plain `arena`.
___
*[crates/derive/src/derive_api.rs, line 130 at r3](https://reviewable.io/reviews/estuary/flow/85#-MPetMlfBhIV0iZG6O89:-MPetMlfBhIV0iZG6O8A:b-5w2ff4) ([raw file](https://github.com/estuary/flow/blob/48a8436358f78b2ef62c4bb03a62431de27507a5/crates/derive/src/derive_api.rs#L130)):*
> ```Rust
> }
> 
> struct Pimpl<'e> {
> ```

:laughing: I guess this naming is just to reflect that this implementation detail is considered a bit unfortunate?
___
*[crates/derive/src/derive_api.rs, line 159 at r3](https://reviewable.io/reviews/estuary/flow/85#-MPev36P730zBQm5AwJ7:-MPev36P730zBQm5AwJ8:bm2z10p) ([raw file](https://github.com/estuary/flow/blob/48a8436358f78b2ef62c4bb03a62431de27507a5/crates/derive/src/derive_api.rs#L159)):*
> ```Rust
> impl<'e> Pimpl<'e> {
>     fn build_runtime() -> tokio::runtime::Runtime {
>         tokio::runtime::Builder::new()
> ```

My gut reaction here is that it would be preferable for all `Pimpl` instances to share references to a single global `Runtime`. Maybe use `lazy_static!` to initialize the tokio runtime and then have `Pimpl` just store a `Handle`? Fine to leave it as is and change that in a later, PR, though.
___
*[crates/derive/src/derive_api.rs, line 483 at r3](https://reviewable.io/reviews/estuary/flow/85#-MPk6YcTA7aBMvx88KzO:-MPk6YcUA43aT4SbuZUQ:b-jy1p53) ([raw file](https://github.com/estuary/flow/blob/48a8436358f78b2ef62c4bb03a62431de27507a5/crates/derive/src/derive_api.rs#L483)):*
> ```Rust
> 
> // Process a continuation block of the derivation's source documents.
> #[tracing::instrument(level = "debug", name = "block", err, skip(block, registers, combiner))]
> ```

I just read the docs on the instrument macro and `err` looks awesome!
___
*[crates/derive/src/derive_api.rs, line 501 at r3](https://reviewable.io/reviews/estuary/flow/85#-MPkGKso9ri1O3Mpl7EM:-MPkGKso9ri1O3Mpl7EN:bjg6cob) ([raw file](https://github.com/estuary/flow/blob/48a8436358f78b2ef62c4bb03a62431de27507a5/crates/derive/src/derive_api.rs#L501)):*
> ```Rust
>         .collect::<Vec<_>>();
> 
>     let b_open_open = bytes::Bytes::from(&b"[["[..]);
> ```

nit: `Bytes` impls `From<&'static str>` so you should be able to use `Bytes::from("[[")` for these.
___
*[crates/derive/src/derive_api.rs, line 521 at r3](https://reviewable.io/reviews/estuary/flow/85#-MPkHmp449w22r4YRbUu:-MPkHmp449w22r4YRbUv:b9w17t3) ([raw file](https://github.com/estuary/flow/blob/48a8436358f78b2ef62c4bb03a62431de27507a5/crates/derive/src/derive_api.rs#L521)):*
> ```Rust
>                 let body: Vec<Result<_, std::convert::Infallible>> = vec>                     Ok(b_open_open.clone()),
>                     Ok(source_column.clone()),
> ```

nit: seems like you could do the `.clone()` in the `trace!` instead of here to avoid copying the source column in the common case.
___
*[crates/derive/src/lib.rs, line 15 at r3 ([raw file](https://github.com/estuary/flow/blob/48a8436358f78b2ef62c4bb03a62431de27507a5/crates/derive/src/lib.rs#L15)):*
> ```Rust
> 
> // TODO(johnny): Move to a common `instrument` crate?
> /// Setup a global racing subscriber using the RUST_LOG env variable.
> ```

I think a common runtime instrumentation crate sounds like a good idea, but can wait until whenever we need it.
Also s/racing/tracing
___
*[crates/derive/src/lib.rs, line 40 at r3](https://reviewable.io/reviews/estuary/flow/85#-MPkOpV73j_wTNJPlvfT:-MPkOpV73j_wTNJPlvfU:b-b80r72) ([raw file](https://github.com/estuary/flow/blob/48a8436358f78b2ef62c4bb03a62431de27507a5/crates/derive/src/lib.rs#L40)):*
> ```Rust
> pub fn build_schema_index(
>     db: &catalog::DB,
> ) -> Result<&'static doc::SchemaIndex<'static>, catalog::Error> {
> ```

Not something I think needs addressed right now, but I keep thinking that we need to make SchemaIndex use `Cow` so it can own schemas directly. Then it's trivial to get a `SchemaIndex<'static>` without leaking anything.
___
*[crates/derive/src/registers.rs, line 178 at r3](https://reviewable.io/reviews/estuary/flow/85#-MPkPd92EnL-wR3mwrAc:-MPkPd93D4nJiMCMAiMi:b-ncq34z) ([raw file](https://github.com/estuary/flow/blob/48a8436358f78b2ef62c4bb03a62431de27507a5/crates/derive/src/registers.rs#L178)):*
> ```Rust
> 
> // Checkpoint key is the key encoding under which a marshalled checkpoint is stored.
> pub const CHECKPOINT_KEY: &[u8] = b"\0\0\0\0checkpoint";
> ```

wat? Probably a good candidate for a comment. Are these null bytes are reserving space for something?
___
*[crates/doc/src/ptr.rs, line 292 at r3](https://reviewable.io/reviews/estuary/flow/85#-MPkQRkZEZiefniYBMKW:-MPkQRk_5hWzG3uf_X1w:bnb3qvg) ([raw file](https://github.com/estuary/flow/blob/48a8436358f78b2ef62c4bb03a62431de27507a5/crates/doc/src/ptr.rs#L292)):*
> ```Rust
>         for tok in self.iter() {
>             match tok {
>                 Token::Index(ind) => f.write_fmt(format_args!("/{}", ind))?,
> ```

nit: you can also `write!(f, "/{}", ind)?`
___
*[go/fdb/tuple/tuple_iface.go, line 5 at r7](https://reviewable.io/reviews/estuary/flow/85#-MPkUKTj48HDDkcCoWB0:-MPkUKTj48HDDkcCoWB1:brt86b3) ([raw file](https://github.com/estuary/flow/blob/6fc13b4e59262d7b319c81d6c4a91ff1af2ead11/go/fdb/tuple/tuple_iface.go#L5)):*
> ```Go
> // ToInterface maps a Tuple to a []interface{}.
> // This is useful within interfaces that take []interface{} splats,
> // like SQL drivers.
> ```

Indeed! I wrote one of these, too. I'll clean up mine after rebasing, since I'm sure this will merge first.
___
*[go/flow/mapping.go, line 54 at r7](https://reviewable.io/reviews/estuary/flow/85#-MPkWZS64TvdYGsoHAle:-MPkWZS64TvdYGsoHAlf:bagbevh) ([raw file](https://github.com/estuary/flow/blob/6fc13b4e59262d7b319c81d6c4a91ff1af2ead11/go/flow/mapping.go#L54)):*
> ```Go
> }
> 
> // Map the Mappable, which must be an IndexedCombineResponse, into a physical journal partition
> ```

looks like this comment needs updated
___
*[go/bindings/upper_case_test.go, line 119 at r6](https://reviewable.io/reviews/estuary/flow/85#-MPkTyjp8jZ6PDaWWtDZ:-MPkTyjp8jZ6PDaWWtD_:bun1eqv) ([raw file](https://github.com/estuary/flow/blob/5b901b62ed4eb1894bd5e2ba1948187a2f182189/go/bindings/upper_case_test.go#L119)):*
> ```Go
> 	assert.Equal(t, expect, data)
> 
> 	_, _, err = svc.invoke(789, []byte("whoops"))
> ```

wouldn't want to surprise anyone with that ! :stuck_out_tongue_winking_eye:

jgraettinger (commented):
Thanks for slogging through this. Response / updates below, PTAL

> *[Reviewable](https://reviewable.io/reviews/estuary/flow/85#-:-MPllZCx3eBYJqdoKhG5:bgc7k4k)* status: 20 of 71 files reviewed, 3 unresolved discussions (waiting on @psFried)

___
*[Makefile, line 168 at r9](https://reviewable.io/reviews/estuary/flow/85#-MPkaclt3cWlHWBf_CNE:-MPle2RV69VmUkxfd_2T:biay1rk) ([raw file](https://github.com/estuary/flow/blob/0651fb482411de118601367191520d1e26a74633/Makefile#L168)):*

I looked into that at one point and found that github actions don't give you that kind of control unless you use self-hosted runners, which is problematic for public repositories.
</blockquote></details>

understood
___
*[Makefile, line 200 at r9](https://reviewable.io/reviews/estuary/flow/85#-MPkbAgZDM0C6h0Izs-A:-MPleB6cDZpgVAgRyCeV:bsog8fe) ([raw file](https://github.com/estuary/flow/blob/0651fb482411de118601367191520d1e26a74633/Makefile#L200)):*

good idea
</blockquote></details>

also required, now, in order to plumb through the ROCKSDB_LIB_DIR environment variable. I wish there were a way to do that via Cargo...
___
*[crates/bindings/src/service.rs, line 160 at r4](https://reviewable.io/reviews/estuary/flow/85#-MPkRg-r-2rvRTTlf8LV:-MPleO8j0GYj_slj75Ck:b-1mmai1) ([raw file](https://github.com/estuary/flow/blob/e3655435ac3a926451c5d19e5798c5d29c1e2b5b/crates/bindings/src/service.rs#L160)):*

Does this just provide a better formatting?
</blockquote></details>

Yea, in particular it sends the complete error chain rather than the top-most error string.
___
*[crates/derive/src/combine_api.rs, line 119 at r3](https://reviewable.io/reviews/estuary/flow/85#-MPelUin0pfYpqDK3bjd:-MPlf0tv8k48nppf7F8c:b-jubjnm) ([raw file](https://github.com/estuary/flow/blob/48a8436358f78b2ef62c4bb03a62431de27507a5/crates/derive/src/combine_api.rs#L119)):*

Nit: this "re-borrow" pattern is ambiguous here because it could either result in a mutable slice or a mutable reference to a vec (which `arena` already is). Not sure what the rules are for deciding which one wins, but I think we definitely want just the mutable reference to the vec since it'll allow it to grow if needed. So IMO just get rid of the `&mut *` and pass the plain `arena`.
</blockquote></details>

yea, i'd tried that initially but the compiler considers it a move of the borrow.

i _believe_ the reason this form is allowed is that it's creating a new derivative borrow, and won't allow use of the parent borrow until it can prove that the child goes away (which it can prove).

kinda funky: serde_json::to_writer takes a io::Write by-value, while pack() below takes a &mut io::Write. The former has this issue but the latter doesn't.

Added an explicit typed let to at least remove the ambiguity. 
___
*[crates/derive/src/derive_api.rs, line 130 at r3](https://reviewable.io/reviews/estuary/flow/85#-MPetMlfBhIV0iZG6O89:-MPlhJdVAIWw89XeomJM:b-1hp318) ([raw file](https://github.com/estuary/flow/blob/48a8436358f78b2ef62c4bb03a62431de27507a5/crates/derive/src/derive_api.rs#L130)):*

:laughing: I guess this naming is just to reflect that this implementation detail is considered a bit unfortunate?
</blockquote></details>

Hah, this is in reference to the C++ pimpl pattern, which is what i was thinking of when i wrote it. I still don't have a better name for it. Switched to APIInner :confused: 

What i really wanted to do was `impl cgo::Service for Option<API<'static>>`, but you can't do that :(
___
*[crates/derive/src/derive_api.rs, line 159 at r3](https://reviewable.io/reviews/estuary/flow/85#-MPev36P730zBQm5AwJ7:-MPli8yA9w4bG3Kt-QL-:b-9dly6a) ([raw file](https://github.com/estuary/flow/blob/48a8436358f78b2ef62c4bb03a62431de27507a5/crates/derive/src/derive_api.rs#L159)):*

My gut reaction here is that it would be preferable for all `Pimpl` instances to share references to a single global `Runtime`. Maybe use `lazy_static!` to initialize the tokio runtime and then have `Pimpl` just store a `Handle`? Fine to leave it as is and change that in a later, PR, though.
</blockquote></details>

this choice was deliberate, actually. i think it's desirable to have each shard receive a single dedicated thread in which it's event loop runs.
___
*[crates/derive/src/derive_api.rs, line 501 at r3](https://reviewable.io/reviews/estuary/flow/85#-MPkGKso9ri1O3Mpl7EM:-MPlik4x9s77hOC5eZvd:bb7f2mf) ([raw file](https://github.com/estuary/flow/blob/48a8436358f78b2ef62c4bb03a62431de27507a5/crates/derive/src/derive_api.rs#L501)):*

nit: `Bytes` impls `From<&'static str>` so you should be able to use `Bytes::from("[[")` for these.
</blockquote></details>

done
___
*[crates/derive/src/derive_api.rs, line 521 at r3](https://reviewable.io/reviews/estuary/flow/85#-MPkHmp449w22r4YRbUu:-MPljC3b3nhY8KCvyTZk:b-enygms) ([raw file](https://github.com/estuary/flow/blob/48a8436358f78b2ef62c4bb03a62431de27507a5/crates/derive/src/derive_api.rs#L521)):*

nit: seems like you could do the `.clone()` in the `trace!` instead of here to avoid copying the source column in the common case.
</blockquote></details>

got a little clone() happy. These are cheap clones (just reference count adjustments), but removed several.
___
*[crates/derive/src/lib.rs, line 40 at r3](https://reviewable.io/reviews/estuary/flow/85#-MPkOpV73j_wTNJPlvfT:-MPljfSG552QF_gq_L2h:b-pb4j33) ([raw file](https://github.com/estuary/flow/blob/48a8436358f78b2ef62c4bb03a62431de27507a5/crates/derive/src/lib.rs#L40)):*

Not something I think needs addressed right now, but I keep thinking that we need to make SchemaIndex use `Cow` so it can own schemas directly. Then it's trivial to get a `SchemaIndex<'static>` without leaking anything.
</blockquote></details>

i don't believe it would have `static lifetime, then, but rather the imputed lifetime of the Cow? not sure that's any different than just borrowing directly.
___
*[crates/derive/src/registers.rs, line 178 at r3](https://reviewable.io/reviews/estuary/flow/85#-MPkPd92EnL-wR3mwrAc:-MPljtCx886yUXnnTFip:bi5js0p) ([raw file](https://github.com/estuary/flow/blob/48a8436358f78b2ef62c4bb03a62431de27507a5/crates/derive/src/registers.rs#L178)):*

wat? Probably a good candidate for a comment. Are these null bytes are reserving space for something?
</blockquote></details>

hmm. well, it's a nod to how `store_rocksdb` encodes it, which uses slighlty-different leading zero's in cockroach encoding to ensure it appears before any user values.

A big difference, though, is that we're storing all registers in a separate column family and checkpoint is the _only_ key in the main database. so this matters less. Removed.
___
*[crates/doc/src/ptr.rs, line 292 at r3](https://reviewable.io/reviews/estuary/flow/85#-MPkQRkZEZiefniYBMKW:-MPlkd707EQUkGuOa71Q:bb7f2mf) ([raw file](https://github.com/estuary/flow/blob/48a8436358f78b2ef62c4bb03a62431de27507a5/crates/doc/src/ptr.rs#L292)):*

nit: you can also `write!(f, "/{}", ind)?`
</blockquote></details>

done
___
*[go/flow/mapping.go, line 54 at r7](https://reviewable.io/reviews/estuary/flow/85#-MPkWZS64TvdYGsoHAle:-MPll-CADWG-Boh6IV5p:b-896fix) ([raw file](https://github.com/estuary/flow/blob/6fc13b4e59262d7b319c81d6c4a91ff1af2ead11/go/flow/mapping.go#L54)):*

looks like this comment needs updated
</blockquote></details>

Done.
___
*[go/bindings/upper_case_test.go, line 119 at r6](https://reviewable.io/reviews/estuary/flow/85#-MPkTyjp8jZ6PDaWWtDZ:-MPlkgYLDQxZuF8bF7B5:b-6fnnxo) ([raw file](https://github.com/estuary/flow/blob/5b901b62ed4eb1894bd5e2ba1948187a2f182189/go/bindings/upper_case_test.go#L119)):*

wouldn't want to surprise anyone with that ! :stuck_out_tongue_winking_eye: 
</blockquote></details>

hah this was cleaning up a go linter error (it's a no-no to end error messages with punctuation)

psFried (approved):
> Reviewed 51 of 51 files at r10, 8 of 8 files at r11, 3 of 3 files at r12, 8 of 8 files at r13, 14 of 14 files at r14, 7 of 7 files at r15, 6 of 6 files at r16.
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/85)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jgraettinger)

___
*[Makefile, line 200 at r9](https://reviewable.io/reviews/estuary/flow/85#-MPkbAgZDM0C6h0Izs-A:-MPo45F7-vWXqm-DLhiE:bx4rs1y) ([raw file](https://github.com/estuary/flow/blob/0651fb482411de118601367191520d1e26a74633/Makefile#L200)):*

also required, now, in order to plumb through the ROCKSDB_LIB_DIR environment variable. I wish there were a way to do that via Cargo...
</blockquote></details>

I was initially referring to the addition of the `--locked` argument, but now that's not showing up in the diff in reviewable (it is in GH though). Anyway, I'm not sure how that env var is being plumbed through now, but it is possible to read and set env variables in `build.rs`, for example: https://github.com/estuary/flow/blob/master/crates/flowctl/build.rs#L11
___
*[crates/derive/src/combine_api.rs, line 119 at r3](https://reviewable.io/reviews/estuary/flow/85#-MPelUin0pfYpqDK3bjd:-MPo5kSb78SilAN71Cm4:b-wjiee0) ([raw file](https://github.com/estuary/flow/blob/48a8436358f78b2ef62c4bb03a62431de27507a5/crates/derive/src/combine_api.rs#L119)):*

yea, i'd tried that initially but the compiler considers it a move of the borrow.

i _believe_ the reason this form is allowed is that it's creating a new derivative borrow, and won't allow use of the parent borrow until it can prove that the child goes away (which it can prove).

kinda funky: serde_json::to_writer takes a io::Write by-value, while pack() below takes a &mut io::Write. The former has this issue but the latter doesn't.

Added an explicit typed let to at least remove the ambiguity. 
</blockquote></details>

Yeah, this is definitely one of the hairy edges of rust. Somewhere in std there's an `impl<'a, W: Write> Write for &'a mut W`, which is supposed to allow passing an `&mut Vec<u8>` even to `to_writer`. But it seems like something is preventing that from working in this case, and it's probably not worth the time to dig into it. Extracting the variable is a fine solution.
___
*[crates/derive/src/derive_api.rs, line 159 at r3](https://reviewable.io/reviews/estuary/flow/85#-MPev36P730zBQm5AwJ7:-MPo71UGAC8aP1wj3PSk:b-sdx3if) ([raw file](https://github.com/estuary/flow/blob/48a8436358f78b2ef62c4bb03a62431de27507a5/crates/derive/src/derive_api.rs#L159)):*

this choice was deliberate, actually. i think it's desirable to have each shard receive a single dedicated thread in which it's event loop runs.
</blockquote></details>

OK. My own intuition is that it may be preferable to have them all share a single reactor thread and a pool of executor threads, but I really don't think it's worth worrying about at the moment. Easy to change later if we come up with a reason for it.
___
*[crates/derive/src/lib.rs, line 40 at r3](https://reviewable.io/reviews/estuary/flow/85#-MPkOpV73j_wTNJPlvfT:-MPoCLHM9NBQKU3Lm8I3:bub8vpu) ([raw file](https://github.com/estuary/flow/blob/48a8436358f78b2ef62c4bb03a62431de27507a5/crates/derive/src/lib.rs#L40)):*

i don't believe it would have `static lifetime, then, but rather the imputed lifetime of the Cow? not sure that's any different than just borrowing directly.
</blockquote></details>

If you guarantee that a Cow is owned, you can give it any lifetime you want, including static. Here's an example: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=639e327409cca58ac4d4e83b47b9dd20 
I don't think this needs to block the PR or anything, just something that's been a little todo in the back of my mind for a while.

jgraettinger (commented):
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/85)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jgraettinger)

___
*[Makefile, line 200 at r9](https://reviewable.io/reviews/estuary/flow/85#-MPkbAgZDM0C6h0Izs-A:-MPoNPJJAQjYhRQWu2W3:botegf5) ([raw file](https://github.com/estuary/flow/blob/0651fb482411de118601367191520d1e26a74633/Makefile#L200)):*

I was initially referring to the addition of the `--locked` argument, but now that's not showing up in the diff in reviewable (it is in GH though). Anyway, I'm not sure how that env var is being plumbed through now, but it is possible to read and set env variables in `build.rs`, for example: https://github.com/estuary/flow/blob/master/crates/flowctl/build.rs#L11
</blockquote></details>

The trouble is that variable is just set for that one crate, no? What's needed is a way to have _our_ crate provide a variable that's exposed to the build.rs of a third-party crate (librocksdb-sys).

Sorry for screwing up the commit history. Reviewable has been really good about rebased force-pushes in the past, not sure what happened here. I should have pushed as extra to-be-squashed commits.
___
*[crates/derive/src/lib.rs, line 40 at r3](https://reviewable.io/reviews/estuary/flow/85#-MPkOpV73j_wTNJPlvfT:-MPoNyRR5mayP7_5GSW9:bnbleh1) ([raw file](https://github.com/estuary/flow/blob/48a8436358f78b2ef62c4bb03a62431de27507a5/crates/derive/src/lib.rs#L40)):*

If you guarantee that a Cow is owned, you can give it any lifetime you want, including static. Here's an example: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=639e327409cca58ac4d4e83b47b9dd20 
I don't think this needs to block the PR or anything, just something that's been a little todo in the back of my mind for a while.
</blockquote></details>

Wouldn't that be the same thing as having the schema index deeply copy every schema that's indexed ? Keep in mind that schemas directly hold their children.

PR #86, opened 2021-01-04
Materialization service

psFried:
Adds an initial Driver service implementation for sqlite and postgres. This PR is pretty bare bones, and only includes the grpc server code and a basic end to end test. There's a few notable pieces that are not included here and will come in subsequent PRs:

- A generic driver tester binary
- The runnable Driver binaries
- Integration with flowctl during builds

This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/86)


jgraettinger (changes requested):
Okay, got through a bunch of this but not all. Sending over comments I do have. Thanks!

> Reviewed 5 of 5 files at r1, 10 of 10 files at r2, 2 of 8 files at r3.
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/86#-:-MQEGx4u0QbvlZxGWx0U:b-bah9mv)* status: 15 of 28 files reviewed, 19 unresolved discussions (waiting on @jgraettinger and @psFried)

___
*[go/materialize/driver/sql.go, line 55 at r3](https://reviewable.io/reviews/estuary/flow/86#-MQE1l0082hPU5ylBahi:-MQE1l0082hPU5ylBahj:b-kk3kld) ([raw file](https://github.com/estuary/flow/blob/d86b680e8b004de2ce4760f6fc7ab461c8e3f6ff/go/materialize/driver/sql.go#L55)):*
> ```Go
> 			{
> 				Name:    GAZETTE_CHECKPOINTS_NONCE_COLUMN,
> 				Comment: "This nonce is used to uniquely identify instances of a shard and prevent them from conflicting.",
> ```

nit: uniquely identify instances of a shard => identify unique process assignments of a shard
___
*[go/materialize/driver/sql.go, line 154 at r3](https://reviewable.io/reviews/estuary/flow/86#-MQE36VN0yojB-Z1MHzb:-MQE36VN0yojB-Z1MHzc:bf9hhrr) ([raw file](https://github.com/estuary/flow/blob/d86b680e8b004de2ce4760f6fc7ab461c8e3f6ff/go/materialize/driver/sql.go#L154)):*
> ```Go
> 		Uri:      req.EndpointUrl,
> 		Table:    req.Target,
> 		CallerId: req.CallerId,
> ```

My own mistake, but I'm beginning to think that "caller ID" was an error in naming, as opposed to "shard ID". The conceit was that driver authors wouldn't need to know about shards. I now don't think that's true, and therefore the name adds overhead without value.
___
*[go/materialize/driver/sql.go, line 254 at r3](https://reviewable.io/reviews/estuary/flow/86#-MQE4hjT0d5y5oq11s4A:-MQE4hjT0d5y5oq11s4B:bg9946n) ([raw file](https://github.com/estuary/flow/blob/d86b680e8b004de2ce4760f6fc7ab461c8e3f6ff/go/materialize/driver/sql.go#L254)):*
> ```Go
> 	}
> 
> 	insertMaterializationSpec, err := driver.SqlGen.DirectInsertStatement(flowMaterializationsTable, handle.Table, string(specJson))
> ```

couldn't this use SqlGen.InsertStatement(), and then immediately invoke it with params ?
___
*[go/materialize/driver/sql.go, line 258 at r3](https://reviewable.io/reviews/estuary/flow/86#-MQE4NZXC4KeNHiwWP4Z:-MQE4Hja7sajUXWo0UCA:b-reynzt) ([raw file](https://github.com/estuary/flow/blob/d86b680e8b004de2ce4760f6fc7ab461c8e3f6ff/go/materialize/driver/sql.go#L258)):*
> ```Go
> 		return nil, err
> 	}
> 	var tableComment = fmt.Sprintf("Holds the fully reduced materialization of the Flow collection %s", req.Collection.Name)
> ```

nit: strike "fully reduced" here? It would need to be defined otherwise, and really it's an implementation detail from the user's perspective.
___
*[go/materialize/driver/sql.go, line 400 at r3](https://reviewable.io/reviews/estuary/flow/86#-MQE6Zkn7WIGN6i5z0oM:-MQE6Zkn7WIGN6i5z0oN:bx0aao) ([raw file](https://github.com/estuary/flow/blob/d86b680e8b004de2ce4760f6fc7ab461c8e3f6ff/go/materialize/driver/sql.go#L400)):*
> ```Go
> 	var response = new(pm.LoadResponse)
> 
> 	// TODO:change materialize.proto to make doc slices embedded and not pointers
> ```

why not do this now?
___
*[go/materialize/driver/sql.go, line 409 at r3](https://reviewable.io/reviews/estuary/flow/86#-MQE6xVGCl-giko06xK_:-MQE6xVGCl-giko06xKa:bi1mglg) ([raw file](https://github.com/estuary/flow/blob/d86b680e8b004de2ce4760f6fc7ab461c8e3f6ff/go/materialize/driver/sql.go#L409)):*
> ```Go
> 		// Each tuple should hold the collection keys in the order provided by the Fields. This will
> 		// match the order in the generated sql query.
> 		var args = tupleToArgSlice(tuple)
> ```

there's a method for this now! :)
___
*[go/materialize/driver/sql.go, line 410 at r3](https://reviewable.io/reviews/estuary/flow/86#-MQE7FOq7s4SGytDyYn6:-MQE7FOq7s4SGytDyYn7:b1htc8t) ([raw file](https://github.com/estuary/flow/blob/d86b680e8b004de2ce4760f6fc7ab461c8e3f6ff/go/materialize/driver/sql.go#L410)):*
> ```Go
> 		// match the order in the generated sql query.
> 		var args = tupleToArgSlice(tuple)
> 		rows, err := stmt.QueryContext(ctx, args...)
> ```

This is giving up a really crucial optimization, which is that Load() has a bunch of keys in-hand and can ask for them all at once rather than sequentially.

It's this issue that IMO motivates looking beyond `driver/sql`. Of course for sqlite it doesn't matter, but if we're talking to postgres over the network, this can be a crushing performance bottleneck.

The [pgx](https://pkg.go.dev/github.com/jackc/pgx/v4#Batch) driver offers an API for building up batch queries.

Another option would be to query via `SELECT` with an inner join against a constant `VALUES` table representing the batch, pared with `INSERT ... VALUES ` and `UPDATE ... FROM` for Store.
___
*[go/materialize/driver/sql.go, line 576 at r3](https://reviewable.io/reviews/estuary/flow/86#-MQEE5Sa6L2hyiPCiyOy:-MQEE5Sa6L2hyiPCiyOz:b-rodqi0) ([raw file](https://github.com/estuary/flow/blob/d86b680e8b004de2ce4760f6fc7ab461c8e3f6ff/go/materialize/driver/sql.go#L576)):*
> ```Go
> 
> 				stmtType = "insert"
> 				result, err = insertStatement.ExecContext(ctx, sqlArgs...)
> ```

same performance concern here. This is executing all updates and inserts sequentially, waiting for the DB to ack each before moving on to the next.
___
*[go/materialize/driver/sql.go, line 595 at r3](https://reviewable.io/reviews/estuary/flow/86#-MQEGTRrA2uFuj03olKD:-MQEGTRs3EXuAm3v8T4E:b-6sy14w) ([raw file](https://github.com/estuary/flow/blob/d86b680e8b004de2ce4760f6fc7ab461c8e3f6ff/go/materialize/driver/sql.go#L595)):*
> ```Go
> }
> 
> func (driver *SqlDriver) persistRows(ctx context.Context, req *pm.StoreRequest_Continue, txn *storeTransactionStarted) error {
> ```

is this used (it doesn't appear so)?
___
*[go/materialize/driver/sql.go, line 632 at r3](https://reviewable.io/reviews/estuary/flow/86#-MQEGpUO1cJ-TMSKaJlu:-MQEGpUO1cJ-TMSKaJlv:bp7ify) ([raw file](https://github.com/estuary/flow/blob/d86b680e8b004de2ce4760f6fc7ab461c8e3f6ff/go/materialize/driver/sql.go#L632)):*
> ```Go
> }
> 
> func tupleToArgSlice(tuple tuple.Tuple) []interface{} {
> ```

nit: remove
___
*[go/materialize/driver/sql.go, line 640 at r3](https://reviewable.io/reviews/estuary/flow/86#-MQEGr51430dKgQs-_cH:-MQEGr51430dKgQs-_cI:bidlk08) ([raw file](https://github.com/estuary/flow/blob/d86b680e8b004de2ce4760f6fc7ab461c8e3f6ff/go/materialize/driver/sql.go#L640)):*
> ```Go
> }
> 
> func (driver *SqlDriver) getCachedSql(ctx context.Context, handle *Handle) (*cachedSql, error) {
> ```

Marking here as where I'm taking a reviewing break.
___
*[go/materialize/driver/sqlgen.go, line 40 at r2](https://reviewable.io/reviews/estuary/flow/86#-MQDkMxyBtpDQp8MTxPK:-MQDkMxyBtpDQp8MTxPL:b-he3wgr) ([raw file](https://github.com/estuary/flow/blob/b588d4d420b52b102d713310b7f94bb7a7ea9b70/go/materialize/driver/sqlgen.go#L40)):*
> ```Go
> 	// PrimaryKey is true if this column is the primary key, or if it is part of a composite key.
> 	PrimaryKey bool
> 	// Type is the application type of the data. This corresponds closely to JSON types. The
> ```

nit: grammar / sentence fragment.
___
*[go/materialize/driver/sqlgen.go, line 55 at r2](https://reviewable.io/reviews/estuary/flow/86#-MQDkkPb0w0YvIULuaWi:-MQDkkPb0w0YvIULuaWj:b-dg3xn0) ([raw file](https://github.com/estuary/flow/blob/b588d4d420b52b102d713310b7f94bb7a7ea9b70/go/materialize/driver/sqlgen.go#L55)):*
> ```Go
> 	// Optional Comment to add to create table statements.
> 	Comment string
> 	// The complete list of Columns that are part of this table. More specifically, this is the complete
> ```

nit: this comment repeats itself but is also unclear (what are "automatic" columns?).
___
*[go/materialize/driver/sqlgen.go, line 77 at r2](https://reviewable.io/reviews/estuary/flow/86#-MQDlB3I82JoeOIZJxmz:-MQDlB3I82JoeOIZJxn-:bmu2746) ([raw file](https://github.com/estuary/flow/blob/b588d4d420b52b102d713310b7f94bb7a7ea9b70/go/materialize/driver/sqlgen.go#L77)):*
> ```Go
> 	// QueryOnPrimaryKey generates a query that has a placeholder parameter for each primary key in
> 	// the order given in the table. Only selectColumns will be selected in the same order as
> 	// provided. An error should be returned if selectColumns is empty.
> ```

aren't we guaranteeing that `selectColumns` isn't empty, before calling into this interface ? it's weird to make this check part of the implementers responsibility.
___
*[go/materialize/driver/sqlgen.go, line 82 at r2](https://reviewable.io/reviews/estuary/flow/86#-MQDlTsw7UW19ABqkvMZ:-MQDlTsw7UW19ABqkvM_:b-rvhes4) ([raw file](https://github.com/estuary/flow/blob/b588d4d420b52b102d713310b7f94bb7a7ea9b70/go/materialize/driver/sqlgen.go#L82)):*
> ```Go
> 	// InsertStatement returns an insert statement for the given table that includes all columns.
> 	// The returns sql will have a parameter placeholder for every column. For most systems, this
> 	// can be a plain insert statement, not an upsert, since we only use the insert statement if a
> ```

nit: grammer. Do we require that placeholders be in a specific order (I'd assume yes)?

should guidance be that we _always_ want implementers to use plain insert statements, since we "know" which is needed and violation checks are a valuable runtime assertion ?
Put differently, what are the exceptions to "For most systems" from your comment ?
___
*[go/materialize/driver/sqlgen.go, line 86 at r2](https://reviewable.io/reviews/estuary/flow/86#-MQDlrfmAa9cdbZkS-oG:-MQDlrfmAa9cdbZkS-oH:b-f3hkcd) ([raw file](https://github.com/estuary/flow/blob/b588d4d420b52b102d713310b7f94bb7a7ea9b70/go/materialize/driver/sqlgen.go#L86)):*
> ```Go
> 	InsertStatement(table *Table) (string, error)
> 
> 	// DirectInsertStatement returns an insert statement without any bound parameters. Only string
> ```

It's not obvious to me why this exists, as opposed to the library combining InsertStatement with placeholder evaluation.
___
*[go/materialize/driver/sqlgen.go, line 123 at r2](https://reviewable.io/reviews/estuary/flow/86#-MQDwT4j1pdrGrmgBAIh:-MQDwT4j1pdrGrmgBAIi:bv4aky6) ([raw file](https://github.com/estuary/flow/blob/b588d4d420b52b102d713310b7f94bb7a7ea9b70/go/materialize/driver/sqlgen.go#L123)):*
> ```Go
> // A TypeMapper resolves a Column to a specific base SQL type. For example, for all "string" type
> // Columns, it may return the "TEXT" sql type. We use a decorator pattern to compose TypeMappers.
> type TypeMapper interface {
> ```

I like this interface & decorator pattern, but I'm trying to square it with `ColumnType` type above, and questioning whether `ColumnType` holds water as a concept. As opposed to having *Column represent unvarnished inference state as extracted / understood by Flow, and using `TypeMapper` as a pure-function mapping from that Flow-understood (or passed-through) state directly through to the raw DB type string.

In particular, how do we account for things like hyperloglogs and other probabilistic digests which might have first-class support in some DBs ? Or for specialized types like vectors of integer types, etc which may be offered by the database ? There could be _lots_ of varieties of these.

Is it possible to **not** have a representational stake in the space in between flow inference and underlying DB types ?

I'm guessing one of the core rationales for ColumnType is that it's the space we work in for semantic change detection. Do we *need* it? Could we, for example, simply persist the raw output of the TypeMapper on a previous invocation and error if the literal database type changes ? That might mean giving up some flexibility (e.x., changing null-ability), but how important is that really?
___
*[go/materialize/driver/sqlgen.go, line 258 at r2](https://reviewable.io/reviews/estuary/flow/86#-MQE0QKI2S7VmTzx-sCq:-MQE0QKI2S7VmTzx-sCr:b2gr8nv) ([raw file](https://github.com/estuary/flow/blob/b588d4d420b52b102d713310b7f94bb7a7ea9b70/go/materialize/driver/sqlgen.go#L258)):*
> ```Go
> // GetPostgresParameterPlaceholder returns $N style parameters where N is the parameter number
> // starting at 1.
> func GetPostgresParameterPlaceholder(parameterIndex int) string {
> ```

nit: [just `PostgresParameterPlaceholder`](https://golang.org/doc/effective_go.html#Getters)
___
*[go/protocols/materialize/materialize_extensions.go, line 6 at r1](https://reviewable.io/reviews/estuary/flow/86#-MQDjZHH5EaqMZ71yXLz:-MQDi-0eEuwG2BvHCE6p:b-gz30cd) ([raw file](https://github.com/estuary/flow/blob/50bbd9d62f09661b9e2dde19a5e574f64c55b800/go/protocols/materialize/materialize_extensions.go#L6)):*
> ```Go
> // return true for FIELD_FORBIDDEN and UNSATISFIABLE, and false for any other constraint type.
> func (constraintType *Constraint_Type) IsForbidden() bool {
> 	switch *constraintType {
> ```

nit: `m` ("message") or `ct` ? it's pretty unconventional to use the entire type name within the receiver name.

psFried (commented):
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/86)* status: 15 of 28 files reviewed, 12 unresolved discussions (waiting on @jgraettinger and @psFried)

___
*[go/materialize/driver/sql.go, line 55 at r3](https://reviewable.io/reviews/estuary/flow/86#-MQE1l0082hPU5ylBahi:-MQEX6RP-AhKHhxj4mKc:bb6un1j) ([raw file](https://github.com/estuary/flow/blob/d86b680e8b004de2ce4760f6fc7ab461c8e3f6ff/go/materialize/driver/sql.go#L55)):*

nit: uniquely identify instances of a shard => identify unique process assignments of a shard
</blockquote></details>

Done
___
*[go/materialize/driver/sql.go, line 154 at r3](https://reviewable.io/reviews/estuary/flow/86#-MQE36VN0yojB-Z1MHzb:-MQEXCjFAh5X7CYF4FPz:bq6hz3w) ([raw file](https://github.com/estuary/flow/blob/d86b680e8b004de2ce4760f6fc7ab461c8e3f6ff/go/materialize/driver/sql.go#L154)):*

My own mistake, but I'm beginning to think that "caller ID" was an error in naming, as opposed to "shard ID". The conceit was that driver authors wouldn't need to know about shards. I now don't think that's true, and therefore the name adds overhead without value.
</blockquote></details>

I do think it's a little weird to have it called CallerId in the protocol and then call it shard_id in the table. That said, I think I still like the name CallerId because it's inclusive of flowctl, which will have to provide some CallerId in order to start a Session. Personally, I feel it's ok as is. WDYT?
___
*[go/materialize/driver/sql.go, line 254 at r3](https://reviewable.io/reviews/estuary/flow/86#-MQE4hjT0d5y5oq11s4A:-MQEYB9RAX971cQlfpSd:brc1sas) ([raw file](https://github.com/estuary/flow/blob/d86b680e8b004de2ce4760f6fc7ab461c8e3f6ff/go/materialize/driver/sql.go#L254)):*

couldn't this use SqlGen.InsertStatement(), and then immediately invoke it with params ?
</blockquote></details>

Yeah, that's exactly what it was doing until I realized that we wouldn't be able to produce a usable `actionDescription` that includes parameters. DirectInsertStatement exists just for this case where we may spit out a script that the user will then execute themselves.
___
*[go/materialize/driver/sql.go, line 258 at r3](https://reviewable.io/reviews/estuary/flow/86#-MQE4NZXC4KeNHiwWP4Z:-MQEYroz5KP2mXqXUsFs:bb6un1j) ([raw file](https://github.com/estuary/flow/blob/d86b680e8b004de2ce4760f6fc7ab461c8e3f6ff/go/materialize/driver/sql.go#L258)):*

nit: strike "fully reduced" here? It would need to be defined otherwise, and really it's an implementation detail from the user's perspective.
</blockquote></details>

Done
___
*[go/materialize/driver/sql.go, line 400 at r3](https://reviewable.io/reviews/estuary/flow/86#-MQE6Zkn7WIGN6i5z0oM:-MQEZ-EiCSC-CtkLrx7K:b-h9mk1o) ([raw file](https://github.com/estuary/flow/blob/d86b680e8b004de2ce4760f6fc7ab461c8e3f6ff/go/materialize/driver/sql.go#L400)):*

why not do this now?
</blockquote></details>

I don't think I follow. Do what now?
___
*[go/materialize/driver/sql.go, line 409 at r3](https://reviewable.io/reviews/estuary/flow/86#-MQE6xVGCl-giko06xK_:-MQEbApC7AfLKpWGqnvX:bb6un1j) ([raw file](https://github.com/estuary/flow/blob/d86b680e8b004de2ce4760f6fc7ab461c8e3f6ff/go/materialize/driver/sql.go#L409)):*

there's a method for this now! :)
</blockquote></details>

Done
___
*[go/materialize/driver/sql.go, line 410 at r3](https://reviewable.io/reviews/estuary/flow/86#-MQE7FOq7s4SGytDyYn6:-MQEbmEI8QLRn9a-H2v8:btnc94q) ([raw file](https://github.com/estuary/flow/blob/d86b680e8b004de2ce4760f6fc7ab461c8e3f6ff/go/materialize/driver/sql.go#L410)):*

This is giving up a really crucial optimization, which is that Load() has a bunch of keys in-hand and can ask for them all at once rather than sequentially.

It's this issue that IMO motivates looking beyond `driver/sql`. Of course for sqlite it doesn't matter, but if we're talking to postgres over the network, this can be a crushing performance bottleneck.

The [pgx](https://pkg.go.dev/github.com/jackc/pgx/v4#Batch) driver offers an API for building up batch queries.

Another option would be to query via `SELECT` with an inner join against a constant `VALUES` table representing the batch, pared with `INSERT ... VALUES ` and `UPDATE ... FROM` for Store.
</blockquote></details>

Yeah, I think this will be important, and something I should have mentioned in the not-done-yet list. I think it makes sense to get the test harness working before adding the postgres-specific optimizations. Thinking also that the test harness may be a good place to put some basic performance testing, which might help too. Also TBD is a good factoring for the internals that allows for maximizing code reuse between the generic version and the postgres-specific one. IMO it makes sense to merge things in the current state and follow on with the rest. How's that plan sound to you?
___
*[go/materialize/driver/sql.go, line 576 at r3](https://reviewable.io/reviews/estuary/flow/86#-MQEE5Sa6L2hyiPCiyOy:-MQEfwsDBtRB0ivKLBJI:b-ds6ehz) ([raw file](https://github.com/estuary/flow/blob/d86b680e8b004de2ce4760f6fc7ab461c8e3f6ff/go/materialize/driver/sql.go#L576)):*

same performance concern here. This is executing all updates and inserts sequentially, waiting for the DB to ack each before moving on to the next.
</blockquote></details>

Yeah, let's discuss the plan in the comment above.
___
*[go/materialize/driver/sql.go, line 595 at r3](https://reviewable.io/reviews/estuary/flow/86#-MQEGTRrA2uFuj03olKD:-MQEgM7S3gPJ8BfO1PXu:ba3dfob) ([raw file](https://github.com/estuary/flow/blob/d86b680e8b004de2ce4760f6fc7ab461c8e3f6ff/go/materialize/driver/sql.go#L595)):*

is this used (it doesn't appear so)?
</blockquote></details>

Removed
___
*[go/materialize/driver/sqlgen.go, line 55 at r2](https://reviewable.io/reviews/estuary/flow/86#-MQDkkPb0w0YvIULuaWi:-MQEhTpa4-YQGT5pLsHE:b8ng8m5) ([raw file](https://github.com/estuary/flow/blob/b588d4d420b52b102d713310b7f94bb7a7ea9b70/go/materialize/driver/sqlgen.go#L55)):*

nit: this comment repeats itself but is also unclear (what are "automatic" columns?).
</blockquote></details>

Fixed. "automatic" was referring to things like rowid.
___
*[go/materialize/driver/sqlgen.go, line 77 at r2](https://reviewable.io/reviews/estuary/flow/86#-MQDlB3I82JoeOIZJxmz:-MQEhmMk0GlS30wKOiGu:b-kyk8kh) ([raw file](https://github.com/estuary/flow/blob/b588d4d420b52b102d713310b7f94bb7a7ea9b70/go/materialize/driver/sqlgen.go#L77)):*

aren't we guaranteeing that `selectColumns` isn't empty, before calling into this interface ? it's weird to make this check part of the implementers responsibility.
</blockquote></details>

Yes, we are guaranteeing that selectColumns is non empty. I'll just remove that 
___
*[go/materialize/driver/sqlgen.go, line 82 at r2](https://reviewable.io/reviews/estuary/flow/86#-MQDlTsw7UW19ABqkvMZ:-MQEiXf12SPw437AHrdl:b4szc67) ([raw file](https://github.com/estuary/flow/blob/b588d4d420b52b102d713310b7f94bb7a7ea9b70/go/materialize/driver/sqlgen.go#L82)):*

nit: grammer. Do we require that placeholders be in a specific order (I'd assume yes)?

should guidance be that we _always_ want implementers to use plain insert statements, since we "know" which is needed and violation checks are a valuable runtime assertion ?
Put differently, what are the exceptions to "For most systems" from your comment ?
</blockquote></details>

Yeah I think the "most" was just my way of being had wavy and non-committal. I was imagining that there might be systems that might prefer to use an upsert for everything for efficiency or ergonomics. Basically I think we want to impart the guidance that you should just use a plain insert, but that technically it doesn't matter as long as the parameters match what we expect (same order and number as the Columns in the Table). I'll try to clarify.
___
*[go/materialize/driver/sqlgen.go, line 86 at r2](https://reviewable.io/reviews/estuary/flow/86#-MQDlrfmAa9cdbZkS-oG:-MQI1lVvEdqa2MZVXpeY:b-m1jxo3) ([raw file](https://github.com/estuary/flow/blob/b588d4d420b52b102d713310b7f94bb7a7ea9b70/go/materialize/driver/sqlgen.go#L86)):*

It's not obvious to me why this exists, as opposed to the library combining InsertStatement with placeholder evaluation.
</blockquote></details>

This is so we can generate a statement that can be included in the ActionDescription, which of course can't use parameters.
___
*[go/materialize/driver/sqlgen.go, line 123 at r2](https://reviewable.io/reviews/estuary/flow/86#-MQDwT4j1pdrGrmgBAIh:-MQI2KTT8PjOqvPUdl_v:b4rwtfh) ([raw file](https://github.com/estuary/flow/blob/b588d4d420b52b102d713310b7f94bb7a7ea9b70/go/materialize/driver/sqlgen.go#L123)):*

I like this interface & decorator pattern, but I'm trying to square it with `ColumnType` type above, and questioning whether `ColumnType` holds water as a concept. As opposed to having *Column represent unvarnished inference state as extracted / understood by Flow, and using `TypeMapper` as a pure-function mapping from that Flow-understood (or passed-through) state directly through to the raw DB type string.

In particular, how do we account for things like hyperloglogs and other probabilistic digests which might have first-class support in some DBs ? Or for specialized types like vectors of integer types, etc which may be offered by the database ? There could be _lots_ of varieties of these.

Is it possible to **not** have a representational stake in the space in between flow inference and underlying DB types ?

I'm guessing one of the core rationales for ColumnType is that it's the space we work in for semantic change detection. Do we *need* it? Could we, for example, simply persist the raw output of the TypeMapper on a previous invocation and error if the literal database type changes ? That might mean giving up some flexibility (e.x., changing null-ability), but how important is that really?
</blockquote></details>

I think the main justification for the current Column representation is just that it's used for both materialization tables and for the `gazette_checkpoints` and `flow_materializations` tables. Notably, `ColumnType` includes a "binary" type, which is used for checkpoint values. Inference can only represent binary as strings that are base64 encoded (or an array of integers, I guess). Additionally, the Column representation allows for only a single unambiguous type (with nullability represented separately). This intermediate means we can generically verify that each materialized projection has a single type, and allows for a simpler SqlGenerator implementation since it doesn't have to deal with the possibility of multiple types and multiple ways to represent that a column may be null.

Basically, I think it's _possible_ to not have a representation stake here, but I think it's better that we do have this separate representation.
___
*[go/materialize/driver/sqlgen.go, line 258 at r2](https://reviewable.io/reviews/estuary/flow/86#-MQE0QKI2S7VmTzx-sCq:-MQI5KUU47all7wzRqp-:bb6un1j) ([raw file](https://github.com/estuary/flow/blob/b588d4d420b52b102d713310b7f94bb7a7ea9b70/go/materialize/driver/sqlgen.go#L258)):*

nit: [just `PostgresParameterPlaceholder`](https://golang.org/doc/effective_go.html#Getters)
</blockquote></details>

Done
___
*[go/protocols/materialize/materialize_extensions.go, line 6 at r1](https://reviewable.io/reviews/estuary/flow/86#-MQDjZHH5EaqMZ71yXLz:-MQI67hA9-absIu3uoCt:b-by5op8) ([raw file](https://github.com/estuary/flow/blob/50bbd9d62f09661b9e2dde19a5e574f64c55b800/go/protocols/materialize/materialize_extensions.go#L6)):*

nit: `m` ("message") or `ct` ? it's pretty unconventional to use the entire type name within the receiver name.
</blockquote></details>

I'll be honest and say that I strongly disagree with the convention of using an initialism unless it's one that's universally understood (e.g. `i` for index in a loop). It can easily hurt readability when you have to pause to figure out what some initialism means, and there's very few cases where spelling out the word makes code more difficult to read. The go linter even has a lint forbidding "non-descriptive" names like "this" or "self", both of which are more descriptive than "m" or "ct", IMO.</rant>
Anyway, I'm not one for dogmatism, so if you still think that I ought to fall in line and start using shorter names for receiver variables in go, then just LMK and I'll start doing so and not lose a minute of sleep over it.

jgraettinger (approved):
> Reviewed 6 of 8 files at r3, 7 of 7 files at r4, 4 of 4 files at r5, 18 of 18 files at r6, 5 of 5 files at r7, 2 of 2 files at r8.
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/86)* status: all files reviewed, 5 unresolved discussions (waiting on @psFried)

___
*[go/materialize/driver/sql.go, line 154 at r3](https://reviewable.io/reviews/estuary/flow/86#-MQE36VN0yojB-Z1MHzb:-MQJI0kr9cpqiAiXxK3s:b-wnci8u) ([raw file](https://github.com/estuary/flow/blob/d86b680e8b004de2ce4760f6fc7ab461c8e3f6ff/go/materialize/driver/sql.go#L154)):*

I do think it's a little weird to have it called CallerId in the protocol and then call it shard_id in the table. That said, I think I still like the name CallerId because it's inclusive of flowctl, which will have to provide some CallerId in order to start a Session. Personally, I feel it's ok as is. WDYT?
</blockquote></details>

resolved over VC, switching to "shard ID".
___
*[go/materialize/driver/sql.go, line 400 at r3](https://reviewable.io/reviews/estuary/flow/86#-MQE6Zkn7WIGN6i5z0oM:-MQJI6Pz5mvGkRg3UxU5:b-p6s1ua) ([raw file](https://github.com/estuary/flow/blob/d86b680e8b004de2ce4760f6fc7ab461c8e3f6ff/go/materialize/driver/sql.go#L400)):*

I don't think I follow. Do what now?
</blockquote></details>

resolved over VC, in reference to embedding Slice to avoid []*Slice's.

I was going through this commit by commit, so was unknowingly seeing your churn.
___
*[go/materialize/driver/sql.go, line 410 at r3](https://reviewable.io/reviews/estuary/flow/86#-MQE7FOq7s4SGytDyYn6:-MQJIDRKEgo3ecXXLN6i:bds0pcm) ([raw file](https://github.com/estuary/flow/blob/d86b680e8b004de2ce4760f6fc7ab461c8e3f6ff/go/materialize/driver/sql.go#L410)):*

Yeah, I think this will be important, and something I should have mentioned in the not-done-yet list. I think it makes sense to get the test harness working before adding the postgres-specific optimizations. Thinking also that the test harness may be a good place to put some basic performance testing, which might help too. Also TBD is a good factoring for the internals that allows for maximizing code reuse between the generic version and the postgres-specific one. IMO it makes sense to merge things in the current state and follow on with the rest. How's that plan sound to you?
</blockquote></details>

resolved over VC
___
*[go/materialize/driver/sqlgen.go, line 123 at r2](https://reviewable.io/reviews/estuary/flow/86#-MQDwT4j1pdrGrmgBAIh:-MQICzIe1VH_1-cQXlHE:b3cxh02) ([raw file](https://github.com/estuary/flow/blob/b588d4d420b52b102d713310b7f94bb7a7ea9b70/go/materialize/driver/sqlgen.go#L123)):*

I think the main justification for the current Column representation is just that it's used for both materialization tables and for the `gazette_checkpoints` and `flow_materializations` tables. Notably, `ColumnType` includes a "binary" type, which is used for checkpoint values. Inference can only represent binary as strings that are base64 encoded (or an array of integers, I guess). Additionally, the Column representation allows for only a single unambiguous type (with nullability represented separately). This intermediate means we can generically verify that each materialized projection has a single type, and allows for a simpler SqlGenerator implementation since it doesn't have to deal with the possibility of multiple types and multiple ways to represent that a column may be null.

Basically, I think it's _possible_ to not have a representation stake here, but I think it's better that we do have this separate representation.
</blockquote></details>

Resolved over VC. We're going to ditch ColumnType as a concept, and leverage the `TypeMapper` concept as a pure function that consumes Flow-produced inferences (including passed-through annotations) to produce literal database types. We discussed that he existing functionality of ColumnTypes can be approached via a TypeMapper which projects to current ColumnType literals, and delegates then delegates to meta-type-specific TypeMappers.
___
*[go/protocols/materialize/materialize_extensions.go, line 6 at r1](https://reviewable.io/reviews/estuary/flow/86#-MQDjZHH5EaqMZ71yXLz:-MQJHvue9V4vN1y-eUFm:bds0pcm) ([raw file](https://github.com/estuary/flow/blob/50bbd9d62f09661b9e2dde19a5e574f64c55b800/go/protocols/materialize/materialize_extensions.go#L6)):*

I'll be honest and say that I strongly disagree with the convention of using an initialism unless it's one that's universally understood (e.g. `i` for index in a loop). It can easily hurt readability when you have to pause to figure out what some initialism means, and there's very few cases where spelling out the word makes code more difficult to read. The go linter even has a lint forbidding "non-descriptive" names like "this" or "self", both of which are more descriptive than "m" or "ct", IMO.</rant>
Anyway, I'm not one for dogmatism, so if you still think that I ought to fall in line and start using shorter names for receiver variables in go, then just LMK and I'll start doing so and not lose a minute of sleep over it.
</blockquote></details>

resolved over VC

PR #87, opened 2021-01-07
crates/models: factored crate focusing on catalog domain

jgraettinger:
This packs a whole bunch of planned catalog changes into a parallel
implementing crate, leaving `catalog` in place for now.

It also introduces a new "visitor" pattern for traversing catalog
entities and sources, de-emphasizing the SQLite database as a primary
state store during catalog building steps.

Future PRs will introduce specialized visitors which perform validation
and catalog build tasks.

Issue #68

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/87)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1605,6 +1605,37 @@ dependencies = [
 ]
 
 [[package]]
+name = "models"
+version = "0.0.0"
+dependencies = [
+ "caseless",
+ "doc",
+ "futures",
+ "humantime-serde",
+ "include_dir",
+ "insta",
+ "itertools 0.9.0",
+ "json",
+ "log",
+ "pretty_env_logger",
+ "protocol",
+ "regex",
+ "reqwest",
+ "rusqlite",
+ "schemars",
+ "serde",
+ "serde_json",
+ "serde_yaml",
+ "strsim 0.10.0",
+ "tempfile",
+ "thiserror",
+ "tracing",
+ "unicode-normalization",
+ "url",
+ "yaml-merge-keys",
+]
+
+[[package]]
 name = "multimap"
 version = "0.8.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
--- a/crates/models/Cargo.toml
+++ b/crates/models/Cargo.toml
@@ -1,0 +1,37 @@
+[package]
+name = "models"
+version = "0.0.0"
+authors = ["Estuary Technologies, Inc"]
+edition = "2018"
+
+[dependencies]
+json = { path = "../json", version = "0.0.0" }
+doc = { path = "../doc", version = "0.0.0" }
+protocol = { path = "../protocol", version = "0.0.0" }
+
+futures = "*"
+caseless = "*"
+humantime-serde = "*"
+include_dir = { version = "*", default-features = false }
+itertools = "*"
+log = "*"
+regex = "*"
+reqwest = { version = "*", default-features = false, features = ["blocking", "native-tls"] }
+rusqlite = { version = "*", features = ["bundled", "collation", "column_decltype", "functions", "serde_json", "url"] }
+schemars = "*"
+serde = { version = "*", features = ["derive"] }
+serde_json = { version =  "*", features = ["raw_value"] }
+serde_yaml = "*"
+strsim = "*"
+tempfile = "*"
+thiserror = "*"
+unicode-normalization = "*"
+url = "*"
+yaml-merge-keys = { version = "*", features = ["serde_yaml"] }
+tracing = "*"
+
+[dev-dependencies]
+protocol = { path = "../protocol", version = "0.0.0" }
+
+insta = "*"
+pretty_env_logger = "*"
--- a/crates/models/src/lib.rs
+++ b/crates/models/src/lib.rs
@@ -1,0 +1,1 @@
+pub mod source;
--- a/crates/models/src/source/loader.rs
+++ b/crates/models/src/source/loader.rs
@@ -1,0 +1,870 @@
+use super::wrappers::{
+    CaptureName, CollectionName, ContentType, EndpointName, EndpointType, JsonPointer,
+    MaterializationName, ShuffleHash, TestName, TransformName,
+};
+use super::{specs, Scope};
+
+use doc::Schema as CompiledSchema;
+use futures::future::{FutureExt, LocalBoxFuture};
+use futures::TryFutureExt;
+use json::schema::{build::build_schema, Application, Keyword};
+use std::cell::RefCell;
+use std::collections::HashSet;
+use std::future::Future;
+use url::Url;
+
+#[derive(thiserror::Error, Debug)]
+pub enum LoadError {
+    #[error("failed to parse URL")]
+    URLParse(#[from] url::ParseError),
+    #[error("failed to fetch resource {uri}")]
+    Fetch {
+        uri: String,
+        #[source]
+        detail: Box<dyn std::error::Error + Send + Sync>,
+    },
+    #[error("failed to parse YAML (location {:?})", .0.location())]
+    YAMLErr(#[from] serde_yaml::Error),
+    #[error("failed to merge YAML alias nodes")]
+    YAMLMergeErr(#[from] yaml_merge_keys::MergeKeyError),
+    #[error("failed to build JSON schema")]
+    SchemaBuild(#[from] json::schema::BuildError),
+    #[error("failed to index JSON schema")]
+    SchemaIndex(#[from] json::schema::index::Error),
+}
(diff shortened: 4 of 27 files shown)


psFried (changes requested):
Looking great so far. I really like the scenario tests. I feel like it's a little unfortunate that the fetch order is significant in those snapshots, as it seems to maybe motivate a fair bit of additional complexity. That said, I don't have a better idea handy, so I think it'll be OK as it is. The new specs look good. I'm really liking how the endpoint configs are shaping up.

> Reviewed 27 of 27 files at r1.
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/87#-:-MQoTY-YBG7S6KQYcpj8:b-tunt0x)* status: all files reviewed, 3 unresolved discussions (waiting on @jgraettinger)

___
*[crates/models/src/source/specs.rs, line 594 at r1](https://reviewable.io/reviews/estuary/flow/87#-MQoL_Cs7AwPL7sgF5cy:-MQoL_Cs7AwPL7sgF5cz:b-ems5h) ([raw file](https://github.com/estuary/flow/blob/eeff2e283aee450bc9ff7fd4d68ec7d5c9ec0dd1/crates/models/src/source/specs.rs#L594)):*
> ```Rust
> #[derive(Serialize, Deserialize, Debug, JsonSchema)]
> #[serde(deny_unknown_fields, rename_all = "camelCase")]
> pub enum Endpoint {
> ```

What do you think about adding a generic endpoint type that just takes a driver URI and pases through everything else? You reckon it's premature?
___
*[crates/models/src/source/specs.rs, line 713 at r1](https://reviewable.io/reviews/estuary/flow/87#-MQoK_597Io7aAGqfDAk:-MQoK_597Io7aAGqfDAl:bpncv07) ([raw file](https://github.com/estuary/flow/blob/eeff2e283aee450bc9ff7fd4d68ec7d5c9ec0dd1/crates/models/src/source/specs.rs#L713)):*
> ```Rust
>     /// # Additional endpoint configuration.
>     /// Configuration is merged into that of the endpoint.
>     #[serde(default = "value_null")]
> ```

nit: `Value` already implements Default to return Null, so you can just use `serde(default)`. Not actually important to change this, but just fyi
___
*[crates/models/src/source/specs.rs, line 738 at r1](https://reviewable.io/reviews/estuary/flow/87#-MQoJyxABlQqvC-kRmVi:-MQoJyxABlQqvC-kRmVj:bkcyexk) ([raw file](https://github.com/estuary/flow/blob/eeff2e283aee450bc9ff7fd4d68ec7d5c9ec0dd1/crates/models/src/source/specs.rs#L738)):*
> ```Rust
>     exclude: Vec<String>,
>     /// # Should recommended projections for the endpoint be used?
>     recommended: bool,
> ```

I like recommended better than 'default' :+1: 
___
*[crates/models/src/source/scenarios/mod.rs, line 66 at r1](https://reviewable.io/reviews/estuary/flow/87#-MQoOym336YZcjgjUwfp:-MQoOym336YZcjgjUwfq:b55rdc4) ([raw file](https://github.com/estuary/flow/blob/eeff2e283aee450bc9ff7fd4d68ec7d5c9ec0dd1/crates/models/src/source/scenarios/mod.rs#L66)):*
> ```Rust
>         .boxed_local();
> 
>     let waker = futures::task::noop_waker();
> ```

I'll admit to some skepticism on the approach for driving the futures here, but I get the impression that this was motivated by concerns over fetches executing in a deterministic order. I think it might be worth circling back here and trying to figure out how to get this working with a "current thread" tokio runtime, even if it's later on in another PR. At a minimum, I think it'd be useful to have a comment explaining why this is being done in this way.
___
*[crates/models/src/source/scenarios/mod.rs, line 72 at r1](https://reviewable.io/reviews/estuary/flow/87#-MQoNtXP96v9E-KLhAO3:-MQoNtXP96v9E-KLhAO4:bvt5jxi) ([raw file](https://github.com/estuary/flow/blob/eeff2e283aee450bc9ff7fd4d68ec7d5c9ec0dd1/crates/models/src/source/scenarios/mod.rs#L72)):*
> ```Rust
>         match fut.poll_unpin(&mut ctx) {
>             Poll::Ready(Ok(())) => {
>                 std::mem::forget(fut);
> ```

did you mean `drop`? `forget` is for when you want to leak memory.
___
*[crates/models/src/source/scenarios/mod.rs, line 77 at r1](https://reviewable.io/reviews/estuary/flow/87#-MQoRZDhBrbEqmngus6i:-MQoRZDi7qjRB7dxdVw4:b-fr1uux) ([raw file](https://github.com/estuary/flow/blob/eeff2e283aee450bc9ff7fd4d68ec7d5c9ec0dd1/crates/models/src/source/scenarios/mod.rs#L77)):*
> ```Rust
>             Poll::Ready(Err(err)) => return Err(err),
>             Poll::Pending if fetches.borrow().is_empty() => {
>                 panic!("future is pending, but started no fetches")
> ```

This assertion doesn't seem necessary, and it might make the test a bit fragile. There's lots of other reasons why a future might go into a pending state, and I can't think of a reason why we'd want to explicitly _not_ tolerate that. Maybe a comment, if it's important?

jgraettinger (commented):
Thanks. Some comments below. As discussed on VC, I'm prepping a new PR which covers this (now significantly re-worked) ground.

> *[Reviewable](https://reviewable.io/reviews/estuary/flow/87#-:-MRfdGKE8swsRb0XZ3Ft:b-d3ghm6)* status: all files reviewed, 3 unresolved discussions (waiting on @psFried)

___
*[crates/models/src/source/specs.rs, line 594 at r1](https://reviewable.io/reviews/estuary/flow/87#-MQoL_Cs7AwPL7sgF5cy:-MRf_XUR3qV8yd2Vax61:b-896fix) ([raw file](https://github.com/estuary/flow/blob/eeff2e283aee450bc9ff7fd4d68ec7d5c9ec0dd1/crates/models/src/source/specs.rs#L594)):*

What do you think about adding a generic endpoint type that just takes a driver URI and pases through everything else? You reckon it's premature?
</blockquote></details>

Done.
___
*[crates/models/src/source/specs.rs, line 713 at r1](https://reviewable.io/reviews/estuary/flow/87#-MQoK_597Io7aAGqfDAk:-MRfap6YAQBc6qT4WQB2:bb7f2mf) ([raw file](https://github.com/estuary/flow/blob/eeff2e283aee450bc9ff7fd4d68ec7d5c9ec0dd1/crates/models/src/source/specs.rs#L713)):*

nit: `Value` already implements Default to return Null, so you can just use `serde(default)`. Not actually important to change this, but just fyi
</blockquote></details>

done
___
*[crates/models/src/source/scenarios/mod.rs, line 66 at r1](https://reviewable.io/reviews/estuary/flow/87#-MQoOym336YZcjgjUwfp:-MRfdETD05C866qLxw5B:b-o37f60) ([raw file](https://github.com/estuary/flow/blob/eeff2e283aee450bc9ff7fd4d68ec7d5c9ec0dd1/crates/models/src/source/scenarios/mod.rs#L66)):*

I'll admit to some skepticism on the approach for driving the futures here, but I get the impression that this was motivated by concerns over fetches executing in a deterministic order. I think it might be worth circling back here and trying to figure out how to get this working with a "current thread" tokio runtime, even if it's later on in another PR. At a minimum, I think it'd be useful to have a comment explaining why this is being done in this way.
</blockquote></details>

addressed via comments
___
*[crates/models/src/source/scenarios/mod.rs, line 72 at r1](https://reviewable.io/reviews/estuary/flow/87#-MQoNtXP96v9E-KLhAO3:-MRfavhb8gNggRvxt_hH:b-896fix) ([raw file](https://github.com/estuary/flow/blob/eeff2e283aee450bc9ff7fd4d68ec7d5c9ec0dd1/crates/models/src/source/scenarios/mod.rs#L72)):*

did you mean `drop`? `forget` is for when you want to leak memory.
</blockquote></details>

Done.
___
*[crates/models/src/source/scenarios/mod.rs, line 77 at r1](https://reviewable.io/reviews/estuary/flow/87#-MQoRZDhBrbEqmngus6i:-MRfdD6_7oDQM53hxplW:b-o37f60) ([raw file](https://github.com/estuary/flow/blob/eeff2e283aee450bc9ff7fd4d68ec7d5c9ec0dd1/crates/models/src/source/scenarios/mod.rs#L77)):*

This assertion doesn't seem necessary, and it might make the test a bit fragile. There's lots of other reasons why a future might go into a pending state, and I can't think of a reason why we'd want to explicitly _not_ tolerate that. Maybe a comment, if it's important?
</blockquote></details>

addressed via comments

jgraettinger:
Dropping this PR in favor of PR #89

PR #88, opened 2021-01-21
Materialize transactions

psFried:
This PR includes a few major pieces on the road to getting materializations working end to end with the generic driver protocol:

- Refactors the materialize protocol to run all load and store operations inside a single `transaction` bi-directional streaming rpc.
- Refactors the `materialize/driver/sql` package to account for the protocol changes, and to better support the in-progress postgres driver.
- Adds the `driver-tester` binary to run tests against any driver implementation.
- Updates our `tokio` dependencies to 1.0! I initially did this while investigating an issue with tonic streaming responses. It turned out not to be necessary, but I figured it'd be worth keeping.

What's _not_ included:

- Postgres driver implementation is still in progress on another branch.
- Transactional test in driver-tester is waiting on the postgres impl, since sqlite can't really pass it.
- Running the driver-tester tests in CI is just TBD, but not blocked on anything.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/88)


jgraettinger (changes requested):
Sending some partial feedback your way. Still working through this, I've reviewed through the rust portions and next up is go/ files.

> Reviewed 29 of 54 files at r1.
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/88#-:-MRgSu7h4aYLCWd5JAR5:b-il5t2u)* status: 29 of 54 files reviewed, 14 unresolved discussions (waiting on @psFried)

___
*[crates/derive/Cargo.toml, line 17 at r1](https://reviewable.io/reviews/estuary/flow/88#-MRgBROH76sd1V2EGvFj:-MRgBROH76sd1V2EGvFk:b-l4vean) ([raw file](https://github.com/estuary/flow/blob/e472e0fe6c992d2d4cad2406bc30a215d9112a6b/crates/derive/Cargo.toml#L17)):*
> ```TOML
> futures = "*"
> http = "*"
> hyper = "0.14"
> ```

Any place where we're setting explicit versions (other than '*'), let's comment on exactly why we're doing it and what needs to happen before we can switch it back to '*'.

E.x., the ideal unachievable world state is that all of our crates are using '*', and we're using the Cargo.lock exclusively to pin things. Only when we can't do that for some reason should we use explicit versions here, and that calls for documenting why we're doing it.
___
*[crates/driver-tester/Cargo.toml, line 34 at r1](https://reviewable.io/reviews/estuary/flow/88#-MRgCxSD5T3-6nSES9nU:-MRgCxSD5T3-6nSES9nV:b-5svfqw) ([raw file](https://github.com/estuary/flow/blob/e472e0fe6c992d2d4cad2406bc30a215d9112a6b/crates/driver-tester/Cargo.toml#L34)):*
> ```TOML
> tempfile = "*"
> tokio = { version = "1.0", features = ["full"] }
> tokio-stream = "0.1"
> ```

tokio-stream seems duplicative with `futures` ? what's driving this ?
___
*[crates/driver-tester/src/request.rs, line 109 at r1](https://reviewable.io/reviews/estuary/flow/88#-MRgF9Po-DXZ1kBoqjJ0:-MRgF9Po-DXZ1kBoqjJ1:binoq6m) ([raw file](https://github.com/estuary/flow/blob/e472e0fe6c992d2d4cad2406bc30a215d9112a6b/crates/driver-tester/src/request.rs#L109)):*
> ```Rust
>         loop {
>             let msg = self.0.recv().await?.ok_or_else(||
>                                                          anyhow::anyhow!("expected a LoadResponse message, but got EOF")
> ```

nit: this is formatted ? weird how indented it is.
___
*[crates/driver-tester/src/request.rs, line 157 at r1](https://reviewable.io/reviews/estuary/flow/88#-MRgFQaB9gHYCiOloZB6:-MRgFQaB9gHYCiOloZB7:bskfatc) ([raw file](https://github.com/estuary/flow/blob/e472e0fe6c992d2d4cad2406bc30a215d9112a6b/crates/driver-tester/src/request.rs#L157)):*
> ```Rust
>     // send messages on the TransactionSender before receiving the first response. Otherwise, this
>     // will essentially deadlock, with the client and driver both blocked awaiting receipt of a message.
>     let fut = tokio::spawn(async move {
> ```

I don't think I understand why this needs to be in a background task. Can't TrasnactionReciever::Pending just wrap this future directly, without spawning it to a task ?

In other words, why does this future need to be driven independently of calls to LoadReciever::recv() ?
___
*[crates/driver-tester/src/request.rs, line 323 at r1](https://reviewable.io/reviews/estuary/flow/88#-MRgHRV9-g1bPEZonxTd:-MRgHRV9-g1bPEZonxTe:b-h23s4a) ([raw file](https://github.com/estuary/flow/blob/e472e0fe6c992d2d4cad2406bc30a215d9112a6b/crates/driver-tester/src/request.rs#L323)):*
> ```Rust
> 
> fn extract_and_pack(ptrs: &[Pointer], json: &Value, arena: &mut Vec<u8>) -> Slice {
>     let json = json.borrow();
> ```

nit: don't think you need this borrow() ?
___
*[crates/driver-tester/src/request.rs, line 327 at r1](https://reviewable.io/reviews/estuary/flow/88#-MRgJ1Cm96D3KBSdh4Fa:-MRgJ1Cm96D3KBSdh4Fb:b3f7so8) ([raw file](https://github.com/estuary/flow/blob/e472e0fe6c992d2d4cad2406bc30a215d9112a6b/crates/driver-tester/src/request.rs#L327)):*
> ```Rust
>     for ptr in ptrs {
>         let value = ptr.query(json).unwrap_or(&Value::Null);
>         // TODO: I'm not sure I understand why the depth is 1 here and not 0.
> ```

It's because FDB tuples encode themselves without a leading byte indicator _only if they're the root-most element_. If they're non-root, they prefix with a "nested" byte indicator.

Since we're manually building a tuple here, depth is one to indicates we're "inside" the top-level tuple. Importantly, that means if that one of these elements is an array, it will write a nesting byte instead of assuming it's root-most and effectively flattening its elements up into the parent.

Another way to look at this is that if you mapped ptrs through to an extracted Vec<&Value> and packed *that*, _then_ the depth would be zero.
___
*[crates/driver-tester/src/tests/mod.rs, line 204 at r1](https://reviewable.io/reviews/estuary/flow/88#-MRgLdKh7DwytmDMbzt4:-MRgLdKh7DwytmDMbzt5:btel6iq) ([raw file](https://github.com/estuary/flow/blob/e472e0fe6c992d2d4cad2406bc30a215d9112a6b/crates/driver-tester/src/tests/mod.rs#L204)):*
> ```Rust
>         let tx = tx.finish_loads().await?;
>         debug!("sent loadEOF");
>         let (rx, received_docs, always_empty) = rx.recv_all().await?;
> ```

Right, there's a deadlock lurking here -- and it's perhaps what lead to wrapping the initial future await in a tokio task? But even that is an insufficient solution.

Suppose the peer is reading each load key, loading it, and immediately sending its document:

The trouble is that we'll stuff the channel with load requests here, but the peer is also stuffing the response channel with loaded documents -- and that return channel will eventually block. Then, because the peer is no longer reading our load requests (it's blocked on sending a loaded document), *we* will block on sending load RPCs, completing the deadlock.

The only correct handling here is to concurrently drive both sending loads and receiving responses at the same time.

I think the fix is really simple: just make rx.recv_all() a tokio task, that's spawned before you start sending loads. This also _definitely_ means the transaction start future doesn't need to be it's own task -- it will be driven by the recv_all task.
___
*[crates/driver-tester/src/tests/mod.rs, line 253 at r1](https://reviewable.io/reviews/estuary/flow/88#-MRgOAcvDWuKpxSJ_qDL:-MRgOAcvDWuKpxSJ_qDM:bhqx23e) ([raw file](https://github.com/estuary/flow/blob/e472e0fe6c992d2d4cad2406bc30a215d9112a6b/crates/driver-tester/src/tests/mod.rs#L253)):*
> ```Rust
> 
>     // Track the keys we've observed in order to ensure that the response contains at most one
>     // document per distinct key. This condition seems pretty unlikely, but for now I'm erring on
> ```

I dunno, I could certainly see this mistake being made, and great to catch it here! 
___
*[crates/materialization/Cargo.toml, line 16 at r1](https://reviewable.io/reviews/estuary/flow/88#-MRgOf4h7FQS3SwJJwk9:-MRgOf4h7FQS3SwJJwkA:b-9hori6) ([raw file](https://github.com/estuary/flow/blob/e472e0fe6c992d2d4cad2406bc30a215d9112a6b/crates/materialization/Cargo.toml#L16)):*
> ```TOML
> itertools = "*"
> log = "*"
> prost-types = "0.7"
> ```

with these and other Cargo.toml's, I'd really like to keep them as "*" unless required
___
*[crates/protocol/src/lib.rs, line 76 at r1](https://reviewable.io/reviews/estuary/flow/88#-MRgPGoR5aNJNTgEOQXU:-MRgPGoR5aNJNTgEOQXV:bj6v4ab) ([raw file](https://github.com/estuary/flow/blob/e472e0fe6c992d2d4cad2406bc30a215d9112a6b/crates/protocol/src/lib.rs#L76)):*
> ```Rust
> 
>     impl CollectionExt for CollectionSpec {
>         fn get_projection(&self, field: impl AsRef<str>) -> Option<&Projection> {
> ```

I'm not sure this is carrying enough water to motivate the trait...
___
*[crates/protocol/src/lib.rs, line 89 at r1](https://reviewable.io/reviews/estuary/flow/88#-MRgPOe6-RiSnMox5Tc6:-MRgPOe6-RiSnMox5Tc7:b-4rkmxt) ([raw file](https://github.com/estuary/flow/blob/e472e0fe6c992d2d4cad2406bc30a215d9112a6b/crates/protocol/src/lib.rs#L89)):*
> ```Rust
> 
>     /// Extension trait with helper functions for working with arenas.
>     pub trait ArenaExt {
> ```

this, however, is great!
___
*[crates/runtime/src/local.rs, line 232 at r1](https://reviewable.io/reviews/estuary/flow/88#-MRgQJsp6cUkoOPzwKK0:-MRgQJsp6cUkoOPzwKK1:b-j7az3s) ([raw file](https://github.com/estuary/flow/blob/e472e0fe6c992d2d4cad2406bc30a215d9112a6b/crates/runtime/src/local.rs#L232)):*
> ```Rust
>             // if id() returns None, then it is because the process has already been awaited.
>             // It is expected that it should always return Some here, but seeing as we're just
>             // trying to stop things, we won't treat it as an error condition.
> ```

nit: I actually would prefer to assert on this.

It's a matter of wide debate, but I feel pretty firmly that if in doubt, and it's a developer bug if we're wrong (e.x. wrong control flow or mental model), crash the program! warnings like this are just so easy to miss and ignore.
___
*[crates/runtime/src/local.rs, line 296 at r1](https://reviewable.io/reviews/estuary/flow/88#-MRgQy6U3HjLlopnWPxj:-MRgQy6U3HjLlopnWPxk:b-skhzzo) ([raw file](https://github.com/estuary/flow/blob/e472e0fe6c992d2d4cad2406bc30a215d9112a6b/crates/runtime/src/local.rs#L296)):*
> ```Rust
> 
>         let lines = BufReader::new(child.stderr.take().unwrap()).lines();
>         tokio_stream::wrappers::LinesStream::new(lines).map_ok(move |l| Log::parse(name.clone(), l))
> ```

nit: if we're brining in tokio_stream just for this, I think my preference would be to drive it using the non-Steam api and avoid the dependency: https://docs.rs/tokio/1.1.0/tokio/io/struct.Lines.html#examples 

This *is* a tokio package, and we're probably going to use it at some point, but in this case i'm not sure what it's bringing (it seems to be predominantly shims for legacy code / not-quite-there integrations with `futures` ?).
___
*[crates/tuple/src/element.rs, line 27 at r1](https://reviewable.io/reviews/estuary/flow/88#-MRgRjaC4qtcV5noi_Kt:-MRgRc9CCbF7dwaIO7gy:b-jix9dt) ([raw file](https://github.com/estuary/flow/blob/e472e0fe6c992d2d4cad2406bc30a215d9112a6b/crates/tuple/src/element.rs#L27)):*
> ```Rust
> }
> 
> impl<'a> Default for Element<'a> {
> ```

Curious, where this is used ?

This crate isn't our code, it's lifted from foundationdb-rs, with bits we don't need removed to avoid bringing in a lot of dependencies.

I made only a couple of surgical changes to lib.rs, marked up with comments, so that we could preserve easy diff-ability with future updates from the upstream.

So: move to a separate module linked from lib.rs ? (e.x. as value.rs is today).

jgraettinger (changes requested):
Still not quite done, but making headway. More feedback. Thanks

> Reviewed 5 of 54 files at r1.
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/88#-:-MRjyxYm87Yp1oySOhkB:bt3gphy)* status: 34 of 54 files reviewed, 19 unresolved discussions (waiting on @psFried)

___
*[go/materialize/driver/sql/interface.go, line 27 at r1](https://reviewable.io/reviews/estuary/flow/88#-MRjgsHkA2x_b46cZNd_:-MRjgsHkA2x_b46cZNda:bqmxxq6) ([raw file](https://github.com/estuary/flow/blob/e472e0fe6c992d2d4cad2406bc30a215d9112a6b/go/materialize/driver/sql/interface.go#L27)):*
> ```Go
> // Transaction represents a database transaction, which will correspond one-to-one with a
> // transaction rpc.
> type Transaction interface {
> ```

I'm going to suggest a channel-driven interface here instead of this imperative one. E.x:

```
type Transaction struct {
  // LoadKeyCh is a channel into which keys to load are written.
  // It's closed when all load keys have been sent, and at that time
  // the Transaction must begin sending loaded documents if it hasn't already.
  // (note the Ch suffixes here are idomatic).
  LoadKeyCh chan<- tuple.Tuple

  // LoadedDocumentCh is a channel into which the transaction implementation
  // sends loaded documents, either as they're processed from LoadKeyCh,
  // or all at once upon reading a LoadKeyCh close. Once all loaded documents
  // have been sent, the channel is closed to indicate so to the driver.
  // A non-nil Error is to be considered terminal and aborts the Transaction.
  LoadedDocumentCh <-chan LoadedDocument

  // StoreDocumentCh is a channel into which StoreDocuments are written
  // by the driver. The implementation is expected to begin processing StoreDocuments
  // only after a successful close of LoadedDocumentCh.
  // The channel is closed by the driver once all StoreDocuments have been sent,
  // indicating the Transaction should commit.
  StoreDocumentCh chan<- StoreDocument
  
  // CommitCh is a channel into which a single, final commit status error is sent.
  // The implementation is expected to send a final nil or error after StoreDocumentCh
  // has closed and the Transaction has fully committed, or failed to commit.
  CommitCh <-chan error

  // Note that there isn't a channel to communicate an "abort" or rollback here.
  // That's because it already exists via the Context sent into StartTransaction.
  // That Context will be cancelled to abort the Transaction.
}

type LoadedDocument struct {
  Error error
  Document json.RawMessage
  // .. can be extended with key & value fields, etc in the future.
}

type StoreDocument struct {
 Update bool
 Key tuple.Tuple
 Values tuple.Tuple
 Document json.RawMessage
}
```

This factoring plays into Go's strengths, which is easy concurrency through channels, and using channel closes to carry semantics around EOF's. This also lets the implementation ignore details of Arenas, etc.

A final comment is that nothing in this Transaction is _about_ SQL specifically. It's more of a casting of the generalized transaction protocol into a Go form. I'd want to use this for Redis, bigtable, etc also.

I'll suggest factoring the guts of a Transaction RPC handler into a pure `runTransaction` function which takes a `pm.Driver_TransactionServer` and a generalized `Transaction` struct to drive. Then the Transaction() implementation would focus just on initial SQL-specific setup and delegate into that function.

EDIT: reading more i'm seeing how the passed values are _after_ a type-mapping has been done. That does make some sense, _but_ what about having the SQL-specific Collection.StartTransaction() interface accept a *capability* for projecting e.x. from a LoadKey / StoreDocument to the final mapped & packed []interface{} (e.x. via a provided mapper closure / interface)  ?
___
*[go/materialize/driver/sql/interface.go, line 78 at r1](https://reviewable.io/reviews/estuary/flow/88#-MRjtDBoChsI8L2-Db-q:-MRjtDBoChsI8L2-Db-r:bqjk2b1) ([raw file](https://github.com/estuary/flow/blob/e472e0fe6c992d2d4cad2406bc30a215d9112a6b/go/materialize/driver/sql/interface.go#L78)):*
> ```Go
> // ConnectionManager hands out Connections for each Handle.
> type ConnectionManager interface {
> 	// Connection returns a Connection for the given Handle.
> ```

I'm not sure what to make of both this Connection() call, and the Connection interface itself taking *Handle.

This feels to me more like a ConnectionBuilder or Factory, which ought to be the one and only place where the Connection interface touches a *Handle.

To that end, isn't ConnectionBuilder then the appropriate place to implement any desired caching that's being done to underlying endpoints ? It's able to more meaningfully examine the endpoint and coalesce underlying connections than we can do here.
___
*[go/materialize/driver/sql/interface.go, line 137 at r1](https://reviewable.io/reviews/estuary/flow/88#-MRjw5SJ7Iqsoec3TWeu:-MRjvbeq3VLycCynmYIq:b7qqujt) ([raw file](https://github.com/estuary/flow/blob/e472e0fe6c992d2d4cad2406bc30a215d9112a6b/go/materialize/driver/sql/interface.go#L137)):*
> ```Go
> // StartSession is part of the DriverServer implementation.
> func (g *GenericDriver) StartSession(ctx context.Context, req *pm.SessionRequest) (*pm.SessionResponse, error) {
> 	var handle = Handle{
> ```

just thinking out loud, as this is moved towards a generalized bag of config, we can extend the bag with Nonce but otherwise seek to pass-through structure it contains to the underlying driver.
___
*[go/materialize/driver/sql/sqlgen.go, line 77 at r1](https://reviewable.io/reviews/estuary/flow/88#-MRjwu4N7L_b7T71SqTo:-MRjwu4N7L_b7T71SqTp:bhds9od) ([raw file](https://github.com/estuary/flow/blob/e472e0fe6c992d2d4cad2406bc30a215d9112a6b/go/materialize/driver/sql/sqlgen.go#L77)):*
> ```Go
> type ParametersConverter []func(interface{}) (interface{}, error)
> 
> func (c ParametersConverter) Convert(values ...interface{}) ([]interface{}, error) {
> ```

nit: should values be a Tuple ?

Also, this looks like it's already the realization of a "conversion capability" as discussed in the other comments, which is great. Could it be passed into Connection.StartTransaction ?
___
*[go/materialize/driver/sql/sqlgen.go, line 169 at r1](https://reviewable.io/reviews/estuary/flow/88#-MRjxOX8A4Dgbf8enx6k:-MRjxOX8A4Dgbf8enx6l:b4c9lyr) ([raw file](https://github.com/estuary/flow/blob/e472e0fe6c992d2d4cad2406bc30a215d9112a6b/go/materialize/driver/sql/sqlgen.go#L169)):*
> ```Go
> }
> 
> type ResolvedColumnType struct {
> ```

:thumbsup:

psFried (commented):
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/88)* status: 34 of 54 files reviewed, 16 unresolved discussions (waiting on @jgraettinger and @psFried)

___
*[crates/derive/Cargo.toml, line 17 at r1](https://reviewable.io/reviews/estuary/flow/88#-MRgBROH76sd1V2EGvFj:-MRqAmqZDFY8yAF_FJjK:b-q85afp) ([raw file](https://github.com/estuary/flow/blob/e472e0fe6c992d2d4cad2406bc30a215d9112a6b/crates/derive/Cargo.toml#L17)):*

Any place where we're setting explicit versions (other than '*'), let's comment on exactly why we're doing it and what needs to happen before we can switch it back to '*'.

E.x., the ideal unachievable world state is that all of our crates are using '*', and we're using the Cargo.lock exclusively to pin things. Only when we can't do that for some reason should we use explicit versions here, and that calls for documenting why we're doing it.
</blockquote></details>

The reason I did this was just because I didn't see another way to update specific depencencies without running a complete `cargo update`. I've since learned that the `-p` option for cargo update allows you to do this (it does _not_ just specify a package within the workspace like it does for all the other subcommands). In any case, I'll revert all the cargo tomls to just use `*` for now. In case you weren't aware, there's an approved RFC that will make managing our dependencies _much_ better: https://rust-lang.github.io/rfcs/2906-cargo-workspace-deduplicate.html?highlight=workspace#workspace-level-dependencies.
___
*[crates/driver-tester/Cargo.toml, line 34 at r1](https://reviewable.io/reviews/estuary/flow/88#-MRgCxSD5T3-6nSES9nU:-MRqDex22gS3Qr7JcFXm:b-v0llik) ([raw file](https://github.com/estuary/flow/blob/e472e0fe6c992d2d4cad2406bc30a215d9112a6b/crates/driver-tester/Cargo.toml#L34)):*

tokio-stream seems duplicative with `futures` ? what's driving this ?
</blockquote></details>

When tokio decided to release 1.0, they removed all the `Stream` related functionality because the std Stream trait is not yet stable. A number of tokio types that used to impl the Stream trait no longer do. All the stream handling stuff that used to be in tokio has been moved into tokio-stream, including some wrappers for tokio types (`Lines` and `mpsc::Receiver` are the two I remember off hand)  to make them implement Stream. The main motivator for this is to be able to pass a `Receiver` to `DriverClient::transaction`.
___
*[crates/driver-tester/src/request.rs, line 109 at r1](https://reviewable.io/reviews/estuary/flow/88#-MRgF9Po-DXZ1kBoqjJ0:-MRqGrpA5Nlge9rNOb7S:b-hqqzs6) ([raw file](https://github.com/estuary/flow/blob/e472e0fe6c992d2d4cad2406bc30a215d9112a6b/crates/driver-tester/src/request.rs#L109)):*

nit: this is formatted ? weird how indented it is.
</blockquote></details>

yeah, I'm not sure why rustfmt thought this was ok, but I'll try wrapping some curly braces around it to see if it formats it any better that way.
___
*[crates/driver-tester/src/request.rs, line 157 at r1](https://reviewable.io/reviews/estuary/flow/88#-MRgFQaB9gHYCiOloZB6:-MRqHOcl55B5ZKkcIrJI:b9shkya) ([raw file](https://github.com/estuary/flow/blob/e472e0fe6c992d2d4cad2406bc30a215d9112a6b/crates/driver-tester/src/request.rs#L157)):*

I don't think I understand why this needs to be in a background task. Can't TrasnactionReciever::Pending just wrap this future directly, without spawning it to a task ?

In other words, why does this future need to be driven independently of calls to LoadReciever::recv() ?
</blockquote></details>

It doesn't. I'll change it to just wrap a Boxed future.
___
*[crates/driver-tester/src/request.rs, line 327 at r1](https://reviewable.io/reviews/estuary/flow/88#-MRgJ1Cm96D3KBSdh4Fa:-MRqJqD48NlQyW5GrzlN:b30ea9c) ([raw file](https://github.com/estuary/flow/blob/e472e0fe6c992d2d4cad2406bc30a215d9112a6b/crates/driver-tester/src/request.rs#L327)):*

It's because FDB tuples encode themselves without a leading byte indicator _only if they're the root-most element_. If they're non-root, they prefix with a "nested" byte indicator.

Since we're manually building a tuple here, depth is one to indicates we're "inside" the top-level tuple. Importantly, that means if that one of these elements is an array, it will write a nesting byte instead of assuming it's root-most and effectively flattening its elements up into the parent.

Another way to look at this is that if you mapped ptrs through to an extracted Vec<&Value> and packed *that*, _then_ the depth would be zero.
</blockquote></details>

Cool, thanks. I'll update the comment.
___
*[crates/driver-tester/src/tests/mod.rs, line 204 at r1](https://reviewable.io/reviews/estuary/flow/88#-MRgLdKh7DwytmDMbzt4:-MRqK89y65nKV9SB7vnE:b-2mplrf) ([raw file](https://github.com/estuary/flow/blob/e472e0fe6c992d2d4cad2406bc30a215d9112a6b/crates/driver-tester/src/tests/mod.rs#L204)):*

Right, there's a deadlock lurking here -- and it's perhaps what lead to wrapping the initial future await in a tokio task? But even that is an insufficient solution.

Suppose the peer is reading each load key, loading it, and immediately sending its document:

The trouble is that we'll stuff the channel with load requests here, but the peer is also stuffing the response channel with loaded documents -- and that return channel will eventually block. Then, because the peer is no longer reading our load requests (it's blocked on sending a loaded document), *we* will block on sending load RPCs, completing the deadlock.

The only correct handling here is to concurrently drive both sending loads and receiving responses at the same time.

I think the fix is really simple: just make rx.recv_all() a tokio task, that's spawned before you start sending loads. This also _definitely_ means the transaction start future doesn't need to be it's own task -- it will be driven by the recv_all task.
</blockquote></details>

I'm glad you brought this up, because I had forgotten to update this. I had meant to change the Sender to an unbounded channel to address that possibility, but spawning a separate task as soon as we start the transaction would also work, and might help clean up test code a bit.
___
*[crates/driver-tester/src/tests/mod.rs, line 253 at r1](https://reviewable.io/reviews/estuary/flow/88#-MRgOAcvDWuKpxSJ_qDL:-MRqLaVID49-ARltFkJO:b-8fx2u7) ([raw file](https://github.com/estuary/flow/blob/e472e0fe6c992d2d4cad2406bc30a215d9112a6b/crates/driver-tester/src/tests/mod.rs#L253)):*

I dunno, I could certainly see this mistake being made, and great to catch it here! 
</blockquote></details>

:+1: 
___
*[crates/materialization/Cargo.toml, line 16 at r1](https://reviewable.io/reviews/estuary/flow/88#-MRgOf4h7FQS3SwJJwk9:-MRqLeYS7fpt5BJ1wUSe:b-nu0hm0) ([raw file](https://github.com/estuary/flow/blob/e472e0fe6c992d2d4cad2406bc30a215d9112a6b/crates/materialization/Cargo.toml#L16)):*

with these and other Cargo.toml's, I'd really like to keep them as "*" unless required
</blockquote></details>

Will do, as discussed above.
___
*[crates/protocol/src/lib.rs, line 76 at r1](https://reviewable.io/reviews/estuary/flow/88#-MRgPGoR5aNJNTgEOQXU:-MRqMA5S2b3hReo0v1G6:b-l4wul3) ([raw file](https://github.com/estuary/flow/blob/e472e0fe6c992d2d4cad2406bc30a215d9112a6b/crates/protocol/src/lib.rs#L76)):*

I'm not sure this is carrying enough water to motivate the trait...
</blockquote></details>

haha yeah, I somehow expected there to be a lot more here than I ended up needing. I'll just remove this trait.
___
*[crates/runtime/src/local.rs, line 232 at r1](https://reviewable.io/reviews/estuary/flow/88#-MRgQJsp6cUkoOPzwKK0:-MRqMOKQCHz82DDeLqRn:bvr5mxu) ([raw file](https://github.com/estuary/flow/blob/e472e0fe6c992d2d4cad2406bc30a215d9112a6b/crates/runtime/src/local.rs#L232)):*

nit: I actually would prefer to assert on this.

It's a matter of wide debate, but I feel pretty firmly that if in doubt, and it's a developer bug if we're wrong (e.x. wrong control flow or mental model), crash the program! warnings like this are just so easy to miss and ignore.
</blockquote></details>

Yeah that's fair. In retrospect, the warning was just me hedging on my understanding of the tokio docs, and I'd rather just crash if I misunderstood those.
___
*[crates/runtime/src/local.rs, line 296 at r1](https://reviewable.io/reviews/estuary/flow/88#-MRgQy6U3HjLlopnWPxj:-MRqMy6jE8IwCytpVpmw:bxojzcp) ([raw file](https://github.com/estuary/flow/blob/e472e0fe6c992d2d4cad2406bc30a215d9112a6b/crates/runtime/src/local.rs#L296)):*

nit: if we're brining in tokio_stream just for this, I think my preference would be to drive it using the non-Steam api and avoid the dependency: https://docs.rs/tokio/1.1.0/tokio/io/struct.Lines.html#examples 

This *is* a tokio package, and we're probably going to use it at some point, but in this case i'm not sure what it's bringing (it seems to be predominantly shims for legacy code / not-quite-there integrations with `futures` ?).
</blockquote></details>

Yeah, the real motivation for tokio-stream is for use with the `DriverClient`. We really don't need it in the `runtime` crate, so I'll just rewrite this as a loop.
___
*[crates/tuple/src/element.rs, line 27 at r1](https://reviewable.io/reviews/estuary/flow/88#-MRgRjaC4qtcV5noi_Kt:-MRqOhOJ-sMcj02dzJhH:bebfbxt) ([raw file](https://github.com/estuary/flow/blob/e472e0fe6c992d2d4cad2406bc30a215d9112a6b/crates/tuple/src/element.rs#L27)):*

Curious, where this is used ?

This crate isn't our code, it's lifted from foundationdb-rs, with bits we don't need removed to avoid bringing in a lot of dependencies.

I made only a couple of surgical changes to lib.rs, marked up with comments, so that we could preserve easy diff-ability with future updates from the upstream.

So: move to a separate module linked from lib.rs ? (e.x. as value.rs is today).
</blockquote></details>

Removed. This was vestigial, but I can't remember exactly why at the moment. Some container type that wanted elements to impl Default, maybe?
___
*[go/materialize/driver/sql/interface.go, line 27 at r1](https://reviewable.io/reviews/estuary/flow/88#-MRjgsHkA2x_b46cZNd_:-MRqP-vv0upy5u_TK2Lq:bxaz59i) ([raw file](https://github.com/estuary/flow/blob/e472e0fe6c992d2d4cad2406bc30a215d9112a6b/go/materialize/driver/sql/interface.go#L27)):*

I'm going to suggest a channel-driven interface here instead of this imperative one. E.x:

```
type Transaction struct {
  // LoadKeyCh is a channel into which keys to load are written.
  // It's closed when all load keys have been sent, and at that time
  // the Transaction must begin sending loaded documents if it hasn't already.
  // (note the Ch suffixes here are idomatic).
  LoadKeyCh chan<- tuple.Tuple

  // LoadedDocumentCh is a channel into which the transaction implementation
  // sends loaded documents, either as they're processed from LoadKeyCh,
  // or all at once upon reading a LoadKeyCh close. Once all loaded documents
  // have been sent, the channel is closed to indicate so to the driver.
  // A non-nil Error is to be considered terminal and aborts the Transaction.
  LoadedDocumentCh <-chan LoadedDocument

  // StoreDocumentCh is a channel into which StoreDocuments are written
  // by the driver. The implementation is expected to begin processing StoreDocuments
  // only after a successful close of LoadedDocumentCh.
  // The channel is closed by the driver once all StoreDocuments have been sent,
  // indicating the Transaction should commit.
  StoreDocumentCh chan<- StoreDocument
  
  // CommitCh is a channel into which a single, final commit status error is sent.
  // The implementation is expected to send a final nil or error after StoreDocumentCh
  // has closed and the Transaction has fully committed, or failed to commit.
  CommitCh <-chan error

  // Note that there isn't a channel to communicate an "abort" or rollback here.
  // That's because it already exists via the Context sent into StartTransaction.
  // That Context will be cancelled to abort the Transaction.
}

type LoadedDocument struct {
  Error error
  Document json.RawMessage
  // .. can be extended with key & value fields, etc in the future.
}

type StoreDocument struct {
 Update bool
 Key tuple.Tuple
 Values tuple.Tuple
 Document json.RawMessage
}
```

This factoring plays into Go's strengths, which is easy concurrency through channels, and using channel closes to carry semantics around EOF's. This also lets the implementation ignore details of Arenas, etc.

A final comment is that nothing in this Transaction is _about_ SQL specifically. It's more of a casting of the generalized transaction protocol into a Go form. I'd want to use this for Redis, bigtable, etc also.

I'll suggest factoring the guts of a Transaction RPC handler into a pure `runTransaction` function which takes a `pm.Driver_TransactionServer` and a generalized `Transaction` struct to drive. Then the Transaction() implementation would focus just on initial SQL-specific setup and delegate into that function.

EDIT: reading more i'm seeing how the passed values are _after_ a type-mapping has been done. That does make some sense, _but_ what about having the SQL-specific Collection.StartTransaction() interface accept a *capability* for projecting e.x. from a LoadKey / StoreDocument to the final mapped & packed []interface{} (e.x. via a provided mapper closure / interface)  ?
</blockquote></details>

This factoring was actually motivated by wanting to copy data directly from the database buffer into the response arena, which I don't think would be doable with a channel-driven approach. I can add a comment, if you agree that's a good enough reason to keep it more or less as is.

In general, I agree with your idea that none of this seems specific to SQL, and would likely be usable for a variety of other systems with a little tweaking. I have some reservations, though. In terms of an approach, I'd rather not try to generalize it too much now, and wait until we learn more about those other systems and client libraries. Would you be OK with deferring that work? I could also make these "generic" types private for now, since them being public gives the impression that these interfaces are more fully baked than they are.
___
*[go/materialize/driver/sql/interface.go, line 78 at r1](https://reviewable.io/reviews/estuary/flow/88#-MRjtDBoChsI8L2-Db-q:-MRqUAlf0nqK1IdLfoUe:bdwzy9u) ([raw file](https://github.com/estuary/flow/blob/e472e0fe6c992d2d4cad2406bc30a215d9112a6b/go/materialize/driver/sql/interface.go#L78)):*

I'm not sure what to make of both this Connection() call, and the Connection interface itself taking *Handle.

This feels to me more like a ConnectionBuilder or Factory, which ought to be the one and only place where the Connection interface touches a *Handle.

To that end, isn't ConnectionBuilder then the appropriate place to implement any desired caching that's being done to underlying endpoints ? It's able to more meaningfully examine the endpoint and coalesce underlying connections than we can do here.
</blockquote></details>

I think I'll need to think about this some more... Initially, this interface only accepted the URI, not the whole handle. And I thought it better to pass the whole handle to allow more flexibility in how connections are managed (e.g. maybe there's a reason to prefer separate connections for each table in some system, or for each unique Nonce value?). I think I like the idea of making this the one place where we pass the handle, though.
___
*[go/materialize/driver/sql/interface.go, line 137 at r1](https://reviewable.io/reviews/estuary/flow/88#-MRjw5SJ7Iqsoec3TWeu:-MRqW1AG8Qz8d9quPwGm:bze7880) ([raw file](https://github.com/estuary/flow/blob/e472e0fe6c992d2d4cad2406bc30a215d9112a6b/go/materialize/driver/sql/interface.go#L137)):*

just thinking out loud, as this is moved towards a generalized bag of config, we can extend the bag with Nonce but otherwise seek to pass-through structure it contains to the underlying driver.
</blockquote></details>

I'm honestly still mulling over how to deal with generalizing the connection info here, especially given the lack of generics. For now, I think it makes sense to hold off on designing for that, since I'm still unsure about just how generic this `GenericDriver` really is. Basically, just echoing my comment above about wanting to hold off on doing too much design work on these interfaces.
___
*[go/materialize/driver/sql/sqlgen.go, line 77 at r1](https://reviewable.io/reviews/estuary/flow/88#-MRjwu4N7L_b7T71SqTo:-MRqabzY7C-qNsIpxWz7:befm2i9) ([raw file](https://github.com/estuary/flow/blob/e472e0fe6c992d2d4cad2406bc30a215d9112a6b/go/materialize/driver/sql/sqlgen.go#L77)):*

nit: should values be a Tuple ?

Also, this looks like it's already the realization of a "conversion capability" as discussed in the other comments, which is great. Could it be passed into Connection.StartTransaction ?
</blockquote></details>

I wish I could remember why I had switched this to accept `[]interface{}` instead of a tuple. I feel like it was maybe some confusion about go typedefs or something, but I dunno anymore. At this point, accepting Tuple seems better, since it may allow us to avoid iterating the tuples twice to convert the values, so I'll try it out and see.

I don't see any reason we couldn't pass the converters to StartTransaction, but I also don't think I see the benefit. At this point it seems likely that every driver would simply pass it through on the transaction struct and do the exact same conversion steps with it. My opinion right now is that we should defer this particular design question until we have something more concrete in front of us. Does that make sense to you?
___
*[go/materialize/driver/sql/sqlgen.go, line 169 at r1](https://reviewable.io/reviews/estuary/flow/88#-MRjxOX8A4Dgbf8enx6k:-MRqka8a626gdJWNrpIE:b-8fx2u7) ([raw file](https://github.com/estuary/flow/blob/e472e0fe6c992d2d4cad2406bc30a215d9112a6b/go/materialize/driver/sql/sqlgen.go#L169)):*

:thumbsup: 
</blockquote></details>

:+1:

jgraettinger (changes requested):
> Reviewed 20 of 54 files at r1.
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/88)* status: all files reviewed, 13 unresolved discussions (waiting on @jgraettinger and @psFried)

___
*[crates/derive/Cargo.toml, line 17 at r1](https://reviewable.io/reviews/estuary/flow/88#-MRgBROH76sd1V2EGvFj:-MRtnflXC7sdpjCEPp28:b-bedpy7) ([raw file](https://github.com/estuary/flow/blob/e472e0fe6c992d2d4cad2406bc30a215d9112a6b/crates/derive/Cargo.toml#L17)):*

The reason I did this was just because I didn't see another way to update specific depencencies without running a complete `cargo update`. I've since learned that the `-p` option for cargo update allows you to do this (it does _not_ just specify a package within the workspace like it does for all the other subcommands). In any case, I'll revert all the cargo tomls to just use `*` for now. In case you weren't aware, there's an approved RFC that will make managing our dependencies _much_ better: https://rust-lang.github.io/rfcs/2906-cargo-workspace-deduplicate.html?highlight=workspace#workspace-level-dependencies.
</blockquote></details>

Thanks. That RFC looks awesome.
___
*[go/materialize/driver/sql/interface.go, line 27 at r1](https://reviewable.io/reviews/estuary/flow/88#-MRjgsHkA2x_b46cZNd_:-MRtoG2sEGfdORt-koR7:b-h01wj5) ([raw file](https://github.com/estuary/flow/blob/e472e0fe6c992d2d4cad2406bc30a215d9112a6b/go/materialize/driver/sql/interface.go#L27)):*
> copy data directly from the database buffer into the response arena

I talk about this in a new comment further down, but this is not buying any performance improvement relative to scanning into a `[]byte` or `json.RawMessage` and handing that over through a `LoadedDocument`. It's still the instance allocated by the database driver during `Scan()`, which is the best you can do.

More broadly:

I'm A-OK with not worry about generalizing right now (was more just pointing out that the proposed interface happened to be nearly there already), but I do have significant reservations with the current interface that mostly stem from how it synchronously mixes handling of what are fundamentally full-duplex rx & tx channels of the transaction. IME this is such a common source of deadlocks and issues that I've evolved a rule of life to always always handle tx and rx as separate decoupled tasks.

I've noted one possible deadlock below. Maybe it's not, and perhaps this interface is A-OK for production. But I have a relatively harder time getting comfortable with it vis-a-vis a decoupled task approach, and do think the onus is on proving why this interface is safe and even _required_ over the alternative.

Even _if_ it turned out that there's indeed a meaty performance improvement by having the driver load into an Arena, I'd _still_ push for a channel approach where the rx task controls the allocation of Arenas, and hands over loaded chunks instead of individual documents.
___
*[go/materialize/driver/sql/interface.go, line 263 at r1](https://reviewable.io/reviews/estuary/flow/88#-MRtqM3Q7hcLpQRXikDR:-MRtqM3Q7hcLpQRXikDS:bc4spkp) ([raw file](https://github.com/estuary/flow/blob/e472e0fe6c992d2d4cad2406bc30a215d9112a6b/go/materialize/driver/sql/interface.go#L263)):*
> ```Go
> 		req, err = stream.Recv()
> 		if err != nil {
> 			return err // EOF is not expected here, so we'd return is as an error.
> ```

nit: add context with fmt.Error(). It can also be mapped into io.UnexpectedEOF.
___
*[go/materialize/driver/sql/interface.go, line 277 at r1](https://reviewable.io/reviews/estuary/flow/88#-MRtqaex6RTAYdFTPGSa:-MRtqaex6RTAYdFTPGSb:b-ulmq5a) ([raw file](https://github.com/estuary/flow/blob/e472e0fe6c992d2d4cad2406bc30a215d9112a6b/go/materialize/driver/sql/interface.go#L277)):*
> ```Go
> 					return err
> 				}
> 				err = transaction.AddLoadKey(ctx, args)
> ```

here's the deadlock i'm talking about above. What happens if the driver tx channel gets stuffed in between here, and when we finally PollLoadResults below ?
___
*[go/materialize/driver/sql/standard.go, line 103 at r1](https://reviewable.io/reviews/estuary/flow/88#-MRtjZ4A4epYhUHliAE5:-MRtjZ4A4epYhUHliAE6:b-l4ggm) ([raw file](https://github.com/estuary/flow/blob/e472e0fe6c992d2d4cad2406bc30a215d9112a6b/go/materialize/driver/sql/standard.go#L103)):*
> ```Go
> 
> // Scan implements sql.Scanner for ArenaAppender
> func (a *ArenaAppender) Scan(src interface{}) error {
> ```

Note that this isn't "directly into" the arena. The driver interface is allocating a []byte which it's scanning into, and which is then being copied into the arena. It'd be just as efficient to scan from the driver into a []byte or string, and later hand that interface{} to the arena.

IME Go network protocols pretty closely mirror the typical Rust pattern of "read into an append-only owned BytesMut, then split off chunks of read Bytes and hand to clients" . The predominant patterns are to either read off the network into a owned append-only buffer, and hand off chunks of that buffer to the client as it parses and understands the framing. Or to have a re-usable internal buffer and to explicitly copy before handing clients []byte frames.

To my knowledge there are _no_ interfaces that actually read directly off the network into a client provided buffer. The `http` package interface _appears_ to do this but is actually lying, there's a copy under the hood.
___
*[go/protocols/materialize/materialize.proto, line 259 at r1](https://reviewable.io/reviews/estuary/flow/88#-MRtlz9kC9Muz_vW2_d0:-MRtlz9kC9Muz_vW2_d1:b4a65ji) ([raw file](https://github.com/estuary/flow/blob/e472e0fe6c992d2d4cad2406bc30a215d9112a6b/go/protocols/materialize/materialize.proto#L259)):*
> ```protobuf
>     // Byte arena of the request.
>     bytes arena = 1 [ (gogoproto.casttype) = "github.com/estuary/flow/go/protocols/flow.Arena" ];
>     // Loaded JSON documents, 1:1 with keys of the LoadRequest.
> ```

I'm thinking that we shouldn't require the driver to track this 1:1 correspondence.

Imagine a driver implementation which spawns of a goroutine for each load which performs a point lookup. Those results could come back in any order. The runtime is happy for them to come back in any order, and doesn't need an explicit NACK for keys that aren't found.

Thus we shouldn't require that the driver re-sequence them.
___
*[go/protocols/materialize/materialize_extensions.go, line 30 at r1](https://reviewable.io/reviews/estuary/flow/88#-MRtnIqgCAxpAW7B5lPl:-MRtnIqgCAxpAW7B5lPm:b-tiqqy7) ([raw file](https://github.com/estuary/flow/blob/e472e0fe6c992d2d4cad2406bc30a215d9112a6b/go/protocols/materialize/materialize_extensions.go#L30)):*
> ```Go
> // Equal returns true if the FieldSelections both represent the same selection of fields in the same
> // order.
> func (f FieldSelection) Equal(other *FieldSelection) bool {
> ```

In my branch I introduce a bag of per-field configuration. What would be the handling here?
___
*[go/materialize/driver/sql/sqlgen.go, line 77 at r1](https://reviewable.io/reviews/estuary/flow/88#-MRjwu4N7L_b7T71SqTo:-MRtuGK3Az3AwyPd8cWc:bnp0aga) ([raw file](https://github.com/estuary/flow/blob/e472e0fe6c992d2d4cad2406bc30a215d9112a6b/go/materialize/driver/sql/sqlgen.go#L77)):*

I wish I could remember why I had switched this to accept `[]interface{}` instead of a tuple. I feel like it was maybe some confusion about go typedefs or something, but I dunno anymore. At this point, accepting Tuple seems better, since it may allow us to avoid iterating the tuples twice to convert the values, so I'll try it out and see.

I don't see any reason we couldn't pass the converters to StartTransaction, but I also don't think I see the benefit. At this point it seems likely that every driver would simply pass it through on the transaction struct and do the exact same conversion steps with it. My opinion right now is that we should defer this particular design question until we have something more concrete in front of us. Does that make sense to you?
</blockquote></details>

The reason would be to avoid coupling converter semantics with the proposed Transaction 
interface, above. Invoking the converter is a one-liner, so it doesn't seem a problem to have each driver do it.

But as discussed above, I'm cool with not worrying about what a generalized interface might look like right now, if you'd rather pass post-conversion arguments.

psFried (commented):
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/88)* status: 40 of 53 files reviewed, 10 unresolved discussions (waiting on @jgraettinger)

___
*[crates/protocol/src/lib.rs, line 89 at r1](https://reviewable.io/reviews/estuary/flow/88#-MRgPOe6-RiSnMox5Tc6:-MS57X3r2z-_zpV46q8S:b-896fix) ([raw file](https://github.com/estuary/flow/blob/e472e0fe6c992d2d4cad2406bc30a215d9112a6b/crates/protocol/src/lib.rs#L89)):*

this, however, is great!
</blockquote></details>

Done.
___
*[go/materialize/driver/sql/interface.go, line 27 at r1](https://reviewable.io/reviews/estuary/flow/88#-MRjgsHkA2x_b46cZNd_:-MS5BcHpDajQ6RDI7w5e:b-896fix) ([raw file](https://github.com/estuary/flow/blob/e472e0fe6c992d2d4cad2406bc30a215d9112a6b/go/materialize/driver/sql/interface.go#L27)):*

> copy data directly from the database buffer into the response arena

I talk about this in a new comment further down, but this is not buying any performance improvement relative to scanning into a `[]byte` or `json.RawMessage` and handing that over through a `LoadedDocument`. It's still the instance allocated by the database driver during `Scan()`, which is the best you can do.

More broadly:

I'm A-OK with not worry about generalizing right now (was more just pointing out that the proposed interface happened to be nearly there already), but I do have significant reservations with the current interface that mostly stem from how it synchronously mixes handling of what are fundamentally full-duplex rx & tx channels of the transaction. IME this is such a common source of deadlocks and issues that I've evolved a rule of life to always always handle tx and rx as separate decoupled tasks.

I've noted one possible deadlock below. Maybe it's not, and perhaps this interface is A-OK for production. But I have a relatively harder time getting comfortable with it vis-a-vis a decoupled task approach, and do think the onus is on proving why this interface is safe and even _required_ over the alternative.

Even _if_ it turned out that there's indeed a meaty performance improvement by having the driver load into an Arena, I'd _still_ push for a channel approach where the rx task controls the allocation of Arenas, and hands over loaded chunks instead of individual documents.
</blockquote></details>

Done.
___
*[go/materialize/driver/sql/interface.go, line 78 at r1](https://reviewable.io/reviews/estuary/flow/88#-MRjtDBoChsI8L2-Db-q:-MS52t-r-Zg_KgKmLZyW:bswx6dd) ([raw file](https://github.com/estuary/flow/blob/e472e0fe6c992d2d4cad2406bc30a215d9112a6b/go/materialize/driver/sql/interface.go#L78)):*

I think I'll need to think about this some more... Initially, this interface only accepted the URI, not the whole handle. And I thought it better to pass the whole handle to allow more flexibility in how connections are managed (e.g. maybe there's a reason to prefer separate connections for each table in some system, or for each unique Nonce value?). I think I like the idea of making this the one place where we pass the handle, though.
</blockquote></details>

I think the best thing here is to simply revert to having ConnectionManager (or whatever it's named) accept only the URI instead of the whole handle. At the moment, cacheing of connections is done solely by URI, and we'd lose the ability to do that generically if we made Connection fundamentally specific to a single Handle. Reverting to just the URI at least makes it clear that a Connection is _for_ a URI, and is not specific to a nonce or table. 
___
*[go/materialize/driver/sql/interface.go, line 277 at r1](https://reviewable.io/reviews/estuary/flow/88#-MRtqaex6RTAYdFTPGSa:-MS504oB7vfuf3k4k8aB:b-5wh6w1) ([raw file](https://github.com/estuary/flow/blob/e472e0fe6c992d2d4cad2406bc30a215d9112a6b/go/materialize/driver/sql/interface.go#L277)):*

here's the deadlock i'm talking about above. What happens if the driver tx channel gets stuffed in between here, and when we finally PollLoadResults below ?
</blockquote></details>

Kind of a moot point now, but neither implementation can deadlock due to that scenario because we don't actually send anything over the network unless we're going to read it immediately. The sqlite implementation kept the keys in memory and executed the query when PollLoadResults was called.
___
*[go/materialize/driver/sql/standard.go, line 103 at r1](https://reviewable.io/reviews/estuary/flow/88#-MRtjZ4A4epYhUHliAE5:-MRu2EJ35ckQmcpsnO4L:bu7t5v2) ([raw file](https://github.com/estuary/flow/blob/e472e0fe6c992d2d4cad2406bc30a215d9112a6b/go/materialize/driver/sql/standard.go#L103)):*

Note that this isn't "directly into" the arena. The driver interface is allocating a []byte which it's scanning into, and which is then being copied into the arena. It'd be just as efficient to scan from the driver into a []byte or string, and later hand that interface{} to the arena.

IME Go network protocols pretty closely mirror the typical Rust pattern of "read into an append-only owned BytesMut, then split off chunks of read Bytes and hand to clients" . The predominant patterns are to either read off the network into a owned append-only buffer, and hand off chunks of that buffer to the client as it parses and understands the framing. Or to have a re-usable internal buffer and to explicitly copy before handing clients []byte frames.

To my knowledge there are _no_ interfaces that actually read directly off the network into a client provided buffer. The `http` package interface _appears_ to do this but is actually lying, there's a copy under the hood.
</blockquote></details>

I was inferring that just based on the Scanner docs, which say that the `[]byte` is "owned by the driver" and "should not be retained". I'd think that the driver wouldn't be doing an extra copy if it's just handing off the `[]byte` to a function that's not supposed to retain any references to it. But in any case, I don't feel that this optimization is really worth trying to squeeze into the channel-based transaction API at this stage.
___
*[go/protocols/materialize/materialize.proto, line 259 at r1](https://reviewable.io/reviews/estuary/flow/88#-MRtlz9kC9Muz_vW2_d0:-MS55Qec7V-fKu6x11BH:btvlwle) ([raw file](https://github.com/estuary/flow/blob/e472e0fe6c992d2d4cad2406bc30a215d9112a6b/go/protocols/materialize/materialize.proto#L259)):*

I'm thinking that we shouldn't require the driver to track this 1:1 correspondence.

Imagine a driver implementation which spawns of a goroutine for each load which performs a point lookup. Those results could come back in any order. The runtime is happy for them to come back in any order, and doesn't need an explicit NACK for keys that aren't found.

Thus we shouldn't require that the driver re-sequence them.
</blockquote></details>

This was an outdated comment. It's fixed now. The behavior is as you described.
___
*[go/protocols/materialize/materialize_extensions.go, line 30 at r1](https://reviewable.io/reviews/estuary/flow/88#-MRtnIqgCAxpAW7B5lPl:-MS55k4xCjeuGiXZ-7Cf:b-er8r6z) ([raw file](https://github.com/estuary/flow/blob/e472e0fe6c992d2d4cad2406bc30a215d9112a6b/go/protocols/materialize/materialize_extensions.go#L30)):*

In my branch I introduce a bag of per-field configuration. What would be the handling here?
</blockquote></details>

Ultimately, I think we should just move this function from materialize_extensions into a driver-specific package, since that's what really determines whether a given pair of FieldSelections are _compatible_. Although, it might be most practical to simply define this function based on strict json equality (of each configuration bag), since that's almost certainly what any of our sql drivers will be doing anyway. I'm kinda leaning that way for now, personally.

jgraettinger (approved):
This looks really great, thanks for your patience on requested changes & the delay. Comments below are all minor, and suggestions for ways to clean things up, etc. LGTM to land

> Reviewed 13 of 14 files at r2.
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/88#-:-MSeyfEX6OSzdWX-YJbz:bizz7ez)* status: all files reviewed, 8 unresolved discussions (waiting on @psFried)

___
*[crates/driver-tester/src/tests/mod.rs, line 204 at r1](https://reviewable.io/reviews/estuary/flow/88#-MRgLdKh7DwytmDMbzt4:-MSel9IEEHRA4qE9Rqyr:b-cbucf7) ([raw file](https://github.com/estuary/flow/blob/e472e0fe6c992d2d4cad2406bc30a215d9112a6b/crates/driver-tester/src/tests/mod.rs#L204)):*

I'm glad you brought this up, because I had forgotten to update this. I had meant to change the Sender to an unbounded channel to address that possibility, but spawning a separate task as soon as we start the transaction would also work, and might help clean up test code a bit.
</blockquote></details>

looks great! :thumbsup: 
___
*[go/materialize/driver/sql/interface.go, line 20 at r2](https://reviewable.io/reviews/estuary/flow/88#-MSelZpc4_cihr1WAFns:-MSelZpc4_cihr1WAFnt:bmeny4d) ([raw file](https://github.com/estuary/flow/blob/9371e76c206c390521c402f6277e43772214433a/go/materialize/driver/sql/interface.go#L20)):*
> ```Go
> 	// required in order to connect to the databases.
> 	// The sqlite driver
> 	_ "github.com/mattn/go-sqlite3"
> ```

nit: could this be moved to a _test.go now ?
___
*[go/materialize/driver/sql/interface.go, line 251 at r2](https://reviewable.io/reviews/estuary/flow/88#-MSeovpB6Q5_U4GDEZMS:-MSeovpB6Q5_U4GDEZMT:b1i4ra8) ([raw file](https://github.com/estuary/flow/blob/9371e76c206c390521c402f6277e43772214433a/go/materialize/driver/sql/interface.go#L251)):*
> ```Go
> 				transaction.StoreDocumentCh = nil
> 			}
> 		}
> ```

nit: why not send |retErr| into |errorCh| right here, and not have the wrapping `handleTxnRecvStream` at all ?
___
*[go/materialize/driver/sql/interface.go, line 270 at r2](https://reviewable.io/reviews/estuary/flow/88#-MSeoRdd8kU80iN-xN_c:-MSeoRdd8kU80iN-xN_d:ba1ozc9) ([raw file](https://github.com/estuary/flow/blob/9371e76c206c390521c402f6277e43772214433a/go/materialize/driver/sql/interface.go#L270)):*
> ```Go
> 				select {
> 				case transaction.LoadKeyCh <- tup:
> 					// we sent the key
> ```

nit: these comments are self apparent / idiomatic. remove ?
___
*[go/materialize/driver/sql/interface.go, line 473 at r2](https://reviewable.io/reviews/estuary/flow/88#-MSeqRKU1ZlvcLfeQ06J:-MSeqRKU1ZlvcLfeQ06K:bswtltp) ([raw file](https://github.com/estuary/flow/blob/9371e76c206c390521c402f6277e43772214433a/go/materialize/driver/sql/interface.go#L473)):*
> ```Go
> 
> 	select {
> 	case recvErr := <-recvCompleteCh:
> ```

A nice way to clean this up is to loop the select while either |sendCompleteCh| or |recvCompleteCh| is non-nil. Then, when the respective channels select, return early if err != nil but otherwise set the channel itself to nil.

This works because a select of a nil channel will never happen, so by nil-ing one of the channels it effectively transitions to a blocking select of the other. (And it will indeed deadlock if you don't test that at least one channel is non-nil before entering the select)
___
*[go/materialize/driver/sql/standard.go, line 103 at r1](https://reviewable.io/reviews/estuary/flow/88#-MRtjZ4A4epYhUHliAE5:-MSet2PY6sVvbmBkH-Of:b-bv5wpu) ([raw file](https://github.com/estuary/flow/blob/e472e0fe6c992d2d4cad2406bc30a215d9112a6b/go/materialize/driver/sql/standard.go#L103)):*

I was inferring that just based on the Scanner docs, which say that the `[]byte` is "owned by the driver" and "should not be retained". I'd think that the driver wouldn't be doing an extra copy if it's just handing off the `[]byte` to a function that's not supposed to retain any references to it. But in any case, I don't feel that this optimization is really worth trying to squeeze into the channel-based transaction API at this stage.
</blockquote></details>

You're 100% right here, my mental model of Scanner behavior was straight up wrong. Sorry.

Sounds good. I think we could certainly work this optimization back in later by having `LoadedDocumentCh` send instances of ArenaAppender instead of individual documents (with perhaps a bit nicer of a wrapper for dealing with chunking etc currently done by the rx load task).
___
*[go/materialize/driver/sql/standard.go, line 127 at r2](https://reviewable.io/reviews/estuary/flow/88#-MSevsR99LsOjN3fotRz:-MSevsR99LsOjN3fotS-:bhpu35p) ([raw file](https://github.com/estuary/flow/blob/9371e76c206c390521c402f6277e43772214433a/go/materialize/driver/sql/standard.go#L127)):*
> ```Go
> // implementing the sql.Scanner interface.
> type ArenaAppender struct {
> 	arena  *pf.Arena
> ```

nit: i don't believe pf.Arena should be a pointer ?

This may be dead code now. But IMO still holds water as a concept, given conversation below .
___
*[go/materialize/driver/sql/standard.go, line 203 at r2](https://reviewable.io/reviews/estuary/flow/88#-MSewjlwCesJSwpAjdVZ:-MSewjlx6qaj_lJjj6y1:b92i01k) ([raw file](https://github.com/estuary/flow/blob/9371e76c206c390521c402f6277e43772214433a/go/materialize/driver/sql/standard.go#L203)):*
> ```Go
> 
> func (t *standardSqlStoreTransaction) runStoreTransaction() (retErr error) {
> 	var commit = false
> ```

nit: this flag is a bit wonky because you're having to deal with the combined possibility space of |commit| and also |retErr|.

What I would recommend is, instead of doing a range over the channel, have a for loop which breaks only on an error or on clean exit.

Then, use `doc, ok = <-storeDocumentCh` and treat !ok as an error (e.g., return io.ErrUnexpectedEOF or a fmt.Errorf that's consolidated with other errors in the defer handling)
___
*[go/protocols/materialize/materialize_extensions.go, line 30 at r1](https://reviewable.io/reviews/estuary/flow/88#-MRtnIqgCAxpAW7B5lPl:-MSeyMmS1flvVb5fH1CC:bf7noqj) ([raw file](https://github.com/estuary/flow/blob/e472e0fe6c992d2d4cad2406bc30a215d9112a6b/go/protocols/materialize/materialize_extensions.go#L30)):*

Ultimately, I think we should just move this function from materialize_extensions into a driver-specific package, since that's what really determines whether a given pair of FieldSelections are _compatible_. Although, it might be most practical to simply define this function based on strict json equality (of each configuration bag), since that's almost certainly what any of our sql drivers will be doing anyway. I'm kinda leaning that way for now, personally.
</blockquote></details>

that sounds fine to me too.

PR #89, opened 2021-01-22
`models` crate

jgraettinger:
This PR implements is a complete re-working of how we model, load, validate, and persist entities of the Flow catalog domain.

It's a parallel implementation (for now) in a separate `models` crate.
 * `models::tables` are in-memory tables which model all of Flow entities and concepts, from "raw inputs" (e.x. resource byte contents) through to finished products (e.x. built collection specifications). The module uses macros that make it easy to define / change table models, and implementations know how to round-trip themselves into a SQLite database.
 * `models::source` defines our source YAML specifications, and implements a `Loader` that concurrently walks sources and maps each into related table rows (including errors), using an abstracted "fetch" API.
 * `models::validation` implements comprehensive validation & integrity checking of source tables, invokes driver validation APIs (again, using a generic callback), and returns "finished" specifications for use by the runtime (which are themselves tables).

Issue #68

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/89)


psFried (commented):
So far, this is looking great, but I'm still only part way through it. I'll pick it back up soon and finish it out.

> Reviewed 2 of 2 files at r1, 8 of 8 files at r2, 3 of 3 files at r3, 2 of 2 files at r4, 24 of 24 files at r5.
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/89#-:-MRgoS5J15vSaQbQyIoV:bl5nkv4)* status: 39 of 85 files reviewed, 1 unresolved discussion (waiting on @jgraettinger)

___
*[crates/doc/src/inference.rs, line 1036 at r1](https://reviewable.io/reviews/estuary/flow/89#-MRfm8wr4OBe1mUf-Lq6:-MRfm8wr4OBe1mUf-Lq7:bev87ha) ([raw file](https://github.com/estuary/flow/blob/991fb94d34c73cc184ae0bd0770ceec11e8c446b/crates/doc/src/inference.rs#L1036)):*
> ```Rust
>         } else {
>             // We deliberately don't handle the "/-" end-of-array case.
>             // It's returned by locations, below, but we don't allow an
> ```

Shout out for some good, informative comments :1st_place_medal: 
___
*[crates/models/src/tables/macros.rs, line 44 at r4](https://reviewable.io/reviews/estuary/flow/89#-MRfwEC90oN8SkfO7bgI:-MRfwECA1lLyK9gfkANy:beoerwb) ([raw file](https://github.com/estuary/flow/blob/195b4f0a3e1cf6e800989652242c02d76131de7b/crates/models/src/tables/macros.rs#L44)):*
> ```Rust
> 
> /// Trait for accepting arguments which may be owned, or can be cloned.
> pub trait OwnOrClone<T> {
> ```

I feel like there should be a way to express this succinctly with std library traits, but the best I could come up with is `param: impl Into<Cow<'_, T>>` and then calling `param.into_owned()`. But I like how `OwnOrClone` reads nicer, so up to you whether you want to keep it or use Cow.
___
*[crates/models/src/validation/mod.rs, line 1 at r6](https://reviewable.io/reviews/estuary/flow/89#-MRgnWRb7SYRjtyiWDv0:-MRgnWRb7SYRjtyiWDv1:buwhm83) ([raw file](https://github.com/estuary/flow/blob/35a79b79bb7a4b0a186979874996f835951b4734/crates/models/src/validation/mod.rs#L1)):*
> ```Rust
> use crate::{collate, source, tables};
> ```

this is a biggun. I don't see any issues in here, but at this point my eyes are pretty tired, so I'm making a note here so I can pick it back up later.
___
*[crates/models/src/validation/mod.rs, line 1192 at r6](https://reviewable.io/reviews/estuary/flow/89#-MRglQsB49EURBZlBt0Y:-MRglQsD78chcFB6_Pmj:bmfei3s) ([raw file](https://github.com/estuary/flow/blob/35a79b79bb7a4b0a186979874996f835951b4734/crates/models/src/validation/mod.rs#L1192)):*
> ```Rust
>         ) {
>             // Selector / driver constraints conflict internally:
>             (true, true, _) => Err(format!("field is both included and excluded by selector")),
> ```

Gotta love that exhaustive pattern matching check in the compiler!

psFried (approved):
> Reviewed 43 of 43 files at r6, 3 of 3 files at r7.
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/89)* status: all files reviewed, 1 unresolved discussion (waiting on @jgraettinger)

jgraettinger (commented):
Thanks! Finally merging this, and will next pick off some new PRs from the massive `catalog-build` branch (which finally has all crates working & passing Rust tests, at least).

> *[Reviewable](https://reviewable.io/reviews/estuary/flow/89#-:-MSdhs43AZVYbqvaH-ps:bdhfvhu)* status: all files reviewed, 1 unresolved discussion (waiting on @psFried)

___
*[crates/models/src/tables/macros.rs, line 44 at r4](https://reviewable.io/reviews/estuary/flow/89#-MRfwEC90oN8SkfO7bgI:-MSdhmZT0FhxOy7qg3Vi:b-vd9u09) ([raw file](https://github.com/estuary/flow/blob/195b4f0a3e1cf6e800989652242c02d76131de7b/crates/models/src/tables/macros.rs#L44)):*

I feel like there should be a way to express this succinctly with std library traits, but the best I could come up with is `param: impl Into<Cow<'_, T>>` and then calling `param.into_owned()`. But I like how `OwnOrClone` reads nicer, so up to you whether you want to keep it or use Cow.
</blockquote></details>

keeping this for now, but happy to revisit later.

PR #91, opened 2021-02-04
protocols: reworked & more comprehensive specifications

jgraettinger:
Add `LambdaSpec` for specified user lambdas.

Update `Shuffle` spec to reflect additional fields needed to model transform,
or by the shuffle server, or by the shuffle client. This lets us drop some
other structs which wrapped Shuffle, and is forward-looking to computed
shuffle keys.

Add `DerivationSpec` spec which fully models a derivation.

Add `JournalRules` and `ShardRules` for patching journal & shard
specifications.

Update bindings APIs for new catalog-free call patterns.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/91)


psFried (changes requested):
> Reviewed 11 of 11 files at r1.
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/91)* status: all files reviewed, 1 unresolved discussion (waiting on @jgraettinger)

___
*[go/protocols/flow/flow.proto, line 249 at r1](https://reviewable.io/reviews/estuary/flow/91#-MTBvQGu9wlNlWkV-f3t:-MTBvQGu9wlNlWkV-f3u:b-2d3ll6) ([raw file](https://github.com/estuary/flow/blob/a530a330751acff30b583360fc7b6bd96bd3b4b5/go/protocols/flow/flow.proto#L249)):*
> ```protobuf
>   // If an applied update causes the register to be invalid against its
>   // schema, should the document roll back instead of failing processing?
>   bool rollback_on_register_conflict = 5;
> ```

Is there a specific use case you're thinking of for the rollback? What are the implications of rolling back? Would processing the derivation continue along just with the old register value? My speculation is that this is perhaps intended to provide the basis for an update lambda to intentionally refuse to update a register by returning some sentinel that's not valid against the schema.

jgraettinger (commented):
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/91)* status: all files reviewed, 1 unresolved discussion (waiting on @psFried)

___
*[go/protocols/flow/flow.proto, line 249 at r1](https://reviewable.io/reviews/estuary/flow/91#-MTBvQGu9wlNlWkV-f3t:-MTC1YEm9ny4WaB6AzV-:bqjrrvd) ([raw file](https://github.com/estuary/flow/blob/a530a330751acff30b583360fc7b6bd96bd3b4b5/go/protocols/flow/flow.proto#L249)):*
> Is there a specific use case you're thinking of for the rollback?

These are the equivalent of transactional checks over registers. A good motivating use case is a derivation which is joining updates of available product inventory, with requests attempting to claim certain amount of product, and which is producing transactional results of either claimed products, or errors that the requested amount isn't available.

Registers are used to model available amounts of each product. They increase with product inventory updates, and are decreased as product requests claim inventory.

But, available product can never drop below zero! I want to constrain the register schema to prevent that from happening.

Now, what happens if a product request produces a register update which *does* drop it below zero? The default is that the pipeline halts with an error re: the schema violation.

Or, if this option is set, the alternative behavior is that a register update which doesn't validate to the schema (e.g. produces a below-zero value) is not applied. I can then check in my `publish` lambda, comparing `register` and `previous` to determine whether my request was allocated product or not.

> What are the implications of rolling back?

The register is left unchanged and the value returned by `update` is ignored.

jgraettinger (commented):
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/91)* status: all files reviewed, 1 unresolved discussion (waiting on @psFried)

___
*[go/protocols/flow/flow.proto, line 249 at r1](https://reviewable.io/reviews/estuary/flow/91#-MTBvQGu9wlNlWkV-f3t:-MTC47d3CW5c5CZeoSJz:beh2fsr) ([raw file](https://github.com/estuary/flow/blob/a530a330751acff30b583360fc7b6bd96bd3b4b5/go/protocols/flow/flow.proto#L249)):*

> Is there a specific use case you're thinking of for the rollback?

These are the equivalent of transactional checks over registers. A good motivating use case is a derivation which is joining updates of available product inventory, with requests attempting to claim certain amount of product, and which is producing transactional results of either claimed products, or errors that the requested amount isn't available.

Registers are used to model available amounts of each product. They increase with product inventory updates, and are decreased as product requests claim inventory.

But, available product can never drop below zero! I want to constrain the register schema to prevent that from happening.

Now, what happens if a product request produces a register update which *does* drop it below zero? The default is that the pipeline halts with an error re: the schema violation.

Or, if this option is set, the alternative behavior is that a register update which doesn't validate to the schema (e.g. produces a below-zero value) is not applied. I can then check in my `publish` lambda, comparing `register` and `previous` to determine whether my request was allocated product or not.

> What are the implications of rolling back?

The register is left unchanged and the value returned by `update` is ignored.
</blockquote></details>

https://github.com/estuary/flow/blob/catalog-build/crates/sources/src/specs.rs#L276-L292

psFried (approved):
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/91)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jgraettinger)

___
*[go/protocols/flow/flow.proto, line 249 at r1](https://reviewable.io/reviews/estuary/flow/91#-MTBvQGu9wlNlWkV-f3t:-MTCBRAD-aTSMw3xlwjJ:b38ck0j) ([raw file](https://github.com/estuary/flow/blob/a530a330751acff30b583360fc7b6bd96bd3b4b5/go/protocols/flow/flow.proto#L249)):*

https://github.com/estuary/flow/blob/catalog-build/crates/sources/src/specs.rs#L276-L292
</blockquote></details>

Ok, that makes sense and seems pretty useful. My one concern is how we communicate to the `publish` lambda whether or not a register was rolled back or not. I'm a little uncomfortable with relying on equality checks for that, especially given our proximity to Javascript and the insanity of equality there. What if we passed a boolean parameter to the `publish` lambda that explicitly indicates whether the register was updated or not? Most publish lambdas could simply take the normal number of arguments, and those that were interested could accept the additional parameter. Or perhaps we should indicate this by passing `register` as `null` if the update was rolled back? I kinda like the idea of forcing the publish lambda to have a different signature if `rollback_on_register_conflict == true`, since it could be pretty disastrous if the publish lambda failed to detect that the register was rolled back. I don't think this bike shed needs to block this PR, though, so feel free to move this discussion elsewhere if you prefer.

jgraettinger (commented):
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/91)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jgraettinger)

___
*[go/protocols/flow/flow.proto, line 249 at r1](https://reviewable.io/reviews/estuary/flow/91#-MTBvQGu9wlNlWkV-f3t:-MTCEtxfEvsVBlCMSJ87:b-neehsr) ([raw file](https://github.com/estuary/flow/blob/a530a330751acff30b583360fc7b6bd96bd3b4b5/go/protocols/flow/flow.proto#L249)):*

Ok, that makes sense and seems pretty useful. My one concern is how we communicate to the `publish` lambda whether or not a register was rolled back or not. I'm a little uncomfortable with relying on equality checks for that, especially given our proximity to Javascript and the insanity of equality there. What if we passed a boolean parameter to the `publish` lambda that explicitly indicates whether the register was updated or not? Most publish lambdas could simply take the normal number of arguments, and those that were interested could accept the additional parameter. Or perhaps we should indicate this by passing `register` as `null` if the update was rolled back? I kinda like the idea of forcing the publish lambda to have a different signature if `rollback_on_register_conflict == true`, since it could be pretty disastrous if the publish lambda failed to detect that the register was rolled back. I don't think this bike shed needs to block this PR, though, so feel free to move this discussion elsewhere if you prefer.
</blockquote></details>

Let's come back to this, but it's worth talking though.

I'm less worried about JS crazy equality , because this isn't JS and TypeScript is _far_ more sane on that count.

It _may_ make sense to have the `previous` argument be `undefined` if no actual update occurred. However I'm also thinking about things like `reduce: {strategy: sum}` clamps that allow one to decrement up to but not beyond a bound, and there you would definitely want to compare before/after registers (e.x. i'm requesting _up to_ a certain amount of product, but could get less).

current feeling is to wait and see here, and not build anything just yet.

PR #95, opened 2021-02-10
models/sources => `sources` crate

jgraettinger:
Add support for loading TypeScript modules, and for JournalRules.

Catalog source specs are now made a private implementation
detail of the `sources` crate.

New serde specs for protobuf types are added, with serde
ergonomics & JSON-schema generation. Each has direct conversion
to equivalent protobuf messages.

JSON-schema generation utility functions are grouped into
a separate file, rather than scattering them alongside type definition.

Make Fetcher a trait instead of a closure.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/95)


psFried (changes requested):
> Reviewed 39 of 39 files at r1.
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/95)* status: all files reviewed, 2 unresolved discussions (waiting on @jgraettinger)

___
*[crates/sources/src/proto_serde.rs, line 84 at r1](https://reviewable.io/reviews/estuary/flow/95#-MTCkNuyDU1db5hlOrU2:-MTCkNuyDU1db5hlOrU3:bixny75) ([raw file](https://github.com/estuary/flow/blob/9a2c19c000c1424a77c2a01626045ba6f3bbf39a/crates/sources/src/proto_serde.rs#L84)):*
> ```XML
> pub struct JournalSpec {
>     /// # Desired replication of journals.
>     /// The default value is three, meaning that three machines in two deployment
> ```

Is the `serde(default)` of `0` interpreted by gazette to always mean 3? Might be worth a bit of clarification in the doc comment here, as well. As it is, I'm wondering what happens if I leave the default when I deploy to a single-node instance.
___
*[crates/sources/src/proto_serde.rs, line 144 at r1](https://reviewable.io/reviews/estuary/flow/95#-MTCm0h59cxtGh5Svw1S:-MTCm0h59cxtGh5Svw1T:b-5uw56p) ([raw file](https://github.com/estuary/flow/blob/9a2c19c000c1424a77c2a01626045ba6f3bbf39a/crates/sources/src/proto_serde.rs#L144)):*
> ```XML
>     ///   "s3://My-AWS-bucket/a/prefix/my/journal/000123-000456-789abcdef.gzip
>     ///
>     /// Multiple stores may be specified, and all stores are reguarly scanned
> ```

Wow this is an absolutely kick ass feature that I just now learned about!
___
*[crates/sources/src/loader.rs, line 100 at r1](https://reviewable.io/reviews/estuary/flow/95#-MTCfBV68OJgXx4jFyMi:-MTCfBV68OJgXx4jFyMj:b-2nxpko) ([raw file](https://github.com/estuary/flow/blob/9a2c19c000c1424a77c2a01626045ba6f3bbf39a/crates/sources/src/loader.rs#L100)):*
> ```XML
>                     .await
>             }
>             Err(err) if matches!(content_type, names::ContentType::TypescriptModule) => {
> ```

nit: this feels like a weird place to do this check.  But it seems like the intention is for the public api to only be called with a content type of `CatalogSpec`, and the only reason typescript modules are getting fetched is just from recursive calls as part of that. If that's the idea, then I think it'd be clearer with a `pub fn load_catalog_spec(...` and making this one private.

jgraettinger (commented):
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/95)* status: all files reviewed, 2 unresolved discussions (waiting on @psFried)

___
*[crates/sources/src/proto_serde.rs, line 84 at r1](https://reviewable.io/reviews/estuary/flow/95#-MTCkNuyDU1db5hlOrU2:-MTCsY5ACgmsbVZhdafq:b-qvqaq7) ([raw file](https://github.com/estuary/flow/blob/9a2c19c000c1424a77c2a01626045ba6f3bbf39a/crates/sources/src/proto_serde.rs#L84)):*

Is the `serde(default)` of `0` interpreted by gazette to always mean 3? Might be worth a bit of clarification in the doc comment here, as well. As it is, I'm wondering what happens if I leave the default when I deploy to a single-node instance.
</blockquote></details>

There's a [baseline journal spec](https://github.com/estuary/flow/blob/catalog-build/go/flow/journals.go#L48) which journal rules are applied on top of. This is documenting that default.

When you run flowctl develop or test, it applies a top-priority journal rule which overrides replication of all journals to 1.
___
*[crates/sources/src/loader.rs, line 100 at r1](https://reviewable.io/reviews/estuary/flow/95#-MTCfBV68OJgXx4jFyMi:-MTCtKEp-6_XDgDiaB_-:b-oc3o9x) ([raw file](https://github.com/estuary/flow/blob/9a2c19c000c1424a77c2a01626045ba6f3bbf39a/crates/sources/src/loader.rs#L100)):*

nit: this feels like a weird place to do this check.  But it seems like the intention is for the public api to only be called with a content type of `CatalogSpec`, and the only reason typescript modules are getting fetched is just from recursive calls as part of that. If that's the idea, then I think it'd be clearer with a `pub fn load_catalog_spec(...` and making this one private.
</blockquote></details>

Sure. I'll come back to this later ?

psFried (approved):
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/95)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jgraettinger)

___
*[crates/sources/src/proto_serde.rs, line 84 at r1](https://reviewable.io/reviews/estuary/flow/95#-MTCkNuyDU1db5hlOrU2:-MTCuFoW4xn5fHSLlVZh:bx7unmz) ([raw file](https://github.com/estuary/flow/blob/9a2c19c000c1424a77c2a01626045ba6f3bbf39a/crates/sources/src/proto_serde.rs#L84)):*

There's a [baseline journal spec](https://github.com/estuary/flow/blob/catalog-build/go/flow/journals.go#L48) which journal rules are applied on top of. This is documenting that default.

When you run flowctl develop or test, it applies a top-priority journal rule which overrides replication of all journals to 1.
</blockquote></details>

Ok, gotcha :+1: 
___
*[crates/sources/src/loader.rs, line 100 at r1](https://reviewable.io/reviews/estuary/flow/95#-MTCfBV68OJgXx4jFyMi:-MTCu18v03RVWjeyEe1s:b2d09dc) ([raw file](https://github.com/estuary/flow/blob/9a2c19c000c1424a77c2a01626045ba6f3bbf39a/crates/sources/src/loader.rs#L100)):*

Sure. I'll come back to this later ?
</blockquote></details>

yeah, this can just come later.

PR #97, opened 2021-02-11
inference::Exists & build / typescript generation

jgraettinger:
Bundling a one-off commit with a couple of related build / typescript generation commits.

Recommend reviewing this one commit-by-commit

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/97)

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,0 +1,61 @@
+module.exports = {
+  root: true,
+  parser: '@typescript-eslint/parser',
+  ignorePatterns: [
+    // Don't lint already-built outputs.
+    "dist",
+    // Don't lint the NPM package directory.
+    "node_modules",
+    // Uncomment to skip linting of external sources.
+    /* "flow_generated/external" */
+  ],
+  parserOptions: {
+    tsconfigRootDir: __dirname,
+    project: ['./tsconfig.json'],
+  },
+  plugins: ['@typescript-eslint'],
+  extends: [
+    // Core eslint recommendations.
+    'eslint:recommended',
+    // Disable eslint:recommend rules covered by the typescript linter plugin.
+    'plugin:@typescript-eslint/eslint-recommended',
+    // Enable recommend typescript rules.
+    'plugin:@typescript-eslint/recommended',
+    // Enable recommend typescript rules which require type information from `tsc`.
+    'plugin:@typescript-eslint/recommended-requiring-type-checking',
+    // Disable rules from @typescript-eslint/eslint-plugin that conflict with prettier
+    'prettier/@typescript-eslint',
+    // Enable eslint-plugin-prettier and eslint-config-prettier.
+    // This will display prettier errors as ESLint errors.
+    // This must be last configuration in the extends array.
+    'plugin:prettier/recommended'
+  ],
+  // Opt-in to several additional rules.
+  rules:  {
+    // Disable camel-case linting, as identifier names are often drawn from JSON-Schemas
+    // which are outside of the author's control.
+    "@typescript-eslint/camelcase": "off",
+    // Allow variables prefixed with underscore to be unused.
+    "@typescript-eslint/no-unused-vars": ["error", { "argsIgnorePattern": "^_.*" }],
+    // Flow lambdas always return promises, but not all implementations need to be async.
+    "@typescript-eslint/require-await": "off",
+    // Require that created promises are used (await'd), and not silently dropped.
+    "@typescript-eslint/no-floating-promises": "error",
+    // Disallow uses of foo?.bar! (it's not possible to know that bar exists, since foo may not).
+    "@typescript-eslint/no-non-null-asserted-optional-chain": "error",
+    // Require functions returning Promise to be async. This avoids needing to handle a non-async
+    // function which can technically throw an Error *OR* return a rejected Promise. With this
+    // lint, a function can throw (if non-async) or reject (if async) but never both.
+    "@typescript-eslint/promise-function-async": "error",
+    // When adding two operands, each must be of type string or number, and cannot mix.
+    "@typescript-eslint/restrict-plus-operands": ["error", { "checkCompoundAssignments": true }],
+    // Switches over enum types should check all cases (or use explicit "default").
+    "@typescript-eslint/switch-exhaustiveness-check": "error",
+    // Warn on superfluous checks of boolean types against boolean literals (if (foo: bool === true) {}).
+    "@typescript-eslint/no-unnecessary-boolean-literal-compare": "warn",
+    // Warn on conditionals which are always truthy or falsy.
+    "@typescript-eslint/no-unnecessary-condition": "warn",
+    // Warn on "for (let i = 0; i != arr.length; i++)" syntax instead of for-of "for(const i in arr)".
+    "@typescript-eslint/prefer-for-of": "warn",
+  },
+};
--- a/.gitignore
+++ b/.gitignore
@@ -1,16 +1,21 @@
+# Outputs from project build that we don't version.
 target/
 .build/
 .build-ci/
 
-# typical outputs from estctl build
+# Outputs from flowctl build that we don't version.
 /catalog.db
-/catalog-nodejs/
+/node_modules
+/dist
 
+# Ordinarily one *would* version package-lock.json.
+# In our case, though, we want early warning if dependencies of the
+# vanilla package.json produce broken builds.
+package-lock.json
 
-# typical vim stuff
+# Typical IDE stuff.
 .vim
 *.swp
 Session.vim
-
 .idea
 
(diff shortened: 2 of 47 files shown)


psFried (approved):
> Reviewed 1 of 1 files at r1, 13 of 13 files at r2, 33 of 33 files at r3.
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/97)* status: all files reviewed, 1 unresolved discussion (waiting on @jgraettinger)

___
*[crates/build/src/lib.rs, line 105 at r2](https://reviewable.io/reviews/estuary/flow/97#-MTGX_Bc7HYs0lrEIeoj:-MTGX_Bc7HYs0lrEIeok:b-6n0urn) ([raw file](https://github.com/estuary/flow/blob/bf59d2092201972b7448848362a3dd26fa56a8f2/crates/build/src/lib.rs#L105)):*
> ```XML
> 
> pub fn generate_typescript_package(tables: &tables::All, dir: &Path) -> Result<(), anyhow::Error> {
>     assert!(dir.is_absolute() && dir.is_dir());
> ```

nit: Why does this need to be absolute? Maybe worth just adding a doc comment on this function, anyway. Actually, all three of these functions could maybe use some doc comments to clarify the expectations of `dir` before the function is called and the state it'll be left in afterward.
___
*[crates/build/src/web_fetcher.rs, line 32 at r2](https://reviewable.io/reviews/estuary/flow/97#-MTGj1Fi0D7tx6ahJTnD:-MTGj1Fi0D7tx6ahJTnE:bfx97yi) ([raw file](https://github.com/estuary/flow/blob/bf59d2092201972b7448848362a3dd26fa56a8f2/crates/build/src/web_fetcher.rs#L32)):*
> ```XML
>             }
>             "http" | "https" => {
>                 let request = hyper::Request::get(resource.as_str())
> ```

I think this is fine for now, but I'll leave some thoughts here for possible future improvements.
It would be good to follow (at least some) redirects. And I'm not sure if `hyper_tls` uses the system's CA certificates, but I don't think it does, and it would be an important feature. I'm thinking it might make sense to switch to the `reqwest` library, since it makes that kind of thing a little easier and also builds on more platforms.
___
*[crates/build/src/nodejs/generators.rs, line 178 at r2](https://reviewable.io/reviews/estuary/flow/97#-MTGxaQKBwReNiRnQby0:-MTGxaQKBwReNiRnQby1:bovw4ia) ([raw file](https://github.com/estuary/flow/blob/bf59d2092201972b7448848362a3dd26fa56a8f2/crates/build/src/nodejs/generators.rs#L178)):*
> ```XML
> // Generated from derivation {scope}.
> // Required to be implemented by {module}.
> export interface {name} {{
> ```

These comments will be really helpful. It seems like a good idea to render them using the doc comment syntax `/** ... */`. I'd expect VSCode would show doc comments in the little info popups for you, and it would be nice to enable the use of tools like tsdoc as well. This could just come in a later, PR, though, since it would just be a small quality of life improvement.
___
*[crates/build/src/nodejs/typescript/ast.rs, line 4 at r2](https://reviewable.io/reviews/estuary/flow/97#-MTH0u-s1YmhE87LdfA7:-MTH0u-s1YmhE87LdfA8:bwdf60l) ([raw file](https://github.com/estuary/flow/blob/bf59d2092201972b7448848362a3dd26fa56a8f2/crates/build/src/nodejs/typescript/ast.rs#L4)):*
> ```XML
> 
> #[derive(Debug, Clone, PartialEq)]
> pub enum AST {
> ```

I just got wondering if perhaps the [swc project](https://github.com/swc-project/swc) might expose some useful stuff that we could use. Seems like it might, but it's totally understandable if you reckon it's not worth investing the time digging into it at this point.
___
*[crates/doc/src/inference.rs, line 983 at r1](https://reviewable.io/reviews/estuary/flow/97#-MTGV8W8-Hv9BsKWId-z:-MTGV8W93dSqQlQ7xFt8:bbq0nqj) ([raw file](https://github.com/estuary/flow/blob/97adc02bf5061da96b01e7ee4d213fc621d8fedd/crates/doc/src/inference.rs#L983)):*
> ```XML
> /// Exists captures an existence constraint of an Shape location.
> #[derive(Debug, Copy, Clone, Eq, PartialEq)]
> pub enum Exists {
> ```

This is great. I like that it also provides a little safety so we don't accidentally end up using a pointer with a `-` in situations where it doesn't make sense.

jgraettinger (commented):
Thanks!

> *[Reviewable](https://reviewable.io/reviews/estuary/flow/97#-:-MTH7q6nEYRekKbvZjBT:b-ucfknh)* status: all files reviewed, 1 unresolved discussion (waiting on @psFried)

___
*[crates/build/src/lib.rs, line 105 at r2](https://reviewable.io/reviews/estuary/flow/97#-MTGX_Bc7HYs0lrEIeoj:-MTH6_Yd27FL28d-xHEB:b-q7f4uc) ([raw file](https://github.com/estuary/flow/blob/bf59d2092201972b7448848362a3dd26fa56a8f2/crates/build/src/lib.rs#L105)):*

nit: Why does this need to be absolute? Maybe worth just adding a doc comment on this function, anyway. Actually, all three of these functions could maybe use some doc comments to clarify the expectations of `dir` before the function is called and the state it'll be left in afterward.
</blockquote></details>

Yep, I'm going to add this as a future fix-up commit if that's okay.
___
*[crates/build/src/web_fetcher.rs, line 32 at r2](https://reviewable.io/reviews/estuary/flow/97#-MTGj1Fi0D7tx6ahJTnD:-MTH6fIxDBFAdnmCwLCV:b6axb2n) ([raw file](https://github.com/estuary/flow/blob/bf59d2092201972b7448848362a3dd26fa56a8f2/crates/build/src/web_fetcher.rs#L32)):*

I think this is fine for now, but I'll leave some thoughts here for possible future improvements.
It would be good to follow (at least some) redirects. And I'm not sure if `hyper_tls` uses the system's CA certificates, but I don't think it does, and it would be an important feature. I'm thinking it might make sense to switch to the `reqwest` library, since it makes that kind of thing a little easier and also builds on more platforms.
</blockquote></details>

Hah we had been using reqwest in `catalog`, and I actually ripped it out because the only value over hyper was redirect following. `hyper_tls` does use the system TLS library ; i was pretty sure it's using the system CA certs as well.

I've actually been thinking that catalog building should be a bridged library , that flowctl should be a Go process, and that the actual responsibility for fetching resources should live within Go. Let's talk about that outside of here.
___
*[crates/build/src/nodejs/typescript/ast.rs, line 4 at r2](https://reviewable.io/reviews/estuary/flow/97#-MTH0u-s1YmhE87LdfA7:-MTH7Smx8pgBY1cUal5b:bp69zfw) ([raw file](https://github.com/estuary/flow/blob/bf59d2092201972b7448848362a3dd26fa56a8f2/crates/build/src/nodejs/typescript/ast.rs#L4)):*

I just got wondering if perhaps the [swc project](https://github.com/swc-project/swc) might expose some useful stuff that we could use. Seems like it might, but it's totally understandable if you reckon it's not worth investing the time digging into it at this point.
</blockquote></details>

? i'm not familiar with it / don't know what it would bring

jgraettinger (commented):
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/97)* status: all files reviewed, 1 unresolved discussion (waiting on @psFried)

___
*[crates/build/src/lib.rs, line 105 at r2](https://reviewable.io/reviews/estuary/flow/97#-MTGX_Bc7HYs0lrEIeoj:-MTH7wFn9ZzLtP83dIVG:b-jlu0xg) ([raw file](https://github.com/estuary/flow/blob/bf59d2092201972b7448848362a3dd26fa56a8f2/crates/build/src/lib.rs#L105)):*

Yep, I'm going to add this as a future fix-up commit if that's okay.
</blockquote></details>

Oh the reason it needs to be absolute is we test resource URLs against it to determine if a given typescript module resource is external or relative.

jgraettinger (commented):
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/97)* status: all files reviewed, 1 unresolved discussion (waiting on @psFried)

___
*[crates/build/src/nodejs/typescript/ast.rs, line 4 at r2](https://reviewable.io/reviews/estuary/flow/97#-MTH0u-s1YmhE87LdfA7:-MTH8efq6Q-T9JW0j9dv:bkiybke) ([raw file](https://github.com/estuary/flow/blob/bf59d2092201972b7448848362a3dd26fa56a8f2/crates/build/src/nodejs/typescript/ast.rs#L4)):*

? i'm not familiar with it / don't know what it would bring
</blockquote></details>

oh i see it's a deno related projection doing "typescript compilation". This seems like pretty dangerous ground -- the whole point of typescript is strong static check guarantees; I don't see how this could possibly be replicating the level of checks done by `tsc`, and am guessing it's primary goal is just compatibility with the syntax.

PR #99, opened 2021-02-11
remaining go-side changes from `catalog-build` branch

jgraettinger:
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/99)


psFried (approved):
> Reviewed 2 of 2 files at r1, 4 of 4 files at r2, 20 of 20 files at r3, 8 of 8 files at r4, 8 of 8 files at r5, 4 of 4 files at r6.
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/99)* status: all files reviewed, 2 unresolved discussions (waiting on @jgraettinger)

___
*[go/flow/catalog.go, line 75 at r3](https://reviewable.io/reviews/estuary/flow/99#-MTHeLF858r-53iYdG7i:-MTHeLF858r-53iYdG7j:b-qpua9x) ([raw file](https://github.com/estuary/flow/blob/8b512f541a081dc9a7707dc345895d38f1c943e5/go/flow/catalog.go#L75)):*
> ```Go
> // one is not found
> func (catalog *Catalog) LoadCollection(name string) (*pf.CollectionSpec, error) {
> 	var row = catalog.db.QueryRow(`
> ```

We don't seem to be checking the `errors` table anywhere on the go side here, unless I'm missing it. I think we probably should, since "partially built" catalog databases are possible, and things could get _really_ weird if someone managed to somehow apply one. I'm thinking maybe add a check in `NewCatalog`? This could be done if a follow on PR, though.
___
*[go/flow/mapping.go, line 82 at r3](https://reviewable.io/reviews/estuary/flow/99#-MTHm917Dbqlz-028gwN:-MTHm917Dbqlz-028gwO:bfy3c8u) ([raw file](https://github.com/estuary/flow/blob/8b512f541a081dc9a7707dc345895d38f1c943e5/go/flow/mapping.go#L82)):*
> ```Go
> 				{
> 					Upsert:            &applySpec,
> 					ExpectModRevision: 0,
> ```

What's the significance of `ExpectModRevision: 0`? I would have expected that to just be the zero value.
___
*[go/protocols/flow/shuffle_extensions.go, line 61 at r2](https://reviewable.io/reviews/estuary/flow/99#-MTHdUeN2mMoLGmPHYzQ:-MTHdUeN2mMoLGmPHYzR:bylte1o) ([raw file](https://github.com/estuary/flow/blob/64c4f6050e97d3b85f1c3923d620a70dcd6ffe5a/go/protocols/flow/shuffle_extensions.go#L61)):*
> ```Go
> 		return pb.NewValidationError("missing SourceSchemaUri")
> 	}
> 	if (len(m.ShuffleKeyPtr) == 0) == (m.ShuffleLambda == nil) {
> ```

nit: we could also validate that `UsesSourceKey == false` if `ShuffleLambda != nil`. Doesn't seem super important that we validate that here, though, so no probs if you reckon it's not worth adding.

jgraettinger (commented):
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/99)* status: all files reviewed, 2 unresolved discussions (waiting on @psFried)

___
*[go/flow/catalog.go, line 75 at r3](https://reviewable.io/reviews/estuary/flow/99#-MTHeLF858r-53iYdG7i:-MTHpkWH4KUyLlZ0_Cp7:b-nxcjd3) ([raw file](https://github.com/estuary/flow/blob/8b512f541a081dc9a7707dc345895d38f1c943e5/go/flow/catalog.go#L75)):*

We don't seem to be checking the `errors` table anywhere on the go side here, unless I'm missing it. I think we probably should, since "partially built" catalog databases are possible, and things could get _really_ weird if someone managed to somehow apply one. I'm thinking maybe add a check in `NewCatalog`? This could be done if a follow on PR, though.
</blockquote></details>

Yep, you're right, that should be added.
___
*[go/flow/mapping.go, line 82 at r3](https://reviewable.io/reviews/estuary/flow/99#-MTHm917Dbqlz-028gwN:-MTHpq5758rl5g8uWOwb:brf8sw) ([raw file](https://github.com/estuary/flow/blob/8b512f541a081dc9a7707dc345895d38f1c943e5/go/flow/mapping.go#L82)):*

What's the significance of `ExpectModRevision: 0`? I would have expected that to just be the zero value.
</blockquote></details>

Yep it is already the zero value; I've made it explicit simply to call it out. The behavior is that we're expecting this specification does not exist, and the transaction will fail if it does.
___
*[go/protocols/flow/shuffle_extensions.go, line 61 at r2](https://reviewable.io/reviews/estuary/flow/99#-MTHdUeN2mMoLGmPHYzQ:-MTHqEaS7Pq8k4nqNV2f:b-83od3n) ([raw file](https://github.com/estuary/flow/blob/64c4f6050e97d3b85f1c3923d620a70dcd6ffe5a/go/protocols/flow/shuffle_extensions.go#L61)):*

nit: we could also validate that `UsesSourceKey == false` if `ShuffleLambda != nil`. Doesn't seem super important that we validate that here, though, so no probs if you reckon it's not worth adding.
</blockquote></details>

Noted

Issue #100, opened 2021-02-11
gazette: recoverylog.FSM support for metadata forks


jgraettinger:
The [design for online shard splits](https://github.com/estuary/flow/issues/15) relies on being able to express lightweight forks of recovery logs.

We're _really_ close to having the pieces needed for this. The primary new change is expanding [recoverylog.Segment](https://github.com/gazette/core/blob/master/consumer/recoverylog/recorded_op.proto#L122) -- and related FSM / Recorder / Player machinery -- to model the log it references (rather than always assuming its a single log.

Then, FSM can track log changes over time, model them in FSMHints, and Player can identify logs to read from hinted Segments.

PR #101, opened 2021-02-13
go/flow: dynamically create missing recovery logs

jgraettinger:
Shards create their own recovery logs using JournalRules, via a new
optional BeginRecoverer interface implementation added to Gazette's
consumer.Application.

Rip out log fixture creation which had been undertaken by `flowctl
test`, as it's no longer needed.

Also update the Gazette pin to bring in support for recoverylog forks.

Issue #100

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/101)


psFried (approved):
> Reviewed 10 of 10 files at r1.
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/101)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jgraettinger)

___
*[crates/protocol/src/recoverylog.rs, line 42 at r1](https://reviewable.io/reviews/estuary/flow/101#-MTax3jy5ILZHkVxdOkn:-MTax3jy5ILZHkVxdOko:bhf4nqs) ([raw file](https://github.com/estuary/flow/blob/0db131287538734dd1e3da6f22b0fa306a9da8be/crates/protocol/src/recoverylog.rs#L42)):*
> ```XML
>     pub write: ::std::option::Option<recorded_op::Write>,
>     /// Property indicates a property file has been created or updated.
>     /// DEPRECATED. Properties are no longer created,
> ```

Only because I'm curious, is there a problem with handling Properties in this way, or is there just no longer a need for them?

jgraettinger (commented):
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/101)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jgraettinger)

___
*[crates/protocol/src/recoverylog.rs, line 42 at r1](https://reviewable.io/reviews/estuary/flow/101#-MTax3jy5ILZHkVxdOkn:-MTfdAa9EmUPGf8XaMcZ:bfvymfh) ([raw file](https://github.com/estuary/flow/blob/0db131287538734dd1e3da6f22b0fa306a9da8be/crates/protocol/src/recoverylog.rs#L42)):*

Only because I'm curious, is there a problem with handling Properties in this way, or is there just no longer a need for them?
</blockquote></details>

There just isn't a need. They date from a time when I had a simpler pruning model in mind for recovery logs -- that we'd essentially just pick a lower-bound offset of the log and delete everything before that. This works as long as every file in the RocksDB has a bounded lifetime, when in theory it does.

The only problem with that -- or so i thought -- was this `IDENTITY` file that rocks creates at db startup and then never writes again. It's just a UUID, so the fix was `Properties`.

In practice, though, it's _usually_ the case that there are big chunks of the LSM tree which are effectively written once and not touched again. Log pruning is now fine grained, and will remove discontinuous chunks from within the log where able. So, `IDENTITY` is no longer a problem 'cause it just means the first fragment of the log hangs around -- no big deal.

But, I don't want to yank support for Properties either given the number of existing recovery logs that have them, the need for a migration strategy if properties is removed altogether, and the fact that _it's not broken_. I'm frankly still on the fence around deprecation, and this may be a long term vestigial thing we just live with.

PR #102, opened 2021-02-17
improvements to document validations and reductions

jgraettinger:
Add SpanContext, a light-weight validation context which does
much less work & allocation as compared to FullContext.

Add new types doc::Validation & doc::Valid for efficient and type-safe
document validations, outcome extraction, and error generation.

Implement (optional) register roll-backs on conflicts.

Implement distinct Combiner reduce-left vs combine-right operations.
Update Combiner bindings to match.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/102)


jgraettinger:
I've added a second commit which adds a shared protobuf `flow.MaterializationSpec` for runtime use, bundling everything that's known about a given materialization.

psFried (changes requested):
> Reviewed 21 of 21 files at r1, 25 of 25 files at r2, 12 of 12 files at r3.
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/102)* status: all files reviewed, 3 unresolved discussions (waiting on @jgraettinger)

___
*[crates/derive/src/registers.rs, line 130 at r1](https://reviewable.io/reviews/estuary/flow/102#-MTlspQLBEVxR2lSC2Rd:-MTloIQJ4Y3OQyjAQQgV:b-olgl0g) ([raw file](https://github.com/estuary/flow/blob/379f2f15fae2901e3a582551a0733f567e06b97a/crates/derive/src/registers.rs#L130)):*
> ```XML
> 
>         let rollback = if rollback_on_conflict {
>             Some(lhs.clone())
> ```

I think this is probably fine, but I want to make sure: Should we be rolling back to the last valid document, or the first valid document (as it is)? I _think_ `deltas` here represents the results from a single update lambda invocation, so we'd want to rollback to the starting `lhs` on validation failure. Does that sound right?
___
*[crates/derive/src/registers.rs, line 155 at r1](https://reviewable.io/reviews/estuary/flow/102#-MTlxhQc9nAEFynEjLI0:-MTlxhQc9nAEFynEjLI1:b545bg0) ([raw file](https://github.com/estuary/flow/blob/379f2f15fae2901e3a582551a0733f567e06b97a/crates/derive/src/registers.rs#L155)):*
> ```XML
>             match rollback {
>                 Some(rollback) => {
>                     *lhs = rollback;
> ```

optional nit: I foresee a desire to log the validation errors that trigger a rollback, so we can easily respond to "omg my registers aren't updating because bugs in flow"
___
*[crates/validation/src/materialization.rs, line 82 at r2](https://reviewable.io/reviews/estuary/flow/102#-MTm0vjlDBBZBnHIrRR0:-MTm0vjlDBBZBnHIrRR1:blu82za) ([raw file](https://github.com/estuary/flow/blob/fc270a6319e8c3093dd98483d23b03548eb7cb55/crates/validation/src/materialization.rs#L82)):*
> ```XML
>                     ep_config,
>                     flow::MaterializationSpec {
>                         collection: None,
> ```

Missing `collection` here. If this is intentional, then I think it's worth a comment.

jgraettinger (commented):
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/102)* status: all files reviewed, 3 unresolved discussions (waiting on @psFried)

___
*[crates/derive/src/registers.rs, line 130 at r1](https://reviewable.io/reviews/estuary/flow/102#-MTlspQLBEVxR2lSC2Rd:-MTnDCCLAY6XxSc4Kjop:b-ht0ukm) ([raw file](https://github.com/estuary/flow/blob/379f2f15fae2901e3a582551a0733f567e06b97a/crates/derive/src/registers.rs#L130)):*

I think this is probably fine, but I want to make sure: Should we be rolling back to the last valid document, or the first valid document (as it is)? I _think_ `deltas` here represents the results from a single update lambda invocation, so we'd want to rollback to the starting `lhs` on validation failure. Does that sound right?
</blockquote></details>

That's right.
___
*[crates/derive/src/registers.rs, line 155 at r1](https://reviewable.io/reviews/estuary/flow/102#-MTlxhQc9nAEFynEjLI0:-MTnDQ2qB9hdsiMx1GB-:b-mqjieo) ([raw file](https://github.com/estuary/flow/blob/379f2f15fae2901e3a582551a0733f567e06b97a/crates/derive/src/registers.rs#L155)):*

optional nit: I foresee a desire to log the validation errors that trigger a rollback, so we can easily respond to "omg my registers aren't updating because bugs in flow"
</blockquote></details>

Definitely agreed something is needed here. I don't think it's straight up validation errors  (these are expensive to produce), but perhaps metrics? Or sampled traces? That's going to be an effort in it's own right.
___
*[crates/validation/src/materialization.rs, line 82 at r2](https://reviewable.io/reviews/estuary/flow/102#-MTm0vjlDBBZBnHIrRR0:-MTnET3JA8vPSU0h-hz2:b-35bmc8) ([raw file](https://github.com/estuary/flow/blob/fc270a6319e8c3093dd98483d23b03548eb7cb55/crates/validation/src/materialization.rs#L82)):*

Missing `collection` here. If this is intentional, then I think it's worth a comment.
</blockquote></details>

It's commented on the field in push_materializations. Added a reference here.

jgraettinger (commented):
Thanks!

> *[Reviewable](https://reviewable.io/reviews/estuary/flow/102#-:-MTnEkA_0gNKAkcLFr-S:b-ucfknh)* status: all files reviewed, 3 unresolved discussions (waiting on @psFried)

PR #103, opened 2021-02-18
flowctl-go

jgraettinger:
A first take at a `flowctl` Go binary, using a CGO bridge build API that trampolines fetches as we discussed this morning.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/103)


jgraettinger:
I've rounded out getting `flowctl-go` fully equivalent to the current `flowctl` -- added onto this PR.

psFried (approved):
> Reviewed 57 of 60 files at r1, 14 of 14 files at r2.
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/103)* status: all files reviewed, 1 unresolved discussion (waiting on @jgraettinger)

___
*[crates/build/src/api.rs, line 107 at r2](https://reviewable.io/reviews/estuary/flow/103#-MUF2JXT2C7zhsMy4f_U:-MUF2JXUDBI8C12la2Ac:b-abj86l) ([raw file](https://github.com/estuary/flow/blob/c4d8d58575737d38f68c71d95a60b04709479725/crates/build/src/api.rs#L107)):*
> ```XML
>             .iter()
>             .enumerate()
>             .find_map(|(index, fetch)| {
> ```

FYI there's a [position](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.position) function that does this a little more succinctly
___
*[crates/build/src/api.rs, line 193 at r2](https://reviewable.io/reviews/estuary/flow/103#-MUF53WaDTU7kjYCcVYN:-MUF53WaDTU7kjYCcVYO:b2urdeo) ([raw file](https://github.com/estuary/flow/blob/c4d8d58575737d38f68c71d95a60b04709479725/crates/build/src/api.rs#L193)):*
> ```XML
> 
>                         // We must have drained all outstanding fetches.
>                         assert!(future.fetch_dispatch.borrow().is_empty());
> ```

Is there any chance that there could still be pending fetches if we're failing the build (i.e. fail-fast)?
___
*[crates/build/src/lib.rs, line 18 at r2](https://reviewable.io/reviews/estuary/flow/103#-MUF9Yq8DgeSOegETuTN:-MUF9Yq8DgeSOegETuTO:b-baqv22) ([raw file](https://github.com/estuary/flow/blob/c4d8d58575737d38f68c71d95a60b04709479725/crates/build/src/lib.rs#L18)):*
> ```XML
> mod nodejs;
> 
> pub async fn configured_build<F, D>(
> ```

There's a lot of mixing of blocking and async function calls here. I think that's OK within the context of manually polling the future from the cgo bridge layer, but it could be a problem if this function were called from within the execution of a normal rust async task, since it could block the entire executor thread. I wonder if it's possible to move the `load_and_validate` out, so this doesn't need to be an `async fn`. Or maybe just add a comment or rename it to clarify that this is only valid to be called from the context of the cgo bridge. Generally, I think it best to limit mixing of async and blocking to as few well defined areas as possible. That may already be the case, but I'm only partway through, and these can be really tough to spot.
___
*[crates/models/src/build/mod.rs, line 230 at r2](https://reviewable.io/reviews/estuary/flow/103#-MUFDgxi6a8aEhX-cBAA:-MUFDgxi6a8aEhX-cBAB:bt7eixb) ([raw file](https://github.com/estuary/flow/blob/c4d8d58575737d38f68c71d95a60b04709479725/crates/models/src/build/mod.rs#L230)):*
> ```XML
>             .iter()
>             .map(|d| d.to_string())
>             .collect::<Vec<_>>()
> ```

nit: Assuming this is the `Itertools::join`, there's no need to `collect`, and probably no need for the `map` either.
___
*[go/flowctl-go/cmd-test.go, line 191 at r2](https://reviewable.io/reviews/estuary/flow/103#-MUFGz7z7ZlN6DKan2F1:-MUFGz7z7ZlN6DKan2F2:b-dsqc8w) ([raw file](https://github.com/estuary/flow/blob/c4d8d58575737d38f68c71d95a60b04709479725/go/flowctl-go/cmd-test.go#L191)):*
> ```Go
> }
> 
> func startTempEtcd(tmpdir string) (*exec.Cmd, *clientv3.Client, error) {
> ```

Something to think about for the future, maybe we should just link etcd like a normal go dependency?

jgraettinger (commented):
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/103)* status: all files reviewed, 1 unresolved discussion (waiting on @psFried)

___
*[crates/build/src/api.rs, line 107 at r2](https://reviewable.io/reviews/estuary/flow/103#-MUF2JXT2C7zhsMy4f_U:-MUFW-UNECK6OTspznJd:b-evivl5) ([raw file](https://github.com/estuary/flow/blob/c4d8d58575737d38f68c71d95a60b04709479725/crates/build/src/api.rs#L107)):*

FYI there's a [position](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.position) function that does this a little more succinctly
</blockquote></details>

noted, this function is gone in a future commit already though.
___
*[crates/build/src/api.rs, line 193 at r2](https://reviewable.io/reviews/estuary/flow/103#-MUF53WaDTU7kjYCcVYN:-MUFWD1J02YU-IpohMI9:b-u5ypd3) ([raw file](https://github.com/estuary/flow/blob/c4d8d58575737d38f68c71d95a60b04709479725/crates/build/src/api.rs#L193)):*

Is there any chance that there could still be pending fetches if we're failing the build (i.e. fail-fast)?
</blockquote></details>

Hmm, I wanted it here as an assertion / sanity check for my mental model, and it's changed already in a future PR, though I'll consider further how to handle it -- probably just relaxing the assertion.

It is actually safe in practice because the whole source loading model treats errors as load artifacts like any other. There is no fail-fast. Same with validations.
___
*[crates/build/src/lib.rs, line 18 at r2](https://reviewable.io/reviews/estuary/flow/103#-MUF9Yq8DgeSOegETuTN:-MUFXsSxAneJAmxolZOA:b-1r3id9) ([raw file](https://github.com/estuary/flow/blob/c4d8d58575737d38f68c71d95a60b04709479725/crates/build/src/lib.rs#L18)):*
You're referring to `std::fs` & `rusqlite::Open` ?

We need to mix together a bunch of inter-related async & sync API's, which pretty much mandates that we need a tangled ball of mud _somewhere_ which does this. This is factored to be that ball of mud so others don't have to.

>  but it could be a problem if this function were called from within the execution of a normal rust async task

I suppose it can make it private, and will, but this is concern over a problem we don't have, nor do we intend to have?
___
*[crates/models/src/build/mod.rs, line 230 at r2](https://reviewable.io/reviews/estuary/flow/103#-MUFDgxi6a8aEhX-cBAA:-MUFvePb5FjdkRkb1Ivw:bgkvzlp) ([raw file](https://github.com/estuary/flow/blob/c4d8d58575737d38f68c71d95a60b04709479725/crates/models/src/build/mod.rs#L230)):*

nit: Assuming this is the `Itertools::join`, there's no need to `collect`, and probably no need for the `map` either.
</blockquote></details>

this isn't using itertools, it's an impl Vec<String> method. to_string is going from serde_json::Value -> String.
___
*[go/flowctl-go/cmd-test.go, line 191 at r2](https://reviewable.io/reviews/estuary/flow/103#-MUFGz7z7ZlN6DKan2F1:-MUFXQ5K2cQGCvOzFJWU:bxy4x2w) ([raw file](https://github.com/estuary/flow/blob/c4d8d58575737d38f68c71d95a60b04709479725/go/flowctl-go/cmd-test.go#L191)):*

Something to think about for the future, maybe we should just link etcd like a normal go dependency?
</blockquote></details>

https://github.com/gazette/core/commit/af54f8eb65beb65af8c4bc506f0db21131ac4970

Haha it was dependency hell.

What could be cool though, is to explore using an archive tool to pack `etcd` into the executable. Same benefit & end-state.

jgraettinger (commented):
confused myself a bit, i see now on github it's marked as approve, is your disposition to merge ?

thanks

> *[Reviewable](https://reviewable.io/reviews/estuary/flow/103#-:-MUG0sJaA0IX5CKvzT9h:b-lr325l)* status: all files reviewed, 1 unresolved discussion (waiting on @psFried)

PR #109, opened 2021-02-25
materialization protocol refactors

jgraettinger:
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/109)


jgraettinger:
A few more fixes got to a passing build, and it doesn't seem a terrible PR -- viewing commits together actually is more readable than individually as it better resolves churn -- so sending your way.

psFried (changes requested):
> Reviewed 43 of 43 files at r1.
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/109)* status: all files reviewed, 2 unresolved discussions (waiting on @jgraettinger)

___
*[go/materialize/driver/sql2/driver.go, line 47 at r1](https://reviewable.io/reviews/estuary/flow/109#-MUOetNi0Q5F9ho8CV68:-MUOetNi0Q5F9ho8CV69:btbc047) ([raw file](https://github.com/estuary/flow/blob/6acf9d2f8364b0b43b3e2046e611234f940a3ecf/go/materialize/driver/sql2/driver.go#L47)):*
> ```Go
> 	}
> 
> 	current, err := endpoint.LoadSpec(false)
> ```

The old implementation of `LoadSpec` called `Ping` in an attempt to differentiate between  common network connection issues and errors that are simply caused by a missing table or row. That was based on the [Open docs](https://golang.org/pkg/database/sql/#Open) saying that calling `Ping` may be required in order to ensure you can connect to the database, since `Open` may not actually connect immediately.

The specific scenario I'm concerned with here is if there's some network issue that temporarily prevents us from connecting to the database, which then causes us to return validation results that don't account for an existing materialization. More generally, though, I think `Validate` has to validate the connection in addition to the spec, as this will be the only way for users to validate that their parameters are correct before actually applying. The old code was admittedly unclear on that, and there may be a better place to validate the connection than `LoadSpec`. I'm thinking we want the same behavior in `Apply` as well.
___
*[go/materialize/driver/sql2/fence.go, line 44 at r1](https://reviewable.io/reviews/estuary/flow/109#-MUOpUVv2T2_frzcuvBS:-MUOpUVv2T2_frzcuvBT:bc09s8w) ([raw file](https://github.com/estuary/flow/blob/6acf9d2f8364b0b43b3e2046e611234f940a3ecf/go/materialize/driver/sql2/fence.go#L44)):*
> ```Go
> 			"UPDATE %s SET fence=fence+1 WHERE shard_fqn=%s;",
> 			e.Tables.Checkpoints,
> 			e.Generator.Placeholder(1),
> ```

Unless you changed it, I think `Placeholder` was defined as being 0-indexed

jgraettinger (commented):
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/109)* status: all files reviewed, 2 unresolved discussions (waiting on @psFried)

___
*[go/materialize/driver/sql2/driver.go, line 47 at r1](https://reviewable.io/reviews/estuary/flow/109#-MUOetNi0Q5F9ho8CV68:-MUOyZmQ14rm-gW5kY1S:b-e7axoc) ([raw file](https://github.com/estuary/flow/blob/6acf9d2f8364b0b43b3e2046e611234f940a3ecf/go/materialize/driver/sql2/driver.go#L47)):*

The old implementation of `LoadSpec` called `Ping` in an attempt to differentiate between  common network connection issues and errors that are simply caused by a missing table or row. That was based on the [Open docs](https://golang.org/pkg/database/sql/#Open) saying that calling `Ping` may be required in order to ensure you can connect to the database, since `Open` may not actually connect immediately.

The specific scenario I'm concerned with here is if there's some network issue that temporarily prevents us from connecting to the database, which then causes us to return validation results that don't account for an existing materialization. More generally, though, I think `Validate` has to validate the connection in addition to the spec, as this will be the only way for users to validate that their parameters are correct before actually applying. The old code was admittedly unclear on that, and there may be a better place to validate the connection than `LoadSpec`. I'm thinking we want the same behavior in `Apply` as well.
</blockquote></details>

Makes sense, and I do think `LoadSpec` is an appropriate place to do this -- essentially we're narrowing the scope of ignorable errors if `mustExist` is false. Will update.
___
*[go/materialize/driver/sql2/fence.go, line 44 at r1](https://reviewable.io/reviews/estuary/flow/109#-MUOpUVv2T2_frzcuvBS:-MUOyu_wDc2YUgZFRwpQ:b-ryxgox) ([raw file](https://github.com/estuary/flow/blob/6acf9d2f8364b0b43b3e2046e611234f940a3ecf/go/materialize/driver/sql2/fence.go#L44)):*

Unless you changed it, I think `Placeholder` was defined as being 0-indexed
</blockquote></details>

Whoops, good catch! Worked because sqlite is '?' I suppose.

jgraettinger:
PTAL

psFried (approved):
> Reviewed 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/109)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jgraettinger)

PR #111, opened 2021-02-26
sqlite ergonomics, concurrency, and materialization soak test

jgraettinger:
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/111)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@ target/
 /catalog.db
 /dist
 /examples/examples.db*
+/flowctl-develop
 /node_modules
 
 # Ordinarily one *would* version package-lock.json.
--- a/.gitignore
+++ b/.gitignore
@@ -3,10 +3,11 @@ target/
 .build/
 .build-ci/
 
-# Outputs from flowctl build that we don't version.
+# Outputs from catalog build & test that we don't version.
 /catalog.db
-/node_modules
 /dist
+/examples/examples.db*
+/node_modules
 
 # Ordinarily one *would* version package-lock.json.
 # In our case, though, we want early warning if dependencies of the
--- a/crates/sources/src/loader.rs
+++ b/crates/sources/src/loader.rs
@@ -388,17 +388,18 @@ impl<F: Fetcher> Loader<F> {
 
         // Collect endpoints.
         for (name, endpoint) in endpoints {
-            let scope = scope
-                .push_prop("endpoints")
-                .push_prop(name.as_ref())
-                .flatten();
+            let scope = scope.push_prop("endpoints");
+            let scope = scope.push_prop(name.as_ref());
+            let endpoint_type = endpoint.endpoint_type();
 
-            self.tables.borrow_mut().endpoints.push_row(
-                scope,
-                name,
-                endpoint.endpoint_type(),
-                endpoint.base_config(),
-            );
+            if let Some(base_config) = self.load_endpoint_config(scope, endpoint) {
+                self.tables.borrow_mut().endpoints.push_row(
+                    scope.flatten(),
+                    name,
+                    endpoint_type,
+                    base_config,
+                );
+            }
         }
 
         // Collect captures.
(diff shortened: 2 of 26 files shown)


jgraettinger:
ping?

psFried (approved):
> Reviewed 6 of 6 files at r1, 1 of 1 files at r2, 10 of 10 files at r3, 4 of 4 files at r4, 11 of 11 files at r5.
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/111)* status: all files reviewed, 1 unresolved discussion (waiting on @jgraettinger)

___
*[examples/flow.yaml, line 11 at r1](https://reviewable.io/reviews/estuary/flow/111#-MUnUVVK9bt-hxwr-KJv:-MUnUVVK9bt-hxwr-KJw:bh7a67d) ([raw file](https://github.com/estuary/flow/blob/9d8c47e416097838a9a9f6042bc902681f6a9502/examples/flow.yaml#L11)):*
> ```YAML
>   examples:
>     sqlite:
>       path: examples.db?_journal=WAL
> ```

Is `_journal=WAL` something that most sqlite endpoints would want to use? Wondering if we should have the driver add it automatically? Doesn't seem like it would hurt, but also not anything that can't be done in a future PR.
___
*[go/materialize/driver/sqlite/sqlite.go, line 41 at r2](https://reviewable.io/reviews/estuary/flow/111#-MUn_90N39jXLyDFt4P1:-MUn_90N39jXLyDFt4P2:byjakma) ([raw file](https://github.com/estuary/flow/blob/a953ffb7a0e8891caae214af6246ae28974a381f/go/materialize/driver/sqlite/sqlite.go#L41)):*
> ```Go
> 				// We can directly pass file:// schemes to SQLite.
> 			} else {
> 				// Path is absolute and non-local (e.x. https://some/database.db).
> ```

I had left another comment, which I'll now delete, about relative paths and remote URIs. This seems like a good solution. One nit pick is that if we left the path separators in there and did `<host>/<path>`, then it might result in a marginally better organization if you import multiple materializations whose paths would share a common prefix. ~~It might also help to keep filenames shorter if someone had an obnoxiously long URL, but that seems pretty unlikely, and maybe not worth worrying about.~~ [This URL](https://raw.githubusercontent.com/estuary/flow/fcebc44d81c0359c39ab032cb70952e2cd40a9cd/examples/int-string-flow.yaml) is only 117 characters, so I'm thinking that really isn't worth worrying about.

jgraettinger (commented):
Thanks

> *[Reviewable](https://reviewable.io/reviews/estuary/flow/111#-:-MUxRCkECNFQ4SKGa9Wm:b-zcjvte)* status: all files reviewed, 1 unresolved discussion (waiting on @psFried)

___
*[examples/flow.yaml, line 11 at r1](https://reviewable.io/reviews/estuary/flow/111#-MUnUVVK9bt-hxwr-KJv:-MUxPYPP8iSX78rg6wOv:b7rpxte) ([raw file](https://github.com/estuary/flow/blob/9d8c47e416097838a9a9f6042bc902681f6a9502/examples/flow.yaml#L11)):*

Is `_journal=WAL` something that most sqlite endpoints would want to use? Wondering if we should have the driver add it automatically? Doesn't seem like it would hurt, but also not anything that can't be done in a future PR.
</blockquote></details>

It's not required (though it may become so with the async-commit behavior discussed yesterday). I considered this, but on balance opted to keep configuration very literal to avoid magic.
___
*[go/materialize/driver/sqlite/sqlite.go, line 41 at r2](https://reviewable.io/reviews/estuary/flow/111#-MUn_90N39jXLyDFt4P1:-MUxQ8yyBtVQ1A4IzmIj:b-rh1wvx) ([raw file](https://github.com/estuary/flow/blob/a953ffb7a0e8891caae214af6246ae28974a381f/go/materialize/driver/sqlite/sqlite.go#L41)):*

I had left another comment, which I'll now delete, about relative paths and remote URIs. This seems like a good solution. One nit pick is that if we left the path separators in there and did `<host>/<path>`, then it might result in a marginally better organization if you import multiple materializations whose paths would share a common prefix. ~~It might also help to keep filenames shorter if someone had an obnoxiously long URL, but that seems pretty unlikely, and maybe not worth worrying about.~~ [This URL](https://raw.githubusercontent.com/estuary/flow/fcebc44d81c0359c39ab032cb70952e2cd40a9cd/examples/int-string-flow.yaml) is only 117 characters, so I'm thinking that really isn't worth worrying about.
</blockquote></details>

The most likely reason for hitting this is because you're just trying out flow, and running `flowctl` against a remote repo. Intended ergonomics are that I can run `flowctl develop` and have a single, obvious DB created that tab-completes from my current working directory.

PR #113, opened 2021-03-02
Add Poll_Request functions to Load/Store Iterators

psFried:
Adds functions to lifecycle.Load/StoreIterator types to allow for
checking whether they contain at least one item to process. This is
intended to provide a way for Driver implementations to avoid doing
work for Loads or Stores when a transaction does not contain any.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/113)

### DIFF
--- a/go/materialize/lifecycle/protocol.go
+++ b/go/materialize/lifecycle/protocol.go
@@ -295,12 +295,15 @@ func NewLoadIterator(stream pm.Driver_TransactionsServer) *LoadIterator {
 	return &LoadIterator{stream: stream}
 }
 
-// Next returns true if there is another Load and makes it available via Key.
-// When a Prepare is read, or if an error is encountered, it returns false
-// and must not be called again.
-func (it *LoadIterator) Next() bool {
+// Poll returns true if there is at least one LoadRequest message with at least one key
+// remaining to be read. This will read the next message from the stream if required. This will not
+// advance the iterator, so it can be used to check whether the LoadIterator contains at least one
+// key to load without actually consuming the next key. If false is returned, then there are no
+// remaining keys and Poll must not be called again. Note that if Poll returns
+// true, then Next may still return false if the LoadRequest message is malformed.
+func (it *LoadIterator) Poll() bool {
 	if it.err != nil || it.req.Prepare != nil {
-		panic("Next called again after having returned false")
+		panic("Poll called again after having returned false")
 	}
 
 	// Must we read another request?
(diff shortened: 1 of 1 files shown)


jgraettinger (approved):
LGTM % comments

> *[Reviewable](https://reviewable.io/reviews/estuary/flow/113#-:-MUo4Khl4xJe27HhJJjo:b-tz1f3e)* status: 0 of 1 files reviewed, 4 unresolved discussions (waiting on @psFried)

___
*[go/materialize/lifecycle/protocol.go, line 304 at r1](https://reviewable.io/reviews/estuary/flow/113#-MUo3pEn9mh87wINUfAj:-MUo3pEn9mh87wINUfAk:bbdxoek) ([raw file](https://github.com/estuary/flow/blob/c17f66a5daad1184b81f222b8b88064e86b0823b/go/materialize/lifecycle/protocol.go#L304)):*
> ```Go
> // remaining keys and PollLoadRequest must not be called again. Note that if PollLoadRequest returns
> // true, then Next may still return false if the LoadRequest message is malformed.
> func (it *LoadIterator) PollLoadRequest() bool {
> ```

how about just `Poll` ?
___
*[go/materialize/lifecycle/protocol.go, line 337 at r1](https://reviewable.io/reviews/estuary/flow/113#-MUo3mzS0hgTA0BuLO8i:-MUo3mzS0hgTA0BuLO8j:bqwrmre) ([raw file](https://github.com/estuary/flow/blob/c17f66a5daad1184b81f222b8b88064e86b0823b/go/materialize/lifecycle/protocol.go#L337)):*
> ```Go
> // and must not be called again.
> func (it *LoadIterator) Next() bool {
> 	if it.err != nil || it.req.Prepare != nil {
> ```

this can be dropped -- let Poll do it.
___
*[go/materialize/lifecycle/protocol.go, line 385 at r1](https://reviewable.io/reviews/estuary/flow/113#-MUo4CYa3xB44dybV4uR:-MUo4CYa3xB44dybV4uS:bxwf69v) ([raw file](https://github.com/estuary/flow/blob/c17f66a5daad1184b81f222b8b88064e86b0823b/go/materialize/lifecycle/protocol.go#L385)):*
> ```Go
> }
> 
> // PollStoreRequest returns true if there is at least one StoreRequest message with at least one
> ```

Poll ?
___
*[go/materialize/lifecycle/protocol.go, line 420 at r1](https://reviewable.io/reviews/estuary/flow/113#-MUo4E1DD5cXhZdqUM1z:-MUo4E1DD5cXhZdqUM2-:b-2guby3) ([raw file](https://github.com/estuary/flow/blob/c17f66a5daad1184b81f222b8b88064e86b0823b/go/materialize/lifecycle/protocol.go#L420)):*
> ```Go
> // and must not be called again.
> func (it *StoreIterator) Next() bool {
> 	if it.err != nil || it.req.Commit != nil {
> ```

same here

jgraettinger (approved):
Still LGTM

> *[Reviewable](https://reviewable.io/reviews/estuary/flow/113#-:-MUok4YJ7yK-gN3aK3rp:b-kwu2jz)* status: 0 of 1 files reviewed, 4 unresolved discussions (waiting on @psFried)

PR #115, opened 2021-03-04
Pg driver sql2

psFried:
Includes the postgres driver and the new functional test ported over from rust.
There's a number of things yet to do, but I think these are safe to wait for a subsequent PR:

[ ] Transactional test ported over from rust to go
[ ] Running the driver tests in CI (I've started looking into it, have some ideas)
[ ] Moving Loads out of the Store transaction (possibly into a separate transaction)
[X] ~~Finish looking into the SendBatch race mentioned here: https://pkg.go.dev/github.com/jackc/pgx@v3.6.2+incompatible#Batch.Send~~ This is supposedly resolved by updating to v4 of the pgx library, which was another thing on the list anyway.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/115)


jgraettinger (changes requested):
By and large this is great. I have a lot of comments below, but as I speak to in the meta-comment, what they're really after is moving us towards a common voice for our Go code.

Otherwise there's one functional comment on when we acquire the transaction lock, and another about inlining the fixture *pf.MaterializationSpec to avoid a bunch of coupling.

Thanks

> Reviewed 13 of 16 files at r1, 3 of 3 files at r3, 10 of 10 files at r4, 8 of 8 files at r5.
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/115#-:-MUxrsIDA6ZHw4N0M5Bk:b-m0z4yh)* status: all files reviewed, 16 unresolved discussions (waiting on @psFried)

___
*[Makefile, line 285 at r5](https://reviewable.io/reviews/estuary/flow/115#-MUxqeIl8eFUNexGNmmk:-MUxqeIl8eFUNexGNmml:b-njcvw) ([raw file](https://github.com/estuary/flow/blob/2786467ff9dc863901316ddd1226c7b9782b728a/Makefile#L285)):*
> ```Makefile
> 
> .PHONY: test-pg-driver
> test-pg-driver: ${RUSTBIN}/libbindings.a crates/bindings/flow_bindings.h ${ROCKSDIR}/${LIBROCKS} ${ROOTDIR}/catalog.db
> ```

Making a note that we should talk through CI testing strategy for drivers. It's still an open / unresolved question for me -- especially for things we can't spin up in Docker.
___
*[examples/materialization-test.flow.yaml, line 1 at r5](https://reviewable.io/reviews/estuary/flow/115#-MUxr_YW8HPfoQtqfAca:-MUxrF510sooksHsMRUO:b-bkfb5j) ([raw file](https://github.com/estuary/flow/blob/2786467ff9dc863901316ddd1226c7b9782b728a/examples/materialization-test.flow.yaml#L1)):*
> ```YAML
> collections:
> ```

Yea ... as per my comment in `tester`, can we simply grab this spec as baked-in fixture? Then the driver tester doesn't need to link against libbindings / rocks / etc -- it can be a stand alone static executable.

Another bit of motivation is that `examples` is a dumping ground we need to clear out / organize / better think through, and I'm loathe to add more stuff here that's magically depended on by tests.
___
*[go/materialize/driver/postgres/config_test.go, line 4 at r3](https://reviewable.io/reviews/estuary/flow/115#-MUxTHQa5_WYlG06GAYD:-MUxTHQa5_WYlG06GAYE:b-njn1gt) ([raw file](https://github.com/estuary/flow/blob/f72aa920d46a5213e0e2d7964512002b0e70b860/go/materialize/driver/postgres/config_test.go#L4)):*
> ```Go
> 
> import (
> 	"github.com/stretchr/testify/require"
> ```

nit: `goimports` ordering (add break b/w stdlib packages and everything else)
___
*[go/materialize/driver/postgres/driver.go, line 104 at r1](https://reviewable.io/reviews/estuary/flow/115#-MUxZmjd9gx-MzmkEk4e:-MUxZmjd9gx-MzmkEk4f:b-atkcik) ([raw file](https://github.com/estuary/flow/blob/ba5f27939b9e97200a59e60d47af60e9157fbedb/go/materialize/driver/postgres/driver.go#L104)):*
> ```Go
> 	})
> 
> 	var targetTable = sqlDriver.TableForMaterialization(endpoint.Tables.Target, "", spec)
> ```

Okay, meta-comment I alluded to earlier, to work on name conventions.

As related resources, see [Code Review Comments](https://github.com/golang/go/wiki/CodeReviewComments#variable-names) and [Effective Go](https://golang.org/doc/effective_go#package-names). Reading the Go standard library is also really good for internalizing Go "taste" w.r.t.

The problem statement is: our Go code should read as if written by one voice, and that voice should be relatively indistinguishable from the Go standard library and collective idiomatic practice. Where there's legitimate disagreement on "taste", the presumptive correct answer is the practice of the standard library.

Same story with Rust code, but there we follow Rust std library, community practice, etc. We must avoid making Rust read like Go, or Go read like Rust, and it's on us to keep two separate models for how to "voice" code in our heads and switch between them at whim (which is hard, but there it is).

So, my intent here _really_ is not to bike shed, but to move us towards this common and shared voice, and I'm channeling what I've internalized myself from these resources, the stdlib, and past practice. I'm not infallible, my own channeled taste may be wrong, and IMO evidence of accepted practice in the community w.r.t. an element of taste always wins.

All that said:

nit: this should just be `target`.

Idiomatic Go is to use shorter names where possible. E.x. unless there's another type of semantic `target` in scope ... just use `target`. And generally speaking, the type of a variable -- as well as breadcrumbs towards its usage or implementation -- should _not_ be included in the variables name. It's readily available to the reader should they want to know, through a bit of reading or inspection tools.

The rationale is that I, as a reader, must do that just once to resolve all the semantic meaning for a variable or type "foo" in a scope/package, and then I can quickly visually recognize "foo" and it's associations thereafter. Whereas I must always unpack a "fooBarBaz" compound in each of its usages, and the name primes me to worry there's a "fooBarBing" floating around I should watch out for.

`keyParamsConverter` -- Effective Go discussed package stutter, but it applies here too -- `keyParamsConverter.Convert()` stutters.

Or `loadJoinQuery` -- there isn't another type of `load`, and the `Join` is an implementation detail of how it works. `loadSQL` ?

Also see "initialisms" in CodeReviewComments: `insertSql` is always `insertSQL`.

Picking on one final example: `storedDocs` below -- we're not storing another type of thing, and the `Docs` compound isn't adding value as a discriminant.

Much of this implementation wraps in my editor -- that's okay, and some wrapping lines are unavoidable -- but as "Line Length" in CRC says, often long lines kind of resolve themselves with more concise naming practices. I'm hesitant to put exact numbers on it, but if a name is a compound of more than two words, that's a smell.
___
*[go/materialize/driver/postgres/driver.go, line 16 at r5](https://reviewable.io/reviews/estuary/flow/115#-MUxlXaD0ltOD8fB4Ipk:-MUxlXaD0ltOD8fB4Ipl:b-dwnoo9) ([raw file](https://github.com/estuary/flow/blob/2786467ff9dc863901316ddd1226c7b9782b728a/go/materialize/driver/postgres/driver.go#L16)):*
> ```Go
> 	pf "github.com/estuary/flow/go/protocols/flow"
> 	pm "github.com/estuary/flow/go/protocols/materialize"
> ```

This looks like goimport -local ordering instead of goimports ?
___
*[go/materialize/driver/postgres/driver.go, line 122 at r5](https://reviewable.io/reviews/estuary/flow/115#-MUxgMfU9zkgU16rD8mw:-MUxgMfU9zkgU16rD8mx:bvomm8r) ([raw file](https://github.com/estuary/flow/blob/2786467ff9dc863901316ddd1226c7b9782b728a/go/materialize/driver/postgres/driver.go#L122)):*
> ```Go
> 		return fmt.Errorf("generating insert statement: %w", err)
> 	}
> 	var updateFields = spec.FieldSelection.Values
> ```

nit: this variable isn't used elsewhere; you can avoid it's introduction to the scope by inlining into UpdateStatement.
___
*[go/materialize/driver/postgres/driver.go, line 151 at r5](https://reviewable.io/reviews/estuary/flow/115#-MUxgfxy5oBMdBErb_7X:-MUxgfxy5oBMdBErb_7Y:b7qzob2) ([raw file](https://github.com/estuary/flow/blob/2786467ff9dc863901316ddd1226c7b9782b728a/go/materialize/driver/postgres/driver.go#L151)):*
> ```Go
> 	var response *pm.TransactionResponse
> 	for {
> 		if txn, err = conn.BeginTx(endpoint.Context, pgx.TxOptions{}); err != nil {
> ```

This can deadlock for reasons we discussed on slack. Check out how `RunSQLTransactions` now scopes the transaction _just_ to fence update / Stores.
___
*[go/materialize/driver/postgres/driver.go, line 154 at r5](https://reviewable.io/reviews/estuary/flow/115#-MUxg5zd580muFifOrby:-MUxg5zd580muFifOrbz:b-24005w) ([raw file](https://github.com/estuary/flow/blob/2786467ff9dc863901316ddd1226c7b9782b728a/go/materialize/driver/postgres/driver.go#L154)):*
> ```Go
> 			return fmt.Errorf("DB.BeginTx: %w", err)
> 		}
> 		var loadIterator = lifecycle.NewLoadIterator(stream)
> ```

nit: `loadIt` ?  `it` is more idiomatic, see examples of https://pkg.go.dev/google.golang.org/api/iterator
___
*[go/materialize/driver/postgres/driver.go, line 176 at r5](https://reviewable.io/reviews/estuary/flow/115#-MUxhX8W9y7LxxDZwh2E:-MUxhX8W9y7LxxDZwh2F:bxcxzw9) ([raw file](https://github.com/estuary/flow/blob/2786467ff9dc863901316ddd1226c7b9782b728a/go/materialize/driver/postgres/driver.go#L176)):*
> ```Go
> 				foundDocs++
> 				var json json.RawMessage
> 				err = rows.Scan(&json)
> ```

idiom is to inline this where new variables aren't being declared and it's reasonably short, as `if err = ... ; err != nil { ` ; there are other places that should do this I'm not commenting individually.

I kind of spit-ball "reasonably short" as "less than 80 characters column length", the length limit for comments in Go (but not SLOC). Where it's more, it's likely to wrap anyway so using multiple lines is cleaner without adding effective line count.
___
*[go/materialize/driver/postgres/driver.go, line 247 at r5](https://reviewable.io/reviews/estuary/flow/115#-MUxkZUYCuefCI1t-4Bc:-MUxkZUYCuefCI1t-4Bd:butjwu2) ([raw file](https://github.com/estuary/flow/blob/2786467ff9dc863901316ddd1226c7b9782b728a/go/materialize/driver/postgres/driver.go#L247)):*
> ```Go
> 		}
> 
> 		// Skip sending the batch if the client didn't send any Store requests
> ```

could the fence update be included in the batch as well ?

then the Batch is always non-empty and this branch can be dropped, plus we don't require a separate RPC for the fence.
___
*[go/materialize/driver/postgres/driver.go, line 274 at r5](https://reviewable.io/reviews/estuary/flow/115#-MUxlDU25IM6gwdgOA06:-MUxlDU25IM6gwdgOA07:bhemaq9) ([raw file](https://github.com/estuary/flow/blob/2786467ff9dc863901316ddd1226c7b9782b728a/go/materialize/driver/postgres/driver.go#L274)):*
> ```Go
> 		err = lifecycle.WriteCommitted(stream, &response)
> 		if err != nil {
> 			// This isn't really a problem since we will return the proper flow_checkpoint on the
> ```

Why should we log here, since we're immediately returning an error ? (which will tear down the shard and report the error, which I think is desired behavior )
___
*[go/materialize/driver/sql2/fence.go, line 17 at r1](https://reviewable.io/reviews/estuary/flow/115#-MUxb8to0XvN5FUhBtYN:-MUxb8tp-ObCAFdfPt2w:b-9iyvpi) ([raw file](https://github.com/estuary/flow/blob/ba5f27939b9e97200a59e60d47af60e9157fbedb/go/materialize/driver/sql2/fence.go#L17)):*
> ```Go
> 	// Checkpoint associated with this Fence.
> 	Checkpoint []byte
> 	// Fence is the current value of the monotonically increasing integer used to identify unique
> ```

I don't think these should be public, but they could be read-only exposed via functions Fence() and ShardFQN()
___
*[go/materialize/tester/docs.go, line 14 at r5](https://reviewable.io/reviews/estuary/flow/115#-MUxnqnr3KskNpDeyxG9:-MUxnqnr3KskNpDeyxGA:beytwwj) ([raw file](https://github.com/estuary/flow/blob/2786467ff9dc863901316ddd1226c7b9782b728a/go/materialize/tester/docs.go#L14)):*
> ```Go
> )
> 
> type testDocGenerator struct {
> ```

sorry to keep hammering on this:
 * we're already in a `tester` package so "test" is implicit. 
 * unless there's another type of thing we're generating, "Doc" is also implicit

Just `generator`.

Below, `tester.TestDoc` stutters (see "Effective Go").
___
*[go/materialize/tester/docs.go, line 88 at r5](https://reviewable.io/reviews/estuary/flow/115#-MUxoYdk61evtIafD9q6:-MUxoYdk61evtIafD9q7:bftxta7) ([raw file](https://github.com/estuary/flow/blob/2786467ff9dc863901316ddd1226c7b9782b728a/go/materialize/tester/docs.go#L88)):*
> ```Go
> }
> 
> func (g *testDocGenerator) GenerateTestDoc() *TestDoc {
> ```

consider how this is used: `generator.GenerateTestDoc()` could be `generator.Next()` or similar
___
*[go/materialize/tester/docs.go, line 119 at r5](https://reviewable.io/reviews/estuary/flow/115#-MUxoykR8uViU0LOifOz:-MUxoykR8uViU0LOifP-:b-w5n403) ([raw file](https://github.com/estuary/flow/blob/2786467ff9dc863901316ddd1226c7b9782b728a/go/materialize/tester/docs.go#L119)):*
> ```Go
> }
> 
> func (g *testDocGenerator) UpdateValues(doc *TestDoc) {
> ```

doc comments? i'm unclear on why UpdateValues exists at first glance (why does generator touch documents it's already generated?)
___
*[go/materialize/tester/fixture.go, line 196 at r5](https://reviewable.io/reviews/estuary/flow/115#-MUxpW5R1qMr2uiJGX6J:-MUxpW5R1qMr2uiJGX6K:b9m5wj2) ([raw file](https://github.com/estuary/flow/blob/2786467ff9dc863901316ddd1226c7b9782b728a/go/materialize/tester/fixture.go#L196)):*
> ```Go
> }
> 
> // NewTestMaterialization returns a MaterializationSpec for use by tests. The collection spec and
> ```

This is great, as I recently found myself wanting a handy MaterializationSpec in order to test Snowflake SQL generation.

I think we can just inline a literal `MaterializationSpec` here though, and avoid questions of catalog build / availability. It could even be lifted from the existing catalog fixture ?

jgraettinger (changes requested):
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/115)* status: all files reviewed, 16 unresolved discussions (waiting on @psFried)

___
*[go/materialize/tester/fixture.go, line 196 at r5](https://reviewable.io/reviews/estuary/flow/115#-MUxpW5R1qMr2uiJGX6J:-MUxuGfy94cMCudfp1GO:b-pcvxdm) ([raw file](https://github.com/estuary/flow/blob/2786467ff9dc863901316ddd1226c7b9782b728a/go/materialize/tester/fixture.go#L196)):*

This is great, as I recently found myself wanting a handy MaterializationSpec in order to test Snowflake SQL generation.

I think we can just inline a literal `MaterializationSpec` here though, and avoid questions of catalog build / availability. It could even be lifted from the existing catalog fixture ?
</blockquote></details>

in case it's helpful, a hack for quickly doing this is to run this as-is, but with a `require.Equal(t, nil, loadedSpec)`, and copy-paste in the literal it spits out

jgraettinger (changes requested):
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/115)* status: all files reviewed, 17 unresolved discussions (waiting on @psFried)

___
*[go/materialize/driver/sql2/sqlgen.go, line 453 at r5](https://reviewable.io/reviews/estuary/flow/115#-MV1eymF9v3arsMcilJw:-MV1eymF9v3arsMcilJx:bgtqq13) ([raw file](https://github.com/estuary/flow/blob/2786467ff9dc863901316ddd1226c7b9782b728a/go/materialize/driver/sql2/sqlgen.go#L453)):*
> ```Go
> 	// Close the primary key paren, then newline and close the create table statement
> 	builder.WriteString(")\n)")
> 	if table.Temporary && table.TempOnCommit != "" {
> ```

I'm working up a similar "load keys into temp table, then join" strategy for SQLite, and I'm realizing that this is a postgres specific extension, making this unfortunate coupling in `Table`. 

More broadly I'm finding the `Table` structure itself is helpful, but that these top-level DDL generation routines aren't carrying their weight. So far, I'm much more typically using them just to get converters for keys or table rows, and then taking responsibility for manually crafting SQL that matches the DB dialect.

Perhaps we should more directly embrace this pattern, and look at `Generator` / `Table` as providing supporting routines / high level structure in the practice of crafting DB-specific SQL ?

jgraettinger (changes requested):
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/115)* status: all files reviewed, 17 unresolved discussions (waiting on @psFried)

___
*[go/materialize/driver/sql2/sqlgen.go, line 453 at r5](https://reviewable.io/reviews/estuary/flow/115#-MV1eymF9v3arsMcilJw:-MV1hYcJAXf5YRf-hEpF:b-qn2q2k) ([raw file](https://github.com/estuary/flow/blob/2786467ff9dc863901316ddd1226c7b9782b728a/go/materialize/driver/sql2/sqlgen.go#L453)):*

I'm working up a similar "load keys into temp table, then join" strategy for SQLite, and I'm realizing that this is a postgres specific extension, making this unfortunate coupling in `Table`. 

More broadly I'm finding the `Table` structure itself is helpful, but that these top-level DDL generation routines aren't carrying their weight. So far, I'm much more typically using them just to get converters for keys or table rows, and then taking responsibility for manually crafting SQL that matches the DB dialect.

Perhaps we should more directly embrace this pattern, and look at `Generator` / `Table` as providing supporting routines / high level structure in the practice of crafting DB-specific SQL ?
</blockquote></details>

BTW not suggesting it for this PR -- I think this is fine just to get this landed -- I'm working around these types for now, but would like to come back and think about the interface more holistically (along with identifier quoting).

psFried (commented):
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/115)* status: 12 of 22 files reviewed, 14 unresolved discussions (waiting on @jgraettinger and @psFried)

___
*[Makefile, line 285 at r5](https://reviewable.io/reviews/estuary/flow/115#-MUxqeIl8eFUNexGNmmk:-MUz19x25g4N6Ma5h5ou:bq98m50) ([raw file](https://github.com/estuary/flow/blob/2786467ff9dc863901316ddd1226c7b9782b728a/Makefile#L285)):*

Making a note that we should talk through CI testing strategy for drivers. It's still an open / unresolved question for me -- especially for things we can't spin up in Docker.
</blockquote></details>

I also still have a lot of open/unresolved questions about how we should be testing drivers, and would like to talk about that soon. I'll refrain from dumping my thoughts on that here, but it's definitely something that's been on my mind.
___
*[examples/materialization-test.flow.yaml, line 1 at r5](https://reviewable.io/reviews/estuary/flow/115#-MUxr_YW8HPfoQtqfAca:-MUz1TbyDLZyrn-uD5m2:b-mtph6q) ([raw file](https://github.com/estuary/flow/blob/2786467ff9dc863901316ddd1226c7b9782b728a/examples/materialization-test.flow.yaml#L1)):*

Yea ... as per my comment in `tester`, can we simply grab this spec as baked-in fixture? Then the driver tester doesn't need to link against libbindings / rocks / etc -- it can be a stand alone static executable.

Another bit of motivation is that `examples` is a dumping ground we need to clear out / organize / better think through, and I'm loathe to add more stuff here that's magically depended on by tests.
</blockquote></details>

Yeah, I think the trick with getting a test to spit out the Go struct representation is exactly the pattern I was looking for. I actually started out typing this all out manually, and only switched to the catalog approach after it became clear what a pain that was going to be. Would be great to decouple all of that.
___
*[go/materialize/driver/postgres/driver.go, line 104 at r1](https://reviewable.io/reviews/estuary/flow/115#-MUxZmjd9gx-MzmkEk4e:-MUzCMut2WzXF7qrdjbg:b-3l426z) ([raw file](https://github.com/estuary/flow/blob/ba5f27939b9e97200a59e60d47af60e9157fbedb/go/materialize/driver/postgres/driver.go#L104)):*

Okay, meta-comment I alluded to earlier, to work on name conventions.

As related resources, see [Code Review Comments](https://github.com/golang/go/wiki/CodeReviewComments#variable-names) and [Effective Go](https://golang.org/doc/effective_go#package-names). Reading the Go standard library is also really good for internalizing Go "taste" w.r.t.

The problem statement is: our Go code should read as if written by one voice, and that voice should be relatively indistinguishable from the Go standard library and collective idiomatic practice. Where there's legitimate disagreement on "taste", the presumptive correct answer is the practice of the standard library.

Same story with Rust code, but there we follow Rust std library, community practice, etc. We must avoid making Rust read like Go, or Go read like Rust, and it's on us to keep two separate models for how to "voice" code in our heads and switch between them at whim (which is hard, but there it is).

So, my intent here _really_ is not to bike shed, but to move us towards this common and shared voice, and I'm channeling what I've internalized myself from these resources, the stdlib, and past practice. I'm not infallible, my own channeled taste may be wrong, and IMO evidence of accepted practice in the community w.r.t. an element of taste always wins.

All that said:

nit: this should just be `target`.

Idiomatic Go is to use shorter names where possible. E.x. unless there's another type of semantic `target` in scope ... just use `target`. And generally speaking, the type of a variable -- as well as breadcrumbs towards its usage or implementation -- should _not_ be included in the variables name. It's readily available to the reader should they want to know, through a bit of reading or inspection tools.

The rationale is that I, as a reader, must do that just once to resolve all the semantic meaning for a variable or type "foo" in a scope/package, and then I can quickly visually recognize "foo" and it's associations thereafter. Whereas I must always unpack a "fooBarBaz" compound in each of its usages, and the name primes me to worry there's a "fooBarBing" floating around I should watch out for.

`keyParamsConverter` -- Effective Go discussed package stutter, but it applies here too -- `keyParamsConverter.Convert()` stutters.

Or `loadJoinQuery` -- there isn't another type of `load`, and the `Join` is an implementation detail of how it works. `loadSQL` ?

Also see "initialisms" in CodeReviewComments: `insertSql` is always `insertSQL`.

Picking on one final example: `storedDocs` below -- we're not storing another type of thing, and the `Docs` compound isn't adding value as a discriminant.

Much of this implementation wraps in my editor -- that's okay, and some wrapping lines are unavoidable -- but as "Line Length" in CRC says, often long lines kind of resolve themselves with more concise naming practices. I'm hesitant to put exact numbers on it, but if a name is a compound of more than two words, that's a smell.
</blockquote></details>

As much as I disagree with some of the go idioms (for completely sensible reasons of course :stuck_out_tongue_winking_eye: ), I would rather conform with those idioms and ensure that we can keep this stuff maintainable over the long term. So I definitely won't take offense to corrections there. I tried to just review all the naming for conformance.
___
*[go/materialize/driver/postgres/driver.go, line 16 at r5](https://reviewable.io/reviews/estuary/flow/115#-MUxlXaD0ltOD8fB4Ipk:-MUz2Bka9Ln5jFznvtn4:b-83z2ub) ([raw file](https://github.com/estuary/flow/blob/2786467ff9dc863901316ddd1226c7b9782b728a/go/materialize/driver/postgres/driver.go#L16)):*

This looks like goimport -local ordering instead of goimports ?
</blockquote></details>

I had been assuming that gofmt takes care of imports, too. But now I'm realizing that goimports is what actually arranges these according to the established conventions. Dunno what introduced the blank line here, but I'll try to get goimports setup to take care of those automatically.
___
*[go/materialize/driver/postgres/driver.go, line 176 at r5](https://reviewable.io/reviews/estuary/flow/115#-MUxhX8W9y7LxxDZwh2E:-MUzEyJiCSNNFAbfGcpF:b-896fix) ([raw file](https://github.com/estuary/flow/blob/2786467ff9dc863901316ddd1226c7b9782b728a/go/materialize/driver/postgres/driver.go#L176)):*

idiom is to inline this where new variables aren't being declared and it's reasonably short, as `if err = ... ; err != nil { ` ; there are other places that should do this I'm not commenting individually.

I kind of spit-ball "reasonably short" as "less than 80 characters column length", the length limit for comments in Go (but not SLOC). Where it's more, it's likely to wrap anyway so using multiple lines is cleaner without adding effective line count.
</blockquote></details>

Done.
___
*[go/materialize/driver/postgres/driver.go, line 247 at r5](https://reviewable.io/reviews/estuary/flow/115#-MUxkZUYCuefCI1t-4Bc:-MV1mchPDYgjdNMsH7lu:b-965atw) ([raw file](https://github.com/estuary/flow/blob/2786467ff9dc863901316ddd1226c7b9782b728a/go/materialize/driver/postgres/driver.go#L247)):*

could the fence update be included in the batch as well ?

then the Batch is always non-empty and this branch can be dropped, plus we don't require a separate RPC for the fence.
</blockquote></details>

Yep, turns out it can. I thought this was going to make a mess because we have to consume all the batch results as part of the `Fence.Update` callback, but it isn't actually too bad. Worth it to save a round trip.
___
*[go/materialize/driver/postgres/driver.go, line 274 at r5](https://reviewable.io/reviews/estuary/flow/115#-MUxlDU25IM6gwdgOA06:-MUzF4ceAJSfpM6mFhpG:bch1d8e) ([raw file](https://github.com/estuary/flow/blob/2786467ff9dc863901316ddd1226c7b9782b728a/go/materialize/driver/postgres/driver.go#L274)):*

Why should we log here, since we're immediately returning an error ? (which will tear down the shard and report the error, which I think is desired behavior )
</blockquote></details>

That's fine, I just wanted to make sure we logged that this error occurring after the commit so we could know about it if it started happening more often than expected. I can just wrap the error.
___
*[go/materialize/driver/sql2/fence.go, line 17 at r1](https://reviewable.io/reviews/estuary/flow/115#-MUxb8to0XvN5FUhBtYN:-MV1gwp28CZ4HdMAwPuI:b7zmzwy) ([raw file](https://github.com/estuary/flow/blob/ba5f27939b9e97200a59e60d47af60e9157fbedb/go/materialize/driver/sql2/fence.go#L17)):*

I don't think these should be public, but they could be read-only exposed via functions Fence() and ShardFQN()
</blockquote></details>

Looks like the only place I ended up using them was to add those fields to the log entry, so I just added a `func (f *Fence) LogEntry() *log.Entry` instead. That might be nice anyway, just to encourage some more consistency in the fields we add to logs. Looking ahead to the possibility that we might have to sort through logs from a busy multi-tenant data plane.
___
*[go/materialize/driver/sql2/sqlgen.go, line 453 at r5](https://reviewable.io/reviews/estuary/flow/115#-MV1eymF9v3arsMcilJw:-MV1iEll74wQi99Z-Y_y:b-qnexwl) ([raw file](https://github.com/estuary/flow/blob/2786467ff9dc863901316ddd1226c7b9782b728a/go/materialize/driver/sql2/sqlgen.go#L453)):*

BTW not suggesting it for this PR -- I think this is fine just to get this landed -- I'm working around these types for now, but would like to come back and think about the interface more holistically (along with identifier quoting).
</blockquote></details>

`ON COMMIT` is definitely supported by other databases, at least by Oracle off the top of my head. That said, I think your point still stands, as it does seem like we might want a way to customize `CREATE TABLE` statements without always adding to a global table type. But I'm not seeing it as a high priority, since the interface seems to get us what we need for now, and I'm guessing a better factoring might be more obvious once we have a specific use case that motivates it.
___
*[go/materialize/tester/docs.go, line 14 at r5](https://reviewable.io/reviews/estuary/flow/115#-MUxnqnr3KskNpDeyxG9:-MV1s9wVCSC61IU9KGBG:b-896fix) ([raw file](https://github.com/estuary/flow/blob/2786467ff9dc863901316ddd1226c7b9782b728a/go/materialize/tester/docs.go#L14)):*

sorry to keep hammering on this:
 * we're already in a `tester` package so "test" is implicit. 
 * unless there's another type of thing we're generating, "Doc" is also implicit

Just `generator`.

Below, `tester.TestDoc` stutters (see "Effective Go").
</blockquote></details>

Done.
___
*[go/materialize/tester/docs.go, line 88 at r5](https://reviewable.io/reviews/estuary/flow/115#-MUxoYdk61evtIafD9q6:-MV1rxx5DTplvdelTP0u:b-896fix) ([raw file](https://github.com/estuary/flow/blob/2786467ff9dc863901316ddd1226c7b9782b728a/go/materialize/tester/docs.go#L88)):*

consider how this is used: `generator.GenerateTestDoc()` could be `generator.Next()` or similar
</blockquote></details>

Done.

jgraettinger (approved):
:lgtm_strong: % a couple tiny nits. Thanks, I'll rebase on this once landed

> Reviewed 13 of 13 files at r6.
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/115#-:-MV2VsDIE7ZzmTheGItA:bvfhj7c)* status: all files reviewed, 4 unresolved discussions (waiting on @psFried)

___
*[go/materialize/driver/postgres/driver.go, line 104 at r1](https://reviewable.io/reviews/estuary/flow/115#-MUxZmjd9gx-MzmkEk4e:-MV2Sibb3Xt9hUeOaY67:bc3i5sw) ([raw file](https://github.com/estuary/flow/blob/ba5f27939b9e97200a59e60d47af60e9157fbedb/go/materialize/driver/postgres/driver.go#L104)):*

As much as I disagree with some of the go idioms (for completely sensible reasons of course :stuck_out_tongue_winking_eye: ), I would rather conform with those idioms and ensure that we can keep this stuff maintainable over the long term. So I definitely won't take offense to corrections there. I tried to just review all the naming for conformance.
</blockquote></details>

:thumbsup: appreciate it
___
*[go/materialize/driver/postgres/driver.go, line 16 at r5](https://reviewable.io/reviews/estuary/flow/115#-MUxlXaD0ltOD8fB4Ipk:-MV2RdeaB675lhpI9jhP:by4yare) ([raw file](https://github.com/estuary/flow/blob/2786467ff9dc863901316ddd1226c7b9782b728a/go/materialize/driver/postgres/driver.go#L16)):*

I had been assuming that gofmt takes care of imports, too. But now I'm realizing that goimports is what actually arranges these according to the established conventions. Dunno what introduced the blank line here, but I'll try to get goimports setup to take care of those automatically.
</blockquote></details>

yea this is `goimport -local` mode, which _also_ separates module packages from other third-party from stdlib. Which is fine, I have no objections to the mode, but the rest of the project is using standard `goimport` (which just splits stdlib vs everything else).
___
*[go/materialize/driver/postgres/driver.go, line 151 at r5](https://reviewable.io/reviews/estuary/flow/115#-MUxgfxy5oBMdBErb_7X:-MV2SEO418DRzRwtcV3z:b93qu52) ([raw file](https://github.com/estuary/flow/blob/2786467ff9dc863901316ddd1226c7b9782b728a/go/materialize/driver/postgres/driver.go#L151)):*

This can deadlock for reasons we discussed on slack. Check out how `RunSQLTransactions` now scopes the transaction _just_ to fence update / Stores.
</blockquote></details>

As discussed via Slack, this will be resolved w/ the upcoming `Transactor` interface.
___
*[go/materialize/driver/postgres/driver.go, line 230 at r6](https://reviewable.io/reviews/estuary/flow/115#-MV2TKWy-pHllFthssn4:-MV2TKWy-pHllFthssn5:b25fdwy) ([raw file](https://github.com/estuary/flow/blob/078c5f6ffba6541c180026806b434f9493ae7998/go/materialize/driver/postgres/driver.go#L230)):*
> ```Go
> 		fence.Checkpoint = prepare.FlowCheckpoint
> 		err = fence.Update(func(ctx context.Context, sql string, args ...interface{}) (int64, error) {
> 			if _, err = txn.Prepare(endpoint.Context, "updateFence", sql); err != nil {
> ```

nit: not a problem per-se to txn.Prepare this, but it's only invoked once and batch.Queue appears to allow the literal SQL to be used instead of `updateFence` ? plus doesn't Prepare require a RPC ?
___
*[go/materialize/driver/sql2/fence.go, line 17 at r1](https://reviewable.io/reviews/estuary/flow/115#-MUxb8to0XvN5FUhBtYN:-MV2UJ2b5sNeKQyAf9dT:b-8fx2u7) ([raw file](https://github.com/estuary/flow/blob/ba5f27939b9e97200a59e60d47af60e9157fbedb/go/materialize/driver/sql2/fence.go#L17)):*

Looks like the only place I ended up using them was to add those fields to the log entry, so I just added a `func (f *Fence) LogEntry() *log.Entry` instead. That might be nice anyway, just to encourage some more consistency in the fields we add to logs. Looking ahead to the possibility that we might have to sort through logs from a busy multi-tenant data plane.
</blockquote></details>

:+1: 
___
*[go/materialize/driver/sql2/sqlgen.go, line 453 at r5](https://reviewable.io/reviews/estuary/flow/115#-MV1eymF9v3arsMcilJw:-MV2UbfJ0vP6n2vP9-Be:b-q3pk6f) ([raw file](https://github.com/estuary/flow/blob/2786467ff9dc863901316ddd1226c7b9782b728a/go/materialize/driver/sql2/sqlgen.go#L453)):*

`ON COMMIT` is definitely supported by other databases, at least by Oracle off the top of my head. That said, I think your point still stands, as it does seem like we might want a way to customize `CREATE TABLE` statements without always adding to a global table type. But I'm not seeing it as a high priority, since the interface seems to get us what we need for now, and I'm guessing a better factoring might be more obvious once we have a specific use case that motivates it.
</blockquote></details>

cool let's land this as-is, and i'll put up a PR showing how i've evolved building SQL on the current types, and we can take it from there. Or not.
___
*[go/materialize/tester/fixture.go, line 196 at r6](https://reviewable.io/reviews/estuary/flow/115#-MV2VKg84eirGte4SnmZ:-MV2VKg84eirGte4Snm_:b-jj1s7w) ([raw file](https://github.com/estuary/flow/blob/078c5f6ffba6541c180026806b434f9493ae7998/go/materialize/tester/fixture.go#L196)):*
> ```Go
> // NewTestMaterialization returns a MaterializationSpec for use by tests. This is a hard coded
> // materialization that includes a field of each type.
> func NewTestMaterialization(endpointType pf.EndpointType, endpointConfigJson string) *pf.MaterializationSpec {
> ```

Awesome, I have immediate need for this / am excited to use it.

supernit: NewTestMaterialization -- Test is implied since we're in "tester"

psFried (commented):
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/115)* status: 20 of 22 files reviewed, 4 unresolved discussions (waiting on @jgraettinger)

___
*[go/materialize/driver/postgres/driver.go, line 151 at r5](https://reviewable.io/reviews/estuary/flow/115#-MUxgfxy5oBMdBErb_7X:-MV2W_wz197Budkz7d_L:b-896fix) ([raw file](https://github.com/estuary/flow/blob/2786467ff9dc863901316ddd1226c7b9782b728a/go/materialize/driver/postgres/driver.go#L151)):*

As discussed via Slack, this will be resolved w/ the upcoming `Transactor` interface.
</blockquote></details>

Done.
___
*[go/materialize/driver/postgres/driver.go, line 230 at r6](https://reviewable.io/reviews/estuary/flow/115#-MV2TKWy-pHllFthssn4:-MV2WfEx7JwmAZIWSDcN:b24koh8) ([raw file](https://github.com/estuary/flow/blob/078c5f6ffba6541c180026806b434f9493ae7998/go/materialize/driver/postgres/driver.go#L230)):*

nit: not a problem per-se to txn.Prepare this, but it's only invoked once and batch.Queue appears to allow the literal SQL to be used instead of `updateFence` ? plus doesn't Prepare require a RPC ?
</blockquote></details>

Ah, I had misunderstood the docs and thought that batch _must_ use the name of a prepared statement. Turns out you don't need to use a prepared statement at all. I'll fix it to just use the sql directly.
___
*[go/materialize/driver/sql2/sqlgen.go, line 453 at r5](https://reviewable.io/reviews/estuary/flow/115#-MV1eymF9v3arsMcilJw:-MV2XLhC033zyuy0Mr_B:b-896fix) ([raw file](https://github.com/estuary/flow/blob/2786467ff9dc863901316ddd1226c7b9782b728a/go/materialize/driver/sql2/sqlgen.go#L453)):*

cool let's land this as-is, and i'll put up a PR showing how i've evolved building SQL on the current types, and we can take it from there. Or not.
</blockquote></details>

Done.
___
*[go/materialize/tester/fixture.go, line 196 at r6](https://reviewable.io/reviews/estuary/flow/115#-MV2VKg84eirGte4SnmZ:-MV2XSPk4OW-kYNuwGn2:b-896fix) ([raw file](https://github.com/estuary/flow/blob/078c5f6ffba6541c180026806b434f9493ae7998/go/materialize/tester/fixture.go#L196)):*

Awesome, I have immediate need for this / am excited to use it.

supernit: NewTestMaterialization -- Test is implied since we're in "tester"
</blockquote></details>

Done.

jgraettinger (approved):
> Reviewed 2 of 2 files at r7.
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/115)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @psFried)

PR #116, opened 2021-03-08
Stop quoting database identifiers

psFried:
For some databases, quoting identifiers (table and column names) in
CREATE TABLE statements forces you to always quote the identifiers in
all queries against that table. Since the purpose of materialization is
often to facilitate easy writing of ad hoc queries, requiring quotes
around all identifiers can be a major usability problem. This is
especially true given the horrific state of error reporting among
databases, which typically do not mention quoting in their error
messages. So the decision is to stop quoting all identifiers by default.
In the future, we may re-enable quoting on an opt-in basis, but the
intent is for this never to be done by default.

This PR does not add any supplemental check for validity of identifiers. Instead we rely on the database to surface an error if any identifiers are invalid due to not being quoted. The rationale for _not_ doing this check is basically just that it's way more difficult that you'd expect to match the behavior of the database. For example, a simple check of "does this identifier contain any disallowed characters" would sometimes prevent users from using an identifier that sqlite would otherwise allow. Additionally, the assumption is that users will likely be fairly familiar with whatever database they want to materialize into, and will already have a decent handle on what identifiers it allows. This conclusion came only after a yak shave, so I still have some code for doing a "basic" identifier validity check, which is sitting on another branch should we decide we want it after all.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/116)


jgraettinger (commented):
I'm confused, what happened to the e.x. `[\p{letter}\p{number}_]+` behavior we discussed ?
Also, the need is for a short simple function to call which will spit out an identifier given a literal (`Ident(string) -> string` ?), for use in composing SQL _outside_ of `Generator`'s existing functions. Thanks

> *[Reviewable](https://reviewable.io/reviews/estuary/flow/116#-:-MVH5gcIAEuhgu-WMFAu:b-tvco05)* status: 0 of 9 files reviewed, all discussions resolved

psFried:
> I'm confused, what happened to the e.x. `[\p{letter}\p{number}_]+` behavior we discussed ?

This is what I was referring to as a "supplemental check for validity of identifiers" in my comment above. I thought it better to just not have this check at all, given the possibility that our validity check might reject an identifier that the database would accept. It might be reasonable to consider such identifiers to be "corner cases" that we sacrifice in order to get better error messages for more common cases. But it seems better to avoid that for now.

> Also, the need is for a short simple function to call which will spit out an identifier given a literal (`Ident(string) -> string` ?), for use in composing SQL _outside_ of `Generator`'s existing functions.

I thought this would only be needed once we support quoting of identifiers, so I left it out of this PR. The rationale for not including it as a "placeholder" function is that I'm not yet clear on how we should thread though user-provided configuration for opting in to quoted identifiers. For example, if the SQL Generator needs the `FieldConfig` in order to know what to quote, then how should we be passing that? One possibility might even be to use the `FieldConfig` only in the conversion from `MaterializationSpec => Table`, making the `Column.Name` _already_ quoted. 

Based on our slack thread, I thought that the most important bit was to first stop quoting anything by default, and then add in a feature to allow quoting on an opt-in basis later. Upon re-reading that thread, now I can see how I might have misinterpreted some of that, so maybe we can touch base on this in our VC today.

jgraettinger:
>  that we sacrifice in order to get better error messages for more common cases. But it seems better to avoid that for now

As a practical matter, this will straight up break existing materializations / tests which utilize `foo/bar` encoding for nested locations. We can think about changing this encoding to use a double underscore, but this behavior pivots that choice from being an orthogonal concern to needing to be done in lockstep with this change.

More broadly, it will be necessary to quote things sometimes. I don't think it's a bad thing to attempt to quote identifiers where we know most every database under the sun won't accept it bare.

The guiding principal w.r.t. user contract in this area, IMO, is "be as ergonomic as possible, _without being surprising_" -- a regex like the above is easy to explain and it's easy to understand when it applies. I'd expect scenarios where this isn't the "right" behavior to be quite rare -- and only then would you need to figure out how to enable/disable fine-grain quoting, and on balance that's a good tradeoff for users.

> I thought this would only be needed once we support quoting of identifiers, so I left it out of this PR. The rationale for not including it as a "placeholder" function is that I'm not yet clear on how we should thread though user-provided configuration for opting in to quoted identifiers.

I have a bunch of places where I'd like to _use_ this function on a working branch. Even if it's a no-op I'd still like to wire it up so I'm not using literal Sprintf("%q") -- these will proliferate further as we add drivers, and be easy to miss when we do this "for real". I'm also unclear re: how it's wired, I was hoping you'd come up with a principled handling :) I'd think either the plain string, or the table Column.

jgraettinger:
> One possibility might even be to use the FieldConfig only in the conversion from MaterializationSpec => Table, making the Column.Name already quoted.

Ohh, I like this idea a lot. We'd want to do the same with Table.Name of course, but that would clean it up. Though it would complicate lookups where we're mapping from `FieldSelection` fields to the Column -- we could just have a `Ident` field alongside `Name` ?

jgraettinger (changes requested):
> Reviewed 18 of 18 files at r1.
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/116)* status: all files reviewed, 3 unresolved discussions (waiting on @psFried)

___
*[go/materialize/driver/sql2/spec_mapping.go, line 23 at r1](https://reviewable.io/reviews/estuary/flow/116#-MVIe-fz-f4tdlog8TkV:-MVIe-fz-f4tdlog8TkW:b-dgjp4s) ([raw file](https://github.com/estuary/flow/blob/16fc492301a41906b206ecef94a9c37c230220c9/go/materialize/driver/sql2/spec_mapping.go#L23)):*
> ```Go
> }
> 
> var unquotedIdentifiers = regexp.MustCompile("^[_a-zA-Z]+[_\\pL\\pN]*$")
> ```

why do we care if the first character is ascii (why not `^[_\\pL][_\\pL\\pN]*$` )? 
___
*[go/materialize/driver/sql2/spec_mapping.go, line 48 at r1](https://reviewable.io/reviews/estuary/flow/116#-MVIf5fYD8ih_8Jr99fp:-MVIf5fYD8ih_8Jr99fq:bvksodg) ([raw file](https://github.com/estuary/flow/blob/16fc492301a41906b206ecef94a9c37c230220c9/go/materialize/driver/sql2/spec_mapping.go#L48)):*
> ```Go
> 	var column = Column{
> 		Name:       projection.Field,
> 		Identifier: toIdentifier(projection.Field, identifierQuotes),
> ```

nit: This line has three "identifier"'s in it.

Repetition of `identifierQuotes *TokenPair` is a bit bleh given the only purpose of TokenPair is quoting AFAIK. Perhaps rename TokenPair => Quotes, and a variable of it is `quotes` ?
___
*[go/materialize/driver/sql2/sqlgen.go, line 127 at r1](https://reviewable.io/reviews/estuary/flow/116#-MVIfg-q2svqvpCBeSuv:-MVIfg-r0-cHhdEYUE9P:b-b7ssvc) ([raw file](https://github.com/estuary/flow/blob/16fc492301a41906b206ecef94a9c37c230220c9/go/materialize/driver/sql2/sqlgen.go#L127)):*
> ```Go
> }
> 
> func (p *TokenPair) Wrap(text string) string {
> ```

nit: Do you have an integrated linter ? E.x. in VSCode this shows up as a yellow squiggly to remind me to write comments on exported methods.

psFried (commented):
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/116)* status: 15 of 18 files reviewed, 3 unresolved discussions (waiting on @jgraettinger)

___
*[go/materialize/driver/sql2/spec_mapping.go, line 23 at r1](https://reviewable.io/reviews/estuary/flow/116#-MVIe-fz-f4tdlog8TkV:-MVIoXzkA3b51uBrF_Gn:b-716vxc) ([raw file](https://github.com/estuary/flow/blob/16fc492301a41906b206ecef94a9c37c230220c9/go/materialize/driver/sql2/spec_mapping.go#L23)):*

why do we care if the first character is ascii (why not `^[_\\pL][_\\pL\\pN]*$` )? 
</blockquote></details>

It seems that I've mis-read the postgres documentation on names! Glad you brought it up. I'll open it up to `\pL`.
___
*[go/materialize/driver/sql2/spec_mapping.go, line 48 at r1](https://reviewable.io/reviews/estuary/flow/116#-MVIf5fYD8ih_8Jr99fp:-MVImfx_1TG4VOFIGc2p:bgull1h) ([raw file](https://github.com/estuary/flow/blob/16fc492301a41906b206ecef94a9c37c230220c9/go/materialize/driver/sql2/spec_mapping.go#L48)):*

nit: This line has three "identifier"'s in it.

Repetition of `identifierQuotes *TokenPair` is a bit bleh given the only purpose of TokenPair is quoting AFAIK. Perhaps rename TokenPair => Quotes, and a variable of it is `quotes` ?
</blockquote></details>

Yeah I agree that this reads a little too java-like. `TokenPair` is also used for wrapping comments. I could make a different struct for the comments, though, if you think that'd be worth it. I kinda thought not, but could be convinced otherwise. I think either way I could at least rename `identifierQuotes` to just `quotes`.
___
*[go/materialize/driver/sql2/sqlgen.go, line 127 at r1](https://reviewable.io/reviews/estuary/flow/116#-MVIfg-q2svqvpCBeSuv:-MVIqJPVDN-NSoXVrxhf:b-tqwa3c) ([raw file](https://github.com/estuary/flow/blob/16fc492301a41906b206ecef94a9c37c230220c9/go/materialize/driver/sql2/sqlgen.go#L127)):*

nit: Do you have an integrated linter ? E.x. in VSCode this shows up as a yellow squiggly to remind me to write comments on exported methods.
</blockquote></details>

Looks like the linter actually a part of gopls, but rather a separate thing. I guess gopls only shows you compiler warnings, not lints. Which linter are you using, so I can use the same one? I see options for `golint`, `staticcheck` and `revive`.

jgraettinger (commented):
> Reviewed 3 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/116)* status: all files reviewed, 2 unresolved discussions (waiting on @psFried)

___
*[go/materialize/driver/sql2/spec_mapping.go, line 48 at r1](https://reviewable.io/reviews/estuary/flow/116#-MVIf5fYD8ih_8Jr99fp:-MVJi3Kt69O8kY2tuAAF:b81wjvk) ([raw file](https://github.com/estuary/flow/blob/16fc492301a41906b206ecef94a9c37c230220c9/go/materialize/driver/sql2/spec_mapping.go#L48)):*

Yeah I agree that this reads a little too java-like. `TokenPair` is also used for wrapping comments. I could make a different struct for the comments, though, if you think that'd be worth it. I kinda thought not, but could be convinced otherwise. I think either way I could at least rename `identifierQuotes` to just `quotes`.
</blockquote></details>

:thumbsup: works for me
___
*[go/materialize/driver/sql2/sqlgen.go, line 127 at r1](https://reviewable.io/reviews/estuary/flow/116#-MVIfg-q2svqvpCBeSuv:-MVJiBX63jVUKcypC0tt:b-2mjlfh) ([raw file](https://github.com/estuary/flow/blob/16fc492301a41906b206ecef94a9c37c230220c9/go/materialize/driver/sql2/sqlgen.go#L127)):*

Looks like the linter actually a part of gopls, but rather a separate thing. I guess gopls only shows you compiler warnings, not lints. Which linter are you using, so I can use the same one? I see options for `golint`, `staticcheck` and `revive`.
</blockquote></details>

I believe just `golint` -- default for the "Go" VSCode extension.

PR #117, opened 2021-03-08
materialize all the things 🚀 

jgraettinger:
This PR adopts a `Transactor` interface for modeling materialization drivers, and implements the interface for SQLite & PostgreSQL ; it also adds new driver implementations for Snowflake and Webhooks.


jgraettinger (commented):
I've updated this for quoting changes. Pushed as merge commit to keep the sequencing a little saner :crossed_fingers: 

> *[Reviewable](https://reviewable.io/reviews/estuary/flow/117#-:-MVJqoWz3usEsmQG9dmx:bgrrjex)* status: 0 of 41 files reviewed, all discussions resolved (waiting on @psFried)

jgraettinger (commented):
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/117)* status: all files reviewed, 2 unresolved discussions (waiting on @psFried)

___
*[Makefile, line 277 at r6](https://reviewable.io/reviews/estuary/flow/117#-MVOCnFb4aUKiXVqKy4y:-MVOJJCo7HvfATMrfMXo:b-j7dbe6) ([raw file](https://github.com/estuary/flow/blob/371a63f34d4ee63cb4d5bd15929aa7cfee30094f/Makefile#L277)):*

Yeah I thought this would always be what I'd want, but it has gotten annoying.
</blockquote></details>

We log a bunch of stuff (and that's okay). The intention behind not using -v is that you should see all of those logs only on test failure, and not otherwise. Otherwise I struggle through the noise to figure out if tests actually passed or not.
___
*[go/materialize/driver/postgres/driver.go, line 189 at r4](https://reviewable.io/reviews/estuary/flow/117#-MVNy0pNCbklLRWsgy0H:-MVOJiBs6QkRolJFxCZB:bg6cu1v) ([raw file](https://github.com/estuary/flow/blob/4cbef81b150d0249796eab559f04464c35bb1f12/go/materialize/driver/postgres/driver.go#L189)):*

Shouldn't the `commitCh` be awaited before the query is executed to return results?
</blockquote></details>

Prepare can _only_ be read because the prior Committed was written by us, and was read by the runtime's client. I didn't want to have two duplicative synchronization mechanisms, and give readers the impression that was actually necessary.

 |commitCh| is needed for cases where one is immediately reading from the store and responding with Loaded. Only then must one gate on |<-commitCh|.

I've updated the interface comments to clarify this.
___
*[go/materialize/driver/webhook/driver.go, line 54 at r6](https://reviewable.io/reviews/estuary/flow/117#-MVO7wjZ1o_7yMp73idY:-MVOLrGTAHovzo1aUVF9:b-pcx8db) ([raw file](https://github.com/estuary/flow/blob/76fc49c7680edc7398ef8b76af250f31320aee73/go/materialize/driver/webhook/driver.go#L54)):*

Is there a reason we couldn't materialize any single field? I could see a desire to materialize a subdocument, or for binary/string data that's attached to documents. Definitely not anything that has to come in this PR, but just curious if you ran into anything that would prevent us from supporting that type of thing.
</blockquote></details>

No reason at all, I was going for the simplest possible interesting implementation.

It would require more active Go involvement in framing / marshalling the response body, rather than literally copying over document bytes, but :shrug: 
___
*[go/materialize/driver/webhook/driver.go, line 125 at r6](https://reviewable.io/reviews/estuary/flow/117#-MVO9Ks093yHvTiky0K8:-MVOM7vEArjVFNqfIoyT:bd0aocu) ([raw file](https://github.com/estuary/flow/blob/76fc49c7680edc7398ef8b76af250f31320aee73/go/materialize/driver/webhook/driver.go#L125)):*

Is there a specific requirement that's driving this array body? I'm still working on forming an opinion on what it means to materialize to a generic "webhook", but the thing that feels like the most common desire would be for 1 document per request (despite the performance implications). Maybe it's just easiest to talk on a VC about the general vision for the webhook materialization, though.
</blockquote></details>

I don't think we know what's required, honestly. Trying to solve this problem generally may be out of scope, and instead we may ask that customers stand up lambdas that receive these calls and scatter out per-document requests.
___
*[go/materialize/driver/webhook/driver.go, line 157 at r6](https://reviewable.io/reviews/estuary/flow/117#-MVOCP1m0fHWyVSv06dO:-MVOMc3wC1KpGINCmVcK:bb6un1j) ([raw file](https://github.com/estuary/flow/blob/76fc49c7680edc7398ef8b76af250f31320aee73/go/materialize/driver/webhook/driver.go#L157)):*

nit: might want to accept a `201` here as well.
</blockquote></details>

Done
___
*[go/protocols/flow/materialize_extensions.go, line 28 at r3](https://reviewable.io/reviews/estuary/flow/117#-MVNu5og0MgDCOLABUeb:-MVOMhRg5vhzxTn66Plj:bbb3nnb) ([raw file](https://github.com/estuary/flow/blob/371a63f34d4ee63cb4d5bd15929aa7cfee30094f/go/protocols/flow/materialize_extensions.go#L28)):*

~~Have these validations been moved elsewhere, or is there some reason why they're no longer needed?~~
Ok I think what's going on here is that there's not actually any reason why you have to materialize the key a webhook materialization. That sound right? Just want to make sure we're still validating that the key and document are included for SQL materializations.
</blockquote></details>

Yea. This change is essentially saying that "you must include all keys" is a choice of the materialization driver, and not a requirement imposed by the flow runtime. Flow has no opinion on what gets included in a materialization. It could be just keys, or just some values, or the document, or literally nothing for all flow cares.

PR #119, opened 2021-03-10
endpoint resource paths

jgraettinger:
This PR implements the spec change discussed on #gazette slack, whereby materialization names and capture names are removed, and endpoint names now serve as the namespace of materializations / captures to which they pertain.

It adds a "resource path" mechanism to endpoints, where Validate RPCs return a driver-specific resource path which fully qualify and identify the materialization shards created in service of the materialization. Implementations are added for each of our drivers.

As a bonus, I included the side-task of having SQL drivers send an explicit, zero-valued Flow checkpoint such that DROP TABLES causes a full rebuild of the materialization (rather than picking up where the internal checkpoint left off).

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/119)


psFried (approved):
> Reviewed 39 of 39 files at r1, 25 of 25 files at r2, 6 of 6 files at r3, 6 of 6 files at r4.
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/119)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jgraettinger)

___
*[crates/models/src/build/mod.rs, line 206 at r1](https://reviewable.io/reviews/estuary/flow/119#-MVRXhitCoXdLjGR6rii:-MVRXhitCoXdLjGR6rij:bp0dxpw) ([raw file](https://github.com/estuary/flow/blob/d5d79e9c9e0ce22f065e06cca1dc99461a58dfc6/crates/models/src/build/mod.rs#L206)):*
> ```XML
> }
> 
> pub fn materialization_name(endpoint_name: &str, endpoint_resource_path: &[String]) -> String {
> ```

The type `&[String]` is a little weird for a public function, since it seems to be that way only to align with the implementation details of a caller (the fact that the caller has owned strings). I think it's generally best to accept `&[impl AsRef<str>]` for a public function, as it allows passing most types of strings.
___
*[crates/models/src/build/mod.rs, line 215 at r1](https://reviewable.io/reviews/estuary/flow/119#-MVRhGmA0cnvoseCKtrA:-MVRhGmA0cnvoseCKtrB:blzy6o9) ([raw file](https://github.com/estuary/flow/blob/d5d79e9c9e0ce22f065e06cca1dc99461a58dfc6/crates/models/src/build/mod.rs#L215)):*
> ```XML
>     name.extend(parts.join("/").chars().map(|c| {
>         match c {
>             // Note that '%' is not included (it must be escaped).
> ```

nit: this match is unnecessary if you create an `AsciiSet` that accounts for these characters. It can even be a `const`.
```
const escape_set: &percent_encoding::AsciiSet = &percent_encoding::NON_ALPHANUMERIC
        .remove(b'-').remove(b'_').remove(b'+').remove(b'/').remove(b'.').remove(b'=');

std::iter::once(endpoint).chain(path.into_iter().map(AsRef::as_ref)).map(|s| {
    utf8_percent_encode(s, escape_set)
}).join("/")
```

This also avoids a lot of allocations, especially for strings that don't actually need escaped.
___
*[go/protocols/materialize/materialize.proto, line 202 at r4](https://reviewable.io/reviews/estuary/flow/119#-MVRobOpBVyFB15U3O__:-MVRobOpBVyFB15U3O_a:b-52udzn) ([raw file](https://github.com/estuary/flow/blob/d6813da145f080cb4e1af8b20b06b2b86cb07e0e/go/protocols/materialize/materialize.proto#L202)):*
> ```protobuf
>     // Flow checkpoint which was previously committed with this |shard_fqn|.
>     // May be nil if the Driver is not stateful, in which case the Flow runtime
>     // will use its most-recent internal checkpoint. Note this internal checkpoint
> ```

Great comment :+1:

jgraettinger (commented):
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/119)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jgraettinger)

___
*[crates/models/src/build/mod.rs, line 206 at r1](https://reviewable.io/reviews/estuary/flow/119#-MVRXhitCoXdLjGR6rii:-MVRvPqp1nGIGI8uesbW:bb6un1j) ([raw file](https://github.com/estuary/flow/blob/d5d79e9c9e0ce22f065e06cca1dc99461a58dfc6/crates/models/src/build/mod.rs#L206)):*

The type `&[String]` is a little weird for a public function, since it seems to be that way only to align with the implementation details of a caller (the fact that the caller has owned strings). I think it's generally best to accept `&[impl AsRef<str>]` for a public function, as it allows passing most types of strings.
</blockquote></details>

Done
___
*[crates/models/src/build/mod.rs, line 215 at r1](https://reviewable.io/reviews/estuary/flow/119#-MVRhGmA0cnvoseCKtrA:-MVRw4vW5dZ_5nzO9aJO:b-aqfq9j) ([raw file](https://github.com/estuary/flow/blob/d5d79e9c9e0ce22f065e06cca1dc99461a58dfc6/crates/models/src/build/mod.rs#L215)):*

nit: this match is unnecessary if you create an `AsciiSet` that accounts for these characters. It can even be a `const`.
```
const escape_set: &percent_encoding::AsciiSet = &percent_encoding::NON_ALPHANUMERIC
        .remove(b'-').remove(b'_').remove(b'+').remove(b'/').remove(b'.').remove(b'=');

std::iter::once(endpoint).chain(path.into_iter().map(AsRef::as_ref)).map(|s| {
    utf8_percent_encode(s, escape_set)
}).join("/")
```

This also avoids a lot of allocations, especially for strings that don't actually need escaped.
</blockquote></details>

Ah, but this isn't just ascii! `is_alphanumeric` is true for _Unicode_ letters and numbers, and we want to pass these through as UTF8 without percent encoding. Updated test to highlight this.

I also switched to a for loop construction which avoids allocations, but it doesn't much matter in this context (called very few times).

PR #122, opened 2021-03-15
Flowctl logs

psFried:
Couple of small commits to make the CLI output easier on the eyes, especially for the `test` subcommand. There's definitely more work that _could_ be done here. It would be nice to have some scenario tests for build and test failures, for example. But I think this is probably a good place to leave it for now. Screenshots of example outputs below, in case you have any feedback on those.

image

image

I also added a `Build Success` message at the end of `flowctl build` just to give it something a little more clear than whatever npm spits out.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/122)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,6 +4,7 @@ go 1.14
 
 require (
 	github.com/bradleyjkemp/cupaloy v2.3.0+incompatible
+	github.com/fatih/color v1.7.0
 	github.com/go-openapi/jsonpointer v0.19.3
 	github.com/gogo/protobuf v1.3.1
 	github.com/golang/protobuf v1.3.5
--- a/go.sum
+++ b/go.sum
@@ -88,6 +88,7 @@ github.com/envoyproxy/go-control-plane v0.9.1-0.20191026205805-5f8ba28d4473/go.m
 github.com/envoyproxy/go-control-plane v0.9.4/go.mod h1:6rpuAdCZL397s3pYoYcLgu1mIlRU8Am5FuJP05cCM98=
 github.com/envoyproxy/protoc-gen-validate v0.1.0/go.mod h1:iSmxcyjqTsJpI2R4NaDN7+kN2VEUnK/pcBlmesArF7c=
 github.com/evanphx/json-patch v4.2.0+incompatible/go.mod h1:50XU6AFN0ol/bzJsmQLiYLvXMP4fmwYFNcr97nuDLSk=
+github.com/fatih/color v1.7.0 h1:DkWD4oS2D8LGGgTQ6IvwJJXSL5Vp2ffcQg58nFV38Ys=
 github.com/fatih/color v1.7.0/go.mod h1:Zm6kSWBoL9eyXnKyktHP6abPY2pDugNf5KwzbycvMj4=
 github.com/form3tech-oss/jwt-go v3.2.2+incompatible h1:TcekIExNqud5crz4xD2pavyTgWiPvpYe4Xau31I0PRk=
 github.com/form3tech-oss/jwt-go v3.2.2+incompatible/go.mod h1:pbq4aXjuKjdthFRnoDwaVPLA+WlJuPGy+QneDUgJi2k=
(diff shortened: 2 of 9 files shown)


jgraettinger (changes requested):
LGTM % some nits

> Reviewed 8 of 9 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/122#-:-MVqY8ydD92NW5yOJ5AF:b8bxqtg)* status: all files reviewed, 3 unresolved discussions (waiting on @psFried)

___
*[go/flowctl/cmd-test.go, line 47 at r2](https://reviewable.io/reviews/estuary/flow/122#-MVqWiHw-r_u-TrEa4Ve:-MVqWiHw-r_u-TrEa4Vf:b-di9oua) ([raw file](https://github.com/estuary/flow/blob/54dcc755d9a18a04cb99e285daab23e0ffe072d6/go/flowctl/cmd-test.go#L47)):*
> ```Go
> 				nFailed = 127
> 			}
> 			os.Exit(nFailed)
> ```

nit: it's _probably_ fine, but this doesn't allow upstack defer's to run their shutdown behavior.

Like build errors, I think we're better off returning the error, and fixing `mainboilerplate` to nicely print but not panic on a returned error
___
*[go/flowctl/logging.go, line 9 at r2](https://reviewable.io/reviews/estuary/flow/122#-MVqXlhfEGlC8_HdIA9S:-MVqXlhfEGlC8_HdIA9T:b-5thcz7) ([raw file](https://github.com/estuary/flow/blob/54dcc755d9a18a04cb99e285daab23e0ffe072d6/go/flowctl/logging.go#L9)):*
> ```Go
> // LogConfig configures handling of application log events. This was copied from mainboilerplate so
> // that we could modify it to have a different default and accept a broader range of level options.
> type LogConfig struct {
> ```

Not for now, but I don't see any reason not to upstream this mbp.LogConfig right ?
___
*[go/flowctl/main.go, line 11 at r2](https://reviewable.io/reviews/estuary/flow/122#-MVqX_h3-VL1IYFWPVx7:-MVqX_h3-VL1IYFWPVx8:b-fq8fa0) ([raw file](https://github.com/estuary/flow/blob/54dcc755d9a18a04cb99e285daab23e0ffe072d6/go/flowctl/main.go#L11)):*
> ```Go
> 
> // GlobalArgs is the set of CLI arguments that apply to all subcommands
> type GlobalArgs struct {
> ```

nit: what motivated this? I don't see `GlobalArgs` referenced anywhere.

psFried (commented):
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/122)* status: all files reviewed, 1 unresolved discussion (waiting on @jgraettinger)

___
*[go/flowctl/cmd-test.go, line 47 at r2](https://reviewable.io/reviews/estuary/flow/122#-MVqWiHw-r_u-TrEa4Ve:-MVqcG0AC8AxcwN4vzCL:b-9qy249) ([raw file](https://github.com/estuary/flow/blob/54dcc755d9a18a04cb99e285daab23e0ffe072d6/go/flowctl/cmd-test.go#L47)):*

nit: it's _probably_ fine, but this doesn't allow upstack defer's to run their shutdown behavior.

Like build errors, I think we're better off returning the error, and fixing `mainboilerplate` to nicely print but not panic on a returned error
</blockquote></details>

Yeah, I agree. I wasn't sure about whether that would be appropriate for all usages of `mbp`, though, since there's a lot of different things using it. Do you think it makes sense to try to make the behavior configurable, or is it safe to change the default to exit non-zero instead of panic on an error?
___
*[go/flowctl/logging.go, line 9 at r2](https://reviewable.io/reviews/estuary/flow/122#-MVqXlhfEGlC8_HdIA9S:-MVqd3RyB1YqptZ8FeFJ:b67bjf8) ([raw file](https://github.com/estuary/flow/blob/54dcc755d9a18a04cb99e285daab23e0ffe072d6/go/flowctl/logging.go#L9)):*

Not for now, but I don't see any reason not to upstream this mbp.LogConfig right ?
</blockquote></details>

The main consideration would be whether we want to change the default value for all usages of `mbp`. In general, I'm on the fence about whether to refactor `mainboilerplate` to make it more useful to flowctl vs just re-implementing the bits we need within flowctl. There seems to be a lot of overlap, but the general needs of a CLI seem like they might diverge a bit from what we'd want for long-running server processes. Also, I've been hesitant to refactor mbp since I'm less familiar with all the usages of it, and since it's part of the public gazette api. TLDR: I don't know. What do you think?
___
*[go/flowctl/main.go, line 11 at r2](https://reviewable.io/reviews/estuary/flow/122#-MVqX_h3-VL1IYFWPVx7:-MVqeUq20iGjyUbYty1D:bared3f) ([raw file](https://github.com/estuary/flow/blob/54dcc755d9a18a04cb99e285daab23e0ffe072d6/go/flowctl/main.go#L11)):*

nit: what motivated this? I don't see `GlobalArgs` referenced anywhere.
</blockquote></details>

Whoops! this was meant to be squashed out. It was part of an earlier factoring that used a struct to wrap stdout. I thought that wrapper struct would be a good place for configuration like whether or not to use colored output, but it turns out that the `color` package just uses global variables for that anyway, so I figured the abstraction wasn't pulling its weight and ripped it out. I'll revert this.

jgraettinger (approved):
> Reviewed 2 of 2 files at r3, 1 of 1 files at r4.
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/122)* status: all files reviewed, 1 unresolved discussion (waiting on @psFried)

___
*[go/flowctl/cmd-test.go, line 47 at r2](https://reviewable.io/reviews/estuary/flow/122#-MVqWiHw-r_u-TrEa4Ve:-MVr0foq2xkAccVfwZxB:b33qsh9) ([raw file](https://github.com/estuary/flow/blob/54dcc755d9a18a04cb99e285daab23e0ffe072d6/go/flowctl/cmd-test.go#L47)):*

Yeah, I agree. I wasn't sure about whether that would be appropriate for all usages of `mbp`, though, since there's a lot of different things using it. Do you think it makes sense to try to make the behavior configurable, or is it safe to change the default to exit non-zero instead of panic on an error?
</blockquote></details>

I think we change the default everywhere. It's panic-ing now for no particularly good reason -- the backtrace has never been useful.
___
*[go/flowctl/logging.go, line 9 at r2](https://reviewable.io/reviews/estuary/flow/122#-MVqXlhfEGlC8_HdIA9S:-MVr19SU3GIjINeBsNBN:bye7bb) ([raw file](https://github.com/estuary/flow/blob/54dcc755d9a18a04cb99e285daab23e0ffe072d6/go/flowctl/logging.go#L9)):*

The main consideration would be whether we want to change the default value for all usages of `mbp`. In general, I'm on the fence about whether to refactor `mainboilerplate` to make it more useful to flowctl vs just re-implementing the bits we need within flowctl. There seems to be a lot of overlap, but the general needs of a CLI seem like they might diverge a bit from what we'd want for long-running server processes. Also, I've been hesitant to refactor mbp since I'm less familiar with all the usages of it, and since it's part of the public gazette api. TLDR: I don't know. What do you think?
</blockquote></details>

Yea I get the hesitancy. IMO it's worth doing simply because I happen to know there aren't better-thought-through reasons why gazette `mbp` is the way it is. Gazette's still pre 1.0 major by intention :) But again, this is fine for now

psFried (commented):
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/122)* status: all files reviewed, 1 unresolved discussion (waiting on @jgraettinger)

___
*[go/flowctl/logging.go, line 9 at r2](https://reviewable.io/reviews/estuary/flow/122#-MVqXlhfEGlC8_HdIA9S:-MVr4W8q1CcrglZKYxpn:b-5jkiou) ([raw file](https://github.com/estuary/flow/blob/54dcc755d9a18a04cb99e285daab23e0ffe072d6/go/flowctl/logging.go#L9)):*

Yea I get the hesitancy. IMO it's worth doing simply because I happen to know there aren't better-thought-through reasons why gazette `mbp` is the way it is. Gazette's still pre 1.0 major by intention :) But again, this is fine for now
</blockquote></details>

Ok great. I'll plan on upstreaming this stuff into gazette sometime soon. I'll update the comments for now.

PR #125, opened 2021-03-18
derive trampoline (aka remove tokio) & catalog keyspace

jgraettinger:
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/125)


psFried (approved):
> Reviewed 4 of 4 files at r1, 10 of 10 files at r2, 13 of 13 files at r3, 10 of 10 files at r4, 35 of 35 files at r5, 24 of 24 files at r6, 14 of 14 files at r7, 9 of 9 files at r8, 1 of 1 files at r9, 6 of 6 files at r10, 13 of 13 files at r11, 1 of 1 files at r12, 9 of 9 files at r13, 6 of 6 files at r14, 4 of 4 files at r15.
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/125)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jgraettinger)

___
*[go/bindings/combine.go, line 65 at r14](https://reviewable.io/reviews/estuary/flow/125#-MW9jPjX-QnEourbf8En:-MW9jPjX-QnEourbf8Eo:b-13o4r) ([raw file](https://github.com/estuary/flow/blob/896f5834588c7f84c19e31a34aa328b38eaa596a/go/bindings/combine.go#L65)):*
> ```Go
> 
> 	// Destroy the held service on collection.
> 	runtime.SetFinalizer(combine, func(c *Combine) {
> ```

Just a thought, no action for now: This pattern of relying on the GC to release resources can be problematic because it conflates memory management with management of other types of resources. I seems like it probably a decent compromise for now, but my past experience with this has taught me to always try to put some monitoring around these. Maybe just a prometheus gauge on the number of Combine instances? A reference counter could also be nice, but I don't know how easy those are in Go since it copies things automatically.
___
*[go/flowctl/cmd-build.go, line 72 at r13](https://reviewable.io/reviews/estuary/flow/125#-MW9hNro3Sw1rhCWWKb_:-MW9hNro3Sw1rhCWWKba:b-2e9pef) ([raw file](https://github.com/estuary/flow/blob/bbe2766f0e94b3ff97bd27de035f5219c4a58414/go/flowctl/cmd-build.go#L72)):*
> ```Go
> }
> 
> func applyDerivationShardsTODO(built *bindings.BuiltCatalog, shards pc.ShardClient) error {
> ```

I'm not really clear on what these TODOs are communicating in these names. Is this just because the functions aren't actually called yet? Is this a "Goism" or a "Johnnyism"? Seems like it's just a convention that I'm not familiar with.
___
*[go/runtime/materialize.go, line 313 at r11](https://reviewable.io/reviews/estuary/flow/125#-MW9gcZY5eLwQ59BE9Ss:-MW9gcZY5eLwQ59BE9St:bbag8pq) ([raw file](https://github.com/estuary/flow/blob/dc9570fb903a442d4c6cbf0e8d46bf8a096d77b8/go/runtime/materialize.go#L313)):*
> ```Go
> 
> // BeginTxn implements Application.BeginTxn and is a no-op.
> func (m *Materialize) BeginTxn(shard consumer.Shard) error {
> ```

Is it worth doing a sanity check here that the combiner has been drained and is ready for reuse?

jgraettinger (commented):
Thanks

> *[Reviewable](https://reviewable.io/reviews/estuary/flow/125#-:-MWAPl-r4IwsPRfNPM9L:b-zcjvte)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jgraettinger)

___
*[go/bindings/combine.go, line 65 at r14](https://reviewable.io/reviews/estuary/flow/125#-MW9jPjX-QnEourbf8En:-MWAK-iP4uwVYYlGmWIR:b-wet1je) ([raw file](https://github.com/estuary/flow/blob/896f5834588c7f84c19e31a34aa328b38eaa596a/go/bindings/combine.go#L65)):*

Just a thought, no action for now: This pattern of relying on the GC to release resources can be problematic because it conflates memory management with management of other types of resources. I seems like it probably a decent compromise for now, but my past experience with this has taught me to always try to put some monitoring around these. Maybe just a prometheus gauge on the number of Combine instances? A reference counter could also be nice, but I don't know how easy those are in Go since it copies things automatically.
</blockquote></details>

+1 agreed that a gauge would be great here.

Go has this weird in-between land where you can _kind of_ get RAII through judicious use of defer, but you can't _require_ that callers be good about it. You may have seen that Derive, for example, has an exported Destroy that's idempotent (and _also_ wired up as a finalizer). It's called explicitly as part of tearing down a derivation shard. At the moment Derive has it, and Combine doesn't, because Combine is a memory-only resource while Derive isn't.

tl;dr agreed this is a weak point and we'll have to explore the right idioms as we go.
___
*[go/flowctl/cmd-build.go, line 72 at r13](https://reviewable.io/reviews/estuary/flow/125#-MW9hNro3Sw1rhCWWKb_:-MWALDc9Bft4hz6Hn90P:b4r5mha) ([raw file](https://github.com/estuary/flow/blob/bbe2766f0e94b3ff97bd27de035f5219c4a58414/go/flowctl/cmd-build.go#L72)):*

I'm not really clear on what these TODOs are communicating in these names. Is this just because the functions aren't actually called yet? Is this a "Goism" or a "Johnnyism"? Seems like it's just a convention that I'm not familiar with.
</blockquote></details>

hah it's just expressing that this is a work in progress / not expected to be its final form.
___
*[go/runtime/materialize.go, line 313 at r11](https://reviewable.io/reviews/estuary/flow/125#-MW9gcZY5eLwQ59BE9Ss:-MWAMS2nADPBH0FTnt8A:baa4z88) ([raw file](https://github.com/estuary/flow/blob/dc9570fb903a442d4c6cbf0e8d46bf8a096d77b8/go/runtime/materialize.go#L313)):*

Is it worth doing a sanity check here that the combiner has been drained and is ready for reuse?
</blockquote></details>

I'm not sure what a check here would be guarding against, honestly. The combiner is in a well defined state, and if that state isn't what we think it is here, we'll see that as a future poll() error.

PR #129, opened 2021-03-21
Temp sensor example

psFried:
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/129)


jgraettinger (approved):
> Reviewed 12 of 12 files at r1.
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/129)* status: all files reviewed, 1 unresolved discussion (waiting on @psFried)

___
*[examples/temp-sensors/flow.yaml, line 8 at r1](https://reviewable.io/reviews/estuary/flow/129#-MWI58j5CLuiV5hzk2xi:-MWI58j5CLuiV5hzk2xj:b-6zh666) ([raw file](https://github.com/estuary/flow/blob/5f7e6c4a978dfab79b675d552ebf944a6aff480f/examples/temp-sensors/flow.yaml#L8)):*
> ```YAML
> 
>   # Stream of readings that come from temperature sensors.
>   temperature/readings:
> ```

nit: other examples have a `examples/` naming prefix. Is that easy to change ? if not, don't worry about it.

psFried (commented):
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/129)* status: all files reviewed, 1 unresolved discussion (waiting on @jgraettinger)

___
*[examples/temp-sensors/flow.yaml, line 8 at r1](https://reviewable.io/reviews/estuary/flow/129#-MWI58j5CLuiV5hzk2xi:-MWKOjqOBTQpy_h1Lts3:b-gepsve) ([raw file](https://github.com/estuary/flow/blob/5f7e6c4a978dfab79b675d552ebf944a6aff480f/examples/temp-sensors/flow.yaml#L8)):*

nit: other examples have a `examples/` naming prefix. Is that easy to change ? if not, don't worry about it.
</blockquote></details>

It is easy to change. The tradeoff is that the collection names were getting a bit long when I reference them inline in a paragraph in documentation. Do you think it's ok to name the collection e.g. `examples/temperature/sensors` and then just reference it in docs paragraphs as just `sensors`, or would that be too confusing? Whatever we do, I'd like to be consistent in both how we name collections and how we reference them in the docs. I think I'll give that idea a try for now, and just create a separate PR if we want to do something else.

psFried (commented):
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/129)* status: all files reviewed, 1 unresolved discussion (waiting on @jgraettinger)

___
*[examples/temp-sensors/flow.yaml, line 8 at r1](https://reviewable.io/reviews/estuary/flow/129#-MWI58j5CLuiV5hzk2xi:-MWKQVMY6zz3jtHoM3e3:b7m13j7) ([raw file](https://github.com/estuary/flow/blob/5f7e6c4a978dfab79b675d552ebf944a6aff480f/examples/temp-sensors/flow.yaml#L8)):*

It is easy to change. The tradeoff is that the collection names were getting a bit long when I reference them inline in a paragraph in documentation. Do you think it's ok to name the collection e.g. `examples/temperature/sensors` and then just reference it in docs paragraphs as just `sensors`, or would that be too confusing? Whatever we do, I'd like to be consistent in both how we name collections and how we reference them in the docs. I think I'll give that idea a try for now, and just create a separate PR if we want to do something else.
</blockquote></details>

Ok, I think I actually want to walk that back. The additional `Examples` prefix also adds length to all the code samples, and it's looking to me like a negative impact on the readability of the tutorials. Also, I'm starting to think that the `examples/` prefix is perhaps redundant anyway, since the context is obvious and there's not going to be a case where you'd `import` any of the examples into another catalog that isn't also already an example. So maybe we just start leaving off the `examples/` prefix across the board?

jgraettinger:
Hmm the context is wanting to establish habit early on for thinking of
collections as globally namespaced. We could also use something like “acme”
as a fictitious owner, or do you have other ideas? Or we just don’t worry
about it here. I’m fine with whatever you want to go with.

psFried (commented):
Yeah good point about wanting to reinforce thinking about them being globally namespaced. I think for right now, I'll proceed with the current names. Both in the interest of expediency and so we can step back and look at the docs and examples as a whole. We may also be able to shorten `temperature` to `temp`, but I had been avoiding that due to possible confusion with `temporary`.

> *[Reviewable](https://reviewable.io/reviews/estuary/flow/129#-:-MWKWErJ1EECv8cy0RHA:b4vja1g)* status: all files reviewed, 1 unresolved discussion (waiting on @jgraettinger)

PR #134, opened 2021-03-25
flowctl-apply: various tweaks and fixes from kubernetes testing

jgraettinger:
* 🐛 It's '"NpmPackage"', not "NpmPackage" (it's a JSON-encoded enum).
  `flowctly apply` now panics if an NPM package fails to be built.

* We can't rely on Etcd being sync-able in `flowctl apply`, as it's
  often behind a `kubectl port-forward`.

* Code format acmeBank.flow.ts

* Add commented-out store URL for kubernetes test clusters.

* Update broker root.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/134)

### DIFF
--- a/crates/build/src/nodejs/mod.rs
+++ b/crates/build/src/nodejs/mod.rs
@@ -212,11 +212,11 @@ pub fn compile_package(package_dir: &path::Path) -> Result<(), anyhow::Error> {
         npm_cmd(package_dir, &["install", "--no-audit", "--no-fund"])?;
     }
     npm_cmd(package_dir, &["run", "compile"])?;
+    npm_cmd(package_dir, &["run", "lint"])?;
     Ok(())
 }
 
 pub fn pack_package(package_dir: &path::Path) -> Result<tables::Resources, anyhow::Error> {
-    npm_cmd(package_dir, &["run", "lint"])?;
     npm_cmd(package_dir, &["pack"])?;
 
     let pack = package_dir.join("catalog-js-transformer-0.0.0.tgz");
--- a/examples/acmeBank.flow.ts
+++ b/examples/acmeBank.flow.ts
@@ -9,8 +9,8 @@ export class AcmeBankBalances implements interfaces.AcmeBankBalances {
     ): collections.AcmeBankBalances[] {
         return [
             // A transfer removes from the sender and adds to the receiver.
-            {account: source.from, amount: -source.amount},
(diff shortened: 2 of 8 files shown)


psFried (approved):
Just one question here. I don't know if that's blocking or not.

> Reviewed 5 of 5 files at r1.
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/134#-:-MWeZMS_CcWNwYA8J3vS:bm308ya)* status: all files reviewed, 1 unresolved discussion (waiting on @jgraettinger)

___
*[go/flowctl/cmd-apply.go, line 80 at r1](https://reviewable.io/reviews/estuary/flow/134#-MWeXoBxCV5zodaEF2Mc:-MWeXoBxCV5zodaEF2Md:b77zirw) ([raw file](https://github.com/estuary/flow/blob/a8d471e977317bd5b884f7127936eccde5574ba4/go/flowctl/cmd-apply.go#L80)):*
> ```Go
> 	if err != nil {
> 		return err
> 	} else if len(built.NPMPackage) == 0 {
> ```

What happens if the catalog doesn't include any derivations? Seems like we might panic here in that case. I think I might be missing a bit of context, though, because it seems like we ought to have returned an error if the Typescript package is missing due to some error in the build process (as opposed to this just being a catalog without any derivations).

jgraettinger (commented):
Thanks

> *[Reviewable](https://reviewable.io/reviews/estuary/flow/134#-:-MWtv0G81cdBrG21n4Yr:b-zcjvte)* status: 4 of 8 files reviewed, 1 unresolved discussion (waiting on @psFried)

___
*[go/flowctl/cmd-apply.go, line 80 at r1](https://reviewable.io/reviews/estuary/flow/134#-MWeXoBxCV5zodaEF2Mc:-MWtulng4KrcPdmUojUj:b-nh4hqw) ([raw file](https://github.com/estuary/flow/blob/a8d471e977317bd5b884f7127936eccde5574ba4/go/flowctl/cmd-apply.go#L80)):*

What happens if the catalog doesn't include any derivations? Seems like we might panic here in that case. I think I might be missing a bit of context, though, because it seems like we ought to have returned an error if the Typescript package is missing due to some error in the build process (as opposed to this just being a catalog without any derivations).
</blockquote></details>

It works fine, because we (currently) _always_ build an NPM package. It may just have no routes if given a trivial catalog input. Tested to confirm.

PR #135, opened 2021-03-25
Add basic prometheus metrics and trace lambdas

psFried:
Gather prometheus metrics for Combine and Derive.
This is a minimal starting point for Flow instrumentation to support
some Grafana dashboards. Instrumentation of rocksdb sizes isn't included
here because the intent is to add that to Gazette.

These may still need changed, so I don't want to merge just yet, but it's probably good to get feedback.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/135)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -19,6 +19,7 @@ require (
 	github.com/mattn/go-sqlite3 v2.0.3+incompatible
 	github.com/nsf/jsondiff v0.0.0-20210303162244-6ea32392771e
 	github.com/pkg/errors v0.9.1
+	github.com/prometheus/client_golang v1.5.1 // indirect
 	github.com/sirupsen/logrus v1.7.0
 	github.com/snowflakedb/gosnowflake v1.4.2-0.20210318070613-b0c023e3afd7
 	github.com/stretchr/testify v1.5.1
--- a/go/bindings/combine.go
+++ b/go/bindings/combine.go
@@ -8,6 +8,8 @@ import (
 	"runtime"
 
 	pf "github.com/estuary/flow/go/protocols/flow"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
 )
 
 // CombineCallback is the callback accepted by Combine.Finish and Derive.Finish.
@@ -28,10 +30,13 @@ type Combine struct {
 	svc         *service
 	drained     []C.Out
 	pinnedIndex *SchemaIndex // Used from Rust.
+	stats       combineStats
+	metrics     combineMetrics
 }
 
 // NewCombiner builds and returns a new Combine.
 func NewCombine(
+	shardFqn string,
 	index *SchemaIndex,
 	schemaURI string,
 	keyPtrs []string,
@@ -59,11 +64,15 @@ func NewCombine(
 		svc:         svc,
 		drained:     nil,
 		pinnedIndex: index,
+		stats:       combineStats{},
+		metrics:     newCombineMetrics(shardFqn),
 	}
 
+	combineNewInstanceCounter.WithLabelValues(shardFqn).Inc()
 	// Destroy the held service on collection.
 	runtime.SetFinalizer(combine, func(c *Combine) {
 		c.svc.destroy()
+		combineDestroyInstanceCounter.WithLabelValues(shardFqn).Inc()
 	})
 
 	return combine, nil
@@ -73,6 +82,8 @@ func NewCombine(
 func (c *Combine) ReduceLeft(doc json.RawMessage) error {
 	c.drained = nil // Invalidate.
 	c.svc.sendBytes(uint32(pf.CombineAPI_REDUCE_LEFT), doc)
+	c.stats.leftDocs++
+	c.stats.leftBytes += len(doc)
 
 	if c.svc.queuedFrames() >= 128 {
 		return pollExpectNoOutput(c.svc)
@@ -84,6 +95,9 @@ func (c *Combine) ReduceLeft(doc json.RawMessage) error {
 func (c *Combine) CombineRight(doc json.RawMessage) error {
 	c.drained = nil // Invalidate.
 	c.svc.sendBytes(uint32(pf.CombineAPI_COMBINE_RIGHT), doc)
+
+	c.stats.rightDocs++
+	c.stats.rightBytes += len(doc)
 
 	if c.svc.queuedFrames() >= 128 {
 		return pollExpectNoOutput(c.svc)
@@ -108,38 +122,46 @@ func (c *Combine) PrepareToDrain() error {
 
 // Drain combined documents, invoking the callback for each distinct group-by document.
 // If Drain returns without error, the Combine may be used again.
-func (c *Combine) Drain(cb CombineCallback) error {
+func (c *Combine) Drain(cb CombineCallback) (err error) {
+	defer c.stats.reset()
 	if c.drained == nil {
-		if err := c.PrepareToDrain(); err != nil {
-			return err
+		if err = c.PrepareToDrain(); err != nil {
+			return
 		}
 	}
-	return drainCombineToCallback(c.svc, &c.drained, cb)
+	c.stats.drainDocs, c.stats.drainBytes, err = drainCombineToCallback(c.svc, &c.drained, cb)
+	if err == nil {
+		c.metrics.recordDrain(&c.stats)
+	}
+	return
 }
 
 func drainCombineToCallback(
 	svc *service,
 	out *[]C.Out,
 	cb CombineCallback,
-) error {
+) (nDocs, nBytes int, err error) {
 	// Sanity check we got triples of output frames.
 	if len(*out)%3 != 0 {
 		panic(fmt.Sprintf("wrong number of output frames (%d; should be %% 3)", len(*out)))
 	}
 
 	for len(*out) >= 3 {
-		if err := cb(
+		var doc = svc.arenaSlice((*out)[0])
+		nDocs++
+		nBytes += len(doc)
+		if err = cb(
 			pf.CombineAPI_Code((*out)[0].code) == pf.CombineAPI_DRAINED_REDUCED_DOCUMENT,
-			svc.arenaSlice((*out)[0]), // Doc.
+			doc,                       // Doc.
 			svc.arenaSlice((*out)[1]), // Packed key.
 			svc.arenaSlice((*out)[2]), // Packed fields.
 		); err != nil {
-			return err
+			return
 		}
 		*out = (*out)[3:]
 	}
 
-	return nil
+	return
 }
 
 func newCombineSvc() *service {
@@ -150,4 +172,96 @@ func newCombineSvc() *service {
 		func(ch *C.Channel, in C.In16) { C.combine_invoke16(ch, in) },
 		func(ch *C.Channel) { C.combine_drop(ch) },
 	)
+}
+
+var combineLeftDocsCounter = promauto.NewCounterVec(prometheus.CounterOpts{
+	Name: "flow_combine_left_docs_total",
+	Help: "Count of documents input as the left hand side of combine operations",
+}, []string{"shard"})
+var combineLeftBytesCounter = promauto.NewCounterVec(prometheus.CounterOpts{
+	Name: "flow_combine_left_bytes_total",
+	Help: "Number of bytes input as the left hand side of combine operations",
+}, []string{"shard"})
+var combineRightDocsCounter = promauto.NewCounterVec(prometheus.CounterOpts{
+	Name: "flow_combine_right_docs_total",
+	Help: "Count of documents input as the right hand side of combine operations",
+}, []string{"shard"})
+var combineRightBytesCounter = promauto.NewCounterVec(prometheus.CounterOpts{
+	Name: "flow_combine_right_bytes_total",
+	Help: "Number of bytes input as the right hand side of combine operations",
+}, []string{"shard"})
+var combineDrainDocsCounter = promauto.NewCounterVec(prometheus.CounterOpts{
+	Name: "flow_combine_drain_docs_total",
+	Help: "Count of documents drained from combiners",
+}, []string{"shard"})
+var combineDrainBytesCounter = promauto.NewCounterVec(prometheus.CounterOpts{
+	Name: "flow_combine_drain_bytes_total",
+	Help: "Number of bytes drained from combiners",
+}, []string{"shard"})
+var combineOpsCounter = promauto.NewCounterVec(prometheus.CounterOpts{
+	Name: "flow_combine_drain_ops_total",
+	Help: "Count of number of combine operations. A single operation may combine any number of documents with any number of distinct keys.",
+}, []string{"shard"})
+
+var combineNewInstanceCounter = promauto.NewCounterVec(prometheus.CounterOpts{
+	Name: "flow_combine_new_total",
+	Help: "Count of new combiner instances created",
+}, []string{"shard"})
+var combineDestroyInstanceCounter = promauto.NewCounterVec(prometheus.CounterOpts{
+	Name: "flow_combine_finalize_total",
+	Help: "Count of combiner instances that have been torn down",
+}, []string{"shard"})
+
+type combineStats struct {
+	leftDocs   int
+	leftBytes  int
+	rightDocs  int
+	rightBytes int
+	drainDocs  int
+	drainBytes int
+}
+
+func (s *combineStats) reset() {
+	*s = combineStats{}
+}
+
+type combineMetrics struct {
+	leftDocs  prometheus.Counter
+	leftBytes prometheus.Counter
+
+	rightDocs  prometheus.Counter
+	rightBytes prometheus.Counter
+
+	drainDocs  prometheus.Counter
+	drainBytes prometheus.Counter
+
+	drainCounter prometheus.Counter
+}
+
+func newCombineMetrics(shard string) combineMetrics {
+	return combineMetrics{
+		leftDocs:  combineLeftDocsCounter.WithLabelValues(shard),
+		leftBytes: combineLeftBytesCounter.WithLabelValues(shard),
+
+		rightDocs:  combineRightDocsCounter.WithLabelValues(shard),
+		rightBytes: combineRightBytesCounter.WithLabelValues(shard),
+
+		drainDocs:  combineDrainDocsCounter.WithLabelValues(shard),
+		drainBytes: combineDrainBytesCounter.WithLabelValues(shard),
+
+		drainCounter: combineOpsCounter.WithLabelValues(shard),
+	}
+}
+
+func (m *combineMetrics) recordDrain(stats *combineStats) {
+	m.leftDocs.Add(float64(stats.leftDocs))
+	m.leftBytes.Add(float64(stats.leftBytes))
+
+	m.rightDocs.Add(float64(stats.rightDocs))
+	m.rightBytes.Add(float64(stats.rightBytes))
+
+	m.drainDocs.Add(float64(stats.drainDocs))
+	m.drainBytes.Add(float64(stats.drainBytes))
+
+	m.drainCounter.Inc()
 }
--- a/go/bindings/combine_test.go
+++ b/go/bindings/combine_test.go
@@ -27,6 +27,7 @@ func TestCombineBindings(t *testing.T) {
 	require.NoError(t, err)
 
 	combiner, err := NewCombine(
+		"test/combineBindings",
 		schemaIndex,
 		collection.SchemaUri,
 		collection.KeyPtrs,
--- a/go/bindings/derive.go
+++ b/go/bindings/derive.go
@@ -19,11 +19,20 @@ import (
 	"unsafe"
 
 	pf "github.com/estuary/flow/go/protocols/flow"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
 	log "github.com/sirupsen/logrus"
 	"github.com/tecbot/gorocksdb"
 	pc "go.gazette.dev/core/consumer/protocol"
 	"go.gazette.dev/core/message"
+	"golang.org/x/net/trace"
 )
+
+var deriveLambdaDurations = promauto.NewHistogramVec(prometheus.HistogramOpts{
+	Name:    "flow_derive_lambda_duration_seconds",
+	Help:    "Duration in seconds of invocation of derive lambdas",
+	Buckets: []float64{0.0005, 0.002, 0.01, 0.05, 0.1, 0.3, 1.0},
+}, []string{"derivation", "lambdaType"})
 
 // Derive is an instance of the derivation workflow.
 type Derive struct {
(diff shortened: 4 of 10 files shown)


jgraettinger (changes requested):
> Reviewed 3 of 8 files at r1.
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/135)* status: 3 of 8 files reviewed, 11 unresolved discussions (waiting on @psFried)

___
*[go/bindings/combine.go, line 18 at r1](https://reviewable.io/reviews/estuary/flow/135#-MWu4qT1-WTX1deDs56n:-MWu4qT1-WTX1deDs56o:ba08nah) ([raw file](https://github.com/estuary/flow/blob/c1e87e601c8c26d56d1358473e4ab99bf2b33146/go/bindings/combine.go#L18)):*
> // drained. We _could_ increment all the counters at once after PrepareToDrain, but doing so doesn't
// seems worth the performance impact of iterating all the documents in the transaction twice.

I'm not groking this comment - why would we need to "iterate twice" ? what would that mean (what's being iterated a second time)?

nit: Also, move this block to the bottom ?
___
*[go/bindings/combine.go, line 38 at r1](https://reviewable.io/reviews/estuary/flow/135#-MWtwvlXE-1xzWpIe1TE:-MWtwvlXE-1xzWpIe1TF:b-w52vap) ([raw file](https://github.com/estuary/flow/blob/c1e87e601c8c26d56d1358473e4ab99bf2b33146/go/bindings/combine.go#L38)):*
> ```Go
> }, []string{"task"})
> var combineOutputDocsCounter = promauto.NewCounterVec(prometheus.CounterOpts{
> 	Name: "flow_combine_output_docs",
> ```

nit: output => drain, since that's the terminology we're using?

nit: mixing "_documents" and "_docs" as unit suffixes here. pick one.
___
*[go/bindings/combine.go, line 45 at r1](https://reviewable.io/reviews/estuary/flow/135#-MWtzDHO1JACjYTQtsXr:-MWtzDHO1JACjYTQtsXs:b-aa0flu) ([raw file](https://github.com/estuary/flow/blob/c1e87e601c8c26d56d1358473e4ab99bf2b33146/go/bindings/combine.go#L45)):*
> ```Go
> 	Help: "Number of bytes drained from combiners",
> }, []string{"task"})
> var combineOpsCounter = promauto.NewCounterVec(prometheus.CounterOpts{
> ```

"ops" is really a drain, as that's the thing that's being counted.

How about `flow_combine_drain_total`, `flow_combine_drain_documents`, `flow_combine_drain_bytes` ?
___
*[go/bindings/combine.go, line 49 at r1](https://reviewable.io/reviews/estuary/flow/135#-MWu-CUh9mxAGfEUPX_f:-MWu-CUh9mxAGfEUPX_g:b4viirv) ([raw file](https://github.com/estuary/flow/blob/c1e87e601c8c26d56d1358473e4ab99bf2b33146/go/bindings/combine.go#L49)):*
> ```Go
> 	Help: "Count of number of combine operations. A single operation may combine any number of documents with any number of distinct keys.",
> }, []string{"task"})
> var combineInstanceGauge = promauto.NewGaugeVec(prometheus.GaugeOpts{
> ```

Thinking `flow_combine_instances` should be two counters: `flow_combine_new_total` and `flow_combine_finalize_total`. That gives you the number of un-GC'd instances, plus a rate of creation.
___
*[go/bindings/combine.go, line 73 at r1](https://reviewable.io/reviews/estuary/flow/135#-MWu09ep-di6U_DKT1NV:-MWu09ep-di6U_DKT1NW:b-25v0dy) ([raw file](https://github.com/estuary/flow/blob/c1e87e601c8c26d56d1358473e4ab99bf2b33146/go/bindings/combine.go#L73)):*
> ```Go
> 	pinnedIndex *SchemaIndex // Used from Rust.
> 
> 	// relevant metrics are stored here to avoid doing a lookup on each usage
> ```

Prometheus is pretty fast but:
 - This tightly couples the Combine implementation to the prometheus client
 - Involves dynamically dispatched interface calls for every metric tick.

As an alternative, look at how consumer/transaction.go structures itself:
https://github.com/gazette/core/blob/master/consumer/transaction.go#L363-L402

Basically, these could be plain int counters which accumulate as the Combine runs, and then dump their count mass in all at once (probably in `drainCombineToCallback`) using inline `WithLabelValues`.  All but the left/right counters can even be locals just in that function. This also resolves the "what happens if it errors" skew you speak to above -- define metrics as only counting successful drains, and don't increment counters at all when returning an error.
___
*[go/bindings/combine.go, line 114 at r1](https://reviewable.io/reviews/estuary/flow/135#-MWu1Usx1Y_y0MMIUmS_:-MWu1Usx1Y_y0MMIUmSa:b6ltjf9) ([raw file](https://github.com/estuary/flow/blob/c1e87e601c8c26d56d1358473e4ab99bf2b33146/go/bindings/combine.go#L114)):*
> ```Go
> 		pinnedIndex: index,
> 
> 		leftDocsCounter:    combineLeftDocsCounter.WithLabelValues(taskName),
> ```

We should grain these on FQN (as in a shard FQN) rather than task name -- it's how consumer metrics are tracked, so would make these directly comparable. Task name can be obtained from the shard FQN but not the other way.

It won't _always_ be a shard though -- e.x. an instance of flow-ingester -- but when it is that's what we should be using.

Looks like gazette is using `shard` as the label for this: https://github.com/gazette/core/blob/master/consumer/interfaces.go#L271

I'm thinking we should stick with `shard` just for consistency, even if it's not _exactly_ right ?
___
*[go/bindings/combine.go, line 186 at r1](https://reviewable.io/reviews/estuary/flow/135#-MWu2DAE8s94I2OPxGmf:-MWu2DAE8s94I2OPxGmg:b6agk46) ([raw file](https://github.com/estuary/flow/blob/c1e87e601c8c26d56d1358473e4ab99bf2b33146/go/bindings/combine.go#L186)):*
> ```Go
> 	}
> 	var instrumented = instrumentCallback(c.outputDocsCounter, c.outputBytesCounter, cb)
> 	return drainCombineToCallback(c.svc, &c.drained, instrumented)
> ```

why does this need to be instrumented? `drainCombineToCallback` can do the metric increments
___
*[go/bindings/combine.go, line 208 at r1](https://reviewable.io/reviews/estuary/flow/135#-MWu2Zmf6VIYw3F-E9Mj:-MWu2Zmg9FRjVbEwjftl:b-880z7f) ([raw file](https://github.com/estuary/flow/blob/c1e87e601c8c26d56d1358473e4ab99bf2b33146/go/bindings/combine.go#L208)):*
> ```Go
> 
> 	for len(*out) >= 3 {
> 		if err := cb(
> ```

as per my comment above: have some locals here that accumulate docs and bytes, and send them to the counter at the end ?
___
*[go/bindings/derive.go, line 31 at r1](https://reviewable.io/reviews/estuary/flow/135#-MWu7Trc9PcXb-QtgKnE:-MWu7TrdEZLNjAeojSCr:b-oxyzlb) ([raw file](https://github.com/estuary/flow/blob/c1e87e601c8c26d56d1358473e4ab99bf2b33146/go/bindings/derive.go#L31)):*
> ```Go
> )
> 
> var deriveInputDocsCounter = promauto.NewCounterVec(prometheus.CounterOpts{
> ```

most comments from combine.go apply here too.
___
*[go/bindings/derive.go, line 56 at r1](https://reviewable.io/reviews/estuary/flow/135#-MWu7kzyDFxAq9U2-qwj:-MWu7kzyDFxAq9U2-qwk:b-wz71ln) ([raw file](https://github.com/estuary/flow/blob/c1e87e601c8c26d56d1358473e4ab99bf2b33146/go/bindings/derive.go#L56)):*
> ```Go
> }, []string{"task"})
> var deriveLambdaDurations = promauto.NewHistogramVec(prometheus.HistogramOpts{
> 	Name:    "flow_derive_lambda_time",
> ```

if you haven't seen it, check out https://prometheus.io/docs/practices/naming/#metric-names . Per its guidance, this should be units of seconds (millis is not a base unit), and have `_seconds` in the name.

--- 

I'm a little torn on this metric, because I suspect there are better ways to capture it. The alternative would be a transparent instrumentation like the [grpc metrics middleware](https://github.com/grpc-ecosystem/go-grpc-prometheus) we use. There isn't an exactly equivalent for `net/http` I can find with quick googling -- something like https://github.com/travelaudience/go-promhttp maybe comes close but I'm wary of the dependency. There's also [promhttp](https://pkg.go.dev/github.com/prometheus/client_golang/prometheus/promhttp#RoundTripperFunc) for lower level instrumentation.

I think I'm fine with it for now, but I expect it will be deprecated pretty quickly.
___
*[go/bindings/derive.go, line 222 at r1](https://reviewable.io/reviews/estuary/flow/135#-MWuCiKu7wNU72owHoLV:-MWuCiKu7wNU72owHoLW:b-d3wfn1) ([raw file](https://github.com/estuary/flow/blob/c1e87e601c8c26d56d1358473e4ab99bf2b33146/go/bindings/derive.go#L222)):*
> ```Go
> 	d.svc.sendBytes(uint32(pf.DeriveAPI_FLUSH_TRANSACTION), nil)
> 	d.txnCounter.Inc()
> 	var instrumented = instrumentCallback(d.outputDocsCounter, d.outputBytesCounter, cb)
> ```

Perhaps the reason you did this instrumentation is to distinguish "drains from a combine" vs "drains from a derive" ? I'm not sure that's necessary -- perhaps they _should_ be the same metric, since they instrument the same system and mean the same thing -- and we're able to distinguish them already on the FQN.

Or, derive and combine can have their own `drainCombineToCallback` implementations. It's only shared because, to date, they've been the same literal implementation.

psFried (commented):
Ok I'm pushing some changes now that hopefully address all the feedback. A lot of the comments were just addressed by changing the approach to increment metrics all at the end of each Drain operation, so I didn't respond to all of them individually. But LMK if you think there's any I should have another look at. I may also have additional updates here as I work on setting up the dashboards, so still considering this sort of untested and waiting to merge until I get all of it working end to end.

> *[Reviewable](https://reviewable.io/reviews/estuary/flow/135#-:-MWzPGai5-d2UmvqkbtU:bbo6k26)* status: 3 of 8 files reviewed, 10 unresolved discussions (waiting on @jgraettinger)

___
*[go/bindings/combine.go, line 18 at r1](https://reviewable.io/reviews/estuary/flow/135#-MWu4qT1-WTX1deDs56n:-MWzGpik65AutxMHz34t:be2ulgd) ([raw file](https://github.com/estuary/flow/blob/c1e87e601c8c26d56d1358473e4ab99bf2b33146/go/bindings/combine.go#L18)):*

> // drained. We _could_ increment all the counters at once after PrepareToDrain, but doing so doesn't
// seems worth the performance impact of iterating all the documents in the transaction twice.

I'm not groking this comment - why would we need to "iterate twice" ? what would that mean (what's being iterated a second time)?

nit: Also, move this block to the bottom ?
</blockquote></details>

No longer relevant now that I've switched the approach to accumulating stats and then publishing the metrics all at once, but it was the documents that would have been iterated twice (once to invoke the callback and again to update the metrics).
___
*[go/bindings/combine.go, line 45 at r1](https://reviewable.io/reviews/estuary/flow/135#-MWtzDHO1JACjYTQtsXr:-MWzHt3JEhdGTu9FA_8X:b-mjy6ba) ([raw file](https://github.com/estuary/flow/blob/c1e87e601c8c26d56d1358473e4ab99bf2b33146/go/bindings/combine.go#L45)):*

"ops" is really a drain, as that's the thing that's being counted.

How about `flow_combine_drain_total`, `flow_combine_drain_documents`, `flow_combine_drain_bytes` ?
</blockquote></details>

is `flow_combine_drain_ops` acceptable instead of `flow_combine_drain_total`? The `_total` seems a little confusing to me.
___
*[go/bindings/combine.go, line 49 at r1](https://reviewable.io/reviews/estuary/flow/135#-MWu-CUh9mxAGfEUPX_f:-MWzPCjLEBC5SqE5y56u:b-896fix) ([raw file](https://github.com/estuary/flow/blob/c1e87e601c8c26d56d1358473e4ab99bf2b33146/go/bindings/combine.go#L49)):*

Thinking `flow_combine_instances` should be two counters: `flow_combine_new_total` and `flow_combine_finalize_total`. That gives you the number of un-GC'd instances, plus a rate of creation.
</blockquote></details>

Done.
___
*[go/bindings/combine.go, line 73 at r1](https://reviewable.io/reviews/estuary/flow/135#-MWu09ep-di6U_DKT1NV:-MWzJM4xBdFDL3XmcyXG:b-5hj4jw) ([raw file](https://github.com/estuary/flow/blob/c1e87e601c8c26d56d1358473e4ab99bf2b33146/go/bindings/combine.go#L73)):*

Prometheus is pretty fast but:
 - This tightly couples the Combine implementation to the prometheus client
 - Involves dynamically dispatched interface calls for every metric tick.

As an alternative, look at how consumer/transaction.go structures itself:
https://github.com/gazette/core/blob/master/consumer/transaction.go#L363-L402

Basically, these could be plain int counters which accumulate as the Combine runs, and then dump their count mass in all at once (probably in `drainCombineToCallback`) using inline `WithLabelValues`.  All but the left/right counters can even be locals just in that function. This also resolves the "what happens if it errors" skew you speak to above -- define metrics as only counting successful drains, and don't increment counters at all when returning an error.
</blockquote></details>

Well, I think the Combine implementation _is_ coupled to prometheus in the sense that it has to update those metrics somehow. I tried ecapsulating that in a separate struct, which I think is better,  but I'm honestly not sure I understand the concern there so LMK what you think. 

I also think that the resolved metrics (results of `WithLabelValues`) are worth keeping around if it's easy to do so. Again I think the encapsulation helps here, but it's ultimately still coupled to the Combine/Derive instances, so IDK if that really addresses the concern.
___
*[go/bindings/combine.go, line 114 at r1](https://reviewable.io/reviews/estuary/flow/135#-MWu1Usx1Y_y0MMIUmS_:-MWzLeqLCw9ykXRBdhE7:b-896fix) ([raw file](https://github.com/estuary/flow/blob/c1e87e601c8c26d56d1358473e4ab99bf2b33146/go/bindings/combine.go#L114)):*

We should grain these on FQN (as in a shard FQN) rather than task name -- it's how consumer metrics are tracked, so would make these directly comparable. Task name can be obtained from the shard FQN but not the other way.

It won't _always_ be a shard though -- e.x. an instance of flow-ingester -- but when it is that's what we should be using.

Looks like gazette is using `shard` as the label for this: https://github.com/gazette/core/blob/master/consumer/interfaces.go#L271

I'm thinking we should stick with `shard` just for consistency, even if it's not _exactly_ right ?
</blockquote></details>

Done.
___
*[go/bindings/combine.go, line 186 at r1](https://reviewable.io/reviews/estuary/flow/135#-MWu2DAE8s94I2OPxGmf:-MWzLkc8A4H_1RFLUZAZ:b-896fix) ([raw file](https://github.com/estuary/flow/blob/c1e87e601c8c26d56d1358473e4ab99bf2b33146/go/bindings/combine.go#L186)):*

why does this need to be instrumented? `drainCombineToCallback` can do the metric increments
</blockquote></details>

Done.
___
*[go/bindings/combine.go, line 208 at r1](https://reviewable.io/reviews/estuary/flow/135#-MWu2Zmf6VIYw3F-E9Mj:-MWzLkNoDVlEBMVDX3ri:b-896fix) ([raw file](https://github.com/estuary/flow/blob/c1e87e601c8c26d56d1358473e4ab99bf2b33146/go/bindings/combine.go#L208)):*

as per my comment above: have some locals here that accumulate docs and bytes, and send them to the counter at the end ?
</blockquote></details>

Done.
___
*[go/bindings/derive.go, line 31 at r1](https://reviewable.io/reviews/estuary/flow/135#-MWu7Trc9PcXb-QtgKnE:-MWzLma16kI_mtpkKRM3:b-896fix) ([raw file](https://github.com/estuary/flow/blob/c1e87e601c8c26d56d1358473e4ab99bf2b33146/go/bindings/derive.go#L31)):*

most comments from combine.go apply here too.
</blockquote></details>

Done.
___
*[go/bindings/derive.go, line 56 at r1](https://reviewable.io/reviews/estuary/flow/135#-MWu7kzyDFxAq9U2-qwj:-MWzNPYL7ddrOqvZQW3f:boo6ro7) ([raw file](https://github.com/estuary/flow/blob/c1e87e601c8c26d56d1358473e4ab99bf2b33146/go/bindings/derive.go#L56)):*

if you haven't seen it, check out https://prometheus.io/docs/practices/naming/#metric-names . Per its guidance, this should be units of seconds (millis is not a base unit), and have `_seconds` in the name.

--- 

I'm a little torn on this metric, because I suspect there are better ways to capture it. The alternative would be a transparent instrumentation like the [grpc metrics middleware](https://github.com/grpc-ecosystem/go-grpc-prometheus) we use. There isn't an exactly equivalent for `net/http` I can find with quick googling -- something like https://github.com/travelaudience/go-promhttp maybe comes close but I'm wary of the dependency. There's also [promhttp](https://pkg.go.dev/github.com/prometheus/client_golang/prometheus/promhttp#RoundTripperFunc) for lower level instrumentation.

I think I'm fine with it for now, but I expect it will be deprecated pretty quickly.
</blockquote></details>

Thanks for pointing out the time unit, that's fixed. I also didn't find anything that looked great as an automatic instrumentation, especially given that we use separate clients for different types of lambdas. Personally, I think we'll likely want to re-evaluate lambda instrumentation as we evolve to support different lambda types, so I'm thinking it's not worth spending much time on this one right now anyway.
___
*[go/bindings/derive.go, line 222 at r1](https://reviewable.io/reviews/estuary/flow/135#-MWuCiKu7wNU72owHoLV:-MWzOOrQ1foMRFKBFbeZ:bynvs49) ([raw file](https://github.com/estuary/flow/blob/c1e87e601c8c26d56d1358473e4ab99bf2b33146/go/bindings/derive.go#L222)):*

Perhaps the reason you did this instrumentation is to distinguish "drains from a combine" vs "drains from a derive" ? I'm not sure that's necessary -- perhaps they _should_ be the same metric, since they instrument the same system and mean the same thing -- and we're able to distinguish them already on the FQN.

Or, derive and combine can have their own `drainCombineToCallback` implementations. It's only shared because, to date, they've been the same literal implementation.
</blockquote></details>

Yeah that was it. The thought being that we'd likely want to aggregate these metrics by type, e.g. combiner throughput of all derivations. I think we could also accomplish that using ruler rules to add labels based on the shard name prefix, so I think it's fine to try using just combine metrics for everything.

jgraettinger (approved):
LGTM % making sure metric names align with the prom conventions, as detailed below.

> Reviewed 3 of 8 files at r1, 7 of 7 files at r2.
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/135#-:-MX01avz8fPQsLGg_pOl:b-xmwvot)* status: all files reviewed, 6 unresolved discussions (waiting on @psFried)

___
*[go/bindings/combine.go, line 45 at r1](https://reviewable.io/reviews/estuary/flow/135#-MWtzDHO1JACjYTQtsXr:-MX-svR1Brg9ol34ySbF:bo2w3bk) ([raw file](https://github.com/estuary/flow/blob/c1e87e601c8c26d56d1358473e4ab99bf2b33146/go/bindings/combine.go#L45)):*

is `flow_combine_drain_ops` acceptable instead of `flow_combine_drain_total`? The `_total` seems a little confusing to me.
</blockquote></details>

not my favorite either, but drawn from the prometheus style guide: https://prometheus.io/docs/practices/naming/#metric-names

| a metric name ... should have a suffix describing the unit, in plural form. Note that an accumulating count has total as a suffix, in addition to the unit if applicable.
| e.x. http_requests_*total* (for a unit-less accumulating count)

gazette metrics use it as e.x. `gazette_shard_transactions_total`

And on reading the guide again, it should probably also be e.x. `flow_combine_drain_docs_total`, since "docs" is the unit and "total" because it's an accumulating count.
___
*[go/bindings/combine.go, line 73 at r1](https://reviewable.io/reviews/estuary/flow/135#-MWu09ep-di6U_DKT1NV:-MX-xuOwEa5r24UqIG3i:b5nvnvw) ([raw file](https://github.com/estuary/flow/blob/c1e87e601c8c26d56d1358473e4ab99bf2b33146/go/bindings/combine.go#L73)):*

Well, I think the Combine implementation _is_ coupled to prometheus in the sense that it has to update those metrics somehow. I tried ecapsulating that in a separate struct, which I think is better,  but I'm honestly not sure I understand the concern there so LMK what you think. 

I also think that the resolved metrics (results of `WithLabelValues`) are worth keeping around if it's easy to do so. Again I think the encapsulation helps here, but it's ultimately still coupled to the Combine/Derive instances, so IDK if that really addresses the concern.
</blockquote></details>

I like this factoring :thumbsup:. Only quibble would be that they don't need to be pointers, to make it zero cost.
___
*[go/bindings/combine.go, line 85 at r2](https://reviewable.io/reviews/estuary/flow/135#-MX-vJGd2y1ljTN4T74-:-MX-vJGd2y1ljTN4T740:b-2kiski) ([raw file](https://github.com/estuary/flow/blob/317f6f3dbaf78d4120ef04f3237688a11a7ee3a6/go/bindings/combine.go#L85)):*
> ```Go
> 	c.drained = nil // Invalidate.
> 	c.svc.sendBytes(uint32(pf.CombineAPI_REDUCE_LEFT), doc)
> 	c.stats.leftDocs++
> ```

i like how the `stats` namespacing reads :thumbsup:  
___
*[go/bindings/combine.go, line 242 at r2](https://reviewable.io/reviews/estuary/flow/135#-MX-yvSBEriFECqHw6yf:-MX-yvSBEriFECqHw6yg:b-4z2fdw) ([raw file](https://github.com/estuary/flow/blob/317f6f3dbaf78d4120ef04f3237688a11a7ee3a6/go/bindings/combine.go#L242)):*
> ```Go
> 
> func newCombineMetrics(shard string) *combineMetrics {
> 	return &combineMetrics{
> ```

not for now, just musing: i'm wondering if there's a repeatable builder pattern to be had, where pointers to plain types get associated with a paired metric as part of building a dynamic instance of a function like `recordDrain` (plus reset?).
___
*[go/bindings/derive.go, line 222 at r1](https://reviewable.io/reviews/estuary/flow/135#-MWuCiKu7wNU72owHoLV:-MX0-1DMB17xRfpBtcQl:b-p9dx9f) ([raw file](https://github.com/estuary/flow/blob/c1e87e601c8c26d56d1358473e4ab99bf2b33146/go/bindings/derive.go#L222)):*

Yeah that was it. The thought being that we'd likely want to aggregate these metrics by type, e.g. combiner throughput of all derivations. I think we could also accomplish that using ruler rules to add labels based on the shard name prefix, so I think it's fine to try using just combine metrics for everything.
</blockquote></details>

The way you updated `drainCombineToCallback` is clever and it could be used to maintain separate derive API counts pretty cleanly :shrug: I don't have a great handle on what is going to be more useful for dashboarding.
___
*[go/flow/mapping.go, line 26 at r2](https://reviewable.io/reviews/estuary/flow/135#-MX01MEy5w3mHMHK6TXB:-MX01MEz9uicveLPG1Lm:b-l9i4cb) ([raw file](https://github.com/estuary/flow/blob/317f6f3dbaf78d4120ef04f3237688a11a7ee3a6/go/flow/mapping.go#L26)):*
> ```Go
> 
> var createdPartitionsCounters = promauto.NewCounterVec(prometheus.CounterOpts{
> 	Name: "flow_collection_partitions_created",
> ```

calling this one out 'cause i'm seeing it (haven't checked comprehensively) but per prometheus metric style this should also be `_total`

PR #137, opened 2021-03-30
Updating the sensors example to break it into two tutorials. 

dyaffe:
First, we'll show off reduction annotations and remote URL's, then registers and testing.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/137)

### DIFF
--- a/examples/temp-sensors/flow.ts
+++ b/examples/temp-sensors/flow.ts
@@ -1,84 +1,85 @@
 import { collections, interfaces, registers } from 'flow/modules';
 
-// Implementation for derivation examples/temp-sensors/flow.yaml#/collections/temperature~1average-by-location/derivation.
+// Implementation for derivation examples/temp-sensors/flow.yaml#/collections/temperature~1averageByLocation/derivation.
 export class TemperatureAverageByLocation implements interfaces.TemperatureAverageByLocation {
-
-    // Updates the register value for each reading document.
-    readingsUpdate(
-        source: collections.TemperatureReadings,
-    ): registers.TemperatureAverageByLocation[] {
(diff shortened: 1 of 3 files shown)


examples/temp-sensors/flow.yaml
```diff
@@ -1,92 +1,145 @@
-collections:
-  # Information about a temperature sensor, which may produce many temperature readings
-  temperature/sensors:
-    schema: schemas.yaml#/$defs/tempSensor
-    key: [/id]
+endpoints:
```
psFried, line 1: We should use the `examples/examples.db` endpoint instead of adding this.

psFried (approved):

examples/temp-sensors/flow.yaml
dyaffe: I updated all of this -- let me know if it looks good to merge.

examples/temp-sensors/flow.ts
```diff
@@ -1,84 +1,85 @@
...
-  }
-}
+// Implementation for derivation examples/temp-sensors/flow.yaml#/collections/temperature~1averageTemps/derivation.
+export class TemperatureAverageTemps implements interfaces.TemperatureAverageTemps {
+    averageTempsPublish(
+        source: collections.TemperatureReadings,
+        register: registers.TemperatureAverageTemps,
+        previous: registers.TemperatureAverageTemps,
+    ): collections.TemperatureAverageTemps[] {
+        // This will execute on every reading, so pass in 1 reading for this document.  Reduction annotations
```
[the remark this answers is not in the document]
dyaffe: Reworded this.

PR #140, opened 2021-04-07
More helpful error message for unquoted nulls

psFried:
It's really easy to forget to quote `null` when it's used as a string in
JSON schemas. This adds a more helpful error message when deserializing
JSON schema `type` values. If you write `type: [string, null]`, you will
now see an error message telling you that "null" must be written with
quotes.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/140)

### DIFF
--- a/crates/json/src/schema/build.rs
+++ b/crates/json/src/schema/build.rs
@@ -24,8 +24,8 @@ pub enum Error {
     ExpectedSchema,
     #[error("unexpected fragment component '{0}' of $id keyword")]
     UnexpectedFragment(String),
-    #[error("expected a type or array of types")]
-    ExpectedType,
+    #[error("expected a type or array of types: {0}")]
+    ExpectedType(sj::Error),
     #[error("expected an unsigned integer")]
     ExpectedUnsigned,
     #[error("expected a number")]
@@ -473,7 +473,7 @@ fn build_curi(curi: url::Url, id: Option<&sj::Value>) -> Result<url::Url, Error>
 }
 
 fn extract_type_mask(v: &sj::Value) -> Result<types::Set, Error> {
-    types::Set::deserialize(v).map_err(|_| ExpectedType)
+    types::Set::deserialize(v).map_err(|e| ExpectedType(e))
 }
 
 fn extract_hash(v: &sj::Value) -> HashedLiteral {
--- a/crates/json/src/schema/types.rs
+++ b/crates/json/src/schema/types.rs
@@ -256,6 +256,8 @@ impl serde::Serialize for Set {
     }
 }
 
+/// Deserializes a Set, which may be represented either as a single string, or as an array of
+/// strings.
 struct SetVisitor;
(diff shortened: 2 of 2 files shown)


jgraettinger (changes requested):
> Reviewed 2 of 2 files at r1.
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/140)* status: all files reviewed, 1 unresolved discussion (waiting on @psFried)

___
*[crates/json/src/schema/types.rs, line 313 at r1](https://reviewable.io/reviews/estuary/flow/140#-MXh66IN6obI0RxzFGPa:-MXh66IO7s6e7tRxO1Cl:b1b6ple) ([raw file](https://github.com/estuary/flow/blob/dee5eb8bb80ef6c14f07a6911111e373619777ce/crates/json/src/schema/types.rs#L313)):*
> ```XML
> }
> 
> /// Exists to provide a deserialize impl that can only accept a single string and provides a more
> ```

This is a lot of extra code for this check.

Could we collapse it back into SetVistor, by having visit_seq deserialize into `Set` rather than `TypeStr` ? Technically this would make sets more permissive than the spec by allowing for any array nesting of type names, e.x. `['integer', ['bool', ['string']]` but I don't see that as a problem ?

jgraettinger (approved):
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/140)* status: all files reviewed, 1 unresolved discussion (waiting on @psFried)

___
*[crates/json/src/schema/types.rs, line 313 at r1](https://reviewable.io/reviews/estuary/flow/140#-MXh66IN6obI0RxzFGPa:-MXhBU3Z8bzESpgQYsDB:b-46rsx3) ([raw file](https://github.com/estuary/flow/blob/dee5eb8bb80ef6c14f07a6911111e373619777ce/crates/json/src/schema/types.rs#L313)):*

Yeah, that `[integer, [bool, [string]]]` is exactly why I broke out the separate type. We also could have modified the deserializer to simply accept `null`. In both of those cases, though, we'd be allowing behavior that diverges from the documented spec, and my concern is that it could "reinforce" an incorrect understanding. IMO, the extra code here isn't a problem since it's private and encapsulated, and it's mostly just type signatures anyway. That said, I'm willing to fold it into `SetVisitor` if you still think it's worth it. And I could have it wrap a boolean state that's used to prevent recursing into nested arrays.
</blockquote></details>

Okay, I'm fine with this as-is.

I'd also be fine with technically allowing `[integer, [bool, [string]]]` but simply not documenting it / making no promises to _continue_ to allow it.

plumbing through a bit is probably more trouble than it's worth, as that gets you into the "seeded" forms of visitors to pass through that state as I recall.

jgraettinger (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/estuary/flow/140#-:-MXhCMNOA3hwDZ3bCBul:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @psFried)

PR #141, opened 2021-04-08
reactive runtime & shuffle updates

jgraettinger:
Best to review this commit-by-commit.

Apologies for the size; thanks!

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/141)


jgraettinger (commented):
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/141)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jgraettinger)

___
*[go/runtime/task_term.go, line 30 at r2](https://reviewable.io/reviews/estuary/flow/141#-MXnG9CA2RRzmlbe4ZV6:-MXnOsLGDS0IqtYG8xcW:br1qq0) ([raw file](https://github.com/estuary/flow/blob/2852fb44f6d3830b798834bc64d2f17ec52c0bd4/go/runtime/task_term.go#L30)):*

Can you help me clarify the distinction between the etcd revision and the commons id as they relate to the term lifecycle? I _think_ the deal is that the etcd revision is the only thing we look at in order to determine if there's been a change to a task. So technically a task could change, but still use the same `commons`, and we'd still consider that as starting a new term? That sound right?
</blockquote></details>

Yep that's right. We consider a different revision to be a different term, even if nothing actually change. And correct it could use the same commons (but doesn't in practice).

The other aspect on how revisions are used is that they tell a shuffle server  --  who needs to look up a commons ID -- the minimum revision they must read through to be assured that the commons exists in their catalog keyspace.

PR #148, opened 2021-04-23
Soak tests

psFried:
Gets soak tests running, albeit without the added tests for network segmentations. I'd like to add that in a separate PR, since I'm still figuring out if that even belongs in this repo or ops.

This PR also introduces a different pattern for managing catalogs for different environments. There are top-level yaml files for each environment that each define the endpoint and journal rules to be used in that environment.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/148)


jgraettinger (approved):
Looks great, LGTM

> Reviewed 17 of 17 files at r1, 19 of 19 files at r2, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/148#-:-MYz0-Hm-_E70MmBAsmN:b-aqbxq4)* status: all files reviewed, 3 unresolved discussions (waiting on @psFried)

___
*[examples/soak-tests/set-ops/generator.go, line 20 at r2](https://reviewable.io/reviews/estuary/flow/148#-MYyxfdhBsL3-kzK6TUu:-MYyxfdhBsL3-kzK6TUv:b-g2nh12) ([raw file](https://github.com/estuary/flow/blob/fb94efa85be68e0d3790eb38487cfefa2d7e1efb/examples/soak-tests/set-ops/generator.go#L20)):*
> ```Go
> 
> // We're using these budget model flags because they can easily be used in tests.
> var streamCount = flag.Int("streams", 100, "Number of concurrent streams")
> ```

Why the switch away from `go-flags` ? We use it everywhere else.
___
*[examples/soak-tests/set-ops/k8s-job.yaml, line 19 at r2](https://reviewable.io/reviews/estuary/flow/148#-MYyzCTs9mYb-EcClV68:-MYyzCTs9mYb-EcClV69:b-jpoid3) ([raw file](https://github.com/estuary/flow/blob/fb94efa85be68e0d3790eb38487cfefa2d7e1efb/examples/soak-tests/set-ops/k8s-job.yaml#L19)):*
> ```YAML
>       containers:
>         - name: set-ops-test
>           image: quay.io/giantswarm/golang:1.16.3
> ```

curious, why this repo and not the [official one](https://hub.docker.com/_/golang) ?
___
*[examples/soak-tests/set-ops/setops_test.go, line 1 at r2](https://reviewable.io/reviews/estuary/flow/148#-MYyzWR98jG13FNssZJ5:-MYyzWR98jG13FNssZJ6:b-42mtuh) ([raw file](https://github.com/estuary/flow/blob/fb94efa85be68e0d3790eb38487cfefa2d7e1efb/examples/soak-tests/set-ops/setops_test.go#L1)):*
> ```Go
> package main
> ```

you can use build flags to avoid having this get run during regular CI runs, eg:

```
// +build soak
```

You could also have this be a regular `main` program and not a test?

Ah, or have `go-test` focus on the ./go/ subdirectory, which it looks like you just pushed. Cool with that too.

psFried (commented):
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/148)* status: all files reviewed, 3 unresolved discussions (waiting on @jgraettinger)

___
*[examples/soak-tests/set-ops/generator.go, line 20 at r2](https://reviewable.io/reviews/estuary/flow/148#-MYyxfdhBsL3-kzK6TUu:-MYz0SDM5rsgoOWTbDeg:b-v473ah) ([raw file](https://github.com/estuary/flow/blob/fb94efa85be68e0d3790eb38487cfefa2d7e1efb/examples/soak-tests/set-ops/generator.go#L20)):*

Why the switch away from `go-flags` ? We use it everywhere else.
</blockquote></details>

Yeah, and `go-flags` seems a lot nicer, too. But the reason is that the tests are run using `go test ...`, which uses the `flag` package, and as best I could tell from the docs, this was the only way to have flags that work with tests. I didn't actually test it with `go-flags`, though, so it's possible that might have worked, too. Just didn't seem worth it at this stage. But I can look into it if you think it's worth trying to stay standardized on `go-flags`.
___
*[examples/soak-tests/set-ops/k8s-job.yaml, line 19 at r2](https://reviewable.io/reviews/estuary/flow/148#-MYyzCTs9mYb-EcClV68:-MYz1ZGh0RAyX3FX1hDN:b-541z36) ([raw file](https://github.com/estuary/flow/blob/fb94efa85be68e0d3790eb38487cfefa2d7e1efb/examples/soak-tests/set-ops/k8s-job.yaml#L19)):*

curious, why this repo and not the [official one](https://hub.docker.com/_/golang) ?
</blockquote></details>

Oh I was just having some temporary network slowness reaching docker hub that was making it unusable, so I switched to a quay image just so it would download quickly. Can probably change it back at this point.
___
*[examples/soak-tests/set-ops/setops_test.go, line 1 at r2](https://reviewable.io/reviews/estuary/flow/148#-MYyzWR98jG13FNssZJ5:-MYz2a2m4GDDAaSqqoSs:b-n0dz7q) ([raw file](https://github.com/estuary/flow/blob/fb94efa85be68e0d3790eb38487cfefa2d7e1efb/examples/soak-tests/set-ops/setops_test.go#L1)):*

you can use build flags to avoid having this get run during regular CI runs, eg:

```
// +build soak
```

You could also have this be a regular `main` program and not a test?

Ah, or have `go-test` focus on the ./go/ subdirectory, which it looks like you just pushed. Cool with that too.
</blockquote></details>

Oh yeah, the build tag would have been good, too. I think I had decided that I'd have to use the `testing` package in order to integrate urkel. But since realized that the chaos testing really ought to be separate from the functional testing pieces, and I just never revisited the `testing` piece. Really, it would have been way easier to just run this as a main program. Then I could use `go-flags` too. Do you think it's worth taking another pass and cleaning that up? I'm kinda on the fence.

jgraettinger (changes requested):
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/148)* status: 28 of 29 files reviewed, 3 unresolved discussions (waiting on @jgraettinger and @psFried)

___
*[examples/soak-tests/set-ops/generator.go, line 20 at r2](https://reviewable.io/reviews/estuary/flow/148#-MYyxfdhBsL3-kzK6TUu:-MYz5gYkDYoVHk99v8fm:b-4h2arr) ([raw file](https://github.com/estuary/flow/blob/fb94efa85be68e0d3790eb38487cfefa2d7e1efb/examples/soak-tests/set-ops/generator.go#L20)):*

Yeah, and `go-flags` seems a lot nicer, too. But the reason is that the tests are run using `go test ...`, which uses the `flag` package, and as best I could tell from the docs, this was the only way to have flags that work with tests. I didn't actually test it with `go-flags`, though, so it's possible that might have worked, too. Just didn't seem worth it at this stage. But I can look into it if you think it's worth trying to stay standardized on `go-flags`.
</blockquote></details>

got it, I'd missed that `go test` will be the thing driving flag parsing.
___
*[examples/soak-tests/set-ops/setops_test.go, line 1 at r2](https://reviewable.io/reviews/estuary/flow/148#-MYyzWR98jG13FNssZJ5:-MYz4zJd83aEEEZ6mawy:bo8ykhq) ([raw file](https://github.com/estuary/flow/blob/fb94efa85be68e0d3790eb38487cfefa2d7e1efb/examples/soak-tests/set-ops/setops_test.go#L1)):*

Oh yeah, the build tag would have been good, too. I think I had decided that I'd have to use the `testing` package in order to integrate urkel. But since realized that the chaos testing really ought to be separate from the functional testing pieces, and I just never revisited the `testing` piece. Really, it would have been way easier to just run this as a main program. Then I could use `go-flags` too. Do you think it's worth taking another pass and cleaning that up? I'm kinda on the fence.
</blockquote></details>

I agree that I don't really see the point of having the job itself be a test (vs having a separate test that injects faults). Didn't feel strongly so didn't bring it up.

I do think it'd overall be cleaner as a regular `func main()` using `go-flags`. If it's quick, go for it ?

PR #152, opened 2021-05-12
initial kinesis poc

psFried:
Draft of kinesis captures. The basics are working&trade; end-to-end, but there's still some things left to do. Here's what I know of so far for a punch list:

- [ ] Add whatever unit tests we can
- [ ] Integration test (need to figure out how we want to approach these)
- [ ] Basic perf test, even just a one-off manual test, with lots of shards to make sure performance is at least in the right ballpark.
- [ ] Investigate issue with flowctl develop sometimes not stopping when it receives sigterm. Maybe a kinesis term isn't ending properly?
- [ ] Determine proper values for GetRecords `Limit` and `dataCh` buffer size.
- [ ] Apply hash function to kinesis shard ids and check if they should be processed by the shard
- [ ] Handle rate-limiting errors returned by the kinesis api by applying a backoff and re-trying
- [ ] Re-try other retry-able errors
- [ ] Cleanup and add comments

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/152)


psFried:
Closing this. New PR is in the [estuary/connectors](https://github.com/estuary/connectors) repo

PR #153, opened 2021-05-18
implement shard split workflow

jgraettinger:


jgraettinger:
Did some more testing trying to reproduce / study / grab stack traces during delayed data around shard splits, and an interesting / odd thing I'm seeing is that there's some significant tail latency in nodeJS invocation. The soak test isn't producing a ton of load on the system, yet there are _hundreds_ (out of ~60K) of requests which have a recorded histogram latency of > 1 second.

Current best guess is that there's something in how we're driving the nodeJS worker that's tickled by new shards / requests coming online.

psFried (approved):
The code here looks good, and I definitely appreciated the voluminous comments. I'm not sure what to do about those couple of TODOs and the nodejs tail latencies. They all seem pretty important to understand, but it doesn't seem like there would be any harm in looking into those after this is merged.

> Reviewed 5 of 5 files at r1, 3 of 3 files at r2, 3 of 3 files at r3, 3 of 3 files at r4, 1 of 1 files at r5, 4 of 4 files at r6, 5 of 5 files at r7, 13 of 13 files at r8.
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/153#-:-MaAJoc51DjaDORsZy4v:badonp9)* status: all files reviewed, 1 unresolved discussion (waiting on @jgraettinger)

___
*[go/labels/labels.go, line 117 at r2](https://reviewable.io/reviews/estuary/flow/153#-MaA1iPt-57oOWntR62l:-MaA1iPt-57oOWntR62m:beuadc7) ([raw file](https://github.com/estuary/flow/blob/1918ac2dba03a42ce998fbd27b31f666e48f262c/go/labels/labels.go#L117)):*
> ```Go
> 
> // BuildShardID builds the ShardID that's implied by the LabelSet.
> func BuildShardID(set pb.LabelSet) (pc.ShardID, error) {
> ```

I really like this factoring :+1: 
___
*[go/runtime/split_workflow.go, line 30 at r8](https://reviewable.io/reviews/estuary/flow/153#-MaA7FZpC3u32kVumehB:-MaA7FZpC3u32kVumehC:bvbvx76) ([raw file](https://github.com/estuary/flow/blob/43304b10bdaa843350be329a8575f4e0c7fd5a3d/go/runtime/split_workflow.go#L30)):*
> ```Go
> 
> 	// Inspect the status of the split, extracting a ready LHS primary if
> 	// there is one. This may block indefinitely if |req| references a
> ```

This is exactly the question I had after reading the previous comment, so :+1:  for good commenting. I'm still not sure what the failure mode looks like in practice, though. If I have a failed shard, for example, what happens if I make a request to split it? Is there a check somewhere else that prevents me from doing that?
___
*[go/runtime/split_workflow.go, line 194 at r8](https://reviewable.io/reviews/estuary/flow/153#-MaA9ZNG7f3bA9TA2_rl:-MaA9ZNG7f3bA9TA2_rm:b-ec9d63) ([raw file](https://github.com/estuary/flow/blob/43304b10bdaa843350be329a8575f4e0c7fd5a3d/go/runtime/split_workflow.go#L194)):*
> ```Go
> 	var lhsRange, rhsRange = parentRange, parentRange
> 
> 	if req.SplitOnKey {
> ```

The request has separate fields for `SplitOnKey` vs `SplitOnRClock`, which makes it seem like they're not mutually exclusive. This code here is more like what I was expecting, which is that you can either split on key or rclock, but not both at the same time. Unless there's something else I'm missing, it may be worth updating the request or else maybe just adding a check in `Validate`.

jgraettinger (commented):
Thanks. The allocator bit is future work that makes operations easier in a future world where there are lots of automated splits. Not a problem we have today.

The data delays TODO is more of a generalized "hey are their weird tail latencies in a loaded down nodeJS and what's causing them?".

> *[Reviewable](https://reviewable.io/reviews/estuary/flow/153#-:-MaANze1CS6X4D8Ymb5M:b-r2oo7z)* status: all files reviewed, 1 unresolved discussion (waiting on @psFried)

___
*[go/runtime/split_workflow.go, line 30 at r8](https://reviewable.io/reviews/estuary/flow/153#-MaA7FZpC3u32kVumehB:-MaAMu175ljZ5Ccn_5O8:b-mh398u) ([raw file](https://github.com/estuary/flow/blob/43304b10bdaa843350be329a8575f4e0c7fd5a3d/go/runtime/split_workflow.go#L30)):*

This is exactly the question I had after reading the previous comment, so :+1:  for good commenting. I'm still not sure what the failure mode looks like in practice, though. If I have a failed shard, for example, what happens if I make a request to split it? Is there a check somewhere else that prevents me from doing that?
</blockquote></details>

there's nothing wrong with making a request to split it -- the `StartSplit` workflow will succeed and will create a RHS shard marked as the split target.

Once assigned, that shard will remain marked as `BACKFILL` and will block (here) until the parent shard becomes healthy again, at which point it can make progress and this unblocks.
___
*[go/runtime/split_workflow.go, line 194 at r8](https://reviewable.io/reviews/estuary/flow/153#-MaA9ZNG7f3bA9TA2_rl:-MaANc9D9h0QCc70aujF:bon7hkw) ([raw file](https://github.com/estuary/flow/blob/43304b10bdaa843350be329a8575f4e0c7fd5a3d/go/runtime/split_workflow.go#L194)):*

The request has separate fields for `SplitOnKey` vs `SplitOnRClock`, which makes it seem like they're not mutually exclusive. This code here is more like what I was expecting, which is that you can either split on key or rclock, but not both at the same time. Unless there's something else I'm missing, it may be worth updating the request or else maybe just adding a check in `Validate`.
</blockquote></details>

There already is a check in Validate (`if SplitOnKey == SplitOnRClock { error }`)

PR #157, opened 2021-06-09
Make Flow key and rclock ranges inclusive

psFried:
The ranges used for key and rclock values were inclusive at the
beginning and exclusive at the end. This ended up being a little tricky
to work with, as it required a special case for handling hashed values
at the MaxUint32 boundary. It also presented challenges when
constructing Flow 32 bit range specs from other (kinesis, for example)
ranges that use more bits to represent their ranges. Just using the high
32 bits from larger representations could result in 32 bit ranges where
begin == end (empty ranges), which require even more special cases.

The solution implemented here is to make these ranges inclusive on both
ends. This allows the ranges to fully represent all possible hashed
values, and requires no special cases for MaxUint32. Ranges where begin
== end are now also valid and non-empty.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/157)

### DIFF
--- a/crates/protocol/src/flow.rs
+++ b/crates/protocol/src/flow.rs
@@ -427,7 +427,7 @@ pub mod test_spec {
 /// is responsible for.
 #[derive(Clone, PartialEq, ::prost::Message)]
 pub struct RangeSpec {
-    /// [begin, end) exclusive range of keys to be shuffled to this reader.
+    /// [begin, end] inclusive range of keys to be shuffled to this reader.
     /// Ranges are with respect to a 32-bit hash of a packed document key.
     ///
     /// The choice of hash function is important: while it need not be
@@ -443,7 +443,7 @@ pub struct RangeSpec {
     pub key_begin: u32,
     #[prost(fixed32, tag="3")]
     pub key_end: u32,
-    /// Rotated [begin, end) exclusive ranges of Clocks.
+    /// Rotated [begin, end] inclusive ranges of Clocks.
     #[prost(fixed32, tag="4")]
     pub r_clock_begin: u32,
     #[prost(fixed32, tag="5")]
--- a/go/labels/labels.go
+++ b/go/labels/labels.go
@@ -22,7 +22,7 @@ const (
 	KeyBegin = "estuary.dev/key-begin"
 	// KeyBeginMin is the minimum possible key.
 	KeyBeginMin = "00000000"
-	// KeyEnd is a hexadecimal encoding of the ending key range (exclusive)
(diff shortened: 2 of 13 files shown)


jgraettinger (changes requested):
One comment on how we compute splits, but otherwise LGTM!

> Reviewed 13 of 13 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/157#-:-MbmdocPCH7DOXs2DgPu:bgnd5ez)* status: all files reviewed, 1 unresolved discussion (waiting on @psFried)

___
*[go/runtime/split_workflow.go, line 195 at r1](https://reviewable.io/reviews/estuary/flow/157#-MbmcLU25wRS0wwLHtdm:-MbmcLU3BnNiMjw8IH86:b1j0shf) ([raw file](https://github.com/estuary/flow/blob/a5744a6c92f38139fc1b8b6357619f9abfead38c/go/runtime/split_workflow.go#L195)):*
> ```Go
> 
> 	if req.SplitOnKey {
> 		var pivot = (parentRange.KeyBegin / 2) + (parentRange.KeyEnd / 2)
> ```

I'm wondering if we should add one to KeyEnd (in 64-bit space) and _then_ divide by two.

I believe this gives slightly more intuitive results: range [0, 1<<32 - 1] splits into [0, 1<<31 -1] and [1<<31, 1<<32-1] for example, and [0, 1<<31 -1 ] into [0, 1<<30 -1], [1<<30, 1<<31 - 1]

psFried (commented):
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/157)* status: 12 of 13 files reviewed, 1 unresolved discussion (waiting on @jgraettinger)

___
*[go/runtime/split_workflow.go, line 195 at r1](https://reviewable.io/reviews/estuary/flow/157#-MbmcLU25wRS0wwLHtdm:-Mbmuo3O1N8oiRPHHtMl:b-yxy11s) ([raw file](https://github.com/estuary/flow/blob/a5744a6c92f38139fc1b8b6357619f9abfead38c/go/runtime/split_workflow.go#L195)):*

I'm wondering if we should add one to KeyEnd (in 64-bit space) and _then_ divide by two.

I believe this gives slightly more intuitive results: range [0, 1<<32 - 1] splits into [0, 1<<31 -1] and [1<<31, 1<<32-1] for example, and [0, 1<<31 -1 ] into [0, 1<<30 -1], [1<<30, 1<<31 - 1]
</blockquote></details>

Yeah good call. I updated this so that now the powers of two always get split as you'd expect.

PR #158, opened 2021-06-11
protocol internals work in support of connectors, and fixing materialization shard split checkpoints

jgraettinger:
* Update many protobuf definitions to better support JSON encoding
* Add a (still work-in-progress) "flow sink" JSON connector protocol driver for materialization.
  - I've tested this using local (non-dockerized) programs, but it's not quite ready for prime time yet.
    I plan to come back to it after putting time into the captures side of the runtime,
    but it's far enough along to merit review now.
* Update materialization-stored checkpoint semantics to correctly handle shard splits.
   Introduce new SQL semantics to store and properly fence the complete tree of checkpoints and key ranges.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/158)


jgraettinger:
I've added a further commit which introduces a capture protocol and integrates it's Validate RPC into the build process.

psFried (approved):
Ok I think I've understood what's here so far. My sense is that there's still some work to be done in order to get things working and integrated end to end (for example, wiring this into flow consumer), and so some details here may change, but that the essentials seem to be right.

> Reviewed 23 of 23 files at r1, 12 of 12 files at r2, 28 of 28 files at r3, 32 of 32 files at r4.
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/158#-:-McAz54p-gPhLy7tHrfw:beu38zo)* status: all files reviewed, 3 unresolved discussions (waiting on @jgraettinger)

___
*[crates/protocol/src/flow.rs, line 955 at r4](https://reviewable.io/reviews/estuary/flow/158#-McAxaihETZflnWu4J4t:-McAxaii43jOkdOERMez:b-gsi3az) ([raw file](https://github.com/estuary/flow/blob/67276aa47379a7bc7b475c8825f414eaa7eb917c/crates/protocol/src/flow.rs#L955)):*
> ```XML
>         TrampolineFetch = 3,
>         /// Trampoline sub-type: Start validation of a materialization.
>         TrampolineValidateCapture = 4,
> ```

nit: comment should say "capture" instead of "materialization"
___
*[go/materialize/driver/sql2/fence.go, line 65 at r3](https://reviewable.io/reviews/estuary/flow/158#-McAeDSKDuRhcxZ-eSjG:-McAeDSKDuRhcxZ-eSjH:b-4aws0e) ([raw file](https://github.com/estuary/flow/blob/5462deb551517ba6f3bb0089f9dcd6dff0065477/go/materialize/driver/sql2/fence.go#L65)):*
> ```Go
> 				SET fence=fence+1
> 				WHERE materialization=%s
> 				AND (
> ```

nit: I think this could be simplified to `key_end >= this_key_begin AND key_begin <= this_key_end` where `this_` refers to the key range of the running materialization shard.
___
*[go/protocols/flow/materialize_extensions.go, line 57 at r1](https://reviewable.io/reviews/estuary/flow/158#-McAS5hS430j3N9oOBdH:-McAS5hS430j3N9oOBdI:bv1b4r3) ([raw file](https://github.com/estuary/flow/blob/ed5365b793cb91360d21a358e0eb865cd6244413/go/protocols/flow/materialize_extensions.go#L57)):*
> ```Go
> 	}
> 	for key := range fields.FieldConfigJson {
> 		if string(fields.FieldConfigJson[key]) != string(other.FieldConfigJson[key]) {
> ```

How's this work if the `FieldConfigJson` values are objects? Are we ensuring that objects are always encoded with keys in the same order?

jgraettinger (commented):
Yep, that's right. Thanks.

> *[Reviewable](https://reviewable.io/reviews/estuary/flow/158#-:-McCPNJh7hhuRDO1dgjs:buatn9t)* status: all files reviewed, 3 unresolved discussions (waiting on @jgraettinger and @psFried)

___
*[go/materialize/driver/sql2/fence.go, line 65 at r3](https://reviewable.io/reviews/estuary/flow/158#-McAeDSKDuRhcxZ-eSjG:-McCNRW57LnZunYXd4Me:b-lb5zhx) ([raw file](https://github.com/estuary/flow/blob/5462deb551517ba6f3bb0089f9dcd6dff0065477/go/materialize/driver/sql2/fence.go#L65)):*

nit: I think this could be simplified to `key_end >= this_key_begin AND key_begin <= this_key_end` where `this_` refers to the key range of the running materialization shard.
</blockquote></details>

Huh. Yea. Math. Nice!
___
*[go/protocols/flow/materialize_extensions.go, line 57 at r1](https://reviewable.io/reviews/estuary/flow/158#-McAS5hS430j3N9oOBdH:-McCOu2JDxHvLTbpeGE-:b9o6kyv) ([raw file](https://github.com/estuary/flow/blob/ed5365b793cb91360d21a358e0eb865cd6244413/go/protocols/flow/materialize_extensions.go#L57)):*

How's this work if the `FieldConfigJson` values are objects? Are we ensuring that objects are always encoded with keys in the same order?
</blockquote></details>

FWIW, this isn't new code but is manually copied from the generated equals extension of gogoprotobuf - which was turned off for this message type because json.RawMessage doesn't support equality.

Yep, this relies on the fact that serde_json is encoding in alphabetical order.

PR #162, opened 2021-07-07
complete Airbyte source driver support, and implement capture runtime

jgraettinger:
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/162)


psFried (approved):
> Reviewed 4 of 4 files at r1, 4 of 4 files at r2, 7 of 7 files at r3, 12 of 12 files at r4, 2 of 2 files at r5, 3 of 3 files at r6, 1 of 1 files at r7, 6 of 6 files at r8, 6 of 6 files at r9, 1 of 1 files at r10.
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/162)* status: all files reviewed, 4 unresolved discussions (waiting on @jgraettinger)

___
*[crates/build/src/nodejs/typescript/mapping.rs, line 111 at r1](https://reviewable.io/reviews/estuary/flow/162#-Me0wEBK5E8oZbhyFjBN:-Me0wEBL52F3GJELOFYn:b-7kabrv) ([raw file](https://github.com/estuary/flow/blob/f35bbfefe21ed06688e48b4ba6e3c9d90d689b69/crates/build/src/nodejs/typescript/mapping.rs#L111)):*
> ```XML
>                 prop.name.clone()
>             } else {
>                 format!("{:?}", prop.name)
> ```

non-blocking (extra pedantic) question here is whether this method of escaping will work in all cases. Specifically, I'm thinking that Rust's unicode escape sequences use a different syntax than JS does, so if there's anything that might get spit out in `\u{xxxx}` form, then that might break something? It may not actually be an issue at all if we already disallow those characters, and I don't think it's worth the effort right now to allow :poop: in a variable name. But maybe worth thinking about if `serde_json` might provide a better way to do this.
___
*[go/capture/driver/airbyte/connector.go, line 120 at r6](https://reviewable.io/reviews/estuary/flow/162#-Me0zeOQBa2aJmedvRAl:-Me0zeOQBa2aJmedvRAm:b-2txold) ([raw file](https://github.com/estuary/flow/blob/f1a531dc72a5f6ed4fb3e7b33cb335680fb9a829/go/capture/driver/airbyte/connector.go#L120)):*
> ```Go
> 	}
> 
> 	log.WithField("fe", fe.unwrap()).Info("connector exited")
> ```

nit: `"error"` may be a better field name for the logs
___
*[go/capture/driver/airbyte/driver.go, line 80 at r6](https://reviewable.io/reviews/estuary/flow/162#-Me18LAh9pXeYl-GHcBW:-Me18LAi13rY0uFDLKuT:belyd4w) ([raw file](https://github.com/estuary/flow/blob/f1a531dc72a5f6ed4fb3e7b33cb335680fb9a829/go/capture/driver/airbyte/driver.go#L80)):*
> ```Go
> 		// No stdin is sent to the connector.
> 		func(w io.Writer) error { return nil },
> 		// Expect to decode Airbyte messages, and a ConnectionStatus specifically.
> ```

nit: comment says "ConnectionStatus" when I think "ConnectorSpecification" was intended
___
*[go/capture/driver/airbyte/driver.go, line 96 at r6](https://reviewable.io/reviews/estuary/flow/162#-Me18XSFE6kWaxn3UVvn:-Me18XSG5slntb0WLA-7:b-28tkah) ([raw file](https://github.com/estuary/flow/blob/f1a531dc72a5f6ed4fb3e7b33cb335680fb9a829/go/capture/driver/airbyte/driver.go#L96)):*
> ```Go
> 	)
> 
> 	// Expect connector spit out a successful ConnectionStatus.
> ```

here too, "ConnectorSpecification"
___
*[go/capture/driver/airbyte/driver.go, line 352 at r6](https://reviewable.io/reviews/estuary/flow/162#-Me197MC5Wn2IslQxU48:-Me197MC5Wn2IslQxU49:bn5yv3n) ([raw file](https://github.com/estuary/flow/blob/f1a531dc72a5f6ed4fb3e7b33cb335680fb9a829/go/capture/driver/airbyte/driver.go#L352)):*
> ```Go
> 
> 	// Write a final commit, followed by EOF.
> 	return lifecycle.WriteCommit(stream, &resp,
> ```

Placeholder: Will this `DriverCheckpointJson` overwrite a previous value? I'm guessing whatever handles the response will just ignore the `nil`, but I haven't gotten to that part yet.
___
*[go/flowctl/cmd-apply.go, line 202 at r9](https://reviewable.io/reviews/estuary/flow/162#-Me1LKy9CboECNhb7y4T:-Me1LKyAAdt5AM-VRdQd:b-r2n312) ([raw file](https://github.com/estuary/flow/blob/6e0f4e5e229062146e2c9f937b43d361e33fb77b/go/flowctl/cmd-apply.go#L202)):*
> ```Go
> 		}
> 
> 		resp, err := consumer.ListShards(context.Background(), shards, &pc.ListRequest{
> ```

Not something that I think needs addressed in this PR, but just want to note that this will need adjusted in order to account for multiple shards, since the RClockBegin label is a part of the id. Eventually, I think we'd probably want to do the shard listing based only on the task name and type. I'm guessing this is already on your radar, but just want to make sure.

jgraettinger (commented):
Thanks!

> *[Reviewable](https://reviewable.io/reviews/estuary/flow/162#-:-Me5-PZc1SrDGdYX4Jz-:b-ucfknh)* status: 3 of 43 files reviewed, 2 unresolved discussions (waiting on @psFried)

___
*[crates/build/src/nodejs/typescript/mapping.rs, line 111 at r1](https://reviewable.io/reviews/estuary/flow/162#-Me0wEBK5E8oZbhyFjBN:-Me4xezIEvVAa7VkI3cm:b-sgu4gd) ([raw file](https://github.com/estuary/flow/blob/f35bbfefe21ed06688e48b4ba6e3c9d90d689b69/crates/build/src/nodejs/typescript/mapping.rs#L111)):*

non-blocking (extra pedantic) question here is whether this method of escaping will work in all cases. Specifically, I'm thinking that Rust's unicode escape sequences use a different syntax than JS does, so if there's anything that might get spit out in `\u{xxxx}` form, then that might break something? It may not actually be an issue at all if we already disallow those characters, and I don't think it's worth the effort right now to allow :poop: in a variable name. But maybe worth thinking about if `serde_json` might provide a better way to do this.
</blockquote></details>

Done (switched to serde_json::Value::String(...).to_string())
___
*[go/capture/driver/airbyte/connector.go, line 120 at r6](https://reviewable.io/reviews/estuary/flow/162#-Me0zeOQBa2aJmedvRAl:-Me4xp46-J3mkQ_vFIGx:bb6un1j) ([raw file](https://github.com/estuary/flow/blob/f1a531dc72a5f6ed4fb3e7b33cb335680fb9a829/go/capture/driver/airbyte/connector.go#L120)):*

nit: `"error"` may be a better field name for the logs
</blockquote></details>

Done
___
*[go/capture/driver/airbyte/driver.go, line 80 at r6](https://reviewable.io/reviews/estuary/flow/162#-Me18LAh9pXeYl-GHcBW:-Me4xwl85wbC7c4ULFnc:bb6un1j) ([raw file](https://github.com/estuary/flow/blob/f1a531dc72a5f6ed4fb3e7b33cb335680fb9a829/go/capture/driver/airbyte/driver.go#L80)):*

nit: comment says "ConnectionStatus" when I think "ConnectorSpecification" was intended
</blockquote></details>

Done
___
*[go/capture/driver/airbyte/driver.go, line 96 at r6](https://reviewable.io/reviews/estuary/flow/162#-Me18XSFE6kWaxn3UVvn:-Me4y-uR9QkAk6pdpo48:bb6un1j) ([raw file](https://github.com/estuary/flow/blob/f1a531dc72a5f6ed4fb3e7b33cb335680fb9a829/go/capture/driver/airbyte/driver.go#L96)):*

here too, "ConnectorSpecification"
</blockquote></details>

Done
___
*[go/capture/driver/airbyte/driver.go, line 352 at r6](https://reviewable.io/reviews/estuary/flow/162#-Me197MC5Wn2IslQxU48:-Me4yEMO-YvGbbEHygGQ:b-wmok5k) ([raw file](https://github.com/estuary/flow/blob/f1a531dc72a5f6ed4fb3e7b33cb335680fb9a829/go/capture/driver/airbyte/driver.go#L352)):*

Placeholder: Will this `DriverCheckpointJson` overwrite a previous value? I'm guessing whatever handles the response will just ignore the `nil`, but I haven't gotten to that part yet.
</blockquote></details>

It will. But this is also only reached if the connector writes output and exits _without_ writing a final state checkpoint.

In that case, this handling means we'll a) commit it's output, because we require a Commit, and b) will start the _next_ invocation as a full refresh.

Expanded on this in the comment.
___
*[go/flowctl/cmd-apply.go, line 202 at r9](https://reviewable.io/reviews/estuary/flow/162#-Me1LKy9CboECNhb7y4T:-Me4zIlZ5NzBBiC4A_L0:bvoslg2) ([raw file](https://github.com/estuary/flow/blob/6e0f4e5e229062146e2c9f937b43d361e33fb77b/go/flowctl/cmd-apply.go#L202)):*

Not something that I think needs addressed in this PR, but just want to note that this will need adjusted in order to account for multiple shards, since the RClockBegin label is a part of the id. Eventually, I think we'd probably want to do the shard listing based only on the task name and type. I'm guessing this is already on your radar, but just want to make sure.
</blockquote></details>

I'm not quite following this. The key and rclock begin are indeed part of the shard ID, but they also are stable for a given shard (only the End may change). There must always be a shard having an ID with key/clock begins of zero.

This apply is also a no-op if the shard already exists.

psFried:
___
*[go/flowctl/cmd-apply.go, line 202 at r9](https://reviewable.io/reviews/estuary/flow/162#-Me1LKy9CboECNhb7y4T:-Me5cEadC4RuweRMGKAN:b-ewrsyr) ([raw file](https://github.com/estuary/flow/blob/6e0f4e5e229062146e2c9f937b43d361e33fb77b/go/flowctl/cmd-apply.go#L202)):*

I'm not quite following this. The key and rclock begin are indeed part of the shard ID, but they also are stable for a given shard (only the End may change). There must always be a shard having an ID with key/clock begins of zero.

This apply is also a no-op if the shard already exists.
</blockquote></details>

Yeah, I get it now. Even if the shard had previously been split, there must always be one that has a key/rclock begin of zero. Just didn't occur to me at the time, but it makes sense now.

PR #163, opened 2021-07-09
flowctl develop --poll

jgraettinger:
Leverage dataflow topology tracing from testing infrastructure to enable a non-block mode of `flowctl develop`.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/163)


psFried (approved):
> Reviewed 35 of 35 files at r1.
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/163)* status: all files reviewed, 1 unresolved discussion (waiting on @jgraettinger)

___
*[go/testing/cluster.go, line 280 at r1](https://reviewable.io/reviews/estuary/flow/163#-MeN1lAe5rWlUbV8wuty:-MeN1lAe5rWlUbV8wutz:bc42dxj) ([raw file](https://github.com/estuary/flow/blob/fefe7e3a099dea5ea435091187ca7e833ea37ba6/go/testing/cluster.go#L280)):*
> ```Go
> 				status  = decoded.AssignmentValue.(*pc.ReplicaStatus)
> 			)
> 			if decoded.Slot == 0 && status.Code != pc.ReplicaStatus_PRIMARY {
> ```

Seems like maybe this should handle the case where the shard is failed. Should this maybe return an error in that case, since the shard will likely never be primary? Or does `WaitForRevision` already return an error if the shard fails?

jgraettinger (commented):
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/163)* status: 34 of 35 files reviewed, all discussions resolved (waiting on @psFried)

___
*[go/testing/cluster.go, line 280 at r1](https://reviewable.io/reviews/estuary/flow/163#-MeN1lAe5rWlUbV8wuty:-MeQ8dfl6QRp9ui_X5TI:bb6un1j) ([raw file](https://github.com/estuary/flow/blob/fefe7e3a099dea5ea435091187ca7e833ea37ba6/go/testing/cluster.go#L280)):*

Seems like maybe this should handle the case where the shard is failed. Should this maybe return an error in that case, since the shard will likely never be primary? Or does `WaitForRevision` already return an error if the shard fails?
</blockquote></details>

Done

PR #164, opened 2021-07-12
Install docker cli in flow image

psFried:
Just a draft for now. Still waiting to see if this actually works OK in practice.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/164)

### DIFF
--- a/go/flow/catalog.go
+++ b/go/flow/catalog.go
@@ -106,10 +106,13 @@ func NewCatalog(ctx context.Context, etcd *clientv3.Client, root string) (Catalo
 	return catalog, nil
 }
 
-// GetTask returns the named CatalogTask, Commons, and its Commons ModRevision.
-func (c Catalog) GetTask(name string) (_ *pf.CatalogTask, _ *Commons, revision int64, _ error) {
+// GetTask returns the named CatalogTask, Commons, and its Commons ModRevision. It first ensures
+// that the given revision has been observed by the keyspace so that the caller can guarantee that a
+// task update has been observed if its revision is known.
+func (c Catalog) GetTask(ctx context.Context, name string, waitForRevision int64) (_ *pf.CatalogTask, _ *Commons, revision int64, _ error) {
 	c.Mu.RLock()
 	defer c.Mu.RUnlock()
+	c.KeySpace.WaitForRevision(ctx, waitForRevision)
 	return c.getTask(name)
 }
 
--- a/go/flow/catalog.go
+++ b/go/flow/catalog.go
@@ -7,6 +7,7 @@ import (
 	"path"
 	"runtime"
 	"sort"
+	"strconv"
 
 	"github.com/estuary/flow/go/bindings"
 	pf "github.com/estuary/flow/go/protocols/flow"
@@ -109,10 +110,16 @@ func NewCatalog(ctx context.Context, etcd *clientv3.Client, root string) (Catalo
 // GetTask returns the named CatalogTask, Commons, and its Commons ModRevision. It first ensures
 // that the given revision has been observed by the keyspace so that the caller can guarantee that a
 // task update has been observed if its revision is known.
-func (c Catalog) GetTask(ctx context.Context, name string, waitForRevision int64) (_ *pf.CatalogTask, _ *Commons, revision int64, _ error) {
+func (c Catalog) GetTask(ctx context.Context, name string, taskCreated string) (_ *pf.CatalogTask, _ *Commons, revision int64, _ error) {
 	c.Mu.RLock()
 	defer c.Mu.RUnlock()
-	c.KeySpace.WaitForRevision(ctx, waitForRevision)
+	if taskCreated != "" {
+		var waitForRevision, err = strconv.ParseInt(taskCreated, 10, 64)
+		if err != nil {
+			return nil, nil, 0, fmt.Errorf("parsing task-created label: %w", err)
+		}
+		c.KeySpace.WaitForRevision(ctx, waitForRevision)
+	}
 	return c.getTask(name)
 }
 
--- a/go/flowctl/cmd-apply.go
+++ b/go/flowctl/cmd-apply.go
@@ -122,15 +122,15 @@ func (cmd cmdApply) Execute(_ []string) error {
 
 	if !cmd.DryRun {
 		// Apply capture shard specs.
-		if err = applyCaptureShards(built, shards, 1); err != nil {
+		if err = applyCaptureShards(built, shards, 1, revision); err != nil {
 			return fmt.Errorf("applying capture shards: %w", err)
 		}
 		// Apply derivation shard specs.
-		if err = applyDerivationShards(built, shards, 1); err != nil {
+		if err = applyDerivationShards(built, shards, 1, revision); err != nil {
 			return fmt.Errorf("applying derivation shards: %w", err)
 		}
 		// Apply materialization shards.
-		if err = applyMaterializationShards(built, shards, 1); err != nil {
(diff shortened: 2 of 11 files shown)


go/runtime/flow_consumer.go
```diff
@@ -297,7 +298,12 @@ func shardApply(ctx context.Context, svc *consumer.Service,
 		if nextSpec != nil && logSpec == nil {
 			// Grab labeled catalog task and its journal rules.
 			var name = nextSpec.LabelSet.ValueOf(labels.TaskName)
-			var _, commons, _, err = catalog.GetTask(name)
+			var revisionStr = nextSpec.LabelSet.ValueOf(labels.CatalogRevision)
+			var catalogRevision, err = strconv.ParseInt(revisionStr, 10, 64)
+			if err != nil {
+				return nil, fmt.Errorf("parsing catalog revision: %w", err)
+			}
+			_, commons, _, err := catalog.GetTaskWithRevision(ctx, name, catalogRevision)
```
jgraettinger, line 306: Oh! It's not even about the reactor starting the shard and not seeing the catalog task -- it's about `flowctl apply` driving the shard Apply RPC before the reactor has read-through the catalog.

This is gross, but I actually am thinking that simply adding a `time.Sleep` in `flowctl apply` to account for this is the right move right now. Wiring up this revision as a shard label works, but is feeling weird to me -- that we haven't gotten the right API. And we're likely to change the means by which the catalog is applied to Etcd anyway while building out the control plane. As an example (I don't think we're doing it this way, but:) if the reactor actually applied catalog tasks to Etcd, then it could trivially read-through the applied catalog KeySpace revision before returning from its RPC, gating a future Shard apply RPC, and this race wouldn't be a problem.

go/labels/labels.go
```diff
@@ -44,6 +44,9 @@ const (
 	TaskTypeDerivation = "derivation"
 	// TaskTypeMaterialization is a "materialization" TaskType.
 	TaskTypeMaterialization = "materialization"
+	// CatalogRevision is a base-10 integer of the minimum ETCD revision that must be observed
+	// before this catalog task can be started.
+	CatalogRevision = "estuary.dev/catalog-revision"
```
jgraettinger, line 49: `estuary.dev/task-created` ? "Catalog" isn't a concrete thing (it's CatalogTasks or the CatalogCommons).

go/flow/catalog.go
```diff
@@ -106,6 +106,21 @@ func NewCatalog(ctx context.Context, etcd *clientv3.Client, root string) (Catalo
 	return catalog, nil
 }
 
+// GetTaskWithRevision returns the named CatalogTask, Commons, and its Commons ModRevision. It will
+// first wait for the named task to be at the given revision.
+func (c Catalog) GetTaskWithRevision(ctx context.Context, name string, minRevision int64) (*pf.CatalogTask, *Commons, int64, error) {
+	c.Mu.RLock()
+	defer c.Mu.RUnlock()
+	var task, commons, revision, err = c.getTask(name)
+	if revision < minRevision {
```
jgraettinger, line 115: Just WaitForRevision prior to reading out the task to remove a code path. It's super cheap if you're already >= the revision.

Also, can this be collapsed into `GetTask` ? Its callers, with the exception of ingest, need to pass in create revision to be fully correct w.r.t. data races, so I like that it makes that ordering constraint explicit.  Places where we really don't care (ingest) can pass zero.

go/labels/labels.go
psFried: Yeah I was being intentionally vague with the definition regarding exactly what this revision corresponds to. Technically, the revision is just whatever is returned from running the etcd transaction that applies the catalog tasks _and_ commons, so I didn't want to assume or imply that it would correspond exactly to the revision of the commons or any specific task. So the meaning I was going for is that you need to have observed up through this revision before you can be guaranteed to observe the updated task and commons. That said, I think `task-created` is a perfectly reasonable name, as long as we clarify that it doesn't correspond exactly to the revision of the specific task.

go/flowctl/cmd-apply.go
```diff
@@ -182,46 +182,49 @@ func scopeToPathAndPtr(dir, scope string) (path, ptr string) {
 	return path, ptr
 }
 
-func applyCaptureShards(built *bindings.BuiltCatalog, client pc.ShardClient, shards int) error {
+func applyCaptureShards(built *bindings.BuiltCatalog, client pc.ShardClient, shards int, commonsRevision int64) error {
```
jgraettinger, line 185: nit: `taskRevision` (or `taskCreated` to align w/ label?) would be more accurate. conceivably the `taskRevision` and `commonsRevision` could be different.

go/runtime/flow_consumer.go
```diff
@@ -297,7 +298,12 @@ func shardApply(ctx context.Context, svc *consumer.Service,
 		if nextSpec != nil && logSpec == nil {
 			// Grab labeled catalog task and its journal rules.
 			var name = nextSpec.LabelSet.ValueOf(labels.TaskName)
-			var _, commons, _, err = catalog.GetTask(name)
+			var revisionStr = nextSpec.LabelSet.ValueOf(labels.TaskCreated)
+			var minTaskRevision, err = strconv.ParseInt(revisionStr, 10, 64)
```
jgraettinger, line 302: nit: `taskCreated` instead of `minTaskRevision` ? Can keep the naming consistent. IMO the "revision" suffix isn't adding anything.

go/testing/driver.go
```diff
@@ -192,7 +192,9 @@ func (c *Cluster) Verify(test *pf.TestSpec, testStep int, from, to *Clock) error
 	}
 
 	// Now feed documents into a combiner, filtering documents which are ACKs.
-	task, commons, _, err := c.Consumer.Catalog.GetTask(step.Collection.String())
+	// The waitForRevision parameter is 0 here because it's already guaranteed that the
+	// keyspace has observed the task because WaitForShardsToAssign has already been called.
```
jgraettinger, line 196: nit: Not _technically_ true. The `catalog` vs consumer service keyspace watches process independently, and it's conceivable that the consumer keyspace could be out ahead of the `catalog` keyspace.

I don't think this needs different behavior (for one, this is the same catalog keyspace that the consumer is using, and it's presumably already waited), but the comment should be clarified that we're assuming the race has already been won.

go/flow/catalog.go
```diff
@@ -106,10 +106,13 @@ func NewCatalog(ctx context.Context, etcd *clientv3.Client, root string) (Catalo
 	return catalog, nil
 }
 
-// GetTask returns the named CatalogTask, Commons, and its Commons ModRevision.
-func (c Catalog) GetTask(name string) (_ *pf.CatalogTask, _ *Commons, revision int64, _ error) {
+// GetTask returns the named CatalogTask, Commons, and its Commons ModRevision. It first ensures
+// that the given revision has been observed by the keyspace so that the caller can guarantee that a
+// task update has been observed if its revision is known.
+func (c Catalog) GetTask(ctx context.Context, name string, waitForRevision int64) (_ *pf.CatalogTask, _ *Commons, revision int64, _ error) {
```
jgraettinger, line 112: pragmatically, given that every usage of this either a) doesn't care about the revision, or b) is drawing it from a shard label, should we 1) have the revision be a string argument directly from the label, 2) do str => int conversion here, and 3) pass "" instead of revision 0 to mean "I don't care"?

jgraettinger (approved):
LGTM % a few more comments

go/flow/catalog.go
psFried: yep, good idea. Done

PR #165, opened 2021-07-14
small fixes from testing `flowctl discover` workflow with nested parser configs

jgraettinger:
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/165)

### DIFF
--- a/go/flowctl/cmd-discover.go
+++ b/go/flowctl/cmd-discover.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"sort"
 	"strings"
 	"unicode"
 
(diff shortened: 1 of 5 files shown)


go/flowctl/cmd-discover.go
```diff
@@ -242,14 +243,20 @@ func writeConfigStub(image string, w io.WriteCloser) error {
 
 	var config interface{}
 
+	// Visit leaf-most schema locations first.
+	// Because we're creating yaml.Nodes instead of []interface{}
```
psFried, line 247: Only because I'm curious, but why is that? Seems like a weird restriction

psFried (approved):
LGTM

go/flowctl/cmd-discover.go
jgraettinger: It's because Create() knows how to create children of []interface{} or map[string]interface{}, but doesn't know anything about yaml.Nodes or how to create sub-entities therein.

PR #170, opened 2021-07-22
Set the ResourcePath in webhook driver validation

saterus:
Webhooks use the endpoint address url as the ResourcePath. This means we need to return that ResourcePath from the webhook driver and thoroughly percent-encode it before sending it over to Gazette.

---

This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/170)

### DIFF
--- a/crates/models/src/build/mod.rs
+++ b/crates/models/src/build/mod.rs
@@ -211,11 +211,16 @@ pub fn derivation_spec(
     }
 }
 
-// encode_resource_path encodes path components into a string which
-// is suitable for use within a Gazette token, such as a Journal
-// name or suffix, or a Shard ID. Tokens are restricted to
-// unicode letters and numbers, plus the symbols `-_+/.=%`.
-// All other runes are percent-encoded.
+/// `encode_resource_path` encodes path components into a string which is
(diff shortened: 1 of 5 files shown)


jgraettinger (approved):
LGTM

crates/models/src/build/mod.rs
```diff
@@ -225,7 +225,7 @@ pub fn encode_resource_path(resource_path: &[impl AsRef<str>]) -> String {
     for c in parts.join("/").chars() {
         match c {
             // Note that '%' is not included (it must be escaped).
-            '-' | '_' | '+' | '/' | '.' | '=' => name.push(c),
+            '-' | '_' | '+' | '.' | '=' => name.push(c),
```
jgraettinger, line 228: nit: Add a comment here, discussing that '/' is an allowed path symbol but we don't include it because there are further restrictions on uses of '/' which an arbitrary resource path may violate.

saterus:
@jgraettinger I think this is ready for another look. I updated some affected tests, the docs you mentioned, and the gazette dependency. I've verified it fixes the problem within estuary/demos-segmentation. :tada:

jgraettinger (approved):
LGTM!

PR #171, opened 2021-07-23
fixes to shard stat data-race and flowctl develop hang

jgraettinger:
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/171)

### DIFF
--- a/examples/local-sqlite.flow.yaml
+++ b/examples/local-sqlite.flow.yaml
@@ -2,14 +2,3 @@
 
 import:
   - all-collections.flow.yaml
-
-journalRules:
-  00_fragment_store:
-    template:
-      fragment:
-        compression_codec: GZIP
-        stores:
-          # Default is to persist to the local file system as a stand-in for cloud storage.
-          - file:///
-        # Max time to wait before writing a fragment file
-        flush_interval: 5m
--- a/go.mod
+++ b/go.mod
@@ -33,4 +33,4 @@ require (
 	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b
 )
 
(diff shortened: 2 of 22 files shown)


psFried (approved):
Lots of great fixes in here :lgtm: 

> Reviewed 1 of 1 files at r1, 18 of 18 files at r2, 1 of 1 files at r3, 2 of 2 files at r4.
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/171#-:-MfYcmX23M4e3UVOeIb0:b8cry5b)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @saterus)

examples/local-sqlite.flow.yaml
```diff
@@ -2,14 +2,3 @@
...
-journalRules:
-  00_fragment_store:
-    template:
-      fragment:
-        compression_codec: GZIP
-        stores:
-          # Default is to persist to the local file system as a stand-in for cloud storage.
-          - file:///
-        # Max time to wait before writing a fragment file
-        flush_interval: 5m
```
saterus, line 15: > examples: drop journalRules, in favor of new flowctl develop default

Is this change to the defaults made elsewhere? Or are you saying that the existing defaults should be used rather than these overrides?

In either case, should we remove these `journalRules` from the demo repos?

jgraettinger: The (new in this PR) existing defaults can be used rather than this override. This is in part a test / confirmation that the added defaults work as expected.

I don't think it's necessary to actively remove these from examples. The reason for added default behavior is so that a user can run the `flowctl discover` workflow and get a well-behaved catalog that's ready to run with `flowctl develop` without needing to first add explicit `journalRules`.

PR #176, opened 2021-07-29
Keep stdin open when Flow runs docker containers

psFried:
When Flow is running on a mac using docker-for-mac, containers will
often not produce any output on stdout unless the `--interactive` flag
is provided. I have absolutely no idea why, but it seems to work
reliably with that flag added, and I don't see a down side.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/176)

### DIFF
--- a/go/materialize/driver/jsonimage/driver.go
+++ b/go/materialize/driver/jsonimage/driver.go
@@ -199,12 +199,16 @@ func invokeConnector(
 ) error {
 
 	/*
-		return exec.Command(
-			"docker",
-			"run",
-			"--rm",
-			c.Image,
-		)
+	   return exec.Command(
+	       "docker",
+	       "run",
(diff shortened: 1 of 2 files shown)


go/materialize/driver/jsonimage/driver.go
```diff
@@ -199,12 +199,16 @@ func invokeConnector(
...
-		return exec.Command(
-			"docker",
-			"run",
-			"--rm",
-			c.Image,
-		)
+	   return exec.Command(
+	       "docker",
+	       "run",
+	       // Without the --interactive flag, running connectors with docker-for-mac will sometimes
```
psFried, line 205: I added it here just so we don't forget about it when we come back and work on this code again.

psFried:
Closing this, as it's no longer needed. Turns out the actual issue was caused by the use of socat to proxy to the docker socket.

PR #178, opened 2021-07-30
avoid copies in connector stdout & more tests

jgraettinger:
Add tests for edge cases of parsing JSONL records from connectors.
No bugs were found in the original implementation.

Then improve parsing to avoid excessive copies of the written buffer.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/178)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -160,7 +160,7 @@ install-tools: ${TOOLBIN}/protoc-gen-gogo ${TOOLBIN}/etcd
 
 .PHONY: rust-build
 rust-build:
-	FLOW_VERSION=${VERSION} cargo build --release --locked -p bindings
+	FLOW_VERSION=${VERSION} cargo build --release --locked
 
 .PHONY: rust-test
 rust-test:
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -160,7 +160,7 @@ install-tools: ${TOOLBIN}/protoc-gen-gogo ${TOOLBIN}/etcd
 
 .PHONY: rust-build
 rust-build:
-	FLOW_VERSION=${VERSION} cargo build --release --locked -p bindings
+	FLOW_VERSION=${VERSION} cargo build --release --locked -p bindings -p librocks-exp
```
saterus, line 163: Curious: Why are we only building these two crates as part of the `rust-build`? I assume the other crates get built by being a dependency of the `bindings` crate, but wouldn't the default of building the entire workspace suffice?

jgraettinger: Yes, I think it would. I honestly don't recall why it is this way... will update

Issue #181, opened 2021-07-30
Updates for AWS S3 Connector Defaults


dyaffe:
We need to make a few updates to the AWS S3 connector to make it work out of the box:

1.  Amazon expects ascendingKeys: false vs ""
2. awsAccessKeyId and awsSecretAccessKey should not be required for public datasets
3. Add an example of "us-east-1" to region to make it obvious what most would be

psFried:
This is resolved by #184 and (https://github.com/estuary/connectors/pull/18)

PR #184, opened 2021-08-03
Issue 181

psFried:
This fixes the first point in issue #181  so that we now render `ascendingKeys` as a boolean instead of a string. Of course this applies generically to all typed fields from the connector spec.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/184)

### DIFF
--- a/go/flowctl/cmd-discover.go
+++ b/go/flowctl/cmd-discover.go
@@ -259,7 +259,8 @@ func writeConfigStub(ctx context.Context, image string, w io.WriteCloser) error 
 		} else if *node == nil {
 
 			var nn = new(yaml.Node)
-			nn.Style = yaml.FoldedStyle
+			nn.Style = yaml.FlowStyle // renders values inline with the keys instead of on the next line
+			nn.Kind = yaml.ScalarNode // seems required to get numbers and bools to render correctly
 
 			nn.FootComment =
 				fmt.Sprintf("%s\n%s", loc.Spec.Description, loc.Spec.Types)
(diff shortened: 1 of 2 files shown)


go/flowctl/cmd-discover.go
```diff
@@ -259,15 +259,38 @@ func writeConfigStub(ctx context.Context, image string, w io.WriteCloser) error
...
-			nn.Style = yaml.FoldedStyle
+			nn.Style = yaml.FlowStyle // renders values inline with the keys instead of on the next line
+			nn.Kind = yaml.ScalarNode // seems required to get numbers and bools to render correctly
 
 			nn.FootComment =
 				fmt.Sprintf("%s\n%s", loc.Spec.Description, loc.Spec.Types)
 
 			if loc.Spec.MustExist {
 				nn.FootComment += " (required)"
 			}
```
saterus, line 270: This is definitely outside the scope of this PR, but it looks like we could render the comments *above* each line, rather than below by using [`Node.HeadComment`](https://pkg.go.dev/gopkg.in/yaml.v3#Node) rather than `FootComment`. It's minor, but it was something that felt rather odd when I first saw it. If we're in here making these generated catalog files make a little more sense, it might be worth doing.

go/flowctl/cmd-discover.go
```diff
@@ -286,6 +309,20 @@ func writeConfigStub(ctx context.Context, image string, w io.WriteCloser) error
 	return nil
 }
 
+// getDefaultType returns the type to use for generating a default value for endpoint configuration.
+// It will always prefer a scalar type if a location allows multiple types.
+func getDefaultType(inference *pf.Inference) string {
```
saterus, line 314: Is this effectively picking `"string"` from `["string", "null"]`? Or do we have cases where options are `["object", "string"]` messes?

saterus (approved):
This looks pretty good. It was an interesting read through a part of the system I've not yet seen.

I ran `flowctl discover` with a variety of our connectors and it appears to do the trick. :100:

go/flowctl/cmd-discover.go
psFried: Yeah, TBH that's my preference as well, but I can also see the appeal of putting it below for users who may not be using an editor with syntax highlighting. I also took an informal survey of one (@dyaffe ), and the response was that comments below felt easier to read. So ... :man_shrugging:

go/flowctl/cmd-discover.go
psFried: I think in practice pulling `"string"` from `["string", "null"]`, but `source-gcs` does have a `googleCredentials` key that takes an object (gcp service account keys are JSON objects).

PR #185, opened 2021-08-05
Revert rocksdb compression back to snappy

### LINKED ISSUE #183: Switch back to snappy compression for derivation registers
Ok, here's a write up of what's going on with rocksdb compression. After #172 we only link LZ4, which is the recommended compression library according to the [rocks docs](https://www.bookstack.cn/read/rocksdb-en/91c87d45b466f901.md). Snappy is the default in rocks, though, and the default behavior is for rocks to use no compression if libsnappy is not available. So I think at this point, we're basically just not using compression at all for registers, since just linking to lz4 isn't enough to make rocksdb use it.
Prior to #172 registers were all using snappy compression. After #172 new derivations are not using compression, and existing derivations are failing to initialize because libsnappy is no longer available and so the rocks database can't be opened.

The decision is to switch back to using only Snappy compression, at least for the short term. If we want to switch to LZ4, then we should create a separate issue for doing that, which also addresses the migration path for existing derivations.

psFried:
resolves #183

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/185)


saterus (approved):
Seems reasonable to me.

There's no Gazette-side configuration of Rocks that we need to do, right? It's already just defaulting to using snappy?

psFried:
> There's no Gazette-side configuration of Rocks that we need to do, right? It's already just defaulting to using snappy?

Right, the default behavior of rocks is to use snappy if it's available, or else fall back to no compression, and that was never actually changed as part of the move to zstd. The code for actually opening the rocks database is [here](https://github.com/estuary/flow/blob/master/crates/derive/src/derive_api.rs#L79).

One thing I forgot to mention in the comment is that, at least on my machine, if I run `nm target/release/libbindings.a` I still see `lz4.o` being linked in. I had thought it was just due to old artifacts in the build directory, but in tracking it down I realized that I actually missed a few other places where the rocksdb crate features had specified lz4. So that should hopefully be fixed in the next commit.

PR #188, opened 2021-08-15
devcontainer: update for 20.04 & Codespaces

jgraettinger:
Update to a Ubuntu 20.04 devcontainer base, with host Docker support which (🤞 ) works on Mac Docker as well as Codespaces.

I've confirmed it works fine on Codespaces.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/188)

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -38,7 +38,7 @@
 		// Protobuf language support.
 		"zxh404.vscode-proto3",
 		// Formatting support (eg, protobuf).
-		"xaver.clang-format"
+		"xaver.clang-format",
 		// Spell checking.
 		"streetsidesoftware.code-spell-checker"
 	],
@@ -53,6 +53,7 @@
 	// We're using a base image which is compatible with our CI runner,
 	// and thus use the same make target.
 	"postCreateCommand": "make extra-ci-runner-setup && go mod download",
-	// Uncomment to run as a non-root user.
-	// "remoteUser": "vscode"
+	// This user matches the user that's created inside the flow image
+	"remoteUser": "flow",
+	"overrideCommand": false
 }
--- a/.devcontainer/docker-compose.yaml
+++ b/.devcontainer/docker-compose.yaml
@@ -1,17 +1,27 @@
-version: "3"
+version: "3.7"
 
 services:
   develop:
     build:
       context: .
-      dockerfile: github-ubuntu-1804.Dockerfile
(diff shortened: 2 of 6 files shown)


psFried (approved):
LGTM

saterus:
I booted up a Codespace and it worked! :tada: 

It is popping up dialogs for installing `gopls` and the Docker extension for VSCode. It might be worth doing just to avoid the popups each time.

This might be a misplaced expectation on my part, but should we actually _build_ `flowctl` inside the Docker image so that it's available when you boot? I know this isn't the release container, but it takes 10+ minutes to build the project within the Codespace (anecdotally, this appears to mostly be the Rocks build). I'm not sure if Docker layer caching would work well for this, but at least you'd end up with an environment where Rocks is already compiled once the Codespace is booted.

We may also want to consider adding `.build/package/bin` to the path inside the container. It's not essential, but it's handy.

Are there similar changes we should make for the `flow-template` and `demos-*` repos so they also work on Codespaces? Is there a reasonable way to share some of these things as a base Dockerfile?

jgraettinger:
> I booted up a Codespace and it worked! 🎉

Awesome!

> It is popping up dialogs for installing gopls and the Docker extension for VSCode. It might be worth doing just to avoid the popups each time.

I think the `gopls` one is unavoidable (it comes from the Go extension). We could do the Docker extension but I do groan a bit at prescribing it. Both are one-time popups at the creation of a new codespace.

> This might be a misplaced expectation on my part, but should we actually build flowctl inside the Docker image so that it's available when you boot? I know this isn't the release container, but it takes 10+ minutes to build the project within the Codespace (anecdotally, this appears to mostly be the Rocks build). I'm not sure if Docker layer caching would work well for this, but at least you'd end up with an environment where Rocks is already compiled once the Codespace is booted.

Totally, it should run the full build at bootup, so that you can create a new codespace, grab a coffee, and be fully ready to go when you're back. Resumptions of a codespace (once all of the initial setup is done) is quite fast.

> We may also want to consider adding .build/package/bin to the path inside the container. It's not essential, but it's handy.

✔️ 

> Are there similar changes we should make for the flow-template and demos-* repos so they also work on Codespaces? Is there a reasonable way to share some of these things as a base Dockerfile?

They more or less already have it -- in fact some of these changes are back-ported from `flow-template`. Re: sharing via base Dockerfile ... maybe? Unclear to me right now how divergent they are or will be.

PR #195, opened 2021-08-16
WIP: Launch connectors on host network

saterus:
This may turn out to be temporary.

Opening a draft PR so I can build the docker image for my local testing.


go/capture/driver/airbyte/connector.go
```diff
@@ -53,6 +54,12 @@ func RunConnector(
 		"run",
 		"--rm",
 	}
+
+	if permissiveConnectorNetworking {
+		imageArgs = append(imageArgs, "--network=host")
+		log.Infof("Granting host network access to %s container", image)
```
jgraettinger, line 60: nit: this log seems redundant / abstracted vs the INFO logging of the connector arguments below

go/capture/driver/airbyte/driver.go
```diff
@@ -57,12 +57,16 @@ func (c ResourceSpec) Validate() error {
 }
 
 // driver implements the pm.DriverServer interface.
-type driver struct{}
+type driver struct {
+	PermissiveConnectorNetworking bool
```
jgraettinger, line 61: nit: `hostNetwork` ? This doesn't need to be exported, no?

go/bindings/build.go
```diff
@@ -122,7 +130,7 @@ func BuildCatalog(args BuildArgs) (*BuiltCatalog, error) {
 				log.WithField("request", request).Debug("capture validation requested")
 
 				var driver, err = args.CaptureDriverFn(ctx, request.EndpointType,
-					json.RawMessage(request.EndpointSpecJson), "")
+					json.RawMessage(request.EndpointSpecJson), "", permissiveConnectorNetworking)
```
jgraettinger, line 133: This should be part of BuildArgs rather than a const.

go/bindings/build.go
```diff
@@ -64,6 +65,13 @@ type BuildArgs struct {
 	MaterializeDriverFn MaterializeDriverFn
 }
 
+// We currently run all validations of a catalog spec from a developer's local
```
jgraettinger, line 68: I think this should be inverted: we should model a concept of the current build/run context, probably as a protobuf enum (name: `pf.Invocation` ? ). We'd then thread that through and make a decision about host networking at the last possible mile, within the guts of the docker connector implementation.

That enum definition can be very literal & tightly coupled to the specific modes of operations we have now ("TEST" vs "DEVELOP" vs "SERVE" for now, perhaps), rather than those we hope to have. Part of the benefit of this factoring is that, when we add / remove modes of operation, those become compile-time errors which are easy to root out & update.

PR #197, opened 2021-08-17
Validate that top level collection schema is an object

### LINKED ISSUE #127: Build should error when the top level collection schema is not `type: object`
This example currently builds successfully:

```
collections:
  foo:
    schema: {type: string}
    key: ['']
```

This ought to be a build error, since the top level of the schema must be an object in order to be able to inject the `_meta` property. Currently, ingesting a record to this collection will cause flow-ingester to panic.

willdonnelly:
Resolves #127

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/197)

### DIFF
--- a/crates/validation/src/collection.rs
+++ b/crates/validation/src/collection.rs
@@ -68,6 +68,13 @@ fn walk_collection(
     let schema = schema_shapes.iter().find(|s| s.schema == *schema).unwrap();
     let _ = schema::walk_composite_key(scope, key, schema, errors);
 
+    if schema.shape.type_ != types::OBJECT {
+        Error::CollectionSchemaNotObject {
+            collection: name.to_string(),
+        }
+        .push(scope, errors);
+    }
+
     let projections = walk_collection_projections(
         collection,
         projections,
--- a/crates/validation/src/errors.rs
+++ b/crates/validation/src/errors.rs
@@ -49,6 +49,8 @@ pub enum Error {
     NoSuchSchema { schema: Url },
     #[error("collection {collection} key cannot be empty")]
     CollectionKeyEmpty { collection: String },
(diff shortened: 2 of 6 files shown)


willdonnelly:
As seen above in the failing CI run, the trivial implementation of this check causes a couple of preexisting tests to fail, specifically `schema_reference_verification` and `schema_fragment_not_found`. I haven't finished understanding exactly what the problem is (specifically, I don't yet know whether it's due to the references themselves or the fact that some references in each test are deliberately invalid), but in both cases the issue is that the top-level schema type is `ANY`.

A trivial fix would be changing the condition for the error to `schema.shape.type_ & types::OBJECT != types::OBJECT`, which would for instance cause `{}` to be a valid schema, but that's probably the wrong approach. Probably need to investigate this some more to come up with a fix.

willdonnelly:
I spent some time digging into the inference logic and now I understand why those two tests are breaking, and it seems like as currently written these tests _should_ be failing the new validation check. So I think the correct solution here is simply to update the "correct" `.snap` files with the new error.

In more detail: the root of the problem with `schema_reference_verification` is that `test://example/int-string-len.schema` is patched to make the `$ref` invalid, but that reference was the only thing narrowing the inferred type from `types::ANY` to `types::OBJECT` (at `inference.rs:748`). And when that reference is invalid, the default shape is used instead (`inference.rs:738`), so the intersection doesn't restrict things any. Likewise in `schema_fragment_not_found` the `testing/int-string` collection is modified to make the `schema:` link invalid, so the collection defaults to `types::ANY` and fails the new validation check.

jgraettinger (approved):
Yep, your analysis is right on, LGTM

PR #198, opened 2021-08-17
Enable debug info in dev builds

willdonnelly:
This should make debugging more consistent. Without this change it was impossible to (for example, on my machine) place a breakpoint in the file `collection.rs`.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/198)


jgraettinger (approved):
LGTM

PR #200, opened 2021-08-18
downgrade from Go 1.16.7 => 1.16.6

### LINKED ISSUE #191: consumer transaction timers are sometimes delayed in Go 1.16.7
Trialing Codespaces, I'm seeing `make catalog-test` complete successfully, but only after multiple long (~minute) stalls between many tests. I've debugged enough to isolate that the stalls are due to shard Stat RPCs which are awaiting a future read-through offset, but don't yet know _why_ it's delayed.

jgraettinger:
Also have Codespaces build the entire project on first initialization.

Fixes #191

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/200)

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -52,7 +52,7 @@
 	// Run extra environment setup required to build & test Flow.
 	// We're using a base image which is compatible with our CI runner,
 	// and thus use the same make target.
-	"postCreateCommand": "make extra-ci-runner-setup && go mod download",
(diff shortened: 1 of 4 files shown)


saterus (approved):
Double checked this with Codespaces. It works. 

I'll admit that the `postCreateCommand` is a bit of a mixed bag. It eventually works, but it doesn't show any logs while it builds. It just shows that it's running that command...for what feels like forever. :shrug:

Issue #173, opened 2021-07-28
materializations: extract from runtime into connectors


jgraettinger:
Current materializations (webhook, postgres, snowflake, sqlite) are directly built into the runtime.
We'd like to extract webhook, snowflake, and _probably_ postgres (but **not** sqlite) into Docker images which live outside of the estuary/flow codebase.

There's a stubbed-out implementation of a dockerized connector which uses a text-based protocol. While that protocol is useful, and is something we want to document and maintain for _other_ people's connectors, there's a lot of existing smarts within our connectors that directly leverage the gRPC materialization protocol, and let us do things like optimistic transaction pipelining without a lot of implementation overhead.

Ideally, the docker image would tell us how to drive it. Either by:
  * Talking over stdin/stdout using the text based protocol (probably the default), or
  * Where supported by the connector, directly talking to it over gRPC.

We could use docker image metadata for this, such as labels.
Or perhaps it should be part of an as-yet-undefined configuration spec workflow ?

jgraettinger:
Done

PR #207, opened 2021-08-24
capture/materialize: extend and unify Spec RPC semantics

jgraettinger:
Introduce resource JSON schemas to the capture Spec RPC,
alongside endpoint config JSON schemas.

Add an equivalent Spec RPC to the materialize protocol.

Implement the new semantics using jsonschema for automatic generation
of suitable JSON schemas from our existing stable of connectors
and drivers.

For Snowflake, add an indirection between the parsed configuration
and the snowflake.Config object.

Issue #173


go/protocols/capture/capture.proto
```diff
@@ -37,16 +38,23 @@ message SpecRequest {
...
-  string spec_schema_json = 1 [
+  string endpoint_spec_schema_json = 1 [
+    (gogoproto.casttype) = "encoding/json.RawMessage",
+    json_name = "endpointchema"
+  ];
+  // JSON schema of a complete endpoint specification.
+  string resource_spec_schema_json = 2 [
     (gogoproto.casttype) = "encoding/json.RawMessage",
     json_name = "endpointchema"
   ];
```
saterus, line 52: Where did we land on the semantics of using this resource spec? Is that still TBD?

Typos:
- The resource spec comment mentions the endpoint.
- I'm guessing we wanted `json_name = "endpointschema"`, rather than "endpointchema".

go/protocols/materialize/materialize.proto
```diff
@@ -84,6 +87,36 @@ service Driver {
...
+message SpecResponse {
+  // JSON schema of a complete endpoint specification.
+  string endpoint_spec_schema_json = 1 [
+    (gogoproto.casttype) = "encoding/json.RawMessage",
+    json_name = "endpointchema"
+  ];
+  // JSON schema of a complete endpoint specification.
+  string resource_spec_schema_json = 2 [
+    (gogoproto.casttype) = "encoding/json.RawMessage",
+    json_name = "endpointchema"
```
saterus, line 114: Similar typos here as with `capture.proto`.

saterus (commented):
I'm reviewing just for my own knowledge/familiarity/understanding. I'll defer to Phil on the final review.

General question: Did I miss where in the code we actually validate the resource spec provided against the schema defined by the connector? For either the materialization or the capture. Is this part still a TODO?

go/protocols/capture/capture.proto
jgraettinger: > Where did we land on the semantics of using this resource spec? Is that still TBD?

The `airbyte` driver does supply a schema for the resource spec, which is generated from the type it's parsed into.

There's no (current) schema validation of the resource spec. It is _pretty_ well covered by doing actual strict parsing of the resource specs as part of Validate RPCs though, for both capture and materialize. The principle motivation for adding resource specs is to make it possible for future UIs and tooling to understand and present the available configuration for a resource binding, so this is pretty forward-looking and done from a sense of wanting to carve out & codify a position on how it _should_ work.

| Typos

Doh! How about I address this in the `protocols` repo, to which these are moving ?

go/protocols/materialize/materialize.proto
jgraettinger: Thanks. Same story -- I'll address in `protocols` repo

jgraettinger:
> Did I miss where in the code we actually validate the resource spec provided against the schema defined by the connector? For either the materialization or the capture. Is this part still a TODO?

As per the other comment, we're not validating against the JSON schema directly, but we are implicitly validating it by doing strict deserialization into the same types from which the schemas are generated.

PR #213, opened 2021-08-31
Provide flowctl.sh script to invoke flowctl inside of a container for development

snowzach:
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/213)


snowzach:
This is mostly working for Linux/WSL but still having issues with MacOS. `/var/run/docker.sock` outside of the container is a symlink to a docker.sock in the user's home directory. That docker.sock has an owner of `0:20` the script properly tracks down the correct file and adds the flow user to group 20 but the sock is owned by root `0:0` once in the container. I am not sure if MacOS docker is mapping in from an even higher up container or what. Other than that it does appear to be working using QEMU emulation on M1 Macs. I am working on scheduling some keyboard time on a Mac to do a little more debugging.

snowzach:
The MacOS issue is related to this issue https://github.com/docker/for-mac/issues/4755

snowzach:
After messing around with this for way too long, I think the easiest/most sane solution is to run the container as the root user when MacOS/Darwin is detected. 

Normally this would be a bad idea, especially from a filesystem standpoint but the way in which MacOS Docker handles mapping directories into the container, even though it's running as root, any files created end up being owned by the UID of the running user on the outer MacOS. 

For simple local machine development I think this is the easiest and most straight forward way to manage this process. We could always setup sudo+socat but I'm not sure it buys us anything.

Just for the sake of not figuring it out again in case we decide to try the sudo+socat method...

Dockerfile
```
FROM quay.io/estuary/flow:dev

USER root

RUN apt-get update -y \
 && apt-get upgrade -y \
 && DEBIAN_FRONTEND=noninteractive apt-get install --no-install-recommends -y sudo socat

RUN echo 'flow ALL=(root) NOPASSWD: /usr/bin/socat -t 3600 UNIX-LISTEN\:/var/run/docker.sock.user\,fork\,mode\=660\,user\=* UNIX-CONNECT\:/var/run/docker.sock' > /etc/sudoers.d/docker-entrypoint \
    && chmod 0440 /etc/sudoers.d/docker-entrypoint

ADD docker-entrypoint.sh /usr/local/bin/docker-entrypoint.sh

USER flow

ENTRYPOINT [ "/usr/local/bin/docker-entrypoint.sh" ]
CMD [ "/bin/bash" ]
```

And a docker-entrypoint.sh script
```
#!/bin/sh

echo "Running script..."

# This is a workaround to help resolve an issue with Docker Desktop on Mac where you cannot map the docker.sock into the container and run as non-root: https://github.com/docker/for-mac/issues/4755
# If the current user cannot write /var/run/docker.sock start socat using sudo proxying /var/run/docker.sock.user -> /var/run/docker.sock and grant permissions to the current user
if [ ! -w /var/run/docker.sock ]; then
    echo "Setting up docker.sock proxy"
    sudo socat -t 3600 UNIX-LISTEN:/var/run/docker.sock.user,fork,mode=660,user=$UID UNIX-CONNECT:/var/run/docker.sock 2>&1 &
    export DOCKER_HOST=unix:///var/run/docker.sock.user
fi

# Start the command passed in
$@
```

scripts/flowctl.sh
```diff
@@ -0,0 +1,208 @@
...
+fi
+
+# Build docker command
+CMD="${DOCKER_EXEC} run -it --rm \
+--user ${DOCKER_UID} \
+-v ${FLOWCTL_DIRECTORY}:${FLOWCTL_CONTAINAER_DIRECTORY} \
+-v ${DOCKER_SOCK}:/var/run/docker.sock \
+--network ${FLOWCTL_NETWORK} \
+${DOCKER_EXTRA_OPTS} \
+-v /var/tmp:/var/tmp -e TMPDIR=/var/tmp \
```
psFried, line 193: It might be worth adding a comment here about why the tempdir needs to be the same path on the host and inside the container, since it's caused some confusion in the past.

psFried (changes requested):
I think this looks great. I left a few minor comments inline, but I think it seems good to merge after that. I tested this out locally, and it seems to also work on my machine.

scripts/flowctl.sh
snowzach: Sure thing. I added explanation of the user as well.

scripts/flowctl.sh
```diff
@@ -0,0 +1,208 @@
+#!/bin/bash
+
+function log_fatal {
+    echo "ERROR: $1"
+    exit 1
+}
+
+# Make sure we have docker
```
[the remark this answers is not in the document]
snowzach: Sounds good to me. I'm no bash expert.

PR #218, opened 2021-09-01
Publish Docker images to ghcr.io

saterus:
We'd like to consolidate on using GitHub's container registry.

We still reference the quay.io hosted images everywhere, so we can't reasonably remove that all at once. So we'll start by publishing the images to both registries, then we can update the quay.io references to use ghcr.io instead, then we can sunset quay.io entirely.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/218)


jgraettinger (approved):
LGTM

PR #219, opened 2021-09-02
Make the default target to build the rust side and then the go side

snowzach:
I struggled a little bit at first to figure out how to build flowctl.. And then realized today that if the rust side changes it must be built beforehand. This just adds a default target and builds them both.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/219)


psFried (approved):
Looks good to me! Merge away

PR #221, opened 2021-09-08
runtime support for materialization connectors

jgraettinger:
See individual commits.

Issue #173

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/221)


jgraettinger:
Any feedback here?

snowzach:
LGTM... My only questions are around using the airbyte.RunConnector. Since these aren't *totally* airbyte connectors do we want to make a more generic connector controller abstraction. In addition, do we want to actually use the docker sdk vs running a cmd docker instance and plumbing it up. It may or may not simplify things a bit.

Otherwise, I see no problem merging this and I'd be happy to start excersizing it with the BigQuery connector. As well, if you see a benefit to using the docker sdk, I'm happy to take a look at that as well.

dyaffe:
Sounds like a good opportunity to rename it from airbyte :)

jgraettinger:
> but I don't think it's worth worrying about now

To clarify, that performance benefit might be significant (or may not exist at all, IDK). I don't mean to shut down exploring it -- part of my bias is wanting to keep docker at arms length because I believe different runtimes in the intermediate future are likely (eg firecracker)

snowzach (approved):
LGTM per previous comments

jgraettinger:
Thanks! Sorry for the delay, working on merging now.

PR #227, opened 2021-09-09
capture: defer a combiner.Destroy

jgraettinger:
To ensure the CGO service is destroyed during garden-path processing.
Current suspicion is that buffers allocated by the CGO service (into
which combined results are marshalled) are what are leaking.

Technically they have a finalizer, but if Go memory pressure doesn't
require a GC they may never actually be reclaimed.

Issue #226

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/227)


psFried (approved):
LGTM

PR #229, opened 2021-09-10
fix two teardown-related edge case bugs in shuffled reads

jgraettinger:
Issue #226

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/229)


psFried (approved):
Tricky stuff with that shuffle read error, but I think the fix looks good.

PR #232, opened 2021-09-12
catalog builds: table ordering refactor, and improved implicit projections

### LINKED ISSUE #231: inference: missing implicit projections for collection key within `additionalItems` array
When using a schema that composes `additionalItems` and `minItems`, where the collection key references an index < minItems, we fail to generate a projection for the collection's key. This fails later, during validation of the CollectionSpec.

For example given collection schema:
```yaml
type: object
properties:
  arr:
    type: array
    additionalItems: {type: integer}
    minItems: 2
required: [arr]
```
And collection key `[/arr/1]`, we (correctly) allow the collection key because we infer its required existence and type.
But we don't create an implicit projection having this location. Currently, downstream bits (and the Go `flow.CollectionSpec` implementation) require that it do exist (and it seems pretty reasonable that it would).

jgraettinger:
Definitely review this commit by commit. They tell the story of how this change set came together.

It was a bit of a rabbit hole to fix a seemingly minor bug. I found myself really, really wanting reliable guarantees that tables were ordered by their natural keys, that could be leveraged for lookups throughout the validation codebase. It's a little more complicated than simply slapping a BTreeSet into the mix, because our tables are actually _multi_sets. We want to represent multiple instances of various entities, so that we can generate good catalog build errors as we walk those duplicates.

Fixes #231

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/232)

### DIFF
--- a/crates/models/src/tables/macros.rs
+++ b/crates/models/src/tables/macros.rs
@@ -231,7 +231,7 @@ macro_rules! replace_expr {
 /// Define row & table structures and related implementations.
 macro_rules! tables {
     ($(
-        table $table:ident ( row $row:ident, sql $sql_name:literal ) {
+        table $table:ident ( row $row:ident, order_by [ $($order_by:ident)* ], sql $sql_name:literal ) {
             $($field:ident: $rust_type:ty,)*
         }
     )*) => {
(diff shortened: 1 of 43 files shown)


crates/models/src/tables/macros.rs
```diff
@@ -318,13 +348,15 @@ macro_rules! tables {
                 let mut stmt = db.prepare(&Self::select_sql())?;
                 self.0.extend(stmt.query_map([], $row::scan)?
                               .collect::<Result<Vec<_>, _>>()?);
+                self.reindex();
```
saterus, line 351: I think you can replace `self.0.extend` + `self.reindex` with `self.extend` now, right?

crates/models/src/tables/macros.rs
```diff
@@ -318,13 +348,15 @@ macro_rules! tables {
...
                 Ok(())
             }
 
             fn load_where(&mut self, db: &rusqlite::Connection, filter: &str, params: &[&dyn rusqlite::types::ToSql]) -> rusqlite::Result<()> {
                 let mut stmt = db.prepare(&format!("{} WHERE {}", Self::select_sql(), filter))?;
                 self.0.extend(stmt.query_map(params, $row::scan)?
                               .collect::<Result<Vec<_>, _>>()?);
+                self.reindex();
```
saterus, line 359: Ditto.

saterus (approved):
After spending a lot of time reading through the catalog build process, I understand the problem and why this fixes it. :+1: 

The refactorings seem like a win too. After looking into `superslice`, I understand why you were keen to have the tables maintain a sorted ordering.

crates/models/src/tables/macros.rs
jgraettinger: Done.

crates/models/src/tables/macros.rs
jgraettinger: Done

PR #233, opened 2021-09-13
Phil/rust metrics

psFried:
Adds a bunch of instrumentation around memory allocations in Rust. At this point, these are all generic metrics. Stats that can be scoped to a particular catalog task will be exposed separately in a later PR as part of the derive, combine, and extract services.

Relates to #226

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/233)

### DIFF
--- a/crates/derive/src/combine_api.rs
+++ b/crates/derive/src/combine_api.rs
@@ -1,4 +1,5 @@
 use super::combiner::{self, Combiner};
+use allocator::ThreadStatsReader;
 
 use doc::{Pointer, Validator};
 use prost::Message;
@@ -66,8 +67,10 @@ impl cgo::Service for API {
             None => return Err(Error::InvalidState),
         };
         tracing::trace!(?code, "invoke");
+        let mem_stats = ThreadStatsReader::new();
+        let initial = mem_stats.current();
 
-        match (code, std::mem::take(&mut self.state)) {
+        let result = match (code, std::mem::take(&mut self.state)) {
             (Code::Configure, _) => {
                 let combine_api::Config {
                     schema_index_memptr,
@@ -128,7 +131,11 @@ impl cgo::Service for API {
                 Ok(())
             }
             _ => Err(Error::InvalidState),
-        }
+        };
+
+        let diff = mem_stats.current() - initial;
+        tracing::trace!(mem_initial = ?initial, mem_changes = ?diff, "mem changes");
+        result
     }
 }
 
@@ -140,7 +147,11 @@ pub fn drain_combiner(
     out: &mut Vec<cgo::Out>,
 ) {
     let key_ptrs = combiner.key().clone();
-
+    tracing::debug!(
+        arenaLen = ?arena.len(),
+        nDocs = ?combiner.len(),
+        "drain_combiner",
+    );
     for (doc, fully_reduced) in combiner.drain_entries(uuid_placeholder_ptr) {
         // Send serialized document.
         let begin = arena.len();
--- a/crates/derive/src/combiner.rs
+++ b/crates/derive/src/combiner.rs
@@ -65,6 +65,10 @@ impl Combiner {
         }
     }
 
+    pub fn len(&self) -> usize {
+        self.entries.len()
+    }
+
     /// Reduce the fully reduced left-hand document with a partially reduced right-hand
     /// document that's already in the Combiner. It's an error if there is already a fully
     /// reduced right-hand document.
--- a/go/bindings/derive.go
+++ b/go/bindings/derive.go
@@ -306,6 +306,7 @@ func (d *Derive) Destroy() {
 
 func newDeriveSvc() *service {
 	return newService(
+		"derive",
 		func() *C.Channel { return C.derive_create() },
 		func(ch *C.Channel, in C.In1) { C.derive_invoke1(ch, in) },
 		func(ch *C.Channel, in C.In4) { C.derive_invoke4(ch, in) },
--- a/go/bindings/metrics.go
+++ b/go/bindings/metrics.go
@@ -1,0 +1,109 @@
+package bindings
+
+/*
+#include "../../crates/bindings/flow_bindings.h"
+*/
+import "C"
+
+import (
+	"sync"
+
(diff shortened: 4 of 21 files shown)


crates/derive/src/combine_api.rs
```diff
@@ -140,7 +147,11 @@ pub fn drain_combiner(
     out: &mut Vec<cgo::Out>,
 ) {
     let key_ptrs = combiner.key().clone();
-
+    tracing::debug!(
+        "drain_combiner: arena.len(): {}, nEntries: {}",
```
jgraettinger, line 151: nit: structured output, like

tracing::debug!(arena = ?arena.len(), docs = ?combiner.doc_count(), "draining combiner")

crates/derive/src/combiner.rs
```diff
@@ -65,6 +65,10 @@ impl Combiner {
         }
     }
 
+    pub fn doc_count(&self) -> usize {
```
jgraettinger, line 68: nit: `len` does seem an appropriate name here ? Since a combiner is-a collection

crates/derive/src/lib.rs
```diff
@@ -9,6 +9,20 @@ mod registers;
 
 pub use extract_api::extract_uuid_parts;
 
+/*
```
jgraettinger, line 12: ?

crates/metrics/src/lib.rs
```diff
@@ -0,0 +1,110 @@
+use jemalloc_ctl::{epoch, epoch_mib, stats, thread};
```
jgraettinger, line 1: Setting the global allocator seems like it should be in the `bindings` crate.

As it is, it adds a jemalloc dependency on the `derive` crate.

crates/metrics/src/lib.rs
```diff
@@ -0,0 +1,110 @@
...
+static ALLOC: jemallocator::Jemalloc = jemallocator::Jemalloc;
+
+/// Statistics related to memory allocations for the entire (rust portion) of the application. The
+/// precise meaning of each field is included in the [jemalloc man
+/// page](http://jemalloc.net/jemalloc.3.html). Each field in this struct can be found in the man
+/// page prefixed by "stats.".
+#[derive(Debug)]
+#[repr(C)]
+pub struct GlobalMemoryStats {
+    pub active: u64,
```
jgraettinger, line 13: I'd expect this to be separately tracking allocated _bytes_ vs allocated _objects_. Both have interesting costs.

Maybe it is already, but it's not clear to me from the naming. And if it's not (currently) possible to have both, I think the names should still clarify which is measured.

go/bindings/derive.go
```diff
@@ -50,7 +50,7 @@ type Derive struct {
 }
 
 // NewDerive instantiates the derivation API using the RocksDB environment and local directory.
-func NewDerive(recorder *recoverylog.Recorder, localDir string) (*Derive, error) {
+func NewDerive(recorder *recoverylog.Recorder, localDir string, shardFqn string) (*Derive, error) {
```
jgraettinger, line 53: ? not used here. This gets plumbed through already later IIRC, in Configure

go/bindings/metrics.go
```diff
@@ -0,0 +1,84 @@
...
+
+// RegisterPrometheusCollector initializes prometheus metric collection that's pulled from the rust
+// libbindings. Subsequent calls to this function will panic.
+func RegisterPrometheusCollector() {
+	prometheus.MustRegister(promCollectorInstance)
+}
+
+var (
+	activeBytesDesc = prometheus.NewDesc(
+		"flow_rust_mem_active_bytes",
```
jgraettinger, line 55: Just confirming, do these names align with the Go memory stats ?

jgraettinger (commented):
Looks awesome, will be very helpful. Some comments below.

crates/metrics/src/lib.rs
psFried: jemalloc seems to only expose stats for bytes, not actual counts of allocations. We could get both by further instrumenting the jemalloc allocator using the `stats_alloc` crate to track individual calls to alloc and dealloc, but I wasn't really sure if that was worth it. It'd be easy to add if you think it's worth the runtime overhead (basically just an atomic increment for every call).

As far as the names go, the names that are there are meant to match the names used by jemalloc in their docs. IMO it's better to use their names exactly, since these are all very specific to jemalloc, and are already well documented.

go/bindings/metrics.go
psFried: No. Jemalloc exposes stats that are kind of specific to jemalloc, so I tried to keep naming of these metrics fairly consistent with the names used by the `stats._` properties from the [jemalloc man page](http://jemalloc.net/jemalloc.3.html). These are the same names used in the `GlobalMemoryStats` struct. I don't honestly see a lot of benefit to having these match the go metric names, since they're mostly tracking different things.

crates/derive/src/combine_api.rs
```diff
@@ -128,7 +131,11 @@ impl cgo::Service for API {
                 Ok(())
             }
             _ => Err(Error::InvalidState),
-        }
+        };
+
+        let diff = mem_stats.current() - initial;
+        tracing::trace!(mem_initial = ?initial, mem_changes = ?diff, "mem changes");
```
saterus, line 137: Just curious about our current thinking: are we planning to leave stat collection like this in as tracing, or are we going to move this to being something we pass back to Go for inclusion in Flow/Gazette backed metrics/logging?

Just as a basis for developing intuition: what values have you been seeing here for these "mem_changes" logs?

jgraettinger (approved):
LGTM! Merging now.

crates/derive/src/combine_api.rs
jgraettinger: This is a really interesting open question. Something we could do is return metrics as information messages back over the CGO bridge, as part of the extract / combine / derive protocols. This would allow us to tie memory allocations directly to catalog tasks and even shards which drove them.

Of course, these are globals and could be affected by other concurrent and unrelated operations. I have no idea whether they'd still be accurate enough to be useful under load 🤷 . I hope so.

PR #234, opened 2021-09-14
various changes from performance work

jgraettinger:
Please review commit-by-commit

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/234)


psFried (approved):
Looks good to me.

crates/json/src/schema/index.rs
```diff
@@ -54,23 +56,61 @@ where
...
+
+/// Index is a packed, sorted index over Schema references.
+/// It provides lookups over Schema canonical and anchor-form URIs.
+pub struct Index<'s, A>
+where
+    A: Annotation,
+{
+    // Store is subdivided between a |fast| and |slow| index.
+    // |fast| items are statically known to be referenced, and there are fewer of them.
+    // |slow| items may still be referenced, and there are more of them.
```
saterus, line 97: Do you have any anecdotal, order-of-magnitude guess at how many items are ending up in `fast` vs `slow` in the workloads you're trying out? I think my intuition about this must be wrong.

crates/protocol/src/cgo.rs
```diff
@@ -135,7 +135,7 @@ impl Trampoline {
             }
         });
 
-        tracing::debug!(?id, ?code, queued = ?queue.1.len(), "queued trampoline task");
+        tracing::debug!(?id, ?code, queued = ?queue.1.len(), awaiting = ?self.awaiting.borrow().len(), "queued trampoline task");
```
saterus, line 138: Similar question to my last one: how big is the `awaiting` queue getting?

crates/json/src/schema/index.rs
jgraettinger: maybe 4 vs 50 or so in my workload? though heavily dependent on the catalog of course.

Basically every single location in a schema file is a schema, with a canonical URL, which _could_ be referenced. That's going into `slow`. Only schemas which appear as a `$ref `of something else go into `fast`.

crates/protocol/src/cgo.rs
jgraettinger: I've seen it get into the 20-30s in normalish operation, and much larger (10's of thousands) when node was getting wedged.

jgraettinger:
> Not critical for right now, but I think that the schema index might benefit from the use of `Cow` instead of just storing plain references. It would allow us to add owned schemas to the index, and thus avoid leaking them, since any owned value can be trivially converted into a `Cow<'static, T>`.

Huh would it be possible to have a collection that includes both an owned schema, and also separate values which are borrowed references to that owned schema ? That's what would be required, because an owned schema also owns everything under it. But AFAIK that's a violation of Rust's borrow model (since you can't guarantee a destruction order of such a collection).

PR #235, opened 2021-09-14
Pass npm-package as a regular path

psFried:
Changes the js worker to pass the npm package as a regular file path
instead of a relative file uri. This is because recent versions of npm
will not handle the relative file uri.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/235)


jgraettinger (approved):
LGTM

PR #237, opened 2021-09-16
Fix panic when running in develop mode with readDelay.

snowzach:
Since the testing cluster is also used by develop, I believe this check needs to be removed or logic needs to be added to handle logic from commit https://github.com/estuary/flow/commit/db72fc41f660ec1f7a5f41f0c2eaca3572c685bc as it triggers this panic when using a catalog that has readDelay (citi-bike example)

```
panic: expected DisableClockTicks to be set [recovered]
        panic: expected DisableClockTicks to be set

goroutine 1 [running]:
go.gazette.dev/core/mainboilerplate.InitDiagnosticsAndRecover.func2()
        /data/go/pkg/mod/go.gazette.dev/core@v0.89.1-0.20210914224834-2d07306a0edb/mainboilerplate/diagnostics.go:46 +0x77
panic(0x2a506c0, 0x30471e0)
        /usr/local/go/src/runtime/panic.go:971 +0x499
github.com/estuary/flow/go/testing.(*Cluster).Advance(0xc0000dd680, 0x9d29229e0000, 0x2d5e539, 0xd)
        /data/git/est/flow/go/testing/driver.go:119 +0x448
github.com/estuary/flow/go/testing.RunTestCase(0xc00127b180, 0x30ac1c0, 0xc0000dd680, 0xc0010a0000, 0xc000ea2720, 0x1, 0x1, 0x0)
        /data/git/est/flow/go/testing/action.go:75 +0x369
main.cmdDevelop.Execute(0xc000042570, 0x21, 0x2b57e13, 0x4, 0x1f900000, 0x1, 0x7ffd24abdba1, 0x9, 0x7ffd24abdbc8, 0x4, ...)
        /data/git/est/flow/go/flowctl/cmd-develop.go:263 +0x1a8d
github.com/jessevdk/go-flags.(*Parser).ParseArgs(0xc0005c21c0, 0xc00003c110, 0x7, 0x7, 0xc0005f28d0, 0xc0008bfd58, 0x10f4289, 0xc00011e2b8, 0x18)
        /data/go/pkg/mod/github.com/jessevdk/go-flags@v1.5.0/parser.go:335 +0x89d
go.gazette.dev/core/mainboilerplate.MustParseArgs(0xc0005c21c0)
        /data/go/pkg/mod/go.gazette.dev/core@v0.89.1-0.20210914224834-2d07306a0edb/mainboilerplate/config.go:49 +0x85
go.gazette.dev/core/mainboilerplate.MustParseConfig(0xc0005c21c0, 0x2d575d9, 0x8)
        /data/go/pkg/mod/go.gazette.dev/core@v0.89.1-0.20210914224834-2d07306a0edb/mainboilerplate/config.go:44 +0x36e
main.main()
        /data/git/est/flow/go/flowctl/main.go:81 +0x785
````

I am not totally sure if this ends up bypassing an important check otherwise but I replaced the check with the log line that was previously there. Feel free to close if this is not the right way to handle this. 

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/237)

### DIFF
--- a/go/testing/action.go
+++ b/go/testing/action.go
@@ -1,9 +1,11 @@
 package testing
 
 import (
+	"errors"
 	"fmt"
 
 	pf "github.com/estuary/protocols/flow"
+	log "github.com/sirupsen/logrus"
 )
 
 // Driver executes test actions.
@@ -17,6 +19,8 @@ type Driver interface {
 	// Advance TestTime by the given delta.
 	Advance(TestTime) error
 }
+
+var ErrAdvanceDisabled = errors.New("advance disabled")
 
 // RunTestCase runs a test case using the given Graph and Driver.
 func RunTestCase(graph *Graph, driver Driver, test *pf.TestSpec) (scope string, err error) {
@@ -69,11 +73,16 @@ func RunTestCase(graph *Graph, driver Driver, test *pf.TestSpec) (scope string, 
 
 		// Advance time to unblock the next PendingStat.
 		if nextReady != -1 {
-			if err := driver.Advance(nextReady); err != nil {
+			err := driver.Advance(nextReady)
+			if err == ErrAdvanceDisabled {
+				log.Warnf("time advance disabled with action in %s.", nextReady)
+			} else if err != nil {
 				return scope, fmt.Errorf("driver.Advance: %w", err)
+			} else {
+				// Advance completed
+				graph.CompletedAdvance(nextReady)
+				continue
 			}
-			graph.CompletedAdvance(nextReady)
-			continue
 		}
 
 		// All steps are completed, and no pending stats remain. All done.
--- a/go/testing/action.go
+++ b/go/testing/action.go
@@ -75,15 +75,16 @@ func RunTestCase(graph *Graph, driver Driver, test *pf.TestSpec) (scope string, 
 		// Advance time to unblock the next PendingStat.
 		if nextReady != -1 {
 			err := driver.Advance(nextReady)
-			if err == ErrAdvanceDisabled {
-				log.WithFields(log.Fields{
-					"delay": nextReady,
-					"task":  nextName,
-				}).Warn("sleeping until next task is ready.")
-			} else if err != nil {
-				return scope, fmt.Errorf("driver.Advance: %w", err)
+			if err != nil {
+				if err == ErrAdvanceDisabled {
+					log.WithFields(log.Fields{
+						"delay": nextReady,
+						"task":  nextName,
+					}).Warn("sleeping until next task is ready.")
+				} else {
+					return scope, fmt.Errorf("driver.Advance: %w", err)
+				}
 			} else {
-				// Advance completed
 				graph.CompletedAdvance(nextReady)
 				continue
 			}
--- a/go/testing/action.go
+++ b/go/testing/action.go
@@ -20,6 +20,7 @@ type Driver interface {
 	Advance(TestTime) error
 }
 
+// ErrAdvanceDisabled is a specific error returned when clock advance is called when advance is disabled.
 var ErrAdvanceDisabled = errors.New("advance disabled")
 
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by jgraettinger** [comment]
This isn't quite right, because `flowctl develop` is running with actual wall-clock timers rather than managing its own synthetic test time.

AdvanceTime is called by the test driver immediately before calling a Stat call that it knows would otherwise block, due to time not being far enough in the future. I think the right behavior is to simply not call AdvanceTime when `c.Config.DisableClockTicks`, and allow the `Stat` call to block for as long as it wants to.

This does mean that `flowctl develop --poll` will block for up to the full `readDelay`. I'm not sure there's a better behavior. Perhaps we should instrument the AdvanceTime callsite (which already needs to check DisableClockTicks) to inform the user that flow may block for a while, and given more context about for how long, and which derivation + transformation is causing the delay.

**Comment by snowzach** [comment]
How about this one... I assume that if DisableClockTicks is false than Advancing is really disabled all together. (right?) In which case I return a specific error from Advance and log a warning that advance is disabled.

File: go/testing/action.go [hunk 1]
```diff
@@ -18,6 +20,8 @@ type Driver interface {
 	Advance(TestTime) error
 }
 
+var ErrAdvanceDisabled = errors.New("advance disabled")
```
jgraettinger, line 23: nit: add comment for exported types do you have a Go linter set up?).

go/testing/action.go
```diff
@@ -69,11 +73,16 @@ func RunTestCase(graph *Graph, driver Driver, test *pf.TestSpec) (scope string,
 
 		// Advance time to unblock the next PendingStat.
 		if nextReady != -1 {
-			if err := driver.Advance(nextReady); err != nil {
+			err := driver.Advance(nextReady)
+			if err == ErrAdvanceDisabled {
+				log.Warnf("time advance disabled with action in %s.", nextReady)
```
jgraettinger, line 78: nit: use structured logging, like:

```
log.WithFields(log.Fields{"delay": nextReady}).Warn("the message")
```

I suspect this warning is going to be pretty confusing to users. We should plumb through more context for the user about _why_ there's a delay. Can we get the name of the catalog task ? Probably means having PopReadyStats pass back more info.

Also note that `nextReady` is an absolute duration that's relative to the test start time, so converting to a deadline might be more useful (particularly because there can be multiple delays).

jgraettinger (commented):
This is looking good to me, definitely the right approach, but some comments below to (I think) make it more helpful to the user.

go/testing/action.go
snowzach: Sorry, I'm used to having linting part of of CI and VS Code seems to be grumpy using staticheck with all the CGooeyness. I tried Golint and it seems happier.

go/testing/action.go
snowzach: I tried to make it more obvious. The time being returned is the actual delta time, so zero if there's something to do otherwise some time.Duration. I just mad the message, `sleeping until next task is ready.` If I'm there, I wanna know how long I am going to be waiting.

go/testing/action.go
```diff
@@ -69,11 +74,19 @@ func RunTestCase(graph *Graph, driver Driver, test *pf.TestSpec) (scope string,
...
+			err := driver.Advance(nextReady)
+			if err == ErrAdvanceDisabled {
+				log.WithFields(log.Fields{
+					"delay": nextReady,
+					"task":  nextName,
+				}).Warn("sleeping until next task is ready.")
+			} else if err != nil {
 				return scope, fmt.Errorf("driver.Advance: %w", err)
+			} else {
+				// Advance completed
```
jgraettinger, line 86: nit: obvious comment is obvious :)

Also, prefer to [indent error flow](https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow).

jgraettinger (approved):
LGTM! % final nit

PR #242, opened 2021-09-20
assorted fixes from attempting to reproduce snowflake timeouts plus backoff cancellation

jgraettinger:
See individual comments. I observed some crashes from `flowctl` due to poorly behaved materialization connectors (because, to be fair, I was kill -9'ing them).

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/242)

### DIFF
--- a/go/capture/adapter.go
+++ b/go/capture/adapter.go
@@ -27,6 +27,8 @@ func CaptureResponseChannel(stream pc.Driver_CaptureClient) <-chan CaptureRespon
 	var ch = make(chan CaptureResponse, 4)
 	go func() {
 		for {
+			// Use Recv because ownership of |m| is transferred to |ch|,
+			// and |m| cannot be reused.
 			var m, err = stream.Recv()
 
 			if err == nil {
@@ -43,6 +45,32 @@ func CaptureResponseChannel(stream pc.Driver_CaptureClient) <-chan CaptureRespon
 	}()
 
 	return ch
+}
+
+// Rx receives from a CaptureResponse channel.
+// It destructures CaptureResponse into its parts,
+// and also returns an explicit io.EOF for channel closures.
+func Rx(ch <-chan CaptureResponse, block bool) (*pc.CaptureResponse, error) {
+	var rx CaptureResponse
+	var ok bool
+
+	if block {
+		rx, ok = <-ch
+	} else {
+		select {
+		case rx, ok = <-ch:
+		default:
+			ok = true
+		}
+	}
+
+	if !ok {
+		return nil, io.EOF
+	} else if rx.Error != nil {
+		return nil, rx.Error
+	} else {
+		return rx.CaptureResponse, nil
+	}
 }
 
 // AdaptServerToClient wraps an in-process DriverServer to provide a DriverClient.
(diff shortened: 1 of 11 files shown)


jgraettinger:
These were small and the review was outstanding, so I've piggybacked a fix for issue #244 on this.

go/capture/driver/airbyte/connector.go
```diff
@@ -361,18 +372,18 @@ func (o *jsonOutput) Close() error {
 }
 
 type connectorStderr struct {
-	err bytes.Buffer
+	delegate io.Writer
```
psFried, line 375: I was intending to forward stderr lines to the logs collection in #243 but I was holding off there because I wasn't sure how to parse the stderr output. I now have a bit more conviction on how to handle it there, so should be able to add that into the PR. I don't think there's any need to change anything in _this_ PR, but just figured I'd mention it since I'd be trampling over this code assuming we merge this first.

go/shuffle/read.go
```diff
@@ -595,8 +617,8 @@ func pickHRW(h uint32, from []shuffleMember, start, stop int) int {
 }
 
 // readChannelCapacity is sized so that sendReadResult will overflow and
-// cancel the read after ~8 seconds of no progress (1 << (14 - 1) millis).
-var readChannelCapacity = 15
+// cancel the read after ~65 seconds of no progress (1<<15 + 1<<14 + 1<<13 ... millis).
```
psFried, line 620: Does this math still add up after changing the channel capacity to 17?

psFried (approved):
Looks good :+1:

go/capture/driver/airbyte/connector.go
jgraettinger: Ack

go/shuffle/read.go
jgraettinger: Yep, it's 1 << (len(ch)-1), and we never let ch fill to capacity

go/capture/adapter.go
```diff
@@ -27,6 +27,8 @@ func CaptureResponseChannel(stream pc.Driver_CaptureClient) <-chan CaptureRespon
 	var ch = make(chan CaptureResponse, 4)
 	go func() {
 		for {
+			// Use Recv because ownership of |m| is transferred to |ch|,
```
jixij, line 30: Just curious, is there any anti-patterns when you add this comment?

jgraettinger: I don't quite follow (anti-patterns about what?).

What this comment is getting at is that gRPC provides another RecvMsg call, which takes an owned instance and fills into it. Often that'll be used because it avoids an allocation. This comment was coped from go/materialize/adapater.go to note that that's not allowed here given the ownership semantics. In practice, it also doesn't help because we're using an adapter channel which wants to returned owned instances anyway.

jixij (commented):
LGTM. 
Like the fixes `driverRx`, as well as the improvements with `drainedCh`. :+1:

go/capture/adapter.go
jixij: got it. Thanks!

PR #245, opened 2021-09-22
Revert nodejs to v14 LTS release

snowzach:
Because of this issue with Docker For Mac https://github.com/docker/for-mac/issues/5831
Using the flowctl.sh script does not work on M1 Macs when it is using Node 16. This PR reverts back to Node V14 which is the LTS release anyhow and does include the nullish collating operator which I understand was the main driver for upgrading Node.
It also adds a command option to ensure the flow docker image is up to date as part of the `flowctl.sh` script.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/245)


jgraettinger (approved):
LGTM. Merging now because this also seems to fix an npm permission issue seen w/ connector tests.

PR #246, opened 2021-09-22
Pin `bitvec` dependency to 0.19

saterus:
Our `json` create directly depends on `bitvec`, but we also on `nom` in a different crate. `nom` also depends on `bitvec`. We've not specified a version, but rather relied on Cargo to resolve a working version for
`bitvec`.

Now, `bitvec`'s `0.20.0` release introduced some semver breaking changes. `nom` has `bitvec` pinned at `0.19`. This causes Cargo to resolve `flow`'s dependency on `bitvec` to be `0.19`, rather than the latest `0.2x.x` release.

However, other crates which depend on our `json` crate to fail to pickup the `nom` constraint. This allows them to resolve `bitvec` to the `0.2x.x` releases. This are incompatible with the code we've written in the `json` crate.

Pinning `bitvec` to the `0.19.x` releases prevents dependent crates from trying to use an incompatible version of `bitvec`. :tada: 

Eventually we'll probably upgrade our `nom` dependency, which no longer depends on `bitvec`. This would remove the necessity of using `0.19.x`, but we'd still have to update our code in `json` to use the new APIs.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/246)

### DIFF
--- a/crates/json/Cargo.toml
+++ b/crates/json/Cargo.toml
@@ -5,7 +5,7 @@ authors = ["Estuary Technologies, Inc"]
 edition = "2018"
 
 [dependencies]
-bitvec = "*"
+bitvec = "0.19"
 fancy-regex = "*"
 fxhash = "*"
 itertools = "*"


psFried (commented):
Looks to me like `nom` is purely a build-time dependency, brought in via `librocksdb-sys -> bindgen -> cexp -> nom`. Not that we _need_ to right now, but it seems like we could just as easily pin bitvec to the newer version and update the json crate to be compatible with that. Does that sound right to you? Do you have a rough sense of what it would take to make our json crate compatible with the newer bitvec?

saterus:
> Looks to me like `nom` is purely a build-time dependency, brought in via `librocksdb-sys -> bindgen -> cexp -> nom`. 

I'm not entirely sure that it matters whether it's a build-only dependency or not. :shrug: I think Cargo knows that it needs to resolve `nom`'s specific `0.19` version and it consolidates any other `bitvec`s to that version.

> Not that we _need_ to right now, but it seems like we could just as easily pin bitvec to the newer version and update the json crate to be compatible with that. Does that sound right to you? Do you have a rough sense of what it would take to make our json crate compatible with the newer bitvec?

Yeah, I don't think there's anything showstopping about that upgrade. The error messages when building the `parser` revolve around [changing `iter()` to return `BitRef`s instead of `&bool`s](https://github.com/bitvecto-rs/bitvec/blob/develop/CHANGELOG.md#0200). I would go ahead and guess that it's all equivalent, but I didn't look into it much further since I didn't think we could upgrade it while keeping `nom` happy.

psFried (approved):
Ok I think I'm following now, and agree that it makes sense to just merge this for now and then deal with the bitvec upgrade later.

Issue #247, opened 2021-09-22
connectors: in-process runtime drivers can't assume memory ownership


jgraettinger:
Implementations of gRPC service interfaces generally get to assume full ownership of their request structures. This happens naturally, since requests are deserialized from byte streams immediately prior to handler invocation.

Flow has gRPC service implementations (notably `go/materialize/driver/image|sqlite` and `go/capture/driver/airbyte`) which are designed to be called in synchronous, in-process contexts in order to minimize ser/de & memory copies. These drivers are often adapters, proxying a flavor of the capture / materialize protocol to a delegated driver.

We need to be careful not to mutate shared state in request structures within these drivers. For example, `materialize/driver/image` is [unwrapping a layer of endpoint configuration](https://github.com/estuary/flow/blob/f80139ecda8dd8ab5630acaca77ff3422128b66f/go/materialize/driver/image/driver.go#L114-L115) before proxying it to a containerized driver ... and this is inadvertently modifying the underlying `pf.CatalogTask` held in the etcd KeySpace !

Fix this, and audit drivers for other cases of non-owned mutation (I'm not aware of any more off-hand).

PR #248, opened 2021-09-23
capture/materialize: in-process drivers must deep copy specifications

jgraettinger:
MaterializationSpec is shared with the catalog KeySpace and must be copied.
Audit and document this restriction in drivers.

Tested:
 * Started with a `flowctl develop` catalog having a built-in Snowflake materialization
 * Executed `flowctl shard split` while that catalog was running (1 -> 2 shards).
 * Transitioned the catalog materialization in-place to using the `materialize-snowflake` connector.
 * Executed `flowctl shard split` 2x while _that_ catalog was running (2 -> 4 shards).

Issue #247

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/248)

### DIFF
--- a/go/capture/driver/airbyte/driver.go
+++ b/go/capture/driver/airbyte/driver.go
@@ -56,6 +56,10 @@ func (c ResourceSpec) Validate() error {
 }
 
 // driver implements the pm.DriverServer interface.
+// Though driver is a gRPC service stub, it's called in synchronous and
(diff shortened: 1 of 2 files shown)


jgraettinger:
Merging because this is a small, scoped change that fixes an active bug. Please leave feedback and I'll address as a follow-up

PR #249, opened 2021-09-23
Add the ability to provide extra docker args to the connector docker call

snowzach:
Add support for a DOCKER_EXTRA_CONNECTOR_OPTS environment variable which will specify additional arguments to the docker command that runs connectors. This is primarily used to pass platform arguments to docker when running our containers on an M1 Mac for example.

We had talked about adding arbitrary options to docker. Turns out we need this to make M1 Mac work. (as well as it will work)

M1 Mac will work, however it seems to produce somewhat random Segfaults from the connector. The better solution might be to cross compile and connectors we can for arm64. I may do that when I get the BigQuery connector rolling.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/249)

### DIFF
--- a/go/capture/driver/airbyte/connector.go
+++ b/go/capture/driver/airbyte/connector.go
@@ -56,20 +56,7 @@ func RunConnector(
 	}
 
 	// Check if we have any extra arguments for docker from the environment and pass them along.
-	// We need to split args properly (by unquoted spaces).
-	var quote rune = 0x00 // 0x00 = not currently inside a quote.
-	if args := strings.FieldsFunc(os.Getenv("DOCKER_EXTRA_CONNECTOR_OPTS"), func(r rune) bool {
-		if r == '"' || r == '\'' {
-			if r == quote {
-				// End of quoted string.
-				quote = 0x00
-			} else {
-				// Start of quoted string.
-				quote = r
-			}
-		}
-		return quote == 0x00 && r == ' ' // Not inside a quote and on a space, return true and split.
-	}); len(args) > 0 {
+	if args := parseArgs(os.Getenv("DOCKER_EXTRA_CONNECTOR_OPTS")); len(args) > 0 {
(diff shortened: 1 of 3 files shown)


jgraettinger (changes requested):
> Reviewed 2 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/249)* status: all files reviewed, 2 unresolved discussions (waiting on @snowzach)

___
*[go/capture/driver/airbyte/connector.go, line 58 at r1](https://reviewable.io/reviews/estuary/flow/249#-Mkw_e_qAxxCszhO1www:-Mkw_e_qAxxCszhO1wwx:b-7i16bx) ([raw file](https://github.com/estuary/flow/blob/7ffaa23875a8487ec933862f803f43d214953e1c/go/capture/driver/airbyte/connector.go#L58)):*
> ```Go
> 	}
> 
> 	// Check if we have any extra arguments for docker from the environment and pass them along.
> ```

Given this capability, do we still want to hard-code `--network` ? I'm not sure how to relate our treatment of networks with this.
___
*[go/capture/driver/airbyte/connector.go, line 61 at r1](https://reviewable.io/reviews/estuary/flow/249#-Mkw_NfnCowuyI3r4UWQ:-Mkw_NfnCowuyI3r4UWR:b-casa3r) ([raw file](https://github.com/estuary/flow/blob/7ffaa23875a8487ec933862f803f43d214953e1c/go/capture/driver/airbyte/connector.go#L61)):*
> ```Go
> 	// We need to split args properly (by unquoted spaces).
> 	var quote rune = 0x00 // 0x00 = not currently inside a quote.
> 	if args := strings.FieldsFunc(os.Getenv("DOCKER_EXTRA_CONNECTOR_OPTS"), func(r rune) bool {
> ```

This is neat. Extract to a function taking a string and returning split arguments, and add a test ?

PR #253, opened 2021-09-27
convert nil resp

jixij:
The `airbyte.RunConnector` is not converting an empty response from places like [this](https://github.com/estuary/connectors/blob/main/materialize-webhook/driver.go#L116), leaving nil resp variable unchanged, which causes `flowctl` to crash on [this line](https://github.com/estuary/flow/blob/master/go/flowctl/cmd-apply.go#L313).

The proposed fix sets the `nil` response variable, if not populated, to an empty response in `Apply` command of the `image` driver.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/253)

### DIFF
--- a/go/capture/driver/airbyte/connector.go
+++ b/go/capture/driver/airbyte/connector.go
@@ -235,6 +235,8 @@ func (o *protoOutput) Write(p []byte) (n int, err error) {
 		if o.next == 0 {
 			if _, err = o.decodeLen(o.rem); err != nil {
 				return 0, err
+			} else if o.next == 0 {
+				o.decodeMsg([]byte{})
 			}
 		} else {
 			if _, err = o.decodeMsg(o.rem); err != nil {
@@ -252,6 +254,8 @@ func (o *protoOutput) Write(p []byte) (n int, err error) {
 				return n, nil
 			} else if p, err = o.decodeLen(p); err != nil {
 				return 0, err
+			} else if o.next == 0 {
+				o.decodeMsg([]byte{})
 			}
 		} else {
 			if len(p) < o.next {
--- a/go/capture/driver/airbyte/connector_test.go
+++ b/go/capture/driver/airbyte/connector_test.go
@@ -95,12 +95,17 @@ func TestProtoRecordBreaks(t *testing.T) {
 	// Create a decoder of CollectionSpec, which expects to see parsed
 	// copies of our fixture, and counts its number of invocations.
 	var verifyCount int
+	var verifyEmptyCount int
 	var s = NewConnectorProtoOutput(new(flow.CollectionSpec), func(m proto.Message) error {
-		require.Equal(t, m, &flow.CollectionSpec{
-			Collection: "a/collection",
-			KeyPtrs:    []string{"/a", "/b", "/c"},
-		})
-		verifyCount++
+		if m.String() != "" {
+			require.Equal(t, m, &flow.CollectionSpec{
+				Collection: "a/collection",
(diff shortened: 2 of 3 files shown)


jgraettinger:
Did the webhook materialization, specifically, cause a crash? Or did you encounter this with another connector that might not have sent an ApplyResponse at all ?

jixij:
> Did the webhook materialization, specifically, cause a crash? Or did you encounter this with another connector that might not have sent an ApplyResponse at all ?

I encountered this in the new parquet connector. If I apply this change, it works well.
I am referring to the webhook b/c there is no pr for the pq connector yet. I can double check on the webhook to see if I can reproduce it there.

jixij:
I can reproduce the issue on webhook with the following testing config
```
collections:
  tests/materialize-s3-parquet/data:
    schema:
      properties:
        id:
          type: integer
        canary:
          type: string
      required:
        - id
        - canary
      type: object
    key:
      - /id
materializations:
  tests/materialize-s3-parquet/materialize:
    endpoint:
      flowSink:
        image: wh:latest
        config: {"address": "http://test"}
    bindings:
      - resource:
          relativePath: aa/bb
        source: tests/materialize-s3-parquet/data
```

jgraettinger:
Can you determine if it's simply not writing an ApplyResponse at all, somehow? I'd like to understand what's actually happening before patching this. The contract at all stages is that a gRPC stub return either an error or a non-nil response, so to the extent there's a fault in `driver.Apply` it's only that it's not returning an error "hey, your connector implementation has an error because it didn't return an ApplyResponse" (and we should make sure other RPCs have a similar check).

The actual bug is elsewhere.

jgraettinger:
FWIW I'm betting the bug is actually in here: https://github.com/estuary/flow/blob/master/go/capture/driver/airbyte/connector.go#L207

We're not distinguishing "I got a header, but it was empty" from "I haven't read a header yet"

jixij:
> Can you determine if it's simply not writing an ApplyResponse at all, somehow? I'd like to understand what's actually happening before patching this. The contract at all stages is that a gRPC stub return either an error or a non-nil response, so to the extent there's a fault in `driver.Apply` it's only that it's not returning an error "hey, your connector implementation has an error because it didn't return an ApplyResponse" (and we should make sure other RPCs have a similar check).
> 
> The actual bug is elsewhere.

The check I have done is that if I make the ApplyResponse to be nontrivial, e.g.
`m.ApplyResponse{ActionDescription: "hello world!!!!!"}`, it works. So I think it is sending the response back.
I can dig more into it and see what was happening there.

jixij:
> FWIW I'm betting the bug is actually in here: https://github.com/estuary/flow/blob/master/go/capture/driver/airbyte/connector.go#L207
> 
> We're not distinguishing "I got a header, but it was empty" from "I haven't read a header yet"

Yes, it seems to be working after I patched the following changes [here](https://github.com/estuary/flow/blob/master/go/capture/driver/airbyte/connector.go#L253). (Might need to check ther other places where `decodeLen` is called..)
```
                        } else if p, err = o.decodeLen(p); err != nil {
                                return 0, err
+                       } else if o.next == 0 {
+                               o.decodeMsg([]byte{})
                        }
```

jgraettinger:
Thanks! I see now this is ready for review ?

jgraettinger (approved):
LGTM!

jixij:
> Thanks! I see now this is ready for review ?

Yes. Thanks a lot for the suggestions and approval!

PR #275, opened 2021-10-19
incremental model work towards a CP API

jgraettinger:
This is a incremental set of patches which move closer to models appropriate for our CP vision.
It builds on PR #265 , and further refines `storageMappings`, so it makes sense to land them together rather than landing an intermediate state we'll migrate from.

Best to review commit by commit. Many of these are snapshot churn and mechanical fallout of model changes which are made.

Sorry for the size. It's hard to change this stuff without touching _a lot_ of the codebase.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/275)


jgraettinger:
Merged through another PR

Issue #206, opened 2021-08-24
Exposing logs and metrics as Flow collections


psFried:
### Discussed in https://github.com/estuary/flow/discussions/175

- [x] Catalog build automatically generates collections for logs and stats
- [ ] ~~Add logs and stats collection names as shard labels~~
- [x] Flow runtime publishes stats for each transaction
- [x] Flow runtime publishes logs
    - [x] From source connectors `LOG` messages
    - [x] From all connectors stderr (Shoudl include parser logs for file source connectors)
    - [x] From Rust services
    - [x] From materialization connectors (after #173 is done)
    - [ ] ~~From nodejs runtime stderr~~
    - [x] Whenever a task term begins
    - [x] On any schema validation error
    - [x] Other errors, as long as we are able to publish messages

Update:
It no longer seems worthwhile to capture logs from the current nodejs runtime, since it is shared by all shards with the same build id. The current idea is to just revisit derivation logs as we add support for Deno and/or WASM.

PR #293, opened 2021-11-16
Return stats from Rust bindings

psFried:
This is an incremental step towards the publishing of stats to the
ops/<tenant>/stats collection. Stats are now returned from the Rust side
of bindings whenever a combiner or derive is drained. These stats are
currently just logged and ignored on the Go side of bindings. In a
subsequent PR, they will be aggregated per transaction and published to
the ops collections.

Part of #206 

While not strictly required to be part of this PR, I included changes to `ops-stats-schema.json` in order to give a better idea of how these are shaping up to look once published. This has evolved considerably since the initial discussion (#175), for a few reasons:

- To allow capture stats to be broken down per binding, which is required in order to do things like derive the number of documents added to a specific collection.
- Allow materialization stats to be broken down per binding, for basically the same reason.
- Allow derivation stats to be broken down by _transform_ because it seems more useful, and because it makes it more similar to how materialization stats are broken down by binding.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/293)


crates/derive/src/pipeline/mod.rs
```diff
@@ -312,10 +364,17 @@ impl Pipeline {
                 Poll::Pending => break,
                 Poll::Ready(None) => break,
                 Poll::Ready(Some(result)) => {
-                    let (mut block, tf_register_deltas) =
+                    let (mut block, update_outputs) =
                         result.map_err(Error::UpdateInvocationError)?;
 
-                    self.update_registers(&mut block, tf_register_deltas)?;
+                    for (transform_index, result) in update_outputs.iter().enumerate() {
+                        println!("invoke stats for {}: {:?}", transform_index, result.stats);
```
jgraettinger, line 371: nit: left-over println ?

crates/derive/src/pipeline/pipeline_test.rs
```diff
@@ -0,0 +1,378 @@
...
+
+    // These should only show up as 2 registers created
+    fixture.add_source_document(2, key("tea"), r#"{"key": "tea", "sum": 9}"#.as_bytes());
+    fixture.add_source_document(2, key("tea"), r#"{"key": "tea", "sum": 1}"#.as_bytes());
+    fixture.add_source_document(2, key("free"), r#"{"key": "free", "sum": 42}"#.as_bytes());
+
+    let (docs, stats) = fixture.poll_to_completion();
+    // Redact the time output on the snapshot so that it's deterministic.
+    insta::assert_yaml_snapshot!(stats, {
+        ".**.total_seconds" => "time-redacted",
```
jgraettinger, line 53: neat!

crates/derive/src/pipeline/pipeline_test.rs
```diff
@@ -0,0 +1,378 @@
...
+}
+
+#[derive(Default)]
+struct TransformFixture {
+    update: Option<Box<dyn FnMut(Value) -> Result<Vec<Value>, ()>>>,
+    /// publish lambda parameters are `(source, prev_register, new_register)`
+    publish: Option<Box<dyn FnMut(Value, Value, Value) -> Result<Vec<Value>, ()>>>,
+}
+
+impl Fixture {
```
jgraettinger, line 78: This is great, thanks for adding it

crates/protocol/build.rs
```diff
@@ -20,6 +20,31 @@ struct TypeAttrs<'a> {
 /// https://docs.rs/prost-build/0.6.1/prost_build/struct.Config.html#arguments
 /// `field_attrs` holds tuples of field name to field attributes.
 static TYPE_ATTRS: &'static [TypeAttrs<'static>] = &[
+    TypeAttrs {
```
jgraettinger, line 23: _do_ these need to serde ? aren't we passing them around using protobuf encoding from the Rust side ?

go/bindings/combine.go
```diff
@@ -151,14 +156,22 @@ func (d *Combine) Destroy() {
 	}
 }
 
+type StatsMessage interface {
```
jgraettinger, line 159: nit: Use `proto.Unmarshaler` instead ?

jgraettinger (approved):
LGTM! Some nits

jgraettinger (commented):
> Reviewed 26 of 26 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/293)* status: all files reviewed, 9 unresolved discussions (waiting on @psFried)

crates/derive/src/pipeline/invocation.rs
```diff
@@ -111,30 +119,104 @@ impl Invocation {
...
-
+                        let total_duration = start.elapsed();
                         let result = result.and_then(|data| {
                             let parsed: Vec<Vec<Value>> = serde_json::from_slice(data)
                                 .context("failed to parse lambda invocation response")?;
-                            Ok(parsed)
+                            let stats = InvokeStats {
+                                output: DocCounter::new(
+                                    parsed.iter().map(Vec::len).sum::<usize>() as u64,
+                                    // TODO: _technically_ we _could_ subtract some bytes here to
```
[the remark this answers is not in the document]
psFried: Agreed. Maybe I'll keep the comment, minus the "TODO" portion, just to help with inevitable questions about exactly how we get these numbers.

crates/protocol/build.rs
psFried: Whoops, I meant to add a comment about that. These only need to serde for the snapshot tests, since the `debug_snapshot` variants don't support redaction. I'll propose adding a comment here and also changing these structs to use `cfg_attr(test, derive(Serialize, Deserialize))` so that it's more clear.

jgraettinger: 👍

crates/build/src/ops/ops-stats-schema.json
```diff
@@ -11,87 +11,130 @@
             "format": "date-time",
             "reduce": {"strategy": "maximize"}
```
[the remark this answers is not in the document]
psFried: Good catch! Fixed.

crates/derive/src/pipeline/pipeline_test.rs
psFried: Thanks! I started feeling a little uncertain about it after seeing how big it got, but hopefully it'll be useful in more tests or benchmarks in the future.

go/bindings/combine.go
psFried: Thanks for sharing. I didn't know that was a thing.

crates/protocol/build.rs
psFried: Just to close the loop on this, using `cfg_addr(test, ...)` doesn't quite work, because the tests are in the `build` crate, and Cargo doesn't compile a crate's dependencies using the `test` config, only the crate that's actually under test. So I did this by adding a `test-support` feature to `protocol`, which is enabled as part of `dev-dependencies` in the `build` crate.

psFried (commented):
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/293)* status: 13 of 29 files reviewed, 6 unresolved discussions (waiting on @jgraettinger)

___
*[crates/build/src/ops/ops-stats-schema.json, line 20 at r1](https://reviewable.io/reviews/estuary/flow/293#-MoeIqG_2aTZmW5TJJep-r1-20:-MpCVO7A4Qokvub8FtD-:b-896fix) ([raw file](https://github.com/estuary/flow/blob/288fcfa88bc6bec44633d380b6e3012190177458/crates/build/src/ops/ops-stats-schema.json#L20)):*

nit: Is there another way to put this that's a bit more noob friendly? "Statistics for each bound collection of the capture" ?
</blockquote></details>

Done.
___
*[crates/derive/src/pipeline/invocation.rs, line 157 at r1](https://reviewable.io/reviews/estuary/flow/293#-MoeIqG_2aTZmW5TJJew-r1-157:-MpCVYPC5o_VbxOOLpJ-:b-896fix) ([raw file](https://github.com/estuary/flow/blob/288fcfa88bc6bec44633d380b6e3012190177458/crates/derive/src/pipeline/invocation.rs#L157)):*

This trait is 👌 
</blockquote></details>

Done.
___
*[crates/derive/src/pipeline/pipeline_test.rs, line 53 at r1](https://reviewable.io/reviews/estuary/flow/293#-MoeIqG_2aTZmW5TJJey-r1-53:-MpCVbC50npzxkgXbFBG:b-896fix) ([raw file](https://github.com/estuary/flow/blob/288fcfa88bc6bec44633d380b6e3012190177458/crates/derive/src/pipeline/pipeline_test.rs#L53)):*

neat!
</blockquote></details>

Done.

PR #294, opened 2021-11-16
connector clients with Acknowledgement semantics, ApplyDelete, and E2E test

### LINKED ISSUE #179: captures: recast ingestion as a form of capture
Currently we offer ingestion APIs (HTTP & Websocket) through a `flow-ingester` component that's separate from the `flow-reactor`.

It has a couple of issues:

 * Multi-collection transaction semantics _require_ that there be a stable state store into which ACK intents are written, which is recovered from and replayed on failure. Today, this is stubbed out / incorrectly implemented in `flow-ingester`.
 * Were we to add a stable state-store to flow-ingester, it raises new design questions of how that store is provisioned and scaled.
 * We _may_ want to allow users to write arbitrary TypeScript lambdas that are invoked to produce captured documents.
   - A motivating use case is allowing user code to inspect & merge HTTP headers alongside the POST body.
   - This isn't a required day-one feature, but it's in tension with `flow-ingester` architecture because ingesters are a shared resource. Each would have to simultaneously serve all possible user lambdas.
 * flow-ingester ingests directly into named collections (encoded in, say, an HTTP path).
   - We should decouple the ingested resource (HTTP path, or Kafka gateway topic, etc) from the literal collection being written to.
   - Decoupling these allows, for example, an established WebHook path to be "cut over" to new written collections.

A really lightweight sketch:

We should model ingestion as a kind of capture, alongside airbyte connectors
 - The task's shard stores provide durable storage for ACK intents written by that shard, as part of the regular consumer processing workflow.
 - We'd (ahem) "hack" the message read loop to wire up events ingested over an API as the source of messages driving consumer transactions.
 - Any flow-reactor could potentially receive an ingestion request. It would use proxy the request to the current primary for the resolved capture task (this is a pretty established practice in the gazette

### LINKED ISSUE #272: capture/materializations: acknowledgements and Apply / Delete RPCs
Some data sources require explicit confirmation that data has been consumed and will not be needed again. As a concrete example, PostgreSQL expects a `flush_lsn` value from replication clients which will allow it to advance the `restart_lsn` of a replication slot, discard old WAL segments, etc. Similarly some message queues require explicit acknowledgement of delivery and will buffer / resend messages until they're consumed.

But we don't want our connectors to just assume that data has been persisted once they write it to `stdout`, that's just asking for trouble. And I'm told that no heuristic (like "wait 5 minutes after emitting a record") could be used that wouldn't be violated under some possible recovery scenario.

Thus we need to actually implement some sort of explicit commit acknowledgement from Flow to the source connector which tells it "your output has been committed up to <this point> and you will never be restarted from an older state checkpoint".

jgraettinger:
Add support for:
 * Acknowledgements in both capture and materialization connectors.
 * `flowctl` support for removing materializations when they're deleted (either by `flowctl deploy`, or `flowctl api delete`).
 * A new end-to-end integration test which runs multiple captures, derivations, and materializations, and verifies their output against an expectation (we previously didn't have good CI coverage of captures and materializations with real connectors).
 * A `Runtime.Push` gRPC which executes streams of pushed documents against the capture shard.

Fixes #272
Fixes #179 

This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/294)


jgraettinger:
Rats, I'm having trouble getting a passing build due to conflicting `go.mod` rebases. It [does build](https://github.com/estuary/flow/actions/runs/1517181138) with #302 , which depends on this PR, and I'll plan to merge them at the same time.

go/flowctl/cmd-api-delete.go
```diff
@@ -9,24 +9,28 @@ import (
...
-	BuildID     string                `long:"build-id" required:"true" description:"ID of this build"`
-	BuildsRoot  string                `long:"builds-root" required:"true" env:"BUILDS_ROOT" description:"Base URL for fetching Flow catalog builds"`
-	Consumer    mbp.ClientConfig      `group:"Consumer" namespace:"consumer" env-namespace:"CONSUMER"`
-	DryRun      bool                  `long:"dry-run" description:"Print actions that would be taken, but don't actually take them"`
-	Names       []string              `long:"name" description:"Name of task or collection to activate. May be repeated many times"`
-	Network     string                `long:"network" default:"host" description:"The Docker network that connector containers are given access to."`
-	Log         mbp.LogConfig         `group:"Logging" namespace:"log" env-namespace:"LOG"`
-	Diagnostics mbp.DiagnosticsConfig `group:"Debug" namespace:"debug" env-namespace:"DEBUG"`
+	All            bool                  `long:"all" description:"Delete all tasks and collections"`
+	AllDerivations bool                  `long:"all-derivations" description:"Activate all derivations"`
```
jixij, line 24: (nit) `Delete`, instead of `Activate`?

go/runtime/push.go
```diff
@@ -0,0 +1,174 @@
...
+	"google.golang.org/grpc"
+)
+
+// TODO(johnny): This implementation builds and I believe it to be correct
+// and what we want, but it's also untested at this time (November 2021).
+// I'm putting it down for the time being, with plans to pick it up as we
+// re-add features around HTTP, WebSocket, and Kafka push ingestion.
+var _ capture.RuntimeServer = (*FlowConsumer)(nil)
+
+func (f *FlowConsumer) Push(stream capture.Runtime_PushServer) error {
```
jixij, line 24: Got this working with an example like [this](https://github.com/estuary/flow/commit/26405051bb28e551b6fbf6802c0986bd7c784dbd). But seems `proxyPush` was not covered in the test example.

jixij (approved):
Look great to me. :+1: 
Through the review, I know better about the captures and the Acknowledge semantics.

go/capture/driver/airbyte/driver.go
```diff
@@ -281,26 +285,45 @@ func (d driver) Validate(ctx context.Context, req *pc.ValidateRequest) (*pc.Vali
 	return resp, nil
 }
 
-func (d driver) Capture(req *pc.CaptureRequest, stream pc.Driver_CaptureServer) error {
+// ApplyUpsert is a no-op (not supported by Airbyte connectors).
+func (d driver) ApplyUpsert(context.Context, *pc.ApplyRequest) (*pc.ApplyResponse, error) {
```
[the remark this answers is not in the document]
jgraettinger: Because we're going through the airbyte spec, these are effectively no-ops.

But they're needed (and it has me considering whether we ought to make the capture protocol the native protocol for our own connectors). A good example of why is replication slots in CDC databases, which are a created entity that consumes DB resources. We want to create slots when the capture is created, and delete it when the capture goes away. But today we can't thread this through w/ the airbyte protocol.

go/runtime/push.go
jgraettinger: Awesome! Glad it's working so far.

go/capture/driver/airbyte/driver.go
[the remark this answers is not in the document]
jixij: I see. Cool!

go/flowctl/cmd-api-delete.go
jgraettinger: Done

PR #302, opened 2021-11-29
integrate `sops` for credential management

jgraettinger:
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/302)


crates/sources/src/loader.rs
```diff
@@ -764,24 +836,58 @@ impl<F: Fetcher> Loader<F> {
...
+            config: models::Config::Url(relative),
+        }) = endpoint
+        {
+            let absolute = self.fallible(scope, scope.resource().join(&relative))?;
+            self.load_import(scope, &absolute, models::ContentType::Config)
+                .await;
+
+            endpoint = AirbyteSource(models::ConnectorConfig {
+                image,
+                config: models::Config::Url(models::RelativeUrl::new(absolute.to_string())),
```
jixij, line 858: (Most likely due to my lack of knowledge in the catalog building process and Rust) 
It looks to me that since the process is `await` on the `load_import`, the config file should be loaded when the execution reaches this line, so is it possible to perform the `to_inline` here, instead of storing the URL... 
I feel there should be some reasons of not doing it, but not sure about the exact.

examples/citi-bike/rides.config.yaml
```diff
@@ -0,0 +1,23 @@
...
+awsSecretAccessKey: ""
+# Name of the bucket managed by CitiBike.
+bucket: tripdata
+# We encrypt the region just to demonstrate the capability (this is "us-east-1").
+# We used a `sops` --encrypted-suffix "-sops" option to selectively encrypt only
+# some values of this configuration. Flow knows to strip this suffix.
+region-sops: ENC[AES256_GCM,data:OnrKzkp1aQGz,iv:NmUy9yeRdH1jvzVFuf/sKzMqAf6hEvOA5ohAsJ50O68=,tag:hUSkjQejD7cwzd4aPGlC1A==,type:str]
+sops:
+    kms: []
+    gcp_kms:
```
jixij, line 12: This might prevent the citi-bike demo working on local machines without GCP settings.

I know this is just the start of `sops` integration, and before the work is completed, is it possible to keep a non-encrypted version of `rides.config.yaml` here, and override it with a encrypted version during integration tests, so that the demo can work for the needs from both local and testing environment?

go/connector/decrypt.go
```diff
@@ -0,0 +1,100 @@
...
+		return decrypted, nil
+	}
+	// We must re-write the document to remove the encrypted suffix.
+	// |decrypted| is now only used as input for this stripping process.
+	defer ZeroBytes(decrypted)
+
+	// We'll use `jq` to do the re-writing. This avoids allocating and parsing
+	// values in our own heap, and is also succinct.
+	// See: https://jqplay.org/s/sQunN3Qc4s
+	stripped, err := decryptCmd(ctx, decrypted, "jq",
```
jixij, line 54: cool!

go/connector/decrypt.go
```diff
@@ -0,0 +1,100 @@
...
+
+	var cmd = exec.CommandContext(ctx, args[0], args[1:]...)
+
+	// Run the command, passing in |input| and gathering output.
+	// This is equivalent to cmd.Output(), but uses our pre-allocated output buffer.
+	cmd.Stdin = bytes.NewReader(input)
+	cmd.Stdout = &stdout
+	cmd.Stderr = &stderr
+
+	if err := cmd.Run(); err != nil {
```
jixij, line 91: I am not very sure about this - while reading the code, I was wondering if it needs to detect and panic for `output overflows` when `err != nil`.

jixij (approved):
LGTM. Great to see `sops` integration with Flow.
Just have some non-blocking questions to help me understanding.

crates/sources/src/loader.rs
jgraettinger: The trouble comes if there are _multiple_ references to the config which are reached concurrently. The first call to `load_import` will indeed block and you could immediately inline it's result. But, the _second_ call observes that a fetch is already happening and will return immediately (not waiting for the fetch to complete).

It would be possible to rework this further and have the second call enqueue a one-shot channel on which it awaits, where the first call notifies all such channels on completion. This is a deeper change, though, which didn't seem worth it to me.

go/connector/decrypt.go
jgraettinger: Good point, we should do that check before handling this error.

examples/citi-bike/rides.config.yaml
jgraettinger: Yea. Today, `examples/` are doing double duty, as both examples and also integration tests.

There's a (very recently added) end-to-end test which is also run by CI. Today it's directly using this citi-bike example, and using sops here provides continuous coverage of that decryption.

But I'm not terribly happy with the end-to-end test in its current state (it's... okay), and would like to eventually revert this as part of improving that.

I'll note also that there are some other decryption tests that do require access to this same key. Not sure what to do about that, assuming we want CI coverage of `sops` usage 🤷

crates/sources/src/loader.rs
jixij: Got it. Thanks for the explanations.

PR #304, opened 2021-12-01
README: update the 'show me the code' example

jgraettinger:
Use the Citi Bike rides schema and projections,
and fix the materialization endpoint (`postgres` was removed).

Add a new "Run It" section with (current) directions.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/304)


README.md
```diff
@@ -38,72 +38,125 @@ With a Flow pipeline, you:
...
+storageMappings:
+  # Flow builds out data lakes for your collections in your cloud storage buckets.
+  # A storage mapping relates a prefix, like examples/citi-bike/, to a storage location.
+  # Here we tell Flow to store everything in one bucket.
+  "": { stores: [{ provider: S3, bucket: my-storage-bucket }] }
+```
+
+### Run It
 
+❗ These workflows are under active development and may change.
```
oliviamiannone, line 124: We should probably add a note here saying that they need to go through the setup process per the docs in order for this to work (might not be the case for codespaces, but I imagine it is in at least some setups. It is the way I'm working). That doc is linked in the top of the readme so pointing them there would suffice.

oliviamiannone: On a related note, it may be good to add codespaces setup to our current docs. I should also try it myself!

oliviamiannone (changes requested):
Only thing is the comment I left re: making sure setup is explicit enough. Other than that looks good to me, thanks for improving my code sample

PR #307, opened 2021-12-07
add network params

jixij:
Encountered the missing network paratemeres when writing tests for elastic search connector.
Although these changes are no longer needed, after some refactor to the connector test, it might worth adding these parameters for completeness. 
( Assuming there are no other considerations in setting the network parameters on the `temp-data-plane` and `comibine` commands.)

This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/307)


saterus (approved):
Looks good. :+1:

PR #308, opened 2021-12-09
Makefile: include `sops` in package artifact

jgraettinger:
Drop snowsql as it's no longer needed (moved to materialize-snowflake
connector).

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/308)


jgraettinger:
Merging so I can use this in `flow-template` without having to swap out the `:dev` tag.

PR #309, opened 2021-12-10
Add a pull request template

psFried:
**Description:**

Add a pull request template, which will pre-populate new pull request descriptions with this standard set of prompts. This PR is also an example of its usage! All subsequent PRs will start with these same headings, which can be filled in as appropriate. In particular, this addresses an important part of our documentation process by prompting us to write about anything that may require documentation work.

**Workflow steps:**

When you create a new PR (after this one is merged), the description should be pre-populated with these headings, and a brief reminder of what goes in each. These should be treated as writing prompts, to ensure that PR descriptions all address a minimal set of topics.

**Documentation links affected:** N/A

**Notes for reviewers:**

Feel free to suggest ways we can improve this!

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/309)

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,16 @@
+**Description:**
+
+(Describe the high level scope of new or changed features)
+
+**Workflow steps:**
+
+(How does one use this feature, and how has it changed)
+
+**Documentation links affected:**
+
+(list any documentation links that you created, or existing ones that you've identified as needing updates, along with a brief description)
+
+**Notes for reviewers:**
+
+(anything that might help someone review this PR)
+
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -8,7 +8,7 @@
 
 **Documentation links affected:**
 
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -0,0 +1,16 @@
...
+
+(Describe the high level scope of new or changed features)
+
+**Workflow steps:**
+
+(How does one use this feature, and how has it changed)
+
+**Documentation links affected:**
+
+(list any documentation links that you created, or existing ones that you've identified as needing updates, along with a brief description)
```
oliviamiannone, line 11: I am putting together a doc about docs here -- https://docs.google.com/document/d/1SRC9VS9zyCzWl3n4HXHbc4wPB1eLxJHkA2rtu9ZNokM/edit?usp=sharing. It's not done but might be useful to link in the template.

oliviamiannone (changes requested):
Overall structure is good. First 2 headings can be multi-purpose - they help engineering communicate, but also give me an easy way to do a conceptual run-through of high-level docs and see if there are any other changes to be made that aren't specifically linked. Other comment is the only change I'd make

PR #310, opened 2021-12-12
require that keyed locations not be `null`

jgraettinger:
**Description:**

Further restrict disallowed types of keyed locations to include NULL (in addition to object, array, and fractional).

**Workflow steps:**

* Catalogs which previously used null-able keyed locations will now error (I'm unaware of any)
* Attempts to use a nullable location as a key (a collection key or shuffle key) will produce an error:

```
$ flowctl check --source flow.yaml 
requests-rollup.flow.yaml error at /collections/acmeCo~1requests-rollup.v1 :
location /deal/id accepts "null", "string" in schema file:///workspaces/acmeCo/requests-rollup.schema.yaml, but "null" is disallowed in locations used as keys
```

https://docs.estuary.dev/concepts/catalog-entities/collections#allowed-key-field-types

Two updates to make to this section:
 * Null is no longer an allowed type for keyed locations
 * Flow *does* allow keyed locations to have multiple types (to be a string or an integer, for example). However certain materialization _connectors_ may prohibit keys participating in a materialization from taking multiple types. A SQL DB can't, for example, represent a primary key that could be either an INTEGER or TEXT, and the connector will therefore report an error at build time if you attempt to materialize the collection using such a connector. We may want to note that, in general, materialization connectors may impose extra typing restrictions which will become build-time errors, and give the example of a SQL primary key.

/cc @oliviamiannone 

**Notes for reviewers:**

Nada 

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/310)

### DIFF
--- a/crates/validation/src/errors.rs
+++ b/crates/validation/src/errors.rs
@@ -68,7 +68,7 @@ pub enum Error {
         type_: types::Set,
         schema: Url,
     },
-    #[error("location {ptr} accepts {type_:?} in schema {schema}, but {disallowed:?} is disallowed in locations used as keys")]
+    #[error("location {ptr} accepts {type_:?} in schema {schema}, but {disallowed:?} is disallowed in locations used as keys (https://go.estuary.dev/CigSvN)")]
     KeyWrongType {
         ptr: String,
         type_: types::Set,
--- a/crates/validation/src/schema.rs
+++ b/crates/validation/src/schema.rs
@@ -347,7 +347,7 @@ pub fn walk_keyed_location(
     }
 
     // Prohibit types not suited to being keys.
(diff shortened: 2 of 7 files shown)


saterus (approved):
Simple change. :+1: 

The only situation I can really think of for allowing null keys is when dealing with known-dirty data. Previously we could at least bring them into the collection and try to clean things up in a derivation, but this would require the user to clean them up on their end ahead of time. Though we could potentially skip bad rows if we're using the parser cli.

jgraettinger:
The user has an "out" even then:  if you have documents that are legit missing the intended natural key, such that you _must_ make them nullable and/or not required in the captured collection schema, then you can still key on the capture location metadata attached by the source connector (for example, the S3 filename and row offset) and filter / clean up in a derivation. The catalog I'm working with that motivated this change does exactly this.

PR #312, opened 2021-12-14
enable snapshot testing workflow

jgraettinger:
**Description:**

Support a snapshot testing workflow through:
1) Allowing `documents` of ingest and verity test steps to be a relative URL, in addition to the current inline array of documents. If given as a URL, the file MUST be a JSON array of document objects.
2) Offering a `--snapshot` flag in `flowctl test` & `flowctl api test`, which will generate nicely formatted JSON encodings (array of documents) of actual documents produced into a collection during the test run.

**Workflow steps:**

Before this PR, users were required to either a) edit document fixtures and expectations by hand within the catalog, or b) copy and past document fixtures into test and verification steps. Users (I) would copy terminal output and tediously re-format it into the catalog test which is under development.

Essentially the whole workflow presumes you're writing novel documents to ingest and verify by hand, which often you are, but certainly not always.

Now:

### I have pre-existing sample fixtures

A user needs to format their example documents as an encoded JSON array in a file. Then the test step names the file by a relative or absolute URL. It's permitted to link to files that exist externally to the current repo.

### I want to run the transformation and inspect its output

It's common to want to run a transformation over an input and then visually inspect it's output for correctness, rather than trying to craft the expected output ahead of time. This is now accomplished by running `flowctl test --snapshot ./snapshots`. Then, if a test verification fails, `flowctl` writes out what it _should_ have been to `./snapshots/the/test/name/verify-2.json`, where 2 is the index of the failed verification step.

In more detail, I might:
* Write a verification step of a test under development by using `documents: []` as a placeholder.
* Run `flowctl test --snapshot` to generate the snapshot of what it _actually_ is
* Inspect the output to ensure it's correctness.

When I'm satisfied that it's correct, I can prevent regressions by:
* Updating the verification step => `documents: ../path/to/snapshots/the/test/name/verify-2.json`
* Version the snapshot in git.

New shortlinks:

https://go.estuary.dev/NGT3es -> https://docs.estuary.dev/replace-me-with-docs-for-document-fixtures
https://go.estuary.dev/qpSUoq -> https://docs.estuary.dev/replace-me-with-docs-for-connector-configs

https://docs.estuary.dev/reference/catalog-reference/captures/endpoint-configurations
https://docs.estuary.dev/reference/catalog-reference/materialization/endpoints
https://docs.estuary.dev/concepts/catalog-entities/tests

**Notes for reviewers:**

Minior protocols update, which i'll merge at the same time.
https://github.com/estuary/protocols/pull/19

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/312)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -147,12 +147,15 @@ go-test-ci:   $(GO_BUILD_DEPS) ${GOBIN}/etcd ${GOBIN}/sops
 	./go.sh test -p ${NPROC} --tags "${GO_BUILD_TAGS}" --race --count=15 --failfast ./go/...
 
 .PHONY: catalog-test
-catalog-test: ${GOBIN}/flowctl ${GOBIN}/gazette ${GOBIN}/etcd ${GOBIN}/sops
+catalog-test: ${GOBIN}/flowctl ${GOBIN}/gazette ${GOBIN}/etcd ${GOBIN}/sops flow.schema.json
 	${GOBIN}/flowctl test --source ${ROOTDIR}/examples/local-sqlite.flow.yaml $(ARGS)
 
 .PHONY: end-to-end-test
 end-to-end-test: ${GOBIN}/flowctl ${GOBIN}/gazette ${GOBIN}/etcd ${GOBIN}/sops
 	PATH="${PATH}:${GOBIN}" ./tests/run-end-to-end.sh
+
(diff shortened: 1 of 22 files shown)


jixij (approved):
LGTM!  The new workflow for producing and using the fixtures in test flow looks great.

PR #314, opened 2021-12-16
flowctl: remove --builds-root flag

jgraettinger:
**Description:**

`flowctl` commands which interact with data planes, now ask the data plane what it's configured `--flow.builds-root` is, instead of requiring a `--builds-root` argument to be explicitly passed.

Similarly, `flowctl temp-data-plane` no longer takes a --builds-root, and instead uses a `./builds/` subdirectory of the temporary directory which is provisioned for the data plane.

**Workflow steps:**

Where before you passed `--builds-root`, now you don't.

The top-level README example is updated, and updates to the `flow-template` repository are required.
To my knowledge we hadn't documented this elsewhere yet.

**Notes for reviewers:**

Enjoy the fewer flags.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/314)

### DIFF
--- a/tests/run-end-to-end.sh
+++ b/tests/run-end-to-end.sh
@@ -35,9 +35,6 @@ ACTUAL="${TESTDIR}/actual_test_results.txt"
 # Expected materialization output.
 EXPECTED="${ROOTDIR}/tests/end-to-end.expected"
 
-# `flowctl` commands look for a BUILDS_ROOT environment variable which roots
-# build databases known to the data plane.
-export BUILDS_ROOT="file://${TESTDIR}/build/"
 # `flowctl` commands which interact with the data plane look for *_ADDRESS
 # variables, which are used by the temp-data-plane we're about to start.
 export BROKER_ADDRESS=unix://localhost${TESTDIR}/gazette.sock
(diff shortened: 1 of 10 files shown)


tests/run-end-to-end.sh
```diff
@@ -55,13 +52,25 @@ flowctl temp-data-plane \
     --unix-sockets \
     &
 DATA_PLANE_PID=$!
+
+# `flowctl temp-data-plane` always uses ./builds/ of --tempdir as its --flow.builds-root.
+# See cmd-temp-data-plane.go.
+export BUILDS_ROOT=${TESTDIR}/builds
```
psFried, line 58: This can be removed now, right?

psFried (changes requested):
The `BUILDS_ROOT` was removed, and there was much rejoicing!
This looks good to me % the one comment.

tests/run-end-to-end.sh
jgraettinger: It's still used, but only locally within the script as part of the `mv` below.

jgraettinger:
Thanks

PR #316, opened 2021-12-16
Publish task stats to ops/<tenant>/stats

psFried:
**Description:**

Statistics from all tasks are now published to the corresponding stats
collection. For captures and derivations, these stats are published
as part of the same transaction that writes their outputs, and thus the
stats are guaranteed to accurately reflect the documents that were
actually committed to the collections.

Materialization stats are currently published on a best-effort basis,
because the stats themselves aren't known until after `StartCommit` has
already been called. Making these transactional is being deferred until
a later point.

In any case, the published stats can be used as a source in both
derivations and materializations, and generally treated just like any
other flow collection. The end-to-end test was modified to include
materialization of both logs and stats.

**Workflow steps:**

The short answer is that the stats collection can be used in derivations and materializations just like any other. Flow takes care of publishing the stats automatically, so all users have to do is read them. There's certainly more that we should address in docs, though, like how to derive common high-level stats from these documents, and generating documentation from the schemas.

I didn't create any short links for this, because it's not entirely clear where they would go. The ops schemas is the only place that comes to mind, but those aren't very visible to end users. I'm open to suggestions, though.

**Notes for reviewers:**

This PR builds on several prior PRs, most notably #293, which implements the collection of stats.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/316)


jgraettinger (changes requested):
Cool!

I'm still reviewing (just skimmed so far), but what about having each of runtime/capture.go, runtime/materialization.go etc pass the task-specific formatting bits into StatsFormatter ?

This would enable rolling back task_term.go having anything to do with the specific task specification, wouldn't it? I think that's desirable -- that only materialization.go deals with pf.MaterializationSpec.

> Reviewed 7 of 19 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/316#-:-Mr9Vaa0CShSHWbYt_cY:bvjzmhf)* status: 7 of 19 files reviewed, 1 unresolved discussion (waiting on @jgraettinger and @psFried)

___
*[go/runtime/task_term.go, line 108 at r1](https://reviewable.io/reviews/estuary/flow/316#-Mr9UQeI2tn_EyP4ewKW:-Mr9UQeI2tn_EyP4ewKX:b-dng8ht) ([raw file](https://github.com/estuary/flow/blob/a34d11d574b971ce01e91d1dcce6d5bc33fc9449/go/runtime/task_term.go#L108)):*
> ```Go
> 	}
> 	if t.LogPublisher.Level() >= log.DebugLevel {
> 		logFields["taskSpec"] = taskSpec
> ```

this is a pretty big struct, I don't think it ever makes sense with every logged field. IIRC we already log its definition when the task is initialized.

psFried (commented):
I actually think that having task_term.go load the specs is the better way to go. Previously, we had to repeat the same code in all three `RestoreCheckpoint` impls, and now we have it consolidated into task term, which is where I'd expect to find it, just based on the name. Is there a downside to that that I'm not seeing?

I could take a look at other factorings for task-specific stats formatting, but I do like having the task spec loading all be handled in the task_term.

> *[Reviewable](https://reviewable.io/reviews/estuary/flow/316#-:-MrPFZ8r6rkWNsRwYaOp:b-63ty43)* status: 7 of 19 files reviewed, 1 unresolved discussion (waiting on @jgraettinger)

___
*[go/runtime/task_term.go, line 108 at r1](https://reviewable.io/reviews/estuary/flow/316#-Mr9UQeI2tn_EyP4ewKW:-MrPEGR20M0wz76-tOWT:b3bpy51) ([raw file](https://github.com/estuary/flow/blob/a34d11d574b971ce01e91d1dcce6d5bc33fc9449/go/runtime/task_term.go#L108)):*

this is a pretty big struct, I don't think it ever makes sense with every logged field. IIRC we already log its definition when the task is initialized.
</blockquote></details>

Yeah, this is only logged once, not on every field. The runtimes were each logging the task spec after it was loaded, so I just move that logging into this file to DRY it up a bit. Previously it was being logged at the debug level, which I think made sense given the size of the struct, so I just made sure to still only log it if debug logging is enabled. Of course I realized I forgot to delete the old logging from capture.go, so I'll push that up now.

jgraettinger (commented):
>  Previously, we had to repeat the same code in all three RestoreCheckpoint impls

This isn't quite right. Right now, in what's checked in, `runtime.Materialize` has a `pf.MaterializeSpec`, `runtime.Derive` has a `pf.DerivaitonSpec`, and `runtime.Capture` has a `pf.CaptureSpec`. Code paths to load those specifications are static and specific to the specification type (catalog.LoadDerivation vs catalog.LoadCapture, etc). `task_term.go` has only things which are truly shared across all task types. A static runtime type holds a static task, which is dynamically upcast to a pf.Task interface as needed.

In this PR, it's switched to a static runtime type holding a dynamic task, which is dynamically downcast to the task-specific type as needed. This is what I'm saying is undesirable.

My baggage here is that I recently did a bunch of refactor work to ensure we _didn't_ have this pattern, and that everything was statically checked. I'm no-doubt sensitive for that reason.

We have divergent views on the role of `task_term.go` which I'll try and put my finger on: You're arguing for it as a place for handling `pf.Task` generally, with dynamic handling of per-variant behavior. My argument is that anything specific to, say, derivations belongs in `runtime/derivation.go`. task_term.go` is a necessary evil for some truly repeated bits of stuff we want to DRY, but we should endeavor to keep as small as possible.

would it help to change the name of `task_term.go`? It's not super well named today.

> *[Reviewable](https://reviewable.io/reviews/estuary/flow/316#-:-MrT2bIc16Ve1v-6h8tM:b-pjnvz8)* status: 7 of 19 files reviewed, 1 unresolved discussion (waiting on @jgraettinger)

psFried (commented):
Ok, well, I was actually going to walk back some of that, anyway, because it started to look more desirable to rip the task spec out of the `StatsFormatter` and the `taskTerm` both, as I got into it.

I do think we have different views of the role of `task_term.go`. Some of that, I'll attribute to my instincts not translating well into a language without generics (if we had generics, I'd likely want to parameterize the `taskTerm` with the type of task). Some of the differences are maybe also attributable to my having a general sense that the current factoring of runtime types is less than ideal. I generally feel that our three task types are more alike than different, and that we may eventually try to unify them under a single type. TBH I'm really trying _not_ to focus on that right now, though, since we have a lot of other priorities. But I do have a sense that a better factoring may become apparent as we delve more deeply into (derivation|runtimey| :wave: ) things.

Anyway, I just pushed a refactor. LMK what you think of it now.

> *[Reviewable](https://reviewable.io/reviews/estuary/flow/316#-:-MrTgvhc7ot-YajeeW5y:b-babijw)* status: 7 of 22 files reviewed, 1 unresolved discussion (waiting on @jgraettinger)

psFried (commented):
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/316)* status: 7 of 22 files reviewed, 1 unresolved discussion (waiting on @jgraettinger)

___
*[go/runtime/task_term.go, line 108 at r1](https://reviewable.io/reviews/estuary/flow/316#-Mr9UQeI2tn_EyP4ewKW:-MrVF0Ek135IKppXTqfY:b-1hzgck) ([raw file](https://github.com/estuary/flow/blob/a34d11d574b971ce01e91d1dcce6d5bc33fc9449/go/runtime/task_term.go#L108)):*

Yeah, this is only logged once, not on every field. The runtimes were each logging the task spec after it was loaded, so I just move that logging into this file to DRY it up a bit. Previously it was being logged at the debug level, which I think made sense given the size of the struct, so I just made sure to still only log it if debug logging is enabled. Of course I realized I forgot to delete the old logging from capture.go, so I'll push that up now.
</blockquote></details>

My earlier comment was written before I refactored this. Now that the task spec loading has been removed from task_term.go, this is no longer a thing.

jgraettinger (approved):
Thanks! LGTM % feedback below.

> Reviewed 6 of 19 files at r1, 9 of 9 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/316#-:-MsjmGah5GznVY7dPehG:b-pg43b6)* status: all files reviewed, 12 unresolved discussions (waiting on @psFried)

___
*[go/bindings/combine.go, line 196 at r2](https://reviewable.io/reviews/estuary/flow/316#-Msjd7Va2_fI9UaEcdNI:-Msjd7Vb9-MaQyPnbfcR:b-yg73s1) ([raw file](https://github.com/estuary/flow/blob/3d6648d9fb5e89073337594ab82845bd5191f31e/go/bindings/combine.go#L196)):*
> ```Go
> }
> 
> /*
> ```

Q: what was the thought process behind rolling these up to the process level, instead of breaking out on shard as we were ? Was there a problem in having them broken out by shard ?
___
*[go/runtime/capture.go, line 38 at r2](https://reviewable.io/reviews/estuary/flow/316#-MsjdtFmA2SNZ2s0xDZo:-MsjdtFmA2SNZ2s0xDZp:b-dwfywg) ([raw file](https://github.com/estuary/flow/blob/3d6648d9fb5e89073337594ab82845bd5191f31e/go/runtime/capture.go#L38)):*
> ```Go
> 	// Specification under which the capture is currently running.
> 	// This is duplicated from the task term to avoid needing type assertions on each usage.
> 	spec *pf.CaptureSpec
> ```

nit: this isn't duplicated in the task term though?

nit: Is there a reason for making this a pointer? It didn't used to be, and AFAIK doesn't need to be nullable
___
*[go/runtime/capture.go, line 68 at r2](https://reviewable.io/reviews/estuary/flow/316#-MsjeHTS85RGDj2NhI8s:-MsjeHTS85RGDj2NhI8t:bwkli6t) ([raw file](https://github.com/estuary/flow/blob/3d6648d9fb5e89073337594ab82845bd5191f31e/go/runtime/capture.go#L68)):*
> ```Go
> 		return pf.Checkpoint{}, err
> 	}
> 	err = c.build.Extract(func(db *sql.DB) error {
> ```

nit: this got reordered as compared to what it was. The previous ordering was intentional and is preferred, I believe, because it ensures the defer traps logs related to loading the specification.
___
*[go/runtime/capture.go, line 79 at r2](https://reviewable.io/reviews/estuary/flow/316#-MsjeVvj91qUWV6dE6LH:-MsjeVvj91qUWV6dE6LI:bnylqgi) ([raw file](https://github.com/estuary/flow/blob/3d6648d9fb5e89073337594ab82845bd5191f31e/go/runtime/capture.go#L79)):*
> ```Go
> 		if err == nil {
> 			c.Log(log.DebugLevel, log.Fields{
> 				"capture":    c.spec,
> ```

nit: this will log the entire protobuf structure instead of the capture name. This reads as unintentional to me (it's a common mistake I've made, to use the specification when I meant the specification's contained name); but that may be what you were after. There used to be a separate log statement which focused on logging the specification which I think is preferred though.

derivation & materialization seem to have the same concern
___
*[go/runtime/capture.go, line 342 at r2](https://reviewable.io/reviews/estuary/flow/316#-MsjjH7pCH740sPId46T:-MsjjH7pCH740sPId46U:b947cbb) ([raw file](https://github.com/estuary/flow/blob/3d6648d9fb5e89073337594ab82845bd5191f31e/go/runtime/capture.go#L342)):*
> ```Go
> 	var captureStats = make(map[string]CaptureBindingStats)
> 	for i, bindingStats := range statsPerBinding {
> 		if bindingStats != nil { // Skip bindings that didn't participate
> ```

supernit: `if bindingStats == nil { continue }` to reduce indent level of important code.
___
*[go/runtime/derive.go, line 32 at r2](https://reviewable.io/reviews/estuary/flow/316#-MsjjbXp6QO1fGdSk0yI:-MsjjbXp6QO1fGdSk0yJ:b-ot70ic) ([raw file](https://github.com/estuary/flow/blob/3d6648d9fb5e89073337594ab82845bd5191f31e/go/runtime/derive.go#L32)):*
> ```Go
> 	recorder *recoverylog.Recorder
> 	// Active derivation specification, updated in RestoreCheckpoint.
> 	// This is duplicated from the task term to avoid needing type assertions on each usage.
> ```

nit: no longer duplicated.

I see and note that derivation was already nil-able here, which is different from capture or materialization. That's probably an error of omission on my part, the intent was to make them non-nullable. Just noting, no action.
___
*[go/runtime/derive.go, line 66 at r2](https://reviewable.io/reviews/estuary/flow/316#-Msjg6Om56AmOTp2_9Cm:-Msjg6Om56AmOTp2_9Cn:b-c2o6bd) ([raw file](https://github.com/estuary/flow/blob/3d6648d9fb5e89073337594ab82845bd5191f31e/go/runtime/derive.go#L66)):*
> ```Go
> 		return pf.Checkpoint{}, err
> 	}
> 	err = d.build.Extract(func(db *sql.DB) error {
> ```

nit: same here; this was re-ordered and will no longer trap log errors related to loading the collection.
___
*[go/runtime/derive.go, line 77 at r2](https://reviewable.io/reviews/estuary/flow/316#-MsjjZTe0iNcZqjeENC5:-MsjjZTe0iNcZqjeENC6:bgb82gn) ([raw file](https://github.com/estuary/flow/blob/3d6648d9fb5e89073337594ab82845bd5191f31e/go/runtime/derive.go#L77)):*
> ```Go
> 		if err == nil {
> 			d.Log(log.DebugLevel, log.Fields{
> 				"derivation": d.derivation,
> ```

nit: same; also logging full structure.
___
*[go/runtime/derive.go, line 193 at r2](https://reviewable.io/reviews/estuary/flow/316#-MsjkJzO5JookagD1jZK:-MsjkJzO5JookagD1jZL:b-hrg5t8) ([raw file](https://github.com/estuary/flow/blob/3d6648d9fb5e89073337594ab82845bd5191f31e/go/runtime/derive.go#L193)):*
> ```Go
> 	var includesUpdate = false
> 	for i, tf := range txnStats.Transforms {
> 		// Don't include transforms that didn't participate in this transaction.
> ```

supernit: `continue` to de-indent important code.
___
*[go/runtime/materialize.go, line 34 at r2](https://reviewable.io/reviews/estuary/flow/316#-MsjgaQHErZ4ZBI6u5cc:-MsjgaQHErZ4ZBI6u5cd:b78eww4) ([raw file](https://github.com/estuary/flow/blob/3d6648d9fb5e89073337594ab82845bd5191f31e/go/runtime/materialize.go#L34)):*
> ```Go
> 	// Specification under which the materialization is currently running.
> 	// This is duplicated from the task term to avoid needing type assertions on each usage.
> 	spec *pf.MaterializationSpec
> ```

nit: this isn't duplicated.
nit: this is nil-able when it wasn't before and doesn't need to be.
___
*[go/runtime/materialize.go, line 35 at r2](https://reviewable.io/reviews/estuary/flow/316#-MsjkfDL-q1JD0gLTCd0:-MsjgU6d-1ojlNKyuBYV:bkk4uxh) ([raw file](https://github.com/estuary/flow/blob/3d6648d9fb5e89073337594ab82845bd5191f31e/go/runtime/materialize.go#L35)):*
> ```Go
> 	// This is duplicated from the task term to avoid needing type assertions on each usage.
> 	spec *pf.MaterializationSpec
> 	// Stats are handled differently for materializations than they are for other task types,
> ```

You don't need to create your own Publisher; you can use the shard's on demand as `shard.Publisher()`
___
*[go/runtime/materialize.go, line 79 at r2](https://reviewable.io/reviews/estuary/flow/316#-MsjgtBQ4uyoYw9mgoJE:-MsjgtBQ4uyoYw9mgoJF:bejl49r) ([raw file](https://github.com/estuary/flow/blob/3d6648d9fb5e89073337594ab82845bd5191f31e/go/runtime/materialize.go#L79)):*
> ```Go
> 		return pf.Checkpoint{}, err
> 	}
> 	err = m.build.Extract(func(db *sql.DB) error {
> ```

nit: reordered as with others.

psFried (commented):
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/316)* status: 16 of 22 files reviewed, 6 unresolved discussions (waiting on @jgraettinger)

___
*[go/bindings/combine.go, line 196 at r2](https://reviewable.io/reviews/estuary/flow/316#-Msjd7Va2_fI9UaEcdNI:-MskpGkCAdK2IKKG_MUz:b4aprsh) ([raw file](https://github.com/estuary/flow/blob/3d6648d9fb5e89073337594ab82845bd5191f31e/go/bindings/combine.go#L196)):*

Q: what was the thought process behind rolling these up to the process level, instead of breaking out on shard as we were ? Was there a problem in having them broken out by shard ?
</blockquote></details>

There wasn't a problem, but it seemed needlessly duplicative, since the ops collections already break things out per shard. My thought was that if we want stats on a per-shard basis, then it would be best to use the ones from the ops collections. Part of my reasoning there was also due to my belief (i'm not actually certain this is right) that having a large number of distinct timeseries is relatively much more costly than having a large number of samples in fewer timeseries. My intention is also to follow this up pretty soon with some way to easily consume the new stats, since the current grafana dashboards have fallen into disrepair anyway.
___
*[go/runtime/capture.go, line 38 at r2](https://reviewable.io/reviews/estuary/flow/316#-MsjdtFmA2SNZ2s0xDZo:-MsktVbaBu4mMe7SW2Wc:bltoirp) ([raw file](https://github.com/estuary/flow/blob/3d6648d9fb5e89073337594ab82845bd5191f31e/go/runtime/capture.go#L38)):*

nit: this isn't duplicated in the task term though?

nit: Is there a reason for making this a pointer? It didn't used to be, and AFAIK doesn't need to be nullable
</blockquote></details>

whoops, yeah I forgot to update these comments after the last round of changes. They are no longer part of the task term. I'm getting rid of the pointers, as we discussed.
___
*[go/runtime/capture.go, line 68 at r2](https://reviewable.io/reviews/estuary/flow/316#-MsjeHTS85RGDj2NhI8s:-Msky9Vd3rZX2BCvJ5r_:bb7f2mf) ([raw file](https://github.com/estuary/flow/blob/3d6648d9fb5e89073337594ab82845bd5191f31e/go/runtime/capture.go#L68)):*

nit: this got reordered as compared to what it was. The previous ordering was intentional and is preferred, I believe, because it ensures the defer traps logs related to loading the specification.
</blockquote></details>

done
___
*[go/runtime/capture.go, line 79 at r2](https://reviewable.io/reviews/estuary/flow/316#-MsjeVvj91qUWV6dE6LH:-MskyEACE5-w1h-KwTtg:b-uugmhl) ([raw file](https://github.com/estuary/flow/blob/3d6648d9fb5e89073337594ab82845bd5191f31e/go/runtime/capture.go#L79)):*

nit: this will log the entire protobuf structure instead of the capture name. This reads as unintentional to me (it's a common mistake I've made, to use the specification when I meant the specification's contained name); but that may be what you were after. There used to be a separate log statement which focused on logging the specification which I think is preferred though.

derivation & materialization seem to have the same concern
</blockquote></details>

This was intentional, to combine the two separate log statements into one, for all three runtime files, and to include the spec in the logging of the outcome.
___
*[go/runtime/capture.go, line 342 at r2](https://reviewable.io/reviews/estuary/flow/316#-MsjjH7pCH740sPId46T:-MslHCW2CZFcWcW_Lo02:b-896fix) ([raw file](https://github.com/estuary/flow/blob/3d6648d9fb5e89073337594ab82845bd5191f31e/go/runtime/capture.go#L342)):*

supernit: `if bindingStats == nil { continue }` to reduce indent level of important code.
</blockquote></details>

Done.
___
*[go/runtime/derive.go, line 193 at r2](https://reviewable.io/reviews/estuary/flow/316#-MsjkJzO5JookagD1jZK:-MslHAF495kbghf1rSpb:b-896fix) ([raw file](https://github.com/estuary/flow/blob/3d6648d9fb5e89073337594ab82845bd5191f31e/go/runtime/derive.go#L193)):*

supernit: `continue` to de-indent important code.
</blockquote></details>

Done.
___
*[go/runtime/materialize.go, line 35 at r2](https://reviewable.io/reviews/estuary/flow/316#-MsjkfDL-q1JD0gLTCd0:-MslF-xG6gNUL83AGoTM:bextym4) ([raw file](https://github.com/estuary/flow/blob/3d6648d9fb5e89073337594ab82845bd5191f31e/go/runtime/materialize.go#L35)):*

You don't need to create your own Publisher; you can use the shard's on demand as `shard.Publisher()`
</blockquote></details>

I'm not seeing that function. The shard only exposes a `JournalClient()` function, which is what I'm using to create this publisher. I do see a `Publisher` that's passed as an argument to `ConsumeMessage` and `FinalizeTxn`, but it's not passed to `StartCommit`, which is where it's being used. I'm leaving this as is for now, but LMK there's something I'm missing.

jgraettinger (approved):
> Reviewed 6 of 6 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/316)* status: all files reviewed, 1 unresolved discussion (waiting on @psFried)

___
*[go/runtime/materialize.go, line 35 at r2](https://reviewable.io/reviews/estuary/flow/316#-MsjkfDL-q1JD0gLTCd0:-MslPSa16cmzt5QVCAeB:b-hqho7i) ([raw file](https://github.com/estuary/flow/blob/3d6648d9fb5e89073337594ab82845bd5191f31e/go/runtime/materialize.go#L35)):*

I'm not seeing that function. The shard only exposes a `JournalClient()` function, which is what I'm using to create this publisher. I do see a `Publisher` that's passed as an argument to `ConsumeMessage` and `FinalizeTxn`, but it's not passed to `StartCommit`, which is where it's being used. I'm leaving this as is for now, but LMK there's something I'm missing.
</blockquote></details>

D'oh, right. Oh well.

PR #324, opened 2022-01-05
Add bulk ingestion docs

oliviamiannone:
**Description:**

Just adding docs from https://github.com/estuary/connectors/pull/91 with significant re-writing on the conceptual side; need to ensure it's still factual.

**Workflow steps:**

NA

No shortlinks

**Notes for reviewers:**

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/324)


site/docs/reference/Connectors/materialization-connectors/Rockset.md
```diff
@@ -42,4 +42,123 @@ materializations:
...
+### Prerequisites
+Before completing this workflow, make sure you have:
+* A working catalog spec including at least one Flow collection with its schema specified.
+* A production or [development](../../../getting-started/installation.md) environment
+::: Tip
+The following is an intermediate workflow. As needed, refer to [this guide](../../../guides/create-dataflow.md) for the basic steps to create and deploy a catalog spec using the GitOps workflow.
+:::
+
+### How to configure a bulk ingestion
+If you already have an [S3 integration](https://rockset.com/docs/amazon-s3/) in Rockset, you can simply direct the connector to it. The connector will create the collection from the given integration, process all objects in the S3 bucket first, and then begin writing additional any additional, newer data that is coming through your pipeline. This ensures that documents are always written to Rockset in the proper order.
```
psFried, line 59: I think this is sorta right in terms of the conceptual understanding of how the connector itself works, but it's not a use case we should document here. The garden path is to start with all your data in a Flow collection, and wanting to materialize that into Rockset, which requires the two step process. _Technically_ you could imagine some use case where you have some other process (besides Flow) that's populating an S3 bucket that you want to bulk ingest, but even in that case, it's not clear how that data should be ordered with respect to the data in the Flow collection. Basically, it'd be a pretty "advanced" use case, with a number of foot-guns.

I think a better approach in terms of this doc would be to explain (maybe at the bottom, even) that you can always create the Rockset collection's yourself, using whatever parameters you want, and that the connector will simply use that collection as long as it has the same name. Basically, documenting that there's an "escape hatch" for Rockset power users. In that case, one _could_ create the collection using whatever integration, and decide to also name the integration in the Flow connector config _if_ they need it await completion of the integration before materializing.

site/docs/reference/Connectors/materialization-connectors/Rockset.md
```diff
@@ -42,4 +42,123 @@ materializations:
...
+            collection: <your rockset collection name>
+            initializeFromS3:
+              integration: <rockset integration name>
+              bucket: example-s3-bucket
+              region: us-east-1
+              prefix: example/s3-prefix/
+          source: example/flow/collection
+  ```
+
+If you don't have an S3 integration set up with your historical data, you can use a two-step process to bulk ingest historical data from a Flow collection. First, you'll port your historical data into an S3 bucket and backfill it into Rockset using Estuary's [materialize-s3-parquet](../materialize-s3-parquet/) connector. Then, you'll implement the Rockset connector, which will pick up the dataflow where the S3-parquet connector left off.
```
psFried, line 85: Wording nits:

> First, you'll ~~port~~ write your historical data into an S3 bucket ~~and backfill it into Rockset~~ using Estuary's

(The backfill into Rockset part technically comes later)

> Then, you'll ~~implement~~ switch to using the Rockset connector

site/docs/reference/Connectors/materialization-connectors/Rockset.md
```diff
@@ -42,4 +42,123 @@ materializations:
...
+            initializeFromS3:
+              integration: <rockset integration name>
+              bucket: example-s3-bucket
+              region: us-east-1
+              prefix: example/s3-prefix/
+          source: example/flow/collection
+  ```
+5. When you activate the new materialization, the connector will create the Rockset collection using the given integration, and wait for it to ingest all of the data from S3 before it continues. During this time, the Flow shards will remain in `STANDBY` status, so `flowctl deploy` is expected to block until the bulk ingestion completes. Once this completes, the materialize-rockset connector will automatically switch over to using the Rockset write API.
+
+## Potential improvements
```
psFried, line 151: The README in the connectors repo was written more with a developer audience in mind. I think this section is better left out of this, more user-facing, doc. This could be a good place to document the "escape hatch" described above, though. Maybe something like:

If you want complete control over how the Rockset Collection is created, then you can always create it yourself in the Rockset console, and Flow's materialize-rockset collection will happily use it as long as the name matches.

psFried (commented):
I left some feedback inline. I think there was perhaps some confusion due to the original README in the connectors repo being a little more geared toward developers rather than users. I think we'll generally want to stick with having the more dev-facing docs be in the READMEs next to each connector, and having more user-facing docs live here. Just in case that helps provide some missing context when interpreting those READMEs.

site/docs/reference/Connectors/materialization-connectors/Rockset.md
```diff
@@ -42,4 +42,123 @@ materializations:
       	workspace: ${namespace_name}
       	collection: ${table_name}
     source: ${tenant}/${source_collection}
+```
+## Bulk ingestion for large backfills of historical data
+
+You can backfill large amounts of historical data into Rockset using a *bulk ingestion*. Bulk ingestion must originate in S3 and requires additional steps in your dataflow. Flow's Rockset connector supports this through the GitOps workflow.
+
+### Prerequisites
```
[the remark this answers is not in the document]
oliviamiannone: Good to know, I'll start doing that!

site/docs/reference/Connectors/materialization-connectors/Rockset.md
oliviamiannone: Ohhhkay. Thank you for reframing all of this for me. I see what you're saying. Given that, it's probably best to not document it at all. It's still possible, and potentially an advanced Rockset user would figure it out and try it on their own. There's always workarounds like that... but as the company we should only document the recommended workflow.

site/docs/reference/Connectors/materialization-connectors/Rockset.md
oliviamiannone: As above, I just took the whole thing out. In general, I've simplified this pretty heavily so as not to provide more info than necessary to the end user. Maybe give it one more read when you can, and then I'll merge?

site/docs/reference/Connectors/materialization-connectors/Rockset.md
```diff
@@ -12,9 +13,11 @@ To use this connector, you'll need :
     * Optional; if none exist, one will be created by the connector.
 
 ## Configuration
+
 To use this connector, begin with a Flow catalog that has at least one **collection**. You'll add a Rockset materialization, which will direct one or more of your Flow collections to your desired Rockset collections. Follow the basic [materialization setup](../../../concepts/materialization.md) and add the required Rockset configuration values per the table below.
```
psFried, line 17: uber nit, but I'll ask just to get your thoughts on it: should we have the `**collection**` here be a link to our docs on Flow collections or how to create them?

site/docs/reference/Connectors/materialization-connectors/Rockset.md
```diff
@@ -42,4 +45,84 @@ materializations:
...
\ No newline at end of file
+```
+## Bulk ingestion for large backfills of historical data
+
+You can backfill large amounts of historical data into Rockset using a *bulk ingestion*. Bulk ingestion must originate in S3 and requires additional steps in your dataflow. Flow's Rockset connector supports this through the GitOps workflow.
+
+### Prerequisites
+
+Before completing this workflow, make sure you have:
+* A working catalog spec including at least one Flow collection with its schema specified.
```
psFried, line 56: nit: I think I'd remove "with its schema specified", since that's automatically the case if you have a collection.

site/docs/reference/Connectors/materialization-connectors/Rockset.md
```diff
@@ -42,4 +45,84 @@ materializations:
...
+* A production or [development](../../../getting-started/installation.md) environment
+::: tip
+The following is an intermediate workflow. As needed, refer to [this guide](../../../guides/create-dataflow.md) for the basic steps to create and deploy a catalog spec using the GitOps workflow.
+:::
+
+### How to configure a bulk ingestion
+
+A bulk ingestion from a Flow collection into Rockset is essentially a two-step process. First, Flow writes your historical data into an S3 bucket using Estuary's [materialize-s3-parquet](../materialize-s3-parquet/) connector. Once the data is caught up, it uses the Rockset connector to backfill the data from S3 into Rockset and then switch to the Rockset Write API for the continuous materialization of new data.
+
+import Mermaid from '@theme/Mermaid';
```
psFried, line 66: I'm not positive, but I think this `import` might need to be at the top of the file. Or at least it probably _should_ be at the top to be consistent with prevailing conventions.

site/docs/reference/Connectors/materialization-connectors/Rockset.md
```diff
@@ -42,4 +45,84 @@ materializations:
...
+### Prerequisites
+
+Before completing this workflow, make sure you have:
+* A working catalog spec including at least one Flow collection with its schema specified.
+* A production or [development](../../../getting-started/installation.md) environment
+::: tip
+The following is an intermediate workflow. As needed, refer to [this guide](../../../guides/create-dataflow.md) for the basic steps to create and deploy a catalog spec using the GitOps workflow.
+:::
+
+### How to configure a bulk ingestion
```
psFried, line 62: nit: I think "perform" may be a better word than "configure" here, since this is a process that requires human interaction at several points, rather than someone you set and forget.

psFried (approved):
I think this seems in good shape. I had a few minor nits that I left comments on, but nothing blocking.

site/docs/reference/Connectors/materialization-connectors/Rockset.md
oliviamiannone: It does work in the test output, and I based this off of how Johnny had set up his diagrams in some of the docs he created... that is a good point about the convention but I think I'll leave it for now because of those other files

site/docs/reference/Connectors/materialization-connectors/Rockset.md
oliviamiannone: That's a good question. Gave it some thought, and I think it is safe to assume that by the time the reader gets to this point, they'll know what a collection is. Just by nature of the pathway, they'll have already created a capture and be familiar from that perspective. Plus it's such a high-level concept that if we linked it in every doc it's mentioned that could get link-heavy

site/docs/reference/Connectors/materialization-connectors/Rockset.md
oliviamiannone: It does work in the test output, and I based this off of how Johnny had set up his diagrams in some of the docs he created... that is a good point about the convention but I think I'll leave it for now because of those other files

PR #326, opened 2022-01-06
flowctl: add `journals split` subcommand

jgraettinger:
**Description:**

Splitting journals is similar to, but simpler than splitting shards.
It's done by evenly sub-dividing the key range of the collection partitions,
splitting a parent collection partition into child partitions each having
a fraction of the responsible range.

As journals don't have recovery logs to content with, this tool can be a bit
more aggressive in allowing more than one forked child in a go. However, we
do still constrain it to be a power-of-two.

**Workflow steps:**

```bash
# Split into two:
flowctl journals split --journal the/name/of/a/collection/journal/pivot=00

# Split into eight:
flowctl journals split --journal the/name/of/a/collection/journal/pivot=00 --splits 8
```

We have yet to document how to use flowctl to split things.
It's unclear if we will, given our intention to automate this.

**Notes for reviewers:**

Nada

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/326)


go/flow/converge.go
```diff
@@ -164,6 +165,50 @@ func MapShardToSplit(task pf.Task, shards []pc.ListResponse_Shard, splitOnKey bo
 	return []pf.LabelSet{lhs, rhs}, nil
 }
 
+// MapPartitionToSplit maps a single partition journal contained in |journals|
+// to a desired split state, where the partition is evenly subdivided into
+// |splits| sub-partitions. |splits| must be a power-of-two.
+func MapPartitionToSplit(collection *pf.CollectionSpec, journals []pb.ListResponse_Journal, splits uint) ([]pf.LabelSet, error) {
+	if len(journals) != 1 {
+		return nil, fmt.Errorf("expected exactly one journal in the response")
+	} else if splits < 2 || bits.OnesCount(splits) != 1 {
```
willdonnelly, line 174: Should there be an upper bound on `splits`? Possibly checked down below after `range_` is computed, because the obvious sanity check would be that you have fewer splits than there are possible keys?

willdonnelly (approved):
LGTM, only one nit regarding the edge case where splits > range.

go/flow/converge.go
jgraettinger: Done

Issue #335, opened 2022-01-13
flowctl: improve `flowctl discover` ergonomics


jgraettinger:
`flowctl discover` is currently awkward for a few reasons:

### ~It presumes a catalog prefix of `acmeCo/`~
_Not true, it has (had) a --prefix flag._

This was always a placeholder, meant to establish the convention of using a leading company (tenant) name,
but it requires manually editing all the generated capture and all associated collections.

**Desired**: a chosen prefix should be taken as a parameter.

### It dumps its output into a single Flow catalog source file

This can include _many_ collection definitions, all with inline schemas which can grow quite large.
It also includes a capture specification with an inline configuration, which is difficult to then protect with `sops`.

**Desired**: the tool should output an organized layout of collection schemas and endpoint configuration, all under a subdirectory created for the capture, which are then referenced from a "flow.yaml" in that subdirectory.

### It cannot be re-run for an endpoint

We write a stub configuration, and then _move_ it into the generated catalog source.
This makes it really awkward to re-run discovery. We will also completely blow away any edits included in your pre-existing catalog source.

**Desired**: the tool should look for a `the-capture-dir/endpoint.config.yaml` file in the capture's directory:
 * If it doesn't exist, a stub file is created and the user is notified
 * If it does exist, it's used to drive discovery, and associated files are updated
   * The user can restrict discovery to a regex (?) of endpoint resource paths, to selectively update only some collections / bindings.

### It doesn't respect configuration defaults

From inference, we know what default is provided for various schema locations.
When creating configuration stubs for the user, we don't correctly plumb these through,
requiring that the user explicitly fill in the default.

**Desired**: we should auto-fill the connector's default values when generating a stub configuration.

PR #338, opened 2022-01-15
flowctl: improve discovery workflow and add sub-api's

jgraettinger:
**Description:**

* Improve static inference to surface default values of document locations, as well as a `secret` annotation.
* Add `flowctl api spec` and `flowctl api discover` for future usage within the control plane.
* Refactor `flowctl discover` to use new api sub-commands. Update it to generate more ergonomic file scaffolding as per #335 

**Workflow steps:**

Run `flowctl discover ...` to generate a stub configuration _within a subdirectory_,
then run it again to generate the catalog source and schema files.

Site documentation for the `flowctl discover` workflow is also updated in this PR.

Concepts > Connectors > Using Connectors > `flowctl discover` section is updated as part of this PR.

**Notes for reviewers:**

Nope.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/338)


saterus (approved):
This is looking really good. It looks like it should work well for the control plane. :+1:

PR #349, opened 2022-01-27
Add RDS section to Postgres connector docs

oliviamiannone:
**Description:**

Added instructions from [slack conversation](https://estuaryworkspace.slack.com/archives/C012S5U0WP2/p1643151353092500?thread_ts=1643151253.092400&cid=C012S5U0WP2) to connect to RDS postgres. 

Modified somewhat to instruct user to choose SSH tunneling since that's all we support right now; removed the bit about IP access because we don't have an IP range.

**Workflow steps:**

Nested under Postgres connector docs with the assumption that users that need it will start there. Links to SSH docs included to make it a complete workflow 

I _think_ this should work but might be missing some details needed to get SSH + RDS to work together

**Notes for reviewers:**

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/349)

### DIFF
--- a/site/docs/reference/Connectors/capture-connectors/PostgreSQL.md
+++ b/site/docs/reference/Connectors/capture-connectors/PostgreSQL.md
@@ -118,27 +118,29 @@ to allow Flow to connect to databases ports in secure networks.
 
 To set up and configure your SSH server, see the [guide](../../../../guides/connect-network/).
 
-## Setup: PostgreSQL on Amazon RDS
+## PostgreSQL on Amazon RDS
 
 Amazon Relational Database Service (RDS) is a managed web service providing cloud-based instances
 of popular relational databases, including PostgreSQL.
 
+### Setup
+
 You can use this connector for PostgreSQL instances on RDS, but the setup requirements are different.
-Once configured, the connection will behave in the same way as standard PostgreSQL.
 
 1. Create a new RDS PostgreSQL instance for testing. You can
-[reference the instructions from Amazon](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/CHAP_GettingStarted.CreatingConnecting.PostgreSQL.html), but use
-**Standard create mode** to ensure you can access the required settings. Configure the following:
-  * **Engine**: PostgreSQL
-  * **DB instance size**: Dev/Test
-  * Enter a unique **DB instance identifier**, **Master username**, and **Master password** of your choosing.
-  * **DB instance class**: db.t3.micro
-  * **Public access**: Yes
-  * **Enable enhanced monitoring**: False
-2. You'll need to configure secure access to the database to enable the Flow capture.
-  Currently, Estuary supports SSH tunneling to allow this.
-  Follow the guide to [configure an SSH server for tunneling](../../../../guides/connect-network/).
-3. Enable logical replication on the database.
+[reference the instructions from Amazon](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/CHAP_GettingStarted.CreatingConnecting.PostgreSQL.html), but if you're using AWS
+Management Console, be sure to select
+**Standard create mode** so you can access the required settings.
+
+  Configure the following:
+   * **Engine**: PostgreSQL
+   * **DB instance size**: Dev/Test
+   * Enter a unique **DB instance identifier**, **Master username**, and **Master password** of your choosing.
+   * **DB instance class**: db.t3.micro
+   * **Public access**: Yes
+   * **Enable enhanced monitoring**: False
+
+2. Enable logical replication on the database.
 
   a. Create a [parameter group](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_WorkingWithParamGroups.html)
   with the following properties:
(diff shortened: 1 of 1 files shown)


site/docs/reference/Connectors/capture-connectors/PostgreSQL.md
```diff
@@ -17,6 +17,14 @@ To use this connector, you'll need a PostgreSQL database setup with the followin
 
 ### Setup
 
+:::info
+These setup instructions are PostgreSQL instances you manage yourself. If you use a cloud-based managed service
```
willdonnelly, line 21: I think you missed a word, perhaps "are for" or "are intended for"?

site/docs/reference/Connectors/capture-connectors/PostgreSQL.md
```diff
@@ -17,6 +17,14 @@ To use this connector, you'll need a PostgreSQL database setup with the followin
...
+for your database, different setup steps are required.
```
willdonnelly, line 22: Perhaps replace "are required" with "may be required"? There are probably some managed PostgreSQL offerings where the normal setup steps would still work, it's just that Amazon RDS specifically has tweaked a couple of important details with their own changes, and if I read "may be" then I'm more inclined to go try it anyway.

site/docs/reference/Connectors/capture-connectors/PostgreSQL.md
```diff
@@ -109,3 +117,64 @@ The PostgreSQL source connector [supports SSH tunneling](../../../concepts/conne
...
+## PostgreSQL on Amazon RDS
+
+Amazon Relational Database Service (RDS) is a managed web service providing cloud-based instances
+of popular relational databases, including PostgreSQL.
+
+### Setup
+
+You can use this connector for PostgreSQL instances on RDS, but the setup requirements are different.
+
+1. Create a new RDS PostgreSQL instance for testing. You can
```
willdonnelly, line 130: I would remove the first step here. In the typical case the user already has a DB with a bunch of data in it that they need captured into Flow, and starting with "here's how you create a database" is just going to cause confusion.

I just included this part in the Slack thread because that's where my notes began, and because some of those settings were non-obvious when I started.

site/docs/reference/Connectors/capture-connectors/PostgreSQL.md
```diff
@@ -109,3 +117,64 @@ The PostgreSQL source connector [supports SSH tunneling](../../../concepts/conne
...
+  CREATE USER flow_capture WITH PASSWORD '<secret>';
+  GRANT rds_replication TO flow_capture;
+  GRANT SELECT ON ALL TABLES IN SCHEMA public TO flow_capture;
+  ALTER DEFAULT PRIVILEGES IN SCHEMA public GRANT SELECT ON TABLES TO flow_capture;
+  CREATE TABLE IF NOT EXISTS public.flow_watermarks (slot TEXT PRIMARY KEY, watermark TEXT);
+  GRANT ALL PRIVILEGES ON TABLE public.flow_watermarks TO flow_capture;
+  CREATE PUBLICATION flow_publication FOR ALL TABLES;
+  ```
+
+4. You'll need to configure secure access to the database to enable the Flow capture.
```
willdonnelly, line 170: For a user-facing guide I would recommend leading with this step, because a lot of the time it's going to be the biggest obstacle and the bits about fiddling with the DB parameters and creating a user are trivial after that.

site/docs/reference/Connectors/capture-connectors/PostgreSQL.md
```diff
@@ -109,3 +117,64 @@ The PostgreSQL source connector [supports SSH tunneling](../../../concepts/conne
...
+  Currently, Estuary supports SSH tunneling to allow this.
+  Follow the guide to [configure an SSH server for tunneling](../../../../guides/connect-network/).
+
+5. Configure your connector as you normally would,
```
willdonnelly, line 174: Perhaps replace "as you normally would" with something like "as described in the 'Configuration' section"? The user reading this probably hasn't done this before and has no "normally".

site/docs/reference/Connectors/capture-connectors/PostgreSQL.md
```diff
@@ -109,3 +117,64 @@ The PostgreSQL source connector [supports SSH tunneling](../../../concepts/conne
...
+with the additional of the `proxy` stanza to enable the SSH tunnel.
+See [Connecting to endpoints on secure networks](../../../concepts/connectors.md#connecting-to-endpoints-on-secure-networks)
+for additional details and a sample.
+You can find the `remoteHost` and `remotePort` in the [RDS console](https://console.aws.amazon.com/rds/) as the Endpoint and Port, respectively.
+
+Once configured, the connection will behave in the same way as standard PostgreSQL.
```
willdonnelly, line 180: I'm not sure if this line is helpful. I see what it's trying to say, but it feels like there's a lot of potential ways of misreading it and not really a lot of value from its inclusion.

willdonnelly (approved):
LGTM

site/docs/reference/Connectors/capture-connectors/PostgreSQL.md
oliviamiannone: Got it, my bad- i was reading and assuming we'd want them to set up a test first before hooking up the production DB but that wouldn't really make sense. 
So for pre-existing DBs, are there any of these settings that are important to change? Specifically thinking of Public Access?

willdonnelly: No, I don't think any of these settings need to be mentioned. The "Public Access" thing was part of making it directly reachable from an external IP. Since we're going to steer users towards an SSH proxy for access that shouldn't be necessary.

PR #353, opened 2022-01-28
Add logs documentation, placeholder for stats

oliviamiannone:
**Description:**

We're not quite ready to document stats because they're not as easy to access/work with yet, but documentation for logs  was also missing. 

Two new pages: Added a conceptual page under Advanced concepts. Broke off the section on accessing logs to be a separate doc under Reference, though I'm open to combining the two or moving it elsewhere. 

In the future, the plan is to cover parallel aspects of stats in subsections of these same two docs.

**Workflow steps:**

No shortlinks, don't think we need to add any either (?)

**Notes for reviewers:**

YAML samples need particular attention.. especially the example using the webhook connector.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/353)

### DIFF
--- a/site/docs/concepts/advanced/logs-stats.md
+++ b/site/docs/concepts/advanced/logs-stats.md
@@ -9,10 +9,10 @@ Access to statistics is still a work in progress. For now, this documentation de
 ## Logs
 
 Each organization, or tenant, that uses Flow has a `logs` collection under the global `ops` prefix.
-For the tenant Acme Co, it would have the name `ops/acmeCo/logs`.
+For the tenant acmeCo, it would have the name `ops/acmeCo/logs`.
 
 These can be thought of as standard application logs:
-they store information about events that occur at catalog build time.
+they store information about events that occur at runtime.
 They’re distinct from [recovery logs](./shards.md#recovery-logs), which track the state of various task shards.
 
 Regardless of how many Flow catalogs your organization has, all logs are stored in the same collection,
--- a/site/docs/concepts/advanced/logs-stats.md
+++ b/site/docs/concepts/advanced/logs-stats.md
@@ -1,5 +1,46 @@
 # Logs and statistics
 
+Flow collects logs and statistics of catalog tasks to aid in debugging and refinement of your workflows.
+
 :::caution
-Page in progress. Check back soon.
+Access to statistics is still a work in progress. For now, this documentation deals strictly with logs.
 :::
+
+## Logs
+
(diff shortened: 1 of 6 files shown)


site/docs/concepts/advanced/logs-stats.md
```diff
@@ -1,5 +1,46 @@
...
-Page in progress. Check back soon.
-:::
\ No newline at end of file
+Access to statistics is still a work in progress. For now, this documentation deals strictly with logs.
+:::
+
+## Logs
+
+Each organization, or tenant, that uses Flow has a `logs` collection under the global `ops` prefix.
+For the tenant Acme Co, it would have the name `ops/acmeCo/logs`.
```
psFried, line 12: nit suggestion: For the tenant ~~Acme Co~~ `acmeCo`, it would have the name `ops/acmeCo/logs`. Otherwise it could be interpreted as implying that we somehow transform the tenant names.

site/docs/concepts/advanced/logs-stats.md
```diff
@@ -1,5 +1,46 @@
...
+
+These can be thought of as standard application logs:
+they store information about events that occur at catalog build time.
```
psFried, line 15: "build time" isn't accurate here. We don't publish build logs in this way, but only runtime logs. Maybe a better way to say this would be: they store events that occur while running your shards.

site/docs/reference/working-logs-stats.md
```diff
@@ -0,0 +1,62 @@
...
+      image: ghcr.io/estuary/materialize-webhook:dev
+      config:
+        address: my.imaginarywebhook.com
+  bindings:
+    - resource:
+        relativePath: /log/wordcount
+      source: ops/acmeCo/logs
+```
+
+If necessary, you can also add [partition selectors](../../concepts/materialization/#partition-selectors) to only materialize logs of specific tasks.
```
psFried, line 62: I think it's worth adding a note to this section about excluding the logs for the materialization task itself. Otherwise, if the materialization connector logs something it could end up in an infinite loop materializing its own logs and logging about it. I think the example should show this exclusion, and just explain the reason for it in a comment or note. What this look like is:

```yaml
bindings:
  - resource:
      relativePath: /logEvents
    source: ops/acmeCo/logs
    # Exclude the logs of this materialization task in order to avoid an infinite loop.
    partitions:
      exclude:
        name: ['acmeCo/postgres/logs']
```

psFried (changes requested):
This looks good % the couple of little inline comments. I like the breakdown of `advanced` docs.

PR #356, opened 2022-02-01
Add database backed connector endpoints

saterus:
**Description:**

Adds endpoints for basic information retrieval/storage:
* `GET /connectors`
* `POST /connectors`
* `GET /connector_images`
* `POST /connector_images`
* `GET /connectors/:connector_id/connector_images`

These endpoints work strictly from the database. Information fetched from the connectors themselves is the next immediate followup to this PR.

This also adds the `id_generator` Postgres function for generating non-sequential ids and a wrapper type for base64 encoding/decoding these values on the API side.

The tests each run in their own completely isolated database, each cloned from the central `control_test` database. Every time the tests run, they are all dropped, recreated from the migrations, and each test clones its database again. These databases are named after their tests so you can connect via psql and poke around between test runs.

**Workflow steps:**

```bash
$ cd FLOW_ROOT/crates/control
$ cargo run
```

```bash
# Create a connector
$ curl -XPOST -H "Content-Type: application/json" -d @- http://localhost:3000/connectors | jq .
{
  "name": "Hello World",
  "owner": "Estuary Technologies",
  "type": "source",
  "description": "An endless stream of greetings"
}

# Fetch a connector
curl localhost:3000/connectors | jq .

# Create an image for that connector (pull the id from your response above)
$ curl -XPOST -H "Content-Type: application/json" -d @- http://localhost:3000/connector_images {
    "connector_id": "AQQALVfHXCY",
    "image": "ghcr.io/estuary/source-hello-world",
    "sha256": "15751ba960870e5ba233ebfe9663fe8a236c8ce213b43fbf4cccc4e485594600",
    "tag": "01fb856"
}

# Fetch all images
$ curl localhost:3000/connector_images | jq .

# Fetch all images belonging to a specific connector
$ curl localhost:3000/connectors/AQQALVfHXCY/connector_images | jq .
```

None yet. Yet....

**Notes for reviewers:**

* As always, reviewing commit by commit is probably going to be simpler.
* I'm not super happy with the tests at the moment. They are rather verbose, even with the snapshot testing doing a lot of work. I haven't decided how I'd like to tackle that yet, but would appreciate any thoughts.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/356)


crates/control/src/config.rs
```diff
@@ -4,6 +4,7 @@ use serde::Deserialize;
 pub mod app_env;
 
 pub use app_env::app_env;
+use tracing::info;
 
 #[derive(Debug, Deserialize)]
 pub struct Settings {
```
psFried, line 10: Looking at how this is evolving, I'm wondering if it might be worth trying to use `clap` for these. It supports using env variables, and that would mean that everything could be defined in the `.env` files, while also allowing you to override those by passing CLI arguments. It also gives a nice consistent way of getting documentation for all the options using `--help`, and would allow for seamless integration with the rest of `flowctl-rs` (we could add generic support for .env files in `flowctl-rs`). WDYT?

crates/control/migrations/20220125162950_create_connector_images.up.sql
```diff
@@ -0,0 +1,14 @@
...
+  "image" TEXT NOT NULL,
+  "tag" TEXT NOT NULL,
+  "sha256" TEXT NOT NULL,
+  "created_at" TIMESTAMPTZ NOT NULL,
+  "updated_at" TIMESTAMPTZ NOT NULL
+);
+
+CREATE INDEX idx_connector_images_image ON connector_images("image");
+CREATE UNIQUE INDEX idx_connector_images_image_tag ON connector_images("image", "tag");
+CREATE UNIQUE INDEX idx_connector_images_sha256 ON connector_images("sha256");
```
psFried, line 14: I'm not sure if this is a good idea. It's allowed and expected for multiple tags to all have the same sha256, and it's not clear that we need to prevent duplicate sha sums. Totally fine to leave this for now, since I gather that this is sort of just a placeholder, anyway, but it's making me think that perhaps it would be a good time to talk through how we want to model connectors in this system a little more holistically.

psFried (approved):
I think this is good to merge. I left some comments inline, but none of them are blocking. More like just pointing out discussions that we'll probably want to have at some point soon.

crates/control/migrations/20220121215901_create_connectors.up.sql
```diff
@@ -0,0 +1,14 @@
+CREATE TABLE connectors (
+  "id" BIGINT NOT NULL DEFAULT id_generator(),
+  PRIMARY KEY (id),
+  "type" TEXT NOT NULL,
+  "name" TEXT NOT NULL,
+  "owner" TEXT NOT NULL,
```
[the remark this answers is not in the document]
saterus: Right now `"owner"` is just a display name. That's how I was using it. I didn't want to add more than that right now, but I wanted to leave a placeholder for it.

As for docs, that's a reasonable point. I'm going to say we should lean on rustdoc, rather than sql comments, for most of this. If we've got a tricky thing in a sql file, it's probably worth leaving a comment about, but the rustdoc will be useful beyond/accessible the migration files. I'll swing back and add them in.

crates/control/src/config.rs
saterus: Yeah, it has really gotten messier than I'd hoped. The `config` crate offers a lot of the functionality I'd want, but `sqlx` doesn't make using it easy. I was thinking that `flowctl-rs` would have a flag for pointing to a config file/directory to avoid needing to make dozens of flags.

Maybe we can chat about this more on a VC?

crates/control/migrations/20220125162950_create_connector_images.up.sql
saterus: That's a really good point. This was just a mistake on my part. I'll remove that index.

I'd agree that some of these things *are* placeholders for this schema, so I'm more than happy to talk through which things I think we don't have strong opinions on yet vs which things are Good Enough for the stage we're at.

crates/control/migrations/20220121215901_create_connectors.up.sql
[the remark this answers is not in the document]
psFried: I might suggest `maintainer` or `vendor` instead of `owner`, just as a nit pick, but I agree that it makes sense to have some sort of display name to tell users where the connector has come from.

PR #358, opened 2022-02-04
Re-enable per-task prometheus combiner metrics

psFried:
**Description:**

As part of #316, the shard and collection labels were removed from the
prometheus metrics. That now seems to have been premature, since we
don't yet have an easy way to get grafana dashboards from the new stats
collections. This adds the labels back onto the prometheus metrics so
that the old grafana dashboards will continue to function, at least
until we're able to implement their replacements.

**Workflow steps:** no change

**Notes for reviewers:** n/a

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/358)


willdonnelly (approved):
LGTM

Issue #355, opened 2022-02-01
runtime: read amplification due to slow reads


jgraettinger:
In some circumstances, we're seeing quite a bit of read amplification, where fragment data is read from cloud storage significantly more often than should be theoretically required in order to serve a dataflow. The particular circumstances involve:

 * A materialization which has frequent stalls (while awaiting commits of the endpoint system)
 * A large number of sourced physical partitions
 
 There are two suspected reasons why this is happening:
 
 ### Cancelled reads due to materialization back-pressure
 
[sendReadResult](https://github.com/estuary/flow/blob/6f4b382164b221764b43dbd8debd7d2e61c964dc/go/shuffle/read.go#L299-L385) in read.go implements distributed deadlock avoidance and will eventually cancel ongoing reads where the client is not making enough progress consuming read documents. We've observed that this is happening, and each cancellation requires that the read be re-started. However, this is only about 10% of the total read amplification that we've observed.
 
??? TBD

jgraettinger:
We nailed this down. There was no actual bug, and it did turn out to be due to read cancellations and their interplay with a) lots of concurrently-read logical partitions and b) aggressive pre-fetching and buffering of read pipelines employed by Flow.

Bumping the read cancellation timeout significantly made read amplification go away entirely.
We added lots of metric and debugging (pprof) instrumentation to enable running down this or similar issues should they arise again.

PR #362, opened 2022-02-09
go/shuffle: audit and update logging, and use ops.Logger

jgraettinger:
**Description:**

Shuffled reads now log to the ops.Logger, and are controllable using a
per-shard log level.

Do a comprehensive audit and update of the debug logging
we have in place around journal reads.

Issue #355

**Workflow steps:**

The `shards: {logLevel: debug}` stanza of a catalog task (derivation or materialization) now authors debug logs into the operations log.

No documentation impacts, as this simply makes ops logs more functional / comprehensive.

**Notes for reviewers:**

nada

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/362)

### DIFF
--- a/go/shuffle/read.go
+++ b/go/shuffle/read.go
@@ -8,9 +8,10 @@ import (
 	"time"
 
 	"github.com/estuary/flow/go/flow"
+	"github.com/estuary/flow/go/flow/ops"
 	"github.com/estuary/flow/go/labels"
 	pf "github.com/estuary/flow/go/protocols/flow"
-	log "github.com/sirupsen/logrus"
+	"github.com/sirupsen/logrus"
 	"go.gazette.dev/core/allocator"
 	pb "go.gazette.dev/core/broker/protocol"
 	"go.gazette.dev/core/consumer"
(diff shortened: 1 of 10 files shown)


go/shuffle/read.go
```diff
@@ -428,15 +444,25 @@ func (r *read) dequeue() message.Envelope {
...
-func (r *read) log() *log.Entry {
-	return log.WithFields(log.Fields{
-		"journal":     r.req.Shuffle.Journal,
+func (r *read) log(lvl logrus.Level, message string, extra ...interface{}) {
+	if lvl > r.logger.Level() {
+		return
+	}
+
+	var fields = logrus.Fields{
+		"build":       r.req.Shuffle.BuildId,
```
psFried, line 453: Do you think it's worth just putting the build ID on _all_ logs in `runtime/ops.go`? I went back and forth on it, myself, but if it seems useful enough to have on all shuffle-related logs, then perhaps it's worth including on all logs? :man_shrugging:

psFried (approved):
LGTM. One question about whether we should log the build id globally, but it's not at all blocking.

go/shuffle/read.go
jgraettinger: An interesting wrinkle is that, for example, in shuffle/coordinator.go we're logging a build that _may not be_ the same build as the task term under which the coordinator was created. It's reading under the build baked into the client's request, which is allowed to differ from the shard's.

PR #363, opened 2022-02-09
Add TOAST section to Postgres connector docs

oliviamiannone:
**Description:**

Postgres connector now mostly handles toasted values for the capture task itself. However, in the event that there are failures (particularly in downstream tasks) users need to know what's going on. New section provides some brief troubleshooting options but mostly a conceptual overview.

no shortlinks

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/363)


jgraettinger (changes requested):
> Reviewed 2 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/363)* status: all files reviewed, 2 unresolved discussions (waiting on @oliviamiannone)

___
*[site/docs/reference/Connectors/capture-connectors/PostgreSQL.md, line 180 at r1](https://reviewable.io/reviews/estuary/flow/363#-MvuJNd03QLYGpKr7Dtj:-MvuJNd03QLYGpKr7Dtk:b-v1x4a1) ([raw file](https://github.com/estuary/flow/blob/b0740710147caaa23b3fc6472de4149a861bc6d9/site/docs/reference/Connectors/capture-connectors/PostgreSQL.md#L180)):*
> ```Markdown
> TOASTed values can sometimes present a challenge for systems that rely on the PostgreSQL write-ahead log (WAL), like this connector.
> If a change event occurs on a row that contains a TOASTed value, _but the TOASTed value itself is unchanged_, it is omitted from the WAL.
> As a result, the connector sees a row update with a missing value, which can cause catalog tasks to fail if not handled properly.
> ```

As a result, the connector _emits a row update with the TOAST value omitted_ ?

I would stay away from "failure" verbiage here, because nothing's actually failed. Rather, it has a (defined, intentional, correct) behavior that's different from what a user might naively expect.

Suggestions: "it might be surprising", or "it might require adjustments to downstream catalog tasks"
___
*[site/docs/reference/Connectors/capture-connectors/PostgreSQL.md, line 187 at r1](https://reviewable.io/reviews/estuary/flow/363#-MvuKjcb2zt6ZAMK74C9:-MvuKjcb2zt6ZAMK74CA:b-qmif9w) ([raw file](https://github.com/estuary/flow/blob/b0740710147caaa23b3fc6472de4149a861bc6d9/site/docs/reference/Connectors/capture-connectors/PostgreSQL.md#L187)):*
> ```Markdown
> to fill in the previous known TOASTed value in cases when that value is omitted from a row update.
> 
> However, due to the event-driven nature of certain tasks in Flow, it's still possible to see failures downstream in your data flow, specifically:
> ```

You won't see failure, but your dataflow might not do what you think it does, because you might not be aware that the value could be missing in an individual change event you're deriving from or pushing via materialization delta-updates.

jgraettinger (approved):
LGTM!

> Reviewed 1 of 1 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/363#-:-Mvz5hd2-IaZic-YjRrW:b-85i648)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @oliviamiannone)

PR #364, opened 2022-02-09
go/protocols: fix PullClient flake

jgraettinger:
Update PullClient to prefer to include ready responses in a current
transaction. It will no longer start to commit a current transaction
when that txn isn't full, and there are ready documents to read.

Tweak test case.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/364)

### DIFF
--- a/go/protocols/capture/pull_client_test.go
+++ b/go/protocols/capture/pull_client_test.go
@@ -47,12 +47,14 @@ func TestPullClientLifecycle(t *testing.T) {
 	// |captured|, and reduces the driver checkpoint into |reducedCheckpoint|.
 	// It models the caller's expected behavior of producing captured documents
 	// into a collection upon notification.
-	var drain = func() {
+	var drain = func() string {
 		var combiner = rpc.Combiners()[0].(*pf.MockCombiner)
+		var n = len(combiner.Combined)
 		captured = append(captured, combiner.Combined...)
 		combiner.Combined = nil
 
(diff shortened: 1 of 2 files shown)


jgraettinger:
Ping?

go/protocols/capture/pull_client_test.go
```diff
@@ -47,12 +47,14 @@ func TestPullClientLifecycle(t *testing.T) {
...
 	// into a collection upon notification.
-	var drain = func() {
+	var drain = func() string {
 		var combiner = rpc.Combiners()[0].(*pf.MockCombiner)
+		var n = len(combiner.Combined)
 		captured = append(captured, combiner.Combined...)
 		combiner.Combined = nil
 
 		require.NoError(t, reducedCheckpoint.Reduce(rpc.DriverCheckpoint()))
+		return fmt.Sprintf("%d => %s", n, string(reducedCheckpoint.DriverCheckpointJson))
```
psFried, line 57: Is there something that guarantees a stable order of the keys in the serialized driver checkpoint? I seem to remember the keys being in lexicographical order, but I'm not 100%.

psFried (approved):
LGTM % one "just checking" comment :+1:

go/protocols/capture/pull_client_test.go
jgraettinger: Good question. I believe it's stable because novel keys are merged into the serialization at the end, which only happens to match the lexicographic order.

PR #366, opened 2022-02-11
read & debugging telemetry

jgraettinger:
**Description:**

Substantially increase the read cancellation timeout, from ~65 seconds to ~35 minutes.

Investigation of read amplification highlighted that mixing (relatively) frequent cancellations,
aggressive buffering, and connector transactions that regularly took ~2-3 minutes results
in pretty extravagant numbers of surplus reads from Gazette.

Relaxing this cancellation appears to have wholly resolved the problem.

Otherwise, add more telemetry.
* Support `pprof.Labels` so that we can associate specific shards, builds, and journals to goroutine traces (`debug=1`) and profiles.
* Add a new metric which surfaces the polling state of each ongoing shuffled journal read undertaken by a shard. This is helpful for understanding which specific journal(s) a shard is waiting for data from, which can then in turn be used to track down what the journal's read loop is doing.

**Workflow steps:**

No new user-facing features.

Developers can take profiles `/debug/pprof` and use `/debug/pprof/goroutine?debug=1` to get stack traces which are tightly associated with originating shards & journals. This **massively** simplifies debugging a Flow reactor with lots of things going on.

None I'm aware of.

**Notes for reviewers:**

Nada

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/366)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -24,7 +24,7 @@ require (
 	github.com/stretchr/testify v1.7.0
 	go.etcd.io/etcd/api/v3 v3.5.0
 	go.etcd.io/etcd/client/v3 v3.5.0
-	go.gazette.dev/core v0.89.1-0.20220211035637-fb71eec18f42
+	go.gazette.dev/core v0.89.1-0.20220212151322-e7c18c1b78cc
 	golang.org/x/net v0.0.0-20210825183410-e898025ed96a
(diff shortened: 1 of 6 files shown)


willdonnelly (approved):
LGTM

PR #361, opened 2022-02-09
Phil/add connector bins

psFried:
**Description:**

Migrates the `parser` and `network-proxy-service` binaries from connectors into this repo as `flow-parser` and `flow-network-proxy`. These binaries are both statically linked using musl, and included in the `.build/package/bin` directory and in the flow docker image. The process for merging these was the one from [this post](https://jrsmith3.github.io/merging-a-subdirectory-from-another-repo-via-git-subtree.html). This preserves the git history for the files being brought in from connectors.

This also migrates the elasticsearch schema builder from the connectors repo into the Flow repo under a new `flow-schemalate` binary, which consolidates the generation of elasticsearch schemas, typescript types, and markdown docs, from JSON schemas. This is in line with #332.

There will be a corresponding PR in the connectors repo, which will remove those projects and update dependents to pull them from this repo instead.

**Workflow steps:** No impact to end-users

**Documentation links affected:** n/a

**Notes for reviewers:**

GH doesn't show a super useful commit graph, so here's what we're looking at:

```
* 74c3f427 Replace rust-build make target with cargo check
* 75583562 Add rust-toolchain file and use it in CI
* bdf65840 Fix a compilation error in parser error_buffer
* 83c3fcc4 Finish migration of network-proxy from connectors
*   05e18054 Add 'crates/network-proxy/' from commit 'e940a0391af6b1ae727b9c6fe5f3262c5a3970e5'
|\
| * e940a039 postgres with ssh tunnel  (#99)
* a67d62f8 Updates to incorporate parser into the flow repo
*   c97ae055 Add 'crates/parser/' from commit '5d75e3ce07ca5456895f364a03a1b451ad714a83'
|\
| * 5d75e3ce CSV missing columns with configured headers
| * 47e74225 Support Avro object container files in parser
| * d7a7c867 Parser does type inference after resolving pointer
| * c569ddd4 Add support for W3C WG extended log file format
| * 5d923ab1 Log all parser output to connector stderr
| * 3e73fda4 Move parser errorThreshold under csv config
| * 809c569a Skip Malformed CSV Rows (#41)
| * 96534104 parser: refine projection field collation and variants
| * e558a7c9 Add a test case for a csv requiring quote config
| * 5665d327 Fix missing fields when merging parser config
| * b4ce358f Update go parser config to match rust declaration
| * 318c1bf5 protocols: extract and refactor for new repository
| * 935776e9 Parser represents missing type information as None
| * 0959c3bd Fix parser handling of overflowing numberic values
[...]

Issue #334, opened 2022-01-13
docs: document task log and metrics collections


jgraettinger:
We make no mention currently that we're automatically building these collections on the user's behalf.
We should cover:
1. That they exist, and why
2. What they look like
3. A quick example of materializing them to a database

oliviamiannone:
Just noticed this issue. I believe this documentation I added recently should cover task logs 
https://docs.estuary.dev/concepts/advanced/logs-stats/
https://docs.estuary.dev/reference/working-logs-stats/
And reviewed by Phil - https://github.com/estuary/flow/pull/353

PR #368, opened 2022-02-14
Add logs subcommmand to flowctl-rs

### LINKED ISSUE #298: Proposal: Introduce `flowctl logs` Command
It's much easier to debug problems now that task logs are being written to journals. However, accessing those logs requires reasonable familiarity with the underlying Gazette concepts and CLI commands.

To view a task's logs right now:

1. `flowctl journals list`
2. `flowctl journals read -l name=ops/acmeCo/logs/kind=capture/name=acmeCo%2Ffirehose/pivot=00`

This is cumbersome and not very obvious or easily discovered by a new user.

I propose that we add a new `flowctl logs` command. This command could read the logs journal for a task via the existing mechanisms for reading journals, but expose a more intuitive interface for doing so.

1. `flowctl logs acmeCo/firehose`

I think we should allow a user to specify the name of the task as it would appear in their catalog yaml. Eg. `/acmeCo/firehose`, rather than `ops/acmeCo/logs/kind=capture/name=acmeCo%2Ffirehose/pivot=00`. Given that all our logs are structured as `ops/:tenant:/logs/kind=:tasktype:/name=:taskname:` and task names are unique across task type, I believe we should be able to figure out which log to read from without needing to have the user specify everything explicitly.

We could expose options for specifying `--tail`, for `--offsets`, and for `--partition` ranges, each with sensible defaults.

## Further Enhancements

It would also be helpful to easily set the log level on a task/shard. While the shard spec holds the log level, it suffers from the same workflow issues as reading the logs. New users certainly won't think to edit the shard (if they know what a shard is) in order to change the logging level of the task.

## Alternatives

I've been using a set of shell functions for making reading from a journal a bit easier. These aren't very distributable to a new user, but are a useful stopgap in the meantime. These require https://github.com/lotabout/skim or some equivalent fuzzy finder to perform the interactive selection.

Fish:
```fish
function read_journal -a extra_args
    set -l journal (flowctl journals list -o json | jq ".journals[].spec.name" -cr | sk)
    if test -z $journal
        echo "No journal given"
        return
    end
    flowctl journals read -l name=$journal $extra_args
end
```

Bash (translated untested from Fish):
```bash
function read_journal() {
    journal=$(flowctl journals list -o json | jq ".journals[].spec.name" -cr | sk)
    if test -z $journal; then
        echo "No journal given"
        return
    fi
    flowctl journals read -l name=$journal $@
}
```

psFried:
**Description:**

Resolves #298 

Adds the `logs` subcommand for reading task logs from flowctl-rs.
This is implemented as an alias for `flowctl journals read` with a label
selector that's generated for the task(s) identified in the arguments.
This also involved a significant refactoring of the flowctl and
flow_cli_common crates around the execution of external binaries.

**Workflow steps:**

Note that this only works with `flowctl-rs`, which is intended to replace `flowctl` soon.

- `flowctl-rs` logs --help` to print usage info
- `flowctl-rs logs --capture acmeCo/my/capture` to read the logs of a single capture. Also `--derivation` and `--materialization`
- `flowctl-rs logs --all-derivations acmeCo` to read the logs of all derivations in the `acmeCo` tenant. Also `--all-captures` and `--all-materializations`
- `flowctl-rs logs --all acmeCo` to read the logs of all tasks in the `acmeCo` tenant

This is as yet undocumented, but should be included when we do #334 

**Notes for reviewers:**

This was implemented on top of #361 , but I made it a separate PR to keep the diff a little more manageable given the refactoring that went into the CLI.


jgraettinger (approved):
LGTM % suggested improvement in how the label selector is built

> Reviewed 9 of 9 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/368#-:-Mvu8UxH4k2T9XKk5YFk:blg8xzg)* status: all files reviewed, 2 unresolved discussions (waiting on @psFried and @saterus)

___
*[crates/flow_cli_common/src/lib.rs, line 10 at r1](https://reviewable.io/reviews/estuary/flow/368#-Mvtw8sc0Bd7oBwL_Lw5:-Mvtw8sc0Bd7oBwL_Lw6:b-7rm3m1) ([raw file](https://github.com/estuary/flow/blob/f4556e7e687321b6d00cc25c8d1ef090bd7bbdb5/crates/flow_cli_common/src/lib.rs#L10)):*
> ```XML
> pub use logging::{init_logging, LogArgs, LogFormat};
> 
> /// A simple arguments container that simply takes everything as a plain string value.
> ```

grammar nit: a lot of "simple" in here. IMO you could remove them all without losing anything.
___
*[crates/flowctl/src/logs.rs, line 21 at r1](https://reviewable.io/reviews/estuary/flow/368#-Mvu6_c6B2jiocjcwjUN:-Mvu6_c6B2jiocjcwjUO:b-plwkgz) ([raw file](https://github.com/estuary/flow/blob/f4556e7e687321b6d00cc25c8d1ef090bd7bbdb5/crates/flowctl/src/logs.rs#L21)):*
> ```XML
>         let LogsArgs { task, other } = self;
>         let tenant = task.tenant_name()?;
>         let mut prefix = format!("prefix=ops/{}/logs/", tenant);
> ```

Rather than using `prefix`, it would be more idiomatic to compose a selector by using some or all of the labels:

* `estuary.dev/field/kind` - one of `capture`, `materialization`, or `derivation`
* `estuary.dev/field/name` - the escaped name of the specific catalog task.
* `estuary.dev/collection` - `ops/the-tenant/logs`

If you want to punt on switching over, that's fine, but I'll request a comment that this factoring is eventually desired.

One immediate thing this would buy is not having to specify ahead-of-time what the type of a task is, which is a little smelly.

psFried:
> One immediate thing this would buy is not having to specify ahead-of-time what the type of a task is, which is a little smelly.

That is a _much_ better way of doing it! And I like that we could then just use `--task` instead of different flags for specific types. IMO it's worth getting that done now, since I think it'll be really quick to fix up. Thanks for the suggestion.

oliviamiannone:
Doc'ed - see https://github.com/estuary/flow/pull/392

PR #386, opened 2022-03-01
source-s3 and source-gcs docs

oliviamiannone:
**Description:**
Adds docs for source-s3 and source-gcs. 

**Notes for reviewers:**

Things that are assumptions-based and could particularly use a fact-check:
- Prerequisites, in general. Anything missing?
- Both `prefix` in  the endpoint config and `stream` in bindings are a place you can narrow down to a specific prefix (?) I wrote out a Tip to cover this and how a user might organize the capture, but honestly I’m still not 100% sure why both are needed. Applies to both s3 and gcs.
- Assuming that GCS credentials are similar to S3 access keys in that they are only required if the bucket isn’t public, and connector will work just as well with a public bucket.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/386)


site/docs/reference/Connectors/capture-connectors/gcs.md
```diff
@@ -0,0 +1,78 @@
...
+          bucket: "my-bucket"
+    bindings:
+      - resource:
+          stream: my-bucket/${PREFIX}
+          syncMode: incremental
+        target: ${TENANT}/${COLLECTION_NAME}
+
+```
+
+Your capture definition may be more complex, with additional bindings for different S3 prefixes within the same bucket.
```
mdibaiee, line 76:
```suggestion
Your capture definition may be more complex, with additional bindings for different GCS prefixes within the same bucket.
```

site/docs/reference/Connectors/capture-connectors/gcs.md
```diff
@@ -0,0 +1,78 @@
+---
+sidebar_position: 4
+---
+# Google Cloud Storage
+
+This connector captures data from an Google Cloud Storage (GCS) bucket.
+
+`ghcr.io/estuary/source-gcs:dev` provides the latest connector image when using the Flow GitOps environment. You can also follow the link in your browser to see past image versions.
```
mdibaiee, line 8: should we actually link the image, since we mention they can follow the link?
```suggestion
[`ghcr.io/estuary/source-gcs:dev`](https://ghcr.io/estuary/source-gcs:dev) provides the latest connector image when using the Flow GitOps environment. You can also follow the link in your browser to see past image versions.
```

oliviamiannone: Probably. The idea was that people are much more likely to be copying it into their console than following it, but I suppose it definitely doesn't hurt, especially since the displayed text doesn't change!

site/docs/reference/Connectors/capture-connectors/amazon-s3.md
```diff
@@ -0,0 +1,81 @@
+---
+sidebar_position: 2
+---
+# Amazon S3
+
+This connector captures data from an Amazon S3 bucket.
+
+[`ghcr.io/estuary/source-s3:dev`](https://ghcr.io/estuary/source-s3:dev) provides the latest connector image when using the Flow GitOps environment. You can also follow the link in your browser to see past image versions.
```
jgraettinger, line 8: nit: remove "when using the Flow GitOps environment"

site/docs/reference/Connectors/capture-connectors/amazon-s3.md
```diff
@@ -0,0 +1,81 @@
...
+To use this connector, either your S3 bucket must be public,
+or you must have access via a root or [IAM user](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_users.html).
+
+* For public buckets, verify that the [access policy](https://docs.aws.amazon.com/AmazonS3/latest/userguide/access-control-overview.html#access-control-resources-manage-permissions-basics) allows anonymous reads.
+* For buckets accessed by a user account, you'll need the AWS **access key** and **secret access key** for the user.
+See the [AWS blog](https://aws.amazon.com/blogs/security/wheres-my-secret-access-key/) for help finding these credentials.
+
+## Configuration
+
+There are various ways to configure and implement connectors. See [connectors](../../../concepts/connectors.md#using-connectors) to learn more about these methods. The values and code sample in this section provide configuration details specific to the Amazon S3 source connector.
```
psFried, line 21: nits: 
- "and implement" seems unnecessary and confusing
- I think plain "example" is preferred over "code sample" since it feels a little weird to call yaml _code_

site/docs/reference/Connectors/capture-connectors/amazon-s3.md
```diff
@@ -0,0 +1,81 @@
...
+
+Alternatively, you can capture the entire bucket by omitting `prefix` in the endpoint configuration and
+setting `stream` to the name of the bucket.
+:::
+
+### Values
+
+#### Endpoint
+
+| Value | Name| Description | Type | Required/Default |
```
jgraettinger, line 37: FYI, and I'm not sure how to best document this, but: `bucket` and `prefix` are used / required for discovery but aren't actually used when the connector is running. Instead, the connector takes each binding `stream` and destructures the bucket and prefix from the `stream` name.

It's even possible to have a single capture which binds datasets living across _multiple_ buckets,
but it's on the user to create correct `stream` names for each binding.

This is true of S3 & GCS, and will eventually be true of Azure Blob Store when we inevitably build that connector.

site/docs/reference/Connectors/capture-connectors/amazon-s3.md
```diff
@@ -0,0 +1,81 @@
...
+
+*To use ascending keys, you must write objects in ascending lexicographic order, such as using RFC-3339 timestamps to record modification times.
+This ensures that key ordering matches the order of changes.
+
+#### Bindings
+
+| Value | Name| Description | Type | Required/Default |
+|---|---|---|---|---|
+| `stream` | Prefix | Path to dataset in the bucket, formatted as `bucket-name/prefix-name` | string | Required |
+| `syncMode` | Sync mode | Connection method. Always set to `incremental`. | string | Required |
```
psFried, line 56: Doesn't have to be a part of this PR, but I think we should talk about `syncMode`. It seems like we should maybe just omit this from all of our docs, and we should maybe also consider whether we can remove it from the catalog models, too.

jgraettinger (approved):
LGTM, but left a comment with detail on `bucket` and `prefix`

site/docs/reference/Connectors/capture-connectors/gcs.md
```diff
@@ -0,0 +1,78 @@
...
+To use this connector, either your GCS bucket must be public, or you must have access via a Google service account.
+
+* For public buckets, verify that objects in the bucket are [publicly readable](https://cloud.google.com/storage/docs/access-control/making-data-public).
+* For buckets accessed by a Google Service Account:
+    * Ensure that the user has been assigned a [role](https://cloud.google.com/iam/docs/understanding-roles) with read access.
+    * Create a [JSON service account key](https://cloud.google.com/iam/docs/creating-managing-service-account-keys#creating). Google's [Application Default Credentials](https://cloud.google.com/docs/authentication/production) will use this file for authentication.
+
+## Configuration
+
+There are various ways to configure and implement connectors. See [connectors](../../../concepts/connectors.md#using-connectors) to learn more about these methods. The values and code sample in this section provide configuration details specific to the GCS source connector.
```
psFried, line 21: same nits I had for S3 apply here

psFried (changes requested):
These look good. I managed to find a few nits to pick, but should be fine to merge after.

site/docs/reference/Connectors/capture-connectors/amazon-s3.md
oliviamiannone: Definitely worth talking about. I didn't have tables for bindings initially, but then I added them because it was raising questions for me. (what's required of a binding? Are they standard or are they defined by connectors? What do these properties _mean_? Etc). Whenever something like that happens, I always assume users will eventually arrive at the same questions and so I add more detail. 
If it's something that a user will see - especially something that's required - it could potentially cause trouble if we never talk about it. Even if it's a vestigial property from Airbyte. 
Removing it from the catalog models, on the other hand, sounds rather appealing. :) Although I'm sure it has plenty of implications.

Issue #344, opened 2022-01-19
control: Design Catchall


saterus:
This is a big system we're working on and there are a lot specific bits we need to work out yet. We've been having regular design discussions to hash out the details, but there's quite a breadth of topics. Some of them will get their own dedicated Issue (like #341) if they warrant a lot of extended discussion.

I'm going to post the notes from our discussions here as a way to keep everyone else in the loop.

saterus:
## 2022-01-13:

**Indexing Entities in Postgres:**
As a general rule, we still want to rely on the Build SQLite databases for detail Entity Specs.
In order to supply the UI with cross-cutting info, we'll need to maintain a few secondary indexes in Postgres.

Specifically:
- Index of (build, entity). This allows us to track all the Builds a particular Entity was included in and vice versa.
- Index of (build, task, collection). This allows us to track which Collections were referenced by a Task at the time of a given Build.
- Log of (build, entity, activated_at). This history of activations allows us to know which Build contains the currently active definition of an Entity.

**Resource Resolution:**
Discussed a desire for UI & CLI file structure compatibility. This allows Catalogs created in the UI to be usefully represented in a CLI workflow. We'd like to avoid a user making a transition to a GitOps workflow sift through a 2000 line yaml file.

We need to resolve references on the client side (UI or CLI). The Catalog payloads will include related Resources as bundled ResourceDefs. This allows the UI to handle them separately and the CLI to manage them as individual files. This should avoid the user needing to sift through unnecessary duplication.

saterus:
## 2022-01-14:

**Brainstormed Questions that the Control Plane will need to be able to answer:**
- Which Build defines Collection XYZ?
- Which version of Capture ABC is active?
- Which Materializations read from Collection JKL?
- What is the history of Builds that involved Entity X?
  - Where it would be activated?
- Which Tasks have used Connector:0xabcd?
  - Security updates
  - User gets the latest version whenever the shard launches
  - Images *can* be pinned by sha explicitly
- Which Builds were built by Flow version=x?
- What are the definitions for all these Entities as a single Catalog?


**Build/Test/Activate:**
- Activations are first class concepts, not part of a Build.
- Tests *are* part of a Build. To re-run tests, resubmit the same Build contents.


**Users/Permissions:**
- Users gain access to an Organization/Tenant through Roles, not inherently.
- Language bikeshed: Organization vs Tenant vs Account vs <other>?


**Events:**
[...]

PR #387, opened 2022-03-02
Add Control Plane Authentication

saterus:
**Description:**

We need to be able to authenticate _who_ is making requests. We'll be using OpenID Connect for this in the future, but we need a simple way to do this for local environments too. A lot of the plumbing is the same, so we'll start with the simple part.

Adds the notion of an `Account` and a `Credential`. Users can use a `Credential` to authenticate themselves to act on behalf of an `Account`. Actions performed by handling a request will be attributed to this `Account` for purposes of authorization.

The API now includes a few new endpoints:
* `POST /sessions/:issuer`: accepts an `auth_token` parameter which can be used to prove the identity of the bearer. There is currently one supported `issuer`: `"local"`. 
    * For the special `"local"` issuer, the `auth_token` is expected to be the `Account`'s `name`. There is no further proof required. This is deliberately not a secure mechanism. For now, if there is no `Account` with the provided name, a new one is created. As the UI progresses, we will prompt the User for additional information during registration.
* `GET /accounts`: lists all `Account`s. This will be removed/revised when tackling authorization, but currently this is useful for inspecting the system.
* `GET /accounts/:id`: shows the details for a single `Account`. This will likely be revised when tackling authorization, but currently showcases basic usage of `CurrentAccount`. It requires the authenticated caller is the same as the specified `Account`.

Almost all requests now require a Basic Auth header which contains the account id and the session token.
Authenticated headers are omitted for login and health check endpoints.

**Workflow steps:**

The high level user workflow is largely documented in [this Design Issue comment](https://github.com/estuary/flow/issues/344#issuecomment-1058344465).

All the curl commands I've listed in these PRs are now broken. :scream: These aren't necessarily a drop-in replacement, but should ease the transition.

###### Fetch a Session Token:

```bash
$ curl -s -H "Content-Type: application/json" localhost:3000/sessions/local -d '{"auth_token": "batman"}' | jq '.'
```

###### One-stop shop for setting an AUTH variable after login:

```bash
$ export AUTH=$(curl -s -H "Content-Type: application/json" localhost:3000/sessions/local -d '{"auth_token": "batman"}' | jq '.data.attributes | "\(.account_id):\(.token)" | @base64' -r)
```

###### Send an authenticated request:

```bash
$ curl -v -H "Authorization: Basic $AUTH" http://127.0.0.1:3000/accounts | jq .
```

None yet

**Notes for reviewers:**

Reviewing commit by commit will probably be best. There's quite a bit, but I tried to keep it reasonably organized. The test changes get a little hairy at the end, but hopefully they make sense. I firmly believe that the easier it is to write tests, the more tests we'll write.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/387)


psFried (approved):
This looks fantastic! It feels like a really solid foundation that we'll be able to build on when it comes time to integrate 3rd party logins. The only nit I think needs resolved is the naming in the accounts table, and I think it'd also be fine to handle that in a separate PR if that's easier

PR #388, opened 2022-03-02
Preserve order of JSON fields in sources crate to avoid sops MAC mismatch

### LINKED ISSUE #303: `sops` document re-ordering can violate MACs
`sops` creates a MAC over the contents and structure of documents it protects.

Meanwhile we parse `sops` YAML or JSON documents using a `serde_json::Value` during catalog build, and re-encode them (always) as JSON. The trouble is that `serde_json::Value` sorts on ascending property name, and it turns out that this changes (violates) the `sops` MAC represented in the document.

My current best idea is to require that config documents having a `sops` stanza are given in already-sorted order, and give a meaningful and actionable error to the user if they're not.

mdibaiee:
**Description:**

Enable `preserve_order` feature of `serde_json` on sources crate to conserve ordering of fields when reading catalog files. This avoids an issue where the MAC of sops encrypted files mismatch after reading and writing of the file during the catalog build. Resolves #303.

**Workflow steps:**

- Clone https://github.com/mdibaiee/hello-flow and checkout to [cb4dea5](https://github.com/mdibaiee/hello-flow/commit/cb4dea545604dd6aa4f41ab7a477ae4a4db18d4b)
- Note that the sops file for this project [is not ordered](https://github.com/mdibaiee/hello-flow/blob/main/acmeCo/source-gcs.config.yaml#L3-L6)
- You will actually need to create your own encrypted config, similar to the original file, since this one is encrypted using a personal key I have.
- Run the flow using this branch, and using master
```
flowctl deploy --source flow.yaml --wait-and-cleanup --log.level=debug
```

- https://docs.estuary.dev/concepts/connectors/ : removed warning that mentioned #303 as an issue and the requirement to order keys manually

**Notes for reviewers:**

I'm not completely aware of any potential issues this might cause elsewhere. Let me know if you think this might have some blast radius outside of this crate. My simple test on a simple flow worked fine.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/388)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3406,6 +3406,7 @@ dependencies = [
  "doc 0.0.0",
  "futures",
  "humantime-serde",
+ "indexmap",
  "insta",
  "json 0.0.0",
  "lazy_static",
--- a/crates/sources/Cargo.toml
+++ b/crates/sources/Cargo.toml
@@ -18,7 +18,8 @@ lazy_static = "*"
 regex = "*"
 schemars = "*"
 serde = { version = "*", features = ["derive"] }
-serde_json = { version =  "*", features = ["raw_value"] }
+indexmap = "*"
+serde_json = { version =  "*", features = ["raw_value", "preserve_order"] }
 serde_yaml = "*"
 thiserror = "*"
 tracing = "*"
--- a/crates/sources/src/loader.rs
+++ b/crates/sources/src/loader.rs
@@ -1,12 +1,12 @@
 use super::Scope;
 use doc::Schema as CompiledSchema;
 use futures::future::{FutureExt, LocalBoxFuture};
+use indexmap::map::IndexMap;
 use json::schema::{build::build_schema, Application, Keyword};
 use models::{self, tables};
 use protocol::flow::test_spec::step::Type as TestStepType;
 use regex::Regex;
 use std::cell::RefCell;
-use std::collections::BTreeMap;
 use url::Url;
 
 #[derive(thiserror::Error, Debug)]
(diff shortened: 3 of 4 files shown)


jgraettinger:
Is it possible to more-tightly bound the order-preserving behavior to _just_ the loading of configuration resources?
I'm concerned about the blast radius of enabling it project-wide. For example, we [assume lexigraphic ordering](https://github.com/estuary/flow/blob/master/crates/json/src/compare.rs#L6-L11) in order to define ordering operators over generic JSON documents.

With the specific exception of `sops` MAC's, IMO it's generally preferrable to keep JSON documents in sorted order as a standard practice. It also holds the door open to a variety of processing techniques that aren't otherwise available (for example, using binary search over a minimally parsed, token-tape representation of a document).

jgraettinger:
[This is where parse the YAML and transliterate it into a persisted JSON representation](https://github.com/estuary/flow/blob/master/crates/sources/src/loader.rs#L127-L135).
It's parsed into a models::Object, which is an alias for a `serde_json::Map`.

That's then stored within the `models::Config::Inline` variant.

Perhaps a possibility would be to have that Inline variant instead use a boxed serde_json::value::RawValue ?
`serde_transcode` could be used to directly transcode the yaml into json (which will preserve YAML/JSON input order AFIAK) into a serialized JSON representation.

mdibaiee:
@jgraettinger if the read JSON is converted into serde_map of the `models` crate, that crate does not have `preserve_order` enabled, so I think it might still be re-ordered once it's converted, and it only preserves its ordering in the `sources` crate while reading the JSON. I will see if I can write tests / or somehow make sure of this

jgraettinger:
Yep for this to work it's necessary to altogether bypass ever representing the configuration document it as a `serde_json::Value`. My thought is we can use `serde_transcode` to deserialize the YAML input directly into a serialized JSON output [approximately here](https://github.com/estuary/flow/blob/master/crates/sources/src/loader.rs#L128-L133), and then either switch to RawValue within the models::Config::Inline variant, or add a new variant using a RawValue.

Issue #296, opened 2021-11-21
proposal: tight integration with `sops` for credential management


jgraettinger:
# Overview

Mozilla [sops](https://github.com/mozilla/sops) is a tool for securing credentials that are distributed through a version control system such as Git. It's highly complementary to Flow's evolving control plane architecture, and I propose that we standardize on `sops` for all of:

1) The specific tool used for automated unwrapping of credentials in started connector invocations, within the managed Flow runtime.
2) A blessed, opinionated strategy for management of credentials in user GitOps Flow catalogs
3) The tool used within the control plane to separate and protect user credentials provided through UI-based workflows.

This is an incremental step towards the full workflow discussed in #278, that appears cheap in implementation effort but high leverage, as we're internally feeling the pain right now re: 2) above. Looking forward sops's design introduces some neat new possibilities not contemplated as part of #278, such as a bring-your-own KMS approach.

# Background

In #278 we proposed a Vault-based architecture for securing credentials within Flow's managed runtime.
We proposed a secure proxy component with which the runtime interacts through our established capture and materialization protocols, which receives a capability from the flow consumer shard to unwrap specific credentials required for a connector to function. The proxy uses this capability to unwrap the connector configuration and delegate to the connector over its chosen protocol.

This "unwrap" was to-date under specified, but essentially is a merge of a more-accessible, less sensitive configuration with a more-sensitive set of credentials. As a concrete example, a Postgres connector might have the DB hostname and user in-the-clear, while protecting the password.

Also in #278, we noted the related issue of credential management in a GitOps-maintained Flow catalog. We elected not to attempt to solve it in that design, but it remains an ongoing pain point.

One other important bit of context is to note that the managed Flow runtime runs with a specific identity (a GCP service account, an AWS IAM role, etc). Today, a user arranges for Flow to write into the user's owned buckets by configuring a storage mapping and granting access to those buckets to this managed Flow identity. This is relevant because this pre-existing identity can be further leveraged to protect and authorize decryption of connector credential materials.

# Details on `sops`

sops is a CLI tool and workflow for protecting version-controlled secrets, which seeks to:

 * Maintain document structure of configuration files as JSON or YAML, by encrypting values but not keys.
 * Provide flexibility in the composition of both encrypted and in-the-clear values within a document.
 * Provide workflow strategies which avoid secrets from ever being written to disk.
 * Support many KMS systems -- GCP, AWS, Azure, and Vault -- including secret sharing across multiple KMS's, with key rotation.
[...]

PR #389, opened 2022-03-02
 Preserve order of JSON fields in sources crate to avoid sops MAC mismatch #388 

### LINKED ISSUE #303: `sops` document re-ordering can violate MACs

mdibaiee:
**Description:**

Enable `preserve_order` feature of `serde_json` on sources crate to conserve ordering of fields when reading catalog files. This avoids an issue where the MAC of sops encrypted files mismatch after reading and writing of the file during the catalog build. Resolves #303.

**Workflow steps:**

- Clone https://github.com/mdibaiee/hello-flow and checkout to [cb4dea5](https://github.com/mdibaiee/hello-flow/commit/cb4dea545604dd6aa4f41ab7a477ae4a4db18d4b)
- Note that the sops file for this project [is not ordered](https://github.com/mdibaiee/hello-flow/blob/main/acmeCo/source-gcs.config.yaml#L3-L6)
- You will actually need to create your own encrypted config, similar to the original file, since this one is encrypted using a personal key I have.
- Run the flow using this branch, and using master
```
flowctl deploy --source flow.yaml --wait-and-cleanup --log.level=debug
```

- https://docs.estuary.dev/concepts/connectors/ : removed warning that mentioned #303 as an issue and the requirement to order keys manually

**Notes for reviewers:**

I'm not completely aware of any potential issues this might cause elsewhere. Let me know if you think this might have some blast radius outside of this crate. My simple test on a simple flow worked fine.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/389)


mdibaiee:
@jgraettinger I tried the transcode approach here https://github.com/estuary/flow/pull/394, but unfortunately I hit a bug in serde: https://github.com/serde-rs/json/issues/497

RawValue untagged enum variants always fail to deserialize 😞
```
flow.yaml error at <root> :
failed to parse YAML (location None)
Caused by:
    data did not match any variant of untagged enum Config
```

I've been thinking about this issue, and I'm wondering, can we make the sops part transparent to flow as soon as possible? I see that we pass the encrypted configuration around, until we send it to the drivers in Go, and that's where they get decrypted and some work is done on them. There might be a reason for this (something about security?), but worth asking: can we decrypt the configuration in Rust, in sources crate, and use the unencrypted configuration from there on, so no other procedure has to care about sops.

jgraettinger:
[Here is background on sops and the current architecture](https://github.com/estuary/flow/issues/296). tl;dr is that it's very intentional that we're passing around the sops document and decrypting it at the last possible mile.

Could a manual impl Serialize / Deserialize help? What about storing in a String rather than RawValue ?

mdibaiee:
@jgraettinger good to know! I'll read that proposal and look into making this work then 👍🏽

crates/sources/src/loader.rs
```diff
@@ -111,11 +118,13 @@ impl<F: Fetcher> Loader<F> {
...
         // Re-write capture and materialization configurations to their inline form.
 
-        let to_inline = |endpoint_spec: &mut serde_json::Value| {
-            let taken = serde_json::from_value(std::mem::take(endpoint_spec))
+        let to_inline = |endpoint_spec: &mut Box<RawValue>| {
+            let taken = serde_json::from_str(endpoint_spec.get())
                 .expect("endpoint spec must be a ConnectorConfig");
 
-            *endpoint_spec = match taken {
+            *endpoint_spec = RawValue::from_string("null".to_owned()).unwrap();
```
mdibaiee, line 125: This replaces the previous `std::mem::take` which would set a default value. There is no default value for `RawValue` AFAIK.

crates/sources/src/loader.rs
```diff
@@ -125,18 +134,45 @@ impl<F: Fetcher> Loader<F> {
...
+                        let transcode_result =
+                            serde_transcode::transcode(deserializer, &mut serializer)
+                                .map_err(|e| LoadError::YAMLTranscodeErr(e));
+
+                        serializer.into_inner().flush().unwrap();
+                        let raw_value_result =
+                            RawValue::from_string(String::from_utf8(buf).unwrap())
+                                .map_err(|e| LoadError::JSONErr(e));
+
+                        // TODO: handle these errors properly (similar to self.fallible)
```
mdibaiee, line 156: I couldn't call `self.fallible` here because we expand it at the beginning of the function, and so we have partially moved it, so it can no longer be used as a whole.

One thing we could do is create a function similar to `self.fallible`, that doesn't rely on `self` as a whole, but accepts the `errors` table as input and operates on that, then we can have `errors` as part of our expansion and use that function I think. WDYT?

jgraettinger (changes requested):
The approach looks great! Some comments below related to code structure. Thanks

> Reviewed 1 of 1 files at r2, 1 of 4 files at r3, 2 of 6 files at r4, 2 of 5 files at r5, 4 of 4 files at r6, 3 of 12 files at r8, 1 of 1 files at r9, 11 of 11 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/389#-:-MxLp_Kk7ZemyhhwcrbC:b-5yrau8)* status: 12 of 14 files reviewed, 8 unresolved discussions (waiting on @jgraettinger and @mdibaiee)

___
*[crates/models/src/config.rs, line 45 at r6](https://reviewable.io/reviews/estuary/flow/389#-MxL82Yx7ddKI8gg-1fz:-MxL82Yx7ddKI8gg-1g-:b-v11gck) ([raw file](https://github.com/estuary/flow/blob/3be5eeb9d093dfc06df3d85f8512fec3c32dd82b/crates/models/src/config.rs#L45)):*
> ```XML
>     /// Inline configuration.
>     Inline(Object),
>     /// Once an external configuration has been inlined, we store it
> ```

Could we use `serde(skip)` here? I don't believe we ever need to serialize /deserialize this. Then `schemars` [would also skip schema generation](https://graham.cool/schemars/deriving/attributes/#skip), and the `impl JsonSchema for RawConfig` shouldn't be required.

(I believe this is moot given the RawConnectorConfig comment below).
___
*[crates/sources/src/loader.rs, line 156 at r6](https://reviewable.io/reviews/estuary/flow/389#-MxA-09RGU5aHdBTd80t-r6-156:-MxLmF2m2wXXqlDUyeTU:b5lthcf) ([raw file](https://github.com/estuary/flow/blob/3be5eeb9d093dfc06df3d85f8512fec3c32dd82b/crates/sources/src/loader.rs#L156)):*
<details><summary><i>Previously, mdibaiee (Mahdi Dibaiee) wrote…</i></summary><blockquote>

I couldn't call `self.fallible` here because we expand it at the beginning of the function ([here](https://github.com/estuary/flow/blob/mahdi/preserve-order/crates/sources/src/loader.rs#L109-L116)), and so we have [partially moved it](https://doc.rust-lang.org/rust-by-example/scope/move/partial_move.html), and it can no longer be used as a whole.

One thing we could do is create a function similar to `self.fallible`, that doesn't rely on `self` as a whole, but accepts the `errors` table as input and operates on that, then we can have `errors` as part of our expansion and use that function I think. WDYT?
</blockquote></details>

As noted in the comments above, you don't need to deal with error handling here (it's done by `load_resource_content` already).

If YAML parsing fails, it's totally acceptable to leave the taken `null` placeholder in place as the `endpoint_spec`, and an error will be reported to the user already. All other results here are infallible / can be unwrap.
___
*[crates/sources/src/loader.rs, line 145 at r7](https://reviewable.io/reviews/estuary/flow/389#-MxLhmjE0PlCAS75RdGx:-MxLhmjE0PlCAS75RdGy:b-o9zhkz) ([raw file](https://github.com/estuary/flow/blob/613389ebf8d980db032d33c3102b2d89f8af6663/crates/sources/src/loader.rs#L145)):*
> ```XML
>                         let deserializer = serde_yaml::Deserializer::from_slice(&r.content);
>                         tracing::debug!("r.content {:?}", r.content);
>                         let mut serializer = serde_json::Serializer::new(writer);
> ```

nit: I believe you can directly build a Serializer around a mut &Vec, without the BufWriter in between.
nit: development debug tracing can be removed.
___
*[crates/sources/src/loader.rs, line 148 at r7](https://reviewable.io/reviews/estuary/flow/389#-MxLlaSYEm3HwmviBStD:-MxLlaSYEm3HwmviBStE:b4y4l6w) ([raw file](https://github.com/estuary/flow/blob/613389ebf8d980db032d33c3102b2d89f8af6663/crates/sources/src/loader.rs#L148)):*
> ```XML
> 
>                         let transcode_result =
>                             serde_transcode::transcode(deserializer, &mut serializer)
> ```

At this point, `load_resource_content` has already attempted to parse `r.content` as YAML, and has reported a parse error (if any) that is surfaced to the user.

That means you don't need to also handle that error reporting here, and I'll suggest that `ok` be chained onto the `serde_transcode::transcode` result => Option, which is returned with this `and_then`.

Then, another chained `and_then` can take a successful transcoding and deal with RawValue conversion and serialization of the new `endpoint_spec`.
___
*[crates/sources/src/loader.rs, line 153 at r7](https://reviewable.io/reviews/estuary/flow/389#-MxLiX2j8TKtkEDgYxXZ:-MxLiX2j8TKtkEDgYxX_:bw6w2uj) ([raw file](https://github.com/estuary/flow/blob/613389ebf8d980db032d33c3102b2d89f8af6663/crates/sources/src/loader.rs#L153)):*
> ```XML
>                         serializer.into_inner().flush().unwrap();
>                         let raw_value_result =
>                             RawValue::from_string(String::from_utf8(buf).unwrap())
> ```

nit: you can unwrap() the RawValue::from_string result, since we just serialized it and know it can only be valid JSON.
___
*[crates/sources/src/loader.rs, line 163 at r7](https://reviewable.io/reviews/estuary/flow/389#-MxLjQcx2fjdGrzh2eKg:-MxLjQcx2fjdGrzh2eKh:bqamaj1) ([raw file](https://github.com/estuary/flow/blob/613389ebf8d980db032d33c3102b2d89f8af6663/crates/sources/src/loader.rs#L163)):*
> ```XML
>                             serde_json::to_string(&models::ConnectorConfig {
>                                 image,
>                                 config: models::Config::LoadedExternal(models::RawConfig(config)),
> ```

Seeing this, I do think that a small private struct would be better:

```rust
#[derive(Serialize)]
struct RawConnectorConfig<'a> {
  image: String,
  config: Box<serde_json::value::RawValue>
}

...

RawValue::from_string(serde_json::to_string(RawConnectorConfig{ image, config }).unwrap()).unwrap()
```

`models` are fundamental and are shared project-wide, and `LoadedExternal` is in practice a tightly-scoped implementation detail that we're instantiating only here, and only to then _immediately_ serialize it and drop it.
___
*[crates/sources/src/loader.rs, line 172 at r7](https://reviewable.io/reviews/estuary/flow/389#-MxLmUTB5MrtP1D0hEwO:-MxLmUTB5MrtP1D0hEwP:bw756bk) ([raw file](https://github.com/estuary/flow/blob/613389ebf8d980db032d33c3102b2d89f8af6663/crates/sources/src/loader.rs#L172)):*
> ```XML
>                 // Pass through the ConnectorConfig as-is.
>                 spec @ _ => RawValue::from_string(serde_json::to_string(&spec).unwrap())
>                     .map_err(|e| LoadError::JSONErr(e))
> ```

nit: this `map_err` doesn't do anything given the following `ok` call. In any case the `RawValue::from_string` cannot ever fail (should be unwrap()) given we know it's JSON.

jgraettinger (approved):
:lgtm:

> Reviewed 2 of 5 files at r5, 5 of 5 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/389#-:-MxPkadJAN5c72o99uJe:bnfp4nl)* status: all files reviewed, 8 unresolved discussions (waiting on @mdibaiee)

PR #390, opened 2022-03-02
build: tighten outputs and paths

jgraettinger:
**Description:**

`flowctl` now requires that external binaries it invokes are co-located
in its same executable directory. This was already the practice with
`flowctl-go`, so the behavior is now consistent for both.

Similarly require that `sops` be co-located in the current executable
directory.

Remove dependencies on GOPATH from Makefile.
Instead, Go binaries are directly built (using go build -o) into the
desired $PKGDIR location.

export PATH is also removed from the Makefile, to coerce binaries to
rely only on co-location within a package directory and to not shell out
to $PATH and hope for the best, as this is a frequent source of
challenges from unexpected / old / stale binaries.

**Workflow steps:**

None.

None.

**Notes for reviewers:**

None.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/390)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,8 +6,6 @@ export FLOW_VERSION = $(shell git describe --dirty --tags)
 DATE    = $(shell date +%F-%T-%Z)
 # Number of available processors for parallel builds.
 NPROC := $(if ${NPROC},${NPROC},$(shell nproc))
-# Configured Go installation path of built targets.
-GOBIN = $(shell go env GOPATH)/bin
 # Configured Rust installation path of built release targets.
 # Caller may override with a CARGO_TARGET_DIR environment variable.
 # See: https://doc.rust-lang.org/cargo/reference/environment-variables.html
@@ -25,12 +23,9 @@ RUST_MUSL_BIN = ${CARGO_TARGET_DIR}/x86_64-unknown-linux-musl/release
 # Location to place intermediate files and output artifacts
 # during the build process. Note the go tool ignores directories
 # with leading '.' or '_'.
-WORKDIR  = .build
+WORKDIR  = $(realpath .)/.build
 # Packaged build outputs.
 PKGDIR = ${WORKDIR}/package
-# All invocations can reference installed tools, Rust, and Go binaries.
-# Each takes precedence over the configured $PATH
-export PATH := ${RUSTBIN}:${RUST_MUSL_BIN}:${GOBIN}:${PATH}
 
 # Etcd release we pin within Flow distributions.
 ETCD_VERSION = v3.4.13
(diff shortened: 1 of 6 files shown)


psFried (approved):
LGTM

jgraettinger:
@psFried this has changed a bit more, and I think it's sound now. PTAL.

Makefile
```diff
@@ -25,12 +23,9 @@ RUST_MUSL_BIN = ${CARGO_TARGET_DIR}/x86_64-unknown-linux-musl/release
 # Location to place intermediate files and output artifacts
 # during the build process. Note the go tool ignores directories
 # with leading '.' or '_'.
-WORKDIR  = .build
+WORKDIR  = $(realpath .)/.build
```
psFried, line 26: Can you shed some light on why this is an absolute path? It seems like it just makes it harder to work with, since the task names become dependent on the location that the repo is checked out to. I had recently removed this and made them relative. Did that break something?

go/pkgbin/exec.go
```diff
@@ -0,0 +1,47 @@
...
+	"os/exec"
+	"path/filepath"
+
+	"github.com/sirupsen/logrus"
+)
+
+// Locate the full path of a binary by:
+// * Looking first for a binary that exist alongside the currently
+//   executing binary, in a common directory.
+// * Only if that's not found, then attempt to find a binary on the $PATH.
```
psFried, line 19: This is nice, and makes me wonder if we should do the same thing in Rust. Definitely doesn't sound _necessary_ though, so IMO we just merge this and consider adding that later if we think of a reason to. The only thing that comes to mind at the moment is that `cargo run -p flowctl` wouldn't be able to find executables, but that's not something I imagine wanting to do as part of my workflow.

psFried (approved):
LGTM % the absolute paths in Makefile

Makefile
jgraettinger: > since the task names become dependent on the location that the repo is checked out to

I'm not quite following this. What are task names (make targets), and what is the relevance of their being dependent on the repo checkout path?

The specific reason for doing this is that `go test` compiles test binaries into arbitrary temporary directories and runs them from there. These binaries won't have co-located exec's in the same directory, and require that they be available on the $PATH.

Meanwhile this PR _removes_ any dependency on `GOPATH` , which I still _think_ is desirable given the constant friction of "which binary is running? the one in ~/go/bin or the one in .build/package/bin?". But that means we can't assume that binaries will be externally configured to be available on the $PATH, and thus Makefile must arrange for $PATH must include the binaries.

But... those test binaries are running in an arbitrary temp directory as their cwd, and a relative `.build/package/bin` is meaningless for them.

psFried: Ah, gotcha. Sounds like it's mostly just about ensuring we have the PATH set to include those. IMO it would be preferable to add the absolute path prefix when setting the PATH instead of changing PKGDIR, since PKGDIR is used in target names. I sometimes use target names on the CLI, either with `-o`, or just to build a single binary, and it's more convenient to use the relative path. Should be easy to have our cake and eat it too, though, so can just update it later.

PR #393, opened 2022-03-02
first steps toward support for foreign collections & storage mappings

jgraettinger:
**Description:**

This is an internal refactor which sets the stage for including foreign entities in catalog builds. There are some improvements around inference as well as error conditions and messaging, but no notable feature or behavior changes.

**Workflow steps:**

Business as usual.

Nada.

**Notes for reviewers:**

Recommend reviewing commit-by-commit.


willdonnelly (approved):
I think I understood maybe 60% of this change, but everything LGTM as far as I can tell.

jgraettinger:
Thanks. What would help improve that %? Perhaps not the most pressing thing, but we could do a code tour?

PR #397, opened 2022-03-04
Strongly Type `Id`s

saterus:
**Description:**

One of the things I wanted to do was to parameterize our `Id` type with the type that the `Id` is for. Instead of `Id`, we now have `Id<Account>`. An `Id<Account>` is not compatible with functions that take `Id<Connector>` or `Id<Build>`. This prevents cryptic bugs where stringly-typed ids are passed around and accidentally used for the wrong purposes. :grimacing: This seemed better to do sooner rather than later, when there is more code and more people working on the project in parallel.

Implementing this was a both easier and trickier than I thought it would be. All our tools Just Worked:tm: as soon as I added the parameter, but none of the `derive` traits worked on the new `Id<T>` type. There's a bit of noise in this where we manually implement things like `Clone`, `PartialEq`, and friends, but they are at least fairly straightforward.

**Workflow steps:**

There's very little extra work on the caller side. Instead of using `Id`, use `Id<IdentifiedType>`.

N/A

**Notes for reviewers:**

Builds on #387. Once that's merged, I'll rebase this. Until then, you can start commit-by-commit reviewing at commit `"Strongly Type Ids"`. This is a bit of a scatter shot. These commits don't all necessarily compile, but they are at least organized so that they should make sense.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/397)


jgraettinger (approved):
LGTM

saterus:
Thanks for the review. :raised_hands: 

I'm writing up some more detailed README instructions based on your VC feedback that I'm going to tack onto this PR.

jgraettinger:
I'm going ahead and merging 'cause I'd like to rebase on this 😁

PR #399, opened 2022-03-04
Docs: add more delta updates info and rockset-specific caveats

oliviamiannone:
**Description:**

In light of https://github.com/estuary/connectors/issues/115, the Rockset connector can handle TOASTed values without issue, but there are important considerations regarding delta updates and reduction strategies that users should be aware of.

na

**Notes for reviewers:**

The usual - needs a fact-check as some assumptions were involved. One specific question noted inline.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/399)

### DIFF
--- a/site/docs/reference/Connectors/materialization-connectors/Rockset.md
+++ b/site/docs/reference/Connectors/materialization-connectors/Rockset.md
@@ -59,17 +59,19 @@ materializations:
 ## Delta updates and reduction strategies
 
 The Rockset connector operates only in [delta updates](../../../concepts/materialization.md#delta-updates) mode.
-This means that Rockset merges documents by key when they arrive, using semantics that are somewhat different than Flow's.
-(FOR REVIEW: include a brief description of the difference here? I don't quite understand the distinction but seems worthwhile to cover in brief)
+This means that Rockset, rather than Flow, performs the document merge.
+In some cases, this will affect how materialized views look in Rockset compared to other systems that use standard updates.
 
-In most cases, this difference won't affect your data. However, using certain
-reduction strategies in the schemas of collections that will be materialized to Rockset _can_ result in invalid data:
+Rockset merges documents by the key defined in the Flow collection schema, and always uses the semantics of [RFC 7396 - JSON merge](https://datatracker.ietf.org/doc/html/rfc7396).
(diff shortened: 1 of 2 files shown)


site/docs/reference/Connectors/materialization-connectors/Rockset.md
```diff
@@ -55,6 +55,22 @@ materializations:
       	collection: ${table_name}
     source: ${tenant}/${source_collection}
 ```
+
+## Delta updates and reduction strategies
+
+The Rockset connector operates only in [delta updates](../../../concepts/materialization.md#delta-updates) mode.
+This means that Rockset merges documents by key when they arrive, using semantics that are somewhat different than Flow's.
```
psFried, line 62: Here's my take on this:

> Rockset will merge documents having the same key, as defined by the Flow collection. The documents will be merged according to semantics of [RFC 7396 - JSON merge](https://datatracker.ietf.org/doc/html/rfc7396). This is different from how Flow reduces documents, and Rockset's merge will _not_ use any reduction annotations from your Flow schema. If your Flow JSON schema has a top-level `merge` reduction strategy and all other fields have the default reduction strategy of `lastWriteWins`, then the results will be equivalent. If your Flow collection's JSON schema uses any other reduction stategies, then the view that's materialized to Rockset will differ from the view of the same collection that's materialized to an RDBMS such as Postgres.

site/docs/reference/Connectors/materialization-connectors/Rockset.md
```diff
@@ -55,6 +55,22 @@ materializations:
...
+(FOR REVIEW: include a brief description of the difference here? I don't quite understand the distinction but seems worthwhile to cover in brief)
+
+In most cases, this difference won't affect your data. However, using certain
```
psFried, line 65: > In most cases, this difference won't affect your data

This isn't going to age well, as I expect that reduction annotations will become much more commonly used than they are now. I'm a little torn about whether to leave it for now, but just FYI

psFried (changes requested):
I think this helps a lot. I gave a suggestion for `FOR REVIEW` paragraph, but everything else looks great

site/docs/reference/Connectors/materialization-connectors/Rockset.md
willdonnelly: s/will differ/may differ/ perhaps?

PR #403, opened 2022-03-09
Enhanced Discovery Endpoint

### LINKED ISSUE #385: control: Revamp Discovery Endpoint
The discovery endpoint should return a set of entity configuration, rather than just the raw bindings from the connector.

saterus:
closes #385 

**Description:**

Previously we were just outputting the bindings directly from the connector. This isn't super helpful to the frontend which would need to piece all this together themselves.

Instead, take the discovered bindings, the config, and the connector information to generate useful catalog definitions. The contents of the payloads should be quite easy to turn into a payload to be consumed by the upcoming `/build` endpoint.

Currently unsupported: materialization discovery. I'm not entirely sure if this is going to end up with a meaningful definition, but it should at least give a useful error message when you try to discover a materialization for now.

**Workflow steps:**

Hit the discovery endpoint like normal, see more useful output.

```bash
xh -j --session=alex \
  http://127.0.0.1:3000/connector_images/JrkGaL2ABAQ/discovery \
  @path/to/postgres-config.json
```

N/A

**Notes for reviewers:**

I tried to pick a few things apart, but this one ended up evolving a bit messily and I couldn't _easily_ untangle some of the refactoring work from the feature enhancement. :grimacing:  It's probably still worth going commit-by-commit, but the last commit is too big.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/403)


crates/control/src/services/connectors.rs
```diff
@@ -1,17 +1,86 @@
...
     cmd.arg("api").arg("spec").arg("--image").arg(image);
 
     cmd
 }
 
-pub fn discovery(image: &str, config_path: &Path) -> Command {
+/// Settings for controlling names generated as part of the discovered catalog.
+pub struct DiscoveryOptions {
+    /// The name given to the generated catalog itself. eg. `postgres` in `postgres.flow.json`
+    pub catalog_name: CatalogName,
```
psFried, line 46: I'm stumbling over the naming here. Based on the comment, I think `resource_name` might be more appropriate. But I'm also not sure why it's a `CatalogName`. Is this also used for generating entity names?

crates/control/src/services/connectors.rs
```diff
@@ -1,17 +1,86 @@
...
+    ConnectorFailed(anyhow::Error),
+    #[error("failed to parse configuration")]
+    MalformedConfig(serde_json::Error),
+    #[error("could not parse the connector output")]
+    MalformedOutput(serde_json::Error),
+    #[error("connector does not support operation: `{0}`")]
+    UnsupportedOperation(ConnectorOperation),
+}
+
+pub async fn spec(image: &ConnectorImage) -> Result<RawJson, ConnectorError> {
```
psFried, line 27: What's the rationale for using `RawJson` instead of a typed struct here? Is that just temporary scaffolding, or is the idea to just try to use `RawJson` where we can and let `flowctl api` control the format?

crates/control/src/services/discovery.rs
```diff
@@ -0,0 +1,190 @@
...
+
+    pub fn name(&self) -> String {
+        format!("{}.flow.json", self.connector.codename())
+    }
+
+    pub fn config_name(&self) -> String {
+        format!("{}.config.json", self.connector.codename())
+    }
+
+    fn capture_definitions(&self) -> BTreeMap<Capture, CaptureDef> {
```
psFried, line 122: Are there cases where we could discover more than one capture? I was surprised to see this being a BTreeMap, and I'm not sure if this is just a bit of future proofing, or if I'm just missing something.

crates/control/src/services/discovery.rs
```diff
@@ -0,0 +1,190 @@
...
+        for binding in self.bindings.iter() {
+            schemas.insert(
+                binding.schema_name(),
+                Schema::Object(binding.document_schema_json.clone()),
+            );
+        }
+        schemas
+    }
+
+    pub fn name(&self) -> String {
```
psFried, line 114: nit: maybe `resource_name` here, too?

crates/control/src/services/discovery.rs
```diff
@@ -0,0 +1,190 @@
...
+
+    fn capture_name(&self) -> Capture {
+        let prefix = &self.options.catalog_prefix;
+        let name = &self.options.catalog_name;
+        Capture::new(format!("{prefix}/{name}"))
+    }
+
+    fn collection_name(&self, binding: &DiscoveredBinding) -> Collection {
+        let prefix = &self.options.catalog_prefix;
+        // TODO: Should this binding's name get wrapped in a `CatalogName::new`
```
psFried, line 185: I think the answer to this question is no. The `CatalogName` only does normalization, which is infallible, so there's no harm in deferring that until after the user decides to actually use (build) the generated collection.

crates/control/src/services/discovery.rs
```diff
@@ -0,0 +1,190 @@
...
+    pub fn schema_name(&self) -> String {
+        format!("{}.schema.json", self.recommended_name)
+    }
+}
+
+pub struct DiscoveredCatalog {
+    connector: Connector,
+    image: ConnectorImage,
+    config: Object,
+    bindings: Vec<DiscoveredBinding>,
```
psFried, line 66: nit: Do we ever ensure that the `recommended_name` of each binding is unique? If not, then we probably should, and this might be a reasonable place to do that. Either that, or have `api discover` do it. I think just adding a `_1` (`_2`, etc) to the ends of duplicate names should do it. But probably best not to assume the connectors will all ensure uniqueness.

psFried (approved):
This is looking really good. I had a few minor nits and questions, but this should be good to merge as soon as those are addressed, or even before.

crates/control/src/services/connectors.rs
saterus: Yeah, I wasn't happy with the naming here either. I'd love to find a better term.

It's used in three places: 
1. generating the file name for `{name}.flow.json`
2. generating the file name for`{name}.config.json`
3. naming the capture itself, eg. `acmeCo/{name}`, which is naming a catalog entity. This one is why it's a `CatalogName`, though to be honest there may be valid `CatalogName`s which are not valid file names... :grimacing: 

Making all of these things all match each other is following the convention of `flowctl discover`'s output, but there may be better options? 

In any case, the best I could think of was that this is the name of the discovered catalog as a whole. That said, I'm not sure `resource_name` is quite it either though. We don't use the term `resource` for this concept either and I think it may raise just as many questions. Upon further reflection, it might be more accurately the `capture_name`? Any other ideas?

---

Slight aside from @travjenkins: by the time he calls this endpoint, he's already collected the `{name}` we're referring to here from the user. We could update the expectations of this endpoint to include the user-given name and use it here. I don't think it solves the question of what we call the variable here, but we wouldn't need to default it to the `connector.codename()`, we could use the value the user has already given us.

saterus: The other thing that we discussed to today's UX/ControlPlane meeting was renaming `/discovery` to `/discovered_catalog`. We're all open to better naming on this front too, but it seemed more sensible to "create a discovered catalog from an image" than to "create a discovery from an image".

crates/control/src/controllers/connector_images/view/discovery.rs
```diff
@@ -0,0 +1,248 @@
...
+                    journals: Default::default(),
+                },
+            );
+        }
+
+        collections
+    }
+
+    fn placeholder_storage_mappings(&self) -> BTreeMap<Prefix, StorageDef> {
+        // TODO: What should we be generating for storage mappings? This is just
```
jgraettinger, line 167: This seems good to me for now. In the future we would drop it altogether.

crates/control/src/services/connectors.rs
```diff
@@ -1,17 +1,109 @@
...
+    pub fn schema_url(&self) -> Schema {
+        Schema::Url(RelativeUrl::new(self.schema_name()))
+    }
+
+    pub fn schema_name(&self) -> String {
+        format!("{}.schema.json", self.recommended_name)
+    }
+}
+
+pub async fn discover(
```
jgraettinger, line 80: For (near) future PR: we should be running actual discovery in a service loop. The controller should create the _intention_ to run a discovery rather than doing so itself.

crates/models/src/resources.rs
```diff
@@ -57,87 +68,153 @@ impl Import {
     fn example_extended() -> Self {
         Self::Extended {
             url: RelativeUrl::new("https://example/schema.json"),
-            content_type: ContentType::JsonSchema,
+            content_type: ContentType::JsonSchema(ContentFormat::Json),
+        }
+    }
+}
+
+/// ContentFormat describes the format for a resource.
```
jgraettinger, line 76: I don't believe this is needed and I'd like to see Json / Yaml distinctions of ContentFormat and json vs yaml-specific ContentType backed out of this PR.

The ContentType of a serialized JSON DOM can continue to simply be ContentType::Catalog or ContentType::JsonSchema. We'll parse the JSON as YAML, but that's totally fine and we already to that all the time.

crates/models/src/resources.rs
```diff
@@ -1,20 +1,25 @@
...
+#[derive(Debug, JsonSchema, Clone)]
 #[serde(deny_unknown_fields, rename_all = "camelCase")]
 pub struct ResourceDef {
     /// # Content type of the Resource.
     pub content_type: ContentType,
-    /// # Byte content of the Resource.
-    #[serde(serialize_with = "as_base64", deserialize_with = "from_base64")]
-    #[schemars(schema_with = "base64_schema")]
+    /// # Content of the Resource.
+    #[schemars(schema_with = "serialization::content_schema")]
```
jgraettinger, line 22: Use an untagged enum for `content` which can be either a Bytes (or String if that's hard, and defer the base64-ing), OR an Object. Don't attempt to conjoin the two during ser/deser.

Instead, match on the type and do the conversion when we're inserting into the `models::tables::Resource`, _not_ `models::ResourceDef` while loading within the `sources` crate.

Prior art for this technique: `models/src/config.rs`, `models/src/schemas.rs`, etc.

crates/models/src/resources/deserialization.rs
```diff
@@ -0,0 +1,230 @@
+use bytes::Bytes;
+use schemars::JsonSchema;
+use serde::de::{MapAccess, Visitor};
+use serde::Deserialize;
+
+use crate::resources::ResourceDef;
+use crate::{ContentFormat, ContentType, Object};
+
+#[derive(Clone, Debug, Deserialize, JsonSchema)]
```
jgraettinger, line 9: This is impressive and looks like it was challenging to write... but as mentioned above I don't think we want to cojoin the Object vs Bytes representation of a ResourceDef at ser/deser time.

Instead use serde's existing support for untagged unions, and then we don't need to maintain custom Serializer / Visitor / Deser logic.

jgraettinger (commented):
Thanks, comments below

PR #409, opened 2022-03-10
connector proxy - source connectors

jixij:
**Description:**

This is the second half of the connector proxy work, which enables it to work with capture connectors that currently speaking in airbyte source protocols.

`airbyte_source_interceptor` was implemented to translate the airbyte messages into Flow capture protocol.

**Workflow steps:**

Once this change is merged, the network proxy configuration that are available to materialize connectors will be available to all capture connectors in theory.

**Notes for reviewers:**

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/409)


crates/connector_proxy/src/apis.rs
```diff
@@ -38,30 +38,68 @@ impl FlowOperation for FlowMaterializeOperation {}
 // InterceptorStream defines the type of input and output streams handled by interceptors.
 pub type InterceptorStream = Pin<Box<dyn Stream<Item = std::io::Result<Bytes>> + Send + Sync>>;
 
-// The generic param "T" below is bounded by FlowOperation.
-// A converter is a function that contains the specific stream-handling logic of an interceptor.
-type ConverterFn<T> = Box<dyn Fn(&T, InterceptorStream) -> Result<InterceptorStream, Error>>;
-// An intercept is characterized by a pair of converters, corresponding to the handling logic of request and response streams, respectively.
-pub type RequestResponseConverterPair<T> = (ConverterFn<T>, ConverterFn<T>);
 pub trait Interceptor<T: FlowOperation> {
-    fn get_converters() -> RequestResponseConverterPair<T> {
```
jixij, line 47: Now the converters are part of the interceptor function calls.
The reason is that - 
1. request and response now have different APIs. The request converter takes in an additional connector PID. 
2. Some shared variables needs to be kept to pass from request to response(e.g. the bindings from [validate request](https://github.com/estuary/flow/blob/master/go/capture/driver/airbyte/driver.go#L297)). And these are easier to be maintained via `Arc<Mutex>` variables in the interceptor object.

crates/connector_proxy/src/apis.rs
```diff
@@ -38,30 +38,68 @@ impl FlowOperation for FlowMaterializeOperation {}
...
-            Box::new(|_op, stream| Ok(stream)),
-        )
+    fn convert_command_args(&mut self, _op: &T, args: Vec<String>) -> Result<Vec<String>, Error> {
+        Ok(args)
+    }
+
+    fn convert_request(
+        &mut self,
+        // The pid of the connector process, which is in stopped state at first, waiting for SIGCONT signals from interceptors.
+        pid: Option<u32>,
```
jixij, line 49: pass in `pid` in the `convert_request`, so that the connector process could be started after the interceptors.

The reason is that that arguments used to start the connectors process are constructed by the interceptor, so it is easier if we avoid the interceptor to depend on the connector process to be started.

crates/connector_proxy/src/main.rs
```diff
@@ -114,54 +128,99 @@ async fn signal_handler() {
...
-
     // TODO(jixiang): add a check to make sure the proxy_command passed in from commandline is consistent with the protocol inferred from image.
     match proxy_command {
-        ProxyCommand::ProxyFlowCapture(c) => proxy_flow_capture(c, image_config).await,
-        ProxyCommand::ProxyFlowMaterialize(m) => proxy_flow_materialize(m, image_config).await,
+        ProxyCommand::ProxyFlowCapture(c) => proxy_flow_capture(c, image_inspect_json_path).await,
+        ProxyCommand::ProxyFlowMaterialize(m) => {
+            proxy_flow_materialize(m, image_inspect_json_path).await
+        }
+        ProxyCommand::DelayedExecute(ba) => delayed_execute(ba.config_file_path).await,
```
jixij, line 140: Not using a bash script to start the `proxy-process`, b/c we cannot assume the shell script are working for all docker images, (e.g. `distroless`). So I structured it as this additional sub-command of the `connector-proxy`, which basically a separate process that will start the real connector after received some signals from the `connector-proxy` that starts it.

crates/connector_proxy/src/interceptors/airbyte_capture_interceptor.rs
```diff
@@ -1,32 +1,438 @@
...
 
 impl AirbyteCaptureInterceptor {
-    fn convert_request(
-        _operation: &FlowCaptureOperation,
-        _in_stream: InterceptorStream,
-    ) -> Result<InterceptorStream, Error> {
-        panic!("TBD AirbyteCaptureInterceptor")
+    pub fn new() -> Self {
+        // TODO: keep tempdir inside container only, not mounted to any local dir.
+        // -- tmpfs mount? https://docs.docker.com/storage/tmpfs/
```
jixij, line 44: just came across `tmpfs` that stores files in host memory being another way of storing secure infomation in docker. Shall we look more into that?

crates/connector_proxy/src/main.rs
```diff
@@ -114,54 +128,99 @@ async fn signal_handler() {
...
         image_config.get_entrypoint(vec![DEFAULT_CONNECTOR_ENTRYPOINT.to_string()]),
-        converter_pair,
+        interceptor,
     )
     .await
 }
+
+async fn delayed_execute(command_config_path: String) -> Result<(), Error> {
+    // Sleep for some time to allow parent process to stop the current process.
+    std::thread::sleep(std::time::Duration::from_millis(100));
```
jixij, line 194: This was a bit hacky. 
The approach I've tried is - let the `delayed_execute` process pause itself, and wait for `sigcont` signal from its parent (connector proxy) process to continue. The issue here was that the - `sigcont` from the parent might came even before the `delayed_execute` stopped and prepared to listen to the `sigcont` signal. If `delayed_execute` stopped after missing the `sigcont` signal, it will stack forever.
The main issue is that - separate the stop/resume actions in two processes is not good, b/c it is difficult to coordinate them.

So in this implementation, both the `stop` and `start` signals are coming from the parent process, so that they could be ordered. And added the sleep logic in the delayed process to given enough time for the parent process to stop it. Alternatives I was thinking -
- Using a locker file. The parent creates a locker file before starting the child process. And the child process will block and ping the locker file until it is deleted.
- Building a binary using other languages, and inject it to the docker image together with `connector-proxy`.  (Or calling to the other languages from Rust). The process API in Rust does not allow us to do much. Hopefully the other language could.

go/capture/driver/airbyte/driver.go
```diff
@@ -443,29 +295,18 @@ func (d driver) Pull(stream pc.Driver_PullServer) error {
...
-						return pc.StagePullDocuments(stream, &resp, b, rec.Record.Data)
-					}
-					return fmt.Errorf("connector record with unknown stream %q", rec.Record.Stream)
+		connector.NewProtoOutput(
+			func() proto.Message { return new(pc.PullResponse) },
+			func(m proto.Message) error {
+				var pullResp = m.(*pc.PullResponse)
+				if c := pullResp.Checkpoint; c != nil {
+					return pc.WritePullCheckpoint(stream, &resp, c)
+				} else if d := pullResp.Documents; d != nil {
```
jixij, line 304: The translation from airbyte message to proto message is done in the Rust connector. However, staging / combining the messages are done in Go side. It this OK? Or should the boundary between the Rust/Go be shifted?

jixij:
@jgraettinger @psFried 
This is a draft PR with lots of TODOs. I am not in a hurry, but if you can take a quick look and give some initial feedbacks, it will be great. Just to make sure this is in the right track. Thank you!

jgraettinger (commented):
Thanks @jixij , I'm not able to review this yet but did at least address your comments below.

> Reviewed 1 of 10 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/409#-:-MxtU4WpCj7iFIv2Zpl7:bbt9111)* status: 1 of 16 files reviewed, 5 unresolved discussions (waiting on @jixij and @psFried)

___
*[crates/connector_proxy/src/apis.rs, line 49 at r6](https://reviewable.io/reviews/estuary/flow/409#-MxozmW7vWCxg_U0eAX6-r6-49:-MxtPUx16lZRwnos9KH9:b8fm8x5) ([raw file](https://github.com/estuary/flow/blob/032edb3160fcbc341fdd07cbc7cb01a9ae5d4753/crates/connector_proxy/src/apis.rs#L49)):*
<details><summary><i>Previously, jixij (Jixiang Jiang) wrote…</i></summary><blockquote>

pass in `pid` in the `convert_request`, so that the connector process could be started after the interceptors.

The reason is that that arguments used to start the connectors process are constructed by the interceptor, so it is easier if we avoid the interceptor to depend on the connector process to be started.
</blockquote></details>

As we've discussed, I think this can all be simplified by _removing_ the Intercepter trait and ComposedInterceptor, and building the concrete stacks as needed, and as composed literals. Interceptor is a leaky abstraction (since pid is only needed for airbyte) that doesn't buy much.

Instead have a top-level switch over the protocol cases that constructs an impl InterceptorStream by composing its component pieces:

```rust
let request_stack = airbyte_request_adapter(
   pid,
   other_arg_specific_to_airbyte,
   some_wrapped_adapter(
      some,
      stuff,
      convert_my_stdin_to_stream(stdin())));

let response_stack = airbyte_response_adapter( ... )
```
___
*[crates/connector_proxy/src/main.rs, line 140 at r6](https://reviewable.io/reviews/estuary/flow/409#-MxozmW7vWCxg_U0eAXE-r6-140:-MxtSKmi2BEYms6-2sSL:bd1p8cn) ([raw file](https://github.com/estuary/flow/blob/032edb3160fcbc341fdd07cbc7cb01a9ae5d4753/crates/connector_proxy/src/main.rs#L140)):*
<details><summary><i>Previously, jixij (Jixiang Jiang) wrote…</i></summary><blockquote>

Not using a bash script to start the `proxy-process`, b/c we cannot assume the shell script are working for all docker images, (e.g. `distroless`). So I structured it as this additional sub-command of the `connector-proxy`, which basically a separate process that will start the real connector after received some signals from the `connector-proxy` that starts it.
</blockquote></details>

The specific "call ourselves" approach is reasonable, but I do want to raise a latent concern that we need to taking bigger steps back to reconsider engineering constraints we're setting for ourselves: ("it must run in distroless", "we can't assume there's a `ssh` binary we can shell to", etc).

We don't _need_ to hit these constraints to have meaningful functionality that works in _most_ contexts, and as we raise the bar for how self-sufficient we're trying to be, it adds exponentially to development time and maintenance.
___
*[crates/connector_proxy/src/main.rs, line 194 at r6](https://reviewable.io/reviews/estuary/flow/409#-MxozmW7vWCxg_U0eAXE-r6-194:-MxtMcjF8zi8xIsujYbE:bs1xkxe) ([raw file](https://github.com/estuary/flow/blob/032edb3160fcbc341fdd07cbc7cb01a9ae5d4753/crates/connector_proxy/src/main.rs#L194)):*
<details><summary><i>Previously, jixij (Jixiang Jiang) wrote…</i></summary><blockquote>

This might look a bit hacky, and here is the reason.

The approach I've tried is - let the `delayed_execute` process pause itself, and wait for `sigcont` signal from its parent (connector proxy) process to continue. The issue here was that the - `sigcont` from the parent might came even before the `delayed_execute` stopped and prepared to listen to the `sigcont` signal. If `delayed_execute` stopped after missing the `sigcont` signal, it will stack forever.
The main issue is that - separate the stop/resume actions in two processes is not good, b/c it is difficult to coordinate them.

So in this implementation, both the `stop` and `start` signals are coming from the parent process, so that they could be ordered, and added the sleep logic in the delayed process to give enough time for the parent process to stop it (it might be messy if the real connector being started before process was stopped).
Not sure if this is ok. Alternatives I was thinking -
- Using a locker file. The parent creates a locker file before starting the child process. And the child process will block and ping the locker file until it is deleted.
- Building a binary using other languages, and inject it to the docker image together with `connector-proxy`.  (Or calling to the other languages from Rust). The process API in Rust does not allow us to do much. Hopefully the other language could.



</blockquote></details>

A generalized pattern to do this is to have the child write to stderr when it's ready (has installed signal handler in this case).

See [flow/js_worker.go](https://github.com/estuary/flow/blob/master/go/flow/js_worker.go) for an example.
The child writes "READY\n" to stderr when it's started. The parent reads this, and then passes through the remainder of child stderr to its own stderr.
___
*[crates/connector_proxy/src/interceptors/airbyte_capture_interceptor.rs, line 44 at r6](https://reviewable.io/reviews/estuary/flow/409#-MxozmW7vWCxg_U0eAX9-r6-44:-MxtONOBAkH9_lv7iENg:b-w00vs0) ([raw file](https://github.com/estuary/flow/blob/032edb3160fcbc341fdd07cbc7cb01a9ae5d4753/crates/connector_proxy/src/interceptors/airbyte_capture_interceptor.rs#L44)):*
<details><summary><i>Previously, jixij (Jixiang Jiang) wrote…</i></summary><blockquote>

just came across `tmpfs` that stores files in host memory being another way of storing secure infomation in docker. Shall we look more into that?
</blockquote></details>

No. KISS. The connector-proxy also shouldn't assume anything about the filesystem it's running on (from its perspective, it's not even "aware" it's running within a container).
___
*[go/capture/driver/airbyte/driver.go, line 304 at r6](https://reviewable.io/reviews/estuary/flow/409#-MxozmW8fjPFjqoSxWRT-r6-304:-MxtOibX-ZAEGuZmpveE:b-tvvdi6) ([raw file](https://github.com/estuary/flow/blob/032edb3160fcbc341fdd07cbc7cb01a9ae5d4753/go/capture/driver/airbyte/driver.go#L304)):*
<details><summary><i>Previously, jixij (Jixiang Jiang) wrote…</i></summary><blockquote>

The translation from airbyte message to proto message is done in the Rust connector. However, staging / combining the messages are done in Go side. It this OK? Or should the boundary between the Rust/Go be shifted? 
</blockquote></details>

Let's not worry about it for now. connector-proxy should eventually do coalescing but that's an optimization we can leave for later.

This can be simplified to pass-through |pullResp| without inspecting it -- the runtime already has error cases over unexpected message types.

jixij:
> Thanks @jixij , I'm not able to review this yet but did at least address your comments below.
> 
> > Reviewed 1 of 10 files at r1, all commit messages.
> > _[Reviewable](https://reviewable.io/reviews/estuary/flow/409#-:-MxtU4WpCj7iFIv2Zpl7:bbt9111)_ status: 1 of 16 files reviewed, 5 unresolved discussions (waiting on @jixij and @psFried)
> 
> _[crates/connector_proxy/src/apis.rs, line 49 at r6](https://reviewable.io/reviews/estuary/flow/409#-MxozmW7vWCxg_U0eAX6-r6-49:-MxtPUx16lZRwnos9KH9:b8fm8x5) ([raw file](https://github.com/estuary/flow/blob/032edb3160fcbc341fdd07cbc7cb01a9ae5d4753/crates/connector_proxy/src/apis.rs#L49)):_
> 
> Previously, jixij (Jixiang Jiang) wrote…
> As we've discussed, I think this can all be simplified by _removing_ the Intercepter trait and ComposedInterceptor, and building the concrete stacks as needed, and as composed literals. Interceptor is a leaky abstraction (since pid is only needed for airbyte) that doesn't buy much.
> 
> Instead have a top-level switch over the protocol cases that constructs an impl InterceptorStream by composing its component pieces:
> 
> ```rust
> let request_stack = airbyte_request_adapter(
>    pid,
>    other_arg_specific_to_airbyte,
>    some_wrapped_adapter(
>       some,
>       stuff,
>       convert_my_stdin_to_stream(stdin())));
> 
> let response_stack = airbyte_response_adapter( ... )
> ```
> 
> _[crates/connector_proxy/src/main.rs, line 140 at r6](https://reviewable.io/reviews/estuary/flow/409#-MxozmW7vWCxg_U0eAXE-r6-140:-MxtSKmi2BEYms6-2sSL:bd1p8cn) ([raw file](https://github.com/estuary/flow/blob/032edb3160fcbc341fdd07cbc7cb01a9ae5d4753/crates/connector_proxy/src/main.rs#L140)):_
> 
> Previously, jixij (Jixiang Jiang) wrote…
> The specific "call ourselves" approach is reasonable, but I do want to raise a latent concern that we need to taking bigger steps back to reconsider engineering constraints we're setting for ourselves: ("it must run in distroless", "we can't assume there's a `ssh` binary we can shell to", etc).
> 
> We don't _need_ to hit these constraints to have meaningful functionality that works in _most_ contexts, and as we raise the bar for how self-sufficient we're trying to be, it adds exponentially to development time and maintenance.
> 
> _[crates/connector_proxy/src/main.rs, line 194 at r6](https://reviewable.io/reviews/estuary/flow/409#-MxozmW7vWCxg_U0eAXE-r6-194:-MxtMcjF8zi8xIsujYbE:bs1xkxe) ([raw file](https://github.com/estuary/flow/blob/032edb3160fcbc341fdd07cbc7cb01a9ae5d4753/crates/connector_proxy/src/main.rs#L194)):_
> 
> Previously, jixij (Jixiang Jiang) wrote…
> A generalized pattern to do this is to have the child write to stderr when it's ready (has installed signal handler in this case).
> 
> See [flow/js_worker.go](https://github.com/estuary/flow/blob/master/go/flow/js_worker.go) for an example. The child writes "READY\n" to stderr when it's started. The parent reads this, and then passes through the remainder of child stderr to its own stderr.
> 
> _[crates/connector_proxy/src/interceptors/airbyte_capture_interceptor.rs, line 44 at r6](https://reviewable.io/reviews/estuary/flow/409#-MxozmW7vWCxg_U0eAX9-r6-44:-MxtONOBAkH9_lv7iENg:b-w00vs0) ([raw file](https://github.com/estuary/flow/blob/032edb3160fcbc341fdd07cbc7cb01a9ae5d4753/crates/connector_proxy/src/interceptors/airbyte_capture_interceptor.rs#L44)):_
> 
> Previously, jixij (Jixiang Jiang) wrote…
> No. KISS. The connector-proxy also shouldn't assume anything about the filesystem it's running on (from its perspective, it's not even "aware" it's running within a container).
> 
> _[go/capture/driver/airbyte/driver.go, line 304 at r6](https://reviewable.io/reviews/estuary/flow/409#-MxozmW8fjPFjqoSxWRT-r6-304:-MxtOibX-ZAEGuZmpveE:b-tvvdi6) ([raw file](https://github.com/estuary/flow/blob/032edb3160fcbc341fdd07cbc7cb01a9ae5d4753/go/capture/driver/airbyte/driver.go#L304)):_
> 
> Previously, jixij (Jixiang Jiang) wrote…
> Let's not worry about it for now. connector-proxy should eventually do coalescing but that's an optimization we can leave for later.
> 
> This can be simplified to pass-through |pullResp| without inspecting it -- the runtime already has error cases over unexpected message types.

Sounds great.  Thanks a lot @jgraettinger  for the confirmations, it is more clear to me now.

jixij:
___
*[crates/connector_proxy/src/apis.rs, line 49 at r6](https://reviewable.io/reviews/estuary/flow/409#-MxozmW7vWCxg_U0eAX6-r6-49:-MyOQ_eFBszG_utiE8-t:b-leee93) ([raw file](https://github.com/estuary/flow/blob/032edb3160fcbc341fdd07cbc7cb01a9ae5d4753/crates/connector_proxy/src/apis.rs#L49)):*

As we've discussed, I think this can all be simplified by _removing_ the Intercepter trait and ComposedInterceptor, and building the concrete stacks as needed, and as composed literals. Interceptor is a leaky abstraction (since pid is only needed for airbyte) that doesn't buy much.

Instead have a top-level switch over the protocol cases that constructs an impl InterceptorStream by composing its component pieces:

```rust
let request_stack = airbyte_request_adapter(
   pid,
   other_arg_specific_to_airbyte,
   some_wrapped_adapter(
      some,
      stuff,
      convert_my_stdin_to_stream(stdin())));

let response_stack = airbyte_response_adapter( ... )
```
</blockquote></details>

Yes, I've removed the these traits. And implemented three runner functions for different types of connector.
`run_flow_capture_connector` and `run_flow_materialize_connector` for connectors that speaks Flow protocols.
`run_airbyte_source_connector` for connectors that speaks airbyte protocols.

They construct their own interceptors (or req/resp stream adaptors), and common logic that shared by them are extracted to separate functions.

jixij:
___
*[crates/connector_proxy/src/main.rs, line 140 at r6](https://reviewable.io/reviews/estuary/flow/409#-MxozmW7vWCxg_U0eAXE-r6-140:-MyORj5RA8vxPJWvEEVi:bt1nf6k) ([raw file](https://github.com/estuary/flow/blob/032edb3160fcbc341fdd07cbc7cb01a9ae5d4753/crates/connector_proxy/src/main.rs#L140)):*

The specific "call ourselves" approach is reasonable, but I do want to raise a latent concern that we need to taking bigger steps back to reconsider engineering constraints we're setting for ourselves: ("it must run in distroless", "we can't assume there's a `ssh` binary we can shell to", etc).

We don't _need_ to hit these constraints to have meaningful functionality that works in _most_ contexts, and as we raise the bar for how self-sufficient we're trying to be, it adds exponentially to development time and maintenance.
</blockquote></details>

Got it.. Thanks!

jixij:
___
*[crates/connector_proxy/src/main.rs, line 194 at r6](https://reviewable.io/reviews/estuary/flow/409#-MxozmW7vWCxg_U0eAXE-r6-194:-MyORuvS-51SUQbcfLzG:b4dn6na) ([raw file](https://github.com/estuary/flow/blob/032edb3160fcbc341fdd07cbc7cb01a9ae5d4753/crates/connector_proxy/src/main.rs#L194)):*

A generalized pattern to do this is to have the child write to stderr when it's ready (has installed signal handler in this case).

See [flow/js_worker.go](https://github.com/estuary/flow/blob/master/go/flow/js_worker.go) for an example.
The child writes "READY\n" to stderr when it's started. The parent reads this, and then passes through the remainder of child stderr to its own stderr.
</blockquote></details>

cool. this functionality was implemented in `invoke_delayed_connector`, which will send back a READY from STDERR and wait for sigcont signal.

go/capture/driver/airbyte/driver.go
```diff
@@ -443,49 +297,31 @@ func (d driver) Pull(stream pc.Driver_PullServer) error {
...
-		return nil // Connector flushed prior to exiting. All done.
-	}
+	// TODO: should the new logic send out a final commit if there are pending documents without a checkpoint?
+
+	//if resp == nil {
+	//	return nil // Connector flushed prior to exiting. All done.
+	//}
 
 	// Write a final commit, followed by EOF.
 	// This happens only when a connector writes output and exits _without_
```
jixij, line 316: This is something that I am not very sure about.
It looks to me the go-logic will append an `EOF` to the stream if there are pending messages that are not checked in. This logic was missing in Rust, should this be back?

crates/connector_proxy/src/libs/airbyte_catalog.rs
```diff
@@ -0,0 +1,239 @@
...
+#[validate(schema(function = "Self::validate_configured_stream"))]
+pub struct ConfiguredStream {
+    #[validate]
+    pub stream: Stream,
+    pub sync_mode: SyncMode,
+    pub destination_sync_mode: DestinationSyncMode,
+    pub cursor_field: Option<Vec<String>>,
+    pub primary_key: Option<Vec<Vec<String>>>,
+
+    // TODO: might be broken if both 'projection' and its alias is present in the JSON data.
```
jixij, line 48: I see the pattern here -
A json object could specify two fields `field` and `estuary.dev/field`, and there values are merged.

In `serde`, `alias` was used here to specify the fields. However, in this implementation, if a json object has both fields present, it will break.
I am not very sure the usecases of these fields. If there are such usecases that needs both fields co-exist in the same object, I'll switch to a customized deserializer.

crates/connector_proxy/src/interceptors/airbyte_source_interceptor.rs
```diff
@@ -0,0 +1,429 @@
...
+}
+
+impl AirbyteSourceInterceptor {
+    pub fn new() -> Self {
+        AirbyteSourceInterceptor {
+            validate_request: Arc::new(Mutex::new(None)),
+            stream_to_binding: Arc::new(Mutex::new(HashMap::new())),
+            tmp_dir: Builder::new()
+                .prefix("airbyte-source-")
+                .tempdir_in("/connector-tmpfs")
```
jixij, line 51: I spent some time on the file location. The struggle here is that it is difficult to find a path that the connector could write to.
- The `tmp` dir was mounted to local, so not safe.
- The proxy is not running as a Root user, so does not have a permission to create a dir in most places in the file system. And the available locations could be image-dependent and complex.

On the other hand, when researching, I found the `tmpfs` is pretty easy to set up, with just a few lines of code change. So I added this as a placehoder. And if storing data as files in host machine memory is not working well for other coonsiderations, especially in k8s env, I'll look into more on this..

jixij:
I've made changes based on the initial feedbacks. There were some control-plan related tests that I am still looking into, but in general I think this is good for another round of review. PTAL! Thank you @jgraettinger

jgraettinger (commented):
Thanks, some comments below, but this looks about there.

jgraettinger:
Also, the `connector_images::spec_test` timeout in CI looks like it's probably related to changes in this PR, and not a flake

.github/workflows/main.yml
```diff
@@ -33,6 +33,9 @@ jobs:
       - name: Install rust toolchain
         run: rustup show
 
+      - name: Install libgcc1
```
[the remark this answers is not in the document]
jixij: There was one stage during the development,  I was trying to build a `RUST_TARGET` instead of `MUSL_TARGET` of the connector-proxy to support unit tests triggered by `make go-test-ci`. And installed this library to support the built rust target.

However, later I decided to move `make go-test-ci` to stage 2, in which both flowctl as a RUST target and `connector-proxy` as a MUSL target are available. And this is no longer needed.

jixij:
___
*[crates/connector_proxy/Cargo.toml, line 23 at r10](https://reviewable.io/reviews/estuary/flow/409#-MxozmW7vWCxg_U0eAX5-r10-23:-Mz0SMud4IPXFJFpDT9z:b-lupeiu) ([raw file](https://github.com/estuary/flow/blob/98f499e66216765a3b3fb47cffe3a305b242f1da/crates/connector_proxy/Cargo.toml#L23)):*

use `doc::Pointer`
</blockquote></details>

Yes, I was using `doc::Pointer` for parsing the json pointers (see `tokenize_jsonpointer` in `crates/connector_proxy/src/libs/json.rs`).

The `json-pointer` is used for another purpose - to escape and  concatenate a list of strings to construct a jsoinpointer. Basically to match this Go-logic (https://github.com/estuary/flow/blob/master/go/capture/driver/airbyte/driver.go#L219-L223), which looks to be missing direct support in `doc::Pointer`. So I was relying on the `json-pointer`. But I can also extend the `doc::Pointer` or wraps around it to avoid this dependency, since this logic is very light-weighted.

crates/connector_proxy/src/interceptors/airbyte_source_interceptor.rs
jixij: Yes, `/var/tmp` works fine. Thank you! We can plan the next steps after this. Removing the `/tmp` binding would be ideal.

go/capture/driver/airbyte/driver.go
```diff
@@ -443,49 +297,31 @@ func (d driver) Pull(stream pc.Driver_PullServer) error {
...
 		logger,
-	); err != nil {
-		return err
-	}
+	)
 
-	if resp == nil {
-		return nil // Connector flushed prior to exiting. All done.
-	}
+	// TODO: should the new logic send out a final commit if there are pending documents without a checkpoint?
```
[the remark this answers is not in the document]
jixij: Yes, I've moved the logic to the connector-proxy

go/connector/run.go
```diff
@@ -99,47 +99,47 @@ func Run(
 		imageArgs = append(imageArgs, fmt.Sprintf("--network=%s", networkName))
 	}
 
+	imageArgs = append(imageArgs, "--mount", "type=tmpfs,destination=/connector-tmpfs")
```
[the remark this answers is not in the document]
jixij: done

jixij:
___
*[go/capture/driver/airbyte/driver.go, line 101 at r10](https://reviewable.io/reviews/estuary/flow/409#-MxozmW8fjPFjqoSxWRT-r10-101:-Mz0Z5DA6ZZ3jOEYGLIR:b-225mxw) ([raw file](https://github.com/estuary/flow/blob/98f499e66216765a3b3fb47cffe3a305b242f1da/go/capture/driver/airbyte/driver.go#L101)):*

don't you need to send in the SpecRequest ?
</blockquote></details>

Good point. I missed the differences between the FlowCapture and AirbyteSource.

jixij:
Thank you again Johnny for the review. I've made changes accordingly. PTAL!

FYI, Yes, the timeout issue in CI is real, and as always, it most likely caused by some race conditions that cause the `sigcont` to be missing after the bouncer process stops itself. It is now more difficult to debug b/c it only occurs on the CI.

As I read about the inter-process communication, I felt a pipe-based communication mechanism would be more robust. And I have switched to that - the delayed process will wait for a "READY" message from the parent process from stdin, before it starts the connector process.  And it seems to be working fine for the CI tests runs so far. If there is any other thoughts, please let me know.

jixij:
___
*[crates/connector_proxy/src/libs/airbyte_catalog.rs, line 48 at r9](https://reviewable.io/reviews/estuary/flow/409#-MxozmW8fjPFjqoSxWRS-r9-48:-Mz0bpND-MCdMpza_IIi:b-q74txj) ([raw file](https://github.com/estuary/flow/blob/141f8e09abb0e53a0717964aa44a9462e6addf49/crates/connector_proxy/src/libs/airbyte_catalog.rs#L48)):*

this is fine, they never co-exist. We're the only ones who use this and afaik it will always come in as `estuary.dev/projections`
</blockquote></details>

great!

mdibaiee:
will continue this work here: https://github.com/estuary/flow/pull/425

PR #411, opened 2022-03-14
Docs - managed DBs for MySql and Postgres; SSH

oliviamiannone:
**Description:**

Adds docs to...
- Mysql connector
- Postgres connector
- SSH setup guide 

...which covers implementations on...
- AWS
- GCloud 
- Azure

... per https://docs.google.com/document/d/18tLqjKmdVSyqGtTgXOdoHXrH_9OsqNumapmP5FVi3SE/edit.

**Notes for reviewers:**

Needs a second set of eyes for completeness and accuracy. It should be noted that while I did spend quite a bit of time reading and referencing the UI for the VMs and managed DBs on the three platforms, I didn't set up complete end-to-end dataflows for each workflow (for the sake of time, since complete testing was done quite recently).

Also looks like some of the changes to config tables I did recently got rolled into this PR (?). No need to review all that.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/411)

### DIFF
--- a/site/docs/guides/connect-network.md
+++ b/site/docs/guides/connect-network.md
@@ -2,10 +2,12 @@
 
 Depending on your enterprise network security, you may need to use [SHH tunneling](https://www.ssh.com/academy/ssh/tunneling/example#local-forwarding), or port forwarding, to allow Flow
 to securely access your endpoint. This is configured within a capture or materialization definition, but
-before you can do this, you'll need a properly configured SSH server on your internal network.
+before you can do this, you'll need a properly configured SSH server on your internal network or cloud hosting platform.
 
-These steps provide a basic roadmap, and should be followed with the help of an IT specialist in your organization,
-as details depend on your organization's policies and practices.
+This guide includes setup steps for popular cloud platforms,
+as well as generalized setup that provide a basic roadmap for on-premise servers or other cloud platforms.
+
+## General setup
 
 1. Activate an [SSH implementation on a server](https://www.ssh.com/academy/ssh/server#availability-of-ssh-servers), if you don't have one already.
 Consult the documentation for your server's operating system and/or cloud service provider, as the steps will vary.
@@ -14,7 +16,7 @@ basic configuration options.
 
 2. Referencing the config files and shell output, collect the following information:
 
-   * The **endpoint** for the SSH server, formatted as `ssh://hostname[:port]`. This may look like the any of following:
+   * The **SSH endpoint** for the SSH server, formatted as `ssh://hostname[:port]`. This may look like the any of following:
      * `ssh://ec2-198-21-98-1.compute-1.amazonaws.com`
      * `ssh://198.21.98.1`
      * `ssh://198.21.98.1:22`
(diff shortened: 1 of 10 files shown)


dyaffe (commented):
For AWS SSH server, step 1, I may just link to our SSH article when say "Find your public SSH key as created per this article", since it still has to be of that format and not just any will work.

Same with Step 1 for Google and Azure.

For adding SSH keys in Google, the easiest means is just to add them in the UI [here](https://console.cloud.google.com/compute/metadata?tab=sshkeys) vs via the api in the article mentioned in step 3.  I know that even Phil used that method.

One other related thing is that I notice you're referencing docs on creating these all via the CLI vs UI.  It probably makes sense for us to choose one and be consistent.  I would *think* that the UI would be simpler but it may be good to get an engineer to verify that.

oliviamiannone:
@dyaffe thanks for the review!

> For AWS SSH server, step 1, I may just link to our SSH article when say "Find your public SSH key as created per this article", since it still has to be of that format and not just any will work.
> 
> Same with Step 1 for Google and Azure.

I'm a bit confused what you mean here. I think you're referring to the various sections of our SSH article, "Configure connections with SSH tunneling." I chose to break that article up into different set up steps for different use-cases: The generalized steps (what's live currently); and tailored steps for AWS, Google, and Azure. So, adding that link would just be linking to a different copy of the same step in the same article. 

I decided to set it up the way I did so that user only has to follow one set of steps and doesn't need to jump around the page. Some of the steps are duplicated for that reason -- so step 3 in the original workflow was copy-pasted into step 1 for the various managed platforms. It's just the ordering that changed. 

Let me know if I've misunderstood what you're trying to say, though. 

>For adding SSH keys in Google, the easiest means is just to add them in the UI [here](https://console.cloud.google.com/compute/metadata?tab=sshkeys) vs via the api in the article mentioned in step 3. I know that even Phil used that method.

I was digging into this and found this [article from Google](https://cloud.google.com/compute/docs/connect/add-ssh-keys). It's interesting because it says that to use the SSH keys for OS login, you have to use the CLI/API method. And when you use the UI (which is the Metadata configuration), they say that's for non-OS-login VMs. So would this mean that we _don't_ actually require OS login? 

> One other related thing is that I notice you're referencing docs on creating these all via the CLI vs UI. It probably makes sense for us to choose one and be consistent. I would think that the UI would be simpler but it may be good to get an engineer to verify that.

Good point - that should be consistent. I checked all the links. Amazon and Google are both pretty good about having instructions for both their consoles and CLI in the same docs, so we're good on all those links. Azure has _much_ more thorough docs on their CLI/API vs their Portal, but I was able to find stuff for the SSH docs. So now, the only CLI-only docs we link to should be in the Postgres connector docs for Azure. It's just one link, and I think that's find because the user will have to be running SQL commands for that, anyway.

dyaffe:
For my first point, I was saying that in the article "Configure connections with SSH tunneling", in the first part of each section that talks about specific cloud providers you have a sentence that begins "Begin by finding your public SSH key on your local machine. In the .ssh".   My suggestion is that that sentence should be replaced with something like "Find or create your public SSH key as created per the General Setup portion of this article".  I agree with your chosen route of breaking the article up, but was just commenting on that one sentence.

re: your comment: "I was digging into this and found this [article from Google](https://cloud.google.com/compute/docs/connect/add-ssh-keys). It's interesting because it says that to use the SSH keys for OS login, you have to use the CLI/API method. And when you use the UI (which is the Metadata configuration), they say that's for non-OS-login VMs. So would this mean that we don't actually require OS login?"  -- @psFried Olivia seems to have a point.  Can you verify?

The rest looks good to me.

oliviamiannone:
Going to go ahead and merge, even with that outstanding question about OS login and SSH keys in Google. Changed the link to the top of Google's article to leave it up to the user which method to use, and it's not like it's going to _hurt_ anything to use OS login even if it's not necessary. We can always adjust later.

PR #412, opened 2022-03-16
Fix malformed `flowctl deploy` example usage

willdonnelly:
**Description:**

The command given for `flowctl deploy` in the "Create a simple data flow" guide was broken in what looked like a simple copy-paste error.

https://docs.estuary.dev/guides/create-dataflow/


oliviamiannone (approved):
Thanks for catching this!

PR #414, opened 2022-03-17
Docs - materialize-postgres

oliviamiannone:
**Description:**

Adds docs for materialize-postgres. 

Tested the connector via SSH to a Google Cloud database, and added docs to cover a few gotcha's there. I intend to test in AWS and Azure too, but don't have time immediately and want to get the essentials pushed out. 

**Notes for reviewers:**

- Config descriptions and titles were added by me, referencing source-postgres, because none were provided by this connector. Similarly, there was no README, which I would otherwise reference.
- The fields that are flagged as required seemed a bit odd to me - I feel like they should _all_ be. But this is what came out of the schemalate markdown generator, so...
-  Another thing that we found during testing that strikes me as odd: SSH to a hosted DB on Gcloud required that the DB use a _private_ IP, while source-postgres uses a public IP. This is probably tangential but noting just in case.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/414)

### DIFF
--- a/site/docs/reference/Connectors/materialization-connectors/PostgreSQL.md
+++ b/site/docs/reference/Connectors/materialization-connectors/PostgreSQL.md
@@ -76,6 +76,6 @@ for additional details and a sample.
 :::tip
 You can find the values for `forwardHost` and `forwardPort` in the following locations in each platform's console:
 * Amazon RDS: `forwardHost` as Endpoint; `forwardPort` as Port.
-* Google Cloud SQL: `forwardHost` as Public IP Address; `forwardPort` is always `5432`.
+* Google Cloud SQL: `forwardHost` as Private IP Address; `forwardPort` is always `5432`. You may need to [configure private IP](https://cloud.google.com/sql/docs/postgres/configure-private-ip) on your database.
 * Azure Database: `forwardHost` as Server Name; `forwardPort` under Connection Strings (usually `5432`).
 :::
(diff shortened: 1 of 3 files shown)


site/docs/reference/Connectors/materialization-connectors/PostgreSQL.md
```diff
@@ -0,0 +1,81 @@
...
+
+[`ghcr.io/estuary/materialize-postgres:dev`](https://ghcr.io/estuary/materialize-postgres:dev) provides the latest connector image. You can also follow the link in your browser to see past image versions.
+
+## Prerequisites
+
+To use this connector, you'll need:
+
+* An existing catalog spec that includes at least one collection.
+* A Postgres database to which to materialize, and user credentials.
+  Optionally, you can choose existing tables to materialize to; otherwise, the tables you specify will be created by the connector.
```
mdibaiee, line 12: There was a discussion on the Firebolt PR, my understanding of which is that materialize-ostgres always tries to create a table unless materialize-postgres itself has made the table before, which means it does not allow using of a table made by users, see the discussion here: https://github.com/estuary/flow/pull/407#discussion_r834579173

So I think we might have to edit this part to say that the connector creates the table unless. cc @jgraettinger

oliviamiannone: Thank you for pointing this out - I'm sure that's correct, I was making a bit of an assumption here.

oliviamiannone:
Merging for now in the interest of getting minimal connector docs up - although further testing would be ideal in the future

PR #416, opened 2022-03-18
Adding search to docs

travjenkins:
**Description:**

Adding a basic local search to the docs. The search index is created up during build time and then the search is handled in the browser. This should scale for a little while but eventually we might end up with larger docs that this is not as usable. 

Also removed the blog files as those were getting deployed and I feel like we probably don't want that happening.

**Workflow steps:**

N/A

No links affected but this changes Docs overall by adding search

**Notes for reviewers:**

To test locally you have to do a `build` and then `serve` as this does not work through dev mode.

This library does not support darkmode right now. Someone has an issue open with a solution and the maintainers are asking them to open up a PR with it. We can wait for that, implement our own, or just leave it. It does not feel too terrible to me.

Please use the search and see if it feels usable and okay. I mainly searched for captures, materializations, and catalogs and that felt okay to me.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/416)

**Screenshots:**

image

image


site/docs/concepts/advanced/shards.md
```diff
@@ -53,7 +53,7 @@ Shard stores use
 [recovery logs](https://gazette.readthedocs.io/en/latest/consumers-concepts.html#recovery-logs)
 to replicate updates and implement transaction semantics.
 
-Recovery logs are regular [journals](/workspaces/flow/site/docs/concepts/advanced/journals.md),
+Recovery logs are regular [journals](./journals.md),
```
travjenkins, line 56: This was a link that the build was complaining about so switched it.

site/docusaurus.config.js
```diff
@@ -18,6 +18,17 @@ const config = {
...
+    [
+      require.resolve('docusaurus-lunr-search'), 
+      {
+        // @ts-ignore
+        excludeRoutes: [
+            'blog/**/*',
+        ]
+    }]
+  ],
+
```
travjenkins, line 31: Enabling the new plugin. Also - for now I think excluding the blog is best as it is being served and the search made it easy to get to on accident.

I also just ignored the type checking stuff for this library as it does not have anything type related  I could find.

site/src/css/custom.css
```diff
@@ -26,3 +26,7 @@
 html[data-theme='dark'] .docusaurus-highlight-code-line {
   background-color: rgba(0, 0, 0, 0.3);
 }
+
+.navbar__search {
+  padding-left: var(--ifm-navbar-padding-horizontal) !important;
+}
```
travjenkins, line 32: By default the search bar smashes into the darkmode slider.

oliviamiannone (approved):
Looks great! Tested a bunch of terms and seems to be indexing everything as it should, and the way the results are laid out in the pop-up looks really nice. I agree that it's probably not the final solution - for one thing, we'll get to a point where when you hit "enter" you want it to take you to a full page of results rather than the first result that's there. But agree that for the near-term this will totally meet our needs.

PR #432, opened 2022-04-05
Elasticsearch and S3-parquet connector docs

oliviamiannone:
**Description:**

Adds docs for `materialize-s3-parquet` and `materialize-elasticsearch`. This catches us up on connector docs. 

**Notes for reviewers:**

With the exception of most of the `field_overrides` in elasticsearch, I wrote the property descriptions - they're either not present in the config or I wasn't able to find them.

Some outstanding questions because the original developer isn't here and there are no READMEs. Specifically, with Elastic, I used up my free trial a while ago and wasn't able to test, so this needs review from someone familiar with that platform. 
* Prerequisites and permissions for elastic - are these correct?
* Is there anything worth mentioning in the intro for the mechanism of this connector that'd be pertinent to the user?

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/432)

### DIFF
--- a/site/docs/reference/Connectors/materialization-connectors/Elasticsearch.md
+++ b/site/docs/reference/Connectors/materialization-connectors/Elasticsearch.md
@@ -28,9 +28,11 @@ If you haven't yet captured your data from its external source, start at the beg
 To use this connector, begin with data in one or more Flow collections.
 Use the below properties to configure an Elasticsearch materialization, which will direct the contents of these Flow collections into Elasticsearch indices.
 
-You must indicate the desired Elasticsearch field type for each field in your Flow collection.
-You configure this in the `field_overrides` array for each binding.
-To do so, provide a JSON pointer to the field in the collection schema, choose the output field type, and specify additional properties, if necessary.
+By default, the connector attempts to map each field in the Flow collection to the most appropriate Elasticsearch [field type](https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-types.html).
+However, because each JSON field type can map to multiple Elasticsearch field types,
+you may want to override the defaults.
+You can configure this by adding `field_overrides` to the collection's [binding](#bindings) in the materialization specification.
+To do so, provide a JSON pointer to the field in the collection schema, choose the output field type, and specify additional properties, if necessary. 
 
 ### Properties
 
--- a/site/docs/reference/Connectors/materialization-connectors/Elasticsearch.md
+++ b/site/docs/reference/Connectors/materialization-connectors/Elasticsearch.md
@@ -13,7 +13,7 @@ This connector materializes Flow collections into indices in an Elasticsearch cl
 To use this connector, you'll need:
 
 * An Elastic cluster with a known [endpoint](https://www.elastic.co/guide/en/elasticsearch/reference/current/getting-started.html#send-requests-to-elasticsearch)
-  * If the cluster is on the Elastic Cloud, you'll need an Elastic user with a role that grants all privileges on indices in the cluster.
+  * If the cluster is on the Elastic Cloud, you'll need an Elastic user with a role that grants all privileges on indices you plan to materialize to within the cluster.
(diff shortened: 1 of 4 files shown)


site/docs/reference/Connectors/materialization-connectors/Elasticsearch.md
```diff
@@ -0,0 +1,96 @@
...
+If you haven't yet captured your data from its external source, start at the beginning of the [guide to create a dataflow](../../../guides/create-dataflow.md). You'll be referred back to this connector-specific documentation at the appropriate steps.
+:::
+
+
+## Configuration
+
+To use this connector, begin with data in one or more Flow collections.
+Use the below properties to configure an Elasticsearch materialization, which will direct the contents of these Flow collections into Elasticsearch indices.
+
+You must indicate the desired Elasticsearch field type for each field in your Flow collection.
```
mdibaiee, line 31: This is optional: Since Elasticsearch itself is a JSON storage, by default the connector saves the JSON document by translating its types to Elasticsearch as best as it can. `field_overrides` gives users the ability to override the type translation for a specific field (there are more than a single type in Elasticsearch for each JSON data type, for example a string in JSON can be translated to either `text` or `keyword` or both in Elasticsearch, hence the `field_overrides` which lets the user customise the translation).

mdibaiee (approved):
The Elasticsearch one LGTM! I don't know much about Parquet though

site/docs/reference/Connectors/materialization-connectors/Parquet.md
```diff
@@ -0,0 +1,81 @@
...
+```yaml
+materializations:
+  tenant/mat_name:
+	  endpoint:
+        connector:
+          config:
+            awsAccessKeyId: AKIAIOSFODNN7EXAMPLE
+            awsSecretAccessKey: wJalrXUtnFEMI/K7MDENG/bPxRfiCYSECRET
+            bucket: my-bucket
+            uploadIntervalInSeconds: 10
```
psFried, line 67: nit: `uploadIntervalInSeconds` should probably be `300` here, just so it's a more typical example. We should probably also update the connector to disallow values below some reasonable threshold, but I think we'll leave that for a future exercise. A realistic lower bound might be 1 minute.

site/docs/reference/Connectors/materialization-connectors/Elasticsearch.md
```diff
@@ -0,0 +1,98 @@
...
+| _`/field_overrides/-/es_type/keyword_spec/ignore_above`_ | Ignore above | Strings longer than the ignore&#x5F;above setting will not be indexed or stored. See [Elasticsearch docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/ignore-above.html) | integer |  |
+| _`/field_overrides/-/es_type/text_spec`_ | Text specifications | Configuration for the text field, effective if field&#x5F;type is &#x27;text&#x27;. | object |  |
+| _`/field_overrides/-/es_type/text_spec/dual_keyword`_ | Dual keyword | Whether or not to specify the field as text&#x2F;keyword dual field. | boolean |  |
+| _`/field_overrides/-/es_type/text_spec/keyword_ignore_above`_ | Ignore above | Effective only if Dual Keyword is enabled. Strings longer than the ignore&#x5F;above setting will not be indexed or stored. See [Elasticsearch docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/ignore-above.html). | integer |  |
+| _`/field_overrides/-/pointer`_ | Pointer | A &#x27;&#x2F;&#x27;-delimitated json pointer to the location of the overridden field. | string |  |
```
psFried, line 62: sp: delimitated => delimited

psFried (approved):
I think this all looks good for now. Most of the things I'd take issue with here are things that need addressed in the connector itself. But the docs seem to accurately reflect the current state, and they seem pretty easy to follow. :rocket:

PR #440, opened 2022-04-12
sql/validate: don't allow field names >63 bytes in length

mdibaiee:
**Description:**

- Field names over 63 bytes are not allowed in Postgres and will get truncated
- If there are multiple fields with such long names and they share a prefix that's 63 bytes long, they will become duplicate column names
- This PR introduces a limit of 63 bytes for all SQL materializations as a general rule

**Workflow steps:**

N/A

N/A

**Notes for reviewers:**

Also see: https://estuary-dev.slack.com/archives/C01G7CFNA8K/p1649772559301679

Solves https://github.com/estuary/connectors/issues/166

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/440)


jgraettinger (approved):
LGTM

PR #443, opened 2022-04-13
materialize/sql: reserved words in validation logic

mdibaiee:
**Description:**

- Allows SQL materializations to specify a set of reserved words that will be used in validating projections

**Workflow steps:**

- Specify `ReservedWords` when creating a SQL materialization using our materialize SQL library

N/A

**Notes for reviewers:**

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/443)

### DIFF
--- a/go/materialize/driver/sqlite/validate_test.go
+++ b/go/materialize/driver/sqlite/validate_test.go
@@ -17,6 +17,7 @@ import (
 )
 
 func TestValidations(t *testing.T) {
+	var driver = &sqlDriver.Driver{}
 	var args = bindings.BuildArgs{
 		Context:  context.Background(),
 		FileRoot: "./testdata",
@@ -38,7 +39,7 @@ func TestValidations(t *testing.T) {
 		t.Run(
 			fmt.Sprintf("NewSQLProjections-%s", path.Base(spec.Collection.String())),
 			func(t *testing.T) {
-				constraints := sqlDriver.ValidateNewSQLProjections(spec, false)
+				constraints := driver.ValidateNewSQLProjections(spec, false)
 				cupaloy.SnapshotT(t, constraints)
 			})
 	}
@@ -74,7 +75,8 @@ func testMatchesExisting(t *testing.T, collection *pf.CollectionSpec) {
 	var stringProjection = proposed.GetProjection("string")
 	stringProjection.Inference.Exists = pf.Inference_MUST
 
-	var constraints = sqlDriver.ValidateMatchesExisting(&existingSpec, &proposed)
+	var driver = &sqlDriver.Driver{}
+	var constraints = driver.ValidateMatchesExisting(&existingSpec, &proposed)
 	var req = []string{"theKey", "string", "bool", "flow_document"}
 	for _, field := range req {
 		var constraint, ok = constraints[field]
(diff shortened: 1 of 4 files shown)


jgraettinger (commented):
Thanks. Rather than outright prohibiting these fields, I think we want to quote them.

The way this could instead work, is that the [Renderer.skipWrapper](https://github.com/estuary/flow/blob/master/go/protocols/materialize/sql/renderer.go#L10-L19) would additionally consult the list of reserved keywords and quote any which are present.

I also think it's okay to use a global compile-time const list of keywords for _all_ SQL database ([like this one](https://en.wikipedia.org/wiki/SQL_reserved_words)) if that eases implementation.

> Reviewed 2 of 3 files at r1.
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/443#-:-N-Y2Xok4WlmCesZwfTk:bsyesh3)* status: 2 of 4 files reviewed, all discussions resolved (waiting on @jgraettinger)

jgraettinger:
Some DB's, like Snowflake, essentially require that columns, once quoted, _always_ be accessed through quotes -- which means our choice of reserved keywords has a user-facing component where they must recognize reserved fields are quoted and do the same at query time.

For that reason I think it's probably better to have a global, external, easily-consulted list like the Wikipedia one that they can reference to understand what's gonna be quoted and why, even if it over-quotes with respect to some specific DB dialect.

psFried:
I like the idea of just quoting reserved keywords instead of restricting them.

> For that reason I think it's probably better to have a global, external, easily-consulted list like the Wikipedia one that they can reference to understand what's gonna be quoted and why, even if it over-quotes with respect to some specific DB dialect.

TBH I don't think this will improve the user experience over having separate reserved word lists for each database. My experience is that the majority of users are actually pretty familiar with _their_ databases SQL dialect, but have very little understanding or expectation that other databases do things differently. For example, if we quoted a column called `uid` in Oracle (where it's a reserved keyword), I'd think most oracle users wouldn't bat an eye. But if you were a Snowflake user and you're already accustomed to querying a `uid` column without quotes, you might be a bit miffed to find that we've quoted it. Companies also tend to stick with one or two database systems, and often standardize on column names that avoid reserved words in the SQL dialects they use, so over-quoting could change someone's workflow from never/rarely using quotes to frequently using them. Just my 2 cents.

jgraettinger:
Fair. I agree the ideal would be separate, documented and easily discovered reserved keyword lists for each connector. It's hard to justify the relative difference of work required to accomplish this, however. Note also that materializations are adding _new_ tables, so even if they're used to not quoting `uid` with some _other_ table, we're deviating from but not _breaking_ that convention.

jgraettinger:
@mdibaiee let's take the simpler route for now?

Another consideration is that a more flexible quoting regimen also requires that we have a process for _evolving_ the list of keywords w.r.t. running materializations, which is _definitely_ out of scope right now.

mdibaiee:
@jgraettinger @psFried I think I agree that we better have a minimal list of reserved words for each SQL dialect, but doing this seems straightforward (I tried it here in [this commit](https://github.com/estuary/connectors/pull/170/commits/56347bdff3b926526a930db3cb1f7c957201763d)).

We currently have PostgreSQL, Snowflake and BigQuery. We will soon have more SQL materializations potentially, which might make this more painful to do properly later on. I think I can just do these 3 now and then we can follow the pattern for new materializations as they come. WDYT?

mdibaiee:
Actually it seems for BigQuery we are quoting everything, so we don't need that PR, will close.

jgraettinger:
Sounds good, I yield to the person doing the actual work :)

PR #444, opened 2022-04-13
docs: document PostgreSQL reserved words

mdibaiee:
**Description:**

- Documents PostgreSQL's reserved words

**Workflow steps:**

N/A

https://docs.estuary.dev/reference/Connectors/materialization-connectors/PostgreSQL/

**Notes for reviewers:**

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/444)

### DIFF
--- a/site/docs/reference/Connectors/materialization-connectors/PostgreSQL.md
+++ b/site/docs/reference/Connectors/materialization-connectors/PostgreSQL.md
@@ -78,3 +78,7 @@ You can find the values for `forwardHost` and `forwardPort` in the following loc
 * Google Cloud SQL: `forwardHost` as Private IP Address; `forwardPort` is always `5432`. You may need to [configure private IP](https://cloud.google.com/sql/docs/postgres/configure-private-ip) on your database.
 * Azure Database: `forwardHost` as Server Name; `forwardPort` under Connection Strings (usually `5432`).
 :::
+
+## Reserved Words
+
+PostgreSQL has a list of reserved words that must be quoted in order to be used as an identifier. Flow automatically quotes fields that are in the reserved words list. You can find this list in PostgreSQL's documentation [here](https://www.postgresql.org/docs/current/sql-keywords-appendix.html) as well as [here](https://go.estuary.dev/6DrTYE) in the connector source code. Flow considers all the reserved words which are marked as "reserved" in either of the columns in the official documentation.


site/docs/reference/Connectors/materialization-connectors/PostgreSQL.md
```diff
@@ -77,4 +77,10 @@ You can find the values for `forwardHost` and `forwardPort` in the following loc
 * Amazon RDS: `forwardHost` as Endpoint; `forwardPort` as Port.
 * Google Cloud SQL: `forwardHost` as Private IP Address; `forwardPort` is always `5432`. You may need to [configure private IP](https://cloud.google.com/sql/docs/postgres/configure-private-ip) on your database.
 * Azure Database: `forwardHost` as Server Name; `forwardPort` under Connection Strings (usually `5432`).
-:::
\ No newline at end of file
+:::
+
+## Notes
```
oliviamiannone, line 82: I'd suggest just making "reserved words" the H2 and take this out.

site/docs/reference/Connectors/materialization-connectors/PostgreSQL.md
```diff
@@ -77,4 +77,10 @@ You can find the values for `forwardHost` and `forwardPort` in the following loc
...
+
+### Reserved Words
+
+PostgreSQL has a list of reserved words that must be quoted in order to be used as an identifier. Flow automatically quotes fields that are in the reserved words list. You can find this list in PostgreSQL's documentation [here](https://www.postgresql.org/docs/current/sql-keywords-appendix.html) as well as [here](https://go.estuary.dev/6DrTYE) in the connector source code.
```
oliviamiannone, line 86: I'm looking at the table in Postgres docs and the list in the code, and seems like there are significantly fewer reserved words in Flow than in postgres. From my perspective, it's hard to figure out why, as it doesn't seem to align with what any of the subsequent columns are saying in the postgres docs. Could you clarify that for me?

Either way, in light of that, and although it'd be super long, I might actually suggest putting a list of our reserved words directly in the docs here. Especially for people who are using Flow from UI-only in the future, it's a better UX to be able to just come to our docs and see the reserve words and not have to look at the Go file. I can add this myself after this PR is merged, just cause it might take some time to play with formatting and make sure it looks good.

oliviamiannone, line 86: Pretty sure this is expected since it's not merged, but I do want to call out that the link to the connector source code is broken currently.

oliviamiannone (commented):
Thanks for doc'ing this! A few suggestions/questions.

site/docs/reference/Connectors/materialization-connectors/PostgreSQL.md
mdibaiee: The link should be okay now!

mdibaiee: I added an explanation there, but basically if any word is marked as "reserved" in any of the columns in the official docs, we consider it as reserved. I personally don't understand either why there are some rows which are not marked as "reserved" in any of the columns 🤔

site/docs/reference/Connectors/materialization-connectors/PostgreSQL.md
mdibaiee: Done 👍🏽

site/docs/reference/Connectors/materialization-connectors/PostgreSQL.md
mdibaiee: I also agree with your idea that UI users will prefer seeing the table directly here 👍🏽

oliviamiannone: Ah, thanks for clearing that up. Looks good, think this is good to go!

Issue #455, opened 2022-04-20
Connector Proxy: forward stderr of connector out


mdibaiee:
Currently connector proxy is swallowing up the stderr of connector if it fails. We can just forward it to help with debugging:
https://github.com/estuary/flow/blob/master/crates/connector_proxy/src/main.rs#L207

PR #456, opened 2022-04-20
connector-proxy: let delayed process inherit stderr

mdibaiee:
**Description:**

- We used to eat up stderr of connector and only output it in case of error. With this change it's just forwarded to the parent stderr.

Solves #455 

**Workflow steps:**

N/A

N/A

**Notes for reviewers:**

N/A

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/456)


mdibaiee:
Got approval from Johnny on Slack.

PR #458, opened 2022-04-21
connector-proxy: inherit stderr for child and parallelize streaming of I/O

mdibaiee:
**Description:**

- Use `Stdio::inherit()` for stderr of child process in connector-proxy, this simplifies the logic and removes the need for a manual copy in of stderr. We don't process anything of the stderr anyway.
- Because we used to join our stdin/stdout streams concurrently, rather than in parallel, if one of them was stuck, it would block all the others from working as well. So I changed that to use a parallel copy which allows the stdin to reach the process.
- Added a test with `tail -f` (which will not terminate its stdout until an EOF in stdin is received) to test this: with the concurrent approach, this gets blocked forever, with the parallel approach and stderr being inherited, this allows the EOF in stdin to reach the process without being blocked by the stdout and exits as soon as EOF is reached in stdin.

**Workflow steps:**

N/A

N/A

**Notes for reviewers:**

My main concern with this PR is use of `tail -f` which is available on most Linux and BSD machines through coreutils, but I don't know if there is any guarantee of that.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/458)

### DIFF
--- a/crates/connector_proxy/src/connector_runner.rs
+++ b/crates/connector_proxy/src/connector_runner.rs
@@ -9,7 +9,7 @@ use crate::libs::command::{
     check_exit_status, invoke_connector_delayed, invoke_connector_direct, parse_child,
 };
 use tokio::io::copy;
-use tokio::process::{ChildStderr, ChildStdin, ChildStdout};
+use tokio::process::{ChildStdin, ChildStdout};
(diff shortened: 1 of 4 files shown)


jgraettinger (approved):
LGTM!

PR #459, opened 2022-04-21
Add service account impersonation section to source-gcs docs

oliviamiannone:
**Description:**

Documents the workflow enabled by https://github.com/estuary/connectors/pull/168

Just a few explanatory paragraphs, and then basically used the same config sample from the PR comment (with slight changes). This is a workflow for advanced users/admins that should be documented but doesn't need to be accessible for all users, so docs aren't very in-depth.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/459)

### DIFF
--- a/site/docs/reference/Connectors/capture-connectors/gcs.md
+++ b/site/docs/reference/Connectors/capture-connectors/gcs.md
@@ -62,7 +62,7 @@ captures:
       connector:
         image: ghcr.io/estuary/source-gcs:dev
         config:
-          bucket: "my-bucket"
+          bucket: my-bucket
     bindings:
       - resource:
           stream: my-bucket/${PREFIX}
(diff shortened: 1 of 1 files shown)


site/docs/reference/Connectors/capture-connectors/gcs.md
```diff
@@ -73,4 +73,35 @@ captures:
 
 Your capture definition may be more complex, with additional bindings for different GCS prefixes within the same bucket.
 
-[Learn more about capture definitions.](../../../concepts/captures.md#pull-captures)
\ No newline at end of file
+[Learn more about capture definitions.](../../../concepts/captures.md#pull-captures)
+
+### Configure Google service account impersonation
```
willdonnelly, line 78: Maybe change the heading to something like "Advanced Usage: Service Account Impersonation" to make it clearer to the casual reader that this isn't a normal/necessary part of using the GCS connector?

site/docs/reference/Connectors/capture-connectors/gcs.md
```diff
@@ -73,4 +73,35 @@ captures:
...
+```yaml
+config:
+  bucket: <bucket-name>
+  googleCredentials_sops:
+    # URL containing the account to impersonate and the associated project
+    service_account_impersonation_url: https://iamcredentials.googleapis.com/v1/projects/-/serviceAccounts/<target-account>@<project>.iam.gserviceaccount.com:generateAccessToken
+    # Credentials for the account that has been configured to impersonate the target
+    source_credentials:
+        client_email: <origin-account>@<anotherproject>.iam.gserviceaccount.com
+        client_id: ...
```
willdonnelly, line 100: I'd snip out all the fields where I just wrote `...`, because that's not really an exhaustive list and the idea that I'm trying to convey there was just "copy-paste the same credentials JSON/YAML you would normally use into the `source_credentials` property".

willdonnelly (approved):
LGTM, a couple of minor notes

Issue #299, opened 2021-11-24
Connector Log Levels


saterus:
I've been doing a lot of debugging with our new logging system. This has exposed a gap in the Airbyte spec related to log level. There's no way to tell a connector what its log level should be. This means that changing a task's log level does not actually cause the connector log level to change. They are stuck at the default level.

Given that there's no way specified in the Airbyte spec, we are left in an uncertain position. We could try and raise an issue in the Airbyte repo with them (to which they may not respond or implement), or we can address this only in our own connectors (which means Airbyte-authored connectors are different, again).

Either way, it would be useful to allow setting the log level. We have a few options for doing this available to us:

1. Add another extension attribute, `estuary.dev/log-level`, and allow setting it in a connector config file.
2. Decide on a well-known environment variable to read in each connector.

In either case, the runtime could launch the connector with the appropriate log level. Changing the shard's log level would cause the connector to restart regardless, so I don't think we'd need to change it on the fly.

psFried:
@saterus  do you have an opinion on which of those two options we should use?

My own 2 cents is that a well-known environment variable (`CONNECTOR_LOG_LEVEL`) is something that we could both start using now, and also propose adding to the spec. Looking at the current spec, I'm not really seeing a benefit to putting in any of those files, especially since some applications may want to configure logging before attempting to parse them.

jgraettinger:
A third option would be to use a field that's part of the connector's specific specification? This would maintain compatibility with Airbyte, makes it explicit whether a specific connector is going to even do anything with the value, and builds in flexible semantics and verification of what values may look like (are the `INFO` vs `info`? do they allow separate levels by module ?) -- which could otherwise be confusing in practice given that it's not a day one built-in.

To be clear, this is a convention only, not a spec extension. It wouldn't be part of the ConfiguredCatalog definition or require a `estuary.dev/log-level` extension. Connectors would include it in their generated spec under a presumably conventional name, but even that isn't required.

saterus:
> A third option would be to use a field that's part of the connector's specific specification? [ ... ] To be clear, this is a convention only, not a spec extension. 

This seems tricky to automate. If we want to control log levels at a shard level, we wouldn't necessarily want/expect the user to add these fields to a catalog yaml. 

[...]

PR #461, opened 2022-04-22
go/connector/run: pass --log.level to connector proxy

mdibaiee:
**Description:**

- Pass `--log.level` in connector runner to connector proxy

**Workflow steps:**

Pass `--log.level` to flowctl and see it being passed to connector proxy.

N/A

**Notes for reviewers:**

N/A

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/461)

### DIFF
--- a/crates/connector_proxy/src/errors.rs
+++ b/crates/connector_proxy/src/errors.rs
@@ -81,7 +81,7 @@ where
         match self {
             Ok(t) => t,
             Err(e) => {
-                tracing::debug!(error_details = ?e);
+                tracing::error!(error_details = ?e);
                 std::process::exit(1);
             }
         }
--- a/go/connector/run.go
+++ b/go/connector/run.go
@@ -135,6 +135,7 @@ func Run(
 
 	args = append([]string{
 		fmt.Sprintf("--image-inspect-json-path=/tmp/%s", imageInspectJsonFileName),
+		fmt.Sprintf("--log.level=%s", logger.Level().String()),
 		protocol.proxyCommand(),
 	}, args...)
 


willdonnelly:
This LGTM as-is, but two thoughts:

1. Have you tested the workflow steps you described above to see if that's where the log level value is coming from? Because I think the relevant log level may actually be controlled by editing the shard spec (which is what we'd prefer, AFAIK).

2. Another thought, since this gets very close to [a bit of functionality](https://github.com/estuary/flow/issues/299) we've been wanting for a while: does the `connector_proxy` command-line parsing look for environment variables the way our Go boilerplate does? Because if so setting `LOG_LEVEL` might be an even better way to plumb this through, because many of our Airbyte connectors would start automagically responding to the setting as well.

mdibaiee:
@willdonnelly

1. I'm actually not 100% sure if the --log.level comes from the deployment or flow or the shard spec, _probably_ the shard spec actually, but I will have to try that out.
2. That's a good point. I think at the moment we do not use `LOG_LEVEL` in connector-proxy, but I can make that happen. For now I'd say let's unblock ourselves with this.

mdibaiee:
When testing locally, the `logger.Level()` comes from this command:

```
flowctl deploy --source deepsync-marketing/source-facebook-marketing.flow.yaml --log.level=info
```

But I am not sure what that corresponds to when deploying to a remote control plane. 🤔

jgraettinger:
The `logger` instance passed to connector.Run is built in runtime/task_term.go, and it does [get its level from the shard spec](
https://github.com/estuary/flow/blob/2c68ea638c2b7029ef51bd13f1976ef7335755e0/go/runtime/ops.go#L107), so I believe this all works fine.

Issue #315, opened 2021-12-16
maybe: support individual schemas with > 64 properties


jgraettinger:
To make schema validation fast, we use an intern table strategy where we track observed properties using assigned bits in a 64-bit word. Then we can compare against required properties of a schema using a bitmask. The intern table is specific to a single JSON schema, and different schemas, or in-place sibling applications of a common parent schema, use separate intern tables which can each be up to 64 items large.

We've seen some... _questionable_ schema documents from third-party connectors that produce schemas with significantly more than 64 properties. "questionable" because there's most certainly an element of garbage-in, garbage-out going on here, and it's not yet clear to me that it's _actually required_ to support larger schemas at this time, but... here's a tracking issue. 🤷

mdibaiee:
One way to work around this is to wrap `allOf` around the JSONSchema when we are compiling it for our own code's usage. We need to do this ensuring that the JSONSchema passed to the connector stays intact because the code that handles the JSONSchema is usually not 100% compatible with spec and we see this kind of code commonly:

```
if x['type'] == 'object':
  props = x['properties']
```

This fails if we modify the JSONSchema below:

```
type: object
properties:
  name:
    type: string
```

To:

```
type: object
allOf:
  - properties:
      name:
        type: string
```

So we need to do this transparently.

jgraettinger:
[Here is where](https://github.com/estuary/flow/blob/master/crates/json/src/schema/build.rs#L258-L266) we process the set of properties of a schema.

We can examine the size of this array, and if > 64 then process it into equal chunks < 64 which are handled as `App::AllOf` applications.

PR #465, opened 2022-04-25
json/schema: fix the 64 field limit

mdibaiee:
**Description:**

- Introduce a new `Inline` variant to JSONSchema build to work around the 64 intern table limit for `properties`

solves #315 

**Workflow steps:**

- Try a schema with more than 64 fields under a `properties` keyword

N/A

**Notes for reviewers:**

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/465)

### DIFF
--- a/crates/json/src/schema/build.rs
+++ b/crates/json/src/schema/build.rs
@@ -1,9 +1,12 @@
+use std::iter::FromIterator;
+
 use crate::schema::{
     intern, keywords, types, Annotation, Application, CoreAnnotation, HashedLiteral, Keyword,
     Schema, Validation,
 };
 use crate::{de, NoopWalker, Number};
 use fancy_regex as regex;
+use itertools::Itertools;
 use serde::Deserialize;
 use serde_json as sj;
 use thiserror;
@@ -120,7 +123,7 @@ where
                 K::RecursiveAnchor => 0,
 
                 // Properties / PatternProperties conditions whether AdditionalProperties applies.
-                K::Application(A::Properties { .. }, _) => 2,
+                K::Application(A::Properties { .. }, _) | K::Application(A::Inline, _) => 2,
                 K::Application(A::PatternProperties { .. }, _) => 3,
                 // AdditionalProperties also conditions whether UnevaluatedProperties applies.
                 K::Application(A::AdditionalProperties, _) => 4,
@@ -256,12 +259,27 @@ where
             keywords::PROPERTY_NAMES => self.add_application(App::PropertyNames, v)?,
             keywords::PROPERTIES => match v {
                 sj::Value::Object(m) => {
-                    for (prop, child) in m {
-                        let app = App::Properties {
-                            name: prop.clone(),
-                            name_interned: self.tbl.intern(prop)?,
-                        };
-                        self.add_application(app, child)?;
+                    if m.len() > intern::MAX_TABLE_SIZE {
+                        let chunks = m.iter().chunks(intern::MAX_TABLE_SIZE);
+                        for group in chunks.into_iter() {
+                            // Create a new `properties` with $MAX_TABLE_SIZE or less items
+                            // and add it as part of an `inline`
+                            let group_children = sj::json!({
+                                "properties": sj::Map::from_iter(
+                                    group.map(|(a, b)| (a.to_owned(), b.to_owned())),
+                                )
+                            });
+
+                            self.add_application(App::Inline, &group_children)?;
+                        }
+                    } else {
+                        for (prop, child) in m.iter() {
+                            let app = App::Properties {
+                                name: prop.clone(),
+                                name_interned: self.tbl.intern(prop)?,
+                            };
+                            self.add_application(app, child)?;
+                        }
                     }
                 }
                 _ => return Err(ExpectedObject),
@@ -611,5 +629,83 @@ impl AnnotationBuilder for CoreAnnotation {
             keywords::WRITE_ONLY => CoreAnnotation::WriteOnly(extract_bool(v)?),
             _ => panic!("unexpected keyword: '{}'", kw),
         })
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use std::iter::FromIterator;
+
+    use super::{super::build::build_schema, super::CoreAnnotation};
+    use crate::schema::{
+        Application::Inline, Application::PatternProperties, Application::Properties, Keyword,
+    };
+    use serde_json::{json, Map};
+
+    #[test]
+    fn test_properties_64() {
+        let sixty_four_properties =
+            (0..64).map(|i: u8| (i.to_string(), serde_json::Value::Object(Map::new())));
+
+        let schema = json!({
+            "type": "object",
+            "properties": Map::from_iter(sixty_four_properties)
+        });
+
+        let curi = url::Url::parse("http://example/schema").unwrap();
+        let result = build_schema::<CoreAnnotation>(curi, &schema);
+
+        assert!(result.is_ok());
+        let kw = result.unwrap().kw;
+        assert!(matches!(kw[0], Keyword::Application(Properties { .. }, _)))
+    }
+
+    #[test]
+    fn test_properties_more_than_64() {
+        let many_properties =
+            (0..200).map(|i: u8| (i.to_string(), serde_json::Value::Object(Map::new())));
+
+        let schema = json!({
+            "type": "object",
+            "properties": Map::from_iter(many_properties)
+        });
+
+        let curi = url::Url::parse("http://example/schema").unwrap();
+        let result = build_schema::<CoreAnnotation>(curi, &schema);
+
+        assert!(result.is_ok());
+        let kw = result.unwrap().kw;
+        assert!(matches!(&kw[0], Keyword::Application(Inline, _)));
+        assert!(matches!(&kw[1], Keyword::Application(Inline, _)));
+        assert!(matches!(&kw[2], Keyword::Application(Inline, _)));
+        assert!(matches!(&kw[3], Keyword::Application(Inline, _)));
+    }
+
+    #[test]
+    fn test_properties_more_than_64_keep_ordering() {
+        let many_properties =
+            (0..200).map(|i: u8| (i.to_string(), serde_json::Value::Object(Map::new())));
+
+        let schema = json!({
+            "type": "object",
+            "properties": Map::from_iter(many_properties),
+            "patternProperties": {
+                "test+": {"const": "a"}
+            }
+        });
+
+        let curi = url::Url::parse("http://example/schema").unwrap();
+        let result = build_schema::<CoreAnnotation>(curi, &schema);
+
+        assert!(result.is_ok());
+        let kw = result.unwrap().kw;
+        assert!(matches!(&kw[0], Keyword::Application(Inline, _)));
+        assert!(matches!(&kw[1], Keyword::Application(Inline, _)));
+        assert!(matches!(&kw[2], Keyword::Application(Inline, _)));
+        assert!(matches!(&kw[3], Keyword::Application(Inline, _)));
+        assert!(matches!(
+            &kw[4],
+            Keyword::Application(PatternProperties { .. }, _)
+        ));
     }
 }
--- a/crates/json/src/schema/intern.rs
+++ b/crates/json/src/schema/intern.rs
@@ -75,4 +75,4 @@ mod test {
     }
 }
 
-const MAX_TABLE_SIZE: usize = std::mem::size_of::<Set>() * 8;
+pub const MAX_TABLE_SIZE: usize = std::mem::size_of::<Set>() * 8;
--- a/crates/doc/src/inference.rs
+++ b/crates/doc/src/inference.rs
@@ -592,6 +592,18 @@ impl Shape {
         let mut unevaluated_properties: Option<Shape> = None;
         let mut unevaluated_items: Option<Shape> = None;
 
+        // Flatten Inline Applications into Applications
+        let inlined_kws = schema
+            .kw
+            .iter()
+            .map(|item| match item {
+                Keyword::Application(Application::Inline, inner_schema) => {
+                    inner_schema.kw.iter().collect()
+                }
+                _ => vec![item],
+            })
+            .concat();
+
         // Does this schema have any keywords which directly affect its validation
         // or annotation result? `$defs` and `definition` are non-operative keywords
         // and have no effect. We would also give a pass to `$id`for the same reason,
@@ -603,7 +615,7 @@ impl Shape {
         // though, Provenance is used to guide generation into static types
         // (whether to nest/inline a definition, or reference an external definition),
         // and excluding these keywords works better for this intended use.
-        if !schema.kw.iter().all(|kw| {
+        if !inlined_kws.iter().all(|kw| {
             matches!(
                 kw,
                 Keyword::Application(Application::Ref(_), _)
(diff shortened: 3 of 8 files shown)


mdibaiee:
This solution actually leads to this error during validation phase:

```
acmeCo/source-hubspot.flow.yaml error at /materializations/acmeCo~1postgres~1materialize-source-hubspot/bindings/25 :
driver error while validating materialization acmeCo/postgres/materialize-source-hubspot

Caused by:
    driver sent constraint for unknown field properties/num_notes
```

Which I suppose is because our view of the schema is actually an `allOf` so the path for that property is:
`allOf/0/properties/num_notes`

mdibaiee:
I pushed another solution using `bit_set` instead of our own `u64` which allows for a growing bitset (backed by [bit-vec](https://docs.rs/bit-vec/latest/bit_vec/)), but if we set a default size of 64, we will not need resizing until something bigger is found. We could also allocate larger (or smaller) ones if we think it's justified. Let me know what you think.

I know performance has been the major concern here, is there a way for us to measure the difference this change will make?

crates/json/src/schema/build.rs
```diff
@@ -613,3 +616,58 @@ impl AnnotationBuilder for CoreAnnotation {
...
+        let result = build_schema::<CoreAnnotation>(curi, &schema);
+
+        assert!(result.is_ok());
+        let kw = result.unwrap().kw;
+        assert!(matches!(kw[0], Keyword::Application(Properties { .. }, _)))
+    }
+
+    #[test]
+    fn test_properties_more_than_64() {
+        let sixty_four_properties =
```
psFried, line 654: nit: s/four/five/

crates/json/src/schema/build.rs
```diff
@@ -613,3 +616,58 @@ impl AnnotationBuilder for CoreAnnotation {
...
+            (0..65).map(|i: u8| (i.to_string(), serde_json::Value::Object(Map::new())));
+
+        let schema = serde_json::Value::Object(Map::from_iter([
```
psFried, line 657: Also just a nit: this would probably read nicer as

```rust
serde_json::json!({
  "type": "object",
  "properties": Map::from_iter(sixty_four_properties),
})
```

crates/json/src/schema/intern.rs
```diff
@@ -49,8 +48,8 @@ impl Table {
     /// is zero-valued.
     pub fn lookup(&self, s: &str) -> Set {
         match self.m.get(s) {
-            Some(&v) => v,
-            None => 0,
+            Some(v) => v.clone(),
```
psFried, line 51: Based on a quick look at the usages of `lookup`, it seems like this could be modified to return a borrowed `BitSlice` instead of cloning.

crates/json/src/schema/intern.rs
```diff
@@ -30,13 +31,11 @@ impl Table {
     /// An Error is returned only if the str overflows Set's capacity to
     /// represent all interned strings (determined by the number of bits
     /// of the Set type).
-    pub fn intern(&mut self, s: &str) -> Result<Set, Error> {
+    pub fn intern(&mut self, s: &str) -> Result<&Set, Error> {
```
psFried, line 34: This seems like it would no longer need to return a `Result`.

psFried (commented):
I think the overall approach here seems reasonable. I think it's worth doing a before/after comparison of the couple of benchmarks that we have, just to be safe. The [critcmp crate](https://github.com/BurntSushi/critcmp) might help with that.

mdibaiee:
The benchmarks say that bitset is 47 and 57% slower on two of our smaller benchmarks (which do not exceed 64 bits either):

```
Gnuplot not found, using plotters backend
rides1x                 time:   [23.387 ms 23.475 ms 23.562 ms]                    
                        change: [+55.133% +57.071% +58.949%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 2 outliers among 100 measurements (2.00%)
  1 (1.00%) low mild
  1 (1.00%) high mild

Benchmarking rides4x: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 11.0s, or reduce sample count to 40.
rides4x                 time:   [109.10 ms 109.30 ms 109.55 ms]                    
                        change: [+46.930% +47.815% +48.523%] (p = 0.00 < 0.05)
                        Performance has regressed
```

psFried:
> The benchmarks say that bitset is 47 and 57% slower on two of our smaller benchmarks

I ran them myself, and found roughly a 40% increase in times, although the measurements themselves were significantly lower. The rides1x avg went from 9.6 to 13.8ms. Overall, I feel like those benchmarks were already showing less than stellar performance before this change, so I'm not quite sure how to feel about the degradation. We _could_ say that we _already_ needed to go in and optimize this, and so we should just merge this and revisit validation perf later. Or we could try to improve perf as part of this pr, perhaps by minimizing clones of the bitvec or else by looking into a different approach for interning properties. @jgraettinger WDYT?

jgraettinger:
@mdibaiee and I have had a sidebar on slack on a different approach, where we break up and inline sub-schemas on an as-needed basis. I'd like to explore that as I think it can lead to zero-cost support that also unblocks further optimization work.

mdibaiee:
hibitset is faster than bitset (50% and 33% slower than our u64, compared to 57% and 47% of bitset. Note that it seem bitset gets worse at a higher rate as our data gets bigger whereas hibitset grows slower, I’d imagine bitset is O(n) but hibitset is O(nlogn) or something given its hierarchical structure).

Still, I have identified one piece of code which is consuming most of the time from hibitset side, this one:

```
            let mut set = BitSet::new();
            for item in vec {
                let prop = extract_str(item)?;
                set |= tbl.intern(prop)?;
            }
```

This keeps modifying the memory in a loop and is the slowest part of the code with respect to our bitset, so I’m just trying to find an at least slightly more optimal approach and see if we can make the performance of this acceptable.

----

Side note: while working on the other approach of an Inline variant, I realised this issue also happens in other places, not only properties: `required` can have more than 64 values, etc., and I think if we can in any way fix this in the bitset level, we will save ourselves a lot of maintenance overload in the future.

jgraettinger:
We _may_ be able to achieve comparable performance to today by reaching for a faster bitset.
But the performance today is also not where it ought to be.

I do have conviction that the better long-term path is going to be the inline approach, as it's zero-cost and also unlocks future optimizations which are not possible today (more compact data structures; parallel evaluation). If we reach for `hibitset` as a near-term crutch that's okay, but full disclosure I'll probably switch it out again the next time I'm in here doing performance work.

> I realised this issue also happens in other places, not only properties

Interning is used for `properties`, `required`, `dependentRequired`, and `dependentSchema`.
I don't think we need to support > N interned values for the latter two though, at least until actually proven otherwise, as they're much rarer.

mdibaiee:
Okay, so I got the inline approach working for majority of cases, however I see some weird behaviour in our `intersect` implementation that I'm starting to doubt is correct.

Take the example of these two schemas, what do you expect to get when you intersect them?
```
type: object
additionalProperties:
  const: addl-prop
```

```
type: object
properties:
  prop:
    const: prop
```

I intuitively expect to get:
```
type: object
properties:
  prop:
    const: prop
additionalProperties:
  const: addl-prop
```

But it seems like our intersection logic doesn't handle the additionalProperties correctly, it instead tries to intersect these two sections:
```
const: addl-prop
```
with:
```
const: prop
```

And ends up with an undefined field `prop` that does not have any value whatsoever (it's an empty `Shape` instance). Seems like a bug to me. So I'll be trying to solve this bug (it's a bit hard to troubleshoot and wrap my head around the code). Any help with understanding why this was implemented the way it is would go a long way, because I'm afraid of breaking some other requirement that has led to this implementation.

The piece of code that does this can be found in traced to these places:
1. https://github.com/estuary/flow/blob/master/crates/doc/src/inference.rs#L261
2. https://github.com/estuary/flow/blob/master/crates/doc/src/inference.rs#L418
3. https://github.com/estuary/flow/blob/master/crates/doc/src/inference.rs#L243

mdibaiee:
This is being parked aside for now, but for the sake of picking it up later, here is some context and what I've done so far:

1. There is a new `Keyword::InlineApplication`, which expects an `Application::Inline` inside it (this probably can be made into requiring only `Application::Inline`, but I found it easier during development because it makes the indexing part slightly easier to match on)
2. It works as far as Schema build and indexing is concerned
3. In `doc::inference` however, it seems to cause issues. What I've done there is that during the inference phase, if I find `Keyword::InlineApplication(Application::Inline, schema)`, I will basically flatten it out by extracting `schema.kw` and putting it beside all other keywords of the current iteration of inference.
4. This however, seems to lead to some issues with `locate` and our `error_collection` tests.
5. The comment above about the intersection seems to be what is causing issue here, however I couldn't understand why this is any different from having the `Application::Properties` directly in that inference level.
6. I tried just appending the `properties` from inline's child to the current `schema.object.properties` in `infer_inner`, however that does not correctly account for things such as reduction strategies.

mdibaiee:
Okay, I ended up realising what the issue was and this is now working with no additional cost (still using u64): We have a new Application variant `Inline` which basically just wraps other schemas.

It is semantically _similar_ to `allOf`, but not the same: during indexing it is ignored and does not have a location of its own (i.e. no `#allOf/0` will be added)

mdibaiee:
Oh, by the way, I have not done the same for `required` which also uses the intern table, only for `properties` at the moment. We can come back and do `required` once we hit an actual issue (I haven't yet, I just have seen this break when running the github_events benchmark)

jgraettinger (changes requested):
> Reviewed 6 of 7 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/465)* status: 6 of 10 files reviewed, 8 unresolved discussions (waiting on @mdibaiee and @psFried)

___
*[`crates/doc/src/inference.rs` line 601 at r6](https://reviewable.io/reviews/estuary/flow/465#-N0_K-9RimN2B1hNWrwk-r6-601:-N0lA1zlAA2ZUGtfgvb1:bwetscj) ([raw file](https://github.com/estuary/flow/blob/7a73db256f5f625cf57a902809229427f099c411/crates/doc/src/inference.rs#L601)):*
<details><summary><i>Previously, mdibaiee (Mahdi Dibaiee) wrote…</i></summary><blockquote>

I'm doing this to get a `Vec<&_>` from `Vec<_>`, is there a more straightforward and less weird way of doing that? Perhaps `map(|a| &a)` just to make this less of a magic trick
</blockquote></details>

nit: `all_kws` or `flat_kws`? These are not just inlined KW's.

It might be a bit simpler to use a traditional for item in `schema.kw` loop with push() / extend() of a mut Vec.
___
*[`crates/json/src/validator.rs` line 317 at r8](https://reviewable.io/reviews/estuary/flow/465#-N0lNliBArrdSZNANnji:-N0lNliBArrdSZNANnjj:b-auz99s) ([raw file](https://github.com/estuary/flow/blob/1394478c131a780a308639ac271483629ba44247/crates/json/src/validator.rs#L317)):*
> ```XML
>                     PatternProperties { re } if regex_matches(re, loc.name) => true,
>                     // AdditionalProperties applies if Properties and PatternProperties haven't.
>                     AdditionalProperties if !evaluated => true,
> ```

This is the specific chunk of code I'm concerned about.

`AdditionalProperties` is evaluated _within_ and not across schemas. The expected behavior is that a matched `Application::Properties` will toggle `evaluated` to true, and the `AdditionalProperties` keyword will then be bypassed and not evaluated.

I expect that new inline behavior is currently breaking this. We need to be checking whether inline applications of the schema match `Properties` as well before allowing `AdditionalProperties` to be evaluated. Which is a bit gross. Perhaps we need a separate mechanism to efficiently determine whether a property is directly matched ... :thinking: 
___
*[`crates/json/src/validator.rs` line 854 at r8](https://reviewable.io/reviews/estuary/flow/465#-N0lLr8JEj5MI1GTB4sd:-N0lLr8JEj5MI1GTB4se:b-16xl64) ([raw file](https://github.com/estuary/flow/blob/1394478c131a780a308639ac271483629ba44247/crates/json/src/validator.rs#L854)):*
> ```XML
>             | App::PropertyNames
>             | App::AdditionalItems
>             | App::Inline => RequiredChild,
> ```

Should `Inline` be `RequiredInPlace`? The distinction is that the schema is being evaluated within the context of its parent location, and not as a sub-location.
___
*[`crates/json/src/schema/build.rs` line 262 at r8](https://reviewable.io/reviews/estuary/flow/465#-N0VGt5R_mCZ1PdOojvI-r8-262:-N0lPGV-BH1sCdl5xO1p:bnjudy2) ([raw file](https://github.com/estuary/flow/blob/1394478c131a780a308639ac271483629ba44247/crates/json/src/schema/build.rs#L262)):*
<details><summary><i>Previously, mdibaiee (Mahdi Dibaiee) wrote…</i></summary><blockquote>

Since inlining is supposed to be completely transparent, we can potentially remove this if condition and always do an inlining with chunks of <= MAX_TABLE_SIZE to simplify this logic. WDYT?
</blockquote></details>

Additional validation scopes do have a cost, so the ideal would be that we put as many keywords into the current schema as possible, and then spill additional keywords into `Inline` applications.

So, that suggests a structure of walking `m` in an enumeration of chunks, where chunks after the first become Inline.
___
*[`crates/json/tests/schema-fixtures/2019-09-additional-props-inline.json` line 1 at r8](https://reviewable.io/reviews/estuary/flow/465#-N0lQss48SiYVfCt6hiZ:-N0lQss48SiYVfCt6hi_:bqtvhwt) ([raw file](https://github.com/estuary/flow/blob/1394478c131a780a308639ac271483629ba44247/crates/json/tests/schema-fixtures/2019-09-additional-props-inline.json#L1)):*
> ```JSON
> {
> ```

A couple of things:

First, this `schema-fixtures` directory is here in order to hold the JSON Schema draft 2019-09 meta-schemas, which many of its official tests reference as I recall. It's here only so that those schema fixtures of the official tests build correctly, and isn't the right place to put supplemental tests we're adding to validate behaviors.

You'll also need to hook up an actual test to be run here, as the JSON file by itself doesn't do anything. I'd recommend adding a test alongside `draft2019_tests.sh` which also uses `validator_test_utils`, but tracks its test fixtures in a separate directory (as official tests come directly from a git submodule of the JSON schema repo).

jgraettinger:
Copied over / tweaked from a Slack :thread:

I think we're probably still in trouble w/ `additionalProperties`. But I'm also realizing there's another solution that I think can be a lot simpler & probably faster, which is to allow Properties to not be interned if the intern table overflows, so long as we can intern all properties appearing in `Application::DependentSchema`, `Validation::Required`, and `Validation::DependentRequired`.

Those schema keywords truly require interning and set behavior, because we need to check whether a particular property was observed long after we've processed it. However, `Properties` doesn't actually require this look-back behavior and we're using interning only for performance.


The restriction would then become that there can't be more than N distinct property names appearing in those keywords **that require look-back**, but you can have as many other properties as you want. I think we're _way_ less likely to run into trouble here with in-the-wild schemas.



Tactically, we can achieve this by prioritizing keywords requiring look-back when initially building up the intern table. It would still be a build-time error if the table overflows while processing keywords that require look-back. Only after they're all processed & interned would we process `properties` keywords. They would also attempt to intern, but if the intern table is full that's not an error and we would leave a sentinel intern value in place (like 0xFFFFFFF) that can't possibly match an intern lookup.



Then, in `push_property`, we check if the `Properties` keyword's intern value is this sentinel and, if so, we fall back to string comparison of the property name. Which is easy to do, because we have the property name handy in that function scope.

PR #469, opened 2022-04-29
json: remove the 64-property limit in schemas

jgraettinger:
For the `properties` and `required` keywords, by:

1) Having Application::Properties use direct string comparison rather
than interning. Interning is not fundamentally required for properties
as the keyword has no look-back requirement (where we must check
if the property was seen long after we've processed it). This lets
schemas scale to an arbitrary number of properties.

2) Adding Application::Inline as an in-place application has no schema
keyword. `required` properties of a schema may be pushed down into
Inline applications on an as-needed basis. This is done by deferring
processing of `required` keywords until the intern table size is known,
and then overflowing extra entries into Inline sub-schemas.

This is still correct because `required` is an embarrisingly-parallel
keyword to evaluate. The same technique could also be applied to
dependentRequired and dependentSchemas if needed.

Enable a GitHub benchmark which had been created but was disabled due to
the 64-field limitation. This surfaced some issues in GitHub's own
schema, which are fixed here. Note that the GitHub schema has more
than 64 `required` properties, which validation is now able to handle.

This patch has no impact on performance:

```
rides1x                 time:   [10.827 ms 10.945 ms 11.077 ms]
                        change: [-1.6814% +0.1011% +1.9073%] (p = 0.91 > 0.05)
                        No change in performance detected.
Found 12 outliers among 100 measurements (12.00%)
  4 (4.00%) high mild
  8 (8.00%) high severe

Benchmarking rides4x: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 5.5s, or reduce sample count to 90.
rides4x                 time:   [54.702 ms 54.993 ms 55.370 ms]
                        change: [-1.1690% +0.0567% +1.0967%] (p = 0.93 > 0.05)
                        No change in performance detected.
Found 10 outliers among 100 measurements (10.00%)
  2 (2.00%) high mild
  8 (8.00%) high severe
```

Credit to @mdibaiee, this is derived from patches in #465 

**Workflow steps:**

Start using schemas with lots of `properties` & `required` properties.

None I know of.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/469)

### DIFF
--- a/crates/doc/src/inference.rs
+++ b/crates/doc/src/inference.rs
@@ -811,7 +811,7 @@ impl Shape {
 
                     shape = Shape::intersect(shape, referent);
                 }
-                Keyword::Application(Application::AllOf { .. }, schema) => {
+                Keyword::Application(Application::AllOf { .. } | Application::Inline, schema) => {
                     shape = Shape::intersect(shape, Shape::infer_inner(schema, index, visited));
                 }
                 Keyword::Application(Application::OneOf { .. }, schema) => {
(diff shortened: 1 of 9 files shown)


mdibaiee (approved):
Looks good! Thank you

PR #472, opened 2022-05-03
Add config-encryption crate for endpoint configs

psFried:
**Description:**

Introduces the flow-config-encryption binary, which is an http server
that encrypts endpoint configurations. The intent is to deploy this as a
cloud-run function. The binary is built into the flow docker image for
expediency, and because it's still unclear what a local setup looks like
with the control plane included. So I decided to keep it easy and close
at hand, until we figure out something better.

The service accepts requests that include both a JSON schema and a
`config` object that conforms to the schema. Assuming the config
validates against the provided schema, it will be encrypted using sops
and returned as either JSON or YAML.

**Workflow steps:**

No user-facing changes.

**Notes for reviewers:**

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/472)


crates/config-encryption/src/encrypt/mod.rs
```diff
@@ -0,0 +1,214 @@
...
+        ));
+    }
+    Ok(())
+}
+
+/// We'd like to avoid keeping sensitive values around in memory, but `serde_json::Value` doesn't
+/// have an easy way to zero out the whole document, so this just traverses the object and
+/// overwrites all the String values. There's no safe way to zero numbers, but they seem far less
+/// likely to hold sensitive information anyway.
+fn drop_plain_config(json: serde_json::Value) {
```
saterus, line 194: Not sure if this helps or not, but you might be able to use https://crates.io/crates/secrecy to help here. It supports deserializing into a `Secret` wrapper that can help with accidentally mishandling a sensitive value. I haven't used it, but it's something you might find interesting.

crates/config-encryption/todo.md
```diff
@@ -0,0 +1,9 @@
+
+
+- TS Client
+- OpenAPI spec
+- Prometheus metrics
+- CORS headers
+- Tests
+- Build image
+- Deploy
```
saterus, line 9: Did you mean to leave this file checked in?

saterus (approved):
This looks great. I tested it out locally and seems to work great. :tada:

crates/config-encryption/src/encrypt/mod.rs
psFried: Yeah I had hoped to be able to use something like that, but the reason it doesn't work is that it only handles owned contiguous memory regions. So it won't actually zero the memory of all the nested values when it's dropped :( I could see it still being useful just as a way of communicating to developers that it's sensitive, though. Thanks :+1:

crates/config-encryption/todo.md
psFried: lol _totally_ ... I'll remove. Thanks for catching that.

PR #474, opened 2022-05-09
connector-proxy: fix logging issues

mdibaiee:
**Description:**

- Move Try `?` calls outside of the tracing macro: the macro's expression was not being evaluated unless the log level was greater than or equal to info.
- Convert logrus log level to flow log level, because Logrus was outputting `warning` instead of `warn` for the log level.
- To avoid this issue ^ from happening again, I made the LogLevel an enum so that we get errors if a wrong value is being passed to connector proxy or any of our Rust components.

**Workflow steps:**

- Pass any log level you want (specifically `warn` and `error` that had issues) and see them outputting the error as expected.

N/A

**Notes for reviewers:**

N/A


jgraettinger (approved):
LGTM

PR #475, opened 2022-05-09
connector-proxy: make sure final checkpoint is sent only if exit successfully

mdibaiee:
**Description:**

- I moved the logic for checking whether we have sent a final checkpoint before concluding our response to Pull, and if not, sending an empty Driver Checkpoint, to a part of the code which allows us to check the exit status before doing so.
- We now check the exit status of the process and only if it is successful, we send an empty DriverCheckpoint.

I have yet to test this, but feel free to take a look at the approach.

**Workflow steps:**

**Notes for reviewers:**

I tried _simpler_ and more readable ways of doing what I did here using interior mutablity patterns, but I was not able to satisfy the borrow-checker of compiler, because we use `Box::pin` for all streams, and it seems to assume a `'static` lifetime for values borrowed in closures passed to the stream. If you can think of a better way please let me know.


crates/connector_proxy/src/connector_runner.rs
```diff
@@ -75,14 +80,68 @@ pub async fn run_airbyte_source_connector(
...
+        Box::pin(stream::try_unfold(
+            (false, res_stream, tp_sender),
+            |(transaction_pending, mut stream, sender)| async move {
+                let (message, raw) = match stream.next().await {
+                    Some(bytes) => {
+                        let bytes = bytes?;
+                        let mut buf = &bytes[..];
+                        let msg = decode_message::<PullResponse, _>(&mut buf)
+                            .await
+                            .unwrap()
```
psFried, line 99: This could do with a comment saying why the decoding is infallible, assuming that's the case.

crates/connector_proxy/src/connector_runner.rs
```diff
@@ -75,14 +80,68 @@ pub async fn run_airbyte_source_connector(
...
+    if exit_status.is_ok() && *op == FlowCaptureOperation::Pull {
+        // the received value (transaction_pending) is true if the connector writes output messages and exits _without_ writing
+        // a final state checkpoint.
+        if tp_receiver.await.unwrap() {
+            // We generate a synthetic commit now, and the empty checkpoint means the assumed behavior
+            // of the next invocation will be "full refresh".
+            tracing::warn!("connector exited without writing a final state checkpoint, flushing the driver checkpoint.");
+            let mut resp = PullResponse::default();
+            resp.checkpoint = Some(DriverCheckpoint {
+                driver_checkpoint_json: Vec::new(),
```
psFried, line 135: I'm thinking that it's worth changing this to return an empty json object for the checkpoint json, and to set it as a merge patch. That will ensure that a prior checkpoint doesn't get overwritten in this case.

Also, I think this block should assert that the airbyte `SyncMode` is set to `full_refresh`. If it's `incremental`, then it seems incorrect to generate a synthetic checkpoint.

tests/run-end-to-end.sh
```diff
@@ -134,6 +138,24 @@ sqlite3 ${OUTPUT_DB} >> ${ACTUAL} <<EOF
 EOF
 echo 'greetings from psql:' >> ${ACTUAL}
 docker-compose --file ${SSH_PSQL_DOCKER_COMPOSE} exec -T -e PGPASSWORD=flow postgres psql -w -U flow -d flow -c 'SELECT message, count FROM greetings ORDER BY count;' --csv -P pager=off >> ${ACTUAL}
+# We _can_ make a precise assertion on the number of documents output from the hello-world-no-state capture
+# because it's configured to output a specific number of documents. So this value should match the
+# `greetings` config in that capture.
+echo 'flow_stats (greetings-no-state docsTotal):' >> ${ACTUAL}
```
psFried, line 144: Nice test! :+1:

psFried (changes requested):
I think this is great, but doesn't quite go far enough. Left a few suggestions inline, but LMK if you want to talk about any of those.

> I can also add that to the source-hello-world if that's okay, what do you think?

IMO it's OK to rely on manual testing for this case for a while. I think we'll definitely want to have a testing connector that can be used to exercise cases like this, but it's probably something we can tackle later on.

crates/connector_proxy/src/connector_runner.rs
mdibaiee: @psFried Do we have any way of being able to do merge patch in a way that works with any value? i.e. if the checkpoint from the connector is an array or a scalar, merging with `{}` will still clobber it.

mdibaiee: @psFried My understanding of the Airbyte protocol is that the state file is generally optional, for all sync modes. I know it seems non-sensical to have an incremental connector that doesn't have a state. Let's imagine the case where an incremental connector does exit successfully without an error, and without having sent out a State message. If we do not append an empty checkpoint, the connector and the deployment will just hang forever, right? Do we want that? 🤔

mdibaiee:
@psFried Okay, I actually realised today that this is more complicated:

So, I'm working on making Facebook connector process multiple AdAccounts, and AdAccounts are something that need to be fetched first-hand before other streams can be made up. The AdAccount stream itself is just this prefetched array of AdAccounts returned as is.

However, even though all streams of Facebook connector are incremental, AdAccounts is not: it's not dependent on start_date and end_date and it might change every run, and so AdAccount stream is a _full refresh_, while the rest of the streams are incremental. This leads to this situation: AdAccounts are not committed until another stream sends a checkpoint.

With the current logic we have in place, if the connector fails before another stream gets a chance to send a checkpoint, AdAccounts are also not committed. AdAccount's commit depends on another stream committing because AdAccount itself does not send a state/checkpoint. Our implementation at the moment is "connector-level", whereas it should be "stream-level", but then, it's hard to know if a "stream" has been successful if there is an error from the connector at some point. It's twisted.

I'm not sure if there is a way for us to be cover all of these cases, it's at least not trivial to handle such a case where a stream is successful without sending a checkpoint, but the connector fails on another stream before ever sending any state.

crates/connector_proxy/src/connector_runner.rs
psFried: In the case that a connector claims to support incremental, we _could_ treat it as an error if it exits without a final state message. That seems pretty reasonable to me. Afterall, how is it supposed to resume correctly without it? That said, I'm not sure whether or not we can actually expect all the airbyte connectors to emit final state messages, so it's fine to have the proxy emit that final state automatically in that case. We should probably log it as a warning, though, just so we can know if it's happening.

I don't think we need to worry about the top-level state being anything other than an object, though. The [airbyte specification](https://github.com/airbytehq/airbyte/blob/master/airbyte-protocol/models/src/main/resources/airbyte_protocol/airbyte_protocol.yaml#L71) requires states to be objects. So I think a merge patch of `{}` would work fine.

psFried:
> Our implementation at the moment is "connector-level", whereas it should be "stream-level", but then, it's hard to know if a "stream" has been successful if there is an error from the connector at some point. It's twisted.

Yeah, Flow considers a state checkpoint to apply to the entire capture, rather than applying to just a specific stream within it. That's because all the documents from all streams are committed in a single gazette transaction. I don't see per-stream transactions being very easy to implement, either.

Would it be possible to have the connector emit a state checkpoint after the AdAccounts? If we re-emit all the AdAccounts on every invocation, then I don't see any harm in committing them as soon as they're written. The state checkpoint could just be `{}`.

mdibaiee:
@psFried that's reasonable, and I tried to do that, but it may come off as weird that given how the Airbyte connectors are written (you give a generator of records to the CDK, and the CDK decides when to send a state), it's actually not trivial to send an empty state after AdAccounts. I'm diggin to see how I can do that though, so we'll see.

crates/connector_proxy/src/connector_runner.rs
mdibaiee: @psFried Okay, we already do emit a warning, and I've already changed the code to do a merge patch instead of overriding the checkpoint.

PR #477, opened 2022-05-12
Phil/schema updates

psFried:
**Description:**

The main change here is the addition of the `advanced` JSON schema annotation. This is intended to mark schema locations that should be hidden or collapsed by default. The `advanced` annotation is then applied to both the network tunneling and the parser configuration schemas.

This also rolls up a few fixes to the network tunnel schema.

**Workflow steps:**

No user-facing usage here. For developers, if you add `"advanced": true` to your json schema, then the section will be collapsed by default in the UI.

**Notes for reviewers:** n/a

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/477)


saterus (approved):
This looks good to me. This should work nicely for cleaning up the UI. :+1:

PR #489, opened 2022-05-23
capture: discover `recommended_name` must be a valid collection name

### LINKED ISSUE #488: sanitize recommended capture names
A discover response that includes an invalid recommended name causes the control-plane agent to exit with an error.
Instead, it should produce an error of the discover operation.

Further, we should sanitize airbyte stream names to ensure they're valid collection suffixes.

jgraettinger:
Validate that is is, and error out on validation failure.

Also update the connector proxy to sanitize Airbyte streams to a valid
recommended name.

Fixes #488

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/489)


mdibaiee (approved):
LGTM!

jgraettinger:
Tested using a local control-plane and UI, and verified it works.

PR #490, opened 2022-05-23
Parser: JSONSchema Generator

### LINKED ISSUE #494: Skeleton of the library and generating schema for scalar values (int, string, null, boolean)
Basic structure for the JSON Schema generator library. Only supports initial scalar type, and returns a `schemars::RootSchema` that can be serialized into json.

pothibo:
Implements the skeleton of what could be used to generate JSONSchema
from a JSON document. `schema_generator::generate(Metadata, &JSONValue)`
generates a JSON Schema from the JSONValue document provided. It is
expected that the document is a JSON object and will return an error if
the JSON value received is not an object (it will error on an Array
too).

All validations are generated inline currently and only support scalar
values. The tests here are built around the current implementation, they
will change as more features are added.

Also, the library itself is not being used anywhere and is inaccessible.
Once the library is going to settle, it will be added as a command in
`flow-parser` and will be usable from the CLI.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/490)

Closes #494

### DIFF
--- a/crates/parser/src/schema_generator/mod.rs
+++ b/crates/parser/src/schema_generator/mod.rs
@@ -1,0 +1,209 @@
+use schemars::{gen::SchemaGenerator, schema::*};
+use serde_json::Value as JSONValue;
+
+#[derive(Debug, Default)]
+pub struct JSONSchema {
+    root: RootSchema,
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum SchemaParseError {
+    #[error("failed to parse the value: {0}")]
+    InvalidValueType(serde_json::Value),
+}
+
+// Generate a JSONSchema from a JSON document.
+// Metadata is used to build the metadata of the root schema
+// so that things like the `title`, `description` and `$id` can be passed to
+// the JSONSchema. The current implementation only supports scalar values. If
+// non-scalar values are provided as a JSON value, an error will be returned.
+//
+// The current implementation is also inlining all object validations that are generated
+// from the JSON. As more complex validations will be generated, it's possible that some
+// validations will be moved into definitions.
+pub fn generate(
+    metadata: Metadata,
+    json_value: &JSONValue,
+) -> Result<JSONSchema, SchemaParseError> {
+    let mut schema = JSONSchema::default();
+    schema.root.schema = SchemaObject::default();
+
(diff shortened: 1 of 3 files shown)


crates/parser/src/schema_generator/mod.rs
```diff
@@ -0,0 +1,208 @@
...
+    // I think it's probably overkill, and it's possible hardcoding
+    // a static' string instead would be better.
+    let sg = SchemaGenerator::default();
+    schema.root.meta_schema = sg.settings().meta_schema.clone();
+
+    let data = if let JSONValue::Object(data) = json_value {
+        data
+    } else {
+        return Err(SchemaParseError::InvalidValueType(json_value.to_owned()));
+    };
```
pothibo, line 44: Not sure about this. Maybe there's a better way to do this.

mdibaiee:
Sorry, I'm probably missing some context around this, but can we not just use [`schema_for_value!`](https://docs.rs/schemars/latest/schemars/macro.schema_for_value.html) macro from schemars to get a JSONSchema for a json document?

I know that schemars does not support merging schemas together, but it can infer JSONSchema of one document at least.

psFried (commented):
My early impressions are positive. I left minor some notes on idiomatic Rust. But I think the biggest open question is really just what we discussed on VC: how would this need to change in order to accommodate generating a schema from _multiple_ input documents. We had some good conversation around that this morning. Just to document the outcome of that, the current idea is to try re-framing that problem in terms of combining two JSON schemas in such a way that any document that validates against _either_ schema will also validate against the combined result schema. That seems like a useful framing because it seems to lend itself well to the issue of evolving collection schemas over time. If a collection starts with schema A, and then a Capture emits a document that doesn't validate against A, then we can derive a schema based on only the new, invalid, document(s), and then combine that result with schema A to produce a new schema for the collection. A potential challenge with that approach is dealing with the complexity of things like combinators, conditionals, etc. So I think a reasonable next step would be to validate that approach, and come up with an alternative if it seems like it won't work out well.

But I think overall this is off to a good start, and is moving in the right direction :+1:

pothibo:
After discussing on Slack, I'm going to merge this PR as it is inaccessible, and it's an OK implementation as it stands, it will change over the course of upcoming PRs and only when the code is complete will it be accessible via CLI.

PR #495, opened 2022-05-25
Add reserved words to Firebolt connector docs

oliviamiannone:
**Description:**

Adds a table from list in #480 

**Notes for reviewers:*

Basically just want to make sure these are expressly forbidden, vs some other systems' reserved words which are permissible when they're quoted.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/495)


mdibaiee (approved):
Thank you!

PR #496, opened 2022-05-25
sqlDriver: Do not reference fields when generating JSONSchema

pothibo:
jsonSchema has the ability to toggle the ability to reference properties
so they aren't embedded in the properties field.

For fields that are defined in go struct's annotation, these are
automatically inlined and never referenced. However, a custom type can
be created and will be, by default, referenced.

So this flag seems to be application, for the most part, only affecting
custom types.

In practice, by enabling the `DoNotReference` flag, the custom type gets rendered differently. With the flag disabled:

```json
"properties": {
  "credentials_json": {
    "$schema": "http://json-schema.org/draft-04/schema#",
    "$ref": "#/definitions/Credential",
  }
},
"additionalProperties": false,
"type": "object",
"definitions": {
  "Credential": {
    "type": "string",
    "title": "Credentials",
    "description": "Google Cloud Service Account JSON credentials in base64 format.",
    "secret": true
  }
}
```
With the flag enabled:
```json
"properties": {
  "credentials_json": {
    "type": "string",
    "title": "Credentials",
    "description": "Google Cloud Service Account JSON credentials in base64 format.",
    "secret": true
  }
}
```

**Workflow steps:**

**Notes for reviewers:**

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/496)

### DIFF
--- a/go/protocols/materialize/sql/driver.go
+++ b/go/protocols/materialize/sql/driver.go
@@ -57,6 +57,7 @@ func (d *Driver) Spec(ctx context.Context, req *pm.SpecRequest) (*pm.SpecRespons
 
 	// Use reflection to build JSON Schemas from endpoint and resource configuration types.
 	var reflector = jsonschema.Reflector{
+		DoNotReference: true,
 		ExpandedStruct: true,
 	}
 	endpointSchema, err := reflector.Reflect(d.EndpointSpecType).MarshalJSON()


psFried (approved):
I think this looks like a good change. One thing to note is just that this may cause some tests to fail in the connectors repo once the dependency is updated there. Would you mind also putting up a PR there to update the dependency and fix any failing tests?

PR #498, opened 2022-05-25
docs/MySQL: Fix some incorrect information

willdonnelly:
**Description:**

The description under "Bindings" was completely incorrect for the `/namespace` field and contained a copy-paste error regarding `/stream`.

Also while I was skimming the docs I noticed that the `binlog_expire_logs_seconds` example was for 7 days, but our recommendation should be to aim for 30 days (since that's the default value that an unmodified MySQL install uses) -- 7 is just the minimum we're prepared to accept.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/498)


oliviamiannone (approved):
Thanks for catching this!

PR #501, opened 2022-05-26
docs/MySQL: Update endpoint config examples to match new structure

willdonnelly:
**Description:**

In https://github.com/estuary/connectors/pull/207 the MySQL connector configuration was refactored. most of the settings that typically won't need to be changed now live under an `advanced` subsection, and the username/password have been grouped into a `login` subsection.

This commit updates the endpoint configuration docs and example to reflect that. The new properties table is deliberately not in alphabetical order, instead it's been ordered roughly by (my best guess of) the likelihood of each setting needing changes.


oliviamiannone (approved):
Looks good - thanks!

PR #511, opened 2022-06-02
Docs: Materialize Google Sheets

oliviamiannone:
**Description:**

Adds documentation for new google sheets materialization connector. 

**Notes for reviewers:**

This one's pretty straightforward.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/511)


dyaffe (changes requested):
For step 3 in the main file,  it's not good enough to give anyone view access.  Rather, we have to directly give the service account edit access (since it's writing).

Otherwise I think this looks good.

PR #516, opened 2022-06-06
Allow catalogs to define ops collections

psFried:
**Description:**

Changes the generation of ops collections to only generate them when
they're not already defined in the sources. This allows a catalog source
yaml to include definitions of ops collections, and thus permits
migration to having them generated by the control plane agent instead of
flowctl.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/516)


willdonnelly (approved):
LGTM

PR #528, opened 2022-06-09
network-tunnel: switch to using openssh for tunnel

mdibaiee:
**Description:**

- Switches from using `thrussh` to shelling out to OpenSSH binaries (`ssh`) for creating a network tunnel
- This means it is now necessary for connectors to have OpenSSH installed. I'm currently testing with: https://github.com/estuary/connectors/pull/258. Will make this into a base image that we can use in all connectors.
- This changes the config spec, since now the user is part of the `ssh_endpoint` instead of being a separate key.

**Workflow steps:**

- Use a connector with network tunnel configuration, or alternatively run `make end-to-end-test`

**Notes for reviewers:**

### DIFF
--- a/crates/network-tunnel/src/sshforwarding.rs
+++ b/crates/network-tunnel/src/sshforwarding.rs
@@ -1,21 +1,14 @@
+use std::os::unix::prelude::PermissionsExt;
+use std::process::Stdio;
+
 use super::errors::Error;
 use super::networktunnel::NetworkTunnel;
 
 use async_trait::async_trait;
-use base64::DecodeError;
-use futures::pin_mut;
 use schemars::JsonSchema;
-use std::net::SocketAddr;
-use std::sync::Arc;
-use thrussh::{
-    client,
-    client::{Handle, Session},
-};
-use thrussh_keys::{key, openssh};
-use tokio::io::{AsyncReadExt, AsyncWriteExt};
-use tokio::net::tcp::ReadHalf;
-use tokio::net::{TcpListener, TcpStream};
-use url::Url;
+use tokio::io::AsyncReadExt;
+use tokio::process::Child;
+use tokio::process::Command;
 
 use serde::{Deserialize, Serialize};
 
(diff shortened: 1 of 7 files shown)


crates/network-tunnel/src/sshforwarding.rs
```diff
@@ -78,210 +69,89 @@ fn port_schema(title: &str, description: &str) -> schemars::schema::Schema {
...
+                // Port forwarding stanza
+                "-L".to_string(),
+                format!("{local_port}:{forward_host}:{forward_port}"),
+                self.config.ssh_endpoint.clone(),
+            ])
+            .stderr(Stdio::piped())
+            .spawn()?;
+
+        // Read stderr of SSH until we find "Local forwarding listening", which means
+        // the ports are open and we are ready to serve requests
```
mdibaiee, line 121: Let me know if there is a better way of doing this... I'll also do a bit of search to see if I can find a better solution

crates/network-tunnel/src/sshforwarding.rs
```diff
@@ -78,210 +72,134 @@ fn port_schema(title: &str, description: &str) -> schemars::schema::Schema {
...
-
-    pub async fn prepare_local_listener(&mut self) -> Result<(), Error> {
-        if self.config.local_port == 0 {
-            return Err(Error::ZeroLocalPort);
+    // We used to have `user` as a field on SSHForwarding config
+    // In order to be backward-compatible, we still allow that field, and if it exists we add
+    // the user ourselves manually
+    fn backward_compatible_forward_host(user: Option<&String>, forward_host: &String) -> String {
+        match user {
+            Some(user) => forward_host.replace("ssh://", &format!("ssh://{user}@")),
```
willdonnelly, line 91: This is not the right field to modify -- `forward_host` is the IP address of the target database, and `ssh_endpoint` is the one we want to add the user field to if present.

willdonnelly:
With this change it looks like I can no longer knock over a `source-postgres` capture by throwing a bunch of load at the target database. Previously a ~10s burst of 10-12kQPS (roughly the maximum the current load-generation setup can produce) would cause `tls: bad record MAC` failures, and now when running the CI-built `flowctl` and `flow-connector-proxy` from this PR (and after tweaking the config to change how the username is specified) the same connector image has no issues.

Now for the bad news: this change also appears to make the whole 'lingering TCP socket to the database so it thinks the replication slot is still in use after a connector restart' problem worse, such that killing and restarting a capture with SSH forwarding reliably gets it into a broken state until something is done to clean that up manually. I think we might need to look into that and figure out how to make shutdown actually solid, or this will end up happening a lot once users get their hands on it.

mdibaiee:
@willdonnelly thanks for testing this! The latter part of your observation is a bit concerning, but not sure if it would happen on production instance. I think we now have enough tests to say that this works™️, we can perhaps merge it, try it out on production and see if we can reproduce the issue you described: does the connector process get terminated properly and do we end up with leftover sockets or not. What do you think?

crates/network-tunnel/src/sshforwarding.rs
mdibaiee: I'm glad you noticed this, that was a late in the day commit 😁

PR #532, opened 2022-06-10
Revert "network-tunnel: switch to using openssh for tunnel"

mdibaiee:
Reverts estuary/flow#528

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/532)


mdibaiee:
will merge those PRs on monday instead

Issue #537, opened 2022-06-10
proposal: Format test and publication logs as json in production


saterus:
We had a discussion a few days ago about the usability of the current logs we present to users. There are two complementary tactics we discussed.

1. Reduce the logging to include only items relevant to users.
2. Present the logs in a way that is easier to consume for both new and power users.

This is a stab at helping with 2. We already have the capability of switching our logging to json within the `flowctl-go` client. This would allow us to more easily consume and format logs in a way that is approachable to new users. This mostly impacts build and activation logs, but also fits in easily with the desire to show logs from the `ops/` logs collections, as those are also formatted as json.

#### Example

Here are a few screenshots from an example of deploying a capture. I'm showing screenshots of my console, rather than the UI as the space within the UI is rather limited.

####### `flowctl-go`'s `--log.format=color` (today's default)

![deploy-default-formatting](https://user-images.githubusercontent.com/119613/173115299-e2c54a64-4699-44ea-8a55-fa058e16b0ce.png)

This is generally usable in the CLI, until a log message embeds json directly. Then the amount of escaping really gets out of hand. 

---

####### `flowctl-go`'s `--log.format=json` piped to `jq --compact-output`

![deploy-json-formatting](https://user-images.githubusercontent.com/119613/173115313-054e392e-088b-469f-952d-1537ae3c77de.png)

This is an improvement, as the Go structs and embedded json are uniformly formatted without extra escaping. It's still rather tricky to read as a human, but now we can work with it programmatically.

---

####### `flowctl-go`'s `--log.format=json` then parsed

![deploy-parsed-json-collapsed](https://user-images.githubusercontent.com/119613/173115360-298aaf0c-7291-47e9-973d-301d44e1d779.png)

I found a json log formatting tool called `pino` to use as an example. It takes the json output and formats it neatly for human consumption. I'm not suggesting we use it, but rather that this is analogous to the type of output we could expect to support in the UI.

*Importantly*, while it looks superficially like the output of `--log.format=color`, the embedded json is not repeatedly escaped.

---

####### `flowctl-go`'s `--log.format=json` then parsed with expanded metadata

![deploy-parsed-json](https://user-images.githubusercontent.com/119613/173115330-331fd908-6567-4955-af2b-b12ad71f6f8b.png)

This final example is showing an expanded view where the metadata is itself pretty-printed. We could allow users to click into individual lines and expand their metadata like this. You can also look at practically any other web-based log platform as an example.

---

#### Summary

[...]

PR #538, opened 2022-06-10
Use logger instead of `fmt.Printf`

saterus:
**Description:**

When testing out some [json log formatting scenarios](https://github.com/estuary/flow/issues/537), I noticed there were quite a few things that were being directly output to the console, rather than going through the logger. This doesn't play nicely with parsing the logs as jsonl.

The fix was simply replace `fmt.Printf` with appropriate log statements.

**Workflow steps:**

N/A

N/A

**Notes for reviewers:**

I'm curious what we should be doing with the output from the node processes. This still bypasses the logger.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/538)

### DIFF
--- a/go/flow/js_worker.go
+++ b/go/flow/js_worker.go
@@ -38,6 +38,7 @@ func NewJSWorker(packageTgz []byte) (*JSWorker, error) {
 	// Bootstrap a Node package with the installed pack.
 	var cmd = exec.Command("npm", "install", packagePath)
 	cmd.Dir = tempdir
+	// TODO: Should we be directly piping npm outputs to the OS? Should these go to ops/ logs and/or logrus logs instead?
 	cmd.Stdout = os.Stdout
 	cmd.Stderr = os.Stderr
 	cmd.SysProcAttr = &syscall.SysProcAttr{Pdeathsig: syscall.SIGTERM}
@@ -93,6 +94,7 @@ func StartCmdAndReadReady(dir, socketPath string, setpgid bool, args ...string) 
 
 	cmd.Dir = dir
 	cmd.Env = append(os.Environ(), "SOCKET_PATH="+socketPath)
+	// TODO: Should we be directly piping npm outputs to the OS? Should these go to ops/ logs and/or logrus logs instead?
 	cmd.Stdout = os.Stdout
 	cmd.Stderr = &readyWriter{delegate: os.Stderr, ch: readyCh}
 
--- a/go/flowctl-go/cmd-api-activate.go
+++ b/go/flowctl-go/cmd-api-activate.go
@@ -100,11 +100,7 @@ func (cmd apiActivate) execute(ctx context.Context) error {
 			return fmt.Errorf("applying capture %q: %w", spec.Capture, err)
 		}
 
-		if response.ActionDescription != "" {
-			fmt.Println("Applying capture ", spec.Capture, ":")
-			fmt.Println(response.ActionDescription)
-		}
-		log.WithFields(log.Fields{"name": spec.Capture}).
+		log.WithFields(log.Fields{"name": spec.Capture, "actionDescription": response.ActionDescription}).
 			Info("applied capture to endpoint")
 	}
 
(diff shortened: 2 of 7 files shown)


jgraettinger (changes requested):
The statements that are updated here are not structured logs, but direct and user-facing program output.

Let's first talk through "should the output of these programs be direct and user-facing?" before considering this change. Right now the checked-in answer is "yes", and IMO structuring these logs is only appropriate if these programs are _not_ trying to be user-facing.

> Reviewed 3 of 7 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/538#-:-N4Y90II9cTbsp6bTdaD:b-ff6wpg)* status: 3 of 7 files reviewed, 5 unresolved discussions (waiting on @jgraettinger and @saterus)

___
*[`go/flow/js_worker.go` line 97 at r1](https://reviewable.io/reviews/estuary/flow/538#-N4Y8GdT6ppxEwcd07JH:-N4Y8GdT6ppxEwcd07JI:bwbxxlz) ([raw file](https://github.com/estuary/flow/blob/e317c3079fd5b0d6322babca5dea46307073294c/go/flow/js_worker.go#L97)):*
> ```Go
> 	cmd.Dir = dir
> 	cmd.Env = append(os.Environ(), "SOCKET_PATH="+socketPath)
> 	// TODO: Should we be directly piping npm outputs to the OS? Should these go to ops/ logs and/or logrus logs instead?
> ```

They'd ideally go into ops/ logs for the derivation.
___
*[`go/flowctl-go/cmd-api-activate.go` line 103 at r1](https://reviewable.io/reviews/estuary/flow/538#-N4Y8W1nD7Hw35cx6woF:-N4Y8W1nD7Hw35cx6woG:bdglhzk) ([raw file](https://github.com/estuary/flow/blob/e317c3079fd5b0d6322babca5dea46307073294c/go/flowctl-go/cmd-api-activate.go#L103)):*
> ```Go
> 		}
> 
> 		log.WithFields(log.Fields{"name": spec.Capture, "actionDescription": response.ActionDescription}).
> ```

This is direct user-facing program output that's shown to the user, so I don't think we want to wrap with JSON here.
___
*[`go/flowctl-go/cmd-api-build.go` line 86 at r1](https://reviewable.io/reviews/estuary/flow/538#-N4Y8nBkC5lTYBBh7cZX:-N4Y8nBkC5lTYBBh7cZY:b-9gab35) ([raw file](https://github.com/estuary/flow/blob/e317c3079fd5b0d6322babca5dea46307073294c/go/flowctl-go/cmd-api-build.go#L86)):*
> ```Go
> 	for _, be := range errors {
> 		var path, ptr = scopeToPathAndPtr(args.Directory, be.Scope)
> 		log.WithFields(log.Fields{"path": path, "ptr": ptr}).Error(be.Error)
> ```

Here too, this is user-facing output and should be left as-is.
___
*[`go/flowctl-go/cmd-api-delete.go` line 102 at r1](https://reviewable.io/reviews/estuary/flow/538#-N4Y977CA3Nkg9_heUoj:-N4Y977CA3Nkg9_heUok:b-qzifyu) ([raw file](https://github.com/estuary/flow/blob/e317c3079fd5b0d6322babca5dea46307073294c/go/flowctl-go/cmd-api-delete.go#L102)):*
> ```Go
> 		}
> 
> 		log.WithFields(log.Fields{"name": spec.Capture, "actionDescription": response.ActionDescription}).
> ```

And here: this is user-facing.
___
*[`go/flowctl-go/cmd-api-test.go` line 109 at r1](https://reviewable.io/reviews/estuary/flow/538#-N4Y9B1aDGaVfcAxaspv:-N4Y9B1aDGaVfcAxaspw:bv19sws) ([raw file](https://github.com/estuary/flow/blob/e317c3079fd5b0d6322babca5dea46307073294c/go/flowctl-go/cmd-api-test.go#L109)):*
> ```Go
> 		} else if scope, err := testing.RunTestCase(ctx, graph, driver, testCase); err != nil {
> 			var path, ptr = scopeToPathAndPtr(config.Directory, scope)
> 			log.WithFields(log.Fields{"path": path, "ptr": ptr}).Error("❌ ", err)
> ```

and here too.

jgraettinger:
Just to be clear, I don't mean to say, "we simply shouldn't do this", but rather "I don't understand how this all fits together within the intended user experience"

PR #542, opened 2022-06-12
captures: separate driver checkpoint update from log commit

### LINKED ISSUE #541: capture: driver state resets on connector EOF with no checkpoints

jgraettinger:
Not all recovery log commits necessarily imply an updated driver
checkpoint. Refactor the capture runtime to update driver checkpoints
within ConsumeMessage, rather than StartCommit.

Update the capture protocol to make these semantics clearer and harder
to mis-use.

Manually tested with a capture that exits EOF without writing a
checkpoint.

Fixes #541

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/542)


mdibaiee (approved):
LGTM 👍🏽

PR #550, opened 2022-06-17
Infer JSON Schemas from JSON Documents

### LINKED ISSUE #493: Implement JSONSchema Generator as a parser command
File storage like GCS and S3 don't have a schema defined for the content they are storing and as such, flow needs to infer a schema from the document it's reading. The JSON Schema generator is going to be a library that can be invoked from the CLI and also be called directly in Rust.

The building of this library is going to be broken up in a couple of iterative steps:

- [ ] Skeleton of the library and generating schema for scalar values (int, string, null, boolean)
- [ ] Add support to ingest multiple documents and generate a JSONSchema that support all document parsed
- [ ] Add CLI support to ingest JSON Document from either STDIN or a file
- [ ] Add support for non-scalar values, i.e. object, array
- [ ] Add support for detecting format for a string value, i.e. timestamp, date, number, boolean, etc.

Some of these steps might be either broken up in smaller pieces or change, as not everything is known at this point and it's possible  that it evolves.

Generally speaking, the goal of this library will be able to ingest any JSON document and return a valid JSONSchema that can be used right away by flow in a collection.

### LINKED ISSUE #509: Add support to ingest multiple documents and generate a JSONSchema that support all document parsed

### LINKED ISSUE #510: Add CLI support to ingest JSON Document from either STDIN or a file

saterus:
**Description:**

Adds a `flow-schema-inference` CLI tool. It currently has a single command, `analyze`, which reads a stream of JSON documents from a file/stdin and infers a basic JSON Schema from those documents.

closes #493 
closes #509 
closes #510 

**Workflow steps:**

```console
# From stdin
$ curl https://swapi.dev/api/people/1 | flow-schema-inference analyze
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "type": "object",
  "required": [
    "birth_year",
    "created",
{{SNIP}}

# From a file
$ curl https://swapi.dev/api/people/1 | jq '.starships[]' -r | xargs -I _ curl _ >> ships.json
$ flow-schema-inference analyze ships.json
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "type": "object",
  "required": [
    "MGLT",
    "cargo_capacity",
    "consumables",
    "cost_in_credits",
    "created",
    "crew",
{{SNIP}}
```

Are the `flow-*` binaries user-facing and documented?

**Notes for reviewers:**

There was quite a bit of churn in the changes I made, so I ended up removing all the files from the history and adding them one at a time back to the history. This builds the tool from the bottom up, while most of the tests are at the top. Highly recommend commit-by-commit as a result.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/550)


crates/schema-inference/src/inference.rs
```diff
@@ -0,0 +1,178 @@
...
+fn infer_array(inner: &[JsonValue]) -> Shape {
+    if let Some(shape) = inner
+        .iter()
+        .map(infer_shape)
+        .reduce(|acc, v| shape::merge(acc, v))
+    {
+        Shape {
+            type_: types::ARRAY,
+            array: ArrayShape {
+                tuple: vec![shape],
```
psFried, line 61: I think this use of `tuple` might not be right. The values in the array are all reduced into a single schema, so it seems like it should be set as `additional`, and leave `tuple` empty.

crates/schema-inference/src/schema.rs
```diff
@@ -0,0 +1,118 @@
...
+        }
+        let object = &mut schema_object.object();
+        object.properties = prop_schemas;
+        object.required = required;
+    }
+
+    if shape.type_.overlaps(types::ARRAY) {
+        let mut array_items = Vec::new();
+        for item in shape.array.tuple.iter() {
+            array_items.push(to_schema(item));
```
psFried, line 58: Going along with the previous `tuple` comment, I think it seems best if we don't make any assumptions about the length of arrays when we're inferring schemas. Given an input array of `["foo", 3, true]`, I'm thinking that it would be better to produce a schema like `{"type": "array", "items": {"type": ["string", "integer", "boolean"]}}`, rather than having `items` be an array of schemas.

crates/schema-inference/src/shape.rs
```diff
@@ -0,0 +1,93 @@
...
+    }
+}
+
+fn merge_array_shapes(lhs: ArrayShape, rhs: ArrayShape) -> ArrayShape {
+    // Deeply _merge_ arrays. This is necessary as _union_ would not properly
+    // merge shapes nested inside this array.
+    let tuple = lhs
+        .tuple
+        .into_iter()
+        .zip_longest(rhs.tuple.into_iter())
```
psFried, line 79: Just noting another place where `tuple` would need changed

crates/schema-inference/src/inference.rs
```diff
@@ -0,0 +1,178 @@
...
+
+    Shape {
+        type_,
+        ..Default::default()
+    }
+}
+
+fn infer_string(_value: &str) -> Shape {
+    Shape {
+        type_: types::STRING,
```
psFried, line 40: No action now, but I just wanted to note that I suspect that in the future we may want to try to infer some string formats (i.e. `date-time`) so that those values can have proper column types when we materialize them. I agree with your decision _not_ to worry about that right now, though.

psFried (changes requested):
I think this looks good % the array `items` being output as an array vs a single schema.

Another random thought is that a good test might be to infer a schema for a sequence of json objects, and then assert that each of the objects actually validates against the generated schema. This could potentially even be a good use for property based testing, but IDK how easy that is to wire up with json values. Definitely not a requirement, though. Just tossing ideas out there.

saterus:
@psFried I think this is ready for another look. I added one commit with the changes we talked about that I believe addresses your previous comments. It also has a small tweak to a test that illustrates the output even more clearly.

PR #557, opened 2022-06-21
Docs - add script for materialize-snowflake setup

oliviamiannone:
**Description:**

Title says it all. Per recent testing. 

**Notes for reviewers:**

Going to skip a review I think because this ended up being essentially copy-paste.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/557)


dyaffe (commented):
I might just add a note that says that they can replace the names on the first 6 lines with the intended fields that they want to use.  That might not be necessary since it's probably super obvious.  Looks good though.

PR #558, opened 2022-06-22
Doc updates for SSH, postgres, mysql

oliviamiannone:
**Description:**

Per feedback from @dyaffe, these updates include:

SSH setup guide
- Instruct user to specify port 22 for ssh_endpoint
- Correct misleading notes referring to port 5432 - elaborate that 5432 is Postgres and 3306 is mysql

Source-postgres
- correct azure setup steps
- ensure that user connects to the correct instance before performing queries
- distinguish Postgres v14 queries from previous versions for cloud platforms (see note below)

Source-mysql
- Add setup steps for azure
- Add setup steps for google cloud

**Notes for reviewers:**

Regarding v14 vs previous version queries for postgres:

As a non-sql-expert, I had some trouble parsing out which queries, exactly, in the pre-v-14 versions for Google Cloud and Amazon steps would be replaced by `GRANT pg_read_all_data `. So, that change was not made there. 

I'm under the impression that `pg_read_all_data` is a new, convenient and pre-defined role, so those old-version commands should still work for all versions, anyway (?). Regardless, the queries could probably use a good review.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/558)

### DIFF
--- a/site/docs/guides/connect-network.md
+++ b/site/docs/guides/connect-network.md
@@ -19,12 +19,18 @@ basic configuration options.
 
 2. Referencing the config files and shell output, collect the following information:
 
-   * The **SSH endpoint** for the SSH server, formatted as `ssh://user@hostname[:port]`. This may look like the any of following:
+  * The SSH **user**, which will be used to log into the SSH server, for example, `sshuser`. You may choose to create a new
+  user for this workflow.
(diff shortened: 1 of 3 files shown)


dyaffe (approved):
This looks good to me.

PR #560, opened 2022-06-23
crates/build: don't generate TS package if there are validation failures

jgraettinger:
Generating TypeScript types can fail if there are user schema errors.
Only proceed if validation is successful.

Tested manually.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/560)


psFried (approved):
This looks good to me. Sorry it fell off my radar for review.

Issue #513, opened 2022-06-04
Potential UX improvement for SSH tunnel configuration


willdonnelly:
This is a literal shower thought so I haven't thought it out fully, but it feels like the configuration flow when using SSH tunneling can be significantly improved. The user shouldn't need to pick an arbitrary local port, and they shouldn't need to remember to point the main connector config at `127.0.0.1` and then put the real destination IP/hostname in the SSH forwarding config. The reason this is tricky is because the underlying connector's configuration is treated as a black box, but it seems like it wouldn't be that complicated to make it a bit smarter. A strawman proposal, feel free to nitpick:

- The 'Connector Config Guidelines' should recommend that any connector taking host and port configuration should represent that as a top-level `/address` property with a value like `"host:port"`. This seems desirable in general for consistency across connectors.
- When asked to do actual work (validate/discover/read/etc), the connector proxy examines the connector configuration, and if `/networkTunnel/sshForwarding/{forwardHost,forwardPort,localPort}` are all unset and an `/address` property is present it:
  1. Uses the host:port information from the `/address` field to fill out its `forwardHost` and `forwardPort` information.
  2. Randomly picks a reasonable value for `localPort`.
  3. Replaces `/address` with `"127.0.0.1:<localPort>"` before giving the configuration to the underling connector.
- **[Optional]** When asked for the config spec, the connector proxy could check whether an `/address` property is described in the underlying connector spec. If so, then the 'Forward Host', 'Forward Port', and 'Local Port' properties would be omitted from the SSH Tunnel portion of the config spec. Or alternatively they could just be marked with `advanced: true` I suppose.

I think that the above would suffice, and in my opinion it would make setting up captures/materializations with SSH tunneling much more pleasant.

willdonnelly:
Would it be reasonable to not even offer the SSH Tunnel option for connectors which don't have a `/address` config property? My gut feeling is that connectors either take an arbitrary `host:port` and can use SSH Tunneling, or they're cloud-service connectors for which it wouldn't do anything useful. I could see some cloud service doing source-IP-based access control, but as currently implemented the SSH tunnel wouldn't change that anyway.

willdonnelly:
Also worth noting that the "Host/Port to Address" change to `materialize-postgres` can be done in a backwards-compatible way by adding the `/address` property, removing the host/port properties from the config specification but not the config struct, and a little bit of logic so it'll use whichever mechanism is present. So it should be possible to do this without any breaking changes to existing captures.

[...]

PR #563, opened 2022-06-24
Simplified SSH Forwarding Configuration

willdonnelly:
**Description:**

This is an attempt to implement the mechanism I described in https://github.com/estuary/flow/issues/513, which allows a top-level endpoint config property named `address` to serve as the `forwardHost/forwardPort` values for SSH tunneling (and replaces the `address` property with `127.0.0.1:$RANDOM_LOCAL_PORT` before invoking the connector).

It includes some additional logic which does a similar job when asked for the endpoint config spec, and omits the `forwardHost`, `forwardPort`, and `localPort` properties from the SSH config portion of the resulting schema IFF the main endpoint config schema contains a top-level `address` property.

It should be noted that the new logic is backwards-compatible in two ways:

- First, any existing captures/materializations which explicitly specify `forwardHost` / `forwardPort` / `localPort` should continue to operate as normal. This means that we can choose to migrate existing uses of `source-postgres` and other connectors whenever it's convenient.
- Second, any connectors which do not have a top-level `address` property will continue to provide the "full" version of the SSH config spec, which includes `forwardHost` and friends. This means that we can choose to modify additional connectors (such as `materialize-postgres`) to use the specific `address` configuration whenever it's convenient.

This whole PR still feels really hacky because of the reliance on a specific endpoint property `address`, but I believe it should work well in practice and I couldn't justify using a more complicated mechanism at this time. The main issues are:

1. To opt in to this behavior, connectors will need to be modified. For instance `materialize-postgres` will need to replace its current `host` and `port` properties with `address`, which is itself a potentially-breaking change that needs some care (either keeping things backwards-compatible or pinning-then-upgrading the live materializations) to not break existing uses.
2. There could theoretically be some connectors out there with a top-level `address` property which does not have the expected semantics, and we would do the wrong thing with those. However I have checked and I do not believe any of our current officially supported connectors have this issue, and it's a bit unlikely anyway.

**Workflow steps:**

Configuration for some connectors (currently `source-postgres` and `source-mysql`) will no longer require `forwardHost` and related properties to be specified. Instead the primary `address` field should always describe the target database, and the fact that it's not directly reachable but goes through an SSH bastion host is only represented by the presence of `sshEndpoint` and `privateKey` configuration.

Probably https://docs.estuary.dev/guides/connect-network/, and it looks like https://docs.estuary.dev/reference/Connectors/capture-connectors/PostgreSQL/#amazon-rds directly mentions `forwardHost` so that should be changed too.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/563)

### DIFF
--- a/crates/connector_proxy/src/libs/network_tunnel.rs
+++ b/crates/connector_proxy/src/libs/network_tunnel.rs
@@ -34,27 +34,21 @@ impl NetworkTunnel {
     // Start the network tunnel. The receiver rx will be dropped to indicate the network tunnel
     // is ready to accept requests.
     async fn start_network_tunnel(
-        config: NetworkTunnelConfig,
+        mut network_tunnel: Box<dyn network_tunnel::networktunnel::NetworkTunnel>,
         rx: Receiver<()>,
-    ) -> Result<(), Error> {
-        let mut network_tunnel = config.new_tunnel();
-        tokio::task::spawn(async move {
-            let result: Result<(), Error> = match network_tunnel.prepare().await {
-                Ok(()) => {
-                    drop(rx);
-                    network_tunnel.start_serve().await.map_err(Into::into)
-                }
-                Err(e) => Err(e.into()),
-            };
-
-            if let Err(ref err) = result {
-                tracing::error!(error=?err, "failed starting network tunnel.");
-                std::process::exit(1);
+    ) {
+        let result: Result<(), Error> = match network_tunnel.prepare().await {
+            Ok(()) => {
+                drop(rx);
+                network_tunnel.start_serve().await.map_err(Into::into)
             }
-        })
-        .await?;
+            Err(e) => Err(e.into()),
+        };
 
-        Ok(())
+        if let Err(ref err) = result {
+            tracing::error!(error=?err, "failed starting network tunnel.");
+            std::process::exit(1);
+        }
     }
 
     pub async fn consume_network_tunnel_config(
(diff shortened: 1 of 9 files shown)


psFried (approved):
I think this looks good.

One question I have is whether it would also make sense to apply this behavior in the presence of separate `host` and `port` properties. There certainly doesn't seem to be any reason why we _couldn't_, though it would add complexity to a feature that we may not be entirely certain about. WDYT?

crates/connector_proxy/src/libs/network_tunnel.rs
```diff
@@ -2,28 +2,58 @@ use crate::errors::Error;
...
+    }
+}
 
 impl NetworkTunnel {
     pub fn extend_endpoint_schema(
         endpoint_spec_schema: Box<RawValue>,
     ) -> Result<Box<RawValue>, Error> {
-        let network_tunnel_schema = create_root_schema::<NetworkTunnelConfig>();
-
+        let network_tunnel_schema = &mut create_root_schema::<NetworkTunnelConfig>();
```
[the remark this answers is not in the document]
willdonnelly: Done, thanks!

psFried:
Thanks for the explanation @willdonnelly , I'm convinced 👍

crates/network-tunnel/src/sshforwarding.rs
```diff
@@ -80,6 +87,13 @@ pub struct SshForwarding {
     process: Option<Child>,
 }
 
+fn split_host_port(hostport: String) -> Option<(String, u16)> {
+    let mut splits = hostport.as_str().splitn(2, ':');
```
mdibaiee, line 91: Should this be `rsplitn`, since we might have username and password in the host? e.g.:

```
mahdi@host.com:8080
```

willdonnelly: Currently none of the connectors for which this logic will trigger support username/password in the `address` field. I'd be open to adding that, but I think it would require some additional code to chop off anything up to an `@` character and then reattach that to the rewritten localhost addres.

PR #566, opened 2022-06-28
Don't use host network by default

psFried:
**Description:**

Changes the default value of the --network argument to flowctl serve
consumer to be empty so that the host network will not be used by
default. This also removes the default from `temp-data-plane`, so
users who use that for running locally may want to pass `--network=host`
in order to connect to services running on localhost.

Using the host network in production would be bad because tasks that use ssh forwarding could end up using conflicting ports. This PR is primarily to address that issue.

**Workflow steps:**

If you run temp-data-plane locally, you'll probably want to add the `--network host` argument. Otherwise, no change.

No user-facing docs changes, since temp-data-plane isn't really a user-facing subcommand anymore.

**Notes for reviewers:** n/a

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/566)

### DIFF
--- a/go/runtime/flow_consumer.go
+++ b/go/runtime/flow_consumer.go
@@ -29,7 +29,7 @@ type FlowConsumerConfig struct {
 	Flow struct {
 		BuildsRoot string `long:"builds-root" required:"true" env:"BUILDS_ROOT" description:"Base URL for fetching Flow catalog builds"`
 		BrokerRoot string `long:"broker-root" required:"true" env:"BROKER_ROOT" default:"/gazette/cluster" description:"Broker Etcd base prefix"`
-		Network    string `long:"network" default:"host" description:"The Docker network that connector containers are given access to."`
+		Network    string `long:"network" description:"The Docker network that connector containers are given access to, defaults to the bridge network"`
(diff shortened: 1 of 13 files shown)


go/runtime/flow_consumer.go
```diff
@@ -29,7 +29,7 @@ type FlowConsumerConfig struct {
 	Flow struct {
 		BuildsRoot string `long:"builds-root" required:"true" env:"BUILDS_ROOT" description:"Base URL for fetching Flow catalog builds"`
 		BrokerRoot string `long:"broker-root" required:"true" env:"BROKER_ROOT" default:"/gazette/cluster" description:"Broker Etcd base prefix"`
-		Network    string `long:"network" default:"host" description:"The Docker network that connector containers are given access to."`
+		Network    string `long:"network" description:"The Docker network that connector containers are given access to, defaults to none"`
```
jgraettinger, line 32: comment is misleading, as "none" means "doesn't have anything but loopback".

Perhaps reword to "defaults to creating a new docker bridge network"?

jgraettinger (approved):

PR #569, opened 2022-07-01
Docs: new parser config for cloud storage

oliviamiannone:
**Description:**

Changes documentation for source-gcs and source-s3 per 
https://github.com/estuary/connectors/pull/276 and https://github.com/estuary/flow/pull/556

Added info on the parser as a section to the connector docs themselves; didn't seem like quite enough info to warrant a separate page.

Config tables and YAML have also been updated.

**Notes for reviewers:**

I noticed a typo in the spec, in the description for `/csv/lineEnding` it's corrected here, but worth fixing at the source. 
* "withe" should be "with"
* Also, the md syntax for italicizing _either_ doesn't output correctly either in the UI or the docs. Might be best to just remove that.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/569)

### DIFF
--- a/site/docs/reference/Connectors/capture-connectors/amazon-s3.md
+++ b/site/docs/reference/Connectors/capture-connectors/amazon-s3.md
@@ -32,51 +32,6 @@ and informs how the capture will behave in production.
 To capture the entire bucket, omit `prefix` in the endpoint configuration and set `stream` to the name of the bucket.
 :::
 
-### Parsing cloud storage data
-
-Cloud storage platforms like S3 can support a wider variety of file types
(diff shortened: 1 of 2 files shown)


psFried (approved):
I think overall this looks good. I'm a little mixed on having the parser configuration be inline with the rest of the endpoint config, now that I've seen it. The intent with the new parser config is that most users shouldn't need to touch it, since the `Auto` defaults are able to handle a lot of common cases. The UI is presenting the parser config as "advanced" (hidden by default), so maybe it makes sense to try to do the same in the docs? I'll defer to your judgement there, now that I've given my 2 cents.

PR #570, opened 2022-07-05
oauth: add OAuth2Spec to flow protocol

mdibaiee:
**Description:**

- Add OAuth2Spec to flow protocol

**Workflow steps:**

- Capture connectors can now include this field in their `SpecResponse`
- Try `flowctl api spec --image ghcr.io/estuary/source-test:dev` and you should get:
```
{
  "type": "capture",
  "documentationURL": "",
  "endpointSpecSchema": { ... },
  "resourceSpecSchema": { ... },
  "oauth2Spec": {
    "provider": "google",
    "authUrlTemplate": "https://accounts.google.com/o/oauth2/auth?access_type=offline&client_id={{ client_id }}&redirect_uri={{ redirect_uri }}&response_type=code&scope=email&state={{ state }}",
    "accessTokenUrlTemplate": "https://oauth2.googleapis.com/token",
    "accessTokenBody": {
      "grant_type": "authorization_code",
      "client_id": "{{ client_id }}",
      "client_secret": "{{ client_secret }}",
      "redirect_uri": "{{ redirect_uri }}",
      "code": "{{ code }}"
    },
    "accessTokenResponseMap": {
      "refresh_token": "/refresh_token",
      "testing": "/access_token"
    }
  }
}
```

**Notes for reviewers:**

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/570)


go/protocols/flow/flow.proto
```diff
@@ -419,6 +419,85 @@ message MaterializationSpec {
   protocol.JournalSpec recovery_log_template = 6;
 }
 
+// OAuth2Spec describes an OAuth2 provider
+message OAuth2Spec {
+  // Name of the OAuth2 provider
+  string provider = 1;
```
psFried, line 425: nit: Is this just a human-readable display name, or is it used as a machine-readable key at some point in the process? It would be nice to clarify in the comment.

go/protocols/flow/flow.proto
```diff
@@ -419,6 +419,85 @@ message MaterializationSpec {
...
+
+  // The templates below are handlebars templates and have a set of variables
+  // available to them, the variables available everywhere are:
+  // client_id: OAuth2 provider client id
+  // redirect_uri: OAuth2 provider client registered redirect URI
+  //
+  // Variables available in Auth URL request:
+  // state: the state parameter
```
psFried, line 433: Just a question to clarify my understanding: what generates the `state`? It's the UI, right? Are there any guarantees about what's in it, or is it supposed to be treated as an opaque blob at this point?

crates/connector_proxy/src/interceptors/airbyte_source_interceptor.rs
```diff
@@ -61,12 +61,14 @@ impl AirbyteSourceInterceptor {
         Box::pin(stream::once(async {
             let message = get_airbyte_response(in_stream, |m| m.spec.is_some()).await?;
             let spec = message.spec.unwrap();
+            let auth_spec = spec.auth_specification.map(|s| serde_json::from_str(s.get())).transpose()?;
```
psFried, line 64: TIL about `transpose`. Neat! :+1:

psFried (approved):
I think this looks good. Had a few minor comments, but nothing blocking.

go/protocols/flow/flow.proto
mdibaiee: This is a machine-readable key, used in the UI to distinguish which button to show. Will clarify

go/protocols/flow/flow.proto
mdibaiee: The state is generated somewhere along the way, doesn't really matter where, what matters is that it must be random and somehow assocaited with a user's "session". In this case, we generate a random hash, and we give it to the UI to store it in localStorage or some similar kind of session storage, and it is then checked against the response received from the OAuth provider.

Here is an example of how an attack might be done on one of our users if we don't have a session-reliant state:
https://estuary-dev.slack.com/archives/C01G7CFNA8K/p1656696977024669?thread_ts=1656696974.766259&cid=C01G7CFNA8K

>1. I initiate an OAuth login request for Google on Estuary
>2. I am redirected to Google’s login page, e.g.: `https://accounts.google.com/o/oauth2/auth?access_type=offline&client_id=ESTUARY_CLIENT_ID&redirect_uri=dashboard.estuary.dev/oauth&response_type=code&scope=email&state=SOME_STATE`
>3. I do not fill the login form, instead I send this to a friend of mine, and they open it up and login, and they are redirected to [dashboard.estuary.dev/oauth](http://dashboard.estuary.dev/oauth) with an access token.
>4. If there is no proper state verification, then my Google connector will have access to my friend’s account.
>
>But if we do have state verification, my friend’s state verification will fail, because they have a different session than mine, and [dashboard.estuary.dev/oauth](http://dashboard.estuary.dev/oauth) will compare the SOME_STATE with something like `session['state']` and it will fail.

PR #579, opened 2022-07-18
Bubble up errors returned from parser consumer

saterus:
**Description:**

This allows the caller to signal that we should halt parsing. Without this change, the parser will continue to read documents from the source until it reaches the end of data even if the caller has encountered an error (or would for its own reasons prefer to stop).

**Workflow steps:**

N/A

N/A

**Notes for reviewers:**

Used by the latest changes to https://github.com/estuary/connectors/pull/278.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/579)

### DIFF
--- a/go/parser/run.go
+++ b/go/parser/run.go
@@ -113,7 +113,8 @@ func (r *parserStdout) Write(p []byte) (int, error) {
 		p = p[pivot+1:]
 	}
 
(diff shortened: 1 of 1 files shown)


psFried (approved):
LGTM

PR #580, opened 2022-07-19
Docs: updated network tunnel config

oliviamiannone:
**Description:**

Updates documentation to reflect the new network tunnel config per https://github.com/estuary/flow/pull/563

Eliminated reference to older config, even though it's backward-compatible, because it's not expected to be pertinent to the user.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/580)

### DIFF
--- a/site/docs/concepts/connectors.md
+++ b/site/docs/concepts/connectors.md
@@ -316,27 +316,15 @@ materializations:
       connector:
         image: ghcr.io/estuary/materialize-postgres:dev
         config:
-          # When using a proxy like SSH tunneling, set to localhost
-          host: localhost
-          # Specify an open port on your local machine to connect to the proxy.
-          port: 15432
+          address: 127.0.0.1:5432
           database: flow
           user: flow_user
           password: secret
           networkTunnel:
             sshForwarding:
-              # Port on the local machine from which you'll connect to the SSH server.
-              # If a port is specified elsewhere in the connector configuration, it must match.
-              localPort: 15432
-              # Host or IP address of the final endpoint to which you’ll
-              # connect via tunneling from the SSH server
-              forwardHost: 127.0.0.1
-              # Port of the final endpoint to which you’ll connect via
-              # tunneling from the SSH server.
-              forwardPort: 5432
               # Location of the remote SSH server that supports tunneling.
               # Formatted as ssh://user@hostname[:port].
-              sshEndpoint: ssh://sshUser@198.21.98.1
+              sshEndpoint: ssh://sshUser@198.21.98.1:22
               # Private key to connect to the SSH server, formatted as multiline plaintext.
               # Use the YAML literal block style with the indentation indicator.
               # See https://yaml-multiline.info/ for details.
--- a/site/docs/concepts/connectors.md
+++ b/site/docs/concepts/connectors.md
@@ -309,12 +309,12 @@ or add the `networkTunnel` stanza directly to the YAML, as shown below.
 
 #### Sample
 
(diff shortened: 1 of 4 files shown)


site/docs/concepts/connectors.md
```diff
@@ -316,27 +316,15 @@ materializations:
       connector:
         image: ghcr.io/estuary/materialize-postgres:dev
         config:
-          # When using a proxy like SSH tunneling, set to localhost
-          host: localhost
-          # Specify an open port on your local machine to connect to the proxy.
-          port: 15432
+          address: 127.0.0.1:5432
```
willdonnelly, line 319: The `materialize-postgres` connector hasn't actually been updated to use `address` yet. But that's on my todo list for the near future, so I can throw that together tomorrow if it'll help to update the docs all at once.

site/docs/guides/connect-network.md
```diff
@@ -1,8 +1,9 @@
 # Configure connections with SSH tunneling
 
-Depending on your enterprise network security, you may need to use [SHH tunneling](https://www.ssh.com/academy/ssh/tunneling/example#local-forwarding), or port forwarding, to allow Flow
-to securely access your endpoint. This is configured within a capture or materialization definition, but
-before you can do this, you'll need a properly configured SSH server on your internal network or cloud hosting platform.
+Flow connects to certain types of endpoints — generally databases — using their IP address and port.
+To keep this connection secure, you can configure [SHH tunneling](https://www.ssh.com/academy/ssh/tunneling/example#local-forwarding), or port forwarding.
```
willdonnelly, line 4: Replace "**SHH** tunneling" with "**SSH** tunneling". Also I'd recommend saying "also known as port forwarding" rather than just "or" to make it clear that this is just one mechanism and not two.

willdonnelly, line 4: Perhaps "For added security" rather than "To keep this connection secure". In many cases the connection is plenty secure even in the absence of SSH tunneling, and it's just an added layer of defense-in-depth.

site/docs/guides/connect-network.md
```diff
@@ -157,34 +140,17 @@ To allow SSH tunneling to a database instance hosted on Azure, you'll need to cr
...
-* **Local Port** / `localPort`: the port on the localhost used to connect to the SSH server
 
-1. Use these to add SSH tunneling to your capture or materialization definition, either by filling in the corresponding fields
-  in a web app, or by working with the YAML directly. Reference the [Connectors](../../concepts/connectors/#connecting-to-endpoints-on-secure-networks) page for a YAML sample.
-
-2. Proxies like SSH are always run on an open port on your localhost,
-so you'll need to re-configure other fields in your capture or materialization definition.
-
-   1. Set the connector's host property to match `localhost` in the SSH configuration.
+* **Private Key** / `privateKey`: the contents of the PEM file
```
willdonnelly, line 153: I know there's a relationship between SSH private keys and PEM files, but if we're being pedantic I don't think all private key files are necessarily PEM files nor that all PEM files are necessarily valid SSH keyfiles.

Perhaps change this to simply "the contents of the SSH private key file"?

willdonnelly (approved):
LGTM. I've suggested tweaks to the phrasing in a couple of spots

site/docs/concepts/connectors.md
oliviamiannone: My bad - to make it easier for now I just changed the example to use `source-postgres`, since it really doesn't need to be one way or the other!

PR #589, opened 2022-07-21
docs: source-http-file 

oliviamiannone:
**Description:**

Adds docs for the new http file source connector, https://github.com/estuary/connectors/pull/288

**Notes for reviewers:**

I added descriptions manually to the credentials config - I would recommend also adding these formally so they reflect in the UI. Let me know if it'd help for me to put up a ticket!

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/589)

### DIFF
--- a/site/docs/reference/Connectors/capture-connectors/http-file.md
+++ b/site/docs/reference/Connectors/capture-connectors/http-file.md
@@ -30,6 +30,13 @@ If necessary, you can specify the correct file type, compression, and other prop
 ## Prerequisites
 
 To use this connector, you'll need the URL to an HTTP endpoint that hosts data of one of the supported types.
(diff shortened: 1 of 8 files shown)


site/docs/reference/Connectors/capture-connectors/http-file.md
```diff
@@ -0,0 +1,136 @@
...
+* ZIP
+* GZIP
+* ZSTD
+
+By default, Flow automatically detects the file type and compression method.
+If necessary, you can specify the correct file type, compression, and other properties (CSV only) using the optional [parser configuration](#advanced-parsing-http-hosted-data).
+
+## Prerequisites
+
+To use this connector, you'll need the URL to an HTTP endpoint that hosts data of one of the supported types.
```
mdibaiee, line 32: One thing that's important is that the endpoint must support [`HEAD`](https://developer.mozilla.org/en-US/docs/web/http/methods/head) HTTP requests which includes a [`Last-Modified`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Last-Modified) header that indicates the last time the document was updated. We use this information to know when to pull the data again.

mdibaiee (approved):
Thank you Olivia! I added one point about requirements for the server serving the file, good to go after that

PR #591, opened 2022-07-22
Phil/parse config discriminator

psFried:
**Description:**

Changes the parser configuration again, in an effort to make it work better in the UI. With this, it finally works well end to end!

**Workflow steps:**

1. Use the S3 or GCS capture connectors.
2. Be amazed at how the parser config form renders all beautifully in the UI.

- [S3](https://docs.estuary.dev/reference/Connectors/capture-connectors/amazon-s3/)
- [GCS](https://docs.estuary.dev/reference/Connectors/capture-connectors/gcs/)

**Notes for reviewers:**

Recommend going commit by commit. Here's how it looks as part of the source-s3 connector:

image

image

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/591)

### DIFF
--- a/crates/parser/src/config/mod.rs
+++ b/crates/parser/src/config/mod.rs
@@ -342,6 +342,15 @@ impl Format {
             _ => Some(self.clone()),
         }
     }
+
+    fn schema(gen: &mut schemars::gen::SchemaGenerator) -> schemars::schema::Schema {
+        let mut schema = gen.subschema_for::<Format>().into_object();
+        schema.extensions.insert(
+            "discriminator".to_string(),
+            serde_json::json!({"propertyName": "name"}),
+        );
+        schema.into()
+    }
 }
 
 /// This value is always an empty JSON object.
@@ -516,6 +525,7 @@ pub struct ParseConfig {
     /// Determines how to parse the contents. The default, 'Auto', will try to determine the format
     /// automatically based on the file extension or MIME type, if available.
     #[serde(default)]
+    #[schemars(schema_with = "Format::schema")]
     pub format: Format,
 
     /// Determines how to decompress the contents. The default, 'Auto', will try to determine the
--- a/crates/parser/src/config/mod.rs
+++ b/crates/parser/src/config/mod.rs
@@ -278,6 +278,7 @@ impl JsonPointer {
 }
 
 #[derive(Debug, PartialEq, Clone, Serialize, Deserialize, JsonSchema)]
+#[serde(tag = "name", content = "config")]
 pub enum Format {
     /// Attempt to determine the format automatically, based on the file extension or associated
     /// content-type.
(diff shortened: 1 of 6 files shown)


crates/parser/src/config/mod.rs
```diff
@@ -306,41 +307,69 @@ pub enum Format {
...
+
+    /// Customizes the generated schema for Format to make it work better in the UI.
+    /// Format still derives `JsonSchema`, and this function delegates to that impl.
+    /// This function is used by setting `schema_with` on the property in ParseConfig.
+    fn schema(gen: &mut schemars::gen::SchemaGenerator) -> schemars::schema::Schema {
+        let mut schema = gen.subschema_for::<Format>().into_object();
+        // The UI doesn't handle the case where a schema specifies both `type` and `const`,
+        // and technically it's redundant to have both. So we remove the `type` property from the
+        // `type` of each variant.
+        for subschema in schema.subschemas().one_of.as_mut().unwrap().iter_mut() {
```
mdibaiee, line 354: Is this `unwrap()` safe?

crates/parser/src/config/mod.rs
```diff
@@ -566,12 +596,40 @@ pub enum ConfigError {
...
+/// by the estuary UI in order to render properly.
+#[derive(Debug, Clone)]
+struct ConstDefaultSchemaVisitor;
+impl schemars::visit::Visitor for ConstDefaultSchemaVisitor {
+    fn visit_schema_object(&mut self, schema: &mut schemagen::SchemaObject) {
+        schemars::visit::visit_schema_object(self, schema);
+
+        if schema.enum_values.as_ref().map(|v| v.len()) == Some(1) {
+            let value = schema.enum_values.take().unwrap().pop().unwrap();
+            schema.const_value = Some(value);
```
mdibaiee, line 610: Do we need to `schema.enum_values = None` now?

crates/parser/src/config/mod.rs
mdibaiee: I think we can use:
```
unwrap_or(vec![])
```

crates/parser/src/config/mod.rs
```diff
@@ -601,31 +659,86 @@ mod test {
...
+    #[test]
+    fn basic_configs_are_deserialized_without_format_config() {
+        // We use adjacently tagged enums to represent the config. This means that any variant
+        // values, like `AdvancedCsvConfig`, cannot use `#[serde(default)]`. If a variant contains
+        // a config object, it must be present in the JSON. This means that we cannot add
+        // configuration to a format that currently has none, or else old configs that are missing
+        // the `config` property will suddently fail to deserialize. This test is here to fail in
+        // the case that someone adds a config object to a variant that currently has none.
+        // If you're here because this is failing, fear not. There's a way to fix it. Add a custom
+        // `Deserialize` impl for `Format` that deserializes the `config` using two passes. One the
```
mdibaiee, line 702: typo
```suggestion
        // `Deserialize` impl for `Format` that deserializes the `config` using two passes. On the
```

crates/parser/src/config/snapshots/parser__config__test__config_schema_is_generated.snap
```diff
@@ -15,86 +15,89 @@ expression: schema
...
           "properties": {
-            "auto": {
-              "default": {},
-              "type": "object"
+            "type": {
+              "default": "auto",
+              "const": "auto"
             }
-          },
-          "additionalProperties": false
```
mdibaiee, line 57: Is it okay that we don't have this anymore?

mdibaiee (commented):
thanks Phil, a bunch of comments and then I think it's good to go

crates/parser/src/config/mod.rs
psFried: In this case, we do. According to the JSON schema spec `enum: ["foo"]` and `const: "foo"` are identical. But JSON Forms treats these two differently, and requires `const` to be used for a `discriminator`. This is just changing the representation from `enum` to `const`.

crates/parser/src/config/snapshots/parser__config__test__config_schema_is_generated.snap
psFried: Yeah, I don't think disallowing additional properties really gets us much benefit, and it makes it more complicated to upgrade connectors since you can never remove properties from the schema without breaking existing configs.

crates/parser/src/config/mod.rs
psFried: Yeah, this is a safe unwrap since `Format` will always have a `oneOf` schema. We want this to panic if that changes for any reason (it would do so when running unit tests).

crates/parser/src/config/mod.rs
psFried: Ahh, sorry, I misunderstood. Setting `enum_values` to `None` is handled by calling [`take()`](https://doc.rust-lang.org/std/option/enum.Option.html#method.take) on the line above.

PR #599, opened 2022-07-28
docs: flowctl and get started

oliviamiannone:
**Description:**

* Adds more thorough documentation of the new flowctl and basic workflows
* Removes old "get started" section that was written for self-hosting and replaces it with a succinct page on how to actually get started currently
* Attempts to remove all outdated references to flowctl workflows
* Slaps beta notes on a couple things that are incorrect but beyond the scope of this work

**Notes for reviewers:**

The flowctl concepts I outline are based on my experimentation and the demo I attended recently- I think they'll provide a good starting point for review and likely will need correction. I'm very open to other opportunities to explain more concepts, too. 

I removed a lot of content from the section on build directories that I'm not 100% on so please take a look at that!

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/599)


site/docs/getting-started/installation.md
```diff
@@ -1,63 +1,43 @@
...
-Download and install the following prerequisites:
+[Learn more about using flowctl.](../concepts/flowctl.md)
 
-* [Docker](https://www.docker.com/get-started)
-* [VS Code](https://code.visualstudio.com)
-* VS Code [Remote Containers extension](https://code.visualstudio.com/docs/remote/containers)
+## Self-hosting Flow
 
-#### Create a Git repository from the Flow Template <a href="#create-a-git-repository-from-the-flow-template" id="create-a-git-repository-from-the-flow-template"></a>
+The Flow runtime is available for non-commercial use under the [Business Source License](https://github.com/estuary/flow/blob/master/LICENSE-BSL). It's possible to self-host Flow using a cloud provider of your choice.
```
psFried, line 28: nit: I think we can remove "for non-commercial use". The license does allow commercial use, as long as you don't try to run something like a Flow SAAS.

psFried (approved):
This looks good to me. I had a few suggestions and comments, but none of them are blocking.

PR #600, opened 2022-07-28
Minor POSIX-compatibility fix to makefile

willdonnelly:
**Description:**

The standard `test` (`[`) binary uses the syntax `[ foo = bar ]` for string equality comparison rather than `[ foo == bar ]`. This is easy to miss because many shells (like bash for instance) use their own builtin replacement which *does* accept `==`, but for whatever reason my setup was having trouble with that in the context of Makefile command execution.

For obvious reasons I have not tested this on a Mac, but in general I'd expect this change to be more likely to work in odd places, so hopefully it's fine.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/600)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -104,7 +104,7 @@ go-protobufs: $(GO_PROTO_TARGETS)
 ${PKGDIR}/bin/etcd:
 	# For Apple M1 we are currently using an unofficial built binary. Once the official binary for
 	# M1 is released we should switch to use that: https://github.com/etcd-io/etcd/issues/14001
-	if [ "$(UNAME)" == "Darwin arm" ]; then \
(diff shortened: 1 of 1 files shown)


willdonnelly:
Looks like the CI build is failing because of the recent `materialize-postgres` changes (https://github.com/estuary/connectors/pull/297) breaking end-to-end tests. I've uploaded https://github.com/estuary/flow/pull/601 which should hopefully address that.

williamhbaker (approved):
Still works for me on mac, nice!

mdibaiee (approved):

psFried (approved):
Good catch! I was actually just noticing that this was causing a problem for @oliviamiannone in codespaces, so this should fix the errors we saw there

PR #602, opened 2022-07-28
docs: updated parser config with discriminator 

oliviamiannone:
**Description:**

Updates docs on parser config per https://github.com/estuary/flow/pull/591

These are in:
* source-s3
* source-gcs
* source-http-file

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/602)

### DIFF
--- a/site/docs/reference/Connectors/capture-connectors/amazon-s3.md
+++ b/site/docs/reference/Connectors/capture-connectors/amazon-s3.md
@@ -46,19 +46,8 @@ To capture the entire bucket, omit `prefix` in the endpoint configuration and se
 | `/matchKeys` | Match Keys | Filter applied to all object keys under the prefix. If provided, only objects whose absolute path matches this regex will be read. For example, you can use &quot;.&#x2A;&#x5C;.json&quot; to only capture json files. | string |  |
 | `/parser` | Parser Configuration | Configures how files are parsed | object |  |
 | `/parser/compression` | Compression | Determines how to decompress the contents. The default, &#x27;Auto&#x27;, will try to determine the compression automatically. | null, string | `null` |
-| `/parser/format` |  | Determines how to parse the contents. The default, &#x27;Auto&#x27;, will try to determine the format automatically based on the file extension or MIME type, if available. | object | `{"auto":{}}` |
-| `/parser/format/auto` |  |  | object | `{}` |
-| `/parser/format/avro` |  |  | object | `{}` |
-| `/parser/format/csv` |  |  | object |  |
-| `/parser/format/csv/delimiter` | Delimiter | The delimiter that separates values within each row. Only single-byte delimiters are supported. | null, string | `null` |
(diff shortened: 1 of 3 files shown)


site/docs/reference/Connectors/capture-connectors/amazon-s3.md
```diff
@@ -138,14 +133,43 @@ Options are:
    For now, use a prefix in the endpoint configuration to limit the scope of each capture to data of a single file type.
    :::
 
-Only CSV data requires further configuration. When capturing CSV data, you must specify:
+#### CSV configuration
+
+Only CSV data allows further configuration. When capturing CSV data, you may additionally specify:
```
psFried, line 138: I think it's worth mentioning that the default is to automatically determine the values for delimiter, escape char, quote char, and encoding on a best-effort basis. Most users won't need to provide any configuration for parsing CSVs unless they're doing something weird.

site/docs/reference/Connectors/capture-connectors/amazon-s3.md
```diff
@@ -138,14 +133,43 @@ Options are:
...
+  * Tab (`"0x09"`)
+  * Vertical tab (`"0x0B"`)
+  * Unit separator (`"0x1F"`)
+  * SOH (`"0x01"`)
+  * Auto
 
-* **Delimiter**
 * **Encoding** type, specified by its [WHATWG label](https://encoding.spec.whatwg.org/#names-and-labels).
+
 * Optionally, an **Error threshold**, as an acceptable percentage of errors.
```
psFried, line 153: I kinda feel like Error Threshold could use more explanation, since it's pretty weird. It allows you to skip CSV rows that fail to parse, which is a pretty niche use case. It could probably do with some sort of warning, too, since by setting this to a value other than `0`, you're opting in to throwing away data in malformed rows (instead of stopping so you can fix them).

site/docs/reference/Connectors/capture-connectors/amazon-s3.md
```diff
@@ -138,14 +133,43 @@ Options are:
...
-* **Escape characters**
+
+* **Escape characters**. Options are:
+  * Backslash (`"\\"`)
+  * Disable escapes (`""`)
+  * Auto
+
 * Optionally, a list of column **Headers**, if not already included in the first row of the CSV file.
```
psFried, line 160: I'll suggest adding:
If any `headers` are provided, then it is assumed that the provided list of `headers` is complete and authoritative. The first row of your CSV file will be assumed to data (not headers), and you must provide a header value for every column in the file.

This is another one that's perhaps a little esoteric (though much less so than 'error threshold'), but it'll be important for some people and they'll likely need a little more guidance on how to use it.

psFried (approved):
LGTM :+1: I had a couple of suggestions, but none of them are blocking.

PR #606, opened 2022-07-29
network-tunnel: cleanup child process in case of success and failure

### LINKED ISSUE #598: Network Tunnel: we might be not releasing the socket when there is a failure
We have seen some instances of a running connector with Network Tunnel where we end up with a socket still listening on the port for network tunnel, after a failure, leading to subsequent tries to initiate a network tunnel to fail.

mdibaiee:
**Description:**

- Attempt to resolve #598
- According to https://doc.rust-lang.org/std/process/struct.Child.html#warning it's possible for processes to become zombies if they are not `wait`ed on. In this case, in the `prepare` step we do not wait on the process, which means if there is any failure in the `prepare` step, we will end up with a zombie. This is likely to be the problem we saw a few days ago where network tunnel initialisation failed during `prepare` and we ended up not being able to spawn a fresh instance because of the socket being still in use.

**Workflow steps:**

- Fail the `prepare` step of network tunnel initialisation, the child `ssh` process must be killed.

N/A

**Notes for reviewers:**

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/606)

### DIFF
--- a/crates/network-tunnel/src/main.rs
+++ b/crates/network-tunnel/src/main.rs
@@ -5,6 +5,7 @@ pub mod sshforwarding;
 
 use errors::Error;
 use flow_cli_common::{init_logging, LogArgs, LogFormat, LogLevel};
+use futures::future::{self, TryFutureExt};
 use std::io::{self};
 
 use interface::NetworkTunnelConfig;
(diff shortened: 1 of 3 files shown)


crates/network-tunnel/src/main.rs
```diff
@@ -22,21 +22,129 @@ async fn main() -> io::Result<()> {
...
+
+        // Write "READY" to stdio to unblock Go logic.
+        // The current workflow assumes that
+        //   1. After tunnel.prepare() is called, the network tunnel is able to accept requests from clients without sending errors back to clients.
+        //   2. The network tunnel is able to process client requests immediately after `tunnel.start_serve` is called.
+        // If either of the assumptions is invalid for any new tunnel type, the READY-logic need to be moved to a separate task, which
+        //    sends out the "READY" signal after making sure the network tunnel is started and working properly.
+        println!("READY");
+
+        prep.and(tunnel.start_serve().await)
```
willdonnelly, line 37: So as written here, I think this will run `tunnel.start_serve()` even if the result of `tunnel.prepare()` is an error. Is this the intended behavior, or should this maybe be `prep.and_then()` so it skips the `start_serve()` when prepare resulted in an error?

crates/network-tunnel/src/sshforwarding.rs
```diff
@@ -235,6 +236,19 @@ impl NetworkTunnel for SshForwarding {
 
         Ok(())
     }
+
+    async fn cleanup(&mut self) -> Result<(), Error> {
+        if let Some(process) = self.process.as_mut() {
+            process.kill().await?;
```
willdonnelly, line 242: Just double-checking that I've understood the logic correctly: if the process has already exited [this will result in an InvalidInput error](https://doc.rust-lang.org/std/process/struct.Child.html#method.kill), but that's okay because the code in `run_and_cleanup` discards any cleanup errors and just returns the result status of the `prepare; start_serve` sequence. Does that sound correct?

willdonnelly (approved):
LGTM % one question and one potential bug

crates/network-tunnel/src/main.rs
mdibaiee: @willdonnelly great catch, you are right it should be `and_then`

crates/network-tunnel/src/sshforwarding.rs
mdibaiee: @willdonnelly `run_and_cleanup` doesn't discard errors of `cleanup`, since it's using `?`. I just updated the code for this cleanup method to avoid erroring in case of `InvalidInput`, thanks for bringing this up!

willdonnelly:
LGTM

williamhbaker (approved):

PR #608, opened 2022-08-01
Allow disabling captures and materializations when the external systems are unreachable

psFried:
**Description:**

The problem that we observed is this. Say you have a capture or a materialization, and the system you're connecting to becomes unreachable in some way. It could be that someone changed the credentials, deleted the database, or anything else. In Flow, we often want to just disable these tasks so that we can reduce the noise of alerts, and then they can be resumed when the system comes back online. This was not possible because Flow would call into the connectors to `Validate` the specs when changing `/shards/disabled` from `false` to `true`. I also discovered that `flowctl api activate` would also call into the connector to `ApplyUpsert` disabled shards.

This PR updates the build step to skip the call to `Validate`, and also the activate step to skip the call to `ApplyUpsert`.

There's an argument to be made for still calling `ApplyUpsert` on disabled tasks, and just ignoring errors on those. It seems potentially useful for a connector to know when a task becomes disabled. But there's no clear use case for this at the moment, and I thought it simpler to just always skip the calls for now.

**Workflow steps:**

Use either the UI or CLI to disable a capture or a materialization. This should now succeed, even when the external system is unreachable.

The previous behavior wasn't documented, and it's probably not necessary to document the new behavior, either, since it's probably what users would expect anyway.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/608)

### DIFF
--- a/go/flowctl-go/cmd-api-activate.go
+++ b/go/flowctl-go/cmd-api-activate.go
@@ -84,6 +84,11 @@ func (cmd apiActivate) execute(ctx context.Context) error {
 		if !ok {
 			continue
 		}
+		if spec.ShardTemplate.Disable {
+			log.WithField("capture", spec.Capture.String()).
+				Info("Will skip applying capture because it's shards are disabled")
+			continue
+		}
 
 		driver, err := capture.NewDriver(ctx,
 			spec.EndpointType, json.RawMessage(spec.EndpointSpecJson), cmd.Network, ops.StdLogger())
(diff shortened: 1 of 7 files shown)


go/flowctl-go/cmd-api-activate.go
```diff
@@ -115,6 +120,11 @@ func (cmd apiActivate) execute(ctx context.Context) error {
 		if !ok {
 			continue
 		}
+		if spec.ShardTemplate.Disable {
+			log.WithField("capture", spec.Materialization.String()).
```
willdonnelly, line 124: s/capture/materialization/ ?

willdonnelly (approved):
LGTM % one copy-paste nit

jgraettinger:
One missing bit: i think we also need to skip deleting tasks where the last-deployed version of that task is disabled. Otherwise, the connector could get a nonsensical CaptureSpec / MaterializationSpec with invalid resource paths and fields. 

This implies that we won't clean up a task which you disable and then delete.

williamhbaker (approved):

Issue #624, opened 2022-08-08
reactor pods stuck in Terminating (capture PullClient)


jgraettinger:
Relevant trace:
```
goroutine 2138 [chan receive, 4052 minutes]:
github.com/estuary/flow/go/protocols/capture.(*adapterStreamServer).RecvMsg(0x2a32628?, {0x2726aa0?, 0xc001663800})
        /home/runner/work/flow/flow/go/protocols/capture/adapter.go:218 +0x45
github.com/estuary/flow/go/connector.protoWriteLoop({0x7f0c9470adb0, 0xc0013ccce0}, {0x2a32628, 0xc00064c330}, 0x2823758, {0x7f0c9470ad90, 0xc0035c3e90})
        /home/runner/work/flow/flow/go/connector/rpc.go:157 +0x1ef
github.com/estuary/flow/go/connector.StreamRPC.func1({0x7f0c9470ad90, 0xc0035c3e90})
        /home/runner/work/flow/flow/go/connector/rpc.go:129 +0x78
github.com/estuary/flow/go/connector.runCommand.func1()
        /home/runner/work/flow/flow/go/connector/run.go:177 +0x99
created by github.com/estuary/flow/go/connector.runCommand
        /home/runner/work/flow/flow/go/connector/run.go:175 +0x28a

goroutine 2302 [chan receive, 4051 minutes]:
github.com/estuary/flow/go/protocols/capture.(*coordinator).Close(0xc0014588c0)
        /home/runner/work/flow/flow/go/protocols/capture/coordinator.go:92 +0x2e
github.com/estuary/flow/go/runtime.(*Capture).Destroy(0xc000530dc0)
        /home/runner/work/flow/flow/go/runtime/capture.go:441 +0x2e
go.gazette.dev/core/consumer.waitAndTearDown(0xc0002e2160, 0xc000bcb2f0)
        /home/runner/go/pkg/mod/go.gazette.dev/core@v0.89.1-0.20220422210424-fa37e21e0464/consumer/shard.go:306 +0x44
created by go.gazette.dev/core/consumer.(*Resolver).cancelShards
        /home/runner/go/pkg/mod/go.gazette.dev/core@v0.89.1-0.20220422210424-fa37e21e0464/consumer/resolver.go:394 +0x6c
```

We get stuck tearing down a PullClient which never had `Serve()` invoked.

PR #625, opened 2022-08-08
protocols/capture: simplify public interface and ensure serve loop runs

jgraettinger:
We encountered some reactor pods which were wedged on shutdown.
On investigation, they had capture PullClients which were instantiated
but never started Serve(). The effect of this was they blocked
indefinitely on teardown awaiting resolution of the `loopOp` future,
which resolves only when the Serve() loop completes -- which could never
happen in these cases.

It was previously possible for a PullClient to be created but for the
shard to be cancelled or the task term changed before Serve() was
invoked. Refactor so that the serve loop is an inner implementation
detail which *must* be spawned for a constructed PullClient to be
returned. Further, defer the instantiation of a PullClient until needed
in StartReadingMessages.

**Description:**

No public impacts.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/625)

### DIFF
--- a/go/protocols/capture/pull_client.go
+++ b/go/protocols/capture/pull_client.go
@@ -26,6 +26,7 @@ func OpenPull(
 	spec *pf.CaptureSpec,
 	version string,
 	tail bool,
+	startCommitFn func(error),
 ) (*PullClient, error) {
 
 	var coordinator, err = newCoordinator(newCombinerFn, range_, spec, version)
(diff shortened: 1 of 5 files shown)


jgraettinger:
Fixes https://github.com/estuary/flow/issues/624

mdibaiee (approved):
Looks good to me

psFried (approved):
I think this looks good to merge.

PR #627, opened 2022-08-10
docs: Document some potential isues when building from scratch on an M1, and their resolutions

jshearer:
These are condensed from the notes I took when getting Flow to build on my new M1 macbook. I left out a few more obvious issues such as the error you get when `go` or `protobuf` aren't installed, as I felt it was fairly obvious what the fix was.

One thought I had is that these issues may not all be M1-specific, especially openssl and llvm. Maybe it should all be under a "getting started" section? My default was to just keep it simple.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/627)

### DIFF
--- a/README.md
+++ b/README.md
@@ -199,9 +199,27 @@ Flow mixes a variety of architectural techniques to achieve great throughput wit
 
 ## Building on M1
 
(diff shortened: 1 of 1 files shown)


mdibaiee (approved):
awesome, thanks a lot Joseph

PR #629, opened 2022-08-10
Use a process-scoped Context for ops logs appender

psFried:
**Description:**

The ops logs appender previously used the `Context` of the shard, which
caused log messages to be silently dropped as soon as the shard was
unassigned. This caused us to miss logs related to connector shutdown
because we generally shutdown connectors only after the shard context is
cancelled.

This change introduces a global `AppendService`, which is used by all
`LogAppender`s. This uses a context that will never be cancelled, and
thus should capture logs during the teardown of shards.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/629)

### DIFF
--- a/go/runtime/flow_consumer.go
+++ b/go/runtime/flow_consumer.go
@@ -55,6 +55,10 @@ type FlowConsumer struct {
 		Now *flow.Timepoint
 		Mu  sync.Mutex
 	}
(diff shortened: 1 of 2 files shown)


willdonnelly (approved):
LGTM

PR #634, opened 2022-08-16
Adds support for protobuf messages to parser

psFried:
**Description:**

Sometimes protocol buffers are used as a serialization format outside of
gRPC. An example, and the motivation for this change, is the
[GTFS realtime](https://gtfs.org/realtime/) data feed format.
This enables the parser to parse protobuf message content.

Only a single message is parsed from each file. The configuration must
provide both a `.proto` definition file and the name of a message
withing that file, which will serve as the target type. A single message
of that type will be deserialized because as far as I can tell, that's
how it's used. If there's a reason to add support for multiple messages
in the future, we can look into parsing the framed stream.

**Workflow steps:**

- Select "protobuf" as the `format` in the parser config (it will not be determined automatically, and requires configuration)
- Paste in the content of a `.proto` file, and specify the name of the message type to parse
- profit!

Source-S3, GCS, and filesource, will all be affected, as the `protobuf` format was added and affect anything using the parser.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/634)


crates/parser/src/config/mod.rs
```diff
@@ -303,6 +304,11 @@ pub enum Format {
     #[schemars(title = "CSV")]
     Csv(character_separated::AdvancedCsvConfig),
 
+    /// Parses a single protobuf message, using the given .proto file in the configuration.
```
willdonnelly, line 307: The whole bit about parsing "a single protobuf message" simultaneously makes perfect sense and bothers me a bit.

On the one hand I get where it's coming from, because a protobuf message as a whole doesn't have any sort of framing which would allow putting multiple into a file, so any file format which did that would be a different type.

Except...that's not quite true. If the top-level message type consists solely of a `repeated SomeOtherType` field, it's basically just a collection of individually-framed `SomeOtherType` messages, each with the standard protobuf length-prefixing. You can do all the normal framed-record things like concatenating and slicing them up and processing them individually.

In conclusion, this is a hard problem and I'm not sure how best to tackle it, I just kind of wanted to write up my thoughts on the matter.

willdonnelly (approved):
LGTM, although see my comment about the handling of `repeated` sub-messages. After I wrote the comment I went and looked at the GTFS proto definition, and it looks like this is exactly this sort of case -- there's one header field, but the majority of the file is actually a sequence of individual records via the `repeated FeedEntity` portion.

crates/parser/src/config/mod.rs
psFried: Yeah, I agree that it seems like something we should be able to handle a little better. One thought is to add an `unrollArray` option to the parser, which would accept a JSON pointer to an array. Elements of the array would each be emitted as separate documents, and values outside of that location would be repeated. It seems potentially useful for some other formats, too, though it may be good to keep it scoped to just protobuf for now.

Given an example config: `unrollArray: '/foo'`

and a document that's output from parsing a protobuf message to json:

```
{ "id": "a", "foo": [1, 2, 3]}
```

Would get turned into:

```
{"id": "a", "foo": 1}
{"id": "a", "foo": 2}
{"id": "a", "foo": 3}
```

willdonnelly: A more generic unrolling feature sounds useful, although part of my earlier point was that _at least in principle_ we can read and emit each individual instance of the repeated field in a streaming fashion, without needing to buffer the entire proto message into memory. (The handling of header fields makes this more tricky, but IIRC the GTFS documentation explicitly states that the header will occur first).

I'm not sure if the same would be true of a more generic `unrollArray` mechanism like you've described? Although I know that there exist streaming JSON parsers which would be able to do this sort of thing to JSON documents too...

jgraettinger: I'm uncomfortable with adding a capture-time unrolling feature, because that's what derivations for and it does violate the "one way of doing things" principle.

psFried: I started playing with this, and one observation is that in this case, the repeated field is not at the top level, but nested inside another (non-repeated) message. I don't think that presents a problem for unrolling, but it does mean that we'd require some configuration in order to know which field to unroll. This config could specify the field to unroll either using protobuf field names and dot notation (`unroll: field_foo.field_bar`), or using a json pointer and the JSON field names (`unroll: /fieldFoo/fieldBar`). Using the protobuf fields and dot notation seems a little more user friendly to me, but I could also see arguments for the json pointer. I'm honestly not sure which of these we should use, but would probably lean toward the protobuf dot notation unless you have a different opinion.

I think overall this seems like a reasonable improvement, and worth adding, though.

psFried: Somehow I didn't see your comment earlier, @jgraettinger. The messages for the Columbus GTFS data are indeed all small enough to be reasonable as a single document. For gigs, I checked the data for NYC, thinking that few feeds were likely to have messages bigger than theirs. Their messages seem to be around 250KB, which seems well within the size that Flow can handle. So we can just move forward without the unrolling.

PR #647, opened 2022-08-23
feature: Implement json schema `format` string validator

### LINKED ISSUE #282: schemas: implement string `format` validations
Right now these are unimplemented.

The [JSON Schema spec](https://json-schema.org/draft/2020-12/json-schema-validation.html#name-implementation-requirements) includes the details of how `format` should be handled.

jshearer:
This adds support for validating strings using the `format` field, as specified by jsonschema. Other than the two international variants described below, all jsonschema "format" optional tests now pass! 

For now, we fail when asked to validate `idn-email` and `idn-hostname`. The full test suite from jsonschema for these formats are absolutely bonkers complex, including specific cases for Japanese (Katagana/Hiragana), Han, Arabic, Hebrew, Virama, etc.

Fixes #282

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/647)


jshearer:
@psFried do you know what these diffs are about? I didn't make them (I think they're autogenerated), and CI is failing because of them

https://github.com/estuary/flow/pull/647/files#diff-354be26132f77cd405085cad8861d5edc1ff615969bb7dc97a487ea6dff43c18L614-R616

psFried:
> @psFried do you know what these diffs are about? I didn't make them (I think they're autogenerated), and CI is failing because of them
> 
> https://github.com/estuary/flow/pull/647/files#diff-354be26132f77cd405085cad8861d5edc1ff615969bb7dc97a487ea6dff43c18L614-R616

Those are generated by the `build.rs` in the `proto-flow` crate. It seems like it must be due to a difference in `protoc` version used at build time. On my machine I have protoc `3.14.0`, which seems to just be the version that's included in the fedora repos. I'd just remove those files from the PR for now, and then see if we can figure out a protoc version that works for everyone (or maybe we can somehow vendor it?).

crates/json/Cargo.toml
```diff
@@ -5,17 +5,24 @@ authors = ["Estuary Technologies, Inc"]
...
 fxhash = "*"
+iri-string = "0.6.0"
 itertools = "*"
+json-pointer = "0.3.4"
+lazy_static = "1.4.0"
 percent-encoding = "*"
 serde = "*"
 serde_json = "*"
 thiserror = "*"
+time = {version = "0.3.13", features=["macros", "formatting", "parsing"]}
```
psFried, line 22: Elsewhere in Flow we've used the `chrono` crate, and it'd be kinda nice to stick with one such library if possible. This is definitely not a big deal though. If it'd be easy enough to consolidate time crates, then it seems like a good idea, but it's by no means a blocker.

crates/json/src/schema/formats.rs
```diff
@@ -0,0 +1,88 @@
...
+        "time" => time::Time::parse(
+            val,
+            &time::macros::format_description!("[hour]:[minute]:[second].[subsecond]Z"),
+        )
+        .is_ok(),
+        "email" => parse_email_address(val).is_ok(),
+        "hostname" => parse_domain_name(val).is_ok(),
+        // The rules/test cases for these are absolutely bonkers
+        // If we end up needing this let's revisit (jshearer)
+        "idn-hostname" | "idn-email" => false,
```
psFried, line 50: It seems like a good idea to at least log a warning in this case. Without knowing the specifics of all the rules, I also wonder if perhaps the same `parse_email_address` and `parse_domain_name` functions would work and just be a little more permissive than what the rules stipulate.

TBH I've never seen anyone use this format in a json schema, so it's probably not worth worrying about much.

jgraettinger:
> > @psFried do you know what these diffs are about? I didn't make them (I think they're autogenerated), and CI is failing because of them
> > https://github.com/estuary/flow/pull/647/files#diff-354be26132f77cd405085cad8861d5edc1ff615969bb7dc97a487ea6dff43c18L614-R616
> 
> Those are generated by the `build.rs` in the `proto-flow` crate. It seems like it must be due to a difference in `protoc` version used at build time. On my machine I have protoc `3.14.0`, which seems to just be the version that's included in the fedora repos. I'd just remove those files from the PR for now, and then see if we can figure out a protoc version that works for everyone (or maybe we can somehow vendor it?).

Just to clarify, the `*.pb.go` files are not generated by Rust / `proto-flow` -- those generate the `.rs` Rust protobuf files (not changed here).

The `*pb.go` files are built by gogo-protobuf and its plugin `protoc-gen-gogo`. However, what could also have happened here is that you may have saved the files and your editor "helpfully" reformatted some of the comment markdown?

crates/json/src/schema/formats.rs
```diff
@@ -0,0 +1,88 @@
...
+    static ref URI_TEMPLATE_RE: Regex = Regex::new(
+        r#"^(?:(?:[^\x00-\x20""'<>%\\^`{|}]|%[0-9a-f]{2})|\{[+#.\/;?&=,!@|]?(?:[a-z0-9_]|%[0-9a-f]{2})+(?:\:[1-9][0-9]{0,3}|\*)?(?:,(?:[a-z0-9_]|%[0-9a-f]{2})+(?:\:[1-9][0-9]{0,3}|\*)?)*\})*$"#
+    )
+    .expect("Is a valid regex");
+    static ref ISO_8601_DURATION_RE: Regex = Regex::new(r"^P(?!$)(\d+(?:\.\d+)?Y)?(\d+(?:\.\d+)?M)?(\d+(?:\.\d+)?W)?(\d+(?:\.\d+)?D)?(T(?=\d)(\d+(?:\.\d+)?H)?(\d+(?:\.\d+)?M)?(\d+(?:\.\d+)?S)?)?$").expect("Is a valid regex");
+    static ref ISO_8601_ONLY_WEEKS_RE: Regex = Regex::new(r"^[0-9P|W]*$").expect("Is a valid regex");
+    static ref ISO_8601_NO_WEEKS_RE: Regex = Regex::new(r"^[^W]*$").expect("Is a valid regex");
+}
+
+pub fn validate_format(format: &String, val: &str) -> bool {
```
jgraettinger, line 25: nit: format could be &str instead of &String

psFried (approved):
This looks great! I think it's definitely worth adding some logging for the `idn-*` cases and removing those generated files, but good to merge after that.

crates/json/src/schema/formats.rs
```diff
@@ -0,0 +1,88 @@
...
+            }
+        }
+        "ipv6" => match IpAddr::from_str(val) {
+            Ok(i) => i.is_ipv6(),
+            Err(_) => false,
+        },
+        // uuid crate supports non-hyphenated inputs, jsonschema does not
+        "uuid" if val.len() == 36 => Uuid::parse_str(val).is_ok(),
+
+        "duration" => match ISO_8601_DURATION_RE.is_match(val) {
```
jgraettinger, line 67: supernit: the match for booleans is a bit odd. In Rust, `if`-blocks can be expressions just like `match` blocks, which might make this a bit clearer?

crates/json/src/schema/formats.rs
```diff
@@ -0,0 +1,88 @@
...
+                false => ISO_8601_NO_WEEKS_RE.is_match(val).unwrap_or(false),
+            },
+            _ => false,
+        },
+        "iri" => iri_string::validate::iri::<IriSpec>(val).is_ok(),
+        "uri" => iri_string::validate::iri::<UriSpec>(val).is_ok(),
+        "uri-reference" => iri_string::validate::iri_reference::<UriSpec>(val).is_ok(),
+        "iri-reference" => iri_string::validate::iri_reference::<IriSpec>(val).is_ok(),
+        "uri-template" => URI_TEMPLATE_RE.is_match(val).unwrap_or(false),
+        "json-pointer" if !val.get(0..1).map_or(false, |first| first.eq("#")) => {
```
jgraettinger, line 81: nit: `json-pointer` could be a regex pretty easily, removing the dependency. [We do this already in the models crate](https://github.com/estuary/flow/blob/master/crates/models/src/references.rs#L35).

FYI we also have a json-pointer implementation [here in the docs crate](https://github.com/estuary/flow/blob/master/crates/doc/src/ptr.rs). Arguably it should be in the `json` create but :shrug:

crates/json/Cargo.toml
```diff
@@ -5,17 +5,24 @@ authors = ["Estuary Technologies, Inc"]
 edition = "2018"
 
 [dependencies]
+addr = {version="0.15.4", features=["idna", "std","psl"]}
 # TODO(johnny): bitvec had a breaking 1.0 release we've not upgraded to yet.
 bitvec = "0.19"
+bytecount = "0.6.3"
```
jgraettinger, line 11: I don't see bytecount used anywhere?

jgraettinger (approved):
This looks great! :tada:

Just some minor nits plus the spurious *.pb.go changes, then LGTM.

crates/json/Cargo.toml
jgraettinger: Yea, which should we be doing? I get the impression that the community is coalescing around `time` but what do you think?

psFried:
> Just to clarify, the `*.pb.go` files are not generated by Rust / `proto-flow` -- those generate the `.rs` Rust protobuf files (not changed here).

Whoops, yea IDK why I saw those and thought they were the rust files :coffee: :coffee: :coffee:

crates/json/Cargo.toml
jshearer: Removed, along with `json-pointer`

crates/json/src/schema/formats.rs
jshearer: I tried referencing the `models` crate from `json`, but that created a circular dependency, so I just copied the json pointer regex. How smelly does this feel to you?

crates/json/Cargo.toml
jshearer: I don't have an opinion on this. Happy to replace time with chrono if that's where we want to go

PR #651, opened 2022-08-25
Parser unrolls top-level JSON arrays

psFried:
**Description:**

Previously flow-parser would error if a top-level json type were
anything other than an object. This updates the parser to also handle
JSON arrays of objects by unrolling them.

**Workflow steps:**

There's no change to how you'd use this, except that now it'll handle arrays in addition to objects.

The previous behavior was undocumented, so there's no change. Though I'm starting to wonder if we should add a general parser page to our docs where we could document things like this.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/651)


jgraettinger (approved):
LGTM % comments

Is it reasonable to add support for top-level arrays when we're sniffing a representation from peeked file bytes?

> Reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/651#-:-NAKwox-BDmnvMlwdI4s:b8v1do9)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @jgraettinger and @psFried)

___
*[`crates/parser/src/format/json.rs` line 65 at r1](https://reviewable.io/reviews/estuary/flow/651#-NAKwZPs3NJ_CAL775SH:-NAKwZPtB4qb49T1mlZc:bgwmaoa) ([raw file](https://github.com/estuary/flow/blob/82c5d8800c00b7d5132563f66573aa4b6395e9e8/crates/parser/src/format/json.rs#L65)):*
> ```XML
> 
> #[derive(Debug, thiserror::Error)]
> #[error("The JSON parser only supports objects or arrays of objects, found scalar value: '{0}'")]
> ```

nit: it could also be two levels of arrays.
___
*[`crates/parser/tests/examples/valid-json-top-level-array.json` line 1 at r1](https://reviewable.io/reviews/estuary/flow/651#-NAKwkw_-eSLcfGn55sY:-NAKwkw_-eSLcfGn55sZ:bq9tz71) ([raw file](https://github.com/estuary/flow/blob/82c5d8800c00b7d5132563f66573aa4b6395e9e8/crates/parser/tests/examples/valid-json-top-level-array.json#L1)):*
> ```JSON
> [
> ```

is this used somewhere? I don't see it.

crates/parser/src/format/json.rs
```diff
@@ -21,7 +22,120 @@ impl Parser for JsonParser {
...
+            }
+            Ok(other) if other.is_object() => Some(Ok(other)),
+            Ok(other) => Some(Err(ParseError::Parse(InvalidJsonType(other).into()))),
+            Err(e) => Some(Err(e.into())),
+        }
+    }
+}
+
+#[derive(Debug, thiserror::Error)]
+#[error("The JSON parser only supports objects or arrays of objects, found scalar value: '{0}'")]
```
willdonnelly, line 65: s/scalar value/value/? Just because it sounds silly if given an input like `[[1,2],[3,4]` to say "scalar value: `[1,2]`".

willdonnelly (approved):
LGTM

psFried:
@jgraettinger That file ([crates/parser/tests/examples/valid-json-top-level-array.json](https://reviewable.io/reviews/estuary/flow/651#-NAKwkw_-eSLcfGn55sY:-NAKwkw_-eSLcfGn55sZ:bq9tz71)) is used by [this test](https://github.com/estuary/flow/blob/master/crates/parser/tests/happy_path_test.rs#L23), which parses every `valid*` example file.

I fixed the wording of the error message, too. Thanks for pointing that out.

PR #654, opened 2022-08-31
docs: add generic derivation guide

oliviamiannone:
**Description:**

Adds a generalized user guide to creating a derivation with flowctl. 
This is much-0needed to accompany the existing [guide to create a ui-only dataflow](https://docs.estuary.dev/guides/create-dataflow/). 

The example scenario / code works, and is provided just to have some sort of basic reference, but it still might be nice to revisit in the future to provide something more compelling (especially because, as I found out today, source-hello-world _always_ outputs infinite greetings).

**Workflow steps:**

Create a derivation, start to finish, as if you were a new user.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/654)

### DIFF
--- a/site/docs/guides/create-derivation.md
+++ b/site/docs/guides/create-derivation.md
@@ -22,7 +22,7 @@ If you need help, see the [guide to create a Data Flow](./create-dataflow.md).
 ## Pull your specification files locally
 
 To begin working in your local environment, you must authenticate Flow from the command line.
-Then, you'll need to add your source collection's specification files to a **draft** and copy the draft locally.
+Then, you'll need to add your source collection's specification files to a **draft** and bring the draft into your local environment for editing.
 
 1. Go to the [Flow web application](https://dashboard.estuary.dev). On the **Admin** page, click the **CLI-API** tab and copy the access token.
 
@@ -81,15 +81,37 @@ Then, you'll need to add your source collection's specification files to a **dra
 
    :::tip
    Use `tree` to visualize your current working directory. This is a helpful tool to understand the [files that underlie your local draft](../concepts/flowctl.md#development-directories).
+   For example:
+
+   ```console
+   .
+   ├── namespace
+   │   └── data-flow-name
+   │       └── flow.yaml
+   ├── flow.yaml
+   ├── flow_generated
+   │   ├── flow
+   │   │   ├── main.ts
+   │   │   ├── routes.ts
+   │   │   └── server.ts
+   │   ├── tsconfig-files.json
+   │   └── types
+   │       └── namespace
+   │           └── data-flow-name
+   │               └── my-collection.d.ts
+   ├── package.json
+   └── tsconfig.json
+   ```
    :::
 
 2. Open the specification file in your preferred editor.
 
-   It will look similar to the following (this example uses the default collection from the Hello World test capture, available in the web app):
+   It will look similar to the following. (This example uses the default collection from the Hello World test capture, available in the web app):
 
    ```yaml
    collections:
-      estuary/data-flow-name/greetings:
+      #The Hello World capture outputs a collection called `greetings`.
+      namespace/data-flow-name/greetings:
          schema:
             properties:
             count:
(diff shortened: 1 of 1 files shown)


site/docs/guides/create-derivation.md
```diff
@@ -0,0 +1,251 @@
+---
+sidebar_position: 2
+---
+# Create a derivation with flowctl
+
+Once you're familiar with creating a basic [Data Flow](../concepts/README.md#essential-concepts), you can take things a step further
+and transform your data with [derivations](../concepts/derivations.md).
+A **derivation** is a kind of Flow collection that results from the transformation of one or more other collections.
```
jshearer, line 8: I think this is a true statement, but I'm not entirely sure it's useful. I would probably frame this more as a derivation being a transformation applied to a collection, where the benefit is that your data is transformed in a consistent way, respecting the order/sequence in which it was captured. That transformation is then stored as a flow collection itself, which has the cute consequence of being able to write derivations _of other_ derivations (in addition to being able to materialize it to any of the supported destinations).

site/docs/guides/create-derivation.md
```diff
@@ -0,0 +1,251 @@
...
+
+* An existing Flow **collection**. Typically, you create this through a **capture** in the Flow web application.
+If you need help, see the [guide to create a Data Flow](./create-dataflow.md).
+
+* **flowctl** installed locally. For help, see the [installation instructions](../concepts/flowctl.md#installation-and-setup).
+
+## Pull your specification files locally
+
+To begin working in your local environment, you must authenticate Flow from the command line.
+Then, you'll need to add your source collection's specification files to a **draft** and copy the draft locally.
```
jshearer, line 25: Should this be
> And ~copy~ **edit** that draft locally

?

jshearer: Ah, I see. Reading below, there is an explicit step to copy the draft locally. I would actually still word this as "edit" not "copy" (since the "copy" part is an implementation detail, whereas editing is what you actually care about), but this is a pretty nit-picky nit /shrug

site/docs/guides/create-derivation.md
```diff
@@ -0,0 +1,251 @@
...
+
+## Add a derivation
+
+1. Locate the specification YAML file for the collection you want to transform.
+
+   In your working directory, you'll see a top-level file called `flow.yaml`.
+   Within a subdirectory that shares the name of your Data Flow, you'll find a second `flow.yaml` — this contains the collection specification.
+
+   :::tip
+   Use `tree` to visualize your current working directory. This is a helpful tool to understand the [files that underlie your local draft](../concepts/flowctl.md#development-directories).
```
jshearer, line 83: I might include an example output of `tree` in a draft directory here, just so people have something to visualize the directory structure. I know that'd be useful for me at least

site/docs/guides/create-derivation.md
```diff
@@ -0,0 +1,251 @@
...
+         key:
+            - /dozens
+         derivation:
+            transform:
+               greetings-by-dozen:
+                  source: {name: estuary/data-flow-name/greetings}
+                  publish: {lambda: typescript}
+            typescript: {module: divide-by-twelve.ts}
+      ```
+   Save the file.
```
jshearer, line 147: I would maybe add some more description of what's going on here. Why did you add `dozens` to the `required` keys? Does that come from the `key: - /dozens` field? What does that field actually do? Do I _have_ to add it? Why? etc

jshearer: Maybe these are explained in more detail somewhere else in the docs? Maybe like a spec for this yaml document, explaining what various fields/stanzas do? If so we should link there

jshearer (approved):
This looks great as is! I made a few comments/suggestions that I think would improve things, feel free to take them or leave them at your discretion :)

site/docs/guides/create-derivation.md
jshearer: @jgraettinger can you fact-check me here? Am saying things that are actually true? :P

jgraettinger: this is all true, but the docs context matters. this is pretty early on in the user's journey, and at this stage it's at the level of "oh derivations is how I munge my data".

I don't know how to pack what you said into a sentence that's as concise and approachable as what you're replacing.

jshearer: This is vaguely what I was thinking... while also totally understanding that writing isn't my job so absolutely feel free to ignore me completely @oliviamiannone 😅 

--- 

A derived collection takes data from an input such as a capture, and applies some transformation to that stream of data, keeping order intact. This transformed stream is itself stored as a flow collection, meaning that you can materialize it exactly the same way you would the original capture. 

::tip::
One consequence of this is that derivations can be based on other derivations, allowing you to build up complex and powerful flows of data, while also enabling the intermediary steps to be used as datasets in their own right.
```

oliviamiannone: Thanks for this - I like where you're going with it and am going to riff on it a bit!

site/docs/guides/create-derivation.md
oliviamiannone: I had linked schema and collection key docs above, but I can see how that'd be easy to just skim past. Added comments inline in the sample and I think it's a lot clearer now

PR #659, opened 2022-09-06
oauth: add oauth to materialize protocol

williamhbaker:
**Description:**

Following up to https://github.com/estuary/flow/commit/a656a84e02e49b2c0a63ed6ffd4e143a6e8eccb7, this adds oauth to the materialization protocol, in a similar way that it was added to the capture protocol. This will help with https://github.com/estuary/connectors/issues/326 & https://github.com/estuary/connectors/issues/327

**Workflow steps:**

This will allow materializations to use oauth via communication of the required values in the oauth spec. I've tested this end-to-end with a new connector (see https://github.com/estuary/connectors/pull/336) with Google oauth and it seems to work well.

N/A

**Notes for reviewers:**

My editor seems to have some different ideas of how the `go/protocols/materialize/materialize.proto` file should be formatted; but very little in that file is actually changing.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/659)


willdonnelly (approved):
LGTM, sorry for the delay in getting to this one

PR #667, opened 2022-09-13
network-tunnel: use custom error with more descriptive messages

mdibaiee:
**Description:**

- Refactor our streams so that our InterceptorStream uses our custom Error enum
- Made the messages more descriptive and added an error code E001, E002, ... to the errors so that we can distinguish between them easier when troubleshooting.

**Workflow steps:**

**Notes for reviewers:**


jgraettinger:
Just a comment on the description: we also have and have used shortlinks as an alternative to user facing error codes. Would that be appropriate here?

mdibaiee:
@jgraettinger we could potentially do that if we plan to document errors on web as well, is that something we want to do?

So you mean instead of E001, we do go.estuary.dev/E001?

jgraettinger:
Yes, or for example go.estuary.dev/xY5tP47

It's not any harder to look up for devs, and can be updated after-the-fact to link to useful documentation.

crates/connector_proxy/src/errors.rs
```diff
@@ -1,64 +1,82 @@
...
-    #[error("invalid endpoint json config.")]
-    InvalidEndpointConfig,
+    #[error("go.estuary.dev/E007: The connector's protocol does not match the requested protocol. Connector protocol is {0}, requested protocol is {1}")]
+    MismatchingRuntimeProtocol(String, &'static str),
 
-    #[error("invalid json pointer '{0}' to config.")]
-    InvalidJsonPointer(String),
-
-    #[error(transparent)]
+    #[error("go.estuary.dev/E008: IO Error: {0}")]
```
jgraettinger, line 30: when you have a wrapped error don't directly format it into your error string. `anyhow` has nice chained-error formatting already where it will walk and format the causal error chain for you.

```
#[error("go.estuary.dev/E008: IO Error")]
IOError(#[from] std::io::Error),
```

(I've made this mistake myself in the past, and there are still places where we've done it that ought to be cleaned up, but here we are :shrug:).

It does mean that the top-level error presentation needs to print the formatted chained error. If you have a `tokio::main` as I believe you do here, though, IIRC you can just return anyhow::Error and it does the right thing:

```
#[tokio::main]
async fn main() -> Result<(), anyhow::Error> {
```

jgraettinger (approved):
LGTM % comment below on (not) inline-formatting causal error chains.

You've already done the work of creating these shortlinks so :shrug: but, had you not yet, I would probably have recommended only creating short-links for errors where there's more to legitimately say beyond the coded error message body, and to use the random-generated codes like `QpEo4F` vs `E123` vs `W123`, etc.

The random codes are obviously random, whereas it's a bit weird if `E123` happens to be re-classified as a warning, or if there are gaps in the sequence.

psFried:
I love the thoroughness here, but I think perhaps it's taking things a bit too far in terms of having links to error types that aren't relevant or actionable for the user.  Looking at the error variants we have, many of them seem like they're actually indicative of either i/o errors or programming errors. These categories seem most appropriate to display to the user as an "internal error" or something like that. For example, I'm skeptical that if a user follows the link in `go.estuary.dev/E010: Decoding protobuf RPC messages:`,  that we'd be able to put any content there that's helpful to them, apart from just directing them to file a bug report. So I think my suggestion here is that, rather than including separate links of all of these, we only do that for the ones that are likely actionable for a user. For the remainder, I think it makes more sense to just make it clear that it's not their fault, and that we'd be happy to investigate the issue.

mdibaiee:
@psFried I see where you are coming from, the reason I have put different error codes for each error even if they all may belong to the wider category of "internal error" is that it makes reporting and diagnosis of the issue much easier if we can pinpoint exactly which of those internal errors happened. In this case, we may choose to redirect all such shortlinks to the same page that says: This error is an internal error, please report the error, and that's okay, but we will have more information about the specific error that happened and we can address it more easily on our side. What do you think about this?

psFried:
For error variants that are always indicative of "internal errors", it sounds like the UX would be for a user to see some error message like `go.estuary.dev/E010: Decoding protobuf RPC messages: ..`, and then copy/paste the link to go to a page that tells them it's an internal error and (presumably) that they need to reach out to us to file a bug report or get help. In that case, it seems simpler just tell them directly in the error message, and save the indirection. I'm assuming here that the error messages are already clear and unambiguous to _us_, since we're able to just grep the code for the message.

Of course for error variants that _are_ indicative of something the user has configured incorrectly, I think the link to a help page is great. Also, I don't actually want to take it as a given that every error variant will map unambiguously into one of those two categories, and in such cases I'd imagine that linking to help pages would also be good.

If we know a given variant indicates a bug, then perhaps it could be better to format a URL that [creates a github issue](https://docs.github.com/en/issues/tracking-your-work-with-issues/creating-an-issue#creating-an-issue-from-a-url-query) with a pre-filled title and description?

**Overall, though, I'm fine with just merging this PR as it is. You're a lot closer to this than I am, so please feel free to ignore any ideas that you don't think are helpful.**

Side note: In the future, it might also be interesting to try to generalize the concept of error codes and links, and fit them into a common structured error representation so that we can present them better.

mdibaiee:
@psFried okay, that makes sense. We can indeed redirect those links directly to a GitHub issue filing + additional information (it would be amazing if we could include additional information in the link such as backtrace, etc.)

PR #671, opened 2022-09-14
Docs: Materialize Google Pub/Sub

oliviamiannone:
**Description:**

Adds docs for the new Google Pub/Sub materialization
(https://github.com/estuary/connectors/pull/336)

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/671)

### DIFF
--- a/site/docs/reference/Connectors/materialization-connectors/README.md
+++ b/site/docs/reference/Connectors/materialization-connectors/README.md
@@ -19,6 +19,9 @@ Estuary is actively developing new connectors, so check back regularly for the l
 * Google BigQuery
   * [Configuration](./BigQuery.md)
   * Package — ghcr.io/estuary/materialize-bigquery:dev
+* Google Cloud Pub/Sub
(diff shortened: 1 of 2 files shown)


williamhbaker:
The only other nuance to this connector is the way it handles message ordering. In-order consumption of messages is possible because the connector publishes messages using [ordering keys](https://cloud.google.com/pubsub/docs/publisher#using-ordering-keys). The ordering key for a message is its [flow collection key](https://docs.estuary.dev/concepts/collections/#keys). This way subscriptions can be set up to [consume messages in-order, per-key](https://cloud.google.com/pubsub/docs/ordering).

I think a brief additional section under a heading like "Message Ordering" with content along the lines of the following would be good:

> Messages are published with ordering keys set to the Flow collection key of the documents being published. This allows for PubSub subscriptions to receive messages in order. See [Google Pubsub Ordering Docs](https://cloud.google.com/pubsub/docs/ordering) for more information.

williamhbaker (approved):

PR #674, opened 2022-09-15
connector-proxy: close stdout when child stdout closes

mdibaiee:
**Description:**

- Here is an attempt to ensure that as soon as the child process exits, we close stdout of connector-proxy
- However... this does not work. From my tests, I think the closing of the stdout is successful in connector-proxy, but my current understanding is that `docker run` does not propagate this closing of the stdout and keeps an open `stdout` even if the entrypoint closes its stdout. I tested with and without `--init` flag. Still investigating...

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/674)

### DIFF
--- a/crates/connector_proxy/src/connector_runner.rs
+++ b/crates/connector_proxy/src/connector_runner.rs
@@ -1,3 +1,5 @@
+use std::ffi::CString;
+
 use crate::apis::{FlowCaptureOperation, FlowMaterializeOperation, InterceptorStream};
 use crate::errors::{create_custom_error, Error};
 use crate::interceptors::{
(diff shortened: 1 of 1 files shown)


mdibaiee:
Okay, I can confirm the issue is in behaviour of `docker run`, here is a rundown of the issue:

You can test the behaviour of docker run using a busybox container, to see how it handles `stdin`:

The example below does not print "test":
```
echo "test" | docker run busybox "head"
```

I tried attaching stdin, but still no "test":
```
echo "test" | docker run --attach stdin --attach stdout busybox "head"
```

By reading the [the code](https://github.com/moby/moby/blob/4118e0c9eebda2412a09ae66e90c34b85fae3275/runconfig/opts/parse.go#L531), it seems that stdin is actually not opened at all unless `--interactive` is specified, and `--interactive` is problematic, because it "Keeps STDIN open even if not attached". In practice though, it seems to keep stdout open as well.

I verified this PR by running a pull RPC using [cdk](https://github.com/estuary/cdk/), and then going inside the container. I can see that the code is correctly replacing stdout file descriptor of the process with `/dev/null`:
```
docker/connector $ cd /proc/1/fd
/proc/1/fd $ ls -lha
lr-x------ 1 nonroot nonroot 64 Sep 15 15:08 0 -> 'pipe:[35608139]'
lr-x------ 1 nonroot nonroot 64 Sep 15 15:08 1 -> /dev/null
```

However stdout of `docker run` stays open...

PR #678, opened 2022-09-16
Phil/flowctl reads

psFried:
**Description:**

Adds a first pass at a Gazette journal client in Rust. This currently only does uncommitted reads of individual journals, but it does handle direct reads of fragment files.

**Workflow steps:**

no user-facing changes here, but this crate will be used by the new `flowctl` to power some new features there.

**Notes for reviewers:**

There's portions of this that are still somewhat incomplete. In some cases, I'm intentionally leaving it that way so that refactoring can be done in the context of planned higher level APIs for reading committed messages from Flow collections. When in doubt, feel free to ask.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/678)


psFried (commented):
> *[Reviewable](https://reviewable.io/reviews/estuary/flow/678)* status: 10 of 12 files reviewed, 5 unresolved discussions (waiting on @jgraettinger)

___
*[`Cargo.lock` line 1 at r4](https://reviewable.io/reviews/estuary/flow/678#-NCWG9BwEiobZTF5sIsv:-NCWeCZr0a1YAucTFg7Y:blwotyl) ([raw file](https://github.com/estuary/flow/blob/4fcc28ec59b033baef9019193daf8c3a2cd18691/Cargo.lock#L1)):*

a lot of new stuff in here, plus multiple versions of things we already have. Any of it avoidable?
</blockquote></details>


Most of the added deps were due to adding a dependency on tonic with the tls feature, but `async-compression` also added a good few.
I _think_ the biggest source of duplicate deps here is actually clap, since we have both v2 and v3, though Mahdi has recently fixed this.
There might be a few that we could clean up here or there, but my overall disposition is to wait until we can use workspace inheritance and start being more explicit with our dependency versions before we go through and really clean up our lockfile.
___
*[`crates/journal-client/Cargo.toml` line 23 at r4](https://reviewable.io/reviews/estuary/flow/678#-NCWGCmo6flsJg8MLpDV:-NCWczwb5FBX72DuXMTH:bb7f2mf) ([raw file](https://github.com/estuary/flow/blob/4fcc28ec59b033baef9019193daf8c3a2cd18691/crates/journal-client/Cargo.toml#L23)):*

nit: cleanup commented out stuff.
</blockquote></details>

done
___
*[`crates/journal-client/src/fragments.rs` line 9 at r4](https://reviewable.io/reviews/estuary/flow/678#-NCWGPWf2B0XS90PPI7l:-NCWcsD-9uOuVgaNnH2v:b-hu2ca) ([raw file](https://github.com/estuary/flow/blob/4fcc28ec59b033baef9019193daf8c3a2cd18691/crates/journal-client/src/fragments.rs#L9)):*

nit: is this a list_fragment RPC (not read)?
</blockquote></details>

:+1:  will fix
___
*[`crates/journal-client/src/lib.rs` line 26 at r4](https://reviewable.io/reviews/estuary/flow/678#-NCWI0X32WNwOBV80oQD:-NCW_gShA1M30QP22h2Z:b-b4soik) ([raw file](https://github.com/estuary/flow/blob/4fcc28ec59b033baef9019193daf8c3a2cd18691/crates/journal-client/src/lib.rs#L26)):*

Tonic has [MetadataMap](https://docs.rs/tonic/latest/tonic/metadata/struct.MetadataMap.html) which (I'm hoping) has built-in client support. gRPC has a "metadata" concept which is actually just implemented as HTTP request headers. Is that a better mechanism for plumbing through authentication?
</blockquote></details>

Yeah, I actually used then for my first attempt at this, following [this example](https://github.com/hyperium/tonic/blob/master/examples/src/authentication/client.rs), which made it all look oh so simple. The problem with that example is that the type of the client become unnameable because the function ends up as a generic type parameter. Having just taken another look, I now see how we could implement the `Interceptor` trait for `WithAuthToken`, which seems worth a try.
___
*[`crates/journal-client/src/list.rs` line 11 at r4](https://reviewable.io/reviews/estuary/flow/678#-NCWIc9h5vQd3hx48nAa:-NCWbGQG0dni0NBE40hp:bg3fle1) ([raw file](https://github.com/estuary/flow/blob/4fcc28ec59b033baef9019193daf8c3a2cd18691/crates/journal-client/src/list.rs#L11)):*

IIRC ListRequest / ListResponse offer pagination fields, but I don't believe the broker ever actually implemented it...
</blockquote></details>

Yeah, there was a comment in the protobuf file about those fields having been removed. My thought it is that for Flow, we really only need to list the journals of a single collection at a time, so hopefully that well keep the total number of journal specs in a response to within reason.
___
*[`crates/journal-client/src/read/mod.rs` line 16 at r4](https://reviewable.io/reviews/estuary/flow/678#-NCWJ2DVAPJMtzAOMUxt:-NCWc18e68ojhmBCGkyE:bvqb5bd) ([raw file](https://github.com/estuary/flow/blob/4fcc28ec59b033baef9019193daf8c3a2cd18691/crates/journal-client/src/read/mod.rs#L16)):*

where / when did `reqwest` enter the fray?
</blockquote></details>

reqwest is used for fetching the fragment files using the pre-signed URLs. This error variant is returned when there's an error executing the HTTP request, or if the response status is not OK.
___
*[`crates/journal-client/src/read/uncommitted.rs` line 26 at r4](https://reviewable.io/reviews/estuary/flow/678#-NCWJWw21p3YIVITnVgL:-NCWcYJR-hpxFTNaDzQh:b2zjyug) ([raw file](https://github.com/estuary/flow/blob/4fcc28ec59b033baef9019193daf8c3a2cd18691/crates/journal-client/src/read/uncommitted.rs#L26)):*

Comment is not accurate. The broker will happily take the request and will start sending data once the write head has caught up to your offset.
</blockquote></details>

:+1: will fix

PR #680, opened 2022-09-19
cargo: remove clap:2 and structopt, replace with clap:3

mdibaiee:
**Description:**

- I tried to use `flow_cli_common` in another project (airbyte-to-flow), but Cargo got confused about the two versions of clap we have in our `Cargo.lock` in this repository, turns out we had some packages still using clap:2 with structopt, which is now part of clap:3 as derive. I just migrated the code to use clap:3 and removed clap:2.

**Workflow steps:**

N/A

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/680)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -54,6 +54,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "anes"
+version = "0.1.6"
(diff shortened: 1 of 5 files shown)


psFried (approved):
LGTM :clap:

PR #682, opened 2022-09-20
connector: use tcp socket for communicating with connectors

mdibaiee:
**Description:**

- use a tcp socket for communicating with connectors
- I could not test UDS due to https://github.com/docker/for-mac/issues/483, but adding it should not be too hard if necessary
- We need a way to migrate our existing deployments to this new method without disruption in service, so I may have to change this code so that it supports both methods somehow, open to suggestions on that
- At the moment this pull-request means network-tunnel doesn't work anymore since I'm bypassing connector-proxy completely. Moving network-tunnel to an independent binary should also be worked before we can fully migrate.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/682)

### DIFF
--- a/go/connector/run.go
+++ b/go/connector/run.go
@@ -72,7 +72,6 @@ func Run(
 ) error {
 	var imageInspectRaw json.RawMessage
 	var inspects = []ImageInspect{ImageInspect{}}
-	var inspect ImageInspect
 	// Pull and inspect the image, saving its output for mounting within the container.
 	if err := PullImage(ctx, image); err != nil {
 		return err
@@ -80,15 +79,13 @@ func Run(
 		return err
 	} else if err := json.Unmarshal(imageInspectRaw, &inspects); err != nil {
 		return fmt.Errorf("go.estuary.dev/E132: parsing container image %q inspect results: %w", image, err)
-	} else {
-		inspect = inspects[0]
 	}
 
 	var tmpProxy, tmpInspect *os.File
 
 	// Find out the port on which the connector will be listening. If there is no
 	// port specified, then fallback to using connector_proxy and stdio
-	var port string = inspect.Config.Labels[ConnectorTCPPortLabel]
+	var port = inspects[0].Config.Labels[ConnectorTCPPortLabel]
 	if port == "" {
 		logger.Log(logrus.WarnLevel, logrus.Fields{}, "go.estuary.dev/W002: container did not specify port label, using stdio. Using stdio is deprecated and will be removed in the future.")
 
@@ -155,6 +152,7 @@ func Run(
 			// Mount the connector-proxy binary, as well as the output of inspecting the docker image.
 			"--mount", fmt.Sprintf("type=bind,source=%s,target=/flow-connector-proxy", tmpProxy.Name()),
 			"--mount", fmt.Sprintf("type=bind,source=%s,target=/image-inspect.json", tmpInspect.Name()),
+			"--interactive",
 		)
 	} else {
 		// Publish the tcp port of the container on a random port on host
--- a/go/connector/run.go
+++ b/go/connector/run.go
@@ -13,8 +13,10 @@ import (
 	"strings"
 	"sync"
 	"syscall"
+	"time"
 
 	"github.com/estuary/flow/go/flow/ops"
+	"github.com/estuary/flow/go/pkgbin"
 	"github.com/gogo/protobuf/proto"
 	"github.com/sirupsen/logrus"
 )
@@ -26,7 +28,19 @@ const (
 	Materialize
 )
 
-const DockerHostInternal = "host.docker.internal"
+const ConnectorTCPPortLabel = "FLOW_TCP_PORT"
+
+// proxyCommand returns a ProxyCommand of crates/connector_proxy/src/main.rs
+func (c Protocol) proxyCommand() string {
+	switch c {
+	case Capture:
+		return "proxy-flow-capture"
+	case Materialize:
+		return "proxy-flow-materialize"
+	default:
+		panic(fmt.Sprintf("go.estuary.dev/E100: unexpected protocol %v", c))
+	}
+}
 
 // Run the given Docker |image| with |args|.
 //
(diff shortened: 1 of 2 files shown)


go/connector/run.go
```diff
@@ -167,29 +142,55 @@ func runCommand(
...
+	// Give ready signal to connector, specifying the socket type and address
+	cmdStdin, err := cmd.StdinPipe()
+	if err != nil {
+		return fmt.Errorf("StdinPipe: %w", err)
+	}
+	_, port, err := net.SplitHostPort(socket.Addr().String())
+	if err != nil {
+		return fmt.Errorf("splitting socket address: %w", err)
+	}
+	cmdStdin.Write([]byte(fmt.Sprintf("tcp %s:%s\n", DockerHostInternal, port)))
```
williamhbaker, line 190: Would it be possible to provide this as an argument to the connector rather than sending it over `stdin`? I'm not sure if it's needed to be used as a "ready" signal here with the section above waiting to receive a connection on the socket before doing anything with it. Using an argument instead of `stdin` seems simpler to me, but not an absolute must-do (especially if it doesn't work).

williamhbaker:
A couple of general thoughts/questions:

- I would tend to think supporting connectors running both ways is going to be necessary (Do materialization connectors use this same code path? I am assuming they do). Maybe we could have the runtime query the container somehow before deciding how to interact with it - something simple like a docker run --entrypoint … on the image to check for the existence of a sentinel file that only connectors supporting sockets have - then branching off to the appropriate method to handle the interactions.

- I’m curious if you considered having the connector start the TCP socket and the runtime connect to that? That feels like it would be more intuitive to me, but I haven’t thought it through extensively and there might be all kinds of tricky problems it would create.

willdonnelly:
> [...] having the connector start the TCP socket and the runtime connect to that

Combined with Johnny's point last week about having a "Ready" message on connector stdout, it feels like there could be a coherent migration plan along the lines of:
1. Modify Flow to ignore a `READY ...\n` message from connector stdout.
2. Modify connectors to emit `READY stdio\n` at startup.
3. Add logic to Flow so that `READY tcp://1.2.3.4:5678\n` causes it to use gRPC over the specified socket instead of stdin/stdout.
4. Modify connectors to open a TCP socket and emit `READY tcp://[...]\n` to signal Flow to connect over TCP. Potentially also support for Unix sockets?
5. Optionally, remove support for the gRPC-over-stdio transport after it's not being used anywhere.

jgraettinger:
An issue with READY is that `docker run` should be in on the game: it has to be told explicitly to publish the port in question.

Another option, slightly less elegant but simpler (and more transferable to firecracker?), would be to use an image label which marks that the container supports TCP transport and identifies the specific port on which it'll be listening. The runtime looks for this label, port-forwards it in the `docker run` invocation and dials it with a retry backoff and reasonable timeout.

mdibaiee:
I'm not sure if I yet see the supporting reason for having the connector listen on a port rather than the runtime (other than it being more intuitive, which I personally didn't have the same sense).

williamhbaker:
> I'm not sure if I yet see the supporting reason for having the connector listen on a port rather than the runtime (other than it being more intuitive, which I personally didn't have the same sense).

One thought: I think it would make connector development & testing simpler. It would be nice to be able to start up a connector container and interact with it via its TCP socket, as opposed to needing to start a TCP server prior to starting the container, and then interacting with the connector through that TCP server.

jgraettinger:
Another reason is security model: we don't want to give untrusted code a capability to dial out to trusted code. The trusted code should dial into the untrusted sandbox.

mdibaiee:
Okay, now:
1. If a docker container includes the label `FLOW_TCP_PORT`, then communication is done through that port, with runtime port-forwarding that port, and dialing to it. One note here is: at the moment I'm port-forwarding to the same port on the runtime side, however I'm not sure if this is okay. Will there potentially be multiple connector runs in the same environment, in which case I'll port-forward to a random port, or is it safe to port-forward as-is?
2. If there is no such label, we continue using stdio for communication (hence all end-to-end tests passing)

There is still no network-tunnel binary, but with this we can start migrating simple airbyte connectors, and our own connectors by updating our libraries. We can meanwhile work on the network tunnel binary, and once that's in, migrate the advanced connectors which need the tunnel.

go/connector/run.go
```diff
@@ -167,29 +204,71 @@ func runCommand(
...
+			var err error
+			for {
+				conn, err = net.Dial("tcp", fmt.Sprintf("localhost:%s", port))
+				if err != nil {
+					if connectionCounter > 5 {
+						fe.onError(fmt.Errorf("dialing connection to port %s: %w", port, err))
+						return
+					} else {
+						// Retry after a short wait
+						connectionCounter = connectionCounter + 1
```
williamhbaker, line 235: Will this reliably be enough time, especially in cases where the connector image is being pulled for the first time? I wonder if we could keep a tiny bit of the stdout behavior to allow the connector to send a signal to the runtime on stdin when its listener is ready.

mdibaiee: @willdonnelly the connector image is pulled before this is invoked, this is invoked as `docker run` is executed.

williamhbaker: Ah yeah, I see where it is pulled earlier now. Still kind of wondering if this is going to be reliable, but if it isn't it won't be because the image has to be pulled for the first time.

mdibaiee: @williamhbaker do we want to indefinitely wait for signal from the connector stdout? I think just the fact that a TCP server starts listening on a specific agreed-on port, is a signal. Whether we have a timeout for waiting for ready signal from stdout, or a timeout for connecting to the agreed port, seems the same to me, unless I'm missing something.

williamhbaker: Good point, we'd need a timeout either way, so having a separate signal wouldn't accomplish anything. Thanks!

go/connector/run.go
```diff
@@ -118,25 +136,43 @@ func Run(
...
-		"--entrypoint", "/flow-connector-proxy",
-		// Mount the connector-proxy binary, as well as the output of inspecting the docker image.
-		"--mount", fmt.Sprintf("type=bind,source=%s,target=/flow-connector-proxy", tmpProxy.Name()),
-		"--mount", fmt.Sprintf("type=bind,source=%s,target=/image-inspect.json", tmpInspect.Name()),
+	}
+
+	if port != "" {
+		imageArgs = append(imageArgs,
+			"--publish",
+			fmt.Sprintf("%s:%s", port, port),
```
willdonnelly, line 144: Are we deliberately exposing this port outside the local machine, or should this maybe be `"localhost:%s:%s"` / `"127.0.0.1:%s:%s"`?

go/connector/run.go
```diff
@@ -167,29 +204,71 @@ func runCommand(
...
 		},
 	}
+
+	var conn net.Conn
+	// If port is specified, use TCP socket
+	if port != "" {
+		// Routine dialing a connection to connector
+		go func() {
+			// Try to connect to the connector with a retry mechanism
+			var connectionCounter = 0
```
willdonnelly, line 225: I'm concerned that the timeout of a `net.Dial` attempt is implicitly doing a lot of work in how this retry logic will behave. It seems like it might be a bit more consistent to instead compute a deadline here and retry until `time.Now().After(deadline)` so we always wait roughly the same amount of time for the connector to become reachable?

go/connector/run.go
```diff
@@ -167,29 +204,71 @@ func runCommand(
...
+			var err error
+			for {
+				conn, err = net.Dial("tcp", fmt.Sprintf("localhost:%s", port))
+				if err != nil {
```
willdonnelly, line 229: This retry loop might be a bit clearer to read if this were inverted to `if err == nil { break }`. This would allow the rest of the logic to be less nested and removes the need for an explicit `continue`.

go/connector/run.go
```diff
@@ -209,22 +288,25 @@ func runCommand(
 		}
 	}(cmd.Process.Signal)
 
-	err = cmd.Wait()
-	var closeErr = cmd.Stdout.(io.Closer).Close()
-	// Ignore error on closing stderr because it's already logged by the forwarder
+	var waitErr = cmd.Wait()
+	if waitErr != nil {
```
willdonnelly, line 292: nit: Maybe `if waitErr := cmd.Wait(); waitErr != nil {` since this error value isn't used past the body of this if statement?

willdonnelly (approved):
LGTM

jgraettinger:
> If a docker container includes the label FLOW_TCP_PORT, then communication is done through that port, with runtime port-forwarding that port, and dialing to it. One note here is: at the moment I'm port-forwarding to the same port on the runtime side, however I'm not sure if this is okay. Will there potentially be multiple connector runs in the same environment, in which case I'll port-forward to a random port, or is it safe to port-forward as-is?

🤔 there would be multiple connectors running, so we'll have to account for this.

Is it possible for us to **not** port-forward, but instead to have prior knowledge of the container name (with DNS resolution provided by docker) or of the container IP, which can be directly dialed at the advertised port?

mdibaiee:
> 🤔 there would be multiple connectors running, so we'll have to account for this.
>
> Is it possible for us to not port-forward, but instead to have prior knowledge of the container name (with DNS resolution provided by docker) or of the container IP, which can be directly dialed at the advertised port?

I will look into this to see if it's possible 🤔

go/connector/run.go
```diff
@@ -209,22 +288,25 @@ func runCommand(
...
-	} else if ctx.Err() == nil {
-		// Expect a clean exit if the context wasn't cancelled.
-		// Log the raw error, since we've already logged everything that was printed to stderr.
-		logger.Log(logrus.ErrorLevel, logrus.Fields{"error": err}, "connector failed")
-		fe.onError(fmt.Errorf("go.estuary.dev/E116: connector failed, with error: %w\nwith stderr:\n\n%s",
-			err, cmd.Stderr.(*connectorStderr).buffer.String()))
-	} else {
-		fe.onError(ctx.Err())
+	if port != "" {
+		var closeErr = conn.Close()
```
williamhbaker, line 306: Is there a case where `conn` is `nil` here where the connector is using a port but the connection can't be established within the timeout above?

williamhbaker (approved):

mdibaiee:
Thank you all for your comments,

1. I checked, it's not possible to connect to the container using its IP, the network is shared between containers, but not with the host (unless you do `--network=host` which is another story): https://stackoverflow.com/questions/38872328/connecting-to-docker-container-using-container-ip
2. So now, I request a random port to be assigned for port-forwarding, and then find that port after the container is run, this introduces another layer of retry mechanism where we wait for the container to run and tell us about its ContainerID, so we can ask for its port.
3. I address some of the other comments in cleaning up the code and addressing some edge cases, thanks for those 👍🏽

PR #690, opened 2022-09-23
feature: Schema inference service

### LINKED ISSUE #687: collection schema inference service
We'd like to extract the schema inference work that's currently done by individual connectors, into a service and API which offers up the inferred schema for a given collection of data on-demand.

jshearer:
**Description:**

This PR introdces the schema inference service, which is part of the work to move inference out of connectors, and into its own post-capture step. As per discussion with @jgraettinger, I put the actual service inside the already-existing `schema-inference` crate. The gRPC API exposed is pretty straightforward:

```protobuf
message InferenceRequest {
    string flow_collection = 1;
}

message InferenceResponse {
    oneof body {
        InferredSchema schema = 1;
    }
}

message InferredSchema {
    string schema_json = 1;
    uint64 documents_read = 2;
}

service InferenceService {
    rpc InferSchema (InferenceRequest) returns (InferenceResponse);
}
```

**Workflow steps:**
Screen Shot 2022-09-23 at 15 38 34

I'm told that the `flow-schema-inference` binary is already included in our docker image, so deploying this ought to be fairly straightforward.

Running on my local machine against a collection coming from the Hello World capture connector (so, all the same document shape), with some 300,000 records, this takes 8.3s to run. I think we could probably significantly reduce that if we parallelized by fragment, and not just journal, but I left that out for the moment. 

I could use a bit of your time @psFried figuring out the right way to test this. Should we... mock `journal-client`? Do something simpler?

Also need to (figure out how to) add this to the API exposed by data-plane-gateway

Fixes #687

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/690)


crates/schema-inference/src/server.rs
```diff
@@ -0,0 +1,220 @@
+use doc::inference::Shape;
```
jshearer, line 1: Does anyone know of a good tool to sort/organize these imports? I'd love something like [`import/order`](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/order.md) from ESLint, to e.g put external imports above workspace imports, above crate imports, etc

psFried: I'd advocate for using `rustmft` for this. We used to have it integrated into CI, and I think perhaps it just got lost at some point during a build refactor. But I think a sensible approach is to check in a `rustfmt.toml` and have CI just assert that everything is formatted correctly (and let devs wire up `rustfmt` into their workflows however they wish).

jshearer: Oh I see, rustfmt does this. I just had line breaks in my imports, so it was sorting them individually, not all together.

crates/schema-inference/src/server.rs
```diff
@@ -0,0 +1,224 @@
...
+
+    while let Some(maybe_doc_body) = doc_bytes_stream.next().await {
+        match maybe_doc_body {
+            Ok(doc_str) => {
+                let parsed: JournalMessage = serde_json::de::from_str(doc_str.as_str())
+                    .map_err(|e| Status::aborted(e.to_string()))?;
+                // There should probably be a higher-level API for this in `journal-client`
+
+                if parsed.meta.ack.is_none() {
+                    let re_serialized = to_value(parsed).unwrap();
```
psFried, line 159: This re-serialization could be mitigated by getting rid of the `JournalMessage` and `JournalMeta` structs entirely and just deserializing directly to a `serde_json::Value`. You can use the `Value::pointer` function to check whether `/_meta/ack` is true.

crates/schema-inference/src/server.rs
```diff
@@ -0,0 +1,224 @@
...
+        let selector = journal_selector(&flow_collection, None);
+        let journals = list_journals(&mut client, &selector).await?;
+
+        tracing::debug!("Got {} journals", journals.len());
+
+        // This lets us control parallelism on the journal axis.
+        // As it stands, we frequently only have one journal so it doesn't really matter,
+        // but it's a good proof-of-concept for splitting up `journal_to_shape` by fragments,
+        // where we'd want to do the same thing for much higher parallelism impact
+        let buffered = futures::stream::iter(journals)
```
psFried, line 70: I think it's fine to take this on in a subsequent PR, but TBH I don't really see any reason not to chunk up this work into individual fragments instead of journals. Most collections will only have a single journal.

crates/schema-inference/src/server.rs
```diff
@@ -0,0 +1,224 @@
...
+            .map(|j| journal_to_shape(j.clone(), client.clone()))
+            .buffer_unordered(3);
+
+        let shapes = buffered.collect::<Vec<_>>().await;
```
psFried, line 74: nit: I don't see a benefit to collecting into a `Vec` here. Seems preferable to just iterate the `buffered` stream directly.

psFried (changes requested):
This looks like a great start. I'm not really sure about the best approach to testing, but just thinking out loud it seems like a high level integration test, or even just manual testing might be a good place to start.

crates/schema-inference/src/server.rs
```diff
@@ -0,0 +1,224 @@
...
+        Ok(())
+    }
+}
+
+struct Authentication {
+    token: String,
+}
+
+fn require_auth(mut req: Request<()>) -> Result<Request<()>, Status> {
+    match req.metadata().clone().get("authorization") {
```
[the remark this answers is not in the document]
jshearer: Nice! Didn't know about that one

crates/schema-inference/src/server.rs
```diff
@@ -0,0 +1,224 @@
...
+    // TODO: Split this request up into parallelizable chunks.
+    // The reason we can't just split in to N even-sized chunks is that
+    // we could very easily split in the middle of a document.
+    // One option is to enumerate fragments in order to determine reasonable
+    // split points, and then use the Reader to read to those in parallel.
+
+    let reader = Reader::start_read(
+        client.clone(),
+        JournalRead::new(journal.name.clone()),
+        NoRetry,
```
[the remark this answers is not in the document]
jshearer: Good point, didn't realize that `ExponentialBackoff` had a specifiable limit. Does... 3 sound good?

crates/schema-inference/src/server.rs
jshearer: Right, so something like this?
```rust
let mut buffered = futures::stream::iter(journals)
            .map(|j| journal_to_shape(j.clone(), client.clone()))
            .buffer_unordered(3);

while let Some(shape) = buffered.next().await {
..
}
```

My question about that is, will that actually run multiple futures in parallel? It feels a bit weird to be awaiting them one at a time, but I kind of think it actually will pre-execute them because of `.buffer_unordered`

crates/schema-inference/src/server.rs
```diff
@@ -0,0 +1,224 @@
...
+    );
+
+    let codec = LinesCodec::new(); // do we want to limit length here? LinesCodec::new_with_max_length(...) does this
```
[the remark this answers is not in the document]
jshearer: Done! https://github.com/estuary/flow/blob/1644779d83fc81b0cde2bd64eb7371905702a944/crates/schema-inference/src/json_decoder.rs

crates/schema-inference/src/server.rs
jshearer: Done! https://github.com/estuary/flow/pull/690/files#diff-8a4f203851e6710c1af18f044b54a78c49ddb6eedaecf6aa5d4804fdf2c58187R139

crates/schema-inference/src/server.rs
jshearer: Seems to be fine, done!

jshearer:
Addressed all review feedback:

* Inference is now parallel across fragments as well as journals
* Inference now runs until a deadline, provided by a CLI flag defaulting to 10s. We gracefully terminate reading all fragments, and return the `Shape` that we've inferred so far.
* Brought in our own `JsonCodec` based on [`asynchronous-codec`](https://github.com/mxinden/asynchronous-codec/blob/master/src/codec/json.rs) (but updated to use `tokio_util`'s `Decoder` and not their own bespoke one... lol) which uses serde's ability to incrementally read JSON from a `&[u8]`.
  **Question: ** Is this the right way to go? I'm worried that this implementation may read the entire fragment file into memory.. or does `FramedRead` handle buffering nicely for us?

crates/schema-inference/src/json_decoder.rs
```diff
@@ -0,0 +1,75 @@
...
+
+/// Decoder impl parses json objects from bytes
+impl<Dec> Decoder for JsonCodec<Dec>
+where
+    for<'de> Dec: Deserialize<'de> + 'static,
+{
+    type Item = Dec;
+    type Error = JsonCodecError;
+
+    fn decode(&mut self, buf: &mut BytesMut) -> Result<Option<Self::Item>, Self::Error> {
```
psFried, line 54: I actually thought that this was fine as it is, but I did a little more digging since you'd asked about it, and discovered that there is an important missing optimization. The [`Decoder` docs](https://docs.rs/tokio-util/latest/tokio_util/codec/trait.Decoder.html#tymethod.decode) seem to indicate that the `FramedRead` may not do some sub-optimal bufffer management unless the `Decoder` explicitly reserves more space in the buffer. Looking at the [framed read source](https://github.com/tokio-rs/tokio/blob/master/tokio-util/src/codec/framed_impl.rs#L216) shows that "sub-optimal" is quite an understatement! So I think we just need to grow the buffer here if we're returning `None` and we're not at EOF.

Given that we're taking on buffer management here is also making me reconsider my previous position on whether we should impose a limit on the maximum document/buffer size. It's seeming like a maximum size may be a good idea. I'll propose something rather absurdly large, like 128MiB, and we can defer making a decision on any sort of global document size limit. If for some reason a document goes over that size, then schema inference can fail :man_shrugging:

cc @jgraettinger, since he may have a more informed opinion on a max document size.

crates/schema-inference/src/server.rs
```diff
@@ -0,0 +1,277 @@
...
+}
+
+struct Authentication {
+    token: String,
+}
+
+fn require_auth(mut req: Request<()>) -> Result<Request<()>, Status> {
+    match req.metadata().get("authorization").cloned() {
+        Some(t) /* TODO: validate JWT token */ => {
+            req.extensions_mut().insert(Authentication {
```
psFried, line 266: Is the idea for this server to call the data-plane-gateway (which would enforce auth), or the other way around? Or is that still undecided? The passing along of the token makes it seem like this service would delegate to DPG for the reads, but then that sort of implies that we have a separate ingress for this service, which also kind of doesn't make sense. I vaguely recall there being some conversation on this topic, so apologies if I've forgotten.

crates/schema-inference/src/server.rs
```diff
@@ -0,0 +1,277 @@
...
+
+        let InferenceRequest {
+            flow_collection: collection_name,
+        } = request.get_ref();
+
+        let (abort_tx, abort_rx) = tokio::sync::broadcast::channel::<()>(1);
+
+        let handle = tokio::spawn(async move {
+            sleep(Duration::from_secs(10)).await;
+            abort_tx.send(()).unwrap();
```
psFried, line 59: Doesn't seem like this a safe `unwrap`. Seems like it's possible for all the `abort_rx` receivers to have been dropped prior to this call, for example if they all finish or error out prior to the timeout expiring. It might happen to work, depending on exactly when the `abort_rx` gets dropped, but it seems pretty tricky to rely on that. I think we can just ignore the error here.

psFried (changes requested):
I think this is looking _much_ better. The majority of my comments here are non-blocking, but I did manage to spot a few things that seem worth changing.

crates/journal-client/src/fragments.rs
```diff
@@ -49,4 +50,18 @@ impl FragmentIter {
         }
         next.map(|r| Ok(r))
     }
+
+    pub fn as_stream<'a>(&'a mut self) -> impl TryStream<Ok = broker::fragments_response::Fragment, Error = Error> + 'a {
```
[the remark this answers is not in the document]
jshearer: I haven't been living in Rust land to intuit the consequences of that. My very vague understanding was to stay away from `'static` because it provides less flexibility, but I don't actually know why that's true or when to apply that rule.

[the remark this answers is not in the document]
jshearer: Done

crates/schema-inference/src/main.rs
```diff
@@ -36,32 +65,19 @@ impl AnalyzeArgs {
...
 }
+#[tokio::main]
+async fn main() -> Result<(), anyhow::Error> {
+    let subscriber = tracing_subscriber::FmtSubscriber::builder()
+        .with_env_filter(
+            EnvFilter::builder()
+                .with_default_directive(LevelFilter::INFO.into())
+                .from_env_lossy(),
+        )
+        // .with_writer(std::io::stderr) // Log to stderr since `analyze` uses stdout
```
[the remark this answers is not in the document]
jshearer: I have no idea why this was commented out. Removed since stderr I believe is the default

crates/schema-inference/src/server.rs
```diff
@@ -0,0 +1,277 @@
...
+        tracing::debug!("Starting inference");
+
+        let InferenceRequest {
+            flow_collection: collection_name,
+        } = request.get_ref();
+
+        let (abort_tx, abort_rx) = tokio::sync::broadcast::channel::<()>(1);
+
+        let handle = tokio::spawn(async move {
+            sleep(Duration::from_secs(10)).await;
```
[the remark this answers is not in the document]
jshearer: derp

crates/schema-inference/src/json_decoder.rs
jshearer: I made it grow by 1mb every time it fails to decode a document without reaching EOF or an error

crates/schema-inference/src/server.rs
jshearer: Welll... I am also not quite sure. This token is passed to `connect_journal_client`, which ultimately is used to set an `AuthHeader` interceptor on the `JournalClient` that gets created, so I assumed it was needed.

psFried: I'm thinking it would be better to have data-plane-gateway handle authz and delegate to schema-inference, which would connect to gazette directly (and thus not need the authorization header). It certainly won't hurt for the authorization header to be present, but I don't think it should be _required_.

That being said, IDK how this all squares with the potentially changing view of this being framed as a runtime concern, similar to how stats are handled. We should discuss that soon, but I don't think it should block merging this PR.

psFried (approved):
I left a few more comments, but this should be fine to merge once those are addressed.

Issue #655, opened 2022-09-01
Materializations fail when materializing from collections which have other failed materializations


williamhbaker:
Trying to create a new materialization from a collection that has a totally different materialization materializing from it seems to fail if that _other_ materialization is in a failed state. It makes sense for changes that would cause something else to break to get rejected, but in this case a materialization is unable to be published (fails tests) because of the status of a totally separate materialization using the same collection.

The code responsible for this is believed to be [here](https://github.com/estuary/animated-carnival/blob/main/crates/agent-sql/src/publications.rs#L226-L287). Ideally we would be able to create a materialization from a collection regardless of wether that collection happens to have separate failed materializations from it.

PR #692, opened 2022-09-26
agent-sql: fix flow expansion to captures & materializations

williamhbaker:
Closes https://github.com/estuary/flow/issues/655

**Description:**

Modifies the the flow expansion behavior when doing a build to not include unneeded captures or materializations. This prevents cases where (for example) an existing materialization failing to validate will prevent any new materializations being created from a shared collection. It may also provide a benefit to how long it takes to do a build in certain cases.

As far as I can tell this matches the intent of the original expansion, per the comments in existing the code.

**Workflow steps:**

In addition to the new unit tests, I ran a scenario locally where I:
- Made a single collection
- Create a Postgres materialization from that collection
- Allowed that materialization to successfully start & run, but then deleted the database so the materialization would fail subsequent validation checks
- Verified that with the existing code, a _new_ materialization from that collection could not be created, since the existing one fails to validate
- Verified that with the new code, a new materialization _could_ be created from that same collection, even with the existing materialization not being able to validate (since it isn't included in the build).

N/A

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/692)

### DIFF
--- a/crates/agent-sql/src/publications.rs
+++ b/crates/agent-sql/src/publications.rs
@@ -244,7 +244,7 @@ pub async fn resolve_expanded_rows(
         -- Expand seed collections to include bound captures & materializations.
         -- The "on" clause ensures we only expand if the seed spec is itself a collection.
         -- The "union" is included here to retain the seed spec if it is a capture or materialization already.
-        -- This has the side effect of including the seed spec if is a collection as well, so the seed spec does
+        -- This has the side effect of including the seed spec if it is a collection, so the seed spec does
         -- not need to be union'd into the pass_two selection subquery.
         -- This pass is non-recursive: The "union" does not involve pass_one_b recursively.
         pass_one_b(id) as (
--- a/crates/agent-sql/src/publications.rs
+++ b/crates/agent-sql/src/publications.rs
@@ -241,20 +241,21 @@ pub async fn resolve_expanded_rows(
             from seeds s join live_spec_flows e
             on s.id = e.source_id and e.flow_type = 'collection'
         ),
-        -- Expand seed collections, captures, and materializations through
-        -- edges that connect captures and materializations to their bound
-        -- collections:
-        --   * A seed capture expands to all bound collections.
-        --   * A seed materialization expands to all bound collections.
-        --   * A seed collection expands to captures or materializations which bind it,
-        --      and from there to all of its other bound collections.
+        -- Expand seed collections to bound captures & materializations.
+        -- This pass is non-recursive.
         pass_one_b(id) as (
-            select id from seeds
-          union
+            select case when s.id = e.source_id then e.target_id else e.source_id end
+              from seeds as s join live_spec_flows as e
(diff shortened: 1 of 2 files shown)


crates/agent-sql/src/publications.rs
```diff
@@ -241,28 +241,36 @@ pub async fn resolve_expanded_rows(
...
             on s.id = e.source_id and e.flow_type = 'collection'
         ),
-        -- Expand seed collections, captures, and materializations through
-        -- edges that connect captures and materializations to their bound
-        -- collections:
-        --   * A seed capture expands to all bound collections.
-        --   * A seed materialization expands to all bound collections.
-        --   * A seed collection expands to captures or materializations which bind it,
-        --      and from there to all of its other bound collections.
+        -- Expand seed collections to include bound captures & materializations.
```
jgraettinger, line 244: This seems ready for a deeper refactor to make it more readable and verifiable. The `pass_*` naming (which I started tbf) is not helping.

What about introducing `bound_captures` & `bound_materializations`?
 * `bound_captures` is sources which have a target in `seeds` with flow_type `capture`.
 * `bound_materializations` is targets which have a source in `seeds` with flow_type `materialization`.

Then, a second stage `bound_collections` would be the union of:
 * IDs which have a target in `bound_captures` union `seeds` with flow_type `capture`.
 * IDs which have a source in `bound_materializations` union `seeds` with flow_type `materialization`.
 * IDs which have a source in `seeds` with flow_type `collection` (this is current `pass_one_a`).

Then, `pass_two` becomes `backprop_derivations_and_tests` or similar and uses `bound_collections` as its base-case.

jgraettinger (commented):
See below comments. AFAIK this change is correct but given its foundational importance I think it's worth spending a bit more time improving the factoring and test coverage. Thanks!

crates/agent-sql/tests/expansions.rs
```diff
@@ -119,6 +119,63 @@ async fn test_materialization_expansions() {
...
+async fn test_shared_collection_expansions() {
+    let mut conn = sqlx::postgres::PgConnection::connect(&FIXED_DATABASE_URL)
+        .await
+        .expect("connect");
+
+    let mut txn = conn.begin().await.unwrap();
+
+    sqlx::query(
+        r#"
+    with specs as (
```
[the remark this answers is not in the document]
williamhbaker: Added some more tests, good idea!

crates/agent-sql/src/publications.rs
williamhbaker: I took a shot at refactoring along the lines you suggested. I do think it is more clear now. I found that I needed to union the `bound_captures`, `bound_materializations`, `bound_captures`, and `seeds` together for the `backprop_derivations_and_tests` base case. The non-recursive expansions don't include their initial elements, so they have to be added back in somehow before doing the recursive expansion.

jgraettinger (approved):
Looks awesome! Thanks

PR #698, opened 2022-09-27
Set flowctl version from FLOW_VERSION

psFried:
**Description:**

Uses the `FLOW_VERSION` env variable to set the version string used by flowctl. This makes it consistent with the rest of the build, and allows non-release builds to use the version that's determined by the Makefile.

Also specifies the rust toolchain for flowctl release CI jobs.

_In theory_, this should get flowctl releases working, but I'll have to test that after this is merged.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/698)

### DIFF
--- a/.github/workflows/flowctl-release.yaml
+++ b/.github/workflows/flowctl-release.yaml
@@ -12,7 +12,7 @@ jobs:
     env:
       ASSET_NAME: ${{ matrix.config.assetName }}
       # build.rs reads this env variable and uses to set the value that's printed by flowctl --version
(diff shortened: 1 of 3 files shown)


williamhbaker (approved):
LGTM. There may be some things to sort out if we ever want to do separate `flowctl` vs. `flow` releases.

psFried:
> There may be some things to sort out if we ever want to do separate flowctl vs. flow releases.

Indeed. At the moment, we don't really have releases of anything other that flowctl. But there's definitely going to be some stuff to sort out once we do want to introduce those other release artifacts.

PR #700, opened 2022-09-27
Use setup-protoc action with authentication in CI

psFried:
**Description:**

Previously, flow builds would install protoc using `apt`, while `flowctl` builds would use the `setup-prococ` action. This standardizes on the `setup-protoc` action because it supports multiple platforms.

Further, the `setup-protoc` action was previously failing due to rate limit errors from the GH API (it uses the API in order to fetch release versions). This commit adds the `repo-token` variable, which allows for a much higher rate limit, since the request is authenticated.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/700)

### DIFF
--- a/.github/workflows/flowctl-release.yaml
+++ b/.github/workflows/flowctl-release.yaml
@@ -24,6 +24,10 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v2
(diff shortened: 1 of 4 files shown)


williamhbaker (approved):
LGTM

PR #701, opened 2022-09-28
connector/run: find port before running docker

mdibaiee:
**Description:**

- I just saw this error on production (intermittent):
```
-- time="2022-09-28T16:43:50Z" level=fatal msg="fatal error" err="inspecting container port failed: go.estuary.dev/E111: getting docker entity \"\"'s port failed: exit status 1"
```
Which means `ReadFile` succeeded, but was empty. So I guess it takes some time in between the creation of the file and writing of the content in it by docker.

- I also saw this error (also intermittent):
```
-- time="2022-09-28T16:47:00Z" level=fatal msg="fatal error" err="inspecting container port failed: go.estuary.dev/E111: getting docker entity \"64ab7e6238c7120e31217107a536a0a7cd1f98c127609ece72f806a1ed2cc354\"'s port failed: exit status 1"
```
which similarly means the docker port command may need to wait until the container is in some sort of ready state. So I added retry logic there.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/701)

### DIFF
--- a/go/connector/run.go
+++ b/go/connector/run.go
@@ -10,6 +10,7 @@ import (
 	"net"
 	"os"
 	"os/exec"
+	"strconv"
 	"strings"
 	"sync"
 	"syscall"
@@ -110,13 +111,15 @@ func Run(
 		networkName = "bridge"
 	}
 
-	// Docker --cidfile expects that the file does not exist in the provided path, so we create
-	// a temporary file and delete it just so we can use the allocated path
-	cidTmp, err := os.CreateTemp("", "connector-id")
-	if err != nil {
-		return fmt.Errorf("go.estuary.dev/E133: creating temporary file for CID: %w", err)
-	} else if err := os.Remove(cidTmp.Name()); err != nil {
-		return fmt.Errorf("go.estuary.dev:E134: cleaning temporary file for CID: %w", err)
+	// Find a free port on local system
+	var localAddress string
+	if port != "" {
+		localPort, err := GetFreePort()
+		if err != nil {
+			return fmt.Errorf("go.estuary.dev/E133: could not get a free port on host: %w", err)
+		}
+
+		localAddress = fmt.Sprintf("127.0.0.1:%s", localPort)
 	}
 
 	var imageArgs = []string{
@@ -128,8 +131,6 @@ func Run(
 		"--init",
 		// Remove the docker container upon its exit.
 		"--rm",
-		// Write the container ID to a temporary file
-		"--cidfile", cidTmp.Name(),
 		// Tell docker not to persist any container stdout/stderr output.
 		// Containers may write _lots_ of output to std streams, and docker's
 		// logging drivers may persist all or some of that to disk, which could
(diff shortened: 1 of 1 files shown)


willdonnelly (approved):
LGTM

go/connector/run.go
```diff
@@ -293,7 +293,7 @@ func runCommand(
 		var cidFileReadDeadline = time.Now().Add(time.Second * 5)
 		for {
 			idBytes, err := os.ReadFile(cidFilePath)
-			if err == nil {
+			if err == nil && len(idBytes) > 0 {
```
williamhbaker, line 296: Is there a way to pick a random port on the host and then explicitly map that above on line 160 when the container is started? That seems like it would simplify needing to query/retry the container to see which port has been randomly chosen.

mdibaiee: @williamhbaker I've been thinking about this too... I just found this: https://github.com/phayes/freeport

This could be a great simplification to be honest, I don't like what is happening here with all of these retries, etc... 🤔 
I'll try that

mdibaiee:
@williamhbaker @willdonnelly this simplifies the code _a lot_, thanks for the push @williamhbaker

mdibaiee:
I’m just thinking, there might now be a small chance for race condition if two connectors are going to be run at the same time, is it possible that they end up finding the same port and try to run it on the same port? Is that a risk we can allow or not?

willdonnelly:
What happens if two connectors randomly select the same port? My intuition is that one would successfully bind the port, the other would be rejected and would then shut down, and 5 minutes later it would try running again and randomly pick a different port.

mdibaiee:
> What happens if two connectors randomly select the same port? My intuition is that one would successfully bind the port, the other would be rejected and would then shut down, and 5 minutes later it would try running again and randomly pick a different port.

That's in case of an already running container, but in case of running `spec` or `discovery` requests, etc., I guess it means the user will see an error and will have to try again. If this is rare enough (which I think it is), then we may accept that

willdonnelly:
I'm inclined to not worry about that. The space of port numbers is ~50k elements, so back-of-the-envelope the chance of two concurrent connectors startups picking the same port is around 0.002%. I'm not sure how many concurrent `spec` and `discover` instances we generally have running, but I suspect it's not a big deal.

williamhbaker:
I think it's even less likely than one might think that two containers would select the same port in this case. On my system, repeated calls to `GetFreePort` from the same process yield incrementally higher ports, so it would have to "wrap around" to start getting any overlap. There might even be some jitter to how it steps through them, since they aren't sequential, but are always increasing, for example:
```
56200
56208
56211
56212
56213
56216
56222
56225
56228
56229
56232
```

mdibaiee:
Ok great, let's stick with this then. Let me know if the PR looks good to you overall

williamhbaker (approved):

PR #703, opened 2022-09-29
Fix add bootstrap steps

dhammika:
**Description:**

Adds build tools and bootstrap steps to readme doc.

**Workflow steps:**

N/A

N/A

**Notes for reviewers:**

This is based on ubuntu 20.04 LTS (focal) base image.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/703)

### DIFF
--- a/control_plane_readme.md
+++ b/control_plane_readme.md
@@ -138,18 +138,26 @@ cd postgresql-14.5 | ./configure --prefix=~/flow-pg | make install
 
 Apply init schema:
 ```console
-curl -OL https://raw.githubusercontent.com/supabase/supabase/master/docker/volumes/db/init/00-initial-schema.sql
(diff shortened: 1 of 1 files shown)


control_plane_readme.md
```diff
@@ -98,6 +125,41 @@ Directly access your postgres database:
 psql postgres://postgres:postgres@localhost:5432/postgres
 ```
 
+### EXPERIMENTAL Start private PG (instead of Supabase for local development):
+
+Install and start pg:
```
psFried, line 130: My 2 cents is that it's easier to manage local postgres databases using docker. I don't think it really matters how you install postgres, though, so I don't think it would hurt anything to have this section :man_shrugging:

psFried (approved):
Looks good to me. Thanks!

control_plane_readme.md
dhammika: That was to run it without docker. Anyways, dropped pg install steps in the patch below.

PR #704, opened 2022-09-29
devcontainer and release fixes

skord:
**Description:**

One cannot use the devcontainer locally on M1 at this moment, nor build a release image. 

This PR:
* Installs git-lfs in the devcontainer that works for supported buster platforms instead of just amd64. 
* Upgrades rust in the devcontainer to 1.64.0
* Upgrades golang to the latest version in a way that supports many platforms.
* Installs docker-ce-cli from a repository so it supports more platforms

I've tested this locally on arm64/M1.

**Workflow steps:**

It should work, after not working before. 

None that I'm aware of. 

**Notes for reviewers:**

It might behoove a review who has a linux workstation to try and run `make docker-image` and also try running the devcontainer locally.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/704)

### DIFF
--- a/.devcontainer/github-ubuntu-2004.Dockerfile
+++ b/.devcontainer/github-ubuntu-2004.Dockerfile
@@ -14,48 +14,51 @@ ARG LOCALE=en_US.UTF-8
 # See the package list in the GitHub reference link above, at the very bottom,
 # which lists installed apt packages.
 RUN apt update -y \
- && apt upgrade -y \
- && DEBIAN_FRONTEND=noninteractive apt install --no-install-recommends -y \
-      bash-completion \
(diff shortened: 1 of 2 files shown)


mdibaiee (approved):

mdibaiee:
To clarify, I don't have a linux machine to try what you suggested, so might want to wait for that:

> It might behoove a review who has a linux workstation to try and run make docker-image and also try running the devcontainer locally.

skord:
I was able to get this working mostly on my Fedora laptop, so that question is answered. To be clear, this does not address that the build doesn't work properly in the dev container, but that the dev container builds at all.

PR #707, opened 2022-09-30
Phil/merged flowctl reads

psFried:
**Description:**

This PR rolls up a few big changes to `flowctl` from the animated-carnival repo. Apologies for the size of it. The main feature that this adds is the ability to read collection data from the flow data plane, using the new `journal-client` crate. Additionally, I refactored how the CLI handles outputting results to stdout. That was done to allow sharing some common code for outputting tables, to allow outputting data in JSON or YAML format, and because I just couldn't help myself.

The presentation of the subcommands here is centered around the concept of Flow collections, rather than Gazette journals. I _think_ this is the right approach, but I'd like to explicitly invite people to discuss disagreements with this approach.

**Workflow steps:**

The complete list of subcommands that were added:

### `flowctl collections`

```
flowctl-collections 
Work with Flow collections

USAGE:
    flowctl collections [OPTIONS] <SUBCOMMAND>

OPTIONS:
    -h, --help               Print help information
    -o, --output <OUTPUT>    How to format CLI output [possible values: json, yaml, table]

SUBCOMMANDS:
    help              Print this message or the help of the given subcommand(s)
    list-fragments    List the journal fragments of a flow collection
    list-journals     List the individual journals of a flow collection
    read              Read data from a Flow collection and output to stdout
```

Note that the `read` subcommand currently only supports reading uncommitted data from a single journal. The plan is to remove those limitations in the future, as we build out more features in `journal-client`. For the time being, though, the command will return an error if you provide a selector that matches more than one journal. The command also _requires_ that you pass the `--uncommitted` argument. This is because the future plan would be to make committed reads the default, but to still allow reading uncommitted data with that argument.

One other caveat is that the `--output` argument cannot currently be used to output collection data in a format other than `json`. Again, the intent here is to eventually support the range of output formats, but doing so requires some additional work in `journal-client` to support framed message-based journal reads.

### `flowctl logs`

```
flowctl-logs 
Prints the runtime logs of a task (capture, derivation, or materialization)

USAGE:
    flowctl logs [OPTIONS] --task <TASK>

OPTIONS:
        --follow             Whether to block waiting for new data. If provided, the read will
                             continue indefinitely until interrupted or ending due to an error
    -h, --help               Print help information
    -o, --output <OUTPUT>    How to format CLI output [possible values: json, yaml, table]
        --since <SINCE>      Start reading from approximately this far in the past. For example
                             `--since 10m` will output all data that was added within the last 10
                             minutes. The actual start of the read will always be at a fragment
                             boundary, and thus may include data from significantly before the
                             requested time period
        --task <TASK>        Read the logs of the task with the given name
```

The `logs` subcommand is really just a bit of sugar for calling `collections read` to read data out of the `ops/$tenant/logs` collection. Note that there is no `--uncommitted` argument because log messages are always implicitly committed anyway.

### `flowctl stats`

```
flowctl-stats 
Prints the runtime stats of a task (capture, derivation, or materialization)

USAGE:
    flowctl stats [OPTIONS] --task <TASK>

OPTIONS:
        --follow             Whether to block waiting for new data. If provided, the read will
                             continue indefinitely until interrupted or ending due to an error
    -h, --help               Print help information
    -o, --output <OUTPUT>    How to format CLI output [possible values: json, yaml, table]
        --since <SINCE>      Start reading from approximately this far in the past. For example
                             `--since 10m` will output all data that was added within the last 10
                             minutes. The actual start of the read will always be at a fragment
                             boundary, and thus may include data from significantly before the
                             requested time period
        --task <TASK>        Read the logs of the task with the given name
        --uncommitted        Read raw data from stats journals, including possibly uncommitted or
                             rolled back transactions. This flag is currently required, but will be
                             made optional in the future as we add support for committed reads,
                             which will become the default
```

Similar to `logs`, `stats` just delegates to `collections read` to pull data out of `ops/$tenant/stats` collections. Stats also require the `--uncommitted` flag for the time being.

## `--output`

The TLDR is that now all commands (except for reading collection data) support the same set of output formats, so you can choose to output `catalog list`, for example, as JSON or YAML in addition to the default table format. This refactor was born out of a desire to have some common code for leveraging the `serde::Serialize` trait and projections for outputting tables, and also from a desire to output things in JSON format so that I could pipe the output through `jq`. It also seemed like a fairly approachable refactor now, but that it would become far less approachable with every added bespoke table ouput implementation.

- https://docs.estuary.dev/reference/working-logs-stats/
- https://docs.estuary.dev/concepts/flowctl/

There might be some other places in docs where it could be help

**Notes for reviewers:**

Commit by commit is probably the easiest way to review. Tests are still missing, but my working plan is to work towards using flowctl for driving e2e test scenarios, which should result in some decent coverate.


jshearer (approved):
I read through all of this, and don't see anything that stands out to me as needing to be changed. The new commands/logic for plumbing through `journal-client` are all quite easy to follow and are nicely documented, and the rest of the changes appear to be around improving command output, which also makes sense. LGTM!

PR #717, opened 2022-10-13
fix: e2e tests failing because of conflicting ports

jshearer:
Okay, so after talking with Johnny and Phil about this earlier, I went on to keep debugging this.
* We figured out the correct hostnames and ports in order to configure the connector’s `sshForwarding` to the postgres container. We got hung up on the fact that the connector binds to port 2222 by [default](https://github.com/estuary/connectors/blob/main/materialize-boilerplate/boilerplate.go#L35), which is _also_ the port that the `openssh-server` container uses. After figuring this out, we moved us on to the next error: 
* `ssh: bind [::1]:16666: Cannot assign requested address`. I found [this blog  post](https://www.electricmonk.nl/log/2014/09/24/ssh-port-forwarding-bind-cannot-assign-requested-address/) which suggested this is actually about ipv6 (note the `[::1]` instead of `127.0.0.1`). I couldn’t figure out an easier way to change this, so I [added](https://github.com/estuary/flow/pull/717/files#diff-dea4484ed61c1ed0c3f3fdf7bd093db8fd240d9caea93844456b67dea1101119R168-R171) `-o AddressFamly inet` to the ssh invocation, which seems to have worked, revealing the next  layer of this onion, which is also where I’m gonna call it for tonight:
* `msg=\"failed to query materialization spec (the table may not be initialized?)\" err=\"ERROR: relation \\\"flow_materializations_v2\\\" does not exist (SQLSTATE 42P01)\" table=flow_materializations_v2"` This seems to have gone away, not sure what caused it
* Needed to change how ssh forwarding is configured, thanks @mdibaiee.

We got a ✅!

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/717)

### DIFF
--- a/crates/network-tunnel/src/networktunnel.rs
+++ b/crates/network-tunnel/src/networktunnel.rs
@@ -6,21 +6,22 @@ use async_trait::async_trait;
 
 #[async_trait]
 pub trait NetworkTunnel: Send + Sync {
-    // Inspect and/or modify the endpoint spec for which this network tunnel is created.
-    // This takes place before the tunnel is started, and may also modify tunnel config
-    // if appropriate.
+    /// Inspect and/or modify the endpoint spec for which this network tunnel is created.
+    ///
+    /// This takes place before the tunnel is started, and may also modify tunnel config
+    /// if appropriate.
     fn adjust_endpoint_spec(
         &mut self,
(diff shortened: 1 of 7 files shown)


jgraettinger (approved):
LGTM 🚢 🎉

PR #718, opened 2022-10-13
Add cnt names

dhammika:
**Description:**

Adds shardIDs as connector container names. 

**Workflow steps:**
N/A

N/A

**Notes for reviewers:**
N/A

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/718)

### DIFF
--- a/go/bindings/build.go
+++ b/go/bindings/build.go
@@ -44,6 +44,7 @@ type CaptureDriverFn func(
 	endpointType pf.EndpointType,
 	endpointSpec json.RawMessage,
 	connectorNetwork string,
+	containerName string,
 	logger ops.Logger,
 ) (pc.DriverClient, error)
 
@@ -54,6 +55,7 @@ type MaterializeDriverFn func(
 	endpointType pf.EndpointType,
 	endpointSpec json.RawMessage,
 	connectorNetwork string,
+	containerName string,
 	logger ops.Logger,
 ) (pm.DriverClient, error)
 
@@ -69,6 +71,8 @@ type BuildArgs struct {
 	CaptureDriverFn CaptureDriverFn
 	// Builder of materialization DriverClients
 	MaterializeDriverFn MaterializeDriverFn
+	// Build container name
+	ContainerName string
 }
 
 // BuildCatalog runs the configured build.
@@ -139,7 +143,7 @@ func BuildCatalog(args BuildArgs) error {
 				log.WithField("request", request).Debug("capture validation requested")
 
 				var driver, err = args.CaptureDriverFn(ctx, request.EndpointType,
-					request.EndpointSpecJson, args.BuildAPI_Config.ConnectorNetwork, ops.StdLogger())
+					request.EndpointSpecJson, args.BuildAPI_Config.ConnectorNetwork, args.ContainerName, ops.StdLogger())
 				if err != nil {
 					return nil, fmt.Errorf("driver.NewDriver: %w", err)
 				}
@@ -172,7 +176,7 @@ func BuildCatalog(args BuildArgs) error {
 				log.WithField("request", request).Debug("materialize validation requested")
 
 				var driver, err = args.MaterializeDriverFn(ctx, request.EndpointType,
-					request.EndpointSpecJson, args.BuildAPI_Config.ConnectorNetwork, ops.StdLogger())
+					request.EndpointSpecJson, args.BuildAPI_Config.ConnectorNetwork, args.ContainerName, ops.StdLogger())
 				if err != nil {
 					return nil, fmt.Errorf("driver.NewDriver: %w", err)
 				}
--- a/go/bindings/build.go
+++ b/go/bindings/build.go
@@ -141,7 +141,7 @@ func BuildCatalog(args BuildArgs) error {
 				log.WithField("request", request).Debug("capture validation requested")
 
 				var driver, err = args.CaptureDriverFn(ctx, request.EndpointType,
-					request.EndpointSpecJson, args.BuildAPI_Config.ConnectorNetwork, "", ops.StdLogger())
+					request.EndpointSpecJson, args.BuildAPI_Config.ConnectorNetwork, request.Capture.String(), ops.StdLogger())
 				if err != nil {
 					return nil, fmt.Errorf("driver.NewDriver: %w", err)
 				}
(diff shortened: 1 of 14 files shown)


go/bindings/build.go
```diff
@@ -69,6 +71,8 @@ type BuildArgs struct {
 	CaptureDriverFn CaptureDriverFn
 	// Builder of materialization DriverClients
 	MaterializeDriverFn MaterializeDriverFn
+	// Build container name
+	ContainerName string
```
psFried, line 75: I don't think `ContainerName` can be provided on `BuildArgs`. The problem is that a build may actually start _many_ containers, and they of course can't all have the same name. Builds will perform a `Validate` RPC for each capture and materialization that's involved in the build. There can be many of those, as a build can include arbitrary subsets of all the tasks in the control plane.

I think the solution here is to add the `ContainerName` parameter to the `CaptureDriverFn` and `MaterializeDriverFn`. Those functions each return a driver (backed by a running container) for a specific connector, and the task name, if applicable, should be known when those functions are called.

go/bindings/build.go
```diff
@@ -139,7 +143,7 @@ func BuildCatalog(args BuildArgs) error {
 				log.WithField("request", request).Debug("capture validation requested")
 
 				var driver, err = args.CaptureDriverFn(ctx, request.EndpointType,
-					request.EndpointSpecJson, args.BuildAPI_Config.ConnectorNetwork, ops.StdLogger())
+					request.EndpointSpecJson, args.BuildAPI_Config.ConnectorNetwork, args.ContainerName, ops.StdLogger())
```
psFried, line 146: The name of the capture being validated here is `request.Capture`. I'm thinking this is where it could call a function that accepts the task and operation names, and returns a container name to pass to `CaptureDriverFn`

go/connector/run.go
```diff
@@ -145,6 +146,16 @@ func Run(
 		"--network", networkName,
 	}
 
+	if containerName != "" {
+		// Replace non-alphanum chars.
+		// Container names must match [a-zA-Z0-9][a-zA-Z0-9_.-]+.
+		// FIXME handle names prefixed with [_.-+/]+.
+		containerName = strings.NewReplacer("+", "_", "/", "_").Replace(containerName)
```
psFried, line 153: I'm realizing that there could be name collisions here. You could have two different Flow tasks, `foo/bar/baz` and `foo/bar_baz`, and both of those would result in the same `containerName`.

I'm not sure what we do about that. One option would be to base64 the whole provided name, but that might make it difficult to correlate the metrics later. Although, now that I look at the [cAdvisor docs](https://github.com/google/cadvisor/blob/master/docs/runtime_options.md#container-labels), I see that it's able to pass through labels! So maybe we can just base64 the provided name in order to get a valid container name, and add the original name as a label to allow the metrics to be correlated with the shard.

psFried (changes requested):
Thanks, I think this is moving in the right direction, though there's more to do before it should merge. LMK what you think about the name sanitization.

I think in the meantime, I might try to temporarily hack in a UUID name for each container, just so I can change over to an explicit `docker stop` command. Then that name can be superceded by the one from this PR once it's ready.

go/connector/run.go
dhammika: Nice, labeling is much more elegant! We can just keep docker assigned IDs and use labels for metrics tagging. 
This setup won't restrict duplicate containers from orchestrator misbehavior though. Tags will look something like, 
```
df6abb0ec43a    keen_ellis      shard=materialize/examples/stats/00000000-00000000,FLOW_RUNTIME_PROTOCOL=materialize,FLOW_TCP_PORT=2222,command=transactions
2b791ba97745    kind_shtern     FLOW_RUNTIME_PROTOCOL=materialize,FLOW_TCP_PORT=2222,command=transactions,shard=materialize/test/hello-world/ssh-postgresl/00000000-00000000
```

go/bindings/build.go
dhammika: Good point on `request.Capture`. I used that instead of `BuildArgs`. 
`ContainerName` is already passed to `CaptureDriverFn` and `MaterializeDriverFn`

go/bindings/build.go
dhammika: Patched below.

dhammika:
I added SHA hashed names and CI is failing with duplicate name collisions. I'll check connector shutdown/eviction path.

psFried (approved):
I think this is looking pretty good. I'd like to address the couple of comments before merging, but I'm not actually certain that additional changes would be required, so LMK what you think.

go/flowctl-go/cmd-api-spec.go
```diff
@@ -36,6 +36,7 @@ type apiSpec struct {
 	Diagnostics mbp.DiagnosticsConfig `group:"Debug" namespace:"debug" env-namespace:"DEBUG"`
 	Image       string                `long:"image" required:"true" description:"Docker image of the connector to use"`
 	Network     string                `long:"network" description:"The Docker network that connector containers are given access to."`
+	Name        string                `long:"name" description:"The Docker container name."`
```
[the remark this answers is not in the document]
dhammika: That flag is just a place holder in both `spec` and `discover` api calls. Gating those calls doesn't sound _right_ and honestly I'm not sure if different agents can make concurrent calls on the same image.

go/flowctl-go/cmd-api-discover.go
```diff
@@ -23,6 +23,7 @@ type apiDiscover struct {
 	Diagnostics mbp.DiagnosticsConfig `group:"Debug" namespace:"debug" env-namespace:"DEBUG"`
 	Image       string                `long:"image" required:"true" description:"Docker image of the connector to use"`
 	Network     string                `long:"network" description:"The Docker network that connector containers are given access to."`
+	Name        string                `long:"name" description:"The Docker container name."`
```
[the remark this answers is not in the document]
dhammika: Same as `spec` comment above.  I'd prefer patching caller (agent) later.

go/flowctl-go/cmd-api-spec.go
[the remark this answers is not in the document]
psFried: OK, sounds good to me. The runtime is where we have the most acute need, so I don't think spec or discover should block merging this PR

go/flowctl-go/cmd-api-discover.go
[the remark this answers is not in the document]
psFried: SGTM :+1:

jgraettinger:
Not a review, but I did skim this and it looks 👍  to me.

PR #724, opened 2022-10-18
Await TCP copy goroutines in connector.Run

psFried:
**Description:**

When running connectors using the TCP mode, it was possible that the goroutines copying data between the socket and the "adapter" wrapper would still be running after the call to `Run` returned. This caused a panic in the capture runtime, due to the wrapper attempting to send a deserialized message on a channel that's already been closed.

The short term (hope springs eternal) fix that's implemented in this PR will await the completion of those goroutines before returning from the `runCommand` function. A better factoring, once we no longer need to support stdin/out connectors, would be to simply eliminate the wrapper altogether, and have the runtime dial the connector as you would any normal gRPC service.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/724)

### DIFF
--- a/go/connector/run.go
+++ b/go/connector/run.go
@@ -244,21 +244,22 @@ func runCommand(
 				fe.onError(connErr)
 				// We're not going to run the copy goroutines, so decrement
 				// the waitgroup now.
-				group.Add(-2)
+				group.Done()
+				group.Done()
 				return
 			}
 
 			// Copy |writeLoop| into socket
 			go func() {
 				fe.onError(writeLoop(conn))
-				group.Add(-1)
+				group.Done()
 			}()
 
 			// Read from socket connection and delegate to output through the error interceptor
 			go func() {
 				var _, err = io.Copy(outputInterceptor, conn)
 				fe.onError(err)
-				group.Add(-1)
+				group.Done()
 			}()
 		}()
 	} else {
@@ -336,7 +337,9 @@ func connectTCP(ctx context.Context, conn net.Conn, localAddress string) (net.Co
 		if err = ctx.Err(); err != nil {
 			return nil, err
 		}
-		var dialer = net.Dialer{}
+		var dialer = net.Dialer{
+			Timeout: time.Second * 10,
+		}
 		conn, err = dialer.DialContext(ctx, "tcp", localAddress)
 		if err == nil {
 			return conn, nil
--- a/go/connector/run.go
+++ b/go/connector/run.go
@@ -252,12 +252,14 @@ func runCommand(
 			// Copy |writeLoop| into socket
 			go func() {
 				fe.onError(writeLoop(conn))
+				os.Stderr.WriteString("writeLoop end\n")
 				group.Done()
 			}()
 
 			// Read from socket connection and delegate to output through the error interceptor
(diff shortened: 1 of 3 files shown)


go/connector/run.go
```diff
@@ -225,38 +225,40 @@ func runCommand(
...
 	}
 
+	var group = sync.WaitGroup{}
 	var conn net.Conn
 	// If port is specified, use TCP socket
 	if port != "" {
+		// We're going to need to await 3 separate async tasks:
+		// The attempt to dial, and the two copy operations in
+		// and out of the established connection.
+		group.Add(3)
```
willdonnelly, line 235: If you use `errgroup` instead you can skip the manual adding and subtracting, although I'm not sure how that would interact with the `fe.onError()` abstraction so feel free to ignore if that seems like too much work.

go/connector/run.go
```diff
@@ -225,38 +225,40 @@ func runCommand(
...
-					fe.onError(fmt.Errorf("dialing connection to %s: %w", localAddress, err))
-					return
-				} else {
-					// Retry after a short wait
-					time.Sleep(1 * time.Second)
-				}
+			// Don't retry on context cancellation
+			// Retry after a short wait
+			conn, connErr := connectTCP(ctx, conn, localAddress)
+			group.Add(-1)
```
willdonnelly, line 242: nit: You should probably use `group.Done()` for this, although it's literally a helper alias for `Add(-1)` so this works fine too.

go/connector/run.go
```diff
@@ -225,38 +225,40 @@ func runCommand(
...
+			if connErr != nil {
+				fe.onError(connErr)
+				// We're not going to run the copy goroutines, so decrement
+				// the waitgroup now.
+				group.Add(-2)
```
willdonnelly, line 247: This feels like a good argument for using `errgroup` instead, since it keeps track of the number of still-running goroutines for you.

willdonnelly (approved):
LGTM with a few nits that amount to "could you use errgroup here instead?"

go/connector/run.go
psFried: The reason for not using `errgroup` is that we don't actually want to await the _first_ error, but rather the completion of _all_ the async tasks, regardless of the result.

willdonnelly: I'm not sure if I understand the problem, `errgroup.Wait()` always waits until all tasks have terminated, it just returns the first non-nil error (if any) that was returned by one of the tasks.

psFried: Apparently I'd misunderstood the docs on that, but it seems you're right. I'm inclined to still keep the existing `sync.WaitGroup` though, because the docs for `errgroup` don't specify the behavior in the case that more goroutines are added after the call to `Wait`. This condition seems unlikely, but I couldn't actually prove to myself that it couldn't happen in practice. The problem case would be that the connector starts up and then crashes before we start the goroutines to copy. I could still definitely refactor this and get `errgroup` working, but it doesn't seem worth the effort at this point.

mdibaiee (approved):
LGTM

psFried:
Closing this, since we've rolled back connectors over TCP for the time being

PR #728, opened 2022-10-19
docs: flowctl updates, logs and stats

oliviamiannone:
**Description:**

Docs updates per #707, including

- Updates to flowctl docs to reflect new commands
- Updates to collections docs (on flowctl access)
- Updates to the concept and reference pages on logs and statistics, including full docs of stats, which were omitted prior
- A few random tweaks to installation and auth docs

don't think there are any shortlinks. 

**Notes for reviewers:**

I decided to handle stats properties (sort of) like connector config properties. I'm very curious to hear feedback on this. It's the sort of thing that I don't expect people to use all the time, but needs to be somewhere. Note the table _is_ simplified, but I didn't think exhaustively detailing the whole stats schema would be necessary or even helpful here.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/728)

### DIFF
--- a/site/docs/concepts/collections.md
+++ b/site/docs/concepts/collections.md
@@ -40,10 +40,10 @@ Use `flowctl collections read --help` to see documentation for all options.
 :::info Beta
 While in beta, this command currently has the following limitations. They will be removed in a later release:
 
-* The `--uncommitted` flag is required. This means that all collection documents are read, regardless of whether they were successfully committed or if they part of a [transaction](../concepts/advanced/shards.md#transactions) that was rolled back or uncommitted.
+* The `--uncommitted` flag is required. This means that all collection documents are read, regardless of whether they were successfully committed or not.
 In the future, reads of committed documents will be the default.
 
-* Only collections comprising a single [journal](../concepts/advanced/journals.md) can be read.
+* Only reads of a single [partition](../concepts/advanced/projections.md#logical-partitions) are supported. If you need to read from a partitioned collection, use `--include-partition` or `--exclude-partition` to narrow down to a single partition.
 
 * The `--output` flag is not usable for this command. Only JSON data can be read from collections.
 :::
--- a/site/docs/concepts/collections.md
+++ b/site/docs/concepts/collections.md
@@ -15,6 +15,39 @@ you define one or more new collections as part of that process.
 Every collection has a key and an associated [schema](#schemas)
 that its documents must validate against.
 
+## Viewing collection data
+
+In many cases, it's not necessary to view your collection data — you're able to materialize it directly to a destination in the correct shape using a [connector](../concepts/README.md#connectors).
+
+However, it can be helpful to view collection data to confirm the source data was captured as expected, or verify a schema change.
+
+#### In the web application
+
+Sign into the Flow web application and click the **Collections** tab. The collections to which you have access are listed.
+Click the **Details** drop down to show a sample of collection data as well as the collection [specification](#specification).
+
+The collection documents are displayed by key. Click the desired key to preview it in its native JSON format.
+
+#### Using the flowctl CLI
+
+In your [authenticated flowctl session](../reference/authentication.md#authenticating-flow-using-the-cli), issue the command `flowctl collections read --collection <full/collection-name> --uncommitted`. For example, `flowctl collections read --collection acmeCo/inventory/anvils --uncommitted`.
+
+Options are available to read a subset of data from collections.
+For example, `--since` allows you to specify an approximate start time from which to read data, and
+`--include-partition` allows you to read only data from a specified [logical partition](../concepts/advanced/projections.md#logical-partitions).
+Use `flowctl collections read --help` to see documentation for all options.
+
+:::info Beta
+While in beta, this command currently has the following limitations. They will be removed in a later release:
+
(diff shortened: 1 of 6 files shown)


site/docs/concepts/collections.md
```diff
@@ -15,6 +15,39 @@ you define one or more new collections as part of that process.
...
+
+Options are available to read a subset of data from collections.
+For example, `--since` allows you to specify an approximate start time from which to read data, and
+`--include-partition` allows you to read only data from a specified [logical partition](../concepts/advanced/projections.md#logical-partitions).
+Use `flowctl collections read --help` to see documentation for all options.
+
+:::info Beta
+While in beta, this command currently has the following limitations. They will be removed in a later release:
+
+* The `--uncommitted` flag is required. This means that all collection documents are read, regardless of whether they were successfully committed or if they part of a [transaction](../concepts/advanced/shards.md#transactions) that was rolled back or uncommitted.
```
psFried, line 43: > or if they part of a [transaction](../concepts/advanced/shards.md#transactions) that was rolled back or uncommitted.

This just seems like a kind of awkward ending, and can be replace with "or not."

site/docs/concepts/collections.md
```diff
@@ -15,6 +15,39 @@ you define one or more new collections as part of that process.
...
+In the future, reads of committed documents will be the default.
+
+* Only collections comprising a single [journal](../concepts/advanced/journals.md) can be read.
```
psFried, line 46: This isn't _quite_ correct. I'll suggest:

- Only reads of a single partition are supported. If you need to read from a partitioned collection, then you can use the `--include-partition` and/or `--exclude-partition` arguments in order to narrow it down to a single partition per read.

site/docs/reference/working-logs-stats.md
```diff
@@ -6,65 +6,50 @@ sidebar_position: 3
...
+Additional options for `flowctl logs` and `flowctl stats` can be accessed through command-line help.
 
-You can materialize your `logs` collection to an external system.
-This is typically the preferred method if you’d like to continuously work with or monitor logs.
-It's easiest to materialize the whole collection, but you can use a [partition selector](../../concepts/materialization/#partition-selectors) to only materialize specific tasks, as the `logs` collection is partitioned on tasks.
+### Accessing logs or stats by materialization
+
+You can materialize your `logs` or `stats` collections to an external system.
+This is typically the preferred method if you’d like to continuously work with or monitor logs or statistics.
+It's easiest to materialize the whole collection, but you can use a [partition selector](../../concepts/materialization/#partition-selectors) to only materialize specific tasks, as the `logs` and `stats` collections are partitioned on tasks.
```
psFried, line 47: > It's easiest to materialize the whole collection

This seems to contradict the advice below about excluding the logs/stats of the materialization itself, since doing so requires a partition selector. I think it'd probably be best to just say that you can materialize the logs/stats from all tasks, or just a subset of them.

site/docs/reference/working-logs-stats.md
```diff
@@ -83,3 +68,51 @@ acmeCo/anvils/logs:
...
+
+Available statistics include information about the amount of data in inputs and outputs of each transaction. They also include temporal information about the transaction. Statistics vary by task type (capture, materialization, or derivation).
+
+A thorough knowledge of Flow's [advanced concepts](../concepts/README.md#advanced-concepts) is necessary to effectively leverage these statistics.
+
+`stats` collection documents include the following properties.
+
+### Shard information
+
+Each `stats` document begin with data about the shard processing the transaction.
```
psFried, line 82: I think it would help to convey that the combination of `/shard/name`, `/shard/keyBegin`, and `/shard/rClockBegin` uniquely identifies a specific task processing shard.

site/docs/reference/working-logs-stats.md
```diff
@@ -83,3 +68,51 @@ acmeCo/anvils/logs:
...
+`stats` documents include information about a transaction: its inputs and outputs,
+the amount of data processed, and the time taken.
+You can use this information to ensure that your Flow tasks are running efficiently,
+and that the amount of data processed matches your expectations.
+
+| Property | Description | Data Type | Applicable Task Type |
+|---|---|---|---|
+| `/ts` | Timestamp corresponding to the start of the transaction, rounded to the nearest minute | string | All |
+| `/openSecondsTotal` | Total time that the transaction was open before starting to commit | number | All |
+| `/txnCount` | Total number of transactions represented by this stats document. Used for reduction; will always be `1`. | integer | All |
```
psFried, line 105: nit: `will always be 1` might be confusing for people, since by the time they see a stats document in either a derivation or materialization, it's very likely to be greater than 1 due to reductions.

site/docs/reference/working-logs-stats.md
```diff
@@ -83,3 +68,51 @@ acmeCo/anvils/logs:
...
+| `/capture` | Capture stats, organized by collection | object | Capture |
+| `/materialize` | Materialization stats, organized by collection | object | Materialization |
+| `/derive` | Derivation statistics | object | Derivation |
+| `/<task-type>/right/`| Input documents from a the task's source | object | Capture, materialization |
```
psFried, line 109: These pointers are missing components for the source/output collections. For captures, we'd expect `/capture/<output-collection-name>/right|out`, and for materializations it would be `/materialize/<source-collection-name>/left|right|out`

psFried (changes requested):
Overall this looks great. I managed to spot a few things worth updating, but should be good to merge once those are addressed.

PR #732, opened 2022-10-21
Enable edge functions to work on newer local Supabase Versions

travjenkins:
**Description:**
We want to use jspm because when it bundles non-es module libraries it uses Node polyfills instead of esm's use of Deno STD polyfills.

Also - there are some formatting changes because I saved the file and my LSP kicked in.

**Workflow steps:**

Start up edge functions with any recent version of Supabase

None

**Notes for reviewers:**

This feels so weird as a fix for the problem and I also still do not know exactly what happened in 1.4.7 that messed stuff up. I looked at the change (https://github.com/supabase/cli/compare/v1.4.6...v1.4.7) and could not see anything that would cause an issue. The version of Deno was already locked so it should not be that.

On version 1.4.6 and before we were good.
On version 1.4.7 we get an error like this:
```
...
...
truncated
...
...
TS2345 [ERROR]: Argument of type '"SOA"' is not assignable to parameter of type 'RecordType'.
    this.#query(name, "SOA").then(({ code, ret }) => {
                      ~~~~~
    at https://deno.land/std@0.153.0/node/internal_binding/cares_wrap.ts:451:23

TS2694 [ERROR]: Namespace 'Deno' has no exported member 'SOARecord'.
          ret[0] as Deno.SOARecord;
                         ~~~~~~~~~
    at https://deno.land/std@0.153.0/node/internal_binding/cares_wrap.ts:456:26

TS2339 [ERROR]: Property 'reason' does not exist on type 'Event'.
          throw event.reason;
                      ~~~~~~
    at https://deno.land/std@0.153.0/node/process.ts:290:23

TS2339 [ERROR]: Property 'reason' does not exist on type 'Event'.
        uncaughtExceptionHandler(event.reason, "unhandledRejection");
                                       ~~~~~~
    at https://deno.land/std@0.153.0/node/process.ts:294:40

TS2339 [ERROR]: Property 'reason' does not exist on type 'Event'.
      process.emit("unhandledRejection", event.reason, event.promise);
                                               ~~~~~~
    at https://deno.land/std@0.153.0/node/process.ts:299:48

TS2339 [ERROR]: Property 'promise' does not exist on type 'Event'.
      process.emit("unhandledRejection", event.reason, event.promise);
                                                             ~~~~~~~
    at https://deno.land/std@0.153.0/node/process.ts:299:62

TS2339 [ERROR]: Property 'error' does not exist on type 'Event'.
      uncaughtExceptionHandler(event.error, "uncaughtException");
                                     ~~~~~
    at https://deno.land/std@0.153.0/node/process.ts:307:38

Found 23 errors.
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/732)

### DIFF
--- a/supabase/functions/oauth/access-token.ts
+++ b/supabase/functions/oauth/access-token.ts
@@ -1,113 +1,114 @@
-import { serve } from "https://deno.land/std@0.131.0/http/server.ts";
-import { createClient } from "https://esm.sh/@supabase/supabase-js";
-import Handlebars from "https://esm.sh/handlebars";
-import jsonpointer from "https://esm.sh/jsonpointer.js";
-import { returnPostgresError, handlebarsHelpers } from "../_shared/helpers.ts";
-import { corsHeaders } from "../_shared/cors.ts";
-import { supabaseClient } from "../_shared/supabaseClient.ts";
+import { serve } from 'https://deno.land/std@0.131.0/http/server.ts';
+import { createClient } from 'https://esm.sh/@supabase/supabase-js';
+import HandlebarsJS from 'https://dev.jspm.io/handlebars@4.7.6';
+import jsonpointer from 'https://esm.sh/jsonpointer.js';
+import { returnPostgresError, handlebarsHelpers } from '../_shared/helpers.ts';
+import { corsHeaders } from '../_shared/cors.ts';
+import { supabaseClient } from '../_shared/supabaseClient.ts';
 
 interface OauthSettings {
-  oauth2_client_id: string,
-  oauth2_client_secret: string,
-  oauth2_injected_values: any,
-  oauth2_spec: any
+    oauth2_client_id: string;
+    oauth2_client_secret: string;
+    oauth2_injected_values: any;
+    oauth2_spec: any;
 }
 
 export async function accessToken(req: Record<string, any>) {
-  const { state, config, redirect_uri, ...params } = req;
-  Handlebars.registerHelper(handlebarsHelpers);
+    const { state, config, redirect_uri, ...params } = req;
+    (HandlebarsJS as any).registerHelper(handlebarsHelpers);
 
-  const decodedState = JSON.parse(atob(state));
-  const { connector_id } = decodedState;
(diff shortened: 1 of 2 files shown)


supabase/functions/oauth/access-token.ts
```diff
@@ -1,113 +1,114 @@
-import { serve } from "https://deno.land/std@0.131.0/http/server.ts";
-import { createClient } from "https://esm.sh/@supabase/supabase-js";
-import Handlebars from "https://esm.sh/handlebars";
-import jsonpointer from "https://esm.sh/jsonpointer.js";
-import { returnPostgresError, handlebarsHelpers } from "../_shared/helpers.ts";
-import { corsHeaders } from "../_shared/cors.ts";
-import { supabaseClient } from "../_shared/supabaseClient.ts";
+import { serve } from 'https://deno.land/std@0.131.0/http/server.ts';
+import { createClient } from 'https://esm.sh/@supabase/supabase-js';
+import HandlebarsJS from 'https://dev.jspm.io/handlebars@4.7.6';
```
travjenkins, line 3: This is the main change here. We want to import from jspm

supabase/functions/oauth/access-token.ts
```diff
@@ -1,113 +1,114 @@
...
-  let headers = {};
-  if (oauth2_spec.accessTokenHeaders) {
-    const headersTemplate = Handlebars.compile(
-      JSON.stringify(oauth2_spec.accessTokenHeaders)
-    );
-    headers = JSON.parse(
-      headersTemplate({
-        redirect_uri,
+    const { state, config, redirect_uri, ...params } = req;
+    (HandlebarsJS as any).registerHelper(handlebarsHelpers);
```
travjenkins, line 18: This is the other change through the code. We want to write `(HandlebarsJS as any)` to prevent typing issues. This approach was taken from the libraries that "ported" Handlebars to Deno: https://github.com/alosaur/handlebars/blob/master/mod.ts

mdibaiee (approved):
Thanks Travis

PR #733, opened 2022-10-24
supabase/db: live_specs needs index on spec_type

mdibaiee:
**Description:**

- I checked stats for our queries on `live_specs_ext` and found that all offending queries have a filter on `spec_type` which does not have an index, and they order by `updated_at DESC NULL LAST` which also does not have an index. With these changes, I can see an improvement from 3.6-4s down to 1.9-2.2s.
- To get statistics, I used:
```
SELECT query, mean_exec_time, total_exec_time, min_exec_time, max_exec_time FROM pg_stat_statements WHERE query LIKE '%live_specs_ext%';
```
- This is still relatively slow for 10 documents, and that's becaues apparently our authentication policy on `live_specs` table runs before any other quals (i.e. before filtering by `spec_type`). I did some research trying to find a way to optimise this, and basically ask postgres: can you please filter by `spec_type` first? I see a few relevant options such as `security_barrier` views and `leakproof` functions, but I couldn't make them work for our use case where the table itself has a policy on it. So I have posted this question with more details on [stackoverflow](https://stackoverflow.com/questions/74183521/postgres-optimisation-run-view-query-before-policy) for now.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/733)

### DIFF
--- a/supabase/migrations/09_publications.sql
+++ b/supabase/migrations/09_publications.sql
@@ -64,6 +64,11 @@ alter table live_specs enable row level security;
 create unique index idx_live_specs_catalog_name on live_specs
   (catalog_name text_pattern_ops);
 
+create index idx_live_specs_spec_type on live_specs
+  (spec_type);
+
+create index idx_live_specs_updated_at on live_specs (updated_at desc nulls last);
+
 create policy "Users must be read-authorized to the specification catalog name"
   on live_specs as permissive for select
   using (auth_catalog(catalog_name, 'read'));


supabase/migrations/09_publications.sql
```diff
@@ -64,6 +64,11 @@ alter table live_specs enable row level security;
 create unique index idx_live_specs_catalog_name on live_specs
   (catalog_name text_pattern_ops);
 
+create index idx_live_specs_spec_type on live_specs
+  (spec_type);
+
+create index idx_live_specs_updated_at on live_specs (updated_at desc nulls last);
```
jgraettinger, line 70: what's the purpose of this second index?

should it be combined with the index on `spec_type` (do queries typically filter on both?)?

jgraettinger (approved):
LGTM

>  that's becaues apparently our authentication policy on live_specs table runs before any other quals (i.e. before filtering by spec_type)

interesting, nice writeup in stack overflow

jgraettinger:
The nature of these queries is asking the DB to do a lot of work, since it must scan _all_ specs (though filtered a bit on type) on the off-chance that the user will be read-authorized to a given spec. The spec_type index will probably only help by a low, constant, and linear value (e.x. a 2/3 reduction in running time).

Ideally the DB would be smart enough to hoist the auth_catalog policy up into a grounded set of permitted catalog prefixes, and then use the index over catalog name, but that's asking an awful lot and clearly isn't happening.

We could potentially update the live_specs_ext view to explicitly join against the set of auth_roles for the current user? That might get the catalog_name index involved a lot earlier in the query plan.

mdibaiee:
Actually, Supabase's presentation of the Query Plan was _very_ misleading because the indentation was lost, with indentation it appears that the filtering on `auth_catalog` has actually negligible cost:

```
| QUERY PLAN                                                                                                                                            |
| ----------------------------------------------------------------------------------------------------------------------------------------------------- |
| Limit  (cost=0.68..71.32 rows=10 width=2118)                                                                                                          |
|   ->  Nested Loop  (cost=0.68..1159.11 rows=164 width=2118)                                                                                           |
|         ->  Nested Loop Left Join  (cost=0.43..1155.58 rows=164 width=2022)                                                                           |
|               Join Filter: (((c.id)::macaddr8 = (t.connector_id)::macaddr8) AND (l.connector_image_tag = t.image_tag))                                |
|               ->  Nested Loop Left Join  (cost=0.43..1026.92 rows=164 width=1979)                                                                     |
|                     ->  Nested Loop Left Join  (cost=0.28..1009.96 rows=164 width=1428)                                                               |
|                           ->  Seq Scan on live_specs l  (cost=0.00..785.86 rows=164 width=1389)                                                       |
|                                 Filter: (auth_catalog((catalog_name)::text, 'read'::grant_capability) AND (spec_type = 'capture'::catalog_spec_type)) |
|                           ->  Index Scan using publication_specs_pkey on publication_specs p  (cost=0.28..1.37 rows=1 width=55)                       |
|                                 Index Cond: (((live_spec_id)::macaddr8 = (l.id)::macaddr8) AND ((pub_id)::macaddr8 = (l.last_pub_id)::macaddr8))      |
|                     ->  Memoize  (cost=0.15..0.34 rows=1 width=586)                                                                                   |
|                           Cache Key: l.connector_image_name                                                                                           |
|                           ->  Index Scan using connectors_image_name_key on connectors c  (cost=0.14..0.33 rows=1 width=586)                          |
|                                 Index Cond: (image_name = l.connector_image_name)                                                                     |
|               ->  Materialize  (cost=0.00..22.55 rows=37 width=55)                                                                                    |
|                     ->  Seq Scan on connector_tags t  (cost=0.00..22.37 rows=37 width=55)                                                             |
|         ->  Function Scan on view_user_profile u  (cost=0.25..0.26 rows=1 width=96)                                                                   |
```

The two more significant costs are the join of `live_specs` and `publication_specs` and function scan of `view_user_profile`. I'm fiddling with these to see if I can improve them.

PR #739, opened 2022-10-24
feature: restrict publishing connector images/tags to those defined in the database

### LINKED ISSUE #727: require that published task images are in `connectors` table

jshearer:
Fixes #727

**Description:**

We want to whitelist the connector images that we allow users to run, for obvious security reasons. This introduces a check during publication that the connector image and requested tag exist in the database.

**Examples:**

When the connector image does not exist:
```
detail=Error validating connector tag 'ghcr.io/estuary/source-http-file:dev' because 'ghcr.io/estuary/source-http-file' is not an authorized connector image
```

When the connector image exists, but the specific tag does not: 
```
detail=Error validating connector tag 'ghcr.io/estuary/source-hello-world:dev' because ':dev' is not an allowed tag
```

When the connector image and tag exist, but the tag status isn't success:
```
detail=Error validating connector tag 'ghcr.io/estuary/source-hello-world:v1' because tag ':v1' status is 'SpecFailed'
```

**Things I'm thinking about still:**

* What about when we're developing connectors locally? Is there some "we're running in temp-data-plane mode, let me live dangerously" flag?
* Should these errors actually be implemented as an enum and link to (as-yet nonexistent) docs pages like some of the other errors do?
* Is there a better way to write the [query](https://github.com/estuary/flow/pull/739/files#diff-9252bb50f2c52256ceb9d4ebf4b42bc55e80d44cc23c11592422083d1abb029aR68-R78) than falling back to the non-macro `query_as`? The reason is that I wanted to use `sqlx::types::Json<...>`, but:
  * the `query_as!` macro we're using everywhere else [explicitly doesn't support `FromRow`](https://github.com/launchbadge/sqlx/issues/514)
  * the `query!` macro doesn't let you name the output struct
  * the `query_as` function (not macro) doesn't do build-time typechecking or automatic argument binding

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/739)


psFried (commented):
>     What about when we're developing connectors locally? Is there some "we're running in temp-data-plane mode, let me live dangerously" flag?

I think my bias here would be that, before we add a danger mode, we'd try to make it as easy as possible to just add the connector images to the database locally. IIRC there's a script to add one pretty easily, so it may not be much of an issue?

>     Should these errors actually be implemented as an enum and link to (as-yet nonexistent) docs pages like some of the other errors do?

IMO it's not necessary. The error message you have is pretty clear already. And we eventually would like to support running arbitrary images, so the limitation hopefully won't be in place for very long.

>     Is there a better way to write the [query](https://github.com/estuary/flow/pull/739/files#diff-9252bb50f2c52256ceb9d4ebf4b42bc55e80d44cc23c11592422083d1abb029aR68-R78) than falling back to the non-macro query_as? The reason is that I wanted to use sqlx::types::Json<...>, but:

It seems like the only reason you need to use `Json<_>` is because you're returning all the image tags as a json array. So I think my first question is whether we really need to look at the tags at all? Would it be good enough to only validate the image name? IMO it seems fine, but I'm sure you've thought about that more than I have.

If you still want to include the tags, I'd expect that it would work if you return tags as a postgres array instead of JSON, which could possibly also require a special annotation in the query, as described in [this SO answer](https://stackoverflow.com/a/69474491).

crates/agent-sql/src/connector_tags.rs
```diff
@@ -42,6 +42,47 @@ pub async fn dequeue(txn: &mut sqlx::Transaction<'_, sqlx::Postgres>) -> sqlx::R
     .await
 }
 
+/// JobStatus is the possible outcomes of a handled connector tag.
+#[derive(Debug, Deserialize, Serialize)]
+#[serde(rename_all = "camelCase", tag = "type")]
+pub enum JobStatus {
```
jgraettinger, line 48: The intent of the separation of these crates is that agent-sql is _only_ queries and structs which are POD of query outcomes.

I see why you moved it, but it's a smell that you had too.

crates/agent/src/connector_tags.rs
```diff
@@ -143,11 +132,8 @@ impl TagHandler {
         .await?;
 
         if let Some(oauth2_spec) = oauth2_spec {
-            agent_sql::connector_tags::update_oauth2_spec(
-                row.connector_id,
-                oauth2_spec,
-                txn,
-            ).await?;
+            agent_sql::connector_tags::update_oauth2_spec(row.connector_id, oauth2_spec, txn)
```
jgraettinger, line 135: Are you running rustfmt with default settings? Seeing some churn here and elsewhere that I wouldn't expect.

crates/agent-sql/src/connector_tags.rs
```diff
@@ -42,6 +42,47 @@ pub async fn dequeue(txn: &mut sqlx::Transaction<'_, sqlx::Postgres>) -> sqlx::R
...
+    Queued,
+    PullFailed,
+    SpecFailed,
+    OpenGraphFailed { error: String },
+    Success,
+}
+
+// ConnectorTags represent all of the tags defined for a particular connector
+#[derive(Debug, FromRow)]
```
jgraettinger, line 57: Right, I think I have an alternative proposal for how to go about this:

First, all we really need to validate is that the connector _image_ is defined in the connectors table. The image alone is what implies ownership, and we very frequently are using alternative or draft tags of images, so it would be a big UX issue even just for us if we enforced tag equivalence.

Second, rather than loading connectors and tags, you can frame the whole thing as a query akin to `resolve_expanded_spec_rows` like:

```
pub async fn resolve_unknown_connectors(
    seed_ids: Vec<Id>,
    txn: &mut sqlx::Transaction<'_, sqlx::Postgres>,
) -> sqlx::Result<Vec<UnknownConnector>> {
```

Where UnknownConnector is a struct with the offending spec catalog_name and the connector.

This would be run _after_ `apply_updates_for_row` has run for all specs, such that the `live_specs` table has already been updated with the provisional connector. See how we run `expanded_specifications` for an example of another place we're using this "query after you've already updated" technique.

What's nice about this approach is that the entire check can be framed as a single SQL query in a stand-alone function which returns zero or more rows, and any row > zero Is mapped into an error.

When we're ready to remove this check, we remove the function and its call.

Note this approach is invariant to whether we're checking images and tags, or just images (but again, we should just check images).

jgraettinger:
Oh, and we also need a similar check over discover operations, as that's another place we're invoking a connector within the agent.

jshearer:
Thanks for the suggestion, I like doing the check as one query. It didn't occur to me because I didn't make the connection that you can make changes and then query those changes all inside a transaction that can subsequently be rolled back if desired. Much cleaner.

**TODO:**
- [ ] Move `JobStatus` back
- [ ] Implement `resolve_unknown_connectors`
- [ ]  Validate discovers
- [ ] Only validate images, not tags

crates/agent-sql/src/connector_tags.rs
jshearer: Done!

crates/agent/src/connector_tags.rs
jshearer: I haven't configured rustfmt, just using the built in vscode plugin... I do see this churn. I'll manually reset these couple of lines, but I _have_ seen this before too especially around autogenerated rust protobuf files...

crates/agent-sql/src/connector_tags.rs
jshearer: Moved it back

crates/agent/src/publications/specs.rs
```diff
@@ -282,8 +284,8 @@ pub fn validate_transition(
 
         // Verify that the live specification has not existed and then been deleted in the past.
         // TODO(johnny): remove once we introduce data plane pet-names.
-        if live_type.is_none() && draft_type.is_some() && *last_pub_id != pub_id   {
```
jshearer, line 285: @jgraettinger so I'm pretty sure that this was actually wrong, and my rustfmt fixed it. The spacing was off, there shouldn't have been two spaces here or two indents below

crates/agent-sql/src/connector_tags.rs
```diff
@@ -42,6 +42,47 @@ pub async fn dequeue(txn: &mut sqlx::Transaction<'_, sqlx::Postgres>) -> sqlx::R
     .await
 }
 
+// ConnectorTags represent all of the tags defined for a particular connector
```
jgraettinger, line 45: nit: comment is out of date

crates/agent/src/publications/specs.rs
jgraettinger: 👍

jgraettinger (approved):
LGTM! % a couple of nits below.

jgraettinger (changes requested):
Oh: I would like to see test coverage added here, though.

See `agent-sql/tests/publications.rs`, for example.

The existing publications and discover handlers are woefully lacking, but see also how directives like `grant.rs` or `beta_onboard.rs` are tested in the `agent` crate for a suggested test framework.

jgraettinger (approved):
Awesome, LGTM

PR #740, opened 2022-10-24
connector: revert to stdio behavior

jgraettinger:
**Description:**

Roll-back the implementation of connector/run.go to stdio-only mode. This is being done out of excess caution only, and not because of a specific known problem in this implementation.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/740)


willdonnelly (approved):
LGTM

PR #742, opened 2022-10-25
live_specs_ext: optimise by joining auth.users instead of using function

mdibaiee:
**Description:**

- Instead of using a function, we can join `auth.users` on `p.user_id = auth.users.id` and from my tests, this makes the query run as fast as 500ms instead of 2s. In order to test this, try running a query against the `live_specs_ext_inline` (a view I created this way) from the UI.

**Workflow steps:**

**Notes for reviewers:**


mdibaiee:
`publication_specs_ext` is also using a lateral join on `view_user_profile` at the moment, so we may also migrate that one once we confirm this is working well for us.

supabase/migrations/10_spec_ext.sql
```diff
@@ -89,14 +89,14 @@ select
...
   u.email as last_pub_user_email,
-  u.full_name as last_pub_user_full_name
+  coalesce(u.raw_user_meta_data->>'full_name', u.raw_user_meta_data->>'name') as last_pub_user_full_name
 from live_specs l
 left outer join publication_specs p on l.id = p.live_spec_id and l.last_pub_id = p.pub_id
 left outer join connectors c on c.image_name = l.connector_image_name
-left outer join connector_tags t on c.id = t.connector_id and l.connector_image_tag = t.image_tag,
-lateral view_user_profile(p.user_id) u
+left outer join connector_tags t on c.id = t.connector_id and l.connector_image_tag = t.image_tag
+inner join auth.users u on u.id = p.user_id;
```
jgraettinger, line 99: 🤔 

The rationale for the `view_user_profile` RPC is to a) package user metadata in a standard way for consumption within the UI and APIs, and b) as a `public` schema work-around for the absence of a row-level security policy on auth.users (and it's not clear what the policy would even be).

Basically, we require that you already know the user ID in order to look up metadata, and then we give it to you in a packaged and opinionated way.

Simply joining directly against `auth.users` though clearly makes a difference on performance. There isn't an RLS policy being evaluated over auth.users, but that's fine because the user is only able to join against users attached to specs that they're allowed to see anyway, and we've defined things such that that's okay: you can see who changed the specs you're entitled to see.

I don't love the inlining of coalesce behavior here though, and have a specific proposal: can we package the reshaping behavior done in `view_user_profile` and (now) here into a view within the `internal` schema (which is not part of our public API), and then a) join against that view here and b) update `view_user_profile` to use that view as well?

Then we have a single consistent place for defining the shape of user metadata in our APIs. Thanks

mdibaiee: @jgraettinger sounds good, applied!

supabase/migrations/10_spec_ext.sql
```diff
@@ -95,8 +102,8 @@ select
 from live_specs l
 left outer join publication_specs p on l.id = p.live_spec_id and l.last_pub_id = p.pub_id
 left outer join connectors c on c.image_name = l.connector_image_name
-left outer join connector_tags t on c.id = t.connector_id and l.connector_image_tag = t.image_tag,
-lateral view_user_profile(p.user_id) u
+left outer join connector_tags t on c.id = t.connector_id and l.connector_image_tag = t.image_tag
+inner join user_profiles u on u.id = p.user_id;
```
jgraettinger, line 106: This should probably also be a left outer join.

it shouldn't matter in practice, but if a user is deleted from the system (we don't do this today, nor have a plan to) we'd still want this row to be surfaced in the view -- just with nulls for `last_pub_user_email` etc.

jgraettinger (approved):
A few more comments then LGTM

supabase/migrations/10_spec_ext.sql
```diff
@@ -17,17 +17,24 @@ create type user_profile as (
...
   select
-    id as user_id,
+    user_id,
     email,
-    coalesce(raw_user_meta_data->>'full_name', raw_user_meta_data->>'name') as full_name,
-    coalesce(raw_user_meta_data->>'picture', raw_user_meta_data->>'avatar_url') as avatar_url
-  from auth.users where id = bearer_user_id;
+    full_name,
+    avatar_url
+  from internal.user_profiles where id = bearer_user_id;
```
jgraettinger, line 37: heads up, this now needs to be `user_id = bearer_user_id`

mdibaiee:
Okay, I tried applying these changes on prod, and I realised my test that showed the query becoming faster was actually wrong:

In order for this to work, we need to run these:
```
alter view internal.user_profiles owner to authenticated;
grant select on auth.users to authenticated;
```
With these, the query runs as slow as before, not any faster. The reason my test was fast was because the owner of the new view I had created was still `postgres` so it skipped the policy altogether 🤦🏽 

However, this one works and is secured by the join:
```
create view live_specs_ext_inline as
select
  distinct on (l.catalog_name) catalog_name,
  l.id,
  l.connector_image_name,
  l.connector_image_tag,
  l.last_build_id,
  l.last_pub_id,
  l.reads_from,
  l.spec,
  l.spec_type,
  l.writes_to,
  l.created_at,
  l.detail,
  l.updated_at,
  c.external_url as connector_external_url,
  c.id as connector_id,
  c.title as connector_title,
  c.short_description as connector_short_description,
  c.logo_url as connector_logo_url,
  c.recommended as connector_recommended,
  t.id as connector_tag_id,
  t.documentation_url as connector_tag_documentation_url,
  p.detail as last_pub_detail,
  p.user_id as last_pub_user_id,
  u.avatar_url as last_pub_user_avatar_url,
  u.email as last_pub_user_email,
  u.full_name as last_pub_user_full_name
from live_specs l
inner join internal.user_roles(auth_uid()) on starts_with(l.catalog_name, role_prefix) and capability >= 'read'
left outer join publication_specs p on l.id = p.live_spec_id and l.last_pub_id = p.pub_id
left outer join connectors c on c.image_name = l.connector_image_name
left outer join connector_tags t on c.id = t.connector_id and l.connector_image_tag = t.image_tag
left outer join internal.user_profiles u on u.user_id = p.user_id;
;
grant select on live_specs_ext_inline to authenticated;
```

jgraettinger:
okay, does that mean that the internal.user_profiles view doesn't actually matter at all then? should we roll it back?

Correct to say that all of the query performance improvement is solely in the way we're evaluating the `live_specs` RLS policy, and extracting it is essentially all of the benefit?

dyaffe:
Anecdotally the UI does seem a lot faster now

PR #743, opened 2022-10-25
sshforwarding: Improve SSH endpoint schema

willdonnelly:
**Description:**

This commit fixes the description of the SSH Endpoint config schema to mention the expected shape of the value. It also adds a `pattern` annotation with a validation regexp.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/743)


mdibaiee:
The new network-tunnel is here: https://github.com/estuary/network-tunnel

We should backport this fix over there as well afterwards, thanks!

willdonnelly:
This PR won't do anything since we plan to instead disable connector proxy network tunneling entirely. I'll see about adding the validation regexp in the right place after that's finished instead.

PR #746, opened 2022-10-26
feature: Implement tenant soft quotas

### LINKED ISSUE #741: soft quotas on tenant tasks & collections, enforced at publication time

jshearer:
**Description:**

This introduces per-tenant quotas for collections, captures, materializations, and derivations. At publication time, we compute the change in usage of each of these resources, and compare that to the tenant quota and current number of resources in use. If everything checks out, we update the database with the new resource-in-use counts.

Fixes #741

Example failure:
Screen Shot 2022-10-25 at 21 30 33

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/746)


crates/agent/src/publications/specs.rs
```diff
@@ -350,6 +350,238 @@ pub fn validate_transition(
     errors
 }
 
+pub async fn enforce_resource_quotas(
```
psFried, line 353: This seems like it needs to account for tasks being disabled

psFried (changes requested):

jshearer:
**TODO:** Switch to `anyhow::Result<Result<HashMap<String, ResourceUsageChanges>>, Vec<Error>>>`

crates/agent-sql/src/publications.rs
```diff
@@ -208,6 +208,89 @@ pub async fn resolve_spec_rows(
...
+            captures_quota,
+            derivations_quota,
+            materializations_quota,
+            collections_quota,
+            captures_used,
+            derivations_used,
+            materializations_used,
+            collections_used
+        from tenants
+        where $1 ILIKE tenant || '%';
```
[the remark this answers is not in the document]
jshearer: Done!

crates/agent/src/publications/specs.rs
jshearer: Done!

crates/agent/src/publications/specs.rs
```diff
@@ -350,6 +350,238 @@ pub fn validate_transition(
...
+        }
+    }
+
+    let mut tenant_grouped_changes: HashMap<String, ResourceUsageChanges> = HashMap::new();
+    let mut tenants: HashMap<String, Tenant> = HashMap::new();
+
+    // Is there a more idiomatic way to do this with mapping and iterators and stuff?
+
+    for (catalog_name, changes) in catalog_change_map.iter() {
+        let res = agent_sql::publications::find_tenant_for_catalog_name(catalog_name, txn)
```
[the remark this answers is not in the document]
jshearer: I rewrote this part

crates/agent-sql/src/publications.rs
```diff
@@ -208,6 +208,123 @@ pub async fn resolve_spec_rows(
...
+}
+
+#[derive(Debug)]
+struct TenantFromDB {
+    pub name: String,
+    pub captures_quota: i32,
+    pub derivations_quota: i32,
+    pub materializations_quota: i32,
+    pub collections_quota: i32,
+    // These are Option<i64> because even if we add `coalesce(.., 0) to the query
```
jgraettinger, line 231: nit: instead of doing this, you can tell sqlx in the query that these are non-null by using `foobar as "foobar!"`.

[here's one example](https://github.com/estuary/flow/blob/master/crates/agent-sql/src/publications.rs#L188-L193)

Then you can get rid of `TenantFromDB`

crates/agent-sql/src/publications.rs
```diff
@@ -208,6 +208,123 @@ pub async fn resolve_spec_rows(
...
+            tenants.captures_quota,
+            tenants.derivations_quota,
+            tenants.materializations_quota,
+            tenants.collections_quota,
+            count(live_specs.catalog_name) filter (where live_specs.spec_type = 'capture') as captures_used,
+            count(live_specs.catalog_name) filter (where live_specs.spec_type = 'collection' and live_specs.spec->'derivation' is not null) as derivations_used,
+            count(live_specs.catalog_name) filter (where live_specs.spec_type = 'materialization') as materializations_used,
+            count(live_specs.catalog_name) filter (where live_specs.spec_type = 'collection') as collections_used
+        from tenants
+        join live_specs on live_specs.catalog_name LIKE tenants.tenant || '%' and (live_specs.spec->'shards'->>'disable')::boolean is not true
```
jgraettinger, line 273: supernit: use starts_with(live_specs.catalog_name, tenants.tenant) for consistency with how we do this everywhere else. This makes code search easier. _Probably_ there's no performance impact if postgres optimizes to this already.

crates/agent-sql/src/publications.rs
```diff
@@ -208,6 +208,123 @@ pub async fn resolve_spec_rows(
...
+        where tenants.tenant = ANY($1)
+        group by tenants.tenant, tenants.captures_quota, tenants.materializations_quota, tenants.derivations_quota, tenants.collections_quota;
```
jgraettinger, line 275: this group by should only be over `tenants.tenant`

crates/agent-sql/src/publications.rs
jgraettinger: No action for now, but FYI: `spec` is stored a postgres `JSON` type, and not `JSONB`. It _must_ be this way because `JSONB` will re-order properties and that would break sops macs. But, in this application is means the query must actually parse all of these documents.

A way we can potentially improve this is by creating an index over the pre-extracted values used in this query. It would need to be confirmed, but postgres is smart about using a covering index to avoid needing to do the parsing work.

crates/agent-sql/src/publications.rs
```diff
@@ -208,6 +208,123 @@ pub async fn resolve_spec_rows(
...
+        "#,
+        &tenant_names[..]
+    )
+    .fetch_all(txn)
+    .await
+    .map(|tenants| tenants.into_iter().map(Tenant::from).collect())
+}
+
+#[derive(Default)]
+pub struct ResourceUsageChanges {
```
jgraettinger, line 285: I don't think we should be baking these usage values into the tenants table.

In the future, we could surface the guts of `find_tenant_quotes` as a view `tenants_ext`, which matches the practice we have with `live_specs_ext`, `draft_specs_ext`, etc.

crates/agent-sql/src/publications.rs
```diff
@@ -208,6 +208,123 @@ pub async fn resolve_spec_rows(
...
+            collections_quota: db.collections_quota,
+            captures_used: db.captures_used.unwrap_or(0) as i32,
+            derivations_used: db.derivations_used.unwrap_or(0) as i32,
+            materializations_used: db.materializations_used.unwrap_or(0) as i32,
+            collections_used: db.collections_used.unwrap_or(0).try_into().unwrap(),
+        }
+    }
+}
+
+pub async fn find_tenant_quotas(
```
jgraettinger, line 255: please add test coverage in `agent-sql/tests/publications.rs`

supabase/migrations/16_tenants.sql
```diff
@@ -4,7 +4,17 @@ create table tenants (
 
   -- TODO(johnny): In the future, we expect to hang billing
   -- and data-plane assignment onto this record.
-  tenant      catalog_tenant unique not null
+  tenant                  catalog_tenant unique not null,
+
+  captures_quota          integer        not null DEFAULT 10,
```
jgraettinger, line 9: supernit: we use lowercase keywords. See comment at the top of the first migration `01_json.sql` for details on style.

crates/agent/src/publications/specs.rs
```diff
@@ -350,6 +350,71 @@ pub fn validate_transition(
...
+            detail: err_str,
+            ..Default::default()
+        })
+        .collect();
+
+    Ok(errors)
+}
+
+pub fn tenant_name_for_catalog_name(catalog_name: String) -> anyhow::Result<String> {
+    catalog_name
```
jgraettinger, line 411: There's a sneaky bug here, which is that this split will drop the trailing '/' and thus won't match any entries of the `tenants` table, which always have a trailing `/`.

An optional suggestion: let the database do more of the heavy lift here. Pass in the spec IDs as `resolve_expanded_rows` does and use a CTE as part of `find_tenant_quotas` which maps that into unique tenants. I think you'll find that this construction is easier to test, and it also removes the possibility of mismatches like this.

crates/agent/src/publications/specs.rs
```diff
@@ -350,6 +350,71 @@ pub fn validate_transition(
     errors
 }
 
+pub async fn enforce_resource_quotas(
```
jgraettinger, line 353: Can you add test coverage for this, akin to how directives are tested? Should be pretty feasible to do since these checks occur so early in the publication process.

jgraettinger (commented):
Looking good, some feedback below.

crates/agent-sql/src/publications.rs
jshearer: Glorious, I tried searching for this but didn't find it

crates/agent-sql/src/publications.rs
jshearer: Am I doing something else wrong in the query then? After refactoring, the query looks like this

```sql
        select
            tenants.tenant as name,
            tenants.tasks_quota,
            tenants.collections_quota,
            count(live_specs.catalog_name) filter (where
                live_specs.spec_type = 'capture' or
                live_specs.spec_type = 'materialization' or
                live_specs.spec_type = 'collection' and live_specs.spec->'derivation' is not null
            ) as "tasks_used: i32!",
            count(live_specs.catalog_name) filter (where live_specs.spec_type = 'collection') as "collections_used: i32!"
        from tenants
        join live_specs on live_specs.catalog_name LIKE tenants.tenant || '%' and (live_specs.spec->'shards'->>'disable')::boolean is not true
        where tenants.tenant = ANY($1)
        group by tenants.tenant;
```

And the error I get is

```
error returned from database: column "tenants.tasks_quota" must appear in the GROUP BY clause or be used in an aggregate function
```

jshearer: The intent is to group all of the `live_specs` by tenants.tenant, and then give me the specific `tasks_quota` and `collections_quota` for each grouping (tenant). 

My thought is that including `tasks_quota` and `collections_quota` in the list of fields to group by is okay because there will only ever be one tenant for a particular live_spec, so there will only ever be one "group".

Is there a better way to express this?

crates/agent-sql/src/publications.rs
jshearer: Done

crates/agent-sql/src/publications.rs
jshearer: Done

crates/agent-sql/src/publications.rs
```diff
@@ -208,6 +208,123 @@ pub async fn resolve_spec_rows(
     .await
 }
 
+#[derive(Debug)]
+pub struct Tenant {
+    pub name: String,
+    pub captures_quota: i32,
```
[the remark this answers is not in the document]
jshearer: Okay cool, done!

crates/agent-sql/src/publications.rs
jshearer: Yup, I had removed all of the logic that does the baking-in of usages, just forgot to remove this struct. Done!

supabase/migrations/16_tenants.sql
jshearer: done!

crates/agent/src/publications/specs.rs
jshearer: Done!

crates/agent/src/publications/specs.rs
jshearer: Done!

crates/agent-sql/src/publications.rs
jshearer: done!

jshearer:
Awesome, thanks for the detailed feedback! Just pushed a commit that:

* Changes usage metrics to just "tasks", and "collections". Tasks are: captures, materializations, and any collection that has a derivation. Collections are: all collections, including those with derivations
* Rewrote the quota usage query to:
  * Take `live_spec` IDs as input, and compute the tenants from there using a CTE
  * Hint that values aren't optional with `as "foo!"` syntax
  * Use `starts_with(..,..)` instead of `LIKE '%' || '..'`
  * Only group by tenant name. **NOTE/QUESTION:** I had to wrap the quota fields in `max(...)` because we're grouping by tenant. Is there a better way to do this?
* Added a test for `agent-sql/publications/find_tenant_quotas()`
* Added a test for `agent/publications/enforce_resource_quotas()`

crates/agent-sql/src/publications.rs
jgraettinger: I would use `sum(case when ... )` instead of `count`, where the current `filter` semantics are a case that either returns `1` or `0`.

crates/connector_proxy/src/connector_runner.rs
```diff
@@ -13,8 +15,19 @@ use proto_flow::capture::PullResponse;
 use proto_flow::flow::DriverCheckpoint;
 use tokio::io::copy;
 use tokio::process::{ChildStdin, ChildStdout};
+use tokio::task::JoinHandle;
 use tokio_util::io::{ReaderStream, StreamReader};
 
+async fn flatten_join_handle<T, E: std::convert::From<tokio::task::JoinError>>(
```
jgraettinger, line 21: I think there's a borked merge with upstream -- I'm seeing a bunch of recent changes now being unintentionally reverted in this PR.

jgraettinger (commented):
Testing work is looking great. Think there's a bad merge with upstream here though.

It looks like you can trivially use `sum` with a predicate to roll up current usage, e.x.:

```
postgres=# select sum((letter = 'a')::int) from (values ('a'), ('b'), ('a'), ('b')) as foo(letter);
 sum 
-----
   2
(1 row)
```

For the CI failure, you may want to have your test case first delete all existing live specs as a CTE, to ensure it's starting from a clean slate. Ideally the test would path against a user's local DB regardless of what they actually have in there.

crates/connector_proxy/src/connector_runner.rs
jshearer: Oh gross...

jshearer:
Hoookay, I think I fixed my mistake. It was even groser than a merge, I had accidentally `git reset --soft origin/master` instead of `git reset --soft HEAD~1`, and so rewrote my commit to have a bunch of unnecessary changes from master in it, when all I wanted to do was remove a file that shouldn't have been committed. Anyway, `git reflog` to the rescue, all should be good now.

jshearer:
@jgraettinger I rewrote the tenant usage query to use another CTE to avoid grouping by anything more than tenant name. Thoughts?

```sql
with tenant_names as (
            select tenants.tenant as tenant_name
            from tenants
            join live_specs on starts_with(live_specs.catalog_name, tenants.tenant)
            where live_specs.id = ANY($1::flowid[])
            group by tenants.tenant
        ),
        tenant_usages as (
            select
                tenant_names.tenant_name,
                count(live_specs.catalog_name) filter (
                    where
                        live_specs.spec_type = 'capture' or
                        live_specs.spec_type = 'materialization' or
                        live_specs.spec_type = 'collection' and live_specs.spec->'derivation' is not null
                ) as tasks_used,
                count(live_specs.catalog_name) filter (
                    where live_specs.spec_type = 'collection'
                ) as collections_used
            from tenant_names
            join live_specs on
                starts_with(live_specs.catalog_name, tenant_names.tenant_name) and
                (live_specs.spec->'shards'->>'disable')::boolean is not true
            group by tenant_names.tenant_name
        )
        select
            tenants.tenant as name,
            tenants.tasks_quota as "tasks_quota!",
            tenants.collections_quota as "collections_quota!",
            tenant_usages.tasks_used as "tasks_used!",
            tenant_usages.collections_used as "collections_used!"
        from tenant_usages
        join tenants on tenants.tenant = tenant_usages.tenant_name
        order by tenants.tenant;
```

jgraettinger:
> I rewrote the tenant usage query to use another CTE to avoid grouping by anything more than tenant name. Thoughts?

I realized I gave you bad advice - you don't need e.x. `task_quota` in a group-by but you _do_ still need some kind of aggregate function wrapping it. Postgres doesn't know if you, for example, might want to sum up the task_quota for _each_ joined live spec (which, to be clear, you don't here), so you have to tell it. I'll suggest using `min()` for this and avoiding the extra CTE.

I was playing with this query and came up with another construction, that I'm offering up here only because I have it already in-hand, and for you to take or leave as you will. `count ... filter` or `sum()` are both fine IMO. Otherwise I think we're veering into SQL code golf 😁 

```sql
with tenants_filtered as (
    -- Select distinct tenants from live_specs participating in the current publication.
    select distinct substring(catalog_name for position('/' in catalog_name)) as tenant
    from live_specs where id = ANY($1::flowid[])
)
select
    t.tenant,
    min(t.tasks_quota) as "tasks_quota!",
    min(t.collections_quota) as "collections_quota!",
    sum((
        (
            s.spec_type = 'capture' or
            s.spec_type = 'materialization' or
            (s.spec_type = 'collection' and s.spec->'derivation' is not null)
        ) and (s.spec->'shards'->>'disable')::boolean is not true
    )::int) as "tasks_used!",
    sum((s.spec_type = 'collection')::int) as "collections_used!"
    from tenants t
    join tenants_filtered f on t.tenant = f.tenant
    join live_specs s on starts_with(s.catalog_name, t.tenant)
    group by t.tenant;
```

crates/agent-sql/src/publications.rs
```diff
@@ -208,6 +208,62 @@ pub async fn resolve_spec_rows(
     .await
 }
 
+#[derive(Debug, Serialize)]
+pub struct Tenant {
+    pub name: String,
+    pub tasks_quota: i32,
```
jgraettinger, line 214: nit: is there a reason for the i32 / i64 distinction here? Could they be `isize` ?

jgraettinger (approved):
Awesome, really well done, LGTM % question about integer type

crates/agent-sql/src/publications.rs
jshearer: Ummm... I tried a couple of different ways to get them to be the same and I thought sqlx's inference complained about each of them. It looks like making them all `i32` actually _does_ work now, so I'll go with that. FWIW `isize` does not work, sqlx doesn't know how to infer it correctly and complains that:

```
the trait bound `isize: sqlx::Decode<'_, Postgres>` is not satisfied
the following other types implement trait `sqlx::Decode<'r, DB>`:
  f32
  f64
  i16
  i32
  i64
  i8
```

PR #748, opened 2022-10-26
Add connector resource limits

### LINKED ISSUE #660: Constrain connector resource usage
A few days ago we had a pretty bad experience trying to get a customer onboard, ultimately caused by the Firestore connector having "unfriendly" memory usage patterns that scale linearly with the size of the dataset being captured. Due to the fact that the flow reactor runs multiple workloads (captures, derivations, materializations) per instance (so, per pod in this case), not only did this specific firestore capture run out of memory, it also brought down all of the other workloads running on whatever flow-reactor instance it was scheduled on due to the entire pod getting oom-killed.

While we should absolutely fix the firestore capture connector to not use so much memory, the fact that one noisy neighbor can impact other unrelated workloads like this is an issue to address.

* **Captures** and **materializations** run as Docker containers run on the `dind` container in the `flow-reactor` pod. Conveniently, these are the easiest to constrain and also the most likely to cause resource exhaustion (is this true?). All we need to do is ask Docker to impose [runtime resource constraints](https://docs.docker.com/config/containers/resource_constraints/)
* **Derivations** run inside the `flow-reactor` main container as separate NodeJS processes, interacted with over a local HTTP api. We could either attempt to constrain these nodejs processes directly via cgroups, or we could move to running these as Docker containers. Given the discussions we're having about eventually running (everything? just derivations?) in Firecracker containers, it might be worth not putting in the effort to change how we run derivations until it's time to do that work.

Almost as important as constraining resource usage to prevent noisy neighbors, at least in my opinion, is monitoring these workloads to see if they actually are reaching their limits, and if they are, alerting about it like we do today with k8s OOM-kills. We already have metrics in Grafana... **should we just add cpu and memory metrics for each task**?

So TL;DR: we should add `--cpus="1" --memory="1g"` (or something) to our invocation of captures and materializations, and possibly monitoring for cpu/memory of captures and materializations

dhammika:
Signed-off-by: Dhammika Pathirana <dhammika@gmail.com>

Adds connector mem/cpu resource quotas.

**Workflow steps:**

N/A 

N/A 

**Notes for reviewers:**

N/A

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/748)

### DIFF
--- a/go/connector/run.go
+++ b/go/connector/run.go
@@ -126,6 +126,9 @@ func Run(
 		// Mount the connector-proxy binary, as well as the output of inspecting the docker image.
 		"--mount", fmt.Sprintf("type=bind,source=%s,target=/flow-connector-proxy", tmpProxy.Name()),
 		"--mount", fmt.Sprintf("type=bind,source=%s,target=/image-inspect.json", tmpInspect.Name()),
+		// Cgroups mem/cpu resource limits.
+		"--memory", "1g",
+		"--cpus", "2",
 	}
 
 	// Name the container to avoid duplicate connectors.


mdibaiee (approved):
LGTM

psFried (commented):
I think this looks good, but I have some reservations about deploying it that I'd like to talk through.

It's of course possible that we have connectors in production that are using more than the allowed memory, and those would break once we deploy this. I do think that our approach should be to fix the connectors rather than just increasing the limit, but it's possible for many connectors to be affected, and fixing them to use less memory could take some time.

In order to avoid interruptions to existing workflows, I think we should first know how much memory our existing connectors are using in production. I think we should hold off on merging and deploying this PR until we're sure that the limit won't result in disruptions to existing workloads. Practically, I think that probably means getting prometheus metrics scraped from connector containers, which is something we obvs want anyway. I'm just assuming here that you haven't already gotten those metrics through some other means, so LMK if that's not right.

dhammika:
> 
> It's of course possible that we have connectors in production that are using more than the allowed memory, and those would break once we deploy this. I do think that our approach should be to fix the connectors rather than just increasing the limit, but it's possible for many connectors to be affected, and fixing them to use less memory could take some time.
> 
Agree, current `dind` mem is about 21G and there are ~120 pids in that container, so ~30 active connectors. 
One option is to set a soft limit (with `memory_reservation`) instead of the hard limit set in this patch. That won't necessarily avoid #660 scenarios but just throw some warn logs when containers go over that limit.

dhammika:
> Agree, current `dind` mem is about 21G and there are ~120 pids in that container, so ~30 active connectors. One option is to set a soft limit (with `memory_reservation`) instead of the hard limit set in this patch. That won't necessarily avoid #660 scenarios but just throw some warn logs when containers go over that limit.

Actually, `memory_reservation` is a hint to the global allocator that only comes into play under total system contention. It's more of an advisory to the allocator to reclaim unused memory in that cgroup (container) and doesn't generate any docker events or logs.

dhammika:
These initial settings look ok and there's plenty head room (2x) based on connector mem on prod. This doesn't account for misbehaving connectors though. I think we should roll this in. 
connector-mem

jgraettinger (approved):
LGTM

PR #750, opened 2022-10-27
connector_proxy: use explicit shutdown_background of tokio::Runtime

jgraettinger:
**Description:**

Our use of tokio::io to read from stdin will actually map those reads into regular, synchronous reads on a blocking thread pool of the tokio::Runtime.

Sometimes those reads will run forever, since stdin is never closed by the parent process without our first exiting or closing our stdout.

However, the Drop behavior of a Runtime is to block forever awaiting all tasks to stop running, which means we can deadlock. Use shutdown_background instead to explicitly and immediately leak blocking tasks which cannot be cancelled.

Tested locally with a reproduction case using a capture connector that deliberately fails.
Before this patch I confirmed that the connector-proxy was attempting to exit but not actually doing so, because it was sitting in Runtime::Drop awaiting a blocking read of stdin.

After this patch, the connector-proxy exits as expected.

I'm still unclear why our existing E2E test isn't actually tickling this behavior. It may be due to differences in polling mode but that's not confirmed.

**Workflow steps:**

connector-proxy no longer silently fails but prevents the shard from being marked as FAILED.

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/750)

### DIFF
--- a/crates/connector_proxy/src/connector_runner.rs
+++ b/crates/connector_proxy/src/connector_runner.rs
@@ -1,5 +1,7 @@
 use crate::apis::{FlowCaptureOperation, FlowMaterializeOperation, InterceptorStream};
-use crate::errors::{Error, self, io_stream_to_interceptor_stream, interceptor_stream_to_io_stream};
+use crate::errors::{
+    self, interceptor_stream_to_io_stream, io_stream_to_interceptor_stream, Error,
+};
 use crate::interceptors::{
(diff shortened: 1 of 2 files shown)


jgraettinger:
Oh, I also removed all uses of `tokio::spawn`. I don't believe those were a causal issue, but we don't actually need them and are single-thread anyway so :shrug:

mdibaiee (approved):
awesome!!

PR #752, opened 2022-10-27
feature: Switch agent to use `LISTEN`/`NOTIFY` instead of polling

### LINKED ISSUE #747: faster builds: switch agent from polling to LISTEN/NOTIFY

jshearer:
**Description:**
One notable change is that each handler will run in a loop until it runs out of tasks to handle, whereas before the handlers would only run one task at a time. It's _possible_ that this can result in a scenario where one particular task monopolizes all of the time and never lets other tasks run, but I kind of think this is unlikely. 

If you disagree please comment and I'll add in some fairly trivial logic to allow handlers to run one task at a time and indicate whether they have more work to do, and have the `serve` function deal with scheduling.

Fixes #747

TODO:

- [x] DB Migration

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/752)


jgraettinger (commented):
Thanks, comments below. I'd also like to see test coverage for this behavior:

* SQL tests that verify the notification triggers
* Verifying an actual handler being triggered due to a table update. Perhaps existing test cases can be refactored to be invoked in this manner? For example you could factor out a routine which reads _one_ notification from the listener and invokes that handler to completion, with a goal of then using that [in a test harness like this one.](https://github.com/estuary/flow/blob/master/crates/agent/src/directives/beta_onboard.rs#L154-L163)

supabase/migrations/03_catalog-types.sql
```diff
@@ -51,3 +51,32 @@ Enumeration of Flow catalog specification types:
...
 ';
 
+CREATE OR REPLACE FUNCTION notify_channel() RETURNS trigger AS $trigger$
+DECLARE
+  rec RECORD;
+  dat RECORD;
+  payload TEXT;
+BEGIN
+
+  -- Set record row depending on operation
```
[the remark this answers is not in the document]
jshearer: Yup, totally was copy-pasta. It's much simpler now

crates/agent/src/lib.rs
```diff
@@ -32,33 +44,80 @@ pub async fn serve<E>(
...
+    };
 
-        now = tokio::time::Instant::now();
-        let next_interval = handler.handle(&pg_pool).await?;
-        tracing::trace!(delay=?now.checked_duration_since(*deadline), ?next_interval, handler = %handler.name(), "invoked handler");
+    let handle_from_queue = async {
+        while let Some(chan) = task_rx.recv().await {
+            let mut handler = handlers_by_channel
+                .get(chan.as_str())
+                .expect(format!("Unexpected task channel {}", chan).as_str())
```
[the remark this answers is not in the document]
jshearer: I did not know you could do that!

Cargo.toml
```diff
@@ -32,7 +32,7 @@ bytelines = "2.4"
 byteorder = "1.4"
 caseless = "0.2"
 chardetng = "0.1"
-chrono = "0.4"
+chrono = { version = "0.4", features = ["serde"] }
```
jgraettinger, line 35: no action right now, but we should probably switch to `time` instead of `chrono`. As I understand `chrono` isn't being updated and depends on an old version of `time` anyway

crates/agent/src/lib.rs
```diff
@@ -32,33 +55,107 @@ pub async fn serve<E>(
...
+    let handlers_by_table = handlers
+        .into_iter()
+        .map(|h| (h.table_name(), Arc::new(Mutex::new(h))))
+        .collect::<HashMap<_, _>>();
+
+    let mut listener = PgListener::connect_with(&pg_pool).await?;
+
+    listener.listen(AGENT_NOTIFICATION_CHANNEL).await?;
+
+    // We use a channel here for two reasons:
```
jgraettinger, line 67: 👍

crates/agent/src/lib.rs
```diff
@@ -32,33 +55,107 @@ pub async fn serve<E>(
...
+    // 1. Because handlers run one task at a time, and can also indicate that they have more work to perform or not,
+    //    we want to balance the time spent processing each type of handler so that no one handler can monopolize resources.
+    // 2. It makes it easy to preemptively schedule at least one run of each handler on boot up to allow for handling requests
+    //    that came in while we weren't running
+    // NOTE: it is critical that we use an unbounded channel here, otherwise we would open ourselves up to a deadlock scenario
+    // Example deadlock when using a bounded channel:
+    // 1. A spike of work comes in
```
jgraettinger, line 74: nit: unfinished comment. IMO it's enough to just say that the potential is there.

crates/agent/src/lib.rs
```diff
@@ -32,33 +55,107 @@ pub async fn serve<E>(
...
+    let listen_to_datbase_notifications = async move {
+        loop {
+            let item = listener.recv().await.map_err(|e| anyhow::Error::from(e))?;
+            let notification: AgentNotification = serde_json::from_str(item.payload())
+                .context("deserializing agent task notification")?;
+            let table: &str = &notification.table;
+
+            tracing::debug!(table = table, "Message received to invoke handler");
+            task_tx_cloned
+                .send(table.to_string())
```
jgraettinger, line 92: supernit: you own `notification` here and can take its `table` without cloning?

crates/agent/src/lib.rs
```diff
@@ -32,33 +55,107 @@ pub async fn serve<E>(
...
         tokio::select! {
             _ = &mut exit => {
                 tracing::debug!("caught signal; exiting...");
                 return Ok(()) // All done.
             }
-            _ = sleep => (),
-        };
+            listener_res = &mut listener_handler => {
+                match listener_res {
+                    // It should be impossible to get here since `listen_to_datbase_notifications` loop has no `return`s or `break`s
```
jgraettinger, line 112: nit: use `unreachable!("the reason")` then? External conditions which can fail should clearly be errors, but it's better IMO to panic immediately if an implementation invariant is violated because it gives a clearer stack trace. Returning an error tends to muddle exactly where the error originated from.

crates/agent/src/lib.rs
```diff
@@ -32,33 +55,107 @@ pub async fn serve<E>(
...
+                    // something went wrong when actually listening to the postgres channel
+                    Ok(Err(e)) => return Err(e.into()),
+                    // If we got a JoinError, the listener failed unexpectedly
+                    Err(e) => return Err(e.into()),
+                }
+            }
+            maybe_handler_table = task_rx.recv() => {
+                match maybe_handler_table {
+                    Some(handler_table) => {
+                        let mut handler = match handlers_by_table.get(&handler_table as &str) {
```
jgraettinger, line 124: nit: the cyclomatic complexity here is getting pretty gnarly. How about having `tokio::select` emit a `handler_table` which is further handled outside of its body?

The new `let ... else` syntax might also help in successive the destructuring here

crates/agent/src/lib.rs
```diff
@@ -106,3 +203,175 @@ async fn upsert_draft_specs(
...
+                    sqlx::query(include_str!("publications/test_resources/happy_path.sql"))
+                        .execute(&mut txn)
+                        .await
+                        .unwrap();
+
+                    // We have to commit the transaction for the NOTIFY to get sent
+                    txn.commit().await.unwrap();
+
+                    // Make sure that our mock publication handler was called within 25ms
+                    let res = tokio::time::timeout(Duration::from_millis(25), handler_notify_rx.recv().map(|_|()))
```
jgraettinger, line 311: this is likely to flake on CI. Do we _need_ to bound how long it will take? It should be enough to just assert that it's delivered as expected.

crates/agent/src/lib.rs
```diff
@@ -106,3 +203,175 @@ async fn upsert_draft_specs(
...
+        sqlx::query(include_str!("publications/test_resources/happy_path.sql"))
+            .execute(&mut txn)
+            .await
+            .unwrap();
+
+        // We have to commit the transaction for the NOTIFY to get sent
+        txn.commit().await.unwrap();
+
+        let notification: AgentNotification = tokio::time::timeout(
+            Duration::from_millis(50),
```
jgraettinger, line 350: same here, this could easily be violated on a time-stealing CI server.

jgraettinger (approved):
I have some comments below, especially around the level of indenting and the likelihood of CI test flakes, but LGTM % those.

crates/agent/src/lib.rs
jshearer: The intent was to prove that it wasn't polling, and instead would react "immediately"... but I do see your point about flaking in CI. I'll remove the timeout, as it's pretty clear from reading the code that it's not polling any more :P

PR #753, opened 2022-10-28
connector_proxy: Don't use NetworkTunnelCaptureInterceptor

willdonnelly:
**Description:**

This is the cleanest cut-point I could find to disable the network tunneling logic in connector_proxy, by removing all mention of the wrapping operations

    NetworkTunnelCaptureInterceptor::adapt_{request,response}_stream

from the connector runner functions.

This commit is deliberately kept to a really small diff so we can wait and verify that everything still works in prod for a bit before we start removing all the code that this change renders dead.

**Workflow steps:**

In theory this should have no noticeable impact (other than fixing the current bug with the `spec` subcommand), as the network tunneling logic has already been moved into the connectors that use it.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/753)


mdibaiee (approved):
LGTM

Issue #709, opened 2022-10-04
control-plane: stats redux


jgraettinger:
Today, we create `ops/$tenant/stats` collections with raw shard stats. We roll these up to hour grain in derivation `ops/$tenant/stats-by-hour`, and then materialize to supabase with `ops/$tenant/stats-by-hour-view`.

We'd like update the behavior of this derivation to, for each input stat document:

 * Drop the shard of the stat (we don't care).
 * Sum up top-level "bytes I read" vs "bytes I wrote" metrics across all of the bound collections of the task.
 * Emit separate documents, with a "h" / "d" / "m" grain flag, which truncate the timestamp to the corresponding grain.
 * Emit inverted documents for each bound collection, keyed on the collection, which have top-level metrics of "bytes written to me" and "bytes read from me", also at each of these grains.

PR #755, opened 2022-10-28
ops-catalog: revised stats derivation

williamhbaker:
**Description:**

Updates the stats derivation & view to include top-level aggregated values for byte and document counts written/read by/from all catalog items including tasks (captures, materializations, derivations) and collections. These counts are aggregated in units of time by hour, day, and month. It allows for the possibility of a catalog item being deleted and re-created as a different catalog item and will retain the detailed task stats for the catalog item.

Closes https://github.com/estuary/flow/issues/709

**Workflow steps:**

This will change will be apparent in the stats table. There will be new columns for the top-level aggregates. Collection stats will also be present in the stats table.

An example of a document created could look like this for a derivation as a "monthly" aggregate. Here, `tenant/test/collection` is a collection with a derivation. The derivation task stats account for the `readBy` and `writtenBy` aggregates. They contribute partially to `readFrom` and `writtenTo` - the difference in these sums is attributed to a materialization source from and a capture feed into the collection.
```json
{
    "_meta": {
        "uuid": "flow-uuid"
    },
    "grain": "monthly",
    "name": "tenant/test/collection",
    "statsSummary": {
        "readBy": {
            "bytesTotal": 50,
            "docsTotal": 11
        },
        "readFrom": {
            "bytesTotal": 150,
            "docsTotal": 10
        },
        "writtenBy": {
            "bytesTotal": 75,
            "docsTotal": 7
        },
        "writtenTo": {
            "bytesTotal": 140,
            "docsTotal": 13
        }
    },
    "taskStats": {
        "derive": {
            "out": {
                "bytesTotal": 75,
                "docsTotal": 7
            },
            "transforms": {
                "transform1": {
                    "input": {
                        "bytesTotal": 20,
                        "docsTotal": 5
                    },
                    "source": "tenant/test/source-collection1",
                    "update": {
                        "out": {
                            "bytesTotal": 1,
                            "docsTotal": 1
                        },
                        "secondsTotal": 1
                    }
                },
                "transform2": {
                    "input": {
                        "bytesTotal": 30,
                        "docsTotal": 6
                    },
                    "publish": {
                        "secondsTotal": 2,
                        "out": {
                            "docsTotal": 2,
                            "bytesTotal": 2
                        }
                    },
                    "source": "tenant/test/source-collection2"
                }
            }
        }
    },
    "ts": "2022-04-01T00:00:00.000Z"
}
```

N/A

**Notes for reviewers:**
Some TODOs still:
- [x] Create migration for the stats table changes
- [x] Complete the column comments in `11_stats.sql`
- [x] Get tests passing

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/755)


supabase/migrations/11_stats.sql
```diff
@@ -1,44 +1,37 @@
...
-';
-comment on column task_stats.task_name is
-    'Name of the catalog task';
-comment on column task_stats.task_type is '
-The type of catalog task to which stats pertain.
-
-One of "capture", "derivation", or "materialization".
-';
+create table catalog_stats (
+    name                catalog_name not null,
```
jgraettinger, line 10: nit: use the more-explicit `catalog_name`, for consistency with e.x. the `live_specs` table?

ops-catalog/catalog-stats.ts
```diff
@@ -0,0 +1,174 @@
...
+        },
+    ];
+};
+
+type StatsData = {
+    [k: string]: Document['statsSummary'];
+};
+
+const newDocumentStats = (): Document['statsSummary'] => ({
+    writtenBy: {
```
jgraettinger, line 66: Is it possible to make this more dynamic? I'm concerned about the space and processing time for all of these zero values.

There's a related issue we'll need to tackle before long, where the stats documents we currently write out have an awful lot of explicit zero-value counters.

I'm guessing this could be a little tricky because of the materialization -- we don't want the materialized columns to have NULL values, as that makes roll-ups weird. Tangentially, it would be nice if the were able to project a `default` json schema annotation keyword all the way into a Postgres column default... (and we do have the `default` [available via inference](https://github.com/estuary/flow/blob/master/go/protocols/flow/flow.proto#L230-L231))

It's okay to punt on this if it's a rabbit hole.

jgraettinger (commented):
Awesome, as far as I can tell you nailed the semantics we were going for here.

A couple of comments below. I'd also like to think further about the readFrom / readBy naming. I found the `From` vs `By` distinction pretty confusing, but a better option wasn't immediately forthcoming.

One option is just an an explicit "me" to clarify the object, as in:  `BytesReadFromMe` / `BytesReadByMe` / `DocsWrittenToMe` / `DocsWrittenByMe` ? It's a mouthful 🤢

ops-catalog/catalog-stats.ts
williamhbaker: We're already creating the table with the schema we want so we can set `default 0`, but the materialization will still try to insert `null` values if the field isn't present in the document. If the column has a `not null` constraint it will fail at this, or it will successfully insert a `null` if there isn't a `not null`, even with a `default 0`. There's probably a way to enhance the connector to enable a configuration where instead of inserting `null` for a column of a new record, it would instead insert whatever the default value was. Perhaps that is what you were getting at, beyond just the column default.

Would it really be that bad to have `null` values in the materialized table? Queries could perhaps do a `coalesce`, or just deal with a `null`.

williamhbaker:
> One option is just an an explicit "me" to clarify the object, as in: `BytesReadFromMe` / `BytesReadByMe` / `DocsWrittenToMe` / `DocsWrittenByMe` ? It's a mouthful 🤢

I updated things to use the "me" at the end. I do think it makes it pretty clear what they are, albeit a bit long. I thought about this some too and couldn't come up with anything better.

ops-catalog/template-common.flow.yaml
```diff
@@ -21,91 +21,286 @@ collections:
...
+    schema: ops-catalog-stats-schema.json
+    key: [/catalogName, /grain, /ts]
     projections:
-      # This derivation aggregates timestamps to UTC hour.
-      hourstamp: /ts
-      shard_split: /shard/split
-      task_name: /shard/name
-      task_type: /shard/kind
+      catalog_name: /catalogName
+      bytes_written_by_me: /statsSummary/writtenByMe/bytesTotal
```
jgraettinger, line 29: 👍 it's wordy, yes, but feels super clear to me as to what's being named.

jgraettinger (approved):
LGTM

Assuming we're aligned on keeping stats `bigint not null` in the public api, there's a glide path where we release this as-is today, with explicit stats, and then update the connector to leverage a `default` annotation, and then update this transformation to make them undefined unless they're actively updated. Sound good?

williamhbaker:
> LGTM
> 
> Assuming we're aligned on keeping stats `bigint not null` in the public api, there's a glide path where we release this as-is today, with explicit stats, and then update the connector to leverage a `default` annotation, and then update this transformation to make them undefined unless they're actively updated. Sound good?

Yep! I created an issue for making use of the `default` annotations that will be a follow-up from this: https://github.com/estuary/flow/issues/764

PR #763, opened 2022-11-02
connector-proxy: exit after underlying connector exits

mdibaiee:
**Description:**

- Do not wait for `streaming_all` to finish if the connector has already exit.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/763)

### DIFF
--- a/crates/connector_proxy/src/connector_runner.rs
+++ b/crates/connector_proxy/src/connector_runner.rs
@@ -35,9 +35,10 @@ pub async fn run_flow_capture_connector(
     let exit_status_task =
         async move { check_exit_status("flow capture connector:", child.wait().await) };
 
-    tokio::try_join!(streaming_all_task, exit_status_task)?;
-
-    Ok(())
+    tokio::select! {
+        Err(e) = streaming_all_task => Err(e),
+        resp = exit_status_task => resp,
+    }
 }
 
 pub async fn run_flow_materialize_connector(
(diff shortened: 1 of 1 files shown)


crates/connector_proxy/src/connector_runner.rs
```diff
@@ -35,9 +35,10 @@ pub async fn run_flow_capture_connector(
     let exit_status_task =
         async move { check_exit_status("flow capture connector:", child.wait().await) };
 
-    tokio::try_join!(streaming_all_task, exit_status_task)?;
-
-    Ok(())
+    tokio::select! {
+        Err(e) = streaming_all_task => Err(e),
```
willdonnelly, line 39: I think this will turn an error from `streaming_all_task` into an error of `run_flow_capture_connector` where previously the error would get swallowed and turned into an `Ok(())` result. That seems like a desirable change though, right?

willdonnelly (approved):
LGTM

crates/connector_proxy/src/connector_runner.rs
mdibaiee: @willdonnelly the error would not previously be swallowed, as the previous `try_join!` was `?`d, so we would get the error from the first task that would error out.

The new logic is as follows:
If `streaming_all_task` errors, bubble it up, otherwise wait for result of `exit_status_task` and return whatever result that one had

willdonnelly: Yeah, that makes sense. I had overlooked the `?` before and was just looking how the the `Ok(())` return value was removed and replaced with `select!` as a value expression. This logic seems exactly correct to me, then.

jgraettinger:
Also LGTM

PR #766, opened 2022-11-03
json/schema: consider format: date-time as ISO8601

mdibaiee:
**Description:**

- Consider `format: date-time` to be ISO8601 instead of RFC3339
- This is because `format: date-time` is sometimes used by connectors where the underlying value is not actually RFC3339, but rather ISO8601. The other option of removing `format: date-time` from such connectors reduces Flow's knowledge about the value, making it a string, whereas in cases such as materializing to postgres, ISO8601 is a valid date time value.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/766)

### DIFF
--- a/crates/json/src/schema/formats.rs
+++ b/crates/json/src/schema/formats.rs
@@ -91,7 +91,7 @@ impl Format {
             }
             Self::DateTime => ValidationResult::from(time::OffsetDateTime::parse(
                 val,
-                &time::format_description::well_known::Rfc3339,
+                &time::format_description::well_known::Iso8601::DEFAULT,
             )),
             Self::Time => ValidationResult::from(time::Time::parse(
                 val,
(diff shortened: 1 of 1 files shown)


jgraettinger:
To confirm the obvious, Iso8601 is a superset such that any RFC3339 is also ISO8601, correct?

mdibaiee:
actually, apparently RFC3339 is not a strict subset of ISO8601 and there are some formats which are supported by RFC3339, but not by ISO8601, that makes things harder: https://ijmacd.github.io/rfc3339-iso8601/

mdibaiee:
In case of HubSpot, I just realised the new code pulled from Airbyte's upstream actually has resolved this by converting the datetimes to RFC3339 in their code before outputting the documents, so I'm not blocked by this.

Given that RFC3339 is not a subset of ISO8601, I am doubtful whether this is a good idea or not now.

PR #769, opened 2022-11-03
fix download link for flowctl

psFried:
Quick doc-only PR to fix the URL in our installation instructions. The other link (https://go.estuary.dev/flowctl) is already pointing to the correct location.


mdibaiee (approved):

oliviamiannone (approved):
Thanks; my bad for missing this previously!

PR #771, opened 2022-11-04
new `doc` crate with vastly improved performance & memory efficiency

### LINKED ISSUE #189: combiner: reduce as merge tree
The `Combiner` uses a basic left-to-right strategy when reducing new RHS documents into a LHS aggregate.

This can exhibit poor performance in some circumstances. A place we especially see this is with a `set` reduction strategy over bikes in the CitiBike example, when processing very large consumer transactions (e.g. an entire captured file). As bikes churn and are added / removed to each station, Flow is essentially accumulating the entire set of bikes into separate `add`/`remove` sorted list at each station.

Each update is merging an O(1) bike ID into a O(N) sorted list of bike IDs -- we're performing an O(N) merge for each of O(N) bike IDs that move within the transaction.

An observation is that merging a O(N) list into an O(N) has the same cost as an O(1) into an O(N). We can trade off slightly more overall combine operations in exchange for making far more of those operations O(M) => O(M) merges where M << N. Ideally we'd only do this dynamically, when it's very likely to offer a performance benefit over the current strategy.

What we can do instead is to organize merges using a dynamic right-branching merge tree:
 * Initially the tree is a single root node into which RHS documents are reduced, equivalent to Combiner behavior today.
 * LHS reductions are always applied to the root.
 * RHS combines are applied to the right-most child node.
 * After a RHS combine, we'd compare the sizes of the former, combined RHS document and the current node into which it was reduced ("size" here is # of nodes in the JSON document; we have this as a by-product of reduction).
   - If they're sufficiently different we'd add a new (empty) right-branch child to the tree, to which future RHS combines are applied.
   - If the updated node aggregate is now approximately the same size as it's parent, then we'd remove a level from the tree by combining into the parent.
   - When draining the Combiner, we'd unwind the tree by combining child nodes into their parent.

An open question is the precise definitions of "sufficiently different" vs "approximately the same" for comparing node sizes. Presumably we're checking whether sizes are within / outside an order of magnitude multiple.

An important common case is that combine operations which _don't increase the document size_ (e.x. `min/max`, or `merge` of shared properties) will never cause a right-branch to be created, and will be processed using the same left-to-right strategy we use today.

### LINKED ISSUE #270: combine: introduce spill strategies
Today, the combiner maintains document states as a single `BTreeSet` of `serde_json::Value`.

This is a problem for _very_ large transactions, as we typically see with materializations into analytical warehouses:
 * There's a lot of heap utilization since `BTreeSet` is entirely in-memory.
 * Values have many small allocations and amplify overall memory usage, as compared to the equivalent serialized form.
Together this puts a lot of pressure on the Rust heap and memory allocator (jemalloc) while the transaction runs.

## Desired outcome

We'd like to introduce spill strategies to the combiner, where a serde_json::Value is "spilled" to a packed representation (CBOR or MessagePack) that's backed by scratch file(s) and possibly, as an intermediate step, memory. The exact plan should be refined through benchmarking.

We definitely want to design this with #189 in mind, and probably tackle them at the same time. For example, different memory levels could be chunks of memory or file-backed document sets.

### LINKED ISSUE #694: field extraction as from serialized documents
At present, we deserialize documents into `serde_json::Value` in order to query out projected document locations (keys, logical partitions, and materialized fields).

Parsing into Value involves a lot of extra allocation which is not ultimately needed in the common cases of fetching out collection key components, etc. A more efficient implementation is to model extraction as a `serde::Deserializer` which tracks document location and fetches out projected fields within a single deserialization pass, and which avoids modeling portions of the document which are not actually needed.

This is also a way-station along the road to more combiner optimizations, as it allows the combiner to extract keys (necessary for grouping) without needing to commit to parsing the entire document.

jgraettinger:
**Description:**

This (mega-)PR introduces a new `doc::AsNode` trait and makes our fundamental document processing -- including JSON pointers, schema validations, reduction behaviors, and more -- generic over that trait rather than concretely using `serde_json::Value`. A new `HeapNode` type is introduced with greatly improved memory efficiency, which is capable of being efficiently archived into an `ArchivedNode` which supports zero-copy deserialization with random access.

`doc::AsNode` is implemented for all of `HeapNode`, `ArchivedNode`, and `serde_json::Value`. The `Combiner` is updated to use far more efficient (in CPU performance and also memory use) techniques to do its thing, and to give us much tighter bounds & knobs over the amount of memory used. The combiner will make use of disk space and spill compacted "segments" of the combiner state to disk as needed, performing a final sorted heap merge to drain final document states.

This is a foundational improvement which we can explore to reduce JSON deserialization throughout the application (for example, in the shuffled reader), though I leave that as a future endeavor.

Where we started:
```
running 1 test
test combiner_perf has been running for over 60 seconds
Rounds: 1000000
Drained: 543246
Elapsed: 132.48573487s
Memory: active 5424MB allocated 5423MB resident 5491MB retained 198MB alloc_ops 107619999 dealloc_ops 107619964 realloc_ops 2996235
test combiner_perf ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 132.49s
```

Where we ended:

```
running 1 test
Rounds: 1000000
Drained: 543246
Elapsed: 22.766193106s
Memory: active 264MB allocated 263MB resident 529MB retained 571MB alloc_ops 1994170 dealloc_ops 1994135 realloc_ops 367004
test combiner_perf ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 22.84s
```

As an additional data-point, the new `flowctl raw combine` command is able to easily combine over a real-world 43GB test data set drawn from wikipedia recent changes, taking about 14-15min and using a fixed < 300MB amount of memory and 12G of disk space.

Memory usage is constant, while disk and CPU usage **scale linearly** with the input (which cannot be said of the prior implementation).

**Workflow steps:**

This is an under-the-hood improvement with no (intended) user-facing changes.

Performance should be better, with far less memory pressure.

None.

**Notes for reviewers:**

I typically like to rebase and refactor into a final set of commits which are easier for the reader, but this has been such an expansive effort I've left many commits that "show my work".

That may be helpful in understanding how we got here, but it also means early commits don't look _that_ much like late ones.

Fixes #270 
Fixes #189 
Fixes #694 

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/771)


crates/doc/src/combine/mod.rs
```diff
@@ -0,0 +1,266 @@
...
+/// Accumulator is a MemTable paired with a File-backed SpillWriter.
+/// As the caller utilizes the MemTable the Accumulator will transparently
+/// spill table contents to its SpillWriter and then re-initializes a new and
+/// empty MemTable, bounding overall memory usage.
+pub struct Accumulator {
+    memtable: Option<MemTable>,
+    spill: SpillWriter<std::fs::File>,
+}
+
+impl Accumulator {
```
psFried, line 39: I like how the Combiner got broken down into two types that encode the state transitions :+1:

crates/doc/src/combine/mod.rs
```diff
@@ -0,0 +1,266 @@
...
+    }
+}
+
+// Bit flags set on HeapDoc::flags to mark combiner processing status.
+// A the moment there's just one: a bit that indicates the document
+// is fully reduced, and not merely right-combined.
+const FLAG_REDUCED: u8 = 1;
+
+// Smash two documents together.
+fn smash<'alloc>(
```
psFried, line 184: :laughing:

crates/doc/src/combine/mod.rs
```diff
@@ -0,0 +1,266 @@
...
+//    for every segment of the spill file.
+//  - Larger, so that LZ4 can compress it reasonably (we typically see 5x
+//    reduction, 10x is probably the theoretical max).
+//  - Larger, so that filesystem reads and writes are amortized for the
+//    LZ4-compressed.
+//
+// We also want a maximum bound so that SpillDrainer isn't required to
+// hold many large chunks in memory, which could push us over our allocation.
+const CHUNK_TARGET_LEN: usize = 1 << 18; // 256KB.
+const CHUNK_MAX_LEN: usize = 1 << 20; // 1MB.
```
psFried, line 257: Can you explain how large documents (larger than the max chunk length) are handled while spilling to disk? I don't see any problems with it, but just want to make sure I'm understanding the behavior. Would we just end up with a chunk that's larger than the max?

crates/doc/src/combine/mod.rs
```diff
@@ -0,0 +1,266 @@
...
+
+fn serialize_as_display<T, S>(thing: T, serializer: S) -> Result<S::Ok, S::Error>
```
psFried, line 259: optional nit: could maybe be worth pulling in [serde_with](https://docs.rs/serde_with/latest/serde_with/)

crates/doc/src/heap_de.rs
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    fn visit_f64<E>(self, v: f64) -> Result<Self::Value, E>
+    where
+        E: de::Error,
+    {
+        Ok(HeapNode::Float(v))
+    }
+
+    fn visit_str<E>(self, v: &str) -> Result<Self::Value, E>
```
psFried, line 70: I'm wondering whether there might be an easy optimization to avoid copying string data in the case where the JSON deserializer has already allocated a new `String`. At a minimum, I'd expect the deserializer to allocate whenever it needs to parse escape sequences. There's the [`visit_string` function](https://docs.rs/serde/latest/serde/de/trait.Visitor.html#method.visit_string). I don't know the serde-json internals well enough to know that this is worth doing, and I'm also not really sure how much there is to be gained from constructing a `BumpStr` from an owned `String`, so feel free to ignore if that doesn't make sense.

crates/flowctl/src/raw.rs
```diff
@@ -73,18 +79,30 @@ pub struct Rpc {
...
+}
+
 impl Advanced {
     pub async fn run(&self, ctx: &mut crate::CliContext) -> Result<(), anyhow::Error> {
         match &self.cmd {
             Command::Get(get) => do_get(ctx, get).await,
             Command::Update(update) => do_update(ctx, update).await,
             Command::Rpc(rpc) => do_rpc(ctx, rpc).await,
             Command::Bundle(bundle) => do_bundle(ctx, bundle).await,
+            Command::Combine(combine) => do_combine(ctx, combine).await,
```
psFried, line 105: :raised_hands:

crates/flowctl/src/raw.rs
```diff
@@ -123,12 +141,77 @@ async fn do_rpc(
...
+async fn do_combine(
     _ctx: &mut crate::CliContext,
-    Bundle { source }: &Bundle,
+    Combine { source, collection }: &Combine,
 ) -> anyhow::Result<()> {
+    let collection = models::Collection::new(collection);
     let catalog = crate::source::bundle(source).await?;
-    serde_json::to_writer_pretty(std::io::stdout(), &catalog)?;
+
+    let Some(models::CollectionDef{schema: models::Schema::Object(schema), key, .. }) = catalog.collections.get(&collection) else {
```
psFried, line 157: Just as an aside, I'm kinda thinking that `flowctl` ought to have a general way of handling references to entities that allows users to either refer to the definition in a local file (like it is here), or refer to the entity by name and have `flowctl` fetch it from the control plane. Totally cool to leave this code as it is right now, but just checking if we're aligned on the vision for `flowctl` UX.

crates/doc/src/heap.rs
```diff
@@ -0,0 +1,136 @@
...
+    bound(
+        serialize = "__S: rkyv::ser::ScratchSpace + rkyv::ser::Serializer + rkyv::ser::SharedSerializeRegistry"
+    )
+)]
+pub enum HeapNode<'alloc> {
+    Array(#[omit_bounds] BumpVec<'alloc, HeapNode<'alloc>>),
+    Bool(bool),
+    Bytes(BumpVec<'alloc, u8>),
+    Float(f64),
+    NegInt(i64),
```
psFried, line 29: I take it the reason for separate +/- int representations is to allow representing unsigned integers that are larger than the `i64` max?

psFried (approved):
LGTM
I left a few questions, but nothing blocking.

crates/doc/src/combine/mod.rs
jgraettinger: > Would we just end up with a chunk that's larger than the max?

Correct, it maps into chunks of one document each, where the chunk is larger than `CHUNK_MAX_LEN`.
There are some tests that cover this (by using a range like 0..1).
Perhaps `CHUNK_MAX_LEN` should be called something slightly different to convey this property....

crates/doc/src/heap_de.rs
jgraettinger: > I'd expect the deserializer to allocate whenever it needs to parse escape sequences

It does, but typically in a re-used scratch buffer. `simd-json` will actually mutate the input buffer in-place.

> I'm also not really sure how much there is to be gained from constructing a BumpStr from an owned String
 
You can't, because a `BumpStr` invariant is that its buffer is allocated into the bump allocator chunk. This is what allows it to not implement proper Drop behavior without leaking (because we just through the whole bump allocator and all its chunks out).

We also need to account for that String's allocation in the heap, which would be lost if it were outside of the bump allocator chunks.

crates/flowctl/src/raw.rs
jgraettinger: 👍

crates/doc/src/heap.rs
jgraettinger: Correct. The precedent here is our `json::Number` type which is itself modeled on `serde_json::Value`.

`HeapNode` flattens it in this way 'cause it turns out that saves a bunch of space in the archived representation, as `rkyv` isn't smart enough to smash together multiple nested enum discriminants

jgraettinger:
I did a new round of end-to-end testing on a local stack, capturing recent wiki changes and materializing them to Postgres with resets a few times, such that it drained from memory and also from disk. Looked good 👍.

PR #775, opened 2022-11-07
Docs: Firestore backfill mode and data model

oliviamiannone:
**Description:**

Per https://github.com/estuary/connectors/pull/395 and https://github.com/estuary/connectors/pull/409, this updates and enhances the Firestore connector docs. 

* Document new backfill mode option in the binding config
* Add information section on FIrestore's data model and how we handle it (basically paraphrasing @willdonnelly's README).

go.estuary.dev/source-firestore - doesn't need to be modified but it'll be more useful now that we have better info.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/775)

### DIFF
--- a/site/docs/reference/Connectors/capture-connectors/google-firestore.md
+++ b/site/docs/reference/Connectors/capture-connectors/google-firestore.md
@@ -116,16 +116,19 @@ There are three options:
 
 * `none`: Skip preexisting data in the Firestore collection. Capture only new documents and changes to existing documents that occur after the capture is published.
 
-* `sync`: Request that Firestore stream all changes to the collection since its creation, in order.
-
-   This mode provides the strongest guarantee against duplicated data, but can cause errors for large datasets as Firestore may terminate the process after about ten minutes.
-
 * `async`: Use two threads to capture data. The first captures new documents, as with `none`.
-The second progressively ingests historical data in chunks.
-
-   This mode is most reliable for Firestore collections of all sizes but provides slightly weaker guarantees against data duplication.
+The second progressively ingests historical data in chunks. This mode is most reliable for Firestore collections of all sizes but provides slightly weaker guarantees against data duplication.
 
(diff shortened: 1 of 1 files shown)


site/docs/reference/Connectors/capture-connectors/google-firestore.md
```diff
@@ -59,15 +89,43 @@ captures:
...
+In each captured collection's [binding configuration](#bindings), you can choose whether and how to backfill historical data.
+There are three options:
+
+* `none`: Skip preexisting data in the Firestore collection. Capture only new documents and changes to existing documents that occur after the capture is published.
+
+* `sync`: Request that Firestore stream all changes to the collection since its creation, in order.
+
+   This mode provides the strongest guarantee against duplicated data, but can cause errors for large datasets as Firestore may terminate the process after about ten minutes.
+
+* `async`: Use two threads to capture data. The first captures new documents, as with `none`.
```
willdonnelly, line 123: I generally list async before sync, just for that marginal extra bit of nudging users to use async backfills if at all reasonable.

site/docs/reference/Connectors/capture-connectors/google-firestore.md
```diff
@@ -59,15 +89,43 @@ captures:
...
+The second progressively ingests historical data in chunks.
+
+   This mode is most reliable for Firestore collections of all sizes but provides slightly weaker guarantees against data duplication.
```
willdonnelly, line 126: Maybe a dumb suggestion, but merging this sentence with the previous line would make the "async" section only two paragraphs, which might (correctly) make it feel a bit simpler than the "sync" option if we add some additional "here's exactly how that can go wrong" details there.

willdonnelly (approved):
LGTM, a few suggestions that all boil down to "I'm really not even sure if sync should be an option going forward, let's at least make it scarier for users".

PR #776, opened 2022-11-07
Adding mustache to compare template output with Handlebars

travjenkins:
**Description:**

As we are looking to move Oauth towards Mustache we need to make sure the template are outputting the same thing as things could easily fail silently.

**Workflow steps:**

Impact all Oauth calls that use templates.

N/A

**Notes for reviewers:**

N/A

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/776)

### DIFF
--- a/supabase/functions/_shared/helpers.ts
+++ b/supabase/functions/_shared/helpers.ts
@@ -1,20 +1,51 @@
-import { corsHeaders } from "./cors.ts";
+import HandlebarsJS from 'https://esm.sh/handlebars';
+import Mustache from 'https://esm.sh/mustache';
+import { corsHeaders } from './cors.ts';
 
 export const returnPostgresError = (error: any) => {
-  return new Response(JSON.stringify({ error }), {
-    headers: {
-      ...corsHeaders,
-      "Content-Type": "application/json",
-    },
-    status: 400,
-  });
+    return new Response(JSON.stringify({ error }), {
+        headers: {
+            ...corsHeaders,
+            'Content-Type': 'application/json',
+        },
+        status: 400,
+    });
 };
 
 export const handlebarsHelpers = {
-  urlencode: function(s: string) {
-    return encodeURIComponent(s)
-  },
-  basicauth: function(user: string, password: string) {
-    return btoa(`${user}:${password}`);
-  }
-}
+    urlencode: function (s: string) {
+        return encodeURIComponent(s);
+    },
+    basicauth: function (user: string, password: string) {
+        return btoa(`${user}:${password}`);
+    },
+};
+
+export const compileTemplate = (template: string, data: any, connector_id: string) => {
+    (HandlebarsJS as any).registerHelper(handlebarsHelpers);
+
+    const compiledTemplate = (HandlebarsJS as any).compile(template);
+    const handlebarsOutput = compiledTemplate(data);
+
+    try {
+        // Override escaping globally for mustache
+        (Mustache.escape as any) = function (text: any) {
+            return text;
+        };
+
+        //Duplicate for testing Mustache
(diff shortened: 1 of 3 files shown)


supabase/functions/_shared/helpers.ts
```diff
@@ -1,20 +1,51 @@
...
-}
+    urlencode: function (s: string) {
+        return encodeURIComponent(s);
+    },
+    basicauth: function (user: string, password: string) {
+        return btoa(`${user}:${password}`);
+    },
+};
+
+export const compileTemplate = (template: string, data: any, connector_id: string) => {
```
travjenkins, line 24: Move the compile call to a stand alone function to keep logging/compare consistent.

supabase/functions/_shared/helpers.ts
```diff
@@ -1,20 +1,51 @@
...
+            ...handlebarsHelpers,
+        });
+
+        if (handlebarsOutput !== mustacheOutput) {
+            console.log(`oauth:templates:mismatch:${connector_id}`, { template });
+        }
+    } catch (e) {
+        // This try catch is just to be super safe
+        console.log(`oauth:templates:error:${connector_id}`, e);
+    }
```
travjenkins, line 48: The logging of the `template` should end up truncated. I am thinking a truncated template and the `connector_id` should be enough to find it in the DB. However, we could log the full one if folks think that is needed.

Decided to be super safe and just wrap in a try catch. Not really sure what could cause this but do not want to blow up prod on accident with these changes.

supabase/functions/_shared/helpers.ts
```diff
@@ -1,20 +1,51 @@
...
+    (HandlebarsJS as any).registerHelper(handlebarsHelpers);
+
+    const compiledTemplate = (HandlebarsJS as any).compile(template);
+    const handlebarsOutput = compiledTemplate(data);
+
+    try {
+        // Override escaping globally for mustache
+        (Mustache.escape as any) = function (text: any) {
+            return text;
+        };
```
mdibaiee, line 34: I think we can just let Mustache handle escaping as it usually does and if it's an error, then we resolve that as part of our transition, what do you think?

travjenkins: That works for me... cause this does not escape everything exactly the same anyway.

travjenkins: Removed.

mdibaiee (approved):
LGTM, thank you

supabase/functions/_shared/helpers.ts
```diff
@@ -1,20 +1,46 @@
...
+        //Duplicate for testing Mustache
+        const mustacheOutput = Mustache.render(template, {
+            ...data,
+            ...handlebarsHelpers,
+        });
+
+        if (handlebarsOutput !== mustacheOutput) {
+            console.log(`oauth:templates:mismatch:${connector_id}`, { template });
+        }
+    } catch (e) {
```
psFried, line 40: Just want to double check whether this could cause an error to be swallowed. Would our handler still return an error if the template fails to compile? Seems like we'd want it to return a 500 in that case.

psFried (approved):
LGTM % one question

supabase/functions/_shared/helpers.ts
mdibaiee: My understanding is that this will swallow compilation error for the mustache template, but not Handlebars. We want to use Handlebars as the default and Mustache for comparison only at this stage, once we reach a stage where Handlebars and Mustache are consistently giving back the same rendering for all templates, then we will remove Handlebars and use Mustache as the main, removing the `try catch`

jgraettinger:
can this be merged?

PR #777, opened 2022-11-08
supabase/live_specs_ext: avoid duplicate specs due to multiple roles

mdibaiee:
**Description:**

- if a user has multiple rows giving them access to the same resource, we would show duplicate rows in `live_specs_ext`, with this PR, we avoid that by filtering with a common table expression that uses `distinct role_prefix`.
- I checked the performance of this query, and it seems to be as good as the previous one

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/777)

### DIFF
--- a/supabase/migrations/10_spec_ext.sql
+++ b/supabase/migrations/10_spec_ext.sql
@@ -84,6 +84,13 @@ comment on view combined_grants_ext is
 
 -- Extended view of live catalog specifications.
 create view live_specs_ext as
+with user_read_access AS (
+    select
+        distinct role_prefix
+        from internal.user_roles(auth_uid())
+    where
+       capability >= 'read'
+)
 select
   l.*,
   c.external_url as connector_external_url,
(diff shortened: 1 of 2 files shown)


supabase/migrations/10_spec_ext.sql
```diff
@@ -84,6 +84,15 @@ comment on view combined_grants_ext is
 
 -- Extended view of live catalog specifications.
 create view live_specs_ext as
+with user_read_access AS (
+    select
+        distinct role_prefix,
+        max(capability) as capability
```
jgraettinger, line 90: nit: drop `capability` as a selected column (it's not used), as well as the `group by` (implied by distinct)

supabase/tests/live_specs_ext.test.sql
```diff
@@ -0,0 +1,85 @@
...
+        order by role_prefix, capability
+    $i$,
+    $i$ VALUES  ('carolCo/','read') $i$,
+    'carol roles'
+  );
+
+  -- seed live_specs, publication_specs, connectors and connector_tags 
+  delete from publications;
+  insert into publications (id, user_id) values
+    ('01:01:01:01:01:01:01:01', '11111111-1111-1111-1111-111111111111');
```
jgraettinger, line 59: optional: the `:` can be omitted, which I personally find a bit more readable: `'0101010101010101'`

jgraettinger (approved):
LGTM % nit

PR #779, opened 2022-11-08
Increase maxMessageSize to 16MB

psFried:
**Description:**

Increases the maximum size of a message in the capture and materialization protocols, to allow Flow to process individual documents that are up to approximately 16MB. This seems like a reasonable maximum document size, and allows Flow to be useful for processing blockchain data, which tends to occasionally have documents in the 8-12MB range.

**Workflow steps:**

No change to the workflow, but things will now work properly if a user captures documents up to about 16MB.

We don't currently document our maximumm document size, but I think we should.

**Notes for reviewers:**

I tested this locally by capturing a ~9MB JSON document from GCS, and also running it through a derivation.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/779)

### DIFF
--- a/go/connector/run.go
+++ b/go/connector/run.go
@@ -458,5 +458,5 @@ func copyToTempFile(r io.Reader, mode os.FileMode) (*os.File, error) {
 }
 
 const maxStderrBytes = 4096
-const maxMessageSize = 1 << 23 // 8 MB.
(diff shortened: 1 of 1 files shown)


jgraettinger (approved):
LGTM

PR #780, opened 2022-11-08
protocols/airbyte: Support `log.level` anywhere in command line

willdonnelly:
**Description:**

Previously the `log.level` flag was implemented as a property of each subcommand, but that meant that it had to come after the subcommand: `./connector --log.level=debug spec` would fail, for instance.

This commit moves the logging config into the global 'Application Options' option group so that it can be parsed at any position in the command line, and then changes the plumbing for how that gets applied during command execution.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/780)


mdibaiee (approved):

williamhbaker (approved):
LGTM!

PR #784, opened 2022-11-09
Fix agent mising some tasks

jshearer:
**Description:**
This adds three things to the agent, both of which should fix the issue with the agent missing some tasks:
* Polling every 30s, and loudly complaining if actual work was found. There is a wrinkle where we don't want to poll currently active tasks, so I had to add a bit of complexity to handle that. This can be removed once we're confident in `LISTEN`/`NOTIFY`
* Explicitly handling message channel disconnects and explicitly waking up every handler. This way if/when messages are dropped on disconnect, we'll still find them.
* Add a glorious hack to emulate TCP keepalive on the `PgListener` connection, so we can guarantee we know when the connection actually disconnects

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/784)

### DIFF
--- a/crates/agent/src/lib.rs
+++ b/crates/agent/src/lib.rs
@@ -17,7 +17,10 @@ use std::{
     },
     time::Duration,
 };
-use tokio::sync::{mpsc, Mutex};
+use tokio::sync::{
+    mpsc::{self, UnboundedSender},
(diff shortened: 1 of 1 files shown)


psFried (approved):
LGTM

PR #785, opened 2022-11-09
docs: s3 bucket policy

oliviamiannone:
**Description:**

Significantly enhances source-s3 docs to have proper setup steps and samples of AWS policy documents. 

Separate steps and samples are included for 

* allowing a specific IAM user to access a bucket (adapted from a customer's recent experience)
* making a bucket public with a policy that allows anonymous reads (tested by me)

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/785)

### DIFF
--- a/site/docs/reference/Connectors/capture-connectors/amazon-s3.md
+++ b/site/docs/reference/Connectors/capture-connectors/amazon-s3.md
@@ -1,6 +1,9 @@
 ---
 sidebar_position: 2
 ---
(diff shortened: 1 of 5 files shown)


jshearer (approved):
This all looks legit to me. The IAM policies don't include `s3:HeadObject`, but I'm 95% sure we determined that `HeadObject` is implied if you have `GetObject`, so that shouldn't be a problem

PR #786, opened 2022-11-10
runtime: swap Derive.Destroy order to prevent use-after-free of npm server

jgraettinger:
**Description:**

The trampoline server of a derive binding may be running concurrent tasks up until the point that binding.Destroy returns. Those tasks may be using the npm / TypeScript server which is managed by the `taskTerm`, so we must destroy the former before the latter.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/786)


dhammika (approved):
:+1:

PR #788, opened 2022-11-10
README update

oliviamiannone:
**Description:**

The README was written back when we were talking about Flow more as a self-hosted solution, and perhaps didn't have as many other resources or docs available. 

The state of affairs has changed now -- we develop in the open but are primarily offering Flow as a managed platform. So, we need to talk about things a bit different. 

This update ...
* emphasizes using the UI and CLI with a free account, not building locally.
* pares down the in-depth conceptual discussion -- it's good stuff but long and is covered in our docs, which we link to. 
* removes outdated "Run it" section.
* removes long YAML spec sample. I don't think it's necessary at this stage, in this location, but I am very open to disagreement here.
* adds more resources- contact options, connector requests, tutorial links, etc

**Workflow steps:**

Pretend you are a potential user or customer. Read the README.

- If you want to try Flow asap, are you satisfied?
- If you're evaluating Flow from a practical business/manager side, did you learn what you need?
- If you're an engineer, is anything lacking (specifically, is the lack of code samples and easy "run it from the command line" section jarring?)

**Notes for reviewers:**

The Build on M1 section seems out of place with the rest of the README now, but this also seems like the only logical place for it to be. Open to suggestions.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/788)


dhammika:
Thanks for adding more example links! 
Option to run it locally without external dependencies was super useful in trying out flow. Perhaps we should stash that content in a separate doc ie. control-plane has a separate readme.

oliviamiannone:
> Thanks for adding more example links! Option to run it locally without external dependencies was super useful in trying out flow. Perhaps we should stash that content in a separate doc ie. control-plane has a separate readme.

@dhammika I thought that section was actually out of date - is that not the case?
Either way, it's definitely important for us to retain anything that's helpful for our dev team without confusing users who would be better off using our SaaS. Stashing in a different README is a good solution.

oliviamiannone:
Old "run it" section has been ported over to control plane readme, along with the instructions for building on M1. It may be worth noting that I tried to follow the latter to test the former, but my M1 seems to have other tricks up its sleeves. Prioritizing getting this in over continuing to test.

PR #789, opened 2022-11-10
refactor: skip running expensive tasks (npm, tests) if they're not neccesary

### LINKED ISSUE #761: faster builds: only invoke NPM when required

jshearer:
In addition to being smart about detecting when to run NPM, I also realized that we were naively always running tests, even if there weren't any tests to run. Skipping this early shaved off ~50% from my build time locally

Fixes #761

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/789)

### DIFF
--- a/crates/build/src/lib.rs
+++ b/crates/build/src/lib.rs
@@ -57,11 +57,6 @@ where
         load_and_validate(root_url.clone(), root_spec, fetcher, drivers, &config).await;
     all_tables.meta.insert_row(config.clone());
 
-    // all_tables will tell us if we have any derivations/NPM modules
-    // places to look:
-    // Resources table: of type npm_module or npm_package for each build
-    // derivations that define a typescript module
-
     let has_typescript_derivations = all_tables
         .derivations
         .iter()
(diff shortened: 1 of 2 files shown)


crates/build/src/lib.rs
```diff
@@ -56,13 +57,26 @@ where
...
+    let typescript_enabled = has_typescript_derivations || has_npm_resources;
+
     // Output database path is implied from the configured directory and ID.
     let output_path = directory.join(&config.build_id);
     let db = rusqlite::Connection::open(&output_path).context("failed to open catalog database")?;
 
     // Generate TypeScript package? Generation should always succeed if the input catalog is valid.
     if all_tables.errors.is_empty()
         && (config.typescript_generate || config.typescript_compile || config.typescript_package)
+        && typescript_enabled
```
mdibaiee, line 79: nit on naming: I read this condition first, and I thought there is a flag somewhere for enabling typescript, and then I read the definition and realised this means that there are either some derivations or npm resources that need to be built, maybe we can rename this to give that message, something like `has_typescript_resources`?

psFried (approved):
LGTM

crates/build/src/lib.rs
jshearer: Yup, great point. Will do

mdibaiee (approved):
Thank you!

PR #791, opened 2022-11-10
fix: Update agent startup logging

### LINKED ISSUE #790: update agent startup logging
Evaluate what the agent is logging on startup and update as needed.

jshearer:
This change completely omits logging _certain_ fields. We could use a custom `Debug` impl instead, but that would mean that we could get out of sync when adding or removing fields from the `Args` struct, which felt smelly.

Alternatively, we could just not log the args at all... I don't really have an opinion on whether it's useful here or not.

Fixes #790

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/791)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -35,6 +35,7 @@ dependencies = [
  "bytes",
  "chrono",
  "clap 3.2.22",
+ "derivative",
  "futures",
  "insta",
  "itertools",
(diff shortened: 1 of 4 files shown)


mdibaiee (approved):

psFried (approved):
LGTM, though I do have one optional nit pick. IMO it's not worth the cost in terms of compile times to bring in a new derive macro dependency just for redacting a single field. IMO either the newtype pattern or a custom Debug impl might be preferable. I honestly don't feel very strongly about it, though, if you're pretty fond of derivative.

PR #796, opened 2022-11-14
ops-catalog: handle cross-tenant stats

williamhbaker:
**Description:**

This allows for the stats materialization to handle flows with "cross-tenant" stats where for example a capture from tenant A is producing documents into a collection of tenant B. Previously this would not work since the stats materialization was materializing directly into the partition for the tenant of the task, and the inverted collection documents for the other tenant are not allowed into that partition.

Now, the stats materialization will materialize into the main `catalog_stats` table, allowing postgres to automatically route the records into the correct tenant partition.

As I've proposed this, the `stats_loader` will own the `catalog_stats` table and will no longer use the `catalog_stat_partitions` schema. The `flow_materializations_v2` and `flow_checkpoints_v1` tables will now be in the `public` schema, rather than the `catalog_stat_partitions` schema. Because of this, I think the easiest migration path would be to delete & re-provision existing tenants similar to what was done in https://github.com/estuary/flow/pull/755, with the extra twist that this time we will want to re-use the task names, so hopefully there is a way to do that.

**Workflow steps:**

I tested this out locally by creating a cross-tenant flow. One tenant was provisioned, and the other was not. The stats materialization initially failed because there was not a partition for the "other" tenant. After creating the partition, the materialization was able to proceed. I was also able to later successfully provision the "other" tenant and have stats correctly populated into the partition.

N/A

**Notes for reviewers:**

~~A migration script for prod is still TODO~~ - Done

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/796)


supabase/migrations/11_stats.sql
```diff
@@ -63,13 +63,15 @@ comment on column catalog_stats.flow_document is
 do $$
 begin
     if not exists (select from pg_catalog.pg_roles where rolname = 'stats_loader') then
-        create role stats_loader with login password 'stats_loader_password';
+        create role stats_loader with login password 'stats_loader_password' bypassrls;
    end if;
 end
 $$;
 
+-- stats_loader loads directly to the catalog_stats table. Postgres routes records to the correct
```
jgraettinger, line 71: Can we leave the owner as postgres? It seems strange that everything _but_ this is owned by postgres. The bypassrls attribute should allow stats_loader to load into it without invoking the RLS policy.

We'd still need to grant insert / select / update on catalog_stats to stats_loader.

supabase/tests/stats.test.sql
```diff
@@ -34,7 +29,7 @@ begin
       5, 1,
       0, 0
     );
-  insert into catalog_stat_partitions.bob_stats (
+  insert into catalog_stats (
```
jgraettinger, line 32: nit: insert multiple `values` tuples rather than re-doing the whole `insert into` stanza.

jgraettinger (approved):
LGTM % ownership of catalog_stats below

supabase/migrations/11_stats.sql
williamhbaker: I had to change the owner to be `stats_loader`, otherwise the materialization(s) will fail when they try to add comments to the table if the user is not the owner of the able. I also think it is strange for postgres not to own this table though so I'll see if I can figure something else out.

williamhbaker: The handling of the target table is in the [connector code](https://github.com/estuary/connectors/blob/24c0a6ef9c0dd523121e66250c00b0362ad43f15/materialize-postgres-rc/sqlgen.go#L71-L92). Adding table & column comments happens no matter what, even if the target table already exists. Ideally I think the connector would leave the table & column comments alone if the target table already exists, but I don't know if this can be easily done. It also doesn't look like there is any way to get around the user being the owner of the table for adding comments, see https://www.postgresql.org/docs/current/sql-comment.html

I could imagine the connector providing an option to user a different templated query for not adding comments. That doesn't seem great either, but maybe the best we can do. What do you think @jgraettinger ?

jgraettinger: Yep I agree with your conclusion. I think the answer for my opening question is "no". Mind updating the migration with a comment summarizing why we're deviating from the norm here? Then LGTM

PR #797, opened 2022-11-14
docs: documents and size

oliviamiannone:
**Description:**

Documentation about documents :) 

Per https://github.com/estuary/flow/pull/779, notes maximum size of documents in collections, captures, and materializations concept docs. 

I realized that previously, we never explicitly stated plainly for our less-technical users that "collections are made up of documents and documents are..." I amended this by adding a small section to the top of the Collections page. 

**Notes for reviewers:**

If there's any other pertinent information we should add about this topic please let me know!

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/797)

### DIFF
--- a/site/docs/concepts/collections.md
+++ b/site/docs/concepts/collections.md
@@ -29,15 +29,16 @@ Flow allows documents up to 16 MB in size, but it's rare for documents to approa
 An example document for a collection with two fields, `name` and `count` is shown below.
 
 ```json
-root:{
-  _meta:{
-    uuid:"xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx"
-  }
-  count:5954
-  message:"Hello #5954"
+{
+  "_meta": {
+    "uuid": "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx"
+  },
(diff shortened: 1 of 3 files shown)


site/docs/concepts/collections.md
```diff
@@ -15,16 +15,40 @@ you define one or more new collections as part of that process.
...
+```json
+root:{
+  _meta:{
+    uuid:"xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx"
+  }
+  count:5954
+  message:"Hello #5954"
+```
+
+The `_meta` object is present in all Flow documents and is used for internal processing only.
```
psFried, line 40:
```suggestion
The `_meta` object is present in all Flow documents, and contains metadata added by Flow. Minimally, every document `_meta` always has a `uuid`, which is a globally unique id for each document. Some capture connectors may add additional `_meta` properties to tie each document to a specific record within the source system. Documents that were captured from cloud storage connectors, for example, will contain `/_meta/file` and `/_meta/offset` properties that tell you where the document came from within your cloud storage bucket.
```

psFried (approved):
Had a few suggestions, but LGTM

PR #802, opened 2022-11-16
fix: avoid deadlock preventing `LISTEN` keepalive loop from running

jshearer:
**Description:**

Because `try_recv` blocks (asynchronously) while waiting for a notification, we were holding the `Mutex` lock open. This prevented the keepalive loop from poking the connection, since it needs to acquire the lock in order to do its job. The solution I came up with is to periodically release the lock held by `try_recv`, so that the keepalive loop can run.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/802)

### DIFF
--- a/crates/agent/src/lib.rs
+++ b/crates/agent/src/lib.rs
@@ -67,13 +67,9 @@ async fn listen_for_tasks(
     table_names: Vec<String>,
     pg_pool: &sqlx::PgPool,
 ) -> anyhow::Result<()> {
-    let listener = Arc::new(Mutex::new(PgListener::connect_with(&pg_pool).await?));
+    let mut listener = PgListener::connect_with(&pg_pool).await?;
 
-    listener
-        .lock()
-        .await
-        .listen(AGENT_NOTIFICATION_CHANNEL)
-        .await?;
+    listener.listen(AGENT_NOTIFICATION_CHANNEL).await?;
 
     // Sqlx does not give us the option to set TCP keepalive on its connections,
     // and this specifc scenario of keeping a long-running connection open to listen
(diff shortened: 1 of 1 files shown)


crates/agent/src/lib.rs
```diff
@@ -107,39 +111,53 @@ async fn listen_for_tasks(
...
-                        table_name: table.clone(),
-                        is_poll: false,
-                    })
-                    .unwrap()
-            });
+        // We need to release the lock on listener every once in a while so that the
+        // keepalive loop above can do its thing, otherwise we'll hold the lock open
+        // waiting on try_recv, which may never complete because the connection is
+        // unknowningly disconnected, which is the whole point of the keepalive loop.
+        match tokio::time::timeout(Duration::from_secs(30), async {
```
jgraettinger, line 118: I think Mutex is the wrong tool for the job here, and what you really want is `tokio::select!`.

You can have a single select where you're blocking for the first of _either_ a `NOTIFY` _or_ for a its-time-to-subscribe-and-unsubscribe-timer to elapse (upon which you'd synchronously subscribe / unsubscribe).

Then, remove the tokio::spawn loop above and `cloned_listener` altogether.

jshearer: Yuuuuuuup that does make more sense

jshearer:
@jgraettinger does that look better?

crates/agent/src/lib.rs
```diff
@@ -17,9 +17,12 @@ use std::{
     },
     time::Duration,
 };
-use tokio::sync::{
-    mpsc::{self, UnboundedSender},
-    Mutex,
+use tokio::{
+    sync::{
+        mpsc::{self, UnboundedSender},
+        Mutex,
```
jgraettinger, line 23: nit: is Mutex still used?

crates/agent/src/lib.rs
```diff
@@ -86,60 +85,50 @@ async fn listen_for_tasks(
...
-                    })
-                    .unwrap()
-            });
+        let should_poke_connection: Result<bool, anyhow::Error> = tokio::select! {
+            res = async {
+                tokio::time::sleep(Duration::from_secs(30)).await;
+                Ok(true)
+            } => {res},
+            res = async {
+                let maybe_item = listener.try_recv().await.map_err(|e| anyhow::Error::from(e))?;
```
jgraettinger, line 95: nit: I don't think you need map_err here and elsewhere. `?` should be doing that for you.

crates/agent/src/lib.rs
```diff
@@ -86,60 +85,50 @@ async fn listen_for_tasks(
...
-            tracing::warn!("LISTEN/NOTIFY stream from postgres lost, waking all handlers and attempting to reconnect");
-            table_names.iter().for_each(|table| {
-                task_tx
-                    .send(HandlerInvocation {
-                        table_name: table.clone(),
-                        is_poll: false,
-                    })
-                    .unwrap()
-            });
+        let should_poke_connection: Result<bool, anyhow::Error> = tokio::select! {
```
jgraettinger, line 89: Okay, a few things:

I've universally found it's better to have as little code inside of a select! as possible. It gets wonky from a rust-analyzer & IDE perspective, type hinting doesn't work well, etc. Plus, it's more confusing to read.

* You can extract the tokio timer to a variable.
* How about a mut bool `should_poke` defined _outside_ of the loop, and handling for that being true at the _top_ of the loop?

Doing these to things gets you to:

```
let timer = tokio::time::sleep(Duration::from_secs(30));
// tokio::pin! as needed
let notify = tokio::select! {
  _ = timer => {
    should_poke = true;
    continue;
  },
  notify = listener.try_recv() => notify,
};

// Handle `notify`.

```

I'll further suggest peeling off the error case below as:

```
let Some(item) = maybe_item else {
  tracing::warn!("LISTEN/NOTIFY stream from postgres lost ...
  ... handling ...
  continue;
}

// Now you can handle `item` as more of a through-line for the reader.
```

crates/agent/src/lib.rs
jgraettinger: You may want to use anyhow::Context to give more context for exactly where the error is coming from. Sometimes, if the error itself is terse, it can be hard to map it back to the code path it took to be bubbled up and I find `.context()` pretty invaluable there.

crates/agent/src/lib.rs
jshearer: [Yup!](https://github.com/estuary/flow/pull/802/files#diff-c89117f7bd70bccf98ba0d48e2ed19e740a8c59c9d1b33cfe33fbc01cd8ebd7eL211)

crates/agent/src/lib.rs
jshearer: I can honestly never tell when I need this map_error and when I don't. You're right, I don't need it here after all... but it certainly fixed a compiler error at some point /shrug

jgraettinger (approved):
LGTM

PR #804, opened 2022-11-17
agent: enforce illegal-tenant-names in beta_onboard

mdibaiee:
**Description:**

- Mark a set of words as illegal along with a reason, and don't allow creation of tenants with these names
- I included the profanity and Fortune 500 list provided by Dave as well as some reserved words that I thought were appropriate for our system. It would be good to get a list of words you think we might need in the future to add to the list. Specially @kiahna-tucker and @travjenkins if there are route names you think we might need in the future that I haven't included.

**Workflow steps:**

```
supabase db reset
```

Open http://localhost:3000/?showBetaOnboard=please_show

Try one of the included words

**Notes for reviewers:**


crates/agent/src/directives/illegal_tenant_names.rs
```diff
@@ -0,0 +1,1481 @@
+use phf::phf_map;
+
+pub const ILLEGAL_TENANT_NAMES: phf::Map<&str, &str> = phf_map! {
```
jgraettinger, line 3: We need this to be managed in the DB, rather than in source code, as we need to be able to update this list and/or react quickly without having to do a code push / build / release.

We had talked about just creating sentinel entries in the tenants table with zero quota which prevent others from creating a tenant of the same name (case invariant). Another option would be to create this as a table in the `internal` schema and update our logic to check it.

jgraettinger: I think I'll push a little more strongly for a separate "block list" table in the `internal` schema of the DB.
While we _could_ put these it into `tenants` directly, I'm just thinking about the potential for confusion with future operations people who might not _entirely_ know what they're doing.

crates/agent/src/directives/beta_onboard.rs
```diff
@@ -36,6 +38,11 @@ pub async fn apply(
...
         )));
     }
+    if let Some(reason) = illegal_tenant_names::ILLEGAL_TENANT_NAMES.get(&requested_tenant.as_str().to_lowercase()) {
+        return Ok(JobStatus::invalid_directive(anyhow::anyhow!(
+            "Cannot provision a new tenant with the requested name: {}.", reason
+        )));
+    }
     if agent_sql::directives::beta_onboard::is_user_provisioned(row.user_id, &mut *txn).await? {
         return Ok(JobStatus::invalid_claims(anyhow::anyhow!(
             "Cannot provision a new tenant because the user has existing grants",
```
jgraettinger, line 48: Subject to the block list being contained in the DB, I'd suggest updating the semantics of `agent_sql::directives::beta_onboard::tenant_exists` to use it.

We do want to leave it ambiguous as to whether a tenant is disallowed because of the block list, or because it actually already exists.

jgraettinger (approved):
LGTM

PR #805, opened 2022-11-17
fix flowctl logs|stats help messages

### LINKED ISSUE #726: Confusing help output for flowctl stats --task
`flowctl stats` was made available in https://github.com/estuary/flow/pull/707/. 

It appears that the help messages for `logs` and `stats` are pulled from the same place, which works except for _one thing:_ when you print the help for flowctl _stats_, for  --task, it says "Read the _logs_ of the task with the given name"

psFried:
**Description:**

fixes #726 

The new ouput of `flowctl logs --help`:

```
flowctl-logs
Prints the runtime logs of a task (capture, derivation, or materialization).

Reads contents from the `ops/<tenant>/logs` collection, selecting the partition that corresponds to
the selected task. This command is essentially equivalent to the much longer: `flowctl collections
read --collection ops/<tenant>/logs --include-partition estuary.dev/field/name=<task> --uncommitted`

USAGE:
    flowctl logs [OPTIONS] --task <TASK>

OPTIONS:
        --task <TASK>
            The name of the task

        --follow
            Whether to block waiting for new data. If provided, the read will continue indefinitely
            until interrupted or ending due to an error

    -o, --output <OUTPUT>
            How to format CLI output

            [possible values: json, yaml, table]

        --since <SINCE>
            Start reading from approximately this far in the past. For example `--since 10m` will
            output all data that was added within the last 10 minutes. The actual start of the read
            will always be at a fragment boundary, and thus may include data from significantly
            before the requested time period

    -h, --help
            Print help information
```

And the new output from `flowctl stats --help`:

```
flowctl-stats
Prints the runtime stats of a task (capture, derivation, or materialization).

Reads contents from the `ops/<tenant>/stats` collection, selecting the partition that corresponds to
the selected task. This command is essentially equivalent to the much longer: `flowctl collections
read --collection ops/<tenant>/stats --include-partition estuary.dev/field/name=<task>`

USAGE:
    flowctl stats [OPTIONS] --task <TASK>

OPTIONS:
        --task <TASK>
            The name of the task

        --follow
            Whether to block waiting for new data. If provided, the read will continue indefinitely
            until interrupted or ending due to an error

    -o, --output <OUTPUT>
            How to format CLI output

            [possible values: json, yaml, table]

        --since <SINCE>
            Start reading from approximately this far in the past. For example `--since 10m` will
            output all data that was added within the last 10 minutes. The actual start of the read
            will always be at a fragment boundary, and thus may include data from significantly
            before the requested time period

        --uncommitted
            Read raw data from stats journals, including possibly uncommitted or rolled back
            transactions. This flag is currently required, but will be made optional in the future
            as we add support for committed reads, which will become the default

    -h, --help
            Print help information
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/805)

### DIFF
--- a/crates/flowctl/src/collection/read/mod.rs
+++ b/crates/flowctl/src/collection/read/mod.rs
@@ -29,7 +29,7 @@ pub struct ReadArgs {
 /// Common definition for arguments specifying the begin and and bounds of a read command.
 #[derive(clap::Args, Debug, Default, Clone)]
 pub struct ReadBounds {
-    /// Whether to block waiting for new data. If provided, the read will continue indefinitely until interrupted or ending due to an error.
+    /// Continue reading indefinitely until interrupted or ending due to an error.
     #[clap(long)]
     pub follow: bool,
 
--- a/crates/flowctl/src/lib.rs
+++ b/crates/flowctl/src/lib.rs
@@ -1,6 +1,7 @@
 use std::fmt::Debug;
 
 use anyhow::Context as _;
+use clap::AppSettings;
 use clap::Parser;
 use proto_flow::flow;
 
@@ -22,7 +23,12 @@ use poll::poll_while_queued;
 
 /// A command-line tool for working with Estuary Flow.
 #[derive(Debug, Parser)]
-#[clap(author, about, version)]
+#[clap(
+    author,
+    about,
+    version,
+    global_setting = AppSettings::DeriveDisplayOrder
+)]
 pub struct Cli {
     /// Configuration profile to use.
     ///
@@ -32,11 +38,11 @@ pub struct Cli {
     #[clap(long, default_value = "default")]
     profile: String,
 
-    #[clap(flatten)]
-    output: Output,
-
     #[clap(subcommand)]
     cmd: Command,
+
+    #[clap(flatten)]
+    output: Output,
 }
 
 #[derive(Debug, clap::Subcommand)]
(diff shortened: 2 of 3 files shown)


oliviamiannone (commented):
Mostly looks good to me! A few suggestions:
* The sentence “this command is essentially equivalent to…” seems unnecessary to me, and may even be distracting users (?)
* The first sentence for —follow (“whether to block waiting for new data”) was hard for me to get my brain around. The second sentence is what allowed me to understand what’s going on — ie, if I provide the flag, the command won’t terminate and will stay open and show me new logs/stats as they come in. Assuming I’m understanding correctly, replacing that sentence with something like “Continue streaming new logs/stats.” could make more sense.

psFried:
Thanks, I updated the help for `--follow`.

> The sentence “this command is essentially equivalent to…” seems unnecessary to me, and may even be distracting users (?)

The goal of having this was to help clear up confusion stemming from there being two separate ways of reading logs collections. Since `flowctl logs` is just a friendlier facade over `collections read`, it seemed helpful to be transparent about that so as not to clash with users' conceptual models of Flow. That being said, I'll definitely trust your judgement here if you still feel like it'd be better to remove it. WDYT?

oliviamiannone:
>The goal of having this was to help clear up confusion stemming from there being two separate ways of reading logs collections. Since flowctl logs is just a friendlier facade over collections read, it seemed helpful to be transparent about that so as not to clash with users' conceptual models of Flow. That being said, I'll definitely trust your judgement here if you still feel like it'd be better to remove it. WDYT?

That's fair. I'm certainly not a frequent CLI user so my POV here probably isn't the most representative. Flow power users / engineering types will likely benefit from this. Let's leave it in!

PR #811, opened 2022-11-19
Phil/new tenant limits

psFried:
**Description:**

This adds the ability to limit the number of new tenants that can be created through the betaOnboarding directive. It does this by enforcing optional limits on the number of times that each directive may be applied. The default is to allow unlimited applications of each directive. If the `uses_remaining` column is non-null, then it is decremented each time the directive is applied, and returns an error if it is negative.

**Workflow steps:**

Not really a user-facing workflow, but here's the steps for a dev:

- To setup a limit of `n` new tenants, first find the `id` of the beta onboarding directive.
- In the DB: `update directives set uses_remaining = <n> where id = <directive-id>;`
- Login with a new account, and navigate to: `/auth?betaOnboarding=please_show`.
- Enter a prefix and hit the CTA.
- If `n` is greater than 0, then the tenant will be created. Otherwise, you will see this error:
- `n` will be decremented on every attempt to apply the directive, so a negative number immediately tells you the number of times an attempt was denied.

image

**Notes for reviewers:**

I went back and forth on the approach for this. The first commit totally works, but requires an index and an additional query. The counter-based approach also seems like it'll be easier for us to manage as a rate-limiter. The idea being to have a cron job that executes roughly the following SQL every hour, for example, if we wanted a rate-limit of 100 new tenants per hour.

```sql
update directives set uses_remaining = 100 where directive_id = '<betaOnboardDirective>';
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/811)

### DIFF
--- a/supabase/migrations/15_directives.sql
+++ b/supabase/migrations/15_directives.sql
@@ -4,7 +4,7 @@ create table directives (
   like internal._model including all,
 
   catalog_prefix  catalog_prefix not null,
-  single_use      boolean not null default false,
+  uses_remaining bigint,
   spec            jsonb_obj not null,
   token           uuid unique default gen_random_uuid(),
 
@@ -47,8 +47,11 @@ Catalog prefix which contains the directive.
 Operations undertaken by a directive are scoped within the catalog prefix,
 and a user must admin the named prefix in order to admin its directives.
 ';
-comment on column directives.single_use is '
-Single-use directives have their token disabled after its first use.
+comment on column directives.uses_remaining is '
+The maximum number of times that this directive may be applied.
+This value gets decremented each time the directive is applied.
+Once it reaches 0, future attempts to apply the directive will fail.
+A null here means that there is no limit.
 ';
 comment on column directives.spec is '
 Specification of the directive.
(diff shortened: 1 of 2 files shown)


jgraettinger:
> instead be a nullable integer of remaining quota (if it's NULL, then quota is unlimited).

I see it's using an absurdly large into to avoid handling null vs not-null distinctions -- that's totally cool too.

psFried:
> I think it should be directly incorporated into the directives table and replace the current single_use mechanism within the implementation of exchange_directive_token

@jgraettinger This seems obvious to me in retrospect, so I'm on board with doing it. I started to implement it, and realized there's kind of an issue with error messages. This trigger-based implementation just uses a hard-coded error message, but that message seems like a poor fit for single-use directives. If the intent is to use single-use directives for things like sending someone an invite to your tenant, then I think we'd want the error message to say something like `This invite has already been accepted`.

So I'm thinking it makes sense to add the error message as another column on the `directives` table. Then if we run out of quota we use the error message for that specific directive. Does that seem reasonable to you?

jgraettinger:
I think we can have a single structured error that the ui can interpret based on the user's context. Each of these directives require bespoke ui work anyway.

psFried:
> I think we can have a single structured error that the ui can interpret based on the user's context. Each of these directives require bespoke ui work anyway.

@jgraettinger  Ok, I left the existing error message, because it works for the use cases we have right now. PTAL

jgraettinger (approved):
LGTM!

PR #818, opened 2022-11-28
capture ACKs are made explicit

jgraettinger:
**Description:**

Make capture acknowledgements an explicit, negotiated portion of the protocol.

Unless a capture opts-in to receiving acks, they will never be sent and the stdin of the connector is immediately closed.

**Workflow steps:**

Capture connectors no longer need to read from stdin.

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/818)


mdibaiee (approved):
LGTM! I did a quick grep on `connectors` and I don't see any capture connectors relying on Acknowledgements, so I think we are good 👍🏽

PR #821, opened 2022-11-30
Phil/flowctl creds

psFried:
**Description:**

This does some significant refactoring around how `flowctl` handles credentials. The main goals of this are to store user credentials more securely using the keychain service provided by the OS, and reduce some friction in the end-to-end process of authenticating the CLI.

This compliments https://github.com/estuary/ui/pull/391 and uses the new `/cli-auth/login` endpoint that was introduced there. The `flowctl auth login` command was added in order to make setting up `flowctl` authentication both obvious and discoverable from the CLI.

The credentials themselves are moved out of the config file, and into the users' keyring. While it's rare for an operating system to not have one, there are of course cases where that will be the case. As a "catch all" to support systems without the ability to use the keyring, the `FLOWCTL_ACCESS_TOKEN` environment variable was added. If set, then the value is used directly, and flowctl will skip any access of the keychain.

The `auth develop` subcommand was also removed. The intended replacement is to use `flowctl auth login` with local installations. The second commit tries to alleviate some of the pain that might cause for developers running `flowctl` against local instances.

**Workflow steps:**

The recommended login steps for `flowctl` will change to:

- Run `flowctl auth login`, which will open your browser to the login page (this respects the `--profile`)
- Login in the browser
- Copy the auth token you're presented with once done
- Paste that token into the terminal where `flowctl` is waiting.
- Profit

Note that `flowctl auth token --token` has not, and will not be changed. An alternative workflow, which will continue to be supported is for the user to navigate to `/cli-auth/login` on their own, and paste the resulting token into `flowctl auth token --token ...`. The `flowctl auth login` approach is functionally no different, and exists primarily to be discoverable in the CLI.

**Notes for reviewers:**

- That first commit has kind of a lot going on, so just LMK if you want a tour.
- There's a complimentary UI PR at estuary/ui#391
- I've done a fair bit of testing locally, but I won't be very confident until the UI PR is merged so that I can test this with oauth logins.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/821)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -25,6 +25,18 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "aae1277d39aeec15cb388266ecc24b11c80469deae6067e17a1a7aa9e5c1f234"
 
 [[package]]
+name = "aes"
+version = "0.7.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9e8b47f52ea9bae42228d07ec09eb676433d7c4ed1ebdf0f1d1c29ed446f1ab8"
+dependencies = [
+ "cfg-if",
+ "cipher",
+ "cpufeatures",
+ "opaque-debug",
+]
+
+[[package]]
 name = "agent"
 version = "0.0.0"
 dependencies = [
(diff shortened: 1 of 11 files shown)


psFried:
Closing this because it doesn't seem compatible with subabase. The supabase-js library doesn't allow for multiple sessions to co-exist in the same browser. And gotrue has no support for auth redirects that use query parameters. It exclusively passes credentials as part of the fragment (after the `#`), which is not sent to the server.

Whenever we're ready to fix auth for the CLI, we should check back in on supabase/supabase#7186 and possibly also take a look at [this workaround](https://gist.github.com/FelixZY/0aef530690458b381b8100afa19202c8) for managing api keys in SQL.

PR #822, opened 2022-11-30
fix: Handle the end of a fragment properly

jshearer:
Previously we'd just sit in an infinite loop :facepalm:

The actual fix is here: https://github.com/estuary/flow/compare/bugfix/schema_inference_fragment_end?expand=1#diff-8a4f203851e6710c1af18f044b54a78c49ddb6eedaecf6aa5d4804fdf2c58187R261

This PR also introduces a bit of plumbing to allow for logging the total number of bytes read when inferring a schema.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/822)

### DIFF
--- a/crates/schema-inference/src/json_decoder.rs
+++ b/crates/schema-inference/src/json_decoder.rs
@@ -88,7 +88,6 @@ where
                 let offset = iter.byte_offset();
                 self.bytes += offset;
                 buf.advance(offset);
-                // tracing::trace!(bytes_advance = offset, "Successfully read document");
 
                 Ok(Some(v))
             }
--- a/crates/schema-inference/src/json_decoder.rs
+++ b/crates/schema-inference/src/json_decoder.rs
@@ -8,6 +8,7 @@ use tokio_util::codec::Decoder;
 
 #[derive(Default)]
 pub struct JsonCodec<Dec = Value> {
+    bytes: usize,
     _dec: PhantomData<Dec>,
 }
 
@@ -17,7 +18,14 @@ where
 {
     /// Creates a new `JsonCodec` with the associated types
     pub fn new() -> JsonCodec<Dec> {
-        JsonCodec { _dec: PhantomData }
+        JsonCodec {
+            bytes: 0,
+            _dec: PhantomData,
+        }
+    }
+
+    pub fn bytes_read(&self) -> usize {
+        self.bytes
     }
 }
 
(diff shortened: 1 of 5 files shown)


crates/schema-inference/src/server.rs
```diff
@@ -231,38 +257,66 @@ async fn fragment_to_shape(
...
 
-    while let Some(shape) = stream.next().await {
-        match shape {
-            Ok(Some((inferred_shape, docs_read))) => {
-                if let Some((accumulated_shape, docs_count)) = accumulator {
-                    accumulator = Some((
-                        shape::merge(accumulated_shape, inferred_shape),
-                        docs_count + docs_read,
-                    ))
+    while let Some(item) = stream.next().await {
```
psFried, line 296: nit: this looks like a job for [`fold`](https://docs.rs/futures/latest/futures/stream/trait.StreamExt.html#method.fold)

crates/schema-inference/src/server.rs
```diff
@@ -188,14 +207,23 @@ async fn journal_to_shape(
 async fn fragment_to_shape(
     client: journal_client::Client,
     fragment: fragments_response::Fragment,
-    abort_signal: Receiver<()>,
-) -> Result<Option<(Shape, u64)>, InferenceError> {
+    abort_signal: Receiver<bool>,
+) -> Result<Option<ShapeAndMeta>, InferenceError> {
+    tokio::pin!(abort_signal);
+    // Let's bail early if we're already aborted
+    if *abort_signal.borrow_and_update() {
```
psFried, line 214: uber nit: This is completely fine, but it does rely on some kinda tricky Rust behavior that's not especially well documented. This works because the compiler drops temporary values for `if` conditions immediately after evaluating the condition. But if you made a seemingly innocuous change, such as changing the `if` to a `match`, then the temporary `Ref` could easily end up living _longer_. IMO it's best to err on the side of explicitness with synchronization types, so maybe something like:

```rust
fn is_aborted(rx: &mut Receiver<bool>) -> bool {
  *rx.borrow_and_update()
}
```

Which makes it clear that the lock is definitely dropped after `is_aborted` returns.

psFried (approved):
LGTM % a couple of minor nit picks

crates/schema-inference/src/json_decoder.rs
```diff
@@ -61,33 +83,70 @@ where
...
+                    // This likely happened because we recently successfully parsed a document
+                    // greater than buffer_target_capacity in size, resulting in a large amount of
+                    // space freeing up in the buffer. In this case, we have no need to grow the buffer's capacity
+                    // and can safely just return Ok(None) to indicate that we need more data before we can try parsing again
+                    std::cmp::Ordering::Equal | std::cmp::Ordering::Greater => 0,
+                };
+
+                if bytes_addl > 0 {
+                    tracing::trace!(
+                        bytes_addl = format!("{:.1} MB", bytes_addl as f32 / 1_000_000f32),
```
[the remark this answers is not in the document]
jshearer: Yup, good point!

Issue #764, opened 2022-11-02
Use values from default annotations when sending documents to materializations


williamhbaker:
When the combiner drains documents to send to a materialization, if a field doesn't exist the projection of that field will currently be a `null` value. We would like to take "default" annotations from the schema into account here, and instead project to the value of the "default" annotation if present & if the field doesn't exist in the document.

This will enable materializations from source documents which do not have every possible field explicitly present to be represented more usefully in a destination store, using the specified "default" value instead of `null`. A good example of this is the new ops catalog stats introduced in https://github.com/estuary/flow/pull/755. Ideally we would not need to publish documents with explicit 0's to avoid having `null` values in the materialization source.

Also as part of this, the build process should verify that the value supplied as a default validates against the allowable type for the field.

mdibaiee:
This is interesting, why don't we apply this default value when we are capturing a document into the collection? Essentially, if we do this at materialization step, the document in the collection is going to be different from the one materialized, which is unintuitive to me 🤔

williamhbaker:
There's maybe a small storage space benefit to not storing the document with the default values. A bigger benefit IMO is so that you can change default values on the schema a (re-)materialize existing documents and have those defaults expressed in the destination. I generally agree that it makes it less clear regarding looking at documents in the collection vs. what gets materialized, but on the other hand this is already a bit of a murky area with reduction annotations.

mdibaiee:
> so that you can change default values on the schema a (re-)materialize existing documents and have those defaults expressed in the destination.

I'm not sure if this is always a benefit, specially if the details of how the default value is applied is not clear to the user (most of the time it isn't). I might expect that documents that have been consumed so far keep their existing default value while new ones use the newly specified default value.

jgraettinger:
> This is interesting, why don't we apply this default value when we are capturing a document into the collection? Essentially, if we do this at materialization step, the document in the collection is going to be different from the one materialized, which is unintuitive to me 🤔

Think of this as an operation that exists only at the boundary of where projections are applied.

The schema can define default values for locations which are _not actually present_ in the document.
If a location isn't present, it can be assumed to have the default. So long as we're working with a JSON document there's no reason for us to materialize this default value in the location. In fact, that would break if the schema default were later changed -- our materialization into a capture document would be incorrect.

However, at the moment where we're _projecting_ a document into fields, we *must* take a position of what the value is. Projection doesn't have a way to represent a location which isn't actually present in the document.

[...]

PR #825, opened 2022-12-01
combine: drainer considers default values during field extraction

williamhbaker:
**Description:**

Teaches the drainer to look for default values from `default` annotations in the schema of draining documents if an actual value is not present when extracting fields. This only takes place for a materialization, and is a means to provide an alternative to materializing `null` values.

Closes https://github.com/estuary/flow/issues/764

**Workflow steps:**

Schemas can be annotated with default values, and materializations will use those default values instead of `null`. The values provided in the schema will be validated during builds to ensure they are compatible with the fields they are specified for.

It would be good to update our schema annotations docs with this: https://docs.estuary.dev/concepts/schemas/#annotations

We should say that `default` annotations are used by materializations in place of a `null` value, and are not used for captures or derivations. They are only used for materializing into the endpoint.

**Notes for reviewers:**

N/A

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/825)

### DIFF
--- a/crates/derive/src/combine_api.rs
+++ b/crates/derive/src/combine_api.rs
@@ -118,9 +118,6 @@ impl cgo::Service for API {
                 };
 
                 let validator = new_validator(&schema_json)?;
-                if validator.schemas().len() != 1 as usize {
-                    panic!("Validator built from bundled JSON schema must have a single schema")
-                };
                 let shape =
                     doc::inference::Shape::infer(&validator.schemas()[0], validator.schema_index());
 
@@ -295,7 +292,7 @@ fn exists_or_default<T>(
             let (inner, _) = shape.locate(p);
 
             match &inner.default {
-                Some(val) => val.pack(arena, TupleDepth::new().increment()),
+                Some((val, _)) => val.pack(arena, TupleDepth::new().increment()),
                 None => {
                     doc::Node::Null::<serde_json::Value>.pack(arena, TupleDepth::new().increment())
                 }
@@ -520,7 +517,7 @@ pub mod test {
 
         let field_ptrs = vec![
             "/intProp".to_owned(),
-            "/numProp".to_owned(), // TODO: Snapshot for this doesn't work very well.
+            "/numProp".to_owned(),
             "/strProp".to_owned(),
             "/boolProp".to_owned(),
             "/objProp".to_owned(),
--- a/crates/derive/src/combine_api.rs
+++ b/crates/derive/src/combine_api.rs
@@ -64,6 +64,8 @@ struct State {
     // JSON-Pointer into which a UUID placeholder should be set,
     // or None if a placeholder shouldn't be set.
     uuid_placeholder_ptr: Option<doc::Pointer>,
+    // Shape of the schema used by the this combiner.
+    shape: doc::inference::Shape,
 }
 
 impl cgo::Service for API {
@@ -115,11 +117,18 @@ impl cgo::Service for API {
                     s => Some(doc::Pointer::from(s)),
                 };
 
+                let validator = new_validator(&schema_json)?;
+                if validator.schemas().len() != 1 as usize {
+                    panic!("Validator built from bundled JSON schema must have a single schema")
+                };
+                let shape =
+                    doc::inference::Shape::infer(&validator.schemas()[0], validator.schema_index());
+
                 let combiner = doc::Combiner::new(
                     key_ptrs.clone(),
                     None,
                     tempfile::tempfile().context("opening temporary spill file")?,
-                    new_validator(&schema_json)?,
+                    validator,
                 )?;
 
                 self.state = Some(State {
@@ -128,6 +137,7 @@ impl cgo::Service for API {
                     key_ptrs,
                     uuid_placeholder_ptr,
                     stats: CombineStats::default(),
+                    shape,
                 });
                 Ok(())
             }
@@ -177,6 +187,7 @@ impl cgo::Service for API {
                     arena,
                     out,
                     &mut state.stats.out,
+                    Some(&state.shape),
                 )?;
 
                 if !more {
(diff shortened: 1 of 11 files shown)


crates/derive/src/combine_api.rs
```diff
@@ -395,6 +415,286 @@ pub mod test {
...
+            "/arrayProp".to_owned(),
+        ];
+
+        let docs = vec![(true, json!({"aKey": "a"}))];
+
+        insta::assert_debug_snapshot!(run_simple_svc(schema_json, key_ptrs, field_ptrs, &docs));
+    }
+
+    #[test]
+    fn test_combine_with_defaults_nested() {
```
williamhbaker, line 521: Since writing these tests, I've realized that the rest of these in particular are really testing `Shape`'s ability to return annotations from located fields. I'm considering moving/merging them into test cases for `doc/inference`.

psFried (commented):
I think this is looking pretty good. The main thing that stands out to me that seems worth changing is just how the `Shape` gets passed to the combine callback. It's calling `Shape::infer` for every document that's drained and then passing the owned shape into the callback. This seems unnecessary, since every document that's drained from a given combiner will have the same schema. So I think perhaps a better approach would be call `Shape::infer` as part of configuring the combiner, and store it in the  `combine_api::State`. It could then be passed to `drain_chunk` as an `Option<&Shape>`, and the callback could use it without needing to pass it as part of the callback parameters. Would that make sense to you?

williamhbaker:
> I think this is looking pretty good. The main thing that stands out to me that seems worth changing is just how the `Shape` gets passed to the combine callback. It's calling `Shape::infer` for every document that's drained and then passing the owned shape into the callback. This seems unnecessary, since every document that's drained from a given combiner will have the same schema. So I think perhaps a better approach would be call `Shape::infer` as part of configuring the combiner, and store it in the `combine_api::State`. It could then be passed to `drain_chunk` as an `Option<&Shape>`, and the callback could use it without needing to pass it as part of the callback parameters. Would that make sense to you?

Yep, that definitely makes sense to me, seems like a much better way to do it. Thanks!

crates/derive/src/combine_api.rs
```diff
@@ -115,11 +117,18 @@ impl cgo::Service for API {
                     s => Some(doc::Pointer::from(s)),
                 };
 
+                let validator = new_validator(&schema_json)?;
+                if validator.schemas().len() != 1 as usize {
```
jgraettinger, line 121: supernit; this check seems a bit superfluous given the context (and Rust will bounds-check already).

crates/derive/src/combine_api.rs
```diff
@@ -225,6 +236,7 @@ pub fn drain_chunk(
     arena: &mut Vec<u8>,
     out: &mut Vec<cgo::Out>,
     stats: &mut DocCounter,
+    shape: Option<&doc::inference::Shape>,
```
jgraettinger, line 239: I'm thinking `shape` should be required here? 🤔 (though this can wait for a future PR). 

In practice, when `drain_chunk` is used in derivations, we use it to extract collection keys and logically partitioned values. Currently we _require_ that those locations exist, so a default value will never be used, and it doesn't _currently_ matter.

But having started down this road, I'm not sure why we _wouldn't_ allow collection key components or logically partitioned values to be omitted so long as they have valid default values.

As a practical benefit, that would allow a user to extend the key of a collection with a new component which might only be present in some documents (say, because it was recently added).

crates/derive/src/combine_api.rs
jgraettinger: 👍 the new service-level testing is very much appreciated, though. I could see shuffling some of the test coverage details around but certainly not a big deal.

jgraettinger (approved):
LGTM!

This is awesome. A couple of questions & nits, and a comment for the future, but well done.

crates/doc/src/validation.rs
```diff
@@ -171,7 +171,7 @@ impl<'schema, 'doc, 'tmp, N: AsNode> Validation<'schema, 'doc, 'tmp, N> {
     }
 }
 
-#[derive(Debug, serde::Serialize)]
+#[derive(Clone, Debug, Eq, PartialEq, serde::Serialize)]
```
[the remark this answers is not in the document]
williamhbaker: It's now a part of `Shape`, which derives `Eq` and `PartialEq`. From the looks of it [schema inference](https://github.com/estuary/flow/blob/80310c90e7144f354421c67d2e1a8b331365c08b/crates/doc/src/inference.rs#L887) does some equality comparisons at a minimum.

crates/derive/src/combine_api.rs
williamhbaker: Good point, will take this out.

crates/schemalate/src/markdown.rs
```diff
@@ -47,14 +47,15 @@ pub fn run(args: Args) -> anyhow::Result<()> {
         let title = shape.title.as_deref().unwrap_or("");
         let desc = shape.description.as_deref().unwrap_or("");
         let type_ = shape.type_.to_vec().join(", ");
+        let def = shape.default.clone().map(|(def, _)| def);
```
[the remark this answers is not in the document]
williamhbaker: I can get rid of the actual `clone()` call here like that, but I still need to derive `Clone` for `FailedValidation` since `Shape` needs it.

crates/derive/src/combine_api.rs
williamhbaker: I think I can see how we could set up the derivation pipeline to provide a `shape` pretty easily. Making use of it the way you describe would take more effort though, so I think I'll leave it as an option & the derive pipeline providing `None` for now until we want to tackle default values for derivations.

crates/derive/src/combine_api.rs
williamhbaker: Yeah I think I'll actually just leave them here. I couldn't find an obvious place to put them closer to `Shape` and they are kind of nice as documentation for how the combiner handles default values, even if they are a bit redundant.

PR #826, opened 2022-12-02
separable read vs write schemas

jgraettinger:
**Description:**

Allow collections to have a `writeSchema` and `readSchema` as an alternative to the current combined `schema`.

Remove the capability for derivation transforms to define an alternative `source` schema, as this behavior is made redundant by collection-level read schemas.

**Workflow steps:**

Optionally transition current collections to have separate read & write schemas.

Existing built specifications will continue to function as intended.

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/826)


crates/assemble/src/npm/generators.rs
```diff
@@ -16,10 +16,11 @@ pub fn module_types(
 ) -> String {
     let mut w = String::with_capacity(4096);
 
-    let doc_mapper = build_mapper(compiled, imports, &collection.schema, true);
+    let w_mapper = build_mapper(compiled, imports, &collection.write_schema, false);
+    let r_mapper = build_mapper(compiled, imports, &collection.read_schema, true);
```
willdonnelly, line 20: So the logic in this function generates types `SourceDocument` and `OutputDocument` but then only the `SourceDocuments` are re-exported as `<Transform>Source` later on, and the `OutputDocument` types are referenced directly in the actual derivation code. Do I have that right?

jgraettinger: Yep, that's right.

And in the combined-schema case we generate `Document` which aliases to `SourceDocument` & `OutputDocument`, which has the convenient property of not breaking existing derivations which import that `Document` type.

crates/schemalate/src/firebolt/firebolt_schema_builder.rs
```diff
@@ -57,7 +57,7 @@ fn build_shape_from_schema(schema_str: &str) -> Result<Shape, Error> {
 pub fn build_firebolt_schema(binding: &Binding) -> Result<TableSchema, Error> {
     let fs = binding.field_selection.as_ref().unwrap();
     let projections = &binding.collection.as_ref().unwrap().projections;
-    let schema_str = &binding.collection.as_ref().unwrap().schema_json;
+    let schema_str = &binding.collection.as_ref().unwrap().write_schema_json;
```
willdonnelly, line 60: Is the write schema the correct thing to be using here? If I understand correctly it seems like the possibly-more-restrictive read schema of the collection is what we'd want to reflect in a materialization.

willdonnelly (approved):
LGTM

crates/flowctl/src/draft/develop.rs
```diff
@@ -99,11 +99,27 @@ pub async fn do_develop(
...
-                        &catalog_url.join(&format!("{base}.schema.{ext}")).unwrap(),
-                        &mut spec.schema,
-                    )?;
+                    // Potentially write out separate schema files for the collection
+                    // schema / write_schema / read_schema.
+                    for (schema, filename) in [
+                        (spec.schema.as_mut(), format!("{base}.schema.{ext}")),
+                        (
+                            spec.write_schema.as_mut(),
+                            format!("{base}.write-schema.{ext}"),
```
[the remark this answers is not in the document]
jgraettinger: Agreed, will go with `{base}.write.schema.yaml`

crates/schemalate/src/firebolt/firebolt_schema_builder.rs
jgraettinger: Done

PR #828, opened 2022-12-02
Phil/flowctl login ux

psFried:
**Description:**

This rolls up a few improvements to the user experience of `flowctl auth` subcommands:

- Removes `flowctl auth develop`
- Adds a default local configuration for use with `start-flow.sh`
- Adds the `flowctl auth login` subcommand

**Workflow steps:**

`flowctl auth login`:

The current authentication workflow will continue to work and there's no reason to discourage it's use. `flowctl auth login` provides an easier entrypoint to that process, which may be easier for CLI users to discover (`flowctl auth --help`) and use. To use this, just run `flowctl auth login`, and it will open the browser to the `/admin/api` page. You copy your token and then paste it into the terminal where `flowctl` is waiting.

`start-flow.sh`:

- `start-flow.sh` will now provide a default config file for the `local`
  profile. You can run `flowctl --profile local` in order to use that
  config file.
- The `--profile` argument may also be provided by the `FLOWCTL_PROFILE`
  environment variable. This allows you to `export FLOWCTL_PROFILE=local`
  to avoid the need to pass `--profile` every time.
- `start-flow.sh` was updated to automatically export `FLOWCTL_PROFILE=local`
  in the tmux windows it opens, so any `flowctl` commands you run
  in those panes will automatically work against the local instance.

We could update the [getting started docs](https://docs.estuary.dev/getting-started/installation/#get-started-with-the-flow-cli), but I think we should wait until estuary/ui#399 is addressed before we make this workflow too public.

**Notes for reviewers:**

Commit by commit will definitely be easiest here.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/828)

### DIFF
--- a/crates/flowctl/src/config.rs
+++ b/crates/flowctl/src/config.rs
@@ -32,6 +32,15 @@ impl Config {
             }
         }
     }
+
+    pub fn set_access_token(&mut self, access_token: String) {
+        // Don't overwrite the other fields of api if they are already present.
+        if let Some(api) = self.api.as_mut() {
+            api.access_token = access_token;
+        } else {
+            self.api = Some(API::managed(access_token));
+        }
+    }
 }
 
 #[derive(Debug, Serialize, Deserialize)]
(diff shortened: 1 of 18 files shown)


crates/flowctl/src/config.rs
```diff
@@ -1,27 +1,67 @@
...
-            Some(api) => {
-                let client = postgrest::Postgrest::new(api.endpoint.as_str());
-                let client = client.insert_header("apikey", &api.public_token);
-                let client =
-                    client.insert_header("Authorization", format!("Bearer {}", &api.access_token));
-                Ok(client)
+    /// Loads the config corresponding to the given named `profile`.
+    /// This loads from:
+    /// - $HOME/.config/flowctl/${profile}.json on linux
+    /// - $HOME/Library/Application Support/flowctl/${profile}.json on macos
```
jshearer, line 23: Is there a reason to use a different path on macos? I would find it much easier if I went looking for my flowctl config and found a `~/.flowctl` folder, rather than digging around in `~/Library/Application Support`, which I tend to think of as opaque and not very user-introspectable

crates/flowctl/src/controlplane.rs
```diff
@@ -0,0 +1,95 @@
...
+        let exp = time::OffsetDateTime::from_unix_timestamp(self.exp).unwrap_or_else(|err| {
+            tracing::error!(exp = self.exp, error = %err, "invalid exp in JWT");
+            time::OffsetDateTime::UNIX_EPOCH
+        });
+        time::OffsetDateTime::now_utc() >= exp
+    }
+}
+
+fn parse_jwt(jwt: &str) -> anyhow::Result<JWT> {
+    let payload = jwt
```
jshearer, line 87: I was going to suggest using a library for decoding JWTs and also verifying the signature at the same time, but it looks like Supabase actually doesn't support public/private keys for their JWTs, so it doesn't seem worth it. https://github.com/supabase/supabase/discussions/5911#discussioncomment-2371302

local/local-flowctl-config.json
```diff
@@ -0,0 +1,8 @@
+{
```
jshearer, line 1: I assume these are specially crafted to live for a very long time, or something like that?

jshearer (approved):
Left a few comments, the only thing I think might be worth changing is the config directory on macos, but other than that LGTM! I think this will certainly improve the UX of getting started w/ flowctl

crates/flowctl/src/config.rs
psFried: FWIW, my own intuition matches yours here. I'm pretty accustomed to using something like `~/.flowctl/` or `~/.config/flowctl/` on OSX, and I'm not aware of any problems with that. But the choice of which directory to use in this case is made by [`dirs::config_dir`](https://docs.rs/dirs/latest/dirs/fn.config_dir.html), and all I wanted to do in _this_ PR is just add some documentation on where that's looking. I could see changing it to `~/.config/flowctl/` in a future PR, but I'd want to check with a couple more mac users first.

crates/flowctl/src/controlplane.rs
psFried: yea, this is a little bit unfortunate, but seemed like the best thing in the short term, especially given our future plans of having more of a legit API key.

local/local-flowctl-config.json
psFried: Yeah, the `public_token` and `endpoint` don't really change, so this should be a one-time operation to copy this file to your config directory, and from that point on we'll merge changes to it as you (re-)authenticate.

Issue #814, opened 2022-11-21
dpg: accept HTTP POST -> ingest capture


jgraettinger:
As a user, I want to be able to POST documents directly into a collection through an ingest capture.

For example, I minimally should be able to POST a single document in a way that broadly works with how web hooks are typically implemented:

```bash
$ curl --data-binary '{"hello":"world"}` https://the-data-plane-gateway-host.estuary.dev/name/of/capture/the-binding-name
```

It would be _nice_ to be able to POST vectors of documents to multiple bindings of the capture as a single transaction:

```bash

$ cat inputs.json
{
  "the-binding-name": [{"doc": 1}, {"doc": 2}],
  "other-binding-name": [{"more":"docs"}]
}

$ curl --data-binary @inputs.json  https://the-data-plane-gateway-host.estuary.dev/name/of/capture
```

In the future we'd like to support a streaming API as well, likely web sockets with full support for exactly once semantics, but this isn't a priority right now.

Breaking this down a bit more, users are able to capture into "ingest" captures, which already exist within the Flow specification. As with any other kind of capture, ingest captures bind to multiple collections and each such binding has a user-defined name (`the-binding-name` above). Only collections explicitly bound by the ingest capture can be ingested into, but the user is able to update the capture to change the bound collection without changing the resource name (this avoids breaking existing web hook URLs that may be in use).

Ingest captures support a bidirectionally streamed gRPC API with driver checkpoints and other bits needed for true exactly-once semantics, and the Flow reactor offers up this API today:

* [Push Runtime server](https://github.com/estuary/flow/blob/master/go/protocols/capture/capture.proto#L79-L130)

It's implemented in the runtime here:

* [Implementation of this interface in the runtime](https://github.com/estuary/flow/blob/master/go/runtime/push.go)

We have a few bits of work left to fully realize ingest captures from end-to-end:

 * There's a small bug of omission in the runtime where we're not currently sending the `Opened` message in response to the client's `Open`. We should fix that first.
 * Then, we need to implement HTTP handlers in the data-plane gateway which offer POST of single documents and (as a fast-follow) multiple documents.

psFried:
This issue was resolved, though in a different way than described here. We went with an approach that allows arbitrary network access to connector containers. This is described in #868. The actual implementation of HTTP POST captures was merged in estuary/connectors#565

PR #836, opened 2022-12-08
refactor: Update the Push capture logic

jshearer:
**Description:**

* Answer `Open` messages with `Opened` messages, allowing clients to get the capture spec. This is important because the client needs to provide the binding _index_ as part of its Push request, which can only be determined given the correct spec
* Pre-validate documents in order to respond with prompt, clear validation errors rather than failing the shard and timing out the request. This has a few yak-shaves, including:
  * Update the extraction API to handle empty UUIDs as an intent to skip UUID extraction, allowing it to be used for purely validation if desired
  * Move the push/pull tests into a separate package in order to fix a circular dependency issue
  * In order to access the extraction service, we have to do some indirection magic with interfaces and factory functions, which is the justification for the new `pf.Extractor` interface, and the whole `newExtractorFn` stuff

Part of #814

Note: I'm going to go through this in a bit and remove some extra-verbose debug logging

**TODO**: figure out how to authorize requests here

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/836)

### DIFF
--- a/go/protocols/capture/push_server.go
+++ b/go/protocols/capture/push_server.go
@@ -5,21 +5,24 @@ import (
 	fmt "fmt"
 
 	pf "github.com/estuary/flow/go/protocols/flow"
+	"github.com/sirupsen/logrus"
 	"go.gazette.dev/core/broker/client"
 )
 
 // PullServer is a server which aides implementations of the Runtime.Push RPC.
 type PushServer struct {
 	coordinator
-	ctx               context.Context   // Context of Serve's lifetime.
-	pushCh            chan readyPush    // Sent to from Push.
-	priorAck, nextAck []chan<- struct{} // Notifications for awaiting RPCs.
+	extractors        []pf.Extractor
+	ctx               context.Context         // Context of Serve's lifetime.
+	pushCh            chan readyPush          // Sent to from Push.
+	priorAck, nextAck []client.AsyncOperation // Notifications for awaiting RPCs.
 }
 
 // NewPushServer builds a new *PushServer using the provided CaptureSpec.
 func NewPushServer(
 	ctx context.Context,
 	newCombinerFn func(*pf.CaptureSpec_Binding) (pf.Combiner, error),
+	newExtractorFn func(*pf.CaptureSpec_Binding) (pf.Extractor, error),
 	range_ pf.RangeSpec,
 	spec *pf.CaptureSpec,
 	version string,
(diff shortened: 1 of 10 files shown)


go/protocols/capture/push_server.go
```diff
@@ -31,19 +34,19 @@ func NewPushServer(
 		return nil, err
 	}
 
-	var combiners [2][]pf.Combiner
```
jshearer, line 34: @jgraettinger As far as I could tell, these combiners were being created and then thrown away, I couldn't find anywhere they were being used. Is there some kind of side-effect of creating a combiner that I'm missing?

psFried: @jshearer I think this was vestigial. Combiner creation happens in the [capture coordinator](https://github.com/estuary/flow/blob/5da9656db62189b74e6935c975d7615ae3f6b34c/go/protocols/capture/coordinator.go#L68), so I think it was ok to remove this.

go/runtime/push.go
```diff
@@ -96,11 +107,18 @@ func (f *FlowConsumer) Push(stream capture.Runtime_PushServer) error {
...
 		for {
 			var docs, checkpoint, err = capture.ReadPushCheckpoint(stream, maxPushByteSize)
+			logrus.Info(fmt.Sprintf("Got %d docs, checkpoint: %v, error: %v", len(docs), checkpoint, err))
+			var ackCh = client.NewAsyncOperation()
 			if err != nil {
 				return fmt.Errorf("reading push checkpoint: %w", err)
 			} else if err = push.Push(docs, checkpoint, ackCh); err != nil {
 				return fmt.Errorf("staging push for capture: %w", err)
 			}
+			// ... Is this the right way to do this?
```
psFried, line 117: I spent some time trying to answer this question, and I don't think this is _wrong_, but I think it can probably be simplified. I _think_ a simpler approach would be to get rid of `ackErrs` entirely and create the `AsyncOperation` up on line 101. Then await the completion of the `AsyncOperation` down on line 137 using `<-asyncOp.Done():` instead of awaiting it in this separate goroutine. Basically, I don't see a benefit to indirecting through `ackErrs`.

psFried (commented):
I think this is pretty close. I left some comments, but this runtime code is pretty tricky, so please push back if any of those aren't making sense to you. I may well have missed some important detail.

jgraettinger (commented):
I wrote up all of the below comments on Friday early afternoon.

Since then, we've had a lot of discussion on this topic and Phil also wrote up comments, which I haven't reviewed. I'm sending them along anyway for general interest, or in case their useful.

Phil and I have been talking about network access to containers, and he has me like 70% convinced that the Push API might not be _that_ long for this world -- that we can tackle it's function through our existing container interface.

For all those reasons, I think we should limit the scope of this change to:
 * Send the Opened response (a current bug)
 * Do better at failing HTTP requests which also fail the shard.
 
I think there's a really narrow change we can make to do this: make sure that the combiner service is flushed _before_ coordinator [service loop yields to `startCommitFn`](https://github.com/estuary/flow/blob/master/go/protocols/capture/coordinator.go#L224). That will cause the coordinator loop to exit with an error, which (AFAIK) _is_ passed back via `PushServer.Push`.

FYI solving this in the 100% correct way is harder (out of scope) and actually requires new plumbing in Gazette (I realized there's no foolproof way for a consumer to subscribe to its own "hey, your shard has failed" notifications that work in all circumstances).

PR #837, opened 2022-12-09
Phil/flowctl ux

### LINKED ISSUE #832: Make it easier to work with specs locally with flowctl
There's a number of UX improvements that seem important for `flowctl`. The goal is to introduce some new subcommands that are more high level for handling common scenarios. These would augment, not replace, our existing subcommands. The names here are placeholders, so feel free to suggest alternate names.

`flowctl catalog pull-specs`:

It should be easier to pull specs from the control plane into a local directory. This subcommand does just that. It pulls directly from `live_specs`, and does not require a draft. Unlike `draft develop`, this command will merge the selected catalog specs with the ones that are in the directory. This behavior is optional, but enabled by default. The goal is to prevent overwriting any new specs that someone added to their local files. Note that the merge behavior is only to merge the set of catalog specs. Individual specs, schemas, etc. are just overwritten if they exist, though there should be flags for bailing out early in that case.

`flowctl publish|test|`:

It should also be easier to take a bunch of specs from your local directory and publish them. Currently, this requires the extra step of creating a draft. There's a number of other tweaks we can make to help with this.

First, we should relax our validations around `import`s. We'll still resolve imports, but no longer require that a given spec's resource `import`s that of another spec. This relaxation makes it easier to work with a subset of specs in your local directory. For example, if you want to publish only the specs in `acmeCo/foo/*`, then you can just do that without first needing to adjust the imports.

Second, you should be able to `publish` a set of specs by naming a directory. This takes advantage of the hierarchical layout to allow things like `flowctl publish --all acmeCo/sales/` to publish all the specs recursively found under the given directory. You'll still be able to pass `--source` to use a single file, but allowing selection of specs based on a recursive directory search is intended to save users from needing to add `import`s in order to publish specs that they've added.

**common selection criteria**

We should establish some common arguments for selecting 0 or more specs from the catalog. These arguments would provide some much needed help for the `catalog list` subcommand, which currently can only list _everything_ you have access to. But we should also use the same common arguments for any other subcommands that select 0 or more specs, such as `catalog pull-specs`. The specific arguments are:

- `--prefix <prefix>` selects all items whose name starts with the given prefix.
- `--<type>` (e.g. `--captures`, `--collections`, etc.) would filter the selection to only include items of the given type
- `--name` select an individual item by name
- `--expand-dependents` expands the selection based on the dependency graph of `reads_from` and `writes_to`. For example, `catalog list --name acmeCo/myCapture --expand-graph` would return all items that transitively depend on that capture. This makes it easy to identify the specs that will need re-created if you need to re-create the capture, and can also help make sense of complex graphs.

psFried:
**Description:**

Mostly fixes #832 (for now).

This rolls up a number of different UX improvements and internal refactors in `flowctl`.
The end result is that you can now run commands such as:

- `flowctl catalog list --prefix acmeCo/sales/ --collections` to list only the collections under the prefix. Listing now supports selectors for type, name, and prefix, and you can combine them arbitrarily.
- `flowctl catalog pull-specs --prefix acmeCo/sales/ --existing=merge-spec` to pull down a selection of live specs into your local directory and merge them with the specs that are already there. This supports all the same selection arguments as `catalog list`.
- `flowctl draft author --source-dir acmeCo/marketing/` to push all of the specs under the given directory (recursively) into your draft. You can even use multiple `--source` and `--source-dir` arguments to include specs from multiple different locations in a single operation.
- `flowctl catalog publish --source flow.yaml` to publish specs from your local directory in a single step. This command supports all the same arguments as `draft author` for selecting specs to publish.
- `flowctl catalog test --source flow.yaml` to run tests in a single step, again supporting those same arguments.

**Changes to imports**

This PR includes a change that removes the requirement that spec resources must `import` other resources when they "depend on" a spec that's defined there. For example, say you had a derivation defined in `foo.flow.yaml` and it reads from a source collection that's defined in `bar.flow.yaml`.
Previously, the `validation` crate would require that `foo.flow.yaml` contains an `import` of `bar.flow.yaml`, and it would result in an error if it did not. That requirement made it a lot more difficult to work with subsets of the specs in a repo, though. For example, say you had the following directories:

```
acmeCo/sales/flow.yaml
acmeCo/marketing/flow.yaml
```

If you wanted to test or publish _only_ `acmeCo/marketing/flow.yaml`, then you'd need to first go remove the `import` of `../../acmeCo/sales/flow.yaml`. When you want to test/publish both again, you'd need to add the `import` back in.

Another example scenario is when you have an existing file that defines `acmeCo/marketing/derivationA` that reads from `acmeCo/sales/source` that's not defined in any of the files in your repo. Then say you want to use `pull-specs` to get the spec for `acmeCo/sales/source` so you can update it and then publish both that and `derivationA` together.
X But now you get an error because `derivationA` didn't `import` the newly added file, so you again need to adjust imports.

Both of those scenarios now "just work". You can run `flowctl catalog publish --source-dir whatever/dir/` and it will never complain about missing imports. Note that `import`s _are_ still resolved. You can still use `import` to include additional catalog resources in your build. The only difference is that now you _never have to_.

**merges with `pull-specs`**

The pull-specs subcommand needs to deal with the case where your local directory already has some some specs in it. I don't think there's a single solution that works for all scenarios, so I came up with a few options that seem like they would cover most common scenarios, and defaulted it to the safest option, which simply aborts if it would try to write to a file that already exists. Other options allow you to keep existing files (discarding the new ones), overwrite existing files (discard the old ones), or merge the two (`--existing=merge-specs`). The `merge-specs` option will merge an existing spec file (`flow.yaml` or `*.flow.yaml`) with the new one that is about to be written. The merge operation is very basic. It does not try to merge properties of individual specs, but only the overall set of specs in each file. For example:

existing `flow.yaml`:
```yaml
collections:
  acmeCo/sales/hammers: { ... }
  acmeCo/sales/anvils: { ... }
```

new `flow.yaml` (from live_specs):
```yaml
collections:
  acmeCo/sales/anvils: { ... }
  acmeCo/sales/tnt: { ... }
```

Then the merged `flow.yaml` would include all three collections, with the `acmeCo/sales/anvils` collection completely overwriting all properties of the spec in the local copy with the ones from live_specs. When we `unbundle` the catalog, we take the inlined schemas, typescript modules, and endpoint configs and write them out as separate files. If `merge-spec` is used, then these schema and config files will always be completely overwritten by the ones from the new catalog. In other words, schemas, typescript modules, and endpoint configs are never merged, as they are considered properties of the specs.

**refactoring**

I extracted functions for bundling and unbundling of catalogs, as these are common operations for working with catalog specs. A "bundled" catalog is one that has had all relative URLs resolved and converted to inlined resources. It is effectively just a giant JSON object containing Flow specs that each fully self-contained. Conversely, unbundling a catalog means replacing the inlined resources with references to separate files.

I'm thinking that a bundled catalog ought to be the primary representation of sets of specs, and that most functions that operate on 0 or more specs ought to accept/return bundled `models::Catalog`s. That's what I've tried to do here. I think it may be beneficial to at some point define separate rust types for bundled and unbundled catalogs, but I'm leaving that for a future exercise.

I added the `flowctl raw unbundle` command, to partner with the existing `flowctl raw bundle`. I've manually tested some examples and verified that you can round trip a catalog through unbundling and re-bundling and get the same result. This would probably make for a good automated end-to-end test case, and I think it'll probably be worth pushing another commit for that.

**Workflow steps:**

The subcommands _are_ the steps, so I'm just going to paste the output of `flowctl catalog --help`:

```
flowctl-catalog 
Work with the current Flow catalog

USAGE:
    flowctl catalog [OPTIONS] <SUBCOMMAND>

OPTIONS:
    -o, --output <OUTPUT>    How to format CLI output [possible values: json, yaml, table]
    -h, --help               Print help information

SUBCOMMANDS:
    list          List catalog specifications
    pull-specs    Pull down catalog specifications into a local directory
    publish       Publish catalog specifications
    test          Test catalog specifications
    history       History of a catalog specification
    draft         Add a catalog specification to your current draft
    help          Print this message or the help of the given subcommand(s)
```

- [The imports docs](https://docs.estuary.dev/concepts/import/) will need updated to reflect the removal of import path requirements
- And of course the [flowctl docs](https://docs.estuary.dev/concepts/flowctl/) :grinning: 

**Notes for reviewers:**

I'm pretty meh on the name `pull-specs`, honestly. Suggestions welcome. And I'm a little worried that `pull-specs`, `test`, and `publish` are also a little out of place being under the `catalog` grouping. Would it perhaps be better to move them under a `flowctl specs` (or `sources`) grouping that's for commands that deal with catalog specs as files/URLs? Suggestions welcome there, too. 

**Things I did _not_ do**

There's a bunch of good ideas for UX improvements that I just couldn't make time for. I'll list the ones here that I'm at least ~75% convinced that we should implement, and please chime in if you have any thoughts on these, as this was roughly what I had in mind with the refactoring I did.

- A `diff` subcommand that diffs a listing of catalog specs (can be a draft or live_specs) with local file specs. I'm also interested in exploring the idea of a semantic diff function/representation that can be compiled to WASM and used by the UI.
- Capability to list catalog entities by traversing the data flow graph (live_spec_flows), so you could `flowctl catalog list --name acmeCo/my-capture --include-dependents` to get your capture and everything that transitively depends on it.
- High-level `disable` and `delete` subcommands that work similarly to `test` and `publish`. This just seems to be something we do a lot, and it would be nice to be able to do it by `--prefix` (or by `--include-dependents`)
- An [interactive fuzzy finder](https://github.com/estuary/flow/issues/832#issuecomment-1340267131) for selecting catalog entities would be amazing. It would give an interactive multi-selection window with a preview pane, which could be used as selection for other commands like `pull-specs` (or `disable` and `delete`).
- Refactor `catalog draft` to accept the same list selection args as the others. It should, but it's slightly more complicated because it deals with publication specs as well as live_specs.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/837)


psFried:
> Refactor catalog draft to accept the same list selection args as the others. It should, but it's slightly more complicated because it deals with publication specs as well as live_specs.

I wanted to expand on this a bit. I don't love the inconsistency between `catalog draft` and the rest of our subcommands. `draft` is able to select specs from a specific `publication_id`, whereas other commands are not. We could certainly add a `--publication-id` argument to `catalog list` (and `pull-specs`, etc.) to allow selecting from `publication_specs` in the same way that `catalog draft` does. But another thing we could do would be to _remove_ `--publication-id` from `catalog draft` and introduce a separate `publications` subcommand group that deals with the `publication_specs` table. So you'd have `flowctl publications draft|list|pull-specs`, and the `catalog` subcommand would be more just for operations on the `live_specs*` table. If we did that, then I'd think it would also make sense for the `history` subcommand to also move under `publications`.

Anyway, I'd been sitting on the fence on that issue, which is why I didn't take action on it in this PR, but I'm starting to lean more toward introducing the separate `publications` grouping in a future PR, so would be curious if others have thoughts on that.

crates/flowctl/src/source/unbundle.rs
```diff
@@ -0,0 +1,338 @@
...
+    spec_filename: &str,
+    json: bool,
+    existing: Existing,
+) -> anyhow::Result<()> {
+    fs::create_dir_all(dir_path).await?;
+    let flow_yaml_path = dir_path.join(spec_filename);
+
+    let maybe_catalog = resolve_catalog_to_write(&flow_yaml_path, new_catalog, existing).await?;
+    let Some(mut new_catalog) = maybe_catalog else {
+        return Ok(());
```
jshearer, line 30: Silently do nothing if we don't pass a valid flow catalog to unbundle? Should this error instead?

crates/flowctl/src/source/unbundle.rs
```diff
@@ -0,0 +1,338 @@
...
+            }
+        }
+    }
+
+    for (name, capture) in new_catalog.captures.iter_mut() {
+        let base = base_name(name);
+
+        if let models::CaptureEndpoint::Connector(connector) = &mut capture.endpoint {
+            // Indirect the connector config to a file, and track its reference.
+            let filename = format!("{base}.config.{ext}");
```
jshearer, line 83: Worth calling these `{base}.capture.{ext}` and `{base}.materialization.{ext}`?

crates/flowctl/src/source/unbundle.rs
```diff
@@ -0,0 +1,338 @@
...
+            Ok(())
+        }
+        Err(err)
+            if err.kind() == std::io::ErrorKind::AlreadyExists && existing == Existing::Keep =>
+        {
+            tracing::info!(path = %path.display(), "skipping writing file because it already exists and --existing=skip");
+            Ok(())
+        }
+        // This branch also handles the case of `Existing::Error` if the file already exists.
+        Err(err) => Err(anyhow::anyhow!("unable to open file: {}", err)),
```
jshearer, line 208: Probably should include the file path here like you did above, just to avoid annoyingly hard to debug error messages. Or is the idea that `err` will contain the path?

crates/flowctl/src/source/unbundle.rs
```diff
@@ -0,0 +1,338 @@
...
+    let cat = sources::parse_catalog_spec(&contents)?;
+    Ok(cat)
+}
+
+/// Returns the last path component for a given catalog name, which must be a valid catalog name.
+fn base_name(catalog_name: &impl AsRef<str>) -> &str {
+    catalog_name
+        .as_ref()
+        .rsplit_once("/")
+        .expect("catalog names have at least one '/'")
```
jshearer, line 238: Is it dangerous to panic here, where everywhere else you're using `anyhow::Result`? i.e will panicking in `base_name` cause some resources to not get cleaned up properly?

crates/flowctl/src/source/unbundle.rs
```diff
@@ -0,0 +1,338 @@
...
+) -> anyhow::Result<()> {
+    let file_path = dir.join(&filename);
+    let contents = serializer.to_vec(&config.config)?;
+
+    write_file(&file_path, &contents, existing).await?;
+    config.config = RawValue::from_string(Value::String(filename).to_string()).unwrap();
+    Ok(())
+}
+
+async fn maybe_indirect_schema(
```
jshearer, line 257: I don't understand what this is about (i.e I can see what's happening, just not why), so maybe a comment would be useful here to explain why this is needed? Or maybe just an answer to this GH comment so I can understand :P

crates/flowctl/src/source/unbundle.rs
```diff
@@ -0,0 +1,338 @@
...
+    }
+
+    let path = dir.join(filename);
+    write_file(&path, &contents, existing).await?;
+    *schema = models::Schema::Url(models::RelativeUrl::new(filename));
+
+    Ok(())
+}
+
+fn merge_catalog(
```
jshearer, line 295: You say
> The merge operation is very basic. It does not try to merge properties of individual specs, but only the overall set of specs in each file.

in the PR description, which matches what's going on here, but is this really sufficient? I'm certainly not thinking of all the ways this could be used, but what about the following scenario:

* You have a spec locally, and you have defined the capture `acmeCo/pizza/my_capture`. This capture exists already, but you're making a change to its config, maybe endpoint config or schema or something
* You realize that you actually need a bunch of other `acmeCo/pizza` specs, so you pull them down with `flowctl catalog pull-specs --prefix acmeCo/pizza/ --existing=merge-spec`.

Does this logic mean that after running `pull-specs`, your local changes to `acmeCo/pizza/my_capture` will be blown away and replaced with the specs as they exist in production?

crates/sources/src/loader.rs
```diff
@@ -68,6 +68,48 @@ pub struct Loader<F: Fetcher> {
...
+}
+
+/// Parses `content` into a JSON DOM that can be further parsed into a Rust type.
+fn load_resource_content_dom(
+    content: &[u8],
+    content_type: flow::ContentType,
+) -> Result<Box<RawValue>, LoadError> {
+    use flow::ContentType as CT;
+
+    // These types are not documents and have a placeholder DOM.
```
jshearer, line 87: This is very likely just some nuance I don't understand, but it looks like you're replacing npm/typescript config with null, and the comment doesn't help me understand why (isn't e.g the npm config actually a document containing packages and versions to install?)

jshearer: Well maybe nevermind, this isn't new code, it was just moved. /shrug I assume it's there for a reason

jshearer (approved):
I think this looks good! I left a few comments, mainly nit-picks or questions for my own edification. 

So, the upshot of this PR is not that we're _removing_ `flowctl draft`, just that we're adding higher-level `flowctl catalog` commands, which will deal with creating/cleaning up the underlying drafts as needed. Right?

One concern I have about flowctl in general is that it doesn't really feel like its arranged very coherently, and maybe this PR will fix that at least for the garden path? Theoretically we could (should?) get rid of `flowctl draft` commands once we're sure that the logic here works properly, right?

crates/flowctl/src/catalog/mod.rs
```diff
@@ -81,13 +252,130 @@ impl Catalog {
...
+            out.push(self.reads_from.iter().flatten().join("\n"));
+            out.push(self.writes_to.iter().flatten().join("\n"));
+        }
+        out
+    }
+}
+
+/// Collects an iterator of `LiveSpecRow`s into a `models::Catalog`. The rows must
+/// all have a `spec` unless they are deleted (`spec_type = null`).
+pub fn collect_specs(
```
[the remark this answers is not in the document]
psFried: The `Result` is what prevents that, since you can't impl `FromIterator` for `Result`.

crates/flowctl/src/catalog/publish.rs
```diff
@@ -0,0 +1,73 @@
...
+    // since that can fail due to missing/expired credentials.
+    let client = ctx.controlplane_client()?;
+
+    anyhow::ensure!(args.auto_approve || std::io::stdin().is_tty(), "The publish command must be run interactively unless the `--auto-approve` flag is provided");
+
+    let sources = args.source_args.resolve_sources().await?;
+    let catalog = crate::source::bundle(sources).await?;
+
+    let draft = draft::create_draft(client.clone()).await?;
+    println!("Created draft: {}", &draft.id);
```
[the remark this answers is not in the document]
psFried: `println!` goes to stdout and is intended to be messages that we always want to show to the user. `tracing::info!` is just application logs, which go to stderr, and info level isn't even enabled by default. Here the `println`s are meant to communicate progress, since otherwise the user may just be waiting for a while with no output.

crates/flowctl/src/source/unbundle.rs
psFried: The error is being propagated here (`?` at the end of line 28). But the successful result of `resolve_catalog_to_write` might still be `None` if the user has opted not to overwrite existing catalog specs with the new one, and there's an existing file already there.

crates/flowctl/src/source/unbundle.rs
psFried: I opted to keep these exactly the same as they are in `draft develop` so that existing repos don't have to deal with files being re-named. On a related note, there's no deletion of existing "unbundled" files like schemas and endpoint configs, so changing the names here would potentially result in both `{base}.capture.{ext}` and `{base}.config.{ext}` existing side by side.

crates/flowctl/src/source/unbundle.rs
psFried: The path is added to the context in `write_file`. `try_write_file` was just created so that `write_file` only has to add that context in a single place instead of attaching it to both fallible calls.

crates/flowctl/src/source/unbundle.rs
psFried: Good question. I think it's probably safe enough, but it also can't hurt to change it to return a Result. I'll update it.

crates/flowctl/src/source/unbundle.rs
psFried: A bundled catalog is one that has all schema resources inlined. The act of unbundling is taking those inlined schemas and writing them out to separate files, replacing the inlined schema with a relative URL pointing to the file. The `maybe_` is because, if the schema is very small, then we just keep it inlined. The overall goal of unbundling is to leave the user with a directory of Flow specs that's easy to navigate and read, so they can edit them and potentially re-publish them.

crates/flowctl/src/source/unbundle.rs
psFried: > Does this logic mean that after running pull-specs, your local changes to acmeCo/pizza/my_capture will be blown away and replaced with the specs as they exist in production?

Yes it does. I'm not really sure whether it's sufficient, but I also don't really see a straight forward solution that's better. We can't merge endpoint configs, and need to just pick either the local copy or the remote one. We could do something like `--existing=merge-specs-but-keep-endpoint-configs`, but it feels to me like that introduces more confusion than it's worth. Now that I'm thinking about it, probably the best default (for interactive shells, at least) would be `--existing=prompt`, which would prompt the user for each existing file. I'll propose that we implement the prompt in the next pass of UX improvements, if that makes sense to you.

crates/sources/src/loader.rs
psFried: Yeah, it's existing, but it's because there's no way to directly represent a typescript module or npm package as JSON. `load_resource_content` could definitely be refactored to remove this weirdness, but I was trying to keep my changes to this crate as minimal as possible in this PR, since such a refactor is pretty unrelated

psFried:
> One concern I have about flowctl in general is that it doesn't really feel like its various commands are arranged very coherently, and maybe this PR will fix that at least for the garden path? Theoretically we could (should?) get rid of flowctl draft commands once we're sure that the logic here works properly, right?

@jshearer I'm also feeling kinda meh on the current arrangement. `flowctl draft` is part of the puzzle, though I hadn't been thinking we'd remove it. At the moment, `draft` still has important capabilities, which are not covered by other subcommands, though. And a "draft" is also an important concept in the control plane, that isn't going anywhere. The `draft` subcommand lets you, for example, work with drafts locally that were created via the UI. _If we did_ want to rip out the `draft` subcommand, I think we'd need to do something like add a `--draft-id <id>` argument to the other subcommands (`catalog list|pull-specs|publish|test`). TBH I think that `draft` should probably stick around, but I'm definitely willing to discuss alternative views on that.

But getting back to the broader issue of the arrangement/grouping of subcommands, I'd definitely like to hear any ideas you have there. One particular issue that I'm still waffling on is: 

> And I'm a little worried that pull-specs, test, and publish are also a little out of place being under the catalog grouping. Would it perhaps be better to move them under a flowctl specs (or sources) grouping that's for commands that deal with catalog specs as files/URLs?

If we went with the `sources` subcommand, then I'd see the overall organization as being:

- `catalog`: subcommands for working with `live_specs`
- `draft`: subcommands for working with `draft_specs`
- `sources`: subcommands for working with specs on the local filesystem

It seems clean, except that some subcommands actually work with various combinations of `live_specs`, `draft_specs`, _and_ local sources. For example, should `pull-specs` live under `sources` or `catalog`, given that it pulls from `live_specs`? I see the _current_ organization as being perhaps a little easier because it's implied that both `catalog` and `draft` have the capability of working with local sources. And that seems reasonable to me, because why else would you install `flowctl` if not to work with local sources? So, while I have a sense that there's room for improvement in the current arrangement of subcommands, I'm not yet seeing an alternative that feels significantly better. But I would like to get your take on things, since I'm not especially attached to looking at things that way.

crates/flowctl/src/source/unbundle.rs
jshearer: Gotcha!

crates/flowctl/src/source/unbundle.rs
jshearer: Ahh ok, missed that!

crates/flowctl/src/source/unbundle.rs
jshearer: Ah sorry, I meant the stuff about `$id` and `$defs`. The part about keeping very small schemas inlined is great though, that should make things much less fragmented

crates/flowctl/src/source/unbundle.rs
jshearer: Okay cool, this makes sense. Yeah, prompting sounds like it'd be better UX, though even a simpler `git`-style failure of "You have pending changes that would be overwritten by this action. Use --force to acknowledge" would probably be good enough too

crates/flowctl/src/source/unbundle.rs
jshearer: Mhmmmmm, I didn't consider the backwards compatibility aspect. Makes sense!

jshearer:
I'm trying to approach this problem from the perspective of a user who's probably not very familiar with Flow, and imagine how I would expect the CLI to function, and what I would want to do with it. Here's an example of a flow I've gone through before:

**I want to work on a _thing_ that one of my colleagues was working on.** 
I know that it was capturing from a database, and I might know the capture name. I need to pull down all of the relevant specs, play with them, test them, publish them, check their status, and then look at their logs when/if something doesn't work As of this PR, this flow would look something like this:

  * **Find the name of the thing I'm looking for**: `flowctl catalog list --prefix acmeCo/ --captures`     
     I'm not sure what catalog means here, that word feels too overloaded. My intent is to get a list of captures that are in production, so something like `flowctl entities list acmeCo/* --captures`, or even `flowctl get acmeCo/* --captures` (or `flowctl list`, not sure if `get` is better or worse) feels more ergonomic. One reason this feels better is it uses the same terminology as other CLI tools, e.g `kubectl`. 
  * **Pull down all of the relevant specs so I can look at/work on them**: `flowctl catalog pull-specs --prefix acmeCo/taskName`
    Again, it feels like I have to think really hard in order to figure out what catalog actually means here. Conceptually, I want to **pull** specs. What about `flowctl entities pull acmeCo/*`, or even `flowctl checkout acmeCo/*` (`flowctl pull` also feels good here, not sure if `checkout` is too much)? I'm very used to `git pull`/`git checkout`, so this feels very easy to understand.
  
    **Note:** I would love to be able to run e.g `flowctl checkout acmeCo/project/captureName --recursive` (this actually feels like it should be the default, with something like `--non-recursive` to indicate only pulling a specific spec or prefix), which would walk all of the dependencies and checkout everything needed to fully specify this entity, and nothing more. But I'm fairly sure you called this out as something you explicitly did _not_ work on here, which does make sense wrt limiting scope of changes.
  * **Publish my changes after working on them**: `flowctl draft author --source-dir acmeCo/marketing/`, `flowctl catalog publish --source flow.yaml`
    I have two suggestions that make these feel more ergonomic:
    * Default `--source-dir` to the current directory and `--source` to the `flow.yaml` in the current directory, much the same as `docker build` defaults to `./Dockerfile`, etc.
    * What about `flowctl entities publish`, or even `flowctl publish`, which by default will run `flowctl draft author`? We should keep `draft author` around in order to facilitate UI interop, but I imagine 95% of the time you'll just want to publish your local changes and not worry about the draft.

crates/flowctl/src/source/unbundle.rs
jgraettinger: Every JSON-schema that we bundle includes a top-level `$id` keyword (part of the JSON-schema spec) which identifies the canonical URI (CURI) from which we got the schema.

Every schema must have a CURI, so in cases where the schema was originally inlined we'll invent a CURI for it using a scheme I'll skip over in this comment.

That CURI often isn't meaningful to the user can can be removed, as it looks weird to leave it. BUT, if $defs are included in the JSON-schema, it may be a bundled representation of _multiple_ source files, and `$id` may play an important role as the base URL against which references are resolved, and it would be incorrect to remove it.

jshearer:
> For example, `flowctl specs list` doesn't really distinguish whether an operation is being undertaken against live specs, or draft specs, which seems like a problem (is it? 🤔).

Why would you want to do anything other than list the live specs?

psFried:
> What happens if flowctl specs is the top-level action?

I think I really like this. Here's what I'm thinking it would look like:

`flowctl specs <SUBCOMMAND>`

- `list`: prints a listing
    - `--draft-id`: selects the specs from the given draft
    - `--prefix`, `--captures`, etc: filter on name or type
- `draft`: adds specs to a draft
    - `--draft-id`: adds specs to the given draft id. If not provided, a new one will be created
    - `--source`, `--source-dir`: add specs from the local directory
    - `--prefix`, `--captures`, etc: add specs from `live_specs` (like the current `flowctl catalog draft`)
    - Thinking here that the name/type arguments would be mutually exclusive with `--source` and `--source-dir`.
- `publish` and `test`: publishes or tests specs, either from a draft or from the local filesystem
    - `--draft-id`: publishes the given draft
    - `--source`, `--source-dir`: publishes specs from the local filesystem, using these inputs
    - I like the idea of having the publish default be either `--source flow.yaml` or `--source-dir .`
- `pull`: writes specs to the local filesystem
    - `--draft-id`: selects the specs from the given draft. If not provided, selects from `live_specs`
    - `--prefix`, `--captures`, etc: filter on name or type
- `history`: print history of the selected specs
    - `--prefix`, `--captures`, etc: filter on name or type

> it "solves" the problem of conceptual distinctions by throwing it all in a single bucket

I think the pros outweigh the cons here. The reality is that our existing subcommands already blur these distinctions. `history` uses `publications`, not `live_specs`. `catalog draft` operates on `draft_specs` and `live_specs` (and `publication_specs`). So it actually seems pretty reasonable to me to just throw them all under a single grouping.

If we did that, then the `catalog` grouping would go away. However, I'm now looking at the `drafts` subcommands, and it's really not looking like we could move them all. We'd still need `draft list` and `draft delete`, at a minimum (aside: I think `select` can probably be removed, along with removing the concept of the current draft id from the config). So I think the question is whether a single `flowctl specs` still sounds good, given that `drafts` would need to stick around :thinking: WDYT?

psFried:
@jshearer and I discussed this today, and he had an idea that I like: just remove the `catalog` subcommand grouping and let `list`, `pull`, `publish`, `test`, and `history` all be top-level subcommands. The `draft` subcommand group would continue to exist, since working directly with drafts is expected to be a somewhat rare and advanced usage of the CLI. Now that we have high level `pull`, `publish`, and `test` subcommands, there's basically just not as much cause for working with drafts directly. The `catalog draft` subcommand would be moved to `draft add`, and the `catalog` grouping would be removed.

Here's the complete breakdown of subcommands we'd be left with:

`flowctl`:

- `list`
- `pull`: instead of `catalog pull-specs`
- `publish`
- `test`
- `history`
- `collections`:
    - `read`
    - `list-journals`
    - `list-fragments`
    - `infer-schema`
- `draft`:
    - `list`
    - `describe`
    - `create`
    - `delete`
    - `add`: moved here from `catalog draft`
- `logs`
- `stats`
- `typescript`
- `raw`
- `auth`:
    - `login`
    - `token`
    - `roles`

I like that it puts the most common commands at the top level and makes them easily discoverable in the output of `flowctl --help`. @jgraettinger I'm curious to get your thoughts on this. Personally, I'm inclined to push another commit on this PR to move them around.

psFried:
I just spoke with @jgraettinger and we decided to just merge this PR as it is right now. We think that there's definitely still room for improvement in the organization of the subcommands, and the above suggestion of flattening them _might_ be the best way forward. But it's also not seeming urgent, and there's ongoing UX discussions that would be nice to resolve before we commit to any particular direction here.

Issue #831, opened 2022-12-06
update ops catalog stats derivations to not include explicit zero values


williamhbaker:
https://github.com/estuary/flow/issues/764 will need to be addressed first, but once that capability is in, we can stop writing explicit 0's for various values that will be included in the field selection of the materialization. The schema for these should be updated to have default values of 0's, so that the materialized columns contain a `0` instead of `null` for easier consumption of the data in the tables.

Explicit zero values as they currently exist in the stats derivation: https://github.com/estuary/flow/blob/4a76a833ecd98320eec967e421ad3ed294de84cd/ops-catalog/catalog-stats.ts#L65-L82

There will need to be some non-trivial updates to the derivation to get it to work in this way. The stats schema will need to have these fields which are currently required be set to optional. The spec stored by the materialization in the destination table will also need to be re-created, since it will have records of the projections as not being nullable.

PR #838, opened 2022-12-09
ops-catalog: stats derivation does not output explicit zero values

williamhbaker:
**Description:**

Following up on https://github.com/estuary/flow/issues/764, this updates the catalog stats derivation to not output explicit zeros. This will save space, since many of these fields are 0 most of the time. Using default values of `0` on `docsAndBytes` we can now materialize into the stats table without having `null` values.

Also in this PR I am updating the stats materialization to use the "normal" postgres connector, since it is now compatible with pgBouncer, see https://github.com/estuary/connectors/pull/439.

Some notes on how this can be rolled out:
- It needs https://github.com/estuary/flow/issues/764 to actually be deployed and running live, otherwise materializations from this pipeline will fail on attempting to insert `null` values into the stats table.
- Existing tenants will need their specs updated _somehow_. We've got it on the list to develop a process for doing this. Since there's currently only two tenants with stats materializations manual updating might be practical as well.
- I believe we will need to finangle the spec that is currently stored in the stats materialization table, as the new projections of the columns will be nullable whereas the current ones are not.

Closes https://github.com/estuary/flow/issues/831

**Workflow steps:**

Running a local stack as a test, the stats materialization will now happily populate the stats table with `0`'s, while the derived collection omits fields with only `0`'s.

N/A

**Notes for reviewers:**

N/A

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/838)

### DIFF
--- a/ops-catalog/catalog-stats.ts
+++ b/ops-catalog/catalog-stats.ts
@@ -62,25 +62,6 @@ type StatsData = {
     [k: string]: Document['statsSummary'];
 };
 
-const newDocumentStats = (): Document['statsSummary'] => ({
-    writtenByMe: {
-        bytesTotal: 0,
-        docsTotal: 0,
-    },
-    readByMe: {
-        bytesTotal: 0,
-        docsTotal: 0,
-    },
-    writtenToMe: {
-        bytesTotal: 0,
-        docsTotal: 0,
-    },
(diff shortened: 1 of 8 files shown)


jgraettinger (approved):
LGTM!

PR #839, opened 2022-12-09
Docs: UI basics and editing

oliviamiannone:
**Description:**

Adds:
* Basic intro to the UI and its functionality to the Concepts docs
* Reference docs on editing considerations

Updates:
* Verbiage throughout docs wrt Discover> Next button

No shortlinks currently

**Notes for reviewers:**

Please fact-check for me and suggest any important bits I may have missed!

I need to review this one more time myself, and I'm sure it'll also need numerous updates in the coming weeks, but I wanted to get this PR up before the weekend so we can publish (hopefully) early next week.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/839)


site/docs/guides/create-dataflow.md
```diff
@@ -94,19 +100,28 @@ Now that you've captured data into one or more collections, you can materialize
...
+2. Choose a unique name for your materialization like you did when naming your capture; for example, `acmeCo/myFirstMaterialization`.
+
+3. Fill out the required properties in the **Endpoint Configuration**.
+
+4. Details of the collections you just created are already filled out in the **Collection Selector**.
+If you'd like, you can remove some of these or add additional collections.
 
-2.  Choose a unique name for your materialization like you did when naming your capture; for example, `acmeCo/myFirstMaterialization`.
+   * To easily find collections, you can use a filter.
+   Hover your cursor within to the Collection Selector table header, next to the **Remove All** button, to reveal an expandable menu icon (three dots).
```
travjenkins, line 111: The menu is still shown on hover of the table header... but that is no longer next to Remove All as Remove All is in a "button row" kind of thing now.

So the three dot menu is now below the Remove All button.

image

site/docs/reference/editing.md
```diff
@@ -0,0 +1,68 @@
...
+For example, say you have capture that writes to a collection called `post/fruity_pebbles/nutritionFacts`.
+You begin to edit the capture using the latest version of the connector.
+The connector detects the same set of nutrition facts data,
+but maps it to a collection called `post/fruity_pebbles/nutrition-facts`.
+If you continue to publish the edited capture, both collections will persist,
+but new data will be written to the new collection.
+
+Before editing, check if a connector has been updated:
+
+* Go to the **Admin** tab and view the list of connectors. Each tile shows the date it was last updated.
```
travjenkins, line 54: This is a good recommendation... we should probably add a sort by last updated in the UI to better support users doing this.

site/docs/concepts/web-app.md
```diff
@@ -0,0 +1,272 @@
...
+![](<./webapp-images/capture-page.png>)
+
+**1:** Select all or deselect all.
+
+**2:** Enable, Disable, and Delete buttons. These actions will be applied to the selected table rows. Choose **Disable** to temporarily pause the flow of data, **Enable** to resume, and **Delete** to permanently remove the capture(s).
+
+**3:** Materialize button. When you click this button, you're directed to the **Create materializations** page.
+All the collections of the selected capture(s) will be added to the materialization.
+
+**4:** Filter captures by name.
```
kiahna-tucker, line 72: I will need to pull this branch down to view the reference images, but I am guessing the reference image is of the full _Captures_ page with numeric labels for its various _segments_. If that assumption is accurate, then the search bar is what is being referenced by the _4_ label.

Granted all the above is true, it is worth highlighting that this search bar allows you to filter captures by catalog name and connector name. Since the catalog names of captures are appended by a connector identifying string (e.g., `/source-firestore`), a connector name query will often match the catalog name in some fashion. However, that is not always the case. For example, the catalog name for a _Google Cloud Storage_ capture is appended by the string `/from-gcs`. While _GCS_ is a commonplace acronym for _Google Cloud Storage_, the explicit connector name (or some variant of it) may be a primary point of reference for a user. A user can easily find all their _Google Cloud Storage_ captures by entering the italicized string (or a subset of it) into the search bar. In the event a user does happen to type a subset of the connector name in the search bar which matches other captures (e.g., typing _Google_ into the search bar can pull captures with an assortment of sources), the table column filters can further facilitate their search.

The same is true for the search bar on the _Materializations_ page.

site/docs/concepts/web-app.md
```diff
@@ -0,0 +1,272 @@
...
+
+**5:** Shard status indicator. Shows the status of the task [shard](./advanced/shards.md) that backs this capture.
+
+* **Primary (Green)**: Data is actively flowing through the capture.
+* **Pending (Yellow)**: The capture is attempting to re-connect. Often, you'll see this after you re-enable the capture as Flow backfills historical data.
+* **Failed (Red)**: The capture has failed with an unrecoverable error.
+* **Disabled (White)**: The capture is disabled.
```
kiahna-tucker, line 79: When identifying disabled and _unknown_ shard status indicators, I would reference their color and fill for both dark and light mode. For example, the shard status indicator for a disabled task is always a circle outline. It is light grey (or white, if that is how it registers to most people) in dark mode and dark blue (or black, if that is how it registers to most people) in light mode. The shard status indicator for an _unknown_ task is always a filled in circle that is light grey (or white) in dark mode and dark blue (or black) in light mode.

site/docs/concepts/web-app.md
```diff
@@ -0,0 +1,272 @@
...
+
+**9:** Publish time. Hover over this value to see the exact time the capture was first published.
+
+**10:** Options. Choose to **View Details** or **Edit Specification**.
+
+### Detail view
+
+When you click **View Details** for a capture, the **Status** and **Specification** viewers are revealed.
+
+The **Status** window shows the full identifier of the shard that backs your capture. If there's an error, you can view its logs.
```
kiahna-tucker, line 96: Since a task can have multiple shards would it make sense to indicate this by saying `shard(s)`? This may be a detail that is generally understood or implied, but asking just in case we would like the language use here to have that _degree of flexibility_, for lack of a better term.

Regarding accessing the logs of erroring shards, do we need to be more explicit and note that an error alert will appear in the details pane that provides access to the logs?

site/docs/concepts/web-app.md
```diff
@@ -0,0 +1,272 @@
...
+
+In the **Specification** window, you can view the specification of the capture itself, as well as each collection to which it writes.
```
kiahna-tucker, line 98: Should we highlight the fact that the left-hand pane of the _Specification window_ can be resized in the event a long catalog name is truncated? This was a point of tension in the UI in the past; people were not aware that you can horizontally resized each side of the _Specification window_. Some tweaks were made to the divider to make this more obvious, but it may be worth noting.

site/docs/concepts/web-app.md
```diff
@@ -0,0 +1,272 @@
...
+
+1. Edit the connection to the source system, if desired. You can either update fields in the **Endpoint Configuration** section or manually update the JSON in the **Specification Editor**.
+
+:::caution
+You may have to re-authenticate with the source system. Be sure to have current credentials on hand before editing the endpoint configuration.
+:::
+
+2. Use the **Collection Selector** to add or remove collections from the capture, if desired.
+
+   To refresh your connection with the source and see an updated list of possible collections, click the **Rediscover All** button,
```
kiahna-tucker, line 117: I believe Dave would prefer for the _Refresh_ CTA (formerly known as _Rediscover All_) to be renamed as _Find New_. I am not sure where he and Johnny stand on that revised name. Currently, _Refresh_ is the label associated with this CTA.

site/docs/concepts/web-app.md
```diff
@@ -0,0 +1,272 @@
...
+
+## Collections page
+
+The **Collections** page shows a read-only table of [collections](./collections.md) to which you have access.
+You can view each collection's specification and see a sample of its data.
+This can help you verify that collection data was captured as expected and that you'll be able to materialize it how you want, and troubleshoot it necessary.
+
+To reveal the **Specification** and **Data Preview** sections, expand **Details** next to a collection name.
+
+The **Specification** section shows the collection specification as JSON in a read-only editor.
```
kiahna-tucker, line 154: In the _Captures page_ section above, I believe you referred to this segment of the UI as the _Specification window_ and here the term is _Specification section_. Calling out for general awareness. I prefer the term _Specification section_ because the term _window_ has certain implications as far as interfaces are concerned.

site/docs/concepts/web-app.md
```diff
@@ -0,0 +1,272 @@
...
+
+You use the table to monitor your materializations. It's nearly identical to the table on the [Captures page](#captures-page), with a few exceptions.
+
+![](<./webapp-images/materialization-page.png>)
+
+**1:** Select all or deselect all.
+
+**2:** Enable, Disable, and Delete buttons. These actions will be applied to the selected table rows. Choose **Disable** to temporarily pause the flow of data, **Enable** to resume, and **Delete** to permanently remove the materialization(s).
+
+**3:** Filter materializations by name.
```
kiahna-tucker, line 177: See the comment left behind in the _Captures page_ section regarding the search bar.

site/docs/concepts/web-app.md
```diff
@@ -0,0 +1,272 @@
...
+
+**4:** Shard status indicator. Shows the status of the task [shard](./advanced/shards.md) that backs this materialization.
+
+* **Primary (Green)**: Data is actively flowing through the materialization.
+* **Pending (Yellow)**: The materialization is attempting to re-connect. Often, you'll see this after you re-enable the materialization as Flow backfills historical data.
+* **Failed (Red)**: The materialization has failed with an unrecoverable error.
+* **Disabled (White)**: The materialization is disabled.
```
kiahna-tucker, line 184: See the comment left behind in the _Captures page_ section regarding disabled and _unknown_ shard status indicators.

kiahna-tucker (commented):
Before publishing these docs, I would like to check in with the team regarding our stance on collection editing in the UI at this point in time. It was my understanding that we wanted to introduce a read-only collection view in the immediate term, encouraging users to edit collections via the CLI. Basic collection editing capabilities in the UI is a feature we would like to have around the open beta release, but we have yet to determine how that should be presented in the UI in task workflows.

site/docs/concepts/web-app.md
oliviamiannone: Added a Tip here, but I didn't make a note in every location where a two-paneled view is used... This is because I do think those UX tweaks made it pretty obvious. If people are still confused going forward I can call this out more frequently.

site/docs/concepts/web-app.md
oliviamiannone: good catch - I think it would definitely be a beneficial exercise at some point to decide on names for different types of UI elements that we use (or maybe y'all already have concrete terminology, in which case please let me know!)

site/docs/concepts/web-app.md
oliviamiannone: Doing this edit really highlighted the difference in naming convention between captures and materializations -- namely that captures have the connector name appended to them, and materializations do not. Search seems like a good reason to have the connector name included for both entities.

oliviamiannone:
@kiahna-tucker @travjenkins Thanks for the reviews! All comments have been addressed. 
I know there are definitely lots of changes coming, large and small, especially with edit/drafts/etc. I'll keep a close eye on your repo, and definitely loop me in as things go to prod. 
I do think, in general, it's better for us to post these docs sooner than later and risk sections going out-of-date rather than wait for things to be 100% stable before posting anything.

site/docs/concepts/web-app.md
oliviamiannone: Question here - I mentioned the use of the `*` wildcard in my revision. Is there a list of supported wildcards that I should call out?

site/docs/concepts/web-app.md
```diff
@@ -93,11 +98,15 @@ All the collections of the selected capture(s) will be added to the materializat
 
 When you click **View Details** for a capture, the **Status** and **Specification** viewers are revealed.
 
-The **Status** window shows the full identifier of the shard that backs your capture. If there's an error, you can view its logs.
+The **Status** window shows the full identifier of the shard(s) that back your capture. If there's an error, you'll see an alert next to the shard identifier. Use the drop-down to open an expanded view of the failed shard's logs.
```
kiahna-tucker, line 101: I would also replace the instance of _window_ here with _section_.
<br />

>you'll see an alert next to the shard identifier.

The reference to the alert's positioning relative to the shard ID feels incorrect. I think someone would anticipate the alert appearing within the _Shard Information_ table, given that is where shard IDs are primarily surfaced. What about something a little more general: `If there's an error, you'll see an alert corresponding to the failing shard(s).` or `If there's an error, you'll see an alert identifying the failing shard(s).`

site/docs/concepts/web-app.md
```diff
@@ -174,15 +183,20 @@ You use the table to monitor your materializations. It's nearly identical to the
 
 **2:** Enable, Disable, and Delete buttons. These actions will be applied to the selected table rows. Choose **Disable** to temporarily pause the flow of data, **Enable** to resume, and **Delete** to permanently remove the materialization(s).
 
-**3:** Filter materializations by name.
+**3:** Filter materializations by name. Type a catalog prefix or unique materialization name to return materializations that match your query.
```
kiahna-tucker, line 186: >Filter captures. Type a catalog prefix, unique capture name, or connector name to return captures that match your query.

The same filter criteria applies for the materialization table. You can filter by connector name here as well.

site/docs/concepts/web-app.md
```diff
@@ -198,7 +212,8 @@ You use the table to monitor your materializations. It's nearly identical to the
 
 When you click **View Details** for a materialization, the **Status** and **Specification** viewers are revealed.
 
-The **Status** window shows the full identifier of the shard that backs your materialization. If there's an error, you can view its logs.
+The **Status** window shows the full identifier of the shard(s) that backs your materialization.
```
kiahna-tucker, line 215: See the comment left behind in the Captures page section regarding the _Status_ section of the details panel.

kiahna-tucker (approved):
Approved with comments. There are a couple tweaks that could be made to the docs, but they are not critical.

site/docs/concepts/web-app.md
kiahna-tucker: For sure! Travis and I have some _working_ names for aspects of the UI. It would be a great exercise to document those terms and refine them with you.

site/docs/concepts/web-app.md
kiahna-tucker: Okay, excellent. Glad to hear the changes made to that divider improved this UX.

site/docs/concepts/web-app.md
kiahna-tucker: That is a great question. @travjenkins can you think of any other wildcards worth mentioning?

site/docs/concepts/web-app.md
kiahna-tucker: Well, right now you can enter a materialization connector name directly in the search bar and it will filter by connector accordingly. The search bar is essentially looking for query matches in the catalog name **and** connector name of the data populating a given task table. For example, if you type _google sheets_ into the search bar of the materialization table, you will see that many of the results are connector name matches (i.e., the catalog name of the match does not contain the string _google sheets_.

site/docs/concepts/web-app.md
travjenkins: This _is_ a great question... I never even tried using them before honestly. 

We use the following query when typing in the search bar: `${param}.ilike.*${searchQuery}*` which handled by PostgREST and replaces the `*` with `%` on the backend. So I feel like anything that would work in a Postgres `ilike` contained with `%` would work here... but have not manually tested all that.

oliviamiannone: I can come back to this and play around with it another time. I definitely think `*` is the most important one to document.

PR #845, opened 2022-12-14
Docs: schema updates

oliviamiannone:
**Description:**

Updates conceptual Schema docs for two recent PRs:

* https://github.com/estuary/flow/pull/826 
* https://github.com/estuary/flow/pull/825

**Notes for reviewers:**

* My schema-writing skills aren't the strongest - I created samples in both new sections but please let me know if they don't make sense / if you have a better, more robust example I can sub in. 
* Writing about `default` annotations caused me to pause on an older line in the docs: _"Flow extends JSON Schema with additional annotation keywords, which provide Flow with further instruction for how documents should be processed."_ This feels like it should be followed by a comprehensive list - I know about `reduce` and now `default` but are there other Flow-specific keywords we should call out? Wasn't able to find in the code.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/845)

### DIFF
--- a/site/docs/concepts/schemas.md
+++ b/site/docs/concepts/schemas.md
@@ -6,7 +6,7 @@ sidebar_position: 7
 Flow documents and [collections](collections.md) always have an associated schema
 that defines the structure, representation, and constraints
 of your documents.
-Collections must have one schema, but [may have two](#write-and-read-schemas) for different types of Flow tasks.
+Collections must have one schema, but [may have two distinct schemas](#write-and-read-schemas): one for when documents are added to the collection, and one for when documents are read from that collection.
 
 Schemas are a powerful tool for data quality.
 Flow verifies every document against its schema whenever it's read or written,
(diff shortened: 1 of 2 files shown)


williamhbaker:
The `default` annotation is actually part of the standard json schema specification, and we're just able to use it in certain circumstances now. So it's not a special added annotation like `reduce`, which isn't part of normal json schema.

As far as I know the special annotations that flow understands are here: https://github.com/estuary/flow/blob/2c0523f0dd4be8894af89a0456534f4c87f241d3/crates/doc/src/annotation.rs#L6-L31

The only one that seems relevant to point out to users of flow would be `reduce`. The rest are used for connector specs.

site/docs/concepts/schemas.md
```diff
@@ -312,6 +351,29 @@ oneOf:
...
+collections:
+  acmeCo/coyotes:
+    schema:
+      type: object
+      required: [id]
+      properties:
+        id: {type: integer}
+        anvils_dropped: {type: integer}
+          reduce: {strategy: sum }
+          default: 0
```
williamhbaker, line 374: This seems like a good example. It could get more complicated than this but for the vast majority of users I think they would be using defaults in exactly this way so I think its good to keep things easy to understand.

williamhbaker (approved):

site/docs/concepts/schemas.md
```diff
@@ -312,6 +351,29 @@ oneOf:
...
+    key: [/id]
+```
+
+`default` annotations are only used for materializations; they're ignored by captures in derivations.
```
jgraettinger, line 378: by captures _and_ derivations

site/docs/concepts/schemas.md
```diff
@@ -211,6 +214,42 @@ but you can also use absolute URLs to a third-party schema like
...
+        id: { type: integer}
+        price: {type: string}
+    readSchema:
+      type: object
+      title: Prices as numbers
+      properties:
+        id: { type: integer}
+        price: {type: number}
+    key: [/id]
+```
```
jgraettinger, line 251: Other bits of further detail to perhaps note:

* The collection key must be defined by both the write and read schema.

* Projections of the collection generally must only exist / be defined by the read schema, but _logical partitions_ must additionally be defined by the write schema.

PR #849, opened 2022-12-15
gh-webhook: trigger a connector_tags re-sync on ghcr package webhook

### LINKED ISSUE #850: listen to docker image changes on ghcr and trigger connector_tag updates
At the moment it's a manual process to ask the agent to update a connector's spec, etc. by updating `job_status:{"type":"queued"}`. 

GitHub API supports [setting up webhooks](https://docs.github.com/en/rest/webhooks?apiVersion=2022-11-28), so we can use that to automatically listen on changes to image updates.

Alternatively we can just poll for updates to a connector image, and trigger a job if we detect it has changed. This approach works for local instances as well.

We want a way for this to work on local instance and GitHub Codespaces too since this is a major friction point when iteratively testing a connector.

mdibaiee:
**Description:**

- Introduce a new supabase function which can be called to trigger a connector_tags job for the agent for a given `package_url`
- The schema of the input is compatible with GitHub webhooks for the event `package`, _however_, GitHub does not allow custom headers to be set when creating a webhook, and Supabase makes `Authorization` header mandatory, so alternatively, I'm using a GitHub action to send a HTTP request, see https://github.com/estuary/airbyte/commit/033d6ce9583eae142f39701efd905451a17df37c
- Resolves #850 

**Workflow steps:**

- Send a HTTP request in the form of:
```
curl -i --location -XPOST 'http://localhost:5431/functions/v1/' --data @gh-webhook.json --header 'Authorization: Bearer TOKEN' --header 'Content-Type: application/json'
```
Where `gh-webhook.json` is:
```
{
  "action": "published",
  "package": {
    "package_version": {
       "package_url": "ghcr.io/estuary/source-hello-world:v1"
     }
  }
}
```

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/849)

### DIFF
--- a/supabase/functions/gh-webhook/index.ts
+++ b/supabase/functions/gh-webhook/index.ts
@@ -1,0 +1,50 @@
+// Follow this setup guide to integrate the Deno language server with your editor:
+// https://deno.land/manual/getting_started/setup_your_environment
+// This enables autocomplete, go to definition, etc.
+
+import { serve } from "https://deno.land/std@0.131.0/http/server.ts"
+import { supabaseClient } from "../_shared/supabaseClient.ts";
+
+console.log("GitHub webhooks functions started.")
+
+serve(async (req) => {
+  const { action, ...body } = await req.json()
+
+  const package_url: string = body.package.package_version.package_url;
+  const [image_name, image_tag] = package_url.split(':');
+  const { data: connector, error }: { data: { id: string, image_name: string } | null, error: any } = await supabaseClient
+    .from("connectors")
+    .select("id,image_name")
+    .eq("image_name", image_name)
+    .single();
+
+  if (error != null) {
+    console.error("error finding connector", error);
+  }
+
+  const { data: connector_tag, error: err }: { data: object | null, error: any } = await supabaseClient
+    .from("connector_tags")
+    .update({ job_status: {"type":"queued"}})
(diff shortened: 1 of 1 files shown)


jshearer (approved):
LGTM

supabase/functions/gh-webhook/index.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  const package_url: string = body.package.package_version.package_url;
+  const [image_name, image_tag] = package_url.split(':');
+  const { data: connector, error }: { data: { id: string, image_name: string } | null, error: any } = await supabaseClient
+    .from("connectors")
+    .select("id,image_name")
+    .eq("image_name", image_name)
+    .single();
+
+  if (error != null) {
```
jshearer, line 21: I would expect a return or throw here, otherwise we're going to continue on to the update request with a null/undefined `connector`, and try to do a `update connector_tags where connector_id = null/undefined` which seems not great

jshearer (changes requested):
Found something that seems worth changing

supabase/functions/gh-webhook/index.ts
mdibaiee: Good point thank you!

PR #853, opened 2022-12-16
fix: Typo in DPG's `--inference-address`

jshearer:
We already append `http://` in DPG, so adding it here too broke things

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/853)


psFried (approved):
LGTM

PR #854, opened 2022-12-16
fix: Fix decrypting empty sops documents

jshearer:
**Description:**

For security reasons, we appear to have strict controls on the size of the buffer that the decrypted document data can be written into. It turns out that in the case of an _empty_ input document that had been encrypted with `sops`, jq was writing a newline, which was exceeding the number of bytes were expecting out.

Turns out the solution was quite simple. `jq` supports the `-j` flag which:

>  #### `--join-output / -j`:
> Like -r but jq won´t print a newline after each output.
> #### `--raw-output / -r`:
> With this option, if the filter´s result is a string then it will be written directly to standard output rather than being formatted as a JSON string with quotes. This can be useful for making `jq` filters talk to non-JSON-based systems.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/854)


go/connector/decrypt.go
```diff
@@ -55,7 +55,9 @@ func DecryptConfig(ctx context.Context, config json.RawMessage) (json.RawMessage
 	// values in our own heap, and is also succinct.
 	// See: https://jqplay.org/s/sQunN3Qc4s
 	stripped, err := decryptCmd(ctx, decrypted, "jq",
-		"-c",
+		// -j puts jq into raw output mode, and additionally stops it from writing newlines
+		// at the end of its output, which were causing buffer overflow issues
+		"-j",
```
jgraettinger, line 60: nit:
 - please use the full argument name: `--join-output`.
 - Rather than saying "which were causing buffer overflow issues", which is historical context that future readers won't know or care about, say "which can otherwise cause the output to be longer than the input, prohibited by `decryptCmd`"

jgraettinger (approved):
Awesome, LGTM % nit below

PR #856, opened 2022-12-16
Add ops_catalog_template to control-plane DB

psFried:
**Description:**

This adds the ops_catalog_template table, some associated SQL functions, and a bash script for generating SQL migrations when the ops catalog changes. The agent now uses the `create_ops_publication` SQL function in order to publish ops catalogs.

Ops catalog updates are now handled as SQL migrations. The `generate-migration.sh` script was added to automatically generate these SQL migrations from the ops catalog sources. After running one of those migrations, there should be queued publications for every existing tenant in the system.

**Workflow steps:**

Whenever the sources in `ops-catalog/` are changed, run:

```
./ops-catalog/generate-migration.sh local | psql 'postgresql://postgres:postgres@localhost:5432/postgres'
```

which will generate the SQL migration. For the production environment, replace `local` with `prod` and switch to the prod DB URL.

**Notes for reviewers:**

I realized while I was working on this that our existing SQL tests were not being run in CI, and they were of course broken. So this includes fixes for the existing tests, and a change to have all SQL tests run in CI.

The ops repo will also need to be updated in order to deploy this, since it's currently downloading and bundling the ops catalog as part of the agent startup command, which isn't going to work anymore.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/856)


supabase/migrations/17_ops_catalogs.sql
```diff
@@ -0,0 +1,128 @@
...
+		from live_specs
+		where catalog_name like ('ops/' || tenant_prefix || '%');
+
+	insert into draft_specs (id, draft_id, catalog_name, spec_type, spec)
+		(
+		select
+			internal.id_generator() as id,
+			new_draft_id as draft_id,
+			"key" as catalog_name,
+			'capture'::catalog_spec_type as spec_type,
```
jgraettinger, line 50: is the `::catalog_spec_type` cast required, or just here as documentation? I'd think that would work implicitly.

supabase/migrations/17_ops_catalogs.sql
```diff
@@ -0,0 +1,128 @@
...
+security definer;
+
+comment on function internal.create_ops_publication is '
+Creates a new publication of the ops catalog template for a specific tenant.
+This publication will include all specs from the bundled_catalog in ops_catalog_template.
+Any _other_ specs under the ops/<tenant>/ prefix will be deleted by this publication.
+';
+
+
+create function internal.update_ops_catalog_template(bundled_catalog_arg json)
```
jgraettinger, line 91: nit: this function doesn't carry a lot of water, IMO, and its body should be lifted to the migration script -- which, as i understand things, would be the only non-test place we ever actually update this. I don't feel very strongly though if you'd rather keep it (but in that case it should probably have test coverage).

supabase/migrations/17_ops_catalogs.sql
```diff
@@ -0,0 +1,128 @@
...
+create function internal.republish_all_ops_catalogs(ops_user_id uuid)
+returns setof flowid as $$
+declare
+	current_tenant tenants;
+	ops_pub_id flowid;
+begin
+	for current_tenant in
+		select * from tenants
+	loop
+		ops_pub_id = internal.create_ops_publication(current_tenant.tenant, ops_user_id);
```
jgraettinger, line 116: FYI `current_tenant.tenant` is already the `catalog_tenant` domain type. The way this is called from the `agent` also should be compatible.

supabase/migrations/17_ops_catalogs.sql
```diff
@@ -0,0 +1,128 @@
...
+end;
+$$ language plpgsql
+security definer;
+
+comment on function internal.update_ops_catalog_template is '
+This function just makes migrations which update the ops catalog template look a little cleaner.
+It just upserts the bundled catalog into ops_catalog_template.
+';
+
+create function internal.republish_all_ops_catalogs(ops_user_id uuid)
```
jgraettinger, line 107: I also don't think this needs to be a function. As with `update_ops_catalog_template`, the _only_ place we expect to call it from is the migration script, so why not have it there? IMO not having this indirection would make the migration script itself more readable.

(I was thinking you had it here so you could test over it, but I'm not seeing that below)

supabase/tests/directives.test.sql
```diff
@@ -14,9 +14,9 @@ begin
     ('aliceCo/', 'otherCo/', 'write'),
     ('aliceCo/','bobCo/',  'read');
 
-  insert into directives (catalog_prefix, spec, token, single_use) values
-    ('aliceCo/', '{"type": "alice"}', 'aaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaaaa', true),
-    ('bobCo/', '{"type": "bob"}',   'bbbbbbbb-bbbb-bbbb-bbbb-bbbbbbbbbbbb', false);
+  insert into directives (catalog_prefix, spec, token, uses_remaining) values
```
jgraettinger, line 17: 👍 thanks

jgraettinger (approved):
Some minor comments below suggesting improvements and quibbling over where stuff lives, but big picture this is perfect.

LGTM

supabase/migrations/17_ops_catalogs.sql
psFried: It was returning an error about mismatched types before I added the cast, but I never really dug into exactly why. I'd also expected it to work without the explicit cast :man_shrugging:

PR #859, opened 2022-12-22
docs: flowctl updates

oliviamiannone:
**Description:**

Documents the set of changes from:
* https://github.com/estuary/flow/pull/837
* https://github.com/estuary/flow/pull/828

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/859)

### DIFF
--- a/site/docs/concepts/flowctl.md
+++ b/site/docs/concepts/flowctl.md
@@ -49,7 +49,7 @@ Important top-level flowctl subcommands are described below.
 It's also how you provision Flow roles and users. Learn more about [authentication](../reference/authentication.md).
 
 * `catalog` allows you to work with your organization's current active catalog entities. You can investigate the current Data Flows,
- or add their specification files to a **draft**, where you can develop them further.f
+ or add their specification files to a **draft**, where you can develop them further.
 
 * `collections` allows you to work with your Flow collections. You can read the data from the collection and output it to stdout, or list the [journals](../concepts/advanced/journals.md) or journal fragments that comprise the collection. [Learn more about reading collections with flowctl](../concepts/collections.md#using-the-flowctl-cli).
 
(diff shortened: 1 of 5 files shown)


psFried (commented):
I think the overall gist of my feedback is that I think we should start to emphasize `flowctl catalog` `test`, `publish`, and `pull-specs` instead of workflows using the `draft` subcommand. The `catalog` subcommands involve fewer steps and are less error prone than the equivalent workflows using `draft`, so I'm thinking we should present `draft` as being there to handle certain niche use cases such as:

- pulling a draft that was created by the UI discover workflow
- deleting catalog items from the CLI
- pushing to a draft in order to save your work for later (? IDK, I'm kinda stretching to think of other use cases)

But any of the common stuff that people will want to do, like pull down some files, edit them and re-publish, would all be easier and less error prone using the `catalog` commands, so I'm thinking that's the workflow we should lead with. Most of my inline comments are just trying to point out specific areas where that might help.

oliviamiannone:
@psFried In light of your comment (and giving this a bit of thought) I decided the best course of action might be to radically simplify things. I don't think before that I fully understood how little most users will _really_ need to use `draft` (deleting entities, IMO, is the most compelling thing to document). Let me know your thoughts on this edit.

I can come back in the future and add a deeper dive into how drafts are working and what the subcommands of `draft` do vs `catalog`... but I'm not sure that's even necessary at all? Curious to get your input here as a developer - would you expect to see that sort of help-system documentation for a CLI tool, or would you just prefer to look at the command line help as you use the tool?

psFried:
> the best course of action might be to radically simplify things

@oliviamiannone yea, I agree, though I also want to summon @dyaffe to get his thoughts on this. Personally, I think it makes sense to leave CLI _reference_ docs just in the `--help` output for now. But I think it also helps a lot to have tutorials and guides on the very basics, and I wonder what y'all think about approaching `flowctl` docs from that angle. I'm imagining having "how to" docs on (very roughly) this list:

- how to install and login
- how to edit a spec locally
- how to troubleshoot a task
- how to edit a draft that was created in the UI

LMK what you think, or if you have other ideas. I'd be happy to help put together some outlines if that would help clarify the new workflows.

dyaffe:
I'd agree that it would be a lot more simple to have a `delete` command @psFried.  

I love the option for the "how to" guides that you outlined @psFried -- maybe another would be "how to create your first derivation".

The reason that we should ever point people to `draft`, at least to a theoretical standpoint, would be to work on multiple entities at the same time.  It would be nice to highlight that to users if possible as an advanced setting and -- if we don't create that `delete` command right away, at least mention it's coming and say something like "for now there's this other clunky method to delete stuff".

Does that make sense @oliviamiannone and @psFried?

oliviamiannone:
@psFried Yeah, I like that idea of fleshing out more guides -- although I think it's ok to still merge this PR as-is and tackle that work separately?

Regarding the guides listed:
* How to install and login - this is covered in https://docs.estuary.dev/getting-started/installation/#get-started-with-the-flow-cli. 
* How to edit a spec locally - I'll add this one; should be straightforward. 
* How to troubleshoot a task & How to edit a draft that was created in the UI - I have a couple different questions about these and will reach out separately. 
* How to create your first derivation (@dyaffe) - This is covered https://docs.estuary.dev/guides/create-derivation/

I can create a sub-node for flowctl guides under **User Guides** to house the new ones and move the derivation one there, too.

psFried:
@oliviamiannone that all sounds good to me.

psFried:
@dyaffe You can work with multiple entities simultaneously with all the new commands. I think the reason to use drafts is really just for interop with the UI, or if you just want to save your work. Currently, you also need to use them to delete stuff, but a high level `delete` should be really easy to add.

psFried (approved):
Lots of good ideas for improvement here, but I think this is good to merge in the meantime.

PR #860, opened 2022-12-23
control-plane: log stats and discovery deep merge

jgraettinger:
**Description:**

* Add log statistics (warnings, errors, and shard failures) to the per-tenant derived stats collection and catalog-stats materialization.

* Introduce deep-merge semantics for discovery operations, and add a new `update_only` flag on discovery operations which refreshes only existing bindings and collections of the capture (without adding new ones).

**Workflow steps:**

* New `warnings`, `errors`, and `failures` statistic columns are available in the `catalog_stats` table and API.
* Discovery will automatically and intelligently merge capture and collection specifications.

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/860)


jgraettinger:
This change seems to break the UI. I suspect it's due to the new behavior with respect to how captures are named?:

image

Clicking next then gives:

image

I've confirmed that the `draft_specs` table is populated with the correct specs:

```
postgres=# select catalog_name from draft_specs where  draft_id = '08:5e:ff:91:f8:02:f8:00';
      catalog_name       
-------------------------
 johnnyCo/path/foobar
 johnnyCo/path/other
 johnnyCo/path/source-db
(3 rows)
```

Note these names are different as compared to what we currently produce, which is likely (?) the problem? Specifically given a discovered capture_name of `johnnyCo/path/source-db`, my capture is named exactly that now (instead of `johnnyCo/path/source-db/source-postgres`, and my collection is `johnnyCo/path/other` instead of `johnnyCo/path/source-db/other`, etc.

crates/agent/src/discovers/specs.rs
```diff
@@ -0,0 +1,373 @@
...
+                    write_schema: None,
+                    read_schema: None,
+                    key: models::CompositeKey::new(Vec::new()),
+                    projections: Default::default(),
+                    derivation: None,
+                    journals: Default::default(),
+                });
+
+        if collection.read_schema.is_some() {
+            collection.write_schema = Some(document_schema);
```
jshearer, line 154: What if the fetched collection already has a write schema? It looks like this will overwrite it? 

Or maybe the thinking is that there's never a circumstance where the write schema would be different from the document schema... but that can't be true because the whole point of write schemas was to allow for more permissive writes and resolve potential schema issues asynchronously on the read side.

jshearer (commented):
Ok so I read through all of this commit-by-commit and only found one place that I have a question about specifically. A lot of the bulk of the changes seemed to be around moving code around vs new/refactored code, so it's entirely possible that I missed something important.

One question I have: where is the change in discovered entity naming? I was looking for that, but maybe it's not obvious?

crates/agent/src/discovers/specs.rs
```diff
@@ -0,0 +1,373 @@
...
+}
+
+pub fn merge_capture(
+    capture_name: &str,
+    endpoint: models::CaptureEndpoint,
+    discovered_bindings: Vec<DiscoveredBinding>,
+    fetched_capture: Option<models::CaptureDef>,
+    update_only: bool,
+) -> (models::CaptureDef, Vec<DiscoveredBinding>) {
+    let capture_prefix = capture_name.rsplit_once("/").unwrap().0;
```
jgraettinger, line 54: This is the most-specific location of the changed behavior: we used to use the capture_name attached to the discover operation as the _entire_ prefix, to which we would suffix the trailing component of the connector name to form the capture name. Now we take it as the literal capture name, and used its base name as the `capture_prefix`.

Compare to the prior behavior of swizzle_response_to_specs (sp ?)

crates/agent/src/discovers/specs.rs
jgraettinger: It will be over-written by this, correct.

Conceptually, the capture connector "owns" the write schema. Connectors SHOULD implement fixed write schemas which are known to always-work (like allowing any json document, so long as it has meta fields of the filename and offset in, say, an S3 connector).

The read schema is owned by the user, and updated manually or perhaps through inference of actually-captured documents. That we definitely _don't_ want to over-write with the connector output, which is what happens today.

jshearer (approved):
Having tested this locally a bunch, it seems to work! LGTM, should be deployed with https://github.com/estuary/ui/pull/436 otherwise entity creation/editing will be broken in the UI

PR #863, opened 2022-12-28
Automatically fetch collections for TS generation

### LINKED ISSUE #862: `flowctl typescript generate` does not work well with read-only collections
Say you want to create a derived collection (`aliceCo/foo`), using a source collection that you only have read access to (`bobCo/bar`). What you'd do today is:

- run `flowctl catalog pull-specs --name bobCo/bar` to pull down the spec of the source collection
- define the spec for `aliceCo/foo` in `flow.yaml` or `aliceCo/flow.yaml`
- Run `flowctl typescript generate`, and implement your lambda(s)
- Run `flowctl test|publish`

The problem is that it's easy for users to end up in a scenario where `flowctl test|publish` will include the `bobCo/bar` spec, which results in an authorization error if you only have read access to it. A user would end up in that situation if they defined the spec for `aliceCo/foo` in `flow.yaml`, which will have an `import` of `bobCo/flow.yaml`. Or they might also have an `import` of `aliceCo/flow.yaml` and define their derivation spec there, which would also result in the authorization error.

I can see at least two perspectives on how we might tackle this, and I don't think they're mutually exclusive.

One perspective would be to say that we should just stop adding `import`s automatically, and thus also stop generating the top-level `flow.yaml` when you run `pull-specs`. The imports are never required, and the ergonomics of `--source flow.yaml` (which the imports afford us) don't seem significantly better than `--source-dir .`. And `--souce-dir` can generally solve the same problem of giving the user a way to apply multiple specs with a single command, so the top-level `flow.yaml` may not be missed.

Another perspective would be to ask why the user should have to run `pull-specs` in the first place. If all they want is to generate typescript classes for use in their derivation code, then we ought to teach `flowctl` to just fetch the source collection specs automatically. This way there's never a need to run `pull-specs` for things that you don't have `admin` access to (we could even emit a warning if you do) and thus you avoid the situation that gave rise to the authorization error.

Having `flowctl typescript generate` automatically fetch whatever specs it needs seems like an obvious win to me, but LMK what you think @dyaffe . I'd also be curious to hear from others on whether the top-level `flow.yaml` still makes sense now that imports are no longer required, or if you think we should implement that instead of, or in addition to, automatic fetching. Of course other ideas are welcome, too.

psFried:
**Description:**

Fixes #862

Teaches the `typescript generate` subcommand to automatically fetch collections that are used as transform sources. This allows users to create derivations of source collections without needing to first pull their specs. Specs will be resolved by making a separate request for collection that's named as the `source` of a transform but not already included in the sources. I used a separate request per collection because it simplifies the process of generating good error messages, and because the number of collections to be fetched is expected to be relatively small. An example error message is:

```
Error: resolving the collection 'estuary/public/wikipedia/recentchangess', referenced from the transform 'fromWiki' in derivation 'phil/test/wikipedia/test-derivation'

Caused by:
    the collection does not exist or you do not have access to it
```

I also updated `typescript generate` to use the same `SourceArgs` as other commands. A minor related change is that the `flow_generated` directory is now always placed relative to the current working directory instead of within the direct parent of the `--source` argument. This plays nicer with the common `SourceArgs` wich allow passing multiple `--source` arguments.

**Workflow steps:**

- Create a derivation spec locally that uses a `source` that _isn't_ defined locally.
- Run `flowctl typescript generate --source-dir .`
- You now have typescript generated for the non-local collection, in addition to the local one.

**Notes for reviewers:** n/a

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/863)

### DIFF
--- a/crates/flowctl/src/typescript/mod.rs
+++ b/crates/flowctl/src/typescript/mod.rs
@@ -1,5 +1,7 @@
+use crate::{catalog::CatalogSpecType, source::SourceArgs};
 use anyhow::Context;
 use proto_flow::flow;
+use std::path::Path;
 
 #[derive(Debug, clap::Args)]
 #[clap(rename_all = "kebab-case")]
@@ -27,11 +29,8 @@ pub enum Command {
 #[derive(Debug, clap::Args)]
 #[clap(rename_all = "kebab-case")]
 pub struct Generate {
-    /// Path to a local Flow catalog source file.
-    ///
-    /// TypeScript project files will be generated within its parent directory.
-    #[clap(long)]
-    source: std::path::PathBuf,
+    #[clap(flatten)]
+    source: SourceArgs,
 }
 
 impl TypeScript {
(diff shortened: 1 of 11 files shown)


crates/flowctl/src/typescript/mod.rs
```diff
@@ -101,8 +100,75 @@ pub async fn do_generate(
...
+            })?;
+            live_specs_catalog
+                .collections
+                .insert(source_name.clone(), resolved);
+        }
+    }
+
+    // If we've fetched any collection specs, then we'll need to load them now.
+    let resolved_tables = if !live_specs_catalog.collections.is_empty() {
+        // Write the bundled catalog to a temporary file. This is a little hacky, but Loader doesn't have a great way to
```
jshearer, line 128: Just for my own understanding when reading this comment, is there any actual file-writing going on here? It looks like we're serializing the catalog to bytes, and then loading those bytes from the "virtual" path `flowctl://resolved-catalog-sources`.

jshearer (approved):
This looks good as far as I can tell, and I love the usability improvement of automatically fetching resources as opposed to having to manually fetch them.

I'm a little concerned about this change
> A minor related change is that the `flow_generated` directory is now always placed relative to the current working directory instead of within the direct parent of the `--source` argument.

This feels like it could get confusing or frustrating if you're e.g `cd`'d into a directory somewhere making changes maybe to a collection spec, and then want to regenerate the typescript definitions. You forget to `cd` back up to the right directory, and now you have some `flow_generated` junk to clean up, as well as worrying about "what other files did this command add in the wrong place?"

I would expect this to work more like `npm install` or ` yarn install`, where it writes to the `node_modules` folder in the same directory as your `package.json` no matter how many directories deep you are. Thoughts?

crates/flowctl/src/typescript/mod.rs
psFried: Whoops, that was an old comment. I just updated it

psFried:
> This feels like it could get confusing or frustrating if you're e.g cd'd into a directory somewhere making changes maybe to a collection spec, and then want to regenerate the typescript definitions. You forget to cd back up to the right directory, and now you have some flow_generated junk to clean up, as well as worrying about "what other files did this command add in the wrong place?"
> 
> I would expect this to work more like npm install or  yarn install, where it writes to the node_modules folder in the same directory as your package.json no matter how many directories deep you are. Thoughts?

To be clear, the behavior in this particular situation hasn't changed. If you cd into a subdirectory and run `flowctl typescript generate --source flow.yaml`, it'll place the `flow_generated` directory right next to that source file both before and after this PR. What has changed is only if you cd into a parent directory. For example, say you have `~/projects/my-flow-project/flow.yaml` and you cd into `~/projects` and run `flowctl typescript generate --source my-flow-project/flow.yaml`. You'd then end up with `~/projects/flow_generated` instead of `~projects/my-flow-project/flow.yaml`.

That said, I also agree that we should probably talk about adding some more formalized concept of a "flow project directory" so that we can more easily deal with all these relative path references. And I think that would resolve the issue when generating typescript from some other directory. I'd been thinking of this as a separate topic, though, as it'll have much broader implications. So I'm inclined to say we should hold our noses for now, and plan on adding some notion of a "project directory" soon. Does that seem reasonable to you?

jshearer:
Yeah I'm fine with that. FWIW I _do_ think that creating `~projects/flow_generated` in that scenario feels worse than `~projects/my-flow-project/flow_generated`, but happy to defer this until we formalize the meaning of a flow project directory.

psFried:
> FWIW I do think that creating ~projects/flow_generated in that scenario feels worse than ~projects/my-flow-project/flow_generated

Agreed. But IMO it's still better than the old behavior, on balance. Consider a typical directory where we lay out specs in subdirectories by prefix (e.g. (`acmeCo/marketing/`, `acmeCo/sales/`). If you want to generate typescript for just one of those, you'd run `flowctl typescript generate --source acmeCo/marketing/flow.yaml` and with the old behavior you'd end up with `./acmeCo/marketing/flow_generated/`, which seems at least as gross.

PR #864, opened 2022-12-28
parser: improve parser schema for UI display

williamhbaker:
**Description:**

Adjusts the generated parser schema so that it renders correctly in the UI.

Recently the UI became more strict about showing errors for schema that it considers invalid. So we now have some connectors that are showing errors in the configuration pages, see https://github.com/estuary/connectors/issues/467. This updates the `parser` schema so that it can be shown correctly. It seems that the UI needs a `"type": "object"` for the `format` property, and also doesn't want the `type` to be removed from the `type` objects it contains.

**Workflow steps:**

Open the parser config in the UI and observe that there are no more error messages shown.

N/A

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/864)

### DIFF
--- a/crates/parser/src/config/mod.rs
+++ b/crates/parser/src/config/mod.rs
@@ -2,7 +2,11 @@ pub mod character_separated;
 pub mod protobuf;
 
 use encoding_rs::Encoding;
-use schemars::{gen, schema as schemagen, JsonSchema};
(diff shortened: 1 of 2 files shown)


psFried (approved):
LGTM

PR #866, opened 2023-01-02
Phil/wasm lib

psFried:
**Description:**

Introduces the `flow-web` crate, which produces a NPM package that can be used by our UI to call Rust code directly in the browser. This is currently an extremely minimal proof of concept, which only exposes a schema inference function. The idea is to land this minimal WASM library so that we can start publishing an NPM package, which we can then iterate on as needed.

The NPM package will theoretically be published to github packages on each successful build from the `master` branch. That actions workflow only runs when files under `crates/flow-web/` are changed, or if the top-level `Cargo.lock` changes.

**Notes for reviewers:**

This includes a few other changes that were required in order to get things compiling to WASM. See those individual commit messages for details.

Some notable things that are missing from this PR:

- There's no meaninful Typescript classes that are generated at this time. The function accepts and returns `any`. This is because the code generation doesn't really work well with structs having embedded arbitrary JSON values. We can probably improve on this in the future if desired.
- We'll likely need to include some more flow crates in the future in order to make this really useful. `validation` is probably a good next goal for WASM.
- The error handling approach here hasn't been especially well thought out, and may need revision once we start integrating this into the UI. I think it makes sense to defer until then, though.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/866)

### DIFF
--- a/.github/workflows/flow-web.yaml
+++ b/.github/workflows/flow-web.yaml
@@ -1,0 +1,50 @@
+name: Flow-web CI
+
+# Controls when the action will run. Triggers the workflow on push
+# or pull request events, but only for the primary branch.
+on:
+  push:
+    branches: [master]
+    paths:
+      - "crates/flow-web/**"
+      - "Cargo.lock"
+  pull_request:
+    branches: [master]
+    paths:
+      - "crates/flow-web/**"
+      - "Cargo.lock"
+
+jobs:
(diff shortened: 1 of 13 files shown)


jgraettinger (approved):
LGTM

😅 A lot of baggage that comes along with WASM

PR #867, opened 2023-01-02
runtime: refactor materialization protocol

jgraettinger:
**Description:**

Tweak the materialization protocol in a backwards-compatible way, to simplify the requirements placed on a materialization connector driver. To wit, a driver is allowed to be fully synchronous if it chooses, but may alternatively elect to commit asynchronously in the background while also processing an optimistically pipelined transaction.

To achieve this:
 * Checkpoints are shifted to a later phase of the protocol.
 * The runtime is required to send `Acknowledge` before any further `Load` requests -- which means that a driver is not obligated to read request from the runtime until it's ready to handle those loads.
 * The ordering of Acknowledge / Acknowledged is loosened to allow either to be written first.

Update the SQLite materialization accordingly. SQLite will be removed soon (and refactored into an image connector), but it's updated here for the improved testing coverage during this transition. Also update the implementation to use an actual gRPC loopback rather than a gRPC simulacrum built on Go channels.

**Migration**: To preserve compatibility with existing connectors, this implementation supports runtime and driver checkpoints in _either_ the old location or the new one. We'll need to update all connectors, switching them over to the new locations, and then we can remove this compatibility shim. This will unblock a further simplification of the runtime (removing handling for stats which cannot be known yet).

**Testing**: Tested with a local stack using:
 * The current `materialize-postgres` connector
 * A delta-updates materialization of `source-hello-world`, having a multi-second minimum transaction duration
 * Manually un-assigning the materialization shard many times, and verifying correctness of the materialized table (no skipped rows or duplicates)

Also add a new reference documentation page on the materialization protocol, with a fancy diagram, under Reference > Connectors > Materialization Protocol.

**Workflow steps:**

No user-facing changes.

Reference documentation page on the materialization protocol under Reference > Connectors > Materialization Protocol.

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/867)


go/protocols/materialize/client.go
```diff
@@ -3,9 +3,10 @@ package materialize
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
-	"math"
+	math "math"
```
psFried, line 9: Can't help but wonder about this. Why repeat the alias?

go/protocols/materialize/materialize.proto
```diff
@@ -529,14 +278,14 @@ message TransactionRequest {
...
-    // Flow checkpoint to begin processing from.
+    // Flow runtime checkpoint to begin processing from.
     // If empty, the most recent checkpoint of the Flow recovery log is used.
     //
     // Or, a driver may send the value []byte{0xf8, 0xff, 0xff, 0xff, 0xf, 0x1}
     // to explicitly begin processing from a zero-valued checkpoint, effectively
     // rebuilding the materialization from scratch. This sentinel is a trivial
     // encoding of the max-value 2^29-1 protobuf tag with boolean true.
-    bytes flow_checkpoint = 1;
+    bytes runtime_checkpoint = 1;
```
psFried, line 288: A _far_ superior name :monocle_face:

site/docs/reference/Connectors/materialization-protocol.md
```diff
@@ -0,0 +1,286 @@
+import Mermaid from '@theme/Mermaid';
+
+# Materialization Protocol
```
psFried, line 3: :+1: nice to see these docs growing up

psFried (approved):
I think this looks good, and is a good step forward both for connectors and the runtime.

go/protocols/materialize/client.go
jgraettinger: No reason, my IDE did this.

PR #870, opened 2023-01-03
Docs: Stats in ui + derivation shard status indicator

oliviamiannone:
**Description:**

Adds documentation for:

* Stats in UI https://github.com/estuary/ui/pull/422
* Shard status in UI for derivations: https://github.com/estuary/ui/pull/428

Stats change is reflected in the stats reference doc and UI concept doc; shard status change is just in the UI concept doc. Images in the UI concept doc were updated.

**Notes for reviewers:**

One or two reviews here will suffice but I included all 3 of you so that there are options. Either way, note that some assumptions were made here (most notably wrt which stats are being pulled).

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/870)


travjenkins (approved):
Everything looks good to me.

site/docs/concepts/web-app.md
```diff
@@ -155,20 +160,37 @@ You can proceed to the materialization, or opt to exit to a different page of th
...
-To reveal the **Specification** and **Data Preview** sections, expand **Details** next to a collection name.
+![](<./webapp-images/collections-page.png>)
 
-The **Specification** section shows the collection specification as JSON in a read-only editor.
-(If you need to modify a collection, edit the [capture](#editing-captures) that it came from.)
+**1:** Shard status indicator. If the collection does not contain a [derivation](./README.md#derivations), the indicator should always show green, and hover text will say "Collection."
+In the event that the server cannot be reached, the indicator will show "Unknown" status (black in light mode and white in dark mode).
 
-![](<./webapp-images/collections-view-schema.png>)
+  If the collection contains a derivation, the status of the derivation's primary task [shard](./advanced/shards.md) will be indicated:
```
kiahna-tucker, line 174: No issue with the content here, but the language used to discuss derivations is intriguing to me. It seems that there is some degree of _mental inception_ at play.

By the time the derivation docs are encountered, it could be argued that the user has the following, high-level modeling of Flow:

* An **entity** is a generic term that applies to all elements of the Flow ecosystem.

* A **task** is a subset of the _entity_ classification that applies to any _active_ element of the Flow ecosystem.

* A capture, materialization, and derivation are _tasks_ while a collection is merely an _entity_. Collections are explicitly labeled as the inert elements of the system.

Consequently, before one familiarizes themselves with the concept of a derivation, the relationship between an _entity_ and a _task_ is the same as that between a rectangle and a square: a _task_ is an _entity_, but an _entity_ is not inherently a _task_. 

From the derivation docs:

>A derivation is a [collection](https://docs.estuary.dev/concepts/#collections) that results from transformations applied to one or more other collections.

Here a derivation is identified as a member of the _collection_ subcategory of _entities_. This assertion cracks the clean partition that existed between _active_ system elements (i.e., tasks) and their _inert_ counterparts (i.e., collections), and the following quote gives the impression that the term _derivation_ can have multiple meanings.

> A derivation has two main components:
> 
> - The new collection, which stores the derived data.
> - A [catalog task](https://docs.estuary.dev/concepts/#tasks), which applies the transformations to source documents as they become available.

Now, a derivation is not simply _a collection_ nor is it a _task_, the classification to which it initially belonged. Instead, it painted as the encapsulation of a collection and a task... which is effectively a unique class of entity.

Reviewing conversations around derivations, I see the following approaches taken:

* **Derivation as a task** --  a derivation is regarded as the producer of a _derived collections_.

* **Derivation as a collection** -- a _pseudo-ambiguous_ transformation is regarded as the producer of a derivation (i.e., the collection).

* **Derivation as a task/collection hybrid** -- the term _derivation_ oscillates between the two _models_ above and depends on context. I can unpack this a bit more if needed, but this comment has now become the grand wall of text and there is some nuance here.

The following segment of the highlighted line of text got the wheels turning on the subject: _If the collection contains a derivation_. Given the language used to unpack the concept of a derivation, I was unsure whether it would be accurate to state that a collection can contain a derivation.

kiahna-tucker (approved):
Approved with a suggestion (and some _food for thought_ around the definition and application of the term _derivation_). I would recommend checking out the review comments before merging, but nothing is blocking the PR from being merged at the time of approval.

site/docs/concepts/web-app.md
```diff
@@ -155,20 +160,37 @@ You can proceed to the materialization, or opt to exit to a different page of th
...
+
+You can use the table to view each collection's specification and see a sample of its data.
 This can help you verify that collection data was captured as expected and that you'll be able to materialize it how you want, and troubleshoot it necessary.
 
-To reveal the **Specification** and **Data Preview** sections, expand **Details** next to a collection name.
+![](<./webapp-images/collections-page.png>)
 
-The **Specification** section shows the collection specification as JSON in a read-only editor.
-(If you need to modify a collection, edit the [capture](#editing-captures) that it came from.)
+**1:** Shard status indicator. If the collection does not contain a [derivation](./README.md#derivations), the indicator should always show green, and hover text will say "Collection."
```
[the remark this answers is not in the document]
oliviamiannone: Thanks for that suggestion - will refer to this generically as "status indicator" throughout. And I'll be on the lookout for updates to the UX.

site/docs/concepts/web-app.md
oliviamiannone: You're right -- this is for sure an area of conceptual confusion, although admittedly not one I've looked at recently. Appreciate you bringing it back up and I'm going to pin for later discussion after merging this PR. 

Originally, we *did* talk about derivations as tasks, which, as you point out, is a much cleaner concept for a new user to place in their worldview. We referred to collections as being either "derived collections" or "captured collections." But somewhere along the way, there was a conversation that led to the current conceptualization of derivation-as-hybrid-entity. Off the top of my head I don't remember the exact "why" behind this. At the basic level I think it had to do with the fact that derivations are defined _within_ collections; they can't exist separately, and that's something that's pretty unique to the Flow world (actually, the discussion in [this conversation](https://estuaryworkspace.slack.com/archives/C012S5U0WP2/p1671729373771889) may be relevant here). 

So then once you have an *advanced* user, referring to a derivation as a task (what once seemed intuitively true) actually becomes *more* misleading. Maybe?

I am also digressing at this point, but definitely pinning this discussion. And plan to clean up the docs on this matter once we've found some more clarity.

PR #874, opened 2023-01-05
docs: add mariadb and timescale

oliviamiannone:
**Description:**

Adds docs for source-mariadb and materialize-timescaledb

Mariadb setup is different enough from source-mysql, so it seemed worth providing a complete page for it (rather than simply linking to source-mysql docs).

 The Timescale docs are pretty similar to materialize-postgres, but for consistency they also get their own full, fleshed-out page.

**Notes for reviewers:**

I'm pretty confident in this one so I don't think it needs an in-depth review.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/874)

### DIFF
--- a/site/docs/reference/Connectors/capture-connectors/mariadb.md
+++ b/site/docs/reference/Connectors/capture-connectors/mariadb.md
@@ -1,23 +1,28 @@
 ---
-sidebar_position: 7
+sidebar_position: 6
 ---
 
-This is a change data capture (CDC) connector that captures change events from a MySQL database via the [Binary Log](https://dev.mysql.com/doc/refman/8.0/en/binary-log.html).
+# MariaDB
 
-It is available for use in the Flow web application. For local development or open-source workflows, [`ghcr.io/estuary/source-mysql:dev`](https://github.com/estuary/connectors/pkgs/container/source-mysql) provides the latest version of the connector as a Docker image. You can also follow the link in your browser to see past image versions.
+This is a change data capture (CDC) connector that captures change events from a MariaDB database via the [Binary Log](https://mariadb.com/kb/en/overview-of-the-binary-log/).
+It's derived from the [MySQL capture connector](./MySQL.md),
+so the same configuration applies, but the setup steps look somewhat different.
+
+This connector is available for use in the Flow web application. For local development or open-source workflows, [`ghcr.io/estuary/source-mariadb:dev`](https://github.com/estuary/connectors/pkgs/container/source-mariadb) provides the latest version of the connector as a Docker image. You can also follow the link in your browser to see past image versions.
 
 ## Prerequisites
-To use this connector, you'll need a MySQL database setup with the following.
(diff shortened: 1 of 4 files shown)


site/docs/reference/Connectors/capture-connectors/mariadb.md
```diff
@@ -0,0 +1,314 @@
...
+SET PERSIST binlog_expire_logs_seconds = 2592000;
+```
+4. Configure the database's time zone. See [below](#setting-the-mariadb-time-zone) for more information.
+```sql
+SET PERSIST time_zone = '-05:00'
+```
+
+### Setting the MariaDB time zone
+
+MariaDB's [`time_zone` server system variable](https://mariadb.com/kb/en/server-system-variables/#system_time_zone) is set to `SYSTEM` by default.
```
willdonnelly, line 63: I know this is copied from the MySQL equivalent, but re-reading it now it feels like it might be worth adding a bit more explanation about why we need time zone information here, and what the consequences are if it's not present. Something like:

"Flow captures `DATETIME` columns as RFC3339 timestamps, but `DATETIME` values have no associated time zone within MariaDB, so this is inferred from the system `time_zone` variable when it is set. If `time_zone` is unset or set to `SYSTEM` it will not be possible to capture tables with `DATETIME` columns, but all other column types will continue to work. Note that `TIMESTAMP` columns do not require the system `time_zone` to be set because they are normalized from the client time zone to UTC at insert time."

site/docs/reference/Connectors/capture-connectors/mariadb.md
```diff
@@ -0,0 +1,314 @@
...
+
+4. Switch to your MariaDB client. Run the following commands to create a new user for the capture with appropriate permissions,
+and set up the watermarks table:
+
+```sql
+CREATE DATABASE IF NOT EXISTS flow;
+CREATE TABLE IF NOT EXISTS flow.watermarks (slot INTEGER PRIMARY KEY, watermark TEXT);
+CREATE USER IF NOT EXISTS flow_capture
+  IDENTIFIED BY 'secret'
+  COMMENT 'User account for Flow MariaDB data capture';
```
willdonnelly, line 215: As above, user accounts can't have comments in MariaDB.

site/docs/reference/Connectors/capture-connectors/mariadb.md
```diff
@@ -0,0 +1,314 @@
...
+```
+
+4. Create the `flow_capture` user with replication permission, the ability to read all tables, and the ability to read and write the watermarks table.
+
+  The `SELECT` permission can be restricted to just the tables that need to be
+  captured, but automatic discovery requires `information_schema` access as well.
+```sql
+CREATE USER IF NOT EXISTS flow_capture
+  IDENTIFIED BY 'secret'
+  COMMENT 'User account for Flow MariaDB data capture';
```
willdonnelly, line 269: Also here.

willdonnelly (approved):
LGTM

PR #882, opened 2023-01-17
docs: source-salesforce

oliviamiannone:
**Description:**

Adds docs for salesforce connector. 
Most of this was originally drafted when we first tried to put this connector in (last summer), but from what I can tell not a lot has changed that's user-facing, so I just did a run-through and some edits throughout. 

We'll want to create a shortlink to use as documentationurl after merging.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/882)


site/docs/reference/Connectors/capture-connectors/salesforce.md
```diff
@@ -0,0 +1,135 @@
+# Salesforce
+
+This connector captures data from [Salesforce standard and custom objects](https://developer.salesforce.com/docs/atlas.en-us.238.0.object_reference.meta/object_reference/sforce_api_objects_concepts.htm) into Flow collections.
```
mdibaiee, line 3: At the moment we have limited the connector to only ingest these objects:

```
"Account",
"Contact",
"User",
"OpportunityFilledHistory",
"LeadHistory",
"Opportunity",
"Campaign",
"Case",
"ContactLineItem",
"Entitlement",
"Lead",
"LiveChatTranscript",
"MessagingSession",
"Quote",
"QuoteLineItem",
"ServiceAppointment",
"ServiceContract",
"Task",
"UserServicePresence",
"WorkOrder",
"WorkOrderLineItem"
```

So at the moment we don't support custom objects, and not all of the standard objects

mdibaiee (commented):
Thank you Olivia, there is one point about the available objects, the rest looks okay!

PR #883, opened 2023-01-18
crates/json: add integer and number string formats

williamhbaker:
**Description:**

Part of https://github.com/estuary/connectors/issues/505

Adds support for string formats of "integer" and "number". These will allow connectors to handle integers and numbers encoded as strings.

See related PR for the SQL materialization connectors that require this change: https://github.com/estuary/connectors/pull/507

**Workflow steps:**

N/A

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/883)

### DIFF
--- a/crates/json/src/schema/formats.rs
+++ b/crates/json/src/schema/formats.rs
@@ -43,6 +43,8 @@ pub enum Format {
     Regex,
     #[serde(rename = "relative-json-pointer")]
     RelativeJsonPointer,
+    Integer,
+    Number,
 }
 
 // Some are from https://github.com/JamesNK/Newtonsoft.Json.Schema/blob/master/Src/Newtonsoft.Json.Schema/Infrastructure/FormatHelpers.cs
@@ -155,6 +157,8 @@ impl Format {
             Self::RelativeJsonPointer => {
                 ValidationResult::from(RELATIVE_JSON_POINTER_RE.is_match(val).unwrap_or(false))
             }
+            Self::Integer => ValidationResult::from(true),
+            Self::Number => ValidationResult::from(true),
         }
     }
 }
--- a/crates/json/src/schema/formats.rs
+++ b/crates/json/src/schema/formats.rs
@@ -1,8 +1,10 @@
 use std::{net::IpAddr, str::FromStr};
 
 use addr::{parse_domain_name, parse_email_address};
+use bigdecimal::BigDecimal;
 use fancy_regex::Regex;
 use iri_string::spec::{IriSpec, UriSpec};
+use num_bigint::BigInt;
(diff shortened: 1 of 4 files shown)


crates/json/src/schema/formats.rs
```diff
@@ -155,6 +157,8 @@ impl Format {
             Self::RelativeJsonPointer => {
                 ValidationResult::from(RELATIVE_JSON_POINTER_RE.is_match(val).unwrap_or(false))
             }
+            Format::Integer => ValidationResult::from(true),
+            Format::Number => ValidationResult::from(true),
```
mdibaiee, line 161: nit: maybe just go with `Self::` to be consistent with the rest of the switch case

mdibaiee: and do we want to actually validate the values here? perhaps try to parse the values as float / integer and validate them as such

williamhbaker: My thought here was to leave validation to the individual connector, since different endpoints will have different requirements on what makes a valid integer or number with respect to the number of digits it can have.

jgraettinger: I think we should do basic validation that it's an integer or number even if we don't bound its length. A regex, perhaps.

Otherwise it's possible to capture data that is truly invalid, and then only find out about it at read-time when you're trying to use it.

jgraettinger: The `bigdecimal` and `bigint` crates would also be reasonable to add.

We'll need these anyway, in order to extend the semantics of the `sum` and related reduction annotations to cover string-encoded integers / numbers.

williamhbaker: I actually hadn't been thinking about the "capturing data" side of things at all so yeah that _does_ make sense to have some basic validation. I'll see what I can come up with.

williamhbaker:
Another validation-related point I need to consider: It should not be possible to format a field with types `[string, integer]` as `format: number`, for example. I'll need to add in validate for that as well.

jgraettinger:
> It should not be possible to format a field with types [string, integer] as format: number

I didn't quite understand this. You're saying it's not possible to use such a schema as a field, and should be? Or something else? Just to confirm, the `format` keyword applies only to a `string` type, so from a JSON schema perspective it's fine to have `{type: [string, integer], format: number}`

williamhbaker:
Ah you're right that `{type: [string, integer], format: number}` is fine from a JSON schema perspective, so we should be fine here. I'll need to think about it on the SQL materialization's side though.

mdibaiee (approved):

jgraettinger (approved):
LGTM

PR #884, opened 2023-01-18
db: Add `estuary_support/` role 

jshearer:
**Description:**

This role will allow us to temporarily grant ourselves access to every tenant for support

~~Actually assigning yourself to this role  currently breaks things because there are too many grants. https://github.com/estuary/ui/pull/452 is the fix for that.~~ https://github.com/estuary/ui/pull/452 is merged

**Notes for reviewers:**

I actually already created this function and trigger on the prod db in response to poking from Dave, but ended up having to unassign the role because it was causing the issue that's mentioned above. So no migration needed.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/884)


psFried (approved):
LGTM

PR #886, opened 2023-01-18
trigger flow-web CI when workflow file changes

psFried:
I'd pushed #885 to specify the credentials for publishing, but the publish workflow only triggers if one of the files in `crates/flow-web/` is modified. This broadens that check so that the workflow will also run if the workflow yaml itself is modified.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/886)


jshearer (approved):
👍

PR #889, opened 2023-01-20
crates/json: add fractional string format

williamhbaker:
**Description:**

Follow-up to d91472b9 which added integer and number formats, this also adds the "fractional" format which is available in Flow collection schemas.

Part of https://github.com/estuary/connectors/issues/505

**Workflow steps:**

N/A for this PR. https://github.com/estuary/connectors/pull/507 will describe the doc updates needed once it is merged.

**Notes for reviewers:**

### DIFF
--- a/crates/json/src/schema/formats.rs
+++ b/crates/json/src/schema/formats.rs
@@ -47,6 +47,7 @@ pub enum Format {
     RelativeJsonPointer,
     Integer,
     Number,
+    Fractional,
 }
 
 // Some are from https://github.com/JamesNK/Newtonsoft.Json.Schema/blob/master/Src/Newtonsoft.Json.Schema/Infrastructure/FormatHelpers.cs
@@ -163,6 +164,10 @@ impl Format {
                 ValidationResult::from(BigInt::parse_bytes(val.as_bytes(), 10).is_some())
             }
             Self::Number => ValidationResult::from(BigDecimal::from_str(val).is_ok()),
+            Self::Fractional => ValidationResult::from(match BigDecimal::from_str(val) {
+                Ok(d) => !d.is_integer(),
+                Err(_) => false,
+            }),
         }
     }
 }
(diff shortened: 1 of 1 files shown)


crates/json/src/schema/formats.rs
```diff
@@ -163,6 +164,10 @@ impl Format {
                 ValidationResult::from(BigInt::parse_bytes(val.as_bytes(), 10).is_some())
             }
             Self::Number => ValidationResult::from(BigDecimal::from_str(val).is_ok()),
+            Self::Fractional => ValidationResult::from(match BigDecimal::from_str(val) {
+                Ok(d) => !d.is_integer(),
```
mdibaiee, line 168: Do we really need to enforce non-zero fractional part here? I couldn't find any definition of `fractional` format, but I would find it unintuitive if I had a field which was in the range `0..1` and I chose `format: fractional` for it, and I see that `"0"` and `"1"` are rejected by the validator.

mdibaiee:
```
    println!("{:?}", BigDecimal::from_str("1").expect("parse").is_integer());
    println!("{:?}", BigDecimal::from_str("1.0").expect("parse").is_integer());
    println!("{:?}", BigDecimal::from_str("1.1").expect("parse").is_integer());
    println!("{:?}", BigDecimal::from_str("0.0").expect("parse").is_integer());
    println!("{:?}", BigDecimal::from_str("0").expect("parse").is_integer());
```
Output:
```
true
true
false
true
true
```

willdonnelly: I'm not familiar with the motivations here, but just from following this discussion I'm now of the opinion that `type: fractional` is so much of a footgun that it shouldn't even exist. And as far as I could tell from 30s of searching it doesn't look like it's a standard type anyway? Is there some specific use-case for which we need this?

williamhbaker:
Going to close this out since there isn't a really compelling reason to include support for formatting strings as fractional at the moment.

Issue #868, opened 2023-01-03
Allowing inbound TCP connections to connector containers


psFried:
### Discussed in https://github.com/estuary/flow/discussions/840

The high-level goal is to allow inbound networking for containers that we run on our platform.
This issue description will be kept up to date with the overall status. Let's use the discussion for resolving additional questions and feedback as it comes up.

### Status: Merged!

**PRs:**

- flow: estuary/flow#893
- data-plane-gateway: estuary/data-plane-gateway#27
- connectors: estuary/connectors#565

### TODO

- [x] Add tests for gRPC proxy services
- [ ] ~~Introduce `flowctl port-forward` subcommand to support protocols such as postgres, which do not use ALPN~~
- [ ] ~~Finish updating `ops/*/stats` collections to include networking stats, and publish stats as part of proxying~~
- [x] Update CertManager in the data-plane to allow using `dns-01` acme challenges, and provision a wildcard certificate for `*.combustible-cronut.estuary-data.dev`
- [x] Add the new data-plane-gateway deployment to combustible-cronut
- [x] Introduce a `source-http-ingest` connector that publishes data from HTTP requests

psFried:
This is now complete. There's still one related oustanding issue (estuary/data-plane-gateway#30), but the basic features are all present and working, so I'm closing this issue.

PR #893, opened 2023-01-24
Phil/connections into connectors

psFried:
**Discussion:** #840 
**Issue:** #868 

Introduces support for accessing connector containers from the public itnernet. This is a complex feature that requires changes in the flow, data-plane-gateway, and connectors repositories.

The basic gist is that connectors may opt to expose certain ports, making them available for connections from the public internet. Connectors indicate the ports they wish to expose by adding `EXPOSE <port-number>` directives to their Dockerfiles.

By default, all exposed ports are private, and the protocol is assumed to be http/1.1. The data-plane-gateway will require a bearer token that authorizes each request to access the task. Connectors may opt to make their port public, in which case no authorization check will be performed. This is done using a `LABEL` directive in the Dockerfile to set the label `dev.estuary.port-public.<port-number>=true` where `<port-number>` matches the numeric value of an `EXPOSE`d port.

Similarly, the connector may opt to use a protocol other than HTTP/1.1 by using `LABEL dev.estuary.port-proto.<port-number>=<protocol>`. Any TCP-based protocol ought to be supported, but data-plane-gateway is only able to enforce authorization when the protocol is `http/1.1` (the default) or `h2`. Thus, if a non-http protocol is used, then the port visibility _must_ be public.

UDP traffic is not supported at this time. TLS is required in order for data-plane-gateway to proxy the traffic.

Network connections are currently handled by tunneling the TCP connection through gRPC services running in the reactor and connector-init. Once data-plane-gateway accepts the TLS connection, it resolves the hostname (provided as the SNI in the tls client hello) to an exposed port of a running shard. It starts a bi-directional streaming gRPC with the reactor to which the shard is currently assigned. That reactor then forwards the streams to another gRPC service that's running in connector-init. The service in connector-init will establish a TCP connection with the target port that the connector is listening on and forward data between that and the reactor, and so on all the way back up the chain.

**Builds**

At build time, Flow now performs a `docker inspect` of all connector images in order to determine the ports they wish to expose. It performs a single `docker inspect` for each distinct image, and surfaces any errors during a later validation of the captures or materializations. The exposed port information is used to derive a set of shard labels containing information about exposed ports. These labels are used at runtime to drive service discovery within data-plane-gateway.

**Shard labels**

The label `estuary.dev/hostname` is added to each shard that exposes any ports. The value is currently just a hash of the _task_ name.

`estuary.dev/expose-port` is added for _each_ exposed port, and the value is the port number (inside the container) that the connector will listen on. Note that shard labels are a multiset, so it will be normal to have multiple labels each exposing a different port.

`estuary.dev/port-public/<port-number>` corresponds to the `dev.estuary.port-public.<port-number>` image label. This label is only added if the visibility is public, as the default is to consider all ports private.

Similarly, `estuary.dev/port-proto/<port-number>` corresponds to the `dev.estuary.port-proto.<port-number>` image label.

**Hostnames**

When a client establishes a connection to the container via data-plane-gateway, it uses a hostname in the format: `<shard-hostname>-<port-number>.<data-plane-hostname>`. The hostname is matched agains the `estuary.dev/hostname` label, and the port is matched against the `estuary.dev/expose-port` label. For example, given a data-plane gateway with the hostname `gcp-us-central1.v1.estuary-data.dev`, and a shard with the labels:

```
estuary.dev/hostname=123abc
estuary.dev/expose-port=9000
```

A client would use the hostname `123abc-9000.gcp-us-central1.v1.estuary-data.dev` to connect to to port 9000 in the running container. In this case, the Dockerfile of the connector would only need to `EXPOSE 9000` in order to allow that. Note that authentication would be required in this case, since there's no explicit configuration making this port public.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/893)


crates/proto-flow/src/flow.rs
```diff
@@ -1199,6 +1198,92 @@ pub struct IngestResponse {
...
+    #[derive(Clone, PartialEq, ::prost::Message)]
+    pub struct Open {
+        #[prost(string, tag="1")]
+        pub shard_id: ::prost::alloc::string::String,
+        #[prost(uint32, tag="2")]
+        pub target_port: u32,
+        /// The network address of the client that is establishing the connection.
+        #[prost(string, tag="3")]
+        pub client_addr: ::prost::alloc::string::String,
+        /// TODO: reorder and comment
```
williamhbaker, line 1219: Comment needs to be addressed?

crates/validation/src/capture.rs
```diff
@@ -39,27 +46,40 @@ pub async fn walk_all_captures<D: Drivers>(
...
         if capture_errors.is_empty() {
-            validations.extend(validation.into_iter());
+            validations.push((capture, bindings, req, port_configs));
         } else {
             errors.extend(capture_errors.into_iter());
         }
     }
 
-    // Run all validations concurrently.
+    // Run all validations concurrently. We do this even if
```
williamhbaker, line 78: Comment is cut off here

crates/validation/src/lib.rs
```diff
@@ -186,3 +200,95 @@ pub async fn validate<D: Drivers>(
...
+        return Err(Error::ImageInspectFailed {
+            image: image.to_owned(),
+            error: anyhow::anyhow!("expected 1 image, got {}", deserialized.len()),
+        });
+    }
+    let mut ports = BTreeMap::new();
+    for (port_config, _) in deserialized[0].config.exposed_ports.iter() {
+        // We're unable to support UDP at this time.
+        if port_config.ends_with("/udp") {
+            continue;
```
williamhbaker, line 243: Would we want to return an error here instead of filtering these out?

go/connector/container.go
```diff
@@ -34,6 +41,20 @@ type Container struct {
 	waitCh chan error
 }
 
+// ExposePorts is a handle to a network configuration process that can expose
+// the ports for a running container. This is currently done by `runtime.ProxyServer`,
+// which proxies traffic throught the `grpc.ClientConn` to the container.
+type ExposePorts interface {
```
williamhbaker, line 47: Is this interface only to resolve an import cycle from `runtime`?

go/ops/local_publisher.go
```diff
@@ -53,7 +53,13 @@ func (*LocalPublisher) PublishLog(log Log) {
 	logrus.StandardLogger().WithFields(fields).Log(level, log.Message)
 }
 
-/// logrusLogLevel maps the current Level of the logrus logger into a pf.LogLevel.
+// PublishStats implements Publisher
+func (*LocalPublisher) PublishStats(event StatsEvent) {
+	// TODO: will this ever get called?
```
williamhbaker, line 58: I'm not sure of the answer to this question myself but it feels weird to leave it like this.

go/ops/local_publisher.go
```diff
@@ -53,7 +53,13 @@ func (*LocalPublisher) PublishLog(log Log) {
...
+	logrus.WithField("stats", event).Error("got local stats event")
+}
+
+// / logrusLogLevel maps the current Level of the logrus logger into a pf.LogLevel.
```
williamhbaker, line 62: There's an extra `/` in front of this comment. The comment itself seems pretty redundant with the self-explanatory code also.

crates/flowctl/src/auth/mod.rs
```diff
@@ -47,6 +47,21 @@ pub enum Command {
     /// Unlike 'read' or 'write', the subject of an 'admin' grant also inherits
     /// capabilities granted to the object role from still-other roles.
     Roles(roles::Roles),
+
+    /// Fetches and prints an auth token that can be used to access a Flow data plane.
```
williamhbaker, line 51: Are the changes in this file related to connector networking? At first glance they seem unrelated and maybe not intentional.

williamhbaker (approved):
LGTM with a few questions & comments

crates/proto-flow/src/flow.rs
```diff
@@ -380,8 +380,6 @@ pub struct FieldSelection {
     pub field_config_json: ::std::collections::HashMap<::prost::alloc::string::String, ::prost::alloc::string::String>,
 }
 /// CaptureSpec describes a collection and its capture from an endpoint.
-///
-/// Next tag: 8.
```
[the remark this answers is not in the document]
psFried: One of those "Next tag" comments was already out-of-date, so I just figured it'd be better to remove them than to check in every PR whether someone forgot to update the comment.

go/labels/shard.go
```diff
@@ -97,3 +118,28 @@ func maybeOne(set pf.LabelSet, name string) (string, error) {
...
+
+func parsePorts(set pf.LabelSet) (map[uint16]*PortConfig, error) {
+	var out = make(map[uint16]*PortConfig)
+
+	for _, value := range set.ValuesOf(ExposePort) {
+		var portNumber, err = strconv.ParseUint(value, 10, 16)
+		if err != nil {
+			return nil, fmt.Errorf("parsing value '%s' of label '%s': %w", value, ExposePort, err)
+		}
+		if portNumber == 0 || portNumber > 65535 {
```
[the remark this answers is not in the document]
psFried: Nope, it should be fine. You still need permissions to bind to ports < 1024, but it's pretty normal for containers to do that, since the default root user has permission to do so.

go/labels/shard.go
```diff
@@ -24,6 +30,15 @@ type ShardLabeling struct {
 	TaskName string
 	// Type of this task (capture, derivation, or materialization).
 	TaskType string
+
+	// TODO: comment and add test for parsing host and ports
```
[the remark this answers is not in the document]
psFried: :facepalm:

go/runtime/proxy.go
```diff
@@ -0,0 +1,434 @@
...
+	"github.com/sirupsen/logrus"
+	pb "go.gazette.dev/core/broker/protocol"
+	"go.gazette.dev/core/consumer"
+	pc "go.gazette.dev/core/consumer/protocol"
+	"golang.org/x/sync/errgroup"
+	"google.golang.org/grpc"
+)
+
+type ProxyServer struct {
+	mu         *sync.Mutex
```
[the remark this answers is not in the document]
psFried: :+1: done

go/runtime/proxy.go
```diff
@@ -0,0 +1,434 @@
...
+func (ps *ProxyServer) checkShardStatus(ctx context.Context, open *pf.TaskNetworkProxyRequest_Open) (status pc.Status, header pb.Header, err error) {
+	ctx, cancelFunc := context.WithTimeout(ctx, 10*time.Second)
+	defer cancelFunc()
+
+	resolution, err := ps.resolver.Resolve(consumer.ResolveArgs{
+		Context:     ctx,
+		ShardID:     open.ShardId,
+		MayProxy:    false,
+		ProxyHeader: open.Header,
+		//ReadThrough: nil, because we don't care about the progress of the shard, only whether this instance is primary
```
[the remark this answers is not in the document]
psFried: This was intended to just document that we're intentionally using the zero-value of `ReadThrough`. Reworded to make that more clear.

go/runtime/ops_publisher.go
```diff
@@ -0,0 +1,202 @@
...
+		Doc:        doc,
+		Partitions: s.partitions,
+	}
+}
+
+// PrepareStatsJournal returns the journal, contentType, and a new Acknowledgment message for the
+// stats journal for this task. The journal is created if it does not exist. This is used in
+// conjunction with Publisher.DeferPublishUncommitted, which requires these things to be provided up
+// front.
+func (s *StatsFormatter) PrepareStatsJournal(mapper flow.Mapper) (journal pb.Journal, contentType string, ack flow.Mappable, err error) {
```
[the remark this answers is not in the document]
psFried: This had been part of `go/runtime/stats.go`, which I'd copied into here without realizing that it's no longer used. We used to have some weird gymnastics as part of publishing materialization stats, but that was removed a while ago with the change to the materialization protocol to have the flow checkpoint sent at a later point in the transaction lifecycle. Anyway, I just removed this whole function.

go/runtime/proxy.go
```diff
@@ -0,0 +1,434 @@
...
+
+func (h *networkConfigHandle) Expose(connection *grpc.ClientConn, logger ops.Publisher) {
+	if h.instanceVersion > 0 {
+		panic("expose called more than once")
+	}
+
+	h.server.mu.Lock()
+	defer h.server.mu.Unlock()
+
+	// copy the ports map to ensure it doesn't get modified while we're using it.
```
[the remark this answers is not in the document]
psFried: Ya know, that's a good point. On reflection, I don't think this copy is necessary at all. They come from the `taskTerm`, which intentionally never modifies them after initialization. I just updated it to use the same ports map instead of creating a new one.

crates/validation/src/lib.rs
psFried: I don't think so. I'm considering cases where you might use `FROM some-external-image` where `some-external-image` happens to expose a UDP port for some optional features, and you can't really control it. And in the event that we want to add support for UDP, I don't see a problem with going from "UDP ports are silently ignored" to "exposed UDP ports are private by default"

go/connector/container.go
psFried: Yeah, pretty much. I'd also considered making it into a more generic `ContainerObserver` interface, but I resisted the urge to prematurely generalize it.

go/ops/local_publisher.go
psFried: Just removed the comment. It's definitely not going to get called at this point.

crates/flowctl/src/auth/mod.rs
psFried: Yeah, this is related and intentional. See [this commit message](https://github.com/estuary/flow/pull/893/commits/86cd537ed02e0ce5e0d832b64c891593f1a28d56) for rationale.

go/runtime/proxy.go
```diff
@@ -0,0 +1,434 @@
...
+
+func (h *networkConfigHandle) Unexpose() {
+	if h.instanceVersion == 0 {
+		return // Expose was never called. This is normal if the container failed quickly.
+	}
+
+	h.server.mu.Lock()
+	defer h.server.mu.Unlock()
+
+	if previousInstance := h.server.containers[h.shardID]; previousInstance != nil {
```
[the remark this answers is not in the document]
psFried: Yep, good call :+1: I added some tests for that.

go/connector/container.go
williamhbaker: Gotcha. Ideally we could use the type directly instead of the interface IMO but not worth the refactoring to resolve the import cycle.

crates/flowctl/src/auth/mod.rs
williamhbaker: Ahhh I see, thanks!

PR #895, opened 2023-01-24
implement formal JSON connector protocol and make it opt-in

jgraettinger:
**Description:**

* Create a formal JSON protocol for captures and materializations, under the new `connector-protocol` crate, with support for JSON schema generation.

* Implement conversions between our existing protobuf protocol and this JSON protocol through a new `proto-convert` crate.

* Implement opt-in support for connectors to use the JSON protocol within `flow-connector-init`, toggled via image `LABEL FLOW_RUNTIME_CODEC=json`.

No changes for connectors that don't elect to use JSON.

I'll post gists shortly, with a "hello world" capture and materialization for demonstration purposes.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/895)

### DIFF
--- a/crates/connector-init/src/capture.rs
+++ b/crates/connector-init/src/capture.rs
@@ -1,4 +1,4 @@
-use super::rpc;
+use super::{codec::Codec, rpc};
 use futures::StreamExt;
 use proto_flow::capture::{
     ApplyRequest, ApplyResponse, DiscoverRequest, DiscoverResponse, PullRequest, PullResponse,
@@ -7,6 +7,7 @@ use proto_flow::capture::{
 
 pub struct Driver {
     pub entrypoint: Vec<String>,
+    pub codec: Codec,
 }
 
 #[tonic::async_trait]
@@ -15,7 +16,8 @@ impl proto_grpc::capture::driver_server::Driver for Driver {
         &self,
         request: tonic::Request<SpecRequest>,
     ) -> Result<tonic::Response<SpecResponse>, tonic::Status> {
-        let message = rpc::unary(&self.entrypoint, "spec", request.into_inner()).await?;
+        let message =
+            rpc::unary(&self.entrypoint, self.codec, "spec", request.into_inner()).await?;
         Ok(tonic::Response::new(message))
     }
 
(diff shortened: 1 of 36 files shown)


crates/connector-init/src/capture.rs
```diff
@@ -15,39 +16,65 @@ impl proto_grpc::capture::driver_server::Driver for Driver {
...
         &self,
         request: tonic::Request<ApplyRequest>,
     ) -> Result<tonic::Response<ApplyResponse>, tonic::Status> {
-        let message = rpc::unary(&self.entrypoint, "apply-delete", request.into_inner()).await?;
+        // For the JSON protocol, there is no apply-delete operation.
+        // Instead, a deletion is an apply with no bindings.
+        let mut request = request.into_inner();
+        if let Codec::Json = self.codec {
+            request.capture.as_mut().unwrap().bindings.clear();
+        }
```
mdibaiee, line 75: Is this a change we want to extend to non-JSON protocol as well? If not, I would prefer if we don't diverge the JSON protocol from the protobuf protocol 🤔

jgraettinger: Yes, the proposal is that we _remove_ deletion as an RPC of the protobuf protocol, and instead make deletion an application with empty bindings.

This is addressing a tension we discussed during the offsite, where connectors are weird because they drop tables / resources if you delete the whole materialization, but not if you delete bindings of the ongoing materialization. The intention is to consolidate into a single thing: from the connector's perspective (*not* the UI or control-plane), the materialization / capture is never "deleted" but it will be asked to apply updates to the bindings over time, and those bindings could be empty.

It's safer to do it first with the JSON protocol, since that's entirely new, and we can update protobuf connectors over time. And, if the JSON protocol is working well for us, I'd actually propose we make it a goal to deprecate and remove the use of protobuf altogether in our connectors (in which case this migration is defacto accomplished), but let's take it one step at a time.

mdibaiee (approved):

mdibaiee:
I just noticed this breaks `flowctl-go` commands such as `flowctl-go discover` because of the renaming of fields (e.g. `endpointSpecSchema` to `configSchema`), I will send a fix for that

jgraettinger:
You sure? I don't think that's the case because flowctl-go only sees the translated protos, and then maps again to json

mdibaiee:
@jgraettinger yeah that was a mistake on my side, `flowctl-go` is working fine

PR #896, opened 2023-01-26
misc/flow.schema.json: use \w instead of \p{Letter}\p{Number}

mdibaiee:
**Description:**

- JSONSchema's specification suggests that ECMA 262, or preferrably a subset of it that is more widely available is used for regex patterns: https://json-schema.org/understanding-json-schema/reference/regular_expressions.html
- My understanding is that `\p{Letter}\p{Number}` would support non-Latin letters and numbers as well, but apparently not all editors support `\p{Letter}\p{Number}`, so I would say we would rather support all editors with Latin names (the common case) rather than breaking some editors to support non-latin characters.

**Workflow steps:**

- Configure your IDE / Editor to use this schema for files in `examples` directory.

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/896)

### DIFF
--- a/flow.schema.json
+++ b/flow.schema.json
@@ -1,5 +1,5 @@
 {
-  "$schema": "https://json-schema.org/draft/2019-09/schema",
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
   "title": "Catalog",
   "description": "Each catalog source defines a portion of a Flow Catalog, by defining collections, derivations, tests, and materializations of the Catalog. Catalog sources may reference and import other sources, in order to collections and other entities that source defines.",
   "type": "object",
@@ -35,7 +35,7 @@
       ],
       "type": "object",
       "patternProperties": {
-        "^[\\p{Letter}\\p{Number}\\-_\\.]+(/[\\p{Letter}\\p{Number}\\-_\\.]+)*$": {
+        "^[a-zA-Z0-9\\-_\\.]+(/[a-zA-Z0-9\\-_\\.]+)*$": {
           "$ref": "#/definitions/CaptureDef"
         }
       },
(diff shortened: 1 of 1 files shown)


psFried:
This is running up against the edges of my understanding of regex character classes, so I'd like to double check that there isn't some subtle problem hiding in here. The goal of this is to allow our catalog JSON schema to be usable by more editors by making the regexes compliant with the ECMA standard as specified by the JSON schema spec. But these regexes are _also_ used by our Rust code internally, which uses the [perl spec for the `\w` character class](https://docs.rs/regex/latest/regex/#perl-character-classes-unicode-friendly). I think the perl `\w` might actually be more permissive than the ECMA `\w`, though I'm not certain whether that's significant for our purposes. For example `¯\_(ツ)_/¯` is accepted by `\w+` for Perl, but not for ECMA. And the Rust regex crate would reject shruggie with our previous pattern, but accepts it with these new ones.

There's also a regex validation as part of the postgres column constraint for `catalog_name`. Ideally, this validation would be consistent across our Rust code, Postgres, and the ECMA-spec regex rules that are used by editors, though I suspect there may always be some edge cases where they might differ. Thinking out loud, this might all just be OK, if the regex in the DB is more restrictive than the one used by our validation crate. But it'd be worth trying to verify whether that's the case.

mdibaiee:
@psFried Okay, I didn't know we use this in flow internally as well, I will take a deeper look to see what can we do here 👍🏽

mdibaiee:
A dirty alternative would be we generate a separate jsonschema for IDEs specifically, where we replace `\p{Letter}\p{Number}` with `a-zA-Z0-9`, but we keep our original schema unicode-friendly.

jgraettinger:
We should make an issue with JetBrains and do nothing.

mdibaiee:
Filed a bug to JetBrains: https://youtrack.jetbrains.com/issue/IDEA-311780/JSONSchema-validator-does-not-support-pLetter-or-pNumber-in-pattern-regex

PR #897, opened 2023-01-26
docs: more flowctl guides

oliviamiannone:
**Description:**

This PR adds a new flowctl-specific node to the Guides section of the docs and populates it with:
* The existing derivation guide (moved)
* Three new guides suggested [in this discussion](https://github.com/estuary/flow/pull/859#issuecomment-1369405044)

plus relevant links and conceptual changes

Also adds a [redirect plugin](https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-client-redirects) to accommodate the moved guide and all future docs pages that move. Note that it only works in production so I haven't been able to verify yet that it works. 

**Notes for reviewers:**

* For all references to authentication, I'm pointing to the `auth token --token` method still, because `auth login` doesn't work on mac... at least on mine.
* Previously in the flowctl docs (mainly the concepts page), I'd been referring to what we internally call a "draft" as a "synced draft" and files pulled locally as a "local draft." This time around I found that to be pretty clunky so I changed it. Now drafts are just drafts, and local source files are just local source files.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/897)


site/docs/guides/flowctl/edit-specification-locally.md
```diff
@@ -0,0 +1,122 @@
...
+
+   * Pull one or more specifications by name, for example: `flowctl catalog pull-specs --name myOrg/marketing/emailList`
+
+   * Pull a group of specifications by prefix or type filter, for example: `flowctl catalog pull-specs --prefix myOrg/marketing --collections`
+
+  The source files are written to your current working directory.
+
+4. Browse the source files.
+
+  The directory structure will look slightly different depending on which specifications you pull.
```
psFried, line 86:
```suggestion
  By default, `flowctl` will put the specs into subdirectories organized by the names of the entities.
```

This wording seems a little too vague to me.

site/docs/guides/flowctl/edit-specification-locally.md
```diff
@@ -0,0 +1,122 @@
...
+
+2. Make changes. For guidance on how to construct Flow specifications, see the documentation for the task type:
+
+   * [Captures](../../concepts/captures.md#specification)
+   * [Collections](../../concepts/collections.md#specification)
+   * [Materializations](../../concepts/materialization.md#specification)
+   * [Derivations](../../concepts/derivations.md#specification)
+   * [Tests](../../concepts/tests.md)
+
+3. When you're done, run a generic test to make sure all the specifications are formatted correctly:
```
psFried, line 106: nit: the `test` command is checking a lot more than just the formatting. And IDK if "generic" is pulling much weight:

```suggestion
3. As you work, you can run `flowctl catalog test --source flow.yaml` to test your changes.
```

site/docs/guides/flowctl/troubleshoot-task.md
```diff
@@ -0,0 +1,61 @@
...
+
+2. Identify the name of the failing task in the web app; for example `myOrg/marketing/leads`.
+Use the tables on the Captures or Materializations pages of the web app to do so.
+
+3. Run `flowctl logs --task <task-name>`. You have several options to get more specific. For example:
+
+   * `flowctl logs --task myOrg/marketing/leads --follow` — If the task hasn't failed, continuously print logs as they're generated.
+
+   * `flowctl logs --task myOrg/marketing/leads --since 1h` — Print logs from approximately the last hour.
+   The actual output window is approximate and may somewhat exceed this time boundary.
```
psFried, line 33: This may output logs from significantly _before_ the requested start time, but the requested start time will always be covered by the logs that were output. It kinda feels like we should try to clarify that :shrug:

site/docs/guides/flowctl/create-derivation.md
```diff
@@ -31,17 +31,13 @@ If you need help, see the [guide to create a Data Flow](./create-dataflow.md).
...
-   ```
-
-   A browser window opens to the CLI-API tab of the Flow web app.
+   1. Go to the [CLI-API tab of the web app](https://dashboard.estuary.dev/admin/api) and copy your access token.
 
-2. Copy the access token and paste it in the Auth Token prompt in your terminal. Press Enter.
+   2. Run `flowctl auth token --token <paste-token-here>`
 
-3. Begin by creating a fresh draft. This is where you'll add the specification files you need from the catalog.
+2. Begin by creating a fresh draft. This is where you'll add the specification files you need from the catalog.
```
psFried, line 40: I think these steps should change for the latest flowctl. There's no longer a need to use a draft in order to pull specs, and there's also no longer a need to pull the source collection specs in the first place. Pretty much all of the steps between here and step 3 of `Add a derivation` are no longer necessary. And I think the stuff about finding the specific file to put your derivation into should perhaps be replaced with a slightly more generic instruction of how to name the file based on the desired name of your derivation, since there likely won't be existing specs in their directory. Totally fine to do this in a subsequent PR, though.

psFried (approved):
This is a nice improvement. I left some comments inline, but should be fine to merge once those are addressed. The "create a derivation" guide I think still needs a bit of an overhaul, though I don't think it needs to hold up merging the rest of this.

site/docs/guides/flowctl/edit-draft-from-webapp.md
```diff
@@ -0,0 +1,93 @@
...
+
+  Each draft has an **ID**, and most have a name in the **Details** column. Note the **# of Specs** column.
+  For drafts created in the web app, materialization drafts will always contain one specification.
+  A number higher than 1 indicates a capture with its associated collections.
+
+4. Copy the draft ID.
+
+5. Select the draft: `flowctl draft select --id <paste-id-here>`.
+
+6. Pull the draft source files to your working directory: `flowctl draft develop`.
```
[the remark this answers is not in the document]
psFried: I think I spoke too soon here. Upon a closer look, it doesn't seem like the benefit justifies the work at this point, so I'm holding off on any changes there.

site/docs/guides/flowctl/create-derivation.md
oliviamiannone: Noted - I wasn't looking at this page too hard here but I should do so soon, and will make these changes at that point.

Issue #891, opened 2023-01-21
catalog-stats architecture redux


jgraettinger:
## Background

Today, for every tenant in the platform, we create collections `ops/$tenant/stats` and `ops/$tenant/logs`, a derivation `ops/$tenant/catalog-stats` which rolls up metrics drawn from stats and logs, and a materialization `ops/$tenant/catalog-stats-view` which pushes rolled-up stats into a database.

We've realized that there's a data-race which is difficult to solve under this architecture: if TenantA shares a collection `TenantA/Shared` read by `TenantB/Task` and `TenantC/Task`, we produce stat rollups of `TenantA/Shared` which include bytes read from _each_ of the `ops/TenantA/catalog-stats` and `ops/TenantB/catalog-stats` derivations, and their separate materializations go on to race updates of this single rows in the common materialized table.

To fix this, we **MUST** have a single derivation and materialization there-of, which accumulates all stat updates across the platform. This is the only way to guarantee that all stat updates are applied exactly-once to the materialized reporting table.

As a secondary rationale, we've observed that many people register and create tenants in order to tour the product and UI, without immediately setting up tasks and data-flows. They may come back later, but it's likely we'll have a long tail of created tenants that have no tasks or data-flows. It would be nice if we could avoid the cost of a dedicated derivation and materialization for every empty tenant.

We have considered (and rejected) a short-term method by which we could achieve this, which would involve a small amount of runtime support to enable a derivation to dynamically discover and process all per-tenant stat and log journals.

But, having a single derivation & materialization introduces scaling challenges which were less of an issue under the derivation-and-materialization-per-tenant model:

* This single derivation must accumulate statistics across every single `ops/$tenant/stats` and `ops/$tenant/logs` journal in the platform.
* Those collections have potentially _many_ logical partitions, one for each task of the tenant.
* Many tenants will have _no_ logical partitions, most will have just a few, and some will have _lots_.
* Having a single derivation read from an unbounded number of journals is infeasible:
  * Every shard of the derivation must book-keep for all read journals. This book-keeping overhead alone is problematic.
  * Having lots of concurrent reads of cloud-storage fragments can result in slow reads and TCP resets for the individual fragments from the cloud provider.
  * We can't even guarantee that the listing of all such journals will reliably fit in memory!
 * Finally, we plan to remove the current property that a flow reactor is aware of all journals in the system once we introduce federated data-planes.
[...]

PR #900, opened 2023-01-27
ops-catalog: multi-staged stats rollups

williamhbaker:
**Description:**

This implements a multi-staged stats rollup architecture as described in https://github.com/estuary/flow/issues/891.

There are some interesting bits of json & string manipulation required for building out the transforms for the level 1 and 2 derivations. There's also some "interesting" string manipulation for building of the publish lambdas of the corresponding typescript modules. Building up the typescript lambdas is kind of ugly, but is the best I could come up with since we currently require a named lambda function for every transform.

This PR includes scripts for migrating production over to this new architecture. Upon completion of the migration, all existing per-tenant stats materializations & derivations will be deleted, and there will be a single level 1 derivation, level 2 derivation, and `stats-view` materialization.

The migration process is as follows:
- Deploy the agent with its small code change. This is backward-compatible.
- Run `supabase/pending/tiered_stats_1.sh` migration. Broadly, this will apply the database changes included in this PR. It will also create a draft that will delete all existing per-tenant stats materializations. A tricky part of this is preventing these deletions from also deleting the `catalog_stats` table. This is accomplished by changing the owner of the table to `postgres` (from `stats_loader`) so that the materialization deletions are not able to delete the table.
- Once the publication from `tiered_stats_1` completes, run `supabase/pending/tiered_stats_2.sql`. This simply changes the owner of `catalog_stats` back to `stats_loader`.
- Run `ops-catalog/generate-migration.sh`. This will create a draft with specs for the level 1 & 2 derivation, `stats-view` materialization, and will re-publish all the individual tenant ops catalogs. The result of this per-tenant will be deleting the per-tenant stats derivation tasks. There will only be 1 level 1 derivation at this point.

I've tested this migration locally by setting up a couple of tenants on the old system, then switching to the new system and running the steps described above. After that, stats continue to work for these existing tenants' tasks. I am then able to create new tenants on the existing `l1_stat_rollup` and have their stats work, and also `alter` the default `l1_stat_rollup` to create new tenants with new L1 derivations and observe that their stats work, as well as all other stats continuing to work.

There are some simple sql tests covering the new postgres functions. I will add more comprehensive tests for the l1 & l2 derivation generation functions as well.

**Workflow steps:**

Run flow locally or in prod and see that stats work.

N/A

**Notes for reviewers:**

See a couple of inline comments for specific points I'd like input on.


ops-catalog/generate-migration.sh
```diff
@@ -19,36 +19,118 @@ if [[ -z "$ENVIRONMENT" ]]; then
...
+	-- Create a draft of ops changes.
+	insert into drafts (id, user_id, detail) values
+	(new_draft_id, ops_user_id, 're-publishing ops catalog');
+
+	-- Queue a publication of the draft.
+	insert into publications (id, user_id, draft_id) values
+	(publication_id, ops_user_id, new_draft_id);
 
-	-- Start a publication of the updated template for each existing tenant.
+	-- Handle the tenant-specific specs for each tenant.
```
williamhbaker, line 78: This is all going into a single draft. There will be specs for every single tenant. I'm not sure if there's a limit to how many specs can be in a single draft. It was a little easier to do it this way, but if we are worried about a single draft becoming too large I could create a separate draft for each tenant's specific specs.

ops-catalog/generate-migration.sh
```diff
@@ -19,36 +19,118 @@ if [[ -z "$ENVIRONMENT" ]]; then
...
+		select new_draft_id, "key", 'capture'::catalog_spec_type, "value"
+		from jsonb_each(jsonb_extract_path(tenant_spec, 'captures'))
+		union all
+		select new_draft_id, "key", 'collection'::catalog_spec_type, "value"
+		from jsonb_each(jsonb_extract_path(tenant_spec, 'collections'))
+		union all
+		select new_draft_id, "key", 'materialization'::catalog_spec_type, "value"
+		from jsonb_each(jsonb_extract_path(tenant_spec, 'materializations'))
+		on conflict (draft_id, catalog_name)
+		do update set spec_type = excluded.spec_type, spec = excluded.spec;
```
williamhbaker, line 104: Here and in the `create_ops_publication` postgres functions I have opted not to include the tests for each tenant. This is partially an attempt to simplify things, and also because I was concerned about a new tenant publication needing to run the tests of every single other tenant attached to that level 1 derivation. I could be convinced to include them though.

jgraettinger (commented):
Thank you! This looks like it was a lot of work to put together. Much appreciated.

There is a lot going on there, though, and I'm concerned about how we maintain it. I'm wanting to reach for a solution with fewer moving parts.

More narrowly and tactically, I think we could elide the L2 derivation and materialization completely. Those can be manually maintained specs which are versioned and applied as regular, non-templated Flow specs. Introducing a new L1 derivation requires a manual action anyway, and at that time we could add the new L1 as a transform of our versioned L2 spec. The materialization itself never needs to change, once applied. We can also re-cast the test as being a non-templated, regular spec which sources only from the fixed `ops/estuary/logs` & stats collections, rather than having it run for _every_ tenant. Finally, it's not a "must" that the L2 spec, test, or materialization be under the `ops/` tenant if that makes the solution easier to operate (they could be under `estuary/` tenant instead, for example).

More broadly, though, we're doing a weird straddle where we're attempting to have the `agent` take some actions (e.g. provisioning ops collections) so that they happen quickly, **BUT** we're constrained by the concurrent and raced nature of the agent. It's fundamentally and intentionally designed for narrowly-scoped actions that are highly parallelize-able. But because of this, we've already developed a manual, global, and slow `generate-migration.sh` script that updates ops tasks of all tenants.

I'm wondering what happens if we lean into having a control-plane component that's distinct from the `agent` that encapsulates global, serialized, and incremental actions that need to be taken to converge _all_ specs towards a desired state. In essence: production-izing `generate-migration.sh` and making it more capable. This is increasingly feeling like an architecture gap.

For example, we could write Rust business logic that:
  * Reads the set of tenants and L1 assignments from the DB.
  * Uses some kind of fast Rust templating solution to render out what a complete set of global specs _ought_ to be.
  * Efficiently compares each generated spec with its live value in the control-plane DB.
  * For each generated spec that differs, it creates narrowly-focused drafts and publications

I think this is pretty tractable, with lots of headroom to scale, so long as the template generation and individual comparisons are fast (for example, just compare an MD5 digest). If nothing's changing, I expect this could be optimized to take ~1sec even with tens of thousands of tenants.

It could also take over unified responsibility for both generating _and migrating_ `ops/TENANT/stats` and logs collections. Its generated templates would include these collections, and the agent would be simplified to only create a `tenants` row when provisioning.

(We might want to have it evaluate specs in a couple "levels", where it first converges all `ops/$tenant/stats` collections before attempting to converge L1 derivation specs which could depend on them).

Next, we need some kind of "watch" loop to regularly evaluate this routine. We could start with a very frequent cron-tab. Or, we could package as a long-lived daemon which implements watches to run whenever a base table changes (i.e. `tenants`).

Again, a key distinction with the `agent` is there would only be _one_ of these running, and it doesn't need to worry about raced changes.

Thoughts?

williamhbaker:
I think your concerns about maintaining this long-term are valid. Moving the L2 derivation & materialization out of the per-tenant automatic part would help a little, but fundamentally the worst parts would still be there in creating the L1 derivation specs (etc) in the `plpgsql` functions.

A new control plane component is an interesting idea. It would certainly be better suited for rendering these templates in a comprehensible, efficient, and testable way. I am a bit hesitant to get fully onboard with the idea of a new component though, because that will have long-term maintenance costs as well & the benefit is solely for our stats reporting. This same argument applies to what I've proposed here in `plpgsql` of course. Neither one feels great to me in that respect.

I believe @psFried had some additional thoughts last week about how we might solve this with enhancements to derivations in general, so that would probably be good to talk about before proceeding.

Issue #904, opened 2023-01-31
Flowctl `catalog delete`  subcommand


psFried:
Deleting specifications currently requires a lot of steps. It would be nice if we had a `flowctl catalog delete` subcommand that could handle all those steps for you. `flowctl catalog delete` should:

- Accept more or less the same arguments as `flowctl catalog list`, such as `--prefix` and `--captures`.
- It should print the selected items and ask for confirmation before proceeding with the deletion.
- If the user wishes to proceed, then it should create a new draft, add `null` sentinels for each selected item, and publish it.

PR #906, opened 2023-01-31
Phil/flowctl moar ux

### LINKED ISSUE #865: `flowctl draft author` should prune specs that aren't in the local sources
Say you run the following steps:

- `flowctl catalog draft --name acmeCo/foo`
- `flowctl draft develop`
- Rename `acmeCo/foo` => `acmeCo/bar` in the local spec files
- `flowctl draft author`
- `flowctl draft publish` :boom: your draft includes both `acmeCo/foo` and `acmeCo/bar`

There's not an obvious (to me) _best_ way to handle this, but any of these options seem significantly better to me:

- `draft author` clears all specs from the draft before adding new ones. This way your existing `acmeCo/foo` remains untouched.
- `draft author` changes all specs in the draft to `null` (triggering the deletion of the live specs upon publication) before adding new ones. This way your existing `acmeCo/foo` will be deleted by the publication.
- `draft author` accepts an argument similar to `--existing abort|keep|overwrite|merge-spec` from `catalog pull-specs`, which explicitly says what to do with an existing draft specs.

Given that there's now, with the introduction of `catalog pull-specs|test|publish`, relatively few reasons to use `draft` at all, my inclination is to have `draft author` simply clear all specs from the draft before adding new ones. It's a super easy change to make, and addresses the biggest foot gun. It would also be easy to add a flag to allow opting in to the current behavior.

psFried:
**Description:**

Resolves #865 and #904

This rolls up a few different minor fixes and improvements to `flowctl`.
- Fixed a bug in confirmation in `catalog publish` that caused confirmation message to sometimes not appear.
- `draft author` now prunes any extra specs automatically (#865)
- `draft develop` no longer adds a bunch of extra imports. It now lays everything out in the same way as `catalog pull-specs`
- Introduced `flowctl catalog delete`, which works in a single-step (with confirmation), so you no longer need to manually deal with drafts during deletion (#904).
- `flowctl catalog list` was including deleted specs by default. This was updated so that deleted specs are never included unless you explicitly ask for them by providing the new `--deleted` flag.

**Workflow steps:**

Deleting specs:
Before:

- `flowctl draft crate`
- `flowctl catalog draft --name acmeCo/foo --delete`
- `flowctl catalog draft --name acmeCo/bar --delete`
- `flowctl draft publish`

After: `flowctl catalog delete --prefix acmeCo/`

These changes seem to mostly avoid impacting existing docs. Delete could maybe use a docs page at some point, but mostly just to help set expectations around materialized tables being deleted.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/906)

### DIFF
--- a/crates/flowctl/src/catalog/delete.rs
+++ b/crates/flowctl/src/catalog/delete.rs
@@ -1,0 +1,132 @@
+use crate::{api_exec, catalog, draft, CliContext};
+use anyhow::Context;
+use serde::Serialize;
+
+#[derive(Debug, clap::Args)]
+pub struct Delete {
+    #[clap(flatten)]
+    pub name_selector: catalog::NameSelector,
+    #[clap(flatten)]
+    pub type_selector: catalog::SpecTypeSelector,
+    /// Proceed with deletion without prompting for confirmation.
+    ///
+    /// Normally, delete will stop and ask for confirmation before it proceeds. This flag disables
+    /// that confirmation. This is sometimes required in order to run flowctl non-interactively,
+    /// such as in a shell script.
+    pub dangerous_auto_approve: bool,
+}
+
+#[derive(Serialize, Debug)]
+struct DraftSpec {
+    draft_id: String,
+    catalog_name: String,
+    expect_pub_id: String,
+    spec_type: serde_json::Value, // always null, since we're deleting
+    spec: serde_json::Value,      // always null, since we're deleting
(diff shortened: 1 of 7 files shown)


jshearer (approved):
LGTM, I can't find anything wrong or confusing. I love how much of `draft/develop.rs` we get to delete here!

PR #907, opened 2023-02-01
doc: new reduction strategy `lastWriteWinsUnlessNull`

mdibaiee:
**Description:**

- Introduce a new reduction strategy `lastWriteWinsUnlessNull` which allows incremental updates to a document with a key.

As an example, when a document needs to be marked deleted, instead of having to have all the fields of the document, one can use the document below using `lastWriteWinsUnlessNull` to update `IsDeleted` field without overriding all the other fields:
```
{"Id": "x", "IsDeleted": true}
```

**Workflow steps:**

- Set up a new collection and mark properties with `reduction: { strategy: "lastWriteWinsUnlessNull" }`
- Emit some documents
- Emit another document with an ID that matches existing documents, but only specifying the fields you want to update on that document

This page would need to be updated to reflect this new strategy: https://docs.estuary.dev/reference/reduction-strategies/

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/907)


psFried:
I think I can see how this would work for individual fields, but I'm having trouble understanding how this would be used in an end-to-end use case. Given the input documents:

```
{"id":"x", "foo": 1, "bar": true}
{"id":"x", "isDeleted": true}
```

and a schema of:

```yaml
type: object
properties:
  id: {type: string}
  foo: { type: string, reduce: { strategy: lastWriteWinsUnlessNull } }
  bar: { type: boolean, reduce: { strategy: lastWriteWinsUnlessNull } }
  isDeleted: {type: boolean}
```

is the goal to have a reduction of those produce `{"id":"x", "isDeleted": true, "foo": 1, "bar": true }`? 

If so, then I'd think the following schema would produce an equivalent result:

```yaml
type: object
reduce: { strategy: merge }
properties:
  id: {type: string}
  foo: { type: string }
  bar: { type: boolean }
  isDeleted: {type: boolean}
required: [id]
```

I feel like I might be missing something with the end-to-end use case, though, so please explain more about that if I'm missing the mark here.

mdibaiee:
@psFried ahh! that's what I've been looking for, and it works, so I don't need this at the moment 👍🏽

PR #908, opened 2023-02-01
docs: Add AlloyDB connectors

oliviamiannone:
**Description:**

Adds docs pages for source-alloydb and materialize-alloydb. Modified from Postgres docs.

We'll need [shortlinks](https://app.short.io/) to these pages to add to the connector for use in the UI.

**Notes for reviewers:**

In here I instruct the user to use SSH tunneling to connect to this one because, from what I can tell, there's no way to connect to AlloyDB directly equivalent to whitelisting an IP. I tried creating a [firewall rule in the VPC](https://cloud.google.com/vpc/docs/using-firewalls) to no avail. It's possible that I'm wrong so please let me know if there's a way to do this.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/908)

### DIFF
--- a/site/docs/reference/Connectors/capture-connectors/PostgreSQL.md
+++ b/site/docs/reference/Connectors/capture-connectors/PostgreSQL.md
@@ -267,7 +267,7 @@ You can use this connector for PostgreSQL instances on Google Cloud SQL using th
         for additional details and a sample.
    :::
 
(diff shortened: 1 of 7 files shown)


williamhbaker (approved):
LGTM, although I'm not sure about the IP whitelisting either. @dyaffe actually tested these out so maybe he knows if its possible to use anything other than SSH tunneling.

dyaffe:
@oliviamiannone and I spoke and it wasn't immediately obvious how to get that working so we're going to ignore it for now, @williamhbaker.

PR #910, opened 2023-02-02
flowctl: catalog delete must require either `name` or `prefix`

mdibaiee:
**Description:**

- clap is smart enough to understand that because `name` and `prefix` conflict with each other, one being marked as `required` is sufficient. 

**Workflow steps:**

- Try `flowctl catalog delete` with no `--name` or `--prefix` arguments

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/910)

### DIFF
--- a/crates/flowctl/src/catalog/mod.rs
+++ b/crates/flowctl/src/catalog/mod.rs
@@ -80,7 +80,7 @@ pub enum Command {
 #[derive(Default, Debug, Clone, clap::Args)]
 pub struct NameSelector {
     /// Select a spec by name. May be provided multiple times.
-    #[clap(long)]
+    #[clap(long, required = true)]
     pub name: Vec<String>,
     /// Select catalog items under the given prefix
     ///


crates/flowctl/src/catalog/mod.rs
```diff
@@ -80,7 +80,7 @@ pub enum Command {
 #[derive(Default, Debug, Clone, clap::Args)]
 pub struct NameSelector {
     /// Select a spec by name. May be provided multiple times.
-    #[clap(long)]
+    #[clap(long, required = true)]
```
psFried, line 83: These arguments are also used by lots of other subcommands, and I don't think we want to require them in all contexts. I'm thinking we should figure out a way for one of these to only be required if the sucommand is `delete`. That _might_ need to be an explicit check inside of `do_delete`, if there isn't something like `required_if_subcommand_eq` in clap. Or I suppose we could add a separate `DeleteNameSelector` that requires on of these args and implements `Into<NameSelector>`.

psFried (changes requested):
Thanks for working on this! I left one comment, which I think needs addressed before we merge, but it should be a pretty straight forward change.

psFried:
Closing in favor #913

PR #920, opened 2023-02-08
auth: introduce refresh-tokens for generating jwt access tokens

mdibaiee:
**Description:**

- Introduce `refresh_tokens` and RPCs for generating access_tokens given a refresh_token and for revoking refresh_tokens.
- Question I have is: currently I'm using our `id_generator` to generate a refresh_token (which is then hashed with a salt): do we want to use a different random entity instead (perhaps something a bit longer), or do we want to apply additional transformation on this uuid so it's not so obviously a uuid? Perhaps a base64 encoding, or even one more level of hashing (e.g. hash once, present to user, hash the result another time with a salt and persist this one). -- It is also confusing since the token ID is also uuid.
- Next step we could have a cron job that cleans up expired refresh_tokens from the database
- This current implementation assumes that the first refresh_token and subsequent ones generated to replace it all have the same validity period. There was a discussion in Slack that the first one might be short-lived while the rest of them are longer-lived, I will work on that too.

**Workflow steps:**

- Check list of existing refresh_tokens:
```
curl -H "apikey: $ANON_KEY" -H "Authorization: Bearer $KEY" 'http://localhost:5431/rest/v1/refresh_tokens?select=id,user_id,valid_for,multi_use'

[]
```

- Create a new refresh_token:
```
curl -XPOST -H "apikey: $ANON_KEY" -H "Authorization: Bearer $KEY" 'http://localhost:5431/rest/v1/rpc/gen_refresh_token' -d '{"multi_use": true, "valid_for": "1 day"}' -H 'Content-Type: application/json'

{"refresh_token":"08:d7:7b:c5:fe:80:38:00"}
```

- Check list of refresh_tokens again:
```
curl -H "apikey: $ANON_KEY" -H "Authorization: Bearer $KEY" 'http://localhost:5431/rest/v1/refresh_tokens?select=id,user_id,valid_for,multi_use'

[{"id":"08:d7:7b:c5:ff:00:3c:00","user_id":"11111111-1111-1111-1111-111111111111","valid_for":"1 day","multi
_use":true}]
```

- Generate an access_token given the refresh_token (note that you don't need a bearer token for this one, you can use anonymous key):
```
curl -XPOST -H "apikey: $ANON_KEY" -H "Authorization: Bearer $ANON_KEY" -H 'Content-Type: application/json' 'http://localhost:5431/rest/v1/rpc/gen_access_token' -d '{"refresh_token": "08:d7:7b:c5:fe:80:38:00"}'

{"access_token":"eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJleHAiIDogMTY3NTk1MzU2NSwgImlhdCIgOiAxNjc1OTQ5OTY1LCAic3ViIiA6ICIxMTExMTExMS0xMTExLTExMTEtMTExMS0xMTExMTExMTExMTEifQ.tX9tn6qg5hs3pdXhdSUNFL0zN72cVOXUVN1-F4OX4Rg","refresh_token":null}
```

To test this new token, you can replace `$KEY` with this new access_token:

```
curl -H "apikey: $ANON_KEY" -H "Authorization: Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJleHAiIDogMTY3NTk1MzU2NSwgImlhdCIgOiAxNjc1OTQ5OTY1LCAic3ViIiA6ICIxMTExMTExMS0xMTExLTExMTEtMTExMS0xMTExMTExMTExMTEifQ.tX9tn6qg5hs3pdXhdSUNFL0zN72cVOXUVN1-F4OX4Rg" 'http://localhost:5431/rest/v1/refresh_tokens?select=id,user_id,valid_for,multi_use'

[{"id":"08:d7:7b:c5:ff:00:3c:00","user_id":"11111111-1111-1111-1111-111111111111","valid_for":"1 day","multi_use":true}]
```

- Revoke the refresh_token:

```
curl -XPOST -H "apikey: $ANON_KEY" -H "Authorization: Bearer $KEY" -H 'Content-Type: application/json' 'http://localhost:5431/rest/v1/rpc/revoke_refresh_token' -d '{"token": "08:d7:7b:c5:fe:80:38:00"}'
```

Or:
```
curl -XPOST -H "apikey: $ANON_KEY" -H "Authorization: Bearer $KEY" -H 'Content-Type: application/json' 'http://localhost:5431/rest/v1/rpc/revoke_refresh_token' -d '{"token_id": "08:d7:7b:c5:ff:00:3c:00"}'
```

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/920)


psFried:
Awesome! I've yet to thoroughly read through the code, but want to offer a bit of early feedback before I dig in.

> Question I have is: currently I'm using our id_generator to generate a refresh_token...

I think we really want to use either the `gen_random_bytes` or `gen_random_uuid` functions from `pgcrypto` to generate these tokens, rather than relying on our primary key generation function. The id generation function is designed to limit collisions, but was _not_ designed to produce unpredictable (in the context of cryptography) output. I think a reasonable sketch would be to use `gen_random_uuid` to generate the token, which is returned to the user directly. Then hash the token (using `pgcrypto`'s `digest` function) and only persist that hashed value to the DB. Curious if that makes sense to @williamhbaker and @jgraettinger ?

> which is then hashed with a salt

This doesn't make sense to me. Normally, you'd add a salt to a _password_ before hashing, to ensure that users having the same password have different hashes, and to mitigate attacks using [rainbow tables](https://en.wikipedia.org/wiki/Rainbow_table). But you need to persist the salt and look up the proper salt for each password, usually based on the username. There's no username provided with the API token, so no good way to lookup the salt. And there's also not really a need to salt the token, as long as it's generated using a cryptographic random function. So I'd think that just taking a digest of the `gen_random_uuid` and storing that directly (without salt) ought to be fine. Once again, though, I'd like to invite others to disagree, because this really isn't my area of expertise.

williamhbaker:
I agree that we should use completely randomly generated values for the actual tokens. An argument could be made that salting a totally random token value is not necessary. However it should be easy enough to do that we might as well. It makes answering questions like "Are your user secrets (passwords, API tokens, etc) stored hashed with a salt?" a lot easier to answer. As far as I understand it the salt value is stored as part of the hashed value as well so we shouldn't need to look it up anywhere else, and pgcrypt just handles it.

Also we should probably use `bf` as the algorithm instead of `md5`. Again perhaps are argument could be made that a work-based algorithm isn't really necessary for an entirely random token string, but since these tokens won't be used on a hot path it makes the story a lot easier to tell rather than having to explain why it is ok that we are using an algorithm that isn't cryptographically secure to store secrets.

supabase/migrations/18_refresh_token.sql
```diff
@@ -0,0 +1,119 @@
+-- Refresh tokens used to generate access tokens
+create table refresh_tokens (
+  like internal._model including all,
+  user_id    uuid references auth.users(id) not null,
+  multi_use  boolean default false,
+  valid_for  interval not null,
+  hash       text not null
```
jgraettinger, line 7: add a 'uses' integer that counts up each time the token is used?

jgraettinger (commented):
Thanks! Some comments below.

jgraettinger:
> which is then hashed with a salt
> This doesn't make sense to me. Normally, you'd add a salt to a password before hashing, to ensure that users having the same password have different hashes, and to mitigate attacks using [rainbow tables](https://en.wikipedia.org/wiki/Rainbow_table).

it works fine in this case, because the hash returned by pgcrypto's `crypt` embeds the salt as a leading component, and you feed the hash back into `crypt` with the secret and it re-uses the salt.

mdibaiee:
Okay, so I've updated the code, now each refresh_token is used with a pair of `id` and `secret`, and the usage looks like this:

Create refresh token:
```
curl -XPOST -H "apikey: $ANON_KEY" -H "Authorization: Bearer $KEY" 'http://localhost:5431/rest/v1/rpc/create_refresh_token' -d '{"multi_use": true, "valid_for": "1 day", "detail": "my refresh token"}' -H 'Content-Type: application/json'

{"secret" : "5dda3d2b-974f-48b6-9a6a-5f98f75cfc46", "id" : "08:d9:c2:34:e2:80:30:00"}
```

List refresh tokens:
```
curl -H "apikey: $ANON_KEY" -H "Authorization: Bearer $KEY" 'http://localhost:5431/rest/v1/refresh_tokens?select=id,user_id,valid_for,multi_use,uses,created_at,detail,updated_at'

[{"id":"08:d9:c2:34:e2:80:30:00","user_id":"11111111-1111-1111-1111-111111111111","valid_for":"1 day","multi_use":true,"uses":0,"created_at":"2023-02-10T10:50:43.521867+00:00","detail":"my refresh token","updated_at":"2023-02-10T10:50:43.521867+00:00"}]
```

Update refresh token:
```
curl -XPATCH -H "apikey: $ANON_KEY" -H "Authorization: Bearer $KEY" 'http://localhost:5431/rest/v1/refresh_tokens?id=eq.08:d9:bb:ce:7b:00:30:00' -d '{"detail": "testing!"}' -H 'content-type: application/json'

curl -H "apikey: $ANON_KEY" -H "Authorization: Bearer $KEY" 'http://localhost:5431/rest/v1/refresh_tokens?select=id,user_id,valid_for,multi_use,uses,created_at,detail,updated_at'

[{"id":"08:d9:c2:34:e2:80:30:00","user_id":"11111111-1111-1111-1111-111111111111","valid_for":"1 day","multi_use":true,"uses":0,"created_at":"2023-02-10T10:50:43.521867+00:00","detail":"testing!","updated_at":"2023-02-10T10:50:43.521867+00:00"}]
```

Generate access token (this is a multi-use token, so no new `secret` is returned):
```
curl -XPOST -H "apikey: $ANON_KEY" -H "Authorization: Bearer $ANON_KEY" -H 'Content-Type: application/json' 'http://localhost:5431/rest/v1/rpc/generate_access_token' -d '{"id": "08:d9:c2:34:e2:80:30:00", "secret": "5dda3d2b-974f-48b6-9a6a-5f98f75cfc46"}'

{"access_token" : "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJleHAiIDogMTY3NjAyOTkzMCwgImlhdCIgOiAxNjc2MDI2MzMw
LCAic3ViIiA6ICIxMTExMTExMS0xMTExLTExMTEtMTExMS0xMTExMTExMTExMTEiLCAicm9sZSIgOiAiYXV0aGVudGljYXRlZCJ9.WGOVrjjnzxtFFBCpeCIAcAR_ob9ZX3su-cs0thQ54xk", "refresh_token" : {"id" : "08:d9:c2:34:e2:80:30:00", "secret" : null}
```

Use new access token and revoke the refresh_token:
```
curl -XDELETE -H "apikey: $ANON_KEY" -H "Authorization: Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJleHAiIDogMTY3NjAyOTkzMCwgImlhdCIgOiAxNjc2MDI2MzMwLCAic3ViIiA6ICIxMTExMTExMS0xMTExLTExMTEtMTExMS0xMTExMTExMTExMTEiLCAicm9sZSIgOiAiYXV0aGVudGljYXRlZCJ9.WGOVrjjnzxtFFBCpeCIAcAR_ob9ZX3su-cs0thQ54xk" 'http://localhost:5431/rest/v1/refresh_tokens?id=eq.08:d9:c2:34:e2:80:30:00'

curl -H "apikey: $ANON_KEY" -H "Authorization: Bearer $KEY" 'http://localhost:5431/rest/v1/refresh_tokens?select=id,user_id,valid_for,multi_use,uses,created_at,detail,updated_at'

[]
```

supabase/migrations/18_refresh_token.sql
```diff
@@ -0,0 +1,103 @@
...
+
+  select coalesce(current_setting('app.settings.jwt_secret', true), 'super-secret-jwt-token-with-at-least-32-characters-long') limit 1
+
+$$ language sql stable security definer;
+
+-- Given a refresh_token, generates a new access_token
+-- if the refresh_token is not multi-use, it is deleted and a new
+-- refresh_token is also created. If the refresh_token is multi-use, we reset
+-- its validity period by updating its `updated_at` column
+create function generate_access_token(id flowid, secret text)
```
jgraettinger, line 58: nit: name `id` => `refresh_token_id`

supabase/migrations/18_refresh_token.sql
```diff
@@ -0,0 +1,103 @@
...
+-- local env, taken from https://github.com/supabase/supabase-js/issues/25#issuecomment-1019935888
+create function internal.access_token_jwt_secret()
+returns text as $$
+
+  select coalesce(current_setting('app.settings.jwt_secret', true), 'super-secret-jwt-token-with-at-least-32-characters-long') limit 1
+
+$$ language sql stable security definer;
+
+-- Given a refresh_token, generates a new access_token
+-- if the refresh_token is not multi-use, it is deleted and a new
```
jgraettinger, line 55: nit: comment is out of date

supabase/migrations/18_refresh_token.sql
```diff
@@ -0,0 +1,103 @@
...
+  rt_new_secret text;
+  access_token text;
+begin
+
+  select * into rt from refresh_tokens where
+    refresh_tokens.id = generate_access_token.id and
+    hash = crypt(secret, hash) and
+    (updated_at + valid_for) > now();
+  if not found then
+    raise 'invalid refresh token';
```
jgraettinger, line 71: can we improve this error message? it's pretty easy to imagine users smacking headlong into it. In particular, as a user, being able to distinguish between:
  a) token ID doesn't exist, vs 
  b) I have the wrong secret value, vs
  c) I have the correct secret but it's no longer valid

... would be very helpful.

supabase/migrations/18_refresh_token.sql
```diff
@@ -0,0 +1,103 @@
...
+    'sub', rt.user_id,
+    'role', 'authenticated'
+  ), internal.access_token_jwt_secret()) into access_token
+  limit 1;
+
+  if rt.multi_use = false then
+    rt_new_secret = gen_random_uuid();
+    update refresh_tokens
+      set
+        hash = crypt(rt_new_secret, gen_salt('bf')),
```
jgraettinger, line 86: you must manually update `updated_at` in this `update` statement and the one below.

supabase/migrations/18_refresh_token.sql
```diff
@@ -0,0 +1,103 @@
...
+  else
+    -- re-set the updated_at timer so the token's validity is refreshed
+    update refresh_tokens set uses = (uses + 1) where refresh_tokens.id = rt.id;
+  end if;
+
+  return json_build_object(
+    'access_token', access_token,
+    'refresh_token', json_build_object(
+      'id', rt.id,
+      'secret', rt_new_secret
```
jgraettinger, line 98: what happens if we take the multi-use branch? does this come through as `secret: null` ?

jgraettinger (commented):
Looking a lot better. A few more comments. Please also work up SQL tests for this? Thanks!

supabase/migrations/18_refresh_tokens.sql
```diff
@@ -0,0 +1,120 @@
...
+create table refresh_tokens (
+  like internal._model including all,
+  user_id    uuid references auth.users(id) not null,
+  multi_use  boolean default false,
+  valid_for  interval not null,
+  uses       int default 0,
+  hash       text not null
+);
+
+create policy "Users can access only their own refreshed_tokens"
```
jgraettinger, line 11: nit: table name is refresh_tokens. Perhaps just "their own refresh tokens".

supabase/migrations/18_refresh_tokens.sql
```diff
@@ -0,0 +1,120 @@
...
+$$ language plpgsql volatile security definer;
+
+-- Returns the secret used for signing JWT tokens, with a default value for
+-- local env, taken from https://github.com/supabase/supabase-js/issues/25#issuecomment-1019935888
+create function internal.access_token_jwt_secret()
+returns text as $$
+
+  select coalesce(current_setting('app.settings.jwt_secret', true), 'super-secret-jwt-token-with-at-least-32-characters-long') limit 1
+
+$$ language sql stable security definer;
```
jgraettinger, line 52: nit: this shouldn't be security definer, as that allows an unprivileged caller to run with the privilege of the definer. It doesn't matter in practice, as this is `internal` schema.

supabase/migrations/18_refresh_tokens.sql
```diff
@@ -0,0 +1,120 @@
...
+
+-- Given a refresh_token, generates a new access_token
```
jgraettinger, line 54: nit: prefer to document public API methods via postgres COMMENT like
```
comment on function generate_access_token is 'Given a refresh token, generate ... ';
```

That both documents the function for the reader here, but it also flows through into supabase-generated API documentation and our OpenAPI response.

supabase/migrations/18_refresh_tokens.sql
```diff
@@ -0,0 +1,120 @@
...
+  rt refresh_tokens;
+  rt_new_secret text;
+  access_token text;
+begin
+
+  select * into rt from refresh_tokens where
+    refresh_tokens.id = refresh_token_id;
+
+  if not found then
+    raise 'could not find refresh_token with the given `refresh_token_id`';
```
jgraettinger, line 69: supersupernit: include appropriate errcode's like:

``` raise 'could not find ... ' using errcode = 'no_data_found';```

[See the appendix of them here.](https://www.postgresql.org/docs/current/errcodes-appendix.html)

These actually flow through PostgREST and are surfaced to users in the HTTP REST error response body.

File: supabase/tests/refresh_tokens.test.sql [hunk 11]
```diff
@@ -0,0 +1,108 @@
+create function tests.startup_refresh_tokens_as_alice()
```
jgraettinger, line 1: this will clober the existing definition in auth.test.sql, and can be removed.

supabase/tests/refresh_tokens.test.sql
```diff
@@ -0,0 +1,108 @@
...
+  select generate_access_token((rt_response->>'id')::flowid, rt_response->>'secret') into response;
+
+  return query select ok(response::jsonb ? 'access_token', 'generate_access_token response has access_token');
+  return query select ok(not (response::jsonb ? 'refresh_token'), 'generate_access_token response does not have refresh_token (multi-use)');
+
+  select * into new_rt from refresh_tokens;
+  return query select is(new_rt.uses, 1, 'refresh_tokens uses bumped');
+  return query select is(new_rt.hash, rt.hash, 'refresh_token hash unchanged (multi-use)');
+
+  -- Unfortunately we can't run this test because `now()` refers to the
```
jgraettinger, line 57: you can use clock_timestamp() instead of now() to get the actual time, even inside of a transaction. IMO it would be reasonable to use clock_timestamp() in the actual function bodies in order to make the tests work (we don't care that it's wall-time vs transaction-start-time in practice).

jgraettinger (approved):
Thanks! Appreciate the back and forth but this looks great to me now. A few comments / nits on ways to further improve things.

PR #922, opened 2023-02-08
docs: salesforce-next

oliviamiannone:
**Description:**

Adds simple reference docs for real-time salesforce connector -- in-depth guide on using the two salesforce connectors in parallel will come later.

**Notes for reviewers:**

A lot of this is copied/slightly modified from the other connector docs. The one thing I really wasn't sure about was the supported salesforce objects.... I couldn't figure out in the code if there was a specific list for this connector, so I used the list from the other one for now.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/922)


mdibaiee (approved):
I think this is a good starting point, we can add more info as we develop the connector and find more use cases, etc.

PR #925, opened 2023-02-09
add support for `sql` lambdas

jgraettinger:
**Description:**

Add a new `sql` variant for V1 derivation `update` and `publish` lambdas.

This lambda variant allows users to write natural `select foo, bar from source` queries to map collection documents, and also join them with registers. It's generally as-capable as TypeScript transformations are, although with fewer guard-rails around type compatibility.

Please see individual commits for more details.

**Workflow steps:**

Start writing `sql` lambdas to implement transforms!

No documentation has been done yet.

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/925)


crates/sqlite-lambda/tests/snapshots/examples__type_conversions.snap
```diff
@@ -0,0 +1,76 @@
+---
+source: crates/sqlite-lambda/tests/examples.rs
+expression: output
+---
+[
+  [
+    {
+      "case": "true",
+      "output": 1
```
williamhbaker, line 9: Does this mean that a sqlite lambda that selects fields out of a document with boolean values will have outputs `0` and `1`? I guess that's how sqlite stores booleans but it probably wouldn't be what a user would expect.

flow.schema.json
```diff
@@ -668,6 +656,9 @@
         "typescript",
         {
           "remote": "http://example/api"
+        },
+        {
+          "remote": "SELECT foo, bar FROM source;"
```
williamhbaker, line 661: Should this key be `sql`?

williamhbaker (approved):
LGTM, with a couple of questions

crates/sqlite-lambda/src/lib.rs
```diff
@@ -0,0 +1,257 @@
...
+        };
+
+        let stmt = db.prepare(query)?;
+
+        // Extract output columns from the prepared statement.
+        let output_columns = if stmt.column_name(0)? == "flow_document" && stmt.column_count() == 1
+        {
+            Vec::new()
+        } else {
+            (0..stmt.column_count())
```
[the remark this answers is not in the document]
jgraettinger: :shrug: it returns &str and we need to avoid having a borrow of `stmt` so we can call `stmt.query(&mut self)`.

Either way it's going to require mapping into strings and re-collecting into a Vec.

crates/sqlite-lambda/tests/snapshots/examples__type_conversions.snap
jgraettinger: That's correct. It's unavoidable, because SQLite doesn't meaningfully distinguish booleans and text in it's returned value types.

Users _can_ work around this using sqlite's json functions and the `flow_document` output special case:
`select json_object('bool', json('true'), 'answer', 42) as flow_document from ... `. SQLite has some internal magic to thread through nested values of json functions that aren't exposed in the public API.

flow.schema.json
jgraettinger: Yep! Done.

PR #926, opened 2023-02-09
ops-catalog: new ops-catalog monitoring process

williamhbaker:
**Description:**

This creates a new `ops-catalog` binary which will be responsible for keeping the reporting infrastructure up to date with the list of tenants in the system. It works by querying the list of tenants from the database, rendering the necessary ops catalog specifications for that list of tenants from handlebars templates, and finally publishing the complete set of specs using `flowctl`.

In order to make the (re-)publishing of the entire ops catalog efficient enough to do on a frequent basis, an optimization was added to `flowctl catalog publish` to omit submitting specs to the control plane for publication that have not changed. This is accomplished by comparing an md5 checksum of the local spec with the md5 checksum of the live spec, requiring only a database query per batched group of specs.

#### _Performance Characteristics_

Since we'll be re-processing specs for every single tenant in the system frequently, it is important that the process be efficient enough given a reasonable amount of scale. I did some tests on my local dev machine to get an idea how many tenants this system can handle.

Considering _just_ template rendering (NOT including `flowctl catalog publish`, but does include querying the list of tenants from the `tenants` table), it handles relatively large amounts of tenants pretty well, generating the complete set of specs under these scenarios:
- 100 tenants / 1 level 1 derivation: ~30 milliseconds
- 10,000 tenants / 100 level 1 derivations (100 tenants each): ~500 milliseconds
- 20,000 tenants / 200 level 1 derivations (100 tenants each): ~1 second

In the case of 20,000 tenants, the rendered `flow.yaml` is about 16 megabytes.

Actually publishing the specs takes a lot longer than rendering the templates. This PR does not include optimizing expansion of the derivations, so publishing a spec for even a single tenant will expand to all others by traversing the level 1 & level 2 derivations. It will be important to implement that in the near future. I observed these ballpark scenarios testing locally for the entire render templates + publish process:
- Publishing the entire catalog for 100 new tenants simultaneously with no existing tenants: ~30 seconds
- Adding a single new tenant with 100 existing tenants all on a single level 1 derivation: ~15 seconds
- With 1,000 tenants in the system on 10 level 1 derivations (100 tenants per l1), a single "no-op" monitoring cycle takes ~500 ms. This includes getting the list of tenants, rendering the specs, & `flowctl catalog publish`. Adding a single new tenant to this existing base of 1,000 tenants results in a publication that takes a little over a minute to complete. Presumably this single tenant publication could be brought closer to the ~15 second range if the derivation expansion optimization were implemented.

#### _Testing_

I tested this locally under a variety of scenarios. The most thorough test was creating 1000 tenants over 100 level 1 derivations via a script, and then manually creating a handful of new tenants either adding to existing level 1 derivations or resulting in an entirely new level 1 derivation for that tenant. For each of these manually creating tenants, I created a hello world capture to cause stats documents to be emitted, and confirmed that stats were being reported correctly into the `catalog-stats` table.

#### _Rollout Strategy_

We will need to execute a few steps to fully roll this out:
1. Deploy the updated version of the agent. Once this is done, any newly created tenants will not include the individual tenant ops catalog publication.
2. Apply the changes in the `new_ops_catalog_setup.sql` script. This will:
    - Nuke the existing `ops_catalog_template` table and the function `internal.create_ops_publication(tenant_prefix catalog_tenant, ops_user_id uuid)`
    - Add the `md5` column to `live_specs` and the column `l1_stat_rollup` to `tenants`
    - Clear out the `catalog_stats` table as well as the materialization metadata tables.
    - Create a publication that will delete all the existing per-tenant ops catalog tasks: `ops/tenant/catalog-stats-view`, `ops/tenant/tests/catalog-stats`, and `ops/tenant/catalog-stats`
    - Add the `md5` column to the `live_specs_ext` view. The best way I could come up with to do this is to drop the existing `live_specs_ext` and re-create it, since it includes a `select live_specs.*` _before_ some other columns, so a `create or replace` won't work. The `draft_specs_ext` view must be dropped & re-created as well since it depends on `live_specs_ext`.
    - *Note*: Once this step is done, stats data will not be available for anyone until the new `ops-catalog` is deployed.
3. Wait for the publication created in (2) to complete.
4. Re-create the `catalog_stats` table.
5. Deploy `ops-catalog`. Upon startup it will do a fresh publication of the entire ops catalog. We've got ~100 tenants in the system currently so doing this all as one big publication will take a minute or so, but should work fine based on my testing.

**Workflow steps:**

In addition to the long-running `monitor` command, there is a `generate` command which can be used to render specs based on a json list of tenants that you provide over stdin. This is for quickly iterating on template changes locally, and supports running tests on the generated catalog using `flowctl-go`.

For example, this command will write out the ops catalog specs for several tenants:
```bash
echo '{"tenant": "hello.there", "l1_stat_rollup": 0}\n{"tenant": "some_tenant/", "l1_stat_rollup": 0}{"tenant": "other-tenant/", "l1_stat_rollup": 1}' | cargo run generate --output-dir generated
```

The resulting catalog can then be tested:
```bash
flowctl-go test --source generated/flow.yaml --directory=generated/test
```

N/A

**Notes for reviewers:**

A few todos still left for this:
- [x] Supabase migration procedures
- [x] Create a readme
- [x] Jsonnet for deployment

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/926)


crates/agent-sql/src/text_json.rs
```diff
@@ -33,19 +33,17 @@ impl<T> sqlx::postgres::PgHasArrayType for TextJson<T> {
     }
 }
 
-impl<T: Serialize> Encode<'_, postgres::Postgres> for TextJson<T> {
+// Encode and Decode pass through to the sqlx::types::Json implementation,
+// but are restricted to the JSON (and not JSONB) postgres types.
+impl<'q, T: Serialize> Encode<'q, postgres::Postgres> for TextJson<T> {
```
williamhbaker, line 38: This ended up not actually preventing the leading space from being encoded with JSON in the database (sqlx does it anyway), but felt a bit cleaner so I left it in.

Makefile
```diff
@@ -175,6 +175,10 @@ ${RUST_MUSL_BIN}/flow-connector-init:
 ${RUST_MUSL_BIN}/flow-network-tunnel:
 	cargo build --target x86_64-unknown-linux-musl --release --locked -p network-tunnel
 
+.PHONY: ${RUST_MUSL_BIN}/flow-ops-catalog
+${RUST_MUSL_BIN}/flow-ops-catalog:
+	cargo build --target x86_64-unknown-linux-musl --release --locked -p ops-catalog
```
jgraettinger, line 180: fwiw I don't think this needs to be musl (it's not running inside someone else's container)

crates/agent-sql/src/text_json.rs
jgraettinger: How did you end up resolving this? I'm not seeing specific handling and I see you added a round-trip test to confirm it's not modified.

jgraettinger (approved):
LGTM! Looks great, I found this very straightforward.

crates/agent-sql/src/text_json.rs
williamhbaker: The column that calculates the hash does a `trim` on the value before hashing it to remove the leading space, see [`supabase/migrations/09_publications.sql`](https://github.com/estuary/flow/pull/926/commits/a7ba22fdd55d66beefa1e27b1824c086e2ad6aee#diff-8e4208014719024769b27f0fba338b404fd0f070e1f8f7cd3716afcde7c3946bR58)

mdibaiee:
I have a nit:
We use Mustache templates for OAuth specification templates (we originally used handlebars but ran into compatibility issue with typescript), now seeing handlebars used it feels a bit icky to me since we will have two different templating engines used in different places (although this one is not user-facing AFAIK, but still it sets us up for some confusion).

What do you think about using [Mustache](https://docs.rs/mustache/latest/mustache/) templates here instead of Handlebars?

williamhbaker:
> What do you think about using [Mustache](https://docs.rs/mustache/latest/mustache/) templates here instead of Handlebars?

We're doing some things in these templates like accessing parent properties with nested loops & using helper functions that wouldn't be directly possible with mustache. I'd imagine it would be possible to manipulate the data on the application side and provide a more extensive context to the templating engine that would produce the same result, but I think it's preferable here that the template syntax can handle things directly. The templating we are doing here is complicated enough (though admittedly - it's not that complicated) that using a different engine feels justified. And as you mentioned they are very different applications so I don't think there is going to be a lot of churn from switching between the two.

williamhbaker:
Added migration scripts and a description of the steps we'll need to take to roll this out to the PR description.

Makefile
williamhbaker: Good call - changed to not build as musl.

jgraettinger:
Since we'll be blowing away and re-creating `catalog_stats` anyway -- should we make it non-partitioned?

I'd originally partitioned it because I though individual stats loaders would materialize to _just_ their partitioned table instance. That clearly didn't work, and instead the materializations now point at `catalog_stats`. It's partitioned nature is feeling pretty vestigial at this point.

When it does come time to "web scale" the `catalog_stats` table, I think we'd be looking at a solution like Clickhouse or something anyway, so the operational overhead of partitioning is not feeling worth it?

williamhbaker:
> Since we'll be blowing away and re-creating `catalog_stats` anyway -- should we make it non-partitioned?
> 
> I'd originally partitioned it because I though individual stats loaders would materialize to _just_ their partitioned table instance. That clearly didn't work, and instead the materializations now point at `catalog_stats`. It's partitioned nature is feeling pretty vestigial at this point.
> 
> When it does come time to "web scale" the `catalog_stats` table, I think we'd be looking at a solution like Clickhouse or something anyway, so the operational overhead of partitioning is not feeling worth it?

That's probably a good idea, the partitions aren't doing us any favors now (and haven't been for a while). I'll take the partition creation out of the agent and update the migrations scripts so we end up with a non-partitioned table.

jgraettinger:
👍 still LGTM

PR #927, opened 2023-02-09
Fix flowctl catalog pull-specs with merge-spec

### LINKED ISSUE #924: Local typescript module corruption after runnin `flowctl catalog pull-specs --existing merge-spec`
Steps to reproduce:

- Start with an empty directory
- Run `flowctl catalog pull-specs --name demo/wikipedia/lines-and-bots`
- Run `flowctl catalog pull-specs --name demo/wikipedia/test-lines-and-bots --existing merge-spec`

Expected behavior is that you just have both the derivation and the test defined in the flow.yaml, and that the typescript module for the derivation is exactly the same as the typescript module from the live spec.

What's actually happening is that the first line of the typescript module (in this case `demo/wikipedia/lines-and-bots.ts`) gets partially overwritten by the relative URL. Here's the actual first line of that typescript module, after running the above commands:

```
lines-and-bots.tsion, OutputDocument, Register, NewFactTableSource } from 'flow/demo/wikipedia/lines-and-bots';
```

The problem is first introduced [here](https://github.com/estuary/flow/blob/04ae85bf66cc961be15f49a4be96dda1f3aa395e/crates/flowctl/src/source/unbundle.rs#L146) when we merge an existing (unbundled) catalog with the new (bundled) catalog. This results in a catalog where the typescript module for an _existing_ derivation (in this case `demo/wikipedia/lines-and-bots`) has `module: lines-and-bots.ts`. When we then write out the catalog, we're partially overwriting the existing module (which is valid Typescript code) with the new value (which is just a relative path).

I'm still not entirely sure, but I _ think_ a reasonable fix here might be to track which entities were pre-existing, and skip writing indirected files for those entities.

psFried:
**Description:**

Fixes #924

When running `flowctl catalog pull-specs --existing merge-spec`, it was possible to mangle the content of indirected resources such as typescript modules, connector configs, schemas, etc. This would happen if you had an existing directory with the spec for `acmeCo/specA` and then ran:
`flowctl catalog pull-specs --existing merge-spec --name acmeCo/specB` In that case, an existing indirected typescript module (or schema or whatever) from `specA` would end up having its first line of content overwritten by the relative path to that content.

The fix here is to just skip re-writing indirected content for specs that are already in the current directory, but not included in the set of specs being pulled.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/927)

### DIFF
--- a/crates/models/src/catalogs.rs
+++ b/crates/models/src/catalogs.rs
@@ -61,6 +61,17 @@ impl Catalog {
         let generator = schemars::gen::SchemaGenerator::new(settings);
         generator.into_root_schema_for::<Self>()
     }
+
+    /// Returns the names of all specs that are directly included within this catalog.
+    /// This does _not_ include specs from imported catalogs.
+    pub fn all_spec_names(&self) -> impl Iterator<Item = &str> {
+        self.collections
+            .keys()
+            .map(AsRef::<str>::as_ref)
+            .chain(self.captures.keys().map(AsRef::<str>::as_ref))
+            .chain(self.materializations.keys().map(AsRef::<str>::as_ref))
+            .chain(self.tests.keys().map(AsRef::<str>::as_ref))
+    }
 }
 
 fn collections_schema(gen: &mut schemars::gen::SchemaGenerator) -> schemars::schema::Schema {
--- a/crates/flowctl/src/source/unbundle.rs
+++ b/crates/flowctl/src/source/unbundle.rs
@@ -2,6 +2,7 @@ use crate::source::Existing;
 use anyhow::Context;
 use serde_json::value::RawValue;
 use serde_json::Value;
+use std::collections::BTreeSet;
 use std::convert::AsRef;
 use std::path::Path;
 use tokio::io::AsyncWriteExt;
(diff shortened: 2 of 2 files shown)


crates/models/src/catalogs.rs
```diff
@@ -61,6 +61,17 @@ impl Catalog {
         let generator = schemars::gen::SchemaGenerator::new(settings);
         generator.into_root_schema_for::<Self>()
     }
+
+    /// Returns the names of all specs that are directly included within this catalog.
+    /// This does _not_ include specs from imported catalogs.
+    pub fn all_spec_names(&self) -> impl Iterator<Item = &str> {
+        self.collections
+            .keys()
+            .map(AsRef::<str>::as_ref)
```
williamhbaker, line 70: Fancy!

williamhbaker (approved):

crates/models/src/catalogs.rs
psFried: lol after our conversation yesterday, I just had to try it the other way :upside_down_face:

PR #929, opened 2023-02-13
control-plane: significantly accelerate authorization checks

jgraettinger:
**Description:**

This is a package of schema changes, test updates, and a new benchmark which quantify and then improve the performance of various authorization primitives and views within the control plane. Based on testing here, we can expect a substantial speedup of queries against the production control plane, even where a user has many imputed roles.

No breaking changes to semantics or public APIs:

 * `auth_roles()` has a new, optional `min_capability` parameter.
 * the role_grants select policy is restricted to require `admin` of the subject and object roles, rather than just `read` of the object role. This was overly permissive and the prior behavior could be considered a bug.

**Workflow steps:**

Apply DB migration and enjoy better dashboard performance.

No user-facing changes.

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/929)

### DIFF
--- a/supabase/migrations/04_auth.sql
+++ b/supabase/migrations/04_auth.sql
@@ -86,6 +86,9 @@ create table user_grants (
 );
 alter table user_grants enable row level security;
 
+-- Index that accelerates operator ^@ (starts-with) for combined_grants_ext view.
+create index idx_user_grants_object_role_spgist on user_grants using spgist (object_role);
+
 comment on table user_grants is
   'Roles and capabilities that the user has been granted';
 comment on column user_grants.user_id is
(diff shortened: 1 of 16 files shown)


supabase/migrations/04_auth.sql
```diff
@@ -123,74 +125,88 @@ comment on column role_grants.capability is
...
 
 -- Policy permissions for role_grants.
-create policy "Users select role grants they recieve or admin the object"
+create policy "Users select role grants where they admin the subject or object"
   on role_grants as permissive for select
-  using (auth_catalog(object_role, 'admin') or auth_catalog(subject_role, 'x_00'));
+  using (exists(
+    select 1 from auth_roles('admin') r
+    where (object_role ^@ r.role_prefix or subject_role ^@ r.role_prefix)
+  ));
```
mdibaiee, line 194: does inlining `catalog` help performance as well?

jgraettinger: I think you're referring to `auth_catalog()` calls here -- and the answer is a resounding "yes".

`auth_catalog(subject_role, 'admin')` is a SQL function, and I naively hoped / assumed that the optimizer would inline it, would see the call to auth_roles() which has no dependent arguments, and that auth_roles() is stable, and would then make a single call to auth_roles() for the query which is then used for evaluating every selected row examined by the policy.

What _actually_ happens, though, is that `auth_catalog(subject_role, 'admin')` is _not_ inlined. I'm pretty sure this is [because of its use of a sub-select](https://wiki.postgresql.org/wiki/Inlining_of_SQL_functions). Whatever the reason, it's not inlined and -- because it has an argument which is dependent on the current row -- it's re-evaluated for every single examined row of the selected table.

What happens now, is that a single call to auth_roles() is made for the lifetime of the query and its results are used over and over again with each examined row. That _does_ mean it's checking each row against the full set of authorized rows ever time -- ideally the optimizer would be able to invert this and identify subsets of the table to even look at in the first place -- but it's much much faster than re-computing auth_roles() every time.

The `combined_specs_ext` and `live_specs_ext` views are faster still, because they _do_ get inverted and the optimizer only bothers to look at rows of the base table which match your authorization scopes.

The included benchmark and `explain analyze` helped a lot in figuring all of this out.

jgraettinger: Just to expand, what I found is that the former `live_specs_ext` was certainly faster than a naive scan, because it avoided re-computing authorized roles with each role. _But_ it was still scanning every row of `live_specs` and comparing each against all of the user's roles.

Now it's faster, because the user's roles are used to determine which chunks of live_specs to look at.

mdibaiee: damn, I also assumed the function would get inlined. thanks for the explanation!

mdibaiee (approved):

williamhbaker (approved):

Issue #813, opened 2022-11-21
Issue with publications that contain more than 1 test


williamhbaker:
Several times I have noticed when I try to publish a catalog that has more than 1 test, the publication fails with:

```
2022-11-21T14:49:45Z temp-data-plane:2> ERRO[0008] protocol error (invalid state or invocation)  error="\"InvalidState\"" logSource=derive
2022-11-21T14:49:45Z test:2> FATA[0005] fatal error                                   err="reseting internal state between test cases: rpc error: code = Unknown desc = clearing registers of shard derivation/ops/ops/catalog-stats/00000000-00000000: protocol error (invalid state or invocation)"
2022-11-21T14:49:46.354789Z ERROR flowctl::draft: scope= detail=one or more test cases failed
Error: failed with status: testFailed
```

Need to find out exactly what circumstances are required to reproduce this, but it seems like a real issue.

If I split the publication up so that each publication only has a single test to run, I am able to get the catalogs running.

williamhbaker:
I have tried coming up with a simple reproducer of this without luck so far.

A sure way to get this error is to do the following with a local stack:
- Provision a new tenant using directives, for example `bobCo/`.
- Give your user an admin grant for both `ops/` and `bobCo/`.
- Create a draft, and add `ops/ops/catalog-stats-view` & `ops/bobCo/catalog-stats-view` to it with `catalog draft --name ...`
- Try to publish this draft and observe the error

I have tried with simple flows, combining different splits of publications, the user that created the thing, etc. and haven't been able to reproduce it. It might have something to do with the complexity of the tests and/or derivation documents, as I have only been using basic "hello world" types of documents. The `catalog-stats` flows have more extensive tests and document structures which might be triggering something.

jgraettinger:
This was due to the special nature of the `ops/$tenant/stats` and `ops/$tenant/logs` collections. Specifically, what happens is that a test over a derivation that reads from either of these collections, where the derivation itself lives under $tenant, causes new writes to those collections to occur -- which then confuse the test runner, which is expecting to strictly control when new data is ingested to the collections under test.

PR #930, opened 2023-02-13
ops-catalog: disable reporting materialization initially

williamhbaker:
**Description:**

This is a tweak to the rollout/migration strategy for the new ops-catalog. Initially, the ops-catalog will publish the final reporting materialization in a disabled state. This, along with an additional step in the rollout, will allow us to create the level 1 and 2 derivations and have them catch up before removing the existing stats materialization and table. This will minimize the amount of stats downtime users see.

The level 1 and 2 derivations will co-exist with the existing per-tenant publications. A complication with this is that the existing per-tenant reporting tests will fail when these derivations attempt to be published to `ops/` (see https://github.com/estuary/flow/issues/813). Because of this, we will actually need to delete the per-tenant tests using the old `generate-migration.sh` process. 4b693145b2f2beba908804c0db0e3c47afa68307 is included in this PR for this purpose: It is a copy of the old `ops-catalog` folder, with the per-tenant tests removed from the template. The commit that immediately follows deletes it again.

So the rollout will work in the following way:
- Delete the existing per-tenant tests via checkout of 4b693145b2f2beba908804c0db0e3c47afa68307 and running `generate-migration.sh`. This will also prevent any new tenants from including these tests.
- Run the `new_ops_catalog_1_initial_migration.sql` migration, which will simply create the `md5` column in `live_specs` and `live_specs_ext`.
- Deploy the `ops-catalog` using the image version that gets pushed from this commit to the main branch. The `agent` should still use an older image, which will continue to create tenants with their individual stats table partitions & reporting publications.
- That `ops-catalog` will publish the level 1 and 2 derivations. Wait for those to process all existing stats documents & "catch up" to the present. This will probably take on the order of ~24 hrs.
- Deploy the `agent` using the image version from this commit. From this point on, new tenants will _not_ be created with a stats table partition or reporting publications.
- Run `new_ops_catalog_2_delete_existing_stats.sql`. This will delete all data from the stats table and create a publication to delete all individual tenant reporting publications.
- After that publication is complete, run `new_ops_catalog_3_create_new_stats_table.sql` to re-create the `catalog_stats` table.
- Deploy the final `ops-catalog` with the reporting materialization enabled. This will require another flow PR to remove disabled shards configuration.

**Workflow steps:**

N/A

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/930)


supabase/pending/new_ops_catalog_1_initial_migration.sql
```diff
@@ -1,50 +1,14 @@
...
-
-    for current_tenant in
-        select * from tenants
-    loop
-        insert into draft_specs (draft_id, catalog_name, spec_type, spec)
-        select new_draft_id, catalog_name, null, null
-        from live_specs
-        where catalog_name in (
-            'ops/' || current_tenant.tenant || 'catalog-stats-view',
-            'ops/' || current_tenant.tenant || 'tests/catalog-stats',
```
jgraettinger, line 40: we _could_ restore this but _only_ delete `tests/catalog-stats`. Up to you whether that's actually expedient.

jgraettinger (approved):
LGTM

PR #931, opened 2023-02-13
ops-catalog: fix token when using local stack

williamhbaker:
**Description:**

Fixes the `ops-catalog` token used in the local stack by changing it to a token signed using the local JWT secret for `ops/` that does not expire until a long time into the future.

**Workflow steps:**

Use local stack and see that it works successfully.

N/A

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/931)


williamhbaker:
Going to close this out since we will be pursuing an alternate strategy.

PR #932, opened 2023-02-14
ops-catalog: use flag for prod instead of local

williamhbaker:
**Description:**

Using the boolean flag `local` with a default of `true` made it effectively impossible to ever have local _not_ be true. This changes that around: Providing a flag of `--prod` will cause the templates to be rendered with the (encrypted) prod credentials, while not passing any flag will continue to render the template with the default local credentials. This way we can actually run with the production credentials in production, instead of being forever doomed to used local credentials.

**Workflow steps:**

N/A

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/932)


williamhbaker:
Going to close this out since we will be pursuing an alternate strategy.

PR #935, opened 2023-02-14
ops-catalog: remove ops-catalog

williamhbaker:
**Description:**

This removes the `ops-catalog` crate and ~most of what was added in https://github.com/estuary/flow/pull/926. It's a bit weird since some of the initial migration steps were completed and I'm not removing _everything_ from that PR.

Here's what is staying:
- agent: remove ops publication from tenant onboarding (75856c7386dc85893776ca399d0a3d21df802b3b): The agent will no longer create a reporting publication for each new tenant. This means they won't have an ops or stats collection published, and also won't have a derivation or materialization for their stats. This is good because new individual stats materializations are broken. I confirmed locally that a tenant is able to create a new task even without the specifically published stats/logs collections - a journal partition is automatically created for these tasks' stats & logs. There's a nice user experience benefit here in that when a user signs up they won't see any pre-existing tasks, which is especially nice in the case where they are failed and/or unusable.
- agent-sql: TextJson serde pass through to sqlx (5457684d52a1dcf1eb8fec8f3876eca48390015e): Not a very consequential change but don't see a reason to revert it either.
- flowctl: skip unchanged specs with catalog publish (de0d1c6e56b42895e47919065259b6aaeece24d3): This keeps the `md5` hashing system into place for live specs and `flowctl` comparing new specs with that to determine if a spec has changed. It's not as important to use now that we won't be using `ops-catalog` but should still be nice to keep in place for future uses.

The pending migration `supabase/pending/remove_ops_catalog.sql` included in this PR can be run after the new agent is deployed to put things as they should be.

**Workflow steps:**

N/A

**Notes for reviewers:**

See inline comments for a bit more explanation that might be helpful in reviewing this.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/935)


local/start-flow.sh
```diff
@@ -40,6 +39,19 @@ function bail() {
 # verify requirements here, before we get into the tmux session.
 command -v nc || bail "netcat must be installed (nc command was not found)"
 
+function copy_local_flowctl_config() {
```
williamhbaker, line 42: Just moving this function back to the way it used to be - in `start-flow.sh` and not called with any particular argument.

supabase/migrations/11_stats.sql
```diff
@@ -22,7 +22,7 @@ create table catalog_stats (
     failures            integer      not null default 0,
     ts                  timestamptz  not null,
     flow_document       json         not null
-);
+) partition by list (substring(catalog_name for position('/' in catalog_name)));
```
williamhbaker, line 25: We never dropped this table or its partitioning, so this migration is reverted back to reflect that.

New tenants won't have stats partitions created for them though, and new tenants won't have stats materializations created for them.

supabase/migrations/16_tenants.sql
```diff
@@ -7,8 +7,7 @@ create table tenants (
   tenant                  catalog_tenant unique not null,
 
   tasks_quota             integer        not null default 10,
-  collections_quota       integer        not null default 100,
-  l1_stat_rollup          integer        not null default 0
```
williamhbaker, line 11: This column was added as an initial migration with the `ops-catalog` rollout, but is now not needed.

jgraettinger (approved):
LGTM

PR #937, opened 2023-02-15
docs: source-sqlserver

oliviamiannone:
**Description:**

Adds docs for new source-sqlserver connector (https://github.com/estuary/connectors/pull/532). 

**Notes for reviewers:**

Included a quick how-to section on updating the collection key with flowctl. I figure that'll probably go away or change pretty soon but good to have for now. 

I did not test this on AWS as of now but hoping to circle back on that soon as time permits.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/937)

### DIFF
--- a/site/docs/reference/Connectors/capture-connectors/sqlserver.md
+++ b/site/docs/reference/Connectors/capture-connectors/sqlserver.md
@@ -1,0 +1,277 @@
+---
+sidebar_position: 6
+---
+# Microsoft SQL Server
+
+This connector uses change data capture (CDC) to continuously capture updates in a Microsoft SQL Server database into one or more Flow collections.
+
+It’s available for use in the Flow web application. For local development or open-source workflows, [`ghcr.io/estuary/source-sqlserver:dev`](https://ghcr.io/estuary/source-sqlserver:dev) provides the latest version of the connector as a Docker image. You can also follow the link in your browser to see past image versions.
+
+## Supported versions and platforms
+
+This connector supports SQL Server 2017 and later on major cloud providers,
+as well as self-hosted instances.
+Setup instructions are provided for the following platforms:
+
+* [Self-hosted SQL Server](#setup-self-hosted-sql-server)
+* [Azure SQL Database](#setup-azure-sql-database)
+* [Amazon RDS for SQL Server](#setup-amazon-rds-for-sql-server)
+* [Google Cloud SQL for SQL Server](#setup-google-cloud-sql-for-sql-server)
+
+## Prerequisites
+
+To capture change events from SQL Server tables using this connector, you need:
+
+* For each table to be captured, a primary key should be specified in the database.
+If a table doesn't have a primary key, you must manually specify a key in the associated Flow collection definition while creating the capture.
+[See detailed steps](#specifying-flow-collection-keys).
+
+* [CDC enabled](https://learn.microsoft.com/en-us/sql/relational-databases/track-changes/enable-and-disable-change-data-capture-sql-server?view=sql-server-ver16)
+on the database and the individual tables to be captured.
+(This creates *change tables* in the database, from which the connector reads.)
+
+* A **watermarks table**. The watermarks table is a small “scratch space” to which the connector occasionally writes a small amount of data to ensure accuracy when backfilling preexisting table contents.
+
+* A user role with:
+  * `SELECT` permissions on the CDC schema and the schemas that contain tables to be captured.
+  * `SELECT` and `UPDATE` permissions on the watermarks table
+
+To meet these requirements, follow the steps for the your hosting type.
+
+* [Self-hosted SQL Server](#setup-self-hosted-sql-server)
+* [Azure SQL Database](#setup-azure-sql-database)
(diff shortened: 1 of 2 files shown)


site/docs/reference/Connectors/capture-connectors/sqlserver.md
```diff
@@ -0,0 +1,277 @@
...
+If a table doesn't have a primary key, you must manually specify a key in the associated Flow collection definition while creating the capture.
+[See detailed steps](#specifying-flow-collection-keys).
+
+* [CDC enabled](https://learn.microsoft.com/en-us/sql/relational-databases/track-changes/enable-and-disable-change-data-capture-sql-server?view=sql-server-ver16)
+on the database and the individual tables to be captured.
+(This creates *change tables* in the database, from which the connector reads.)
+
+* A **watermarks table**. The watermarks table is a small “scratch space” to which the connector occasionally writes a small amount of data to ensure accuracy when backfilling preexisting table contents.
+
+* A user role with:
```
willdonnelly, line 35: There's one additional permission that the capture user needs to have, which is that they must have access to the change data itself [1]. This is controlled through the `@role_name` parameter when CDC is enabled on a particular table, and the capture user needs to either be named explicitly there or be (transitively) a member of whatever role is specified.

[1] This access is distinct from the change tables `SELECT` access for some reason -- if they only have `SELECT` on the `cdc` schema we can poll for change events but the result is just a bunch of `NULL` values.

site/docs/reference/Connectors/capture-connectors/sqlserver.md
```diff
@@ -0,0 +1,277 @@
...
+GRANT SELECT ON SCHEMA :: cdc TO flow_capture;
+--Create the watermarks table, grant permissions, and enable CDC.
+CREATE TABLE dbo.flow_watermarks(slot INTEGER PRIMARY KEY, watermark TEXT);
+GRANT SELECT, UPDATE ON dbo.flow_watermarks TO flow_capture;
+EXEC sys.sp_cdc_enable_table @source_schema = 'dbo', @source_name = 'flow_watermarks', @role_name = 'flow_capture', @capture_instance = 'dbo_flow_watermarks';
+--Add some data to watermarks table to ensure proper connector initiation.
+INSERT INTO dbo.flow_watermarks VALUES (0, 'dummy-value');
+```
+
+2. Set up an [SSH server for tunneling](../../../../guides/connect-network/).
```
willdonnelly, line 70: Should we have an "Alternatively you can whitelist our IP address in your firewall rules" note here, for symmetry with how the managed DB setup instructions call out the SSH tunneling option?

site/docs/reference/Connectors/capture-connectors/sqlserver.md
```diff
@@ -0,0 +1,277 @@
...
+CREATE USER flow_capture FOR LOGIN flow_capture;
+/*Grant the user permissions on the CDC schema and schemas with data.
+This assumes all tables to be captured are in the default schema, `dbo`.
+Add similar queries for any other schemas that contain tables you want to capture.*/
+GRANT SELECT ON SCHEMA :: dbo TO flow_capture;
+GRANT SELECT ON SCHEMA :: cdc TO flow_capture;
+--Create the watermarks table, grant permissions, and enable CDC.
+CREATE TABLE dbo.flow_watermarks(slot INTEGER PRIMARY KEY, watermark TEXT);
+GRANT SELECT, UPDATE ON dbo.flow_watermarks TO flow_capture;
+EXEC sys.sp_cdc_enable_table @source_schema = 'dbo', @source_name = 'flow_watermarks', @role_name = 'flow_capture', @capture_instance = 'dbo_flow_watermarks';
```
willdonnelly, line 65: Yesterday I just copy-pasted you the setup sequence instructions we're currently using our test database initialization. Now that I'm reading through this in the context of user-facing instructions, I think it would be clearer if we changed the last few lines to something like:

```
-- Add some data to the watermarks table to ensure proper connector initiation.
INSERT INTO dbo.flow_watermarks VALUES (0, 'dummy-value');
-- Enable CDC on the watermarks table. You should run this query for all other
-- tables you intend to capture as well.
EXEC sys.sp_cdc_enable_table @source_schema = 'dbo', @source_name = 'flow_watermarks', @role_name = 'flow_capture';
```

(The most important bit is mentioning that they'll need to run `sys.sp_cdc_enable_table` on all the other tables they intend to capture. I figure that moving that command to the end makes that fact a bit more noticeable. Removing the capture instance name is a no-op here since the default is `<schema>_<table>` but it seems marginally easier for the user to generalize the query to other tables if they only have to change the source table name in one spot.)

willdonnelly (approved):
LGTM, a few notes where it seems like things could be made a bit clearer.

site/docs/reference/Connectors/capture-connectors/sqlserver.md
oliviamiannone: Thanks - for some reason I'd been under the impression this was less of a viable option here than on managed cloud platforms but now I think about it I don't see why it would be...

PR #942, opened 2023-02-16
shuffle: partition value tweaks and covered key optimization

jgraettinger:
**Description:**

Two changes in this PR.

First, tweak the encoding of partition values so that they can be round-tripped back into JSON values and key tuples.
Do this without breaking current usage, by continuing to query-encode strings, and by prefixing non-string JSON scalars with a `%_` prefix -- which can never be produced through query-encoding.

While in here, address variations in query-encoding handling between Rust and Go so that both now produce identical values (for all tested strings).

Second, implement a shuffled-read optimization where, if a shuffle key is fully covered by the logical partitions of the source collection, we altogether avoid reading and tracking journals having documents which cannot map into our shard's covered key range. This is a transparent optimization which, if reverted, doesn't affect correctness.

**Workflow steps:**

No changes to product usage.

**Notes for reviewers:**

Recommend reviewing commit by commit.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/942)


williamhbaker (approved):
LGTM

PR #943, opened 2023-02-16
ops: output all tenant logs and stats to a single collection

williamhbaker:
**Description:**

Changes reporting of logs and stats to output to a single collection for all tenants. Journals will be automatically created to partition the singular logs & stats collections for each task. A single reporting derivation will be responsible for aggregating reporting metrics per data-plane, which will then be rolled-up by a higher level derivation and materialization into our stats table.

This is made possible by the optimization in https://github.com/estuary/flow/pull/942, which will allow us to efficiently scale out derivation shards per data plane as the number of journals grows large.

Rolling this out will consist of:
- Deploy this version of flow. All tasks will immediately start writing out logs & stats to partitions of the journals `ops.us-central1.v1/logs` & `ops.us-central1.v1/stats`.
    - The journals will look like `ops.us-central1.v1/stats/kind=capture/name=bobCo%2Fsource-hello-world/pivot=00`
- Run the `1_single_ops_collection.sql` migration. This will rename the existing `catalog_stats` table to preserve its data, rename the existing `ops/` tenant to `ops.us-central1.v1/`, and draft a deletion of all existing per-tenant reporting tasks (which will now be non-functional).
- After that deletion publication is complete, run `2_create_stats_table.sql` which will re-create the `catalog_stats` table. It is important to wait for the individual tenant task deletions to complete, otherwise they might delete the `catalog_stats` table as part of their deletion.
- Publish `ops-catalog/template-prod.flow.yaml` to create the logs & stats collections in the control plane, and deploy the reporting derivations & materialization.

Currently this PR doesn't include publishing the reporting catalog automatically when running a local stack. It can be published manually though using the local credentials for the `support@estuary.dev` account. I'll need to figure out a way for this to happen automatically.

Future work should also be done to enable partition-level authorization so that tenants have access to their specific logs & stats partitions.

Related to https://github.com/estuary/flow/issues/891

**Workflow steps:**

N/A

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/943)


crates/assemble/src/ops/mod.rs
```diff
@@ -23,8 +22,7 @@ macro_rules! gen_schema {
 
 /// Adds ops collections to the given partial built catalog. The tables will be modified in place
 /// to add the resources required for the ops (logs and stats) collections.
-pub fn generate_ops_collections(tables: &mut tables::Sources) {
-    let prefixes = all_top_level_prefixes(&*tables);
+pub fn generate_ops_collections(tables: &mut tables::Sources, dataplane: &str) {
```
jgraettinger, line 25: just a heads up, this `ops` module is essentially deprecated and we'll want to rip it all out once we get planned reporting rolled out, so any changes don't need to be pretty.

ops-catalog/template-common.flow.yaml
```diff
@@ -0,0 +1,65 @@
...
+        logs:
+          source:
+            name: ops.DATAPLANE/logs
+          publish: { lambda: typescript }
+        stats:
+          source:
+            name: ops.DATAPLANE/stats
+          publish: { lambda: typescript }
+
+  ops/catalog-stats-L2:
```
jgraettinger, line 42: perhaps just also have this be in ops.gcp-whatever ?

I don't believe a separate `ops/` tenant makes sense anymore -- we'll pick an arbitrary data-plane and not others to run it in, and the one we have seems reasonable.

supabase/seed.sql
```diff
@@ -4,6 +4,8 @@ insert into auth.users (id, email) values
   -- Root account which provisions other accounts.
   -- It must exist for the agent to function.
   ('ffffffff-ffff-ffff-ffff-ffffffffffff', 'support@estuary.dev'),
+  -- Account that owns the ops.$DATAPLANE collections for stats and logs.
```
jgraettinger, line 7: i think this should still be support@esuary.dev

jgraettinger (commented):
Left some comments below, but my major feedback is: we can simplify further, especially at this stage. Let's do the really dumb thing of `s/ops/ops.gcp-us-whatever/` and not worry _at all_ about the implications of multiple data planes.

crates/agent-sql/src/publications.rs
```diff
@@ -540,10 +541,11 @@ pub async fn resolve_storage_mappings(
         where starts_with(n, m.catalog_prefix)
            or starts_with('recovery/' || n, m.catalog_prefix)
            -- TODO(johnny): hack until we better-integrate ops collections.
-           or m.catalog_prefix = 'ops/'
+           or m.catalog_prefix = 'ops.' || $2 || '/'
```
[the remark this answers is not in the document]
williamhbaker: That will certainly simplify things, and especially makes sense if the agent is for all data-planes.

crates/agent-sql/src/publications.rs
```diff
@@ -540,7 +540,7 @@ pub async fn resolve_storage_mappings(
         where starts_with(n, m.catalog_prefix)
            or starts_with('recovery/' || n, m.catalog_prefix)
            -- TODO(johnny): hack until we better-integrate ops collections.
-           or m.catalog_prefix = 'ops/'
+           or m.catalog_prefix = 'ops.us-central1-c.v1/'
```
jgraettinger, line 543: nit: think there was a crossed wire here -- per the slack thread, we should include the region but not the availability zone

ops-catalog/template-common.flow.yaml
williamhbaker: Changed it to all use the same gcp-specific tenant

supabase/seed.sql
williamhbaker: Update this (among other things) in `seed.sql`. We'll just create the one ops tenant with the specific name we want, which seems to work fine.

ops-catalog/ops-log-schema.json
```diff
@@ -0,0 +1,39 @@
+{
```
jgraettinger, line 1: you could instead point to the schema in `crates/assemble/src/ops/ops-log-schema.json`, right?

They should be identical (and it would be a problem if they weren't!).

supabase/pending/1_single_ops_collection.sql
```diff
@@ -0,0 +1,47 @@
+begin;
+
+-- Rename the existing ops tenant to the new specific ops tenant.
+update tenants set tenant = 'ops.us-central1.v1/' where tenant = 'ops/';
+
+-- Rename the current catalog stats table to preserve its historical stats data.
+alter table catalog_stats owner to postgres;
```
jgraettinger, line 7: I think the table rename, as well as dropping flow_checkpoints & flow_materializations ought to wait until 2_create_stats_table.sql

I'm assuming you've run this migration locally, I'm a bit surprised that deleting `flow_materializstions_v2` doesn't cause a publication error when deleting the older `ops/tenant/catalog-stats-view` materializations, but :shrug:

supabase/pending/1_single_ops_collection.sql
```diff
@@ -0,0 +1,47 @@
+begin;
+
+-- Rename the existing ops tenant to the new specific ops tenant.
+update tenants set tenant = 'ops.us-central1.v1/' where tenant = 'ops/';
```
jgraettinger, line 4: do we need to update the ops role_grants , as well as the user_grant for support@estuary.dev ?

jgraettinger (commented):
A couple of questions below, mostly about migration and release order.

ops-catalog/template-common.flow.yaml
```diff
@@ -0,0 +1,65 @@
...
+        partition: true
+      name:
+        location: /shard/name
+        partition: true
+
+  ops.us-central1.v1/catalog-stats-L1:
+    schema: ops-catalog-stats-schema.json
+    key: [/catalogName, /grain, /ts]
+
+    derivation:
```
jgraettinger, line 28: add

```
shuffle:
  key: [/shard/name]
```

jgraettinger (approved):
LGTM % comments

ops-catalog/template-common.flow.yaml
williamhbaker: Done

supabase/pending/1_single_ops_collection.sql
williamhbaker: The materializations will still delete if the metadata tables aren't there (or their spec isn't in the metadata tables), but if they can't find the metadata, they won't do anything else to the endpoint: https://github.com/estuary/connectors/blob/03951232c66e203fec123c3dd1bcb6661bf9ecd8/materialize-sql/driver.go#L235-L239

I've just realized that this means they won't actually try to delete their target table if the metadata tables are already deleted, so we can indeed re-create the `catalog_stats` table in this same transaction without the deletion publications bothering it.

supabase/pending/1_single_ops_collection.sql
williamhbaker: Yep, good call. Also need to update the storage mappings for the ops tenant.

We need to keep the grants for `ops/` until the per-tenant deletion publications are complete. That along with the note above about re-creating the stats table in the same transaction as it is deleted led me to modify the migrations: There's still a second migration, but it is now all about updating the ops tenant - renaming the tenant itself, its grants, and its storage mappings.

There will be vestigial role grants for `$tenant` to read `ops/$tenant` left over in the system that we should clear out, as well as update the agent to stop creating sometime soon.

ops-catalog/ops-log-schema.json
williamhbaker: I updated `crates/assemble/src/ops` to refer to the schema docs in `ops-catalog` since you mentioned the `ops` module there being effectively deprecated, and also there is an additional schema document for `ops-catalog-stats-schema` in `ops-catalog` so they can all live there instead of being split up.

Interestingly they aren't exactly identical. The schema in `ops-catalog` had a couple of recent-ish updates compared to those in `crates/assemble/src/ops` ([default values](https://github.com/estuary/flow/pull/943/commits/23068a66657a24d5800c703375aac1d4cf0d2261#diff-acf8530839dcae4131d156d15cfeadba04fab44e4abfd07ec96520d33478a437R357), [source property for transforms](https://github.com/estuary/flow/pull/943/commits/23068a66657a24d5800c703375aac1d4cf0d2261#diff-acf8530839dcae4131d156d15cfeadba04fab44e4abfd07ec96520d33478a437R395), [a new log level](https://github.com/estuary/flow/pull/943/commits/23068a66657a24d5800c703375aac1d4cf0d2261#diff-ae74a402356c5dcbade0bdcaa9de6ee9eda7b579c251f9956a8a45957c66c438R8)), and also an interesting [difference in patterns for key/rClock ranges](https://github.com/estuary/flow/pull/943/commits/23068a66657a24d5800c703375aac1d4cf0d2261#diff-acf8530839dcae4131d156d15cfeadba04fab44e4abfd07ec96520d33478a437R491).

I don't know that these differences represent a problem at the moment, and they've been in existence for a while now actually anyway. Let me know what you think.

jgraettinger: Got it -- definitely use the more recent schemas from ops-catalog then. I believe this hasn't been an issue in practice because we don't actually validate documents being written to these collections from tasks. We only validate them on read, and those schemas are correct.

jgraettinger (approved):
LGTM

PR #944, opened 2023-02-16
Run config-encryption in docker locally

psFried:
This is to work around issues connecting from the oauth function to the config-encryption service. I feel fairly confident that this ought to work for Mac and WSL users, but would appreciate some help testing it out.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/944)


jgraettinger (approved):
LGTMM

PR #949, opened 2023-02-22
schema-inference: infer format: integer/number for string values

mdibaiee:
**Description:**

- Try to parse strings as integer or fractional, and based on that suggest `format: integer` or `format: number`.
- Take `NaN`, `Infinity` and `-Infinity` into consideration when doing this

**Workflow steps:**

- Feed in documents with string fields containing numerical values

N/A

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/949)

### DIFF
--- a/crates/schema-inference/src/inference.rs
+++ b/crates/schema-inference/src/inference.rs
@@ -1,6 +1,9 @@
-use doc::inference::{ArrayShape, ObjProperty, ObjShape, Shape};
-use json::schema::types;
+use doc::inference::{ArrayShape, ObjProperty, ObjShape, StringShape, Shape};
+use json::schema::{types, formats::Format};
 use serde_json::Value as JsonValue;
+use bigdecimal::BigDecimal;
+use std::str::FromStr;
+use num_bigint::BigInt;
 
 use crate::shape;
 
@@ -35,9 +38,25 @@ fn infer_number(value: &serde_json::Number) -> Shape {
     }
 }
 
-fn infer_string(_value: &str) -> Shape {
+fn infer_string(value: &str) -> Shape {
+    let (format, additional_type) = if BigInt::parse_bytes(value.as_bytes(), 10).is_some() {
+        (Some(Format::Integer), Some(types::INTEGER))
+    } else if BigDecimal::from_str(value).is_ok() || ["NaN", "Infinity", "-Infinity"].contains(&value) {
+        (Some(Format::Number), Some(types::INT_OR_FRAC))
+    } else {
+        (None, None)
+    };
+
+    let string = StringShape {
+        format,
+        ..Default::default()
+    };
+
+    let type_ = additional_type.map(|t| types::STRING | t).unwrap_or(types::STRING);
+
     Shape {
-        type_: types::STRING,
+        type_,
+        string,
         ..Default::default()
     }
 }
@@ -113,6 +132,26 @@ mod test {
 
         let shape = infer_shape(&json!(null));
         assert_eq!(types::NULL, shape.type_);
+    }
+
+    #[test]
+    fn build_string_types() {
+        let shape = infer_shape(&json!("1"));
+        assert!(shape.type_.overlaps(types::STRING));
+        assert!(shape.type_.overlaps(types::INTEGER));
+        assert_eq!(Some(Format::Integer), shape.string.format);
+
+        let shape = infer_shape(&json!("1.0"));
+        assert!(shape.type_.overlaps(types::STRING));
+        assert!(shape.type_.overlaps(types::INT_OR_FRAC));
+        assert_eq!(Some(Format::Number), shape.string.format);
+
+        for t in ["NaN", "Infinity", "-Infinity"] {
+            let shape = infer_shape(&json!(t));
+            assert!(shape.type_.overlaps(types::STRING));
+            assert!(shape.type_.overlaps(types::INT_OR_FRAC));
+            assert_eq!(Some(Format::Number), shape.string.format);
+        }
     }
 
     #[test]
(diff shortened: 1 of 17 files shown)


crates/schema-inference/src/snapshots/schema_inference__analyze__test__string_format_number_back_to_string.snap
```diff
@@ -0,0 +1,19 @@
...
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "type": "object",
+  "required": [
+    "a"
+  ],
+  "properties": {
+    "a": {
+      "type": [
+        "integer",
+        "string"
```
mdibaiee, line 15: There is a note here worth mentioning:

Documents are:
```
{"a": "1"}
{"a": "blah"}
```

Since one of the documents has a number as its string value, we give it the type `string, integer` and `format: integer`.
After merging it with the next document that shows a non-numeric string, we remove `format`, but it's harder to remove the type `integer` since we don't know at this stage how was that type added in the first place: maybe we did actually see an `integer` value at some point during the inference, or maybe we didn't and we just added this type because we saw strings with numeric values.

To put it another way: while it is valid for a field to be `type: string, integer` and only include strings, it is not valid for a field to have value `"blah"` with `format: integer`, so we can address the latter case, but the first case is ambiguous for us.

I don't think this is a big deal, but worth mentioning. Let me know if you think this can be problematic.

jshearer: I was wondering about this. Is it possible to add some metadata to the inferred schema? Also agree that it's not super important right now

jshearer (approved):
This all LGTM. I'm assuming that this is ultimately to support automatically converting number-like strings into numbers when we capture them?

crates/schema-inference/src/inference.rs
```diff
@@ -35,9 +38,25 @@ fn infer_number(value: &serde_json::Number) -> Shape {
     }
 }
 
-fn infer_string(_value: &str) -> Shape {
+fn infer_string(value: &str) -> Shape {
+    let (format, additional_type) = if BigInt::parse_bytes(value.as_bytes(), 10).is_some() {
+        (Some(Format::Integer), Some(types::INTEGER))
+    } else if BigDecimal::from_str(value).is_ok() || ["NaN", "Infinity", "-Infinity"].contains(&value) {
```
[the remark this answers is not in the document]
mdibaiee: Apparently GRPC also handles `NaN`, `Infinity` and `-Infinity` strings as special cases of float, so we decided we should, too:

https://protobuf.dev/programming-guides/proto3/

> JSON value will be a number or one of the special string values "NaN", "Infinity", and "-Infinity". Either numbers or strings are accepted. Exponent notation is also accepted. -0 is considered equivalent to 0.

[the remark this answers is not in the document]
jshearer: I see. Well, good enough for me!

mdibaiee:
> I'm assuming that this is ultimately to support automatically converting number-like strings into numbers when we capture them?

Yes, we already have this automatic conversion in-place in our materializations, so with the schema inference supporting this we will be able to be smart about this automatic conversion for captures where schema inference is used.

crates/schema-inference/src/snapshots/schema_inference__analyze__test__string_format_number_back_to_string.snap
jgraettinger: > Since one of the documents has a number as its string value, we give it the type string, integer and format: integer.

This is a drive by, but: shouldn't we give it _only_ type `string`, as in `{type: string, format: integer}`? We have yet to see an actual integer.

mdibaiee: @jgraettinger I think in that instance we don't convert that field's values into a number, we only do it if the type permits it.

But it's valid to say unless the data itself contains values which are actually of type integer / number, we do not convert.

mdibaiee: Okay, thinking about this more, there will be another edge case if we decide to only give it type `string`:

Imagine these documents come in:
```
{"a": 1}
{"a": "2"}
{"a": "NaN"}
```

After reading the first two documents, schema inference will suggest:
```
a:
  type: [string, integer]
  format: integer
```

And on reading the third document, we will have:
```
a:
  type: [string, integer]
  format: number
```

But this is wrong, we would expect that on reading the `NaN` which is a special case of number, we would get:
```
a:
  type: [string, number]
  format: number
```

mdibaiee: I think the current implementation's caveat is something we can live with. Unless someone has a strong opinion against this, I will merge this PR tomorrow.

jgraettinger: Given the `1, "2", "NaN" sequence:

```
a:
  type: [string, integer]
  format: number
```
Seems 100% correct to me. We _haven't_ seen a JSON number, only a string-encoded number (as NaN, which is never an integer), making this is the most-correct and tightly-fitting schema.

jgraettinger: Ahh, do we not handle the `{type: [string, integer], format: number}` case downstream within the SQL materializations? Could we?

mdibaiee: @jgraettinger no we can't handle that `format` in this case. We can only do conversion if the conversion fits with the `type`. In this instance the materialized field will have type `string, integer` and the format is basically no-op from our perspective, since `NaN` cannot be materialized into an `integer` field.

mdibaiee: @jgraettinger From my perspective, `1, 2, "NaN"` is the same as `1, 2, 1.1` in this instance, so my expectation of the resulting type is different from yours

PR #953, opened 2023-02-23
flowctl: generate, persist and use refresh_token for continuous access

mdibaiee:
**Description:**

- Use the given `access_token` on at login time to generate a `refresh_token` and use that `refresh_token` to generate new `access_tokens` if they are expired.

**Workflow steps:**

```
flowctl auth token --token='YOUR TOKEN'
# wait for expiry (one day) then the command below should still work
# you can check whether the access_token has been updated by checking the
# config directory
# $HOME/.config/flowctl/${profile}.json on linux
# $HOME/Library/Application Support/flowctl/${profile}.json on macos

flowctl draft list
```

- You can also manually edit the JWT token using https://jwt.io/ debugger and set the expiry date to a time in the past to see the refresh logic in action

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/953)

### DIFF
--- a/crates/flowctl/src/controlplane.rs
+++ b/crates/flowctl/src/controlplane.rs
@@ -1,3 +1,4 @@
+use keyring::Entry;
 use crate::config::{Config, RefreshToken};
 use crate::{CliContext, api_exec};
 
@@ -46,7 +47,7 @@ pub(crate) async fn new_client(config: &mut Config) -> anyhow::Result<Client> {
 
             // Try to give users a more friendly error message if we know their credentials are expired.
             if let Err(e) = check_access_token(&api.access_token) {
-                if let Some(refresh_token) = &api.refresh_token {
+                if let Ok(refresh_token) = retrieve_credentials(&api.endpoint, &api.user_id).await {
                     let response = api_exec::<AccessTokenResponse>(
                         client.rpc("generate_access_token", format!(r#"{{"refresh_token_id": "{}", "secret": "{}"}}"#, refresh_token.id, refresh_token.secret))
                     ).await?;
@@ -68,12 +69,15 @@ pub(crate) async fn new_client(config: &mut Config) -> anyhow::Result<Client> {
 pub async fn configure_new_access_token(ctx: &mut CliContext, token: String) -> anyhow::Result<()> {
     // try to catch issues caused by missing or extra data that may have been accidentally copied
     let jwt = check_access_token(&token)?;
-    ctx.config_mut().set_access_token(token);
+    ctx.config_mut().set_access_token(token, jwt.sub);
     let client = ctx.controlplane_client().await?;
     let refresh_token = api_exec::<RefreshToken>(
         client.rpc("create_refresh_token", r#"{"multi_use": true, "valid_for": "14d", "detail": "Created by flowctl"}"#)
     ).await?;
-    ctx.config_mut().set_refresh_token(refresh_token);
+
+    if let Some(api) = &ctx.config().api {
+        persist_credentials(&api.endpoint, &api.user_id, &refresh_token).await?;
+    }
 
     let message = if let Some(email) = jwt.email {
         format!("Configured access token for user '{email}'")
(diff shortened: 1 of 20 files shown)


psFried:
One thing I'd like to discuss about this is whether we should store this credential in plain text in the config file. TBH my gut reaction is that we should not. This was deemed expedient and "safe enough" when we were only talking about an access token with a limited expiration. But the [keyring crate](https://crates.io/crates/keyring) makes it easy to store this stuff securely, so it seems probably worth doing to me.

The only case where `keyring` seems inappropriate to me would be for some sort of automated process like a CI pipeline. For that case, allowing the use of an env variable seems far more convenient than needing to write the token to a config file, anyway. So IMO the ideal would be to try to read the refresh token from the env variable, and fall back to the user's keychain if that's not set.

This was actually all implemented previously in [this commit](https://github.com/estuary/flow/commit/0de12378b9f54112e3f6fc139464a74be8ee131a), back when we tried to get oauth refresh tokens working. So it might be possible to use that as a starting point. Curious if others also feel that this is worthwhile. I could probably be convinced to handle that in a subsequent PR, though my inclination would be to do it as part of this one.

crates/flowctl/src/controlplane.rs
```diff
@@ -50,27 +66,98 @@ pub(crate) fn new_client(config: &Config) -> anyhow::Result<Client> {
...
+pub async fn configure_new_access_token(ctx: &mut CliContext, token: String) -> anyhow::Result<()> {
     // try to catch issues caused by missing or extra data that may have been accidentally copied
-    let email = check_access_token(&token)?;
-    ctx.config_mut().set_access_token(token);
-    println!("Configured access token for user: '{email}'");
+    let jwt = check_access_token(&token)?;
+    ctx.config_mut().set_access_token(token, jwt.sub);
+    let client = ctx.controlplane_client().await?;
+    let refresh_token = api_exec::<RefreshToken>(
+        client.rpc("create_refresh_token", r#"{"multi_use": true, "valid_for": "14d", "detail": "Created by flowctl"}"#)
```
psFried, line 75: Given that this refresh token can eventually be shown in the UI and revoked at will, I think it's reasonable to give it a very long expiration (or even no expiration) by default.

crates/flowctl/src/controlplane.rs
```diff
@@ -50,27 +66,98 @@ pub(crate) fn new_client(config: &Config) -> anyhow::Result<Client> {
...
     }
-    Ok(jwt.email)
+    Ok(jwt)
+}
+
+async fn retrieve_credentials(
+    endpoint: &url::Url,
+    user_id: &str,
+) -> anyhow::Result<RefreshToken> {
+    let keychain_entry = keychain_auth_entry(endpoint, user_id)?;
```
psFried, line 104: I'm thinking this might be a good place to check for a `FLOW_AUTH_TOKEN` env variable. The keychain is ideal for end-users using flowctl directly on their dev machine, but some CI servers and remote dev environments may not always provide a system keychain, and an env var would be much easier to configure on something like github actions, anyway.

crates/flowctl/src/controlplane.rs
```diff
@@ -50,27 +66,98 @@ pub(crate) fn new_client(config: &Config) -> anyhow::Result<Client> {
     }
 }
 
-pub fn configure_new_access_token(ctx: &mut CliContext, token: String) -> anyhow::Result<()> {
+pub async fn configure_new_access_token(ctx: &mut CliContext, token: String) -> anyhow::Result<()> {
```
psFried, line 69: Just nit picking, but having different types of tokens makes this code a bit hard to read without specific and consistent names for `refresh_token` vs `access_token`. Maybe rename the `token` parameter to `access_token`?

psFried (changes requested):
This is looking good. I managed to find a few more things to nit pick, but should be good to merge once those are addressed.

I'm assuming the CI failure is just due to the gcloud outage.

crates/flowctl/src/controlplane.rs
```diff
@@ -31,15 +32,30 @@ impl Deref for Client {
...
+#[derive(Deserialize)]
+struct AccessTokenResponse {
+    access_token: String,
+}
+
 /// Creates a new client. **you should instead call `CliContext::controlplane_client(&mut Self)`**, which
 /// will re-use the existing client if possible.
-pub(crate) fn new_client(config: &Config) -> anyhow::Result<Client> {
-    match &config.api {
+pub(crate) async fn new_client(config: &mut Config) -> anyhow::Result<Client> {
```
psFried, line 42: Passing a `&mut Config` here means that the config file will be re-written every time a client is created. If this parameter was a `&mut CliContext`, then it could only call `context.config_mut()` in the case that it actually needs to re-fresh the credential.

My thinking is that long term we should also stop storing the `access_token` in the config file, or maybe even not store it at all. I'm totally on board with _not_ changing that right now, to minimize the scope of the change. But I don't think we should go back to always writing the config file on every invocation.

To give some background on why always writing the config file is undesirable, consider the example where you're running flowctl as part of a Job in a kubernetes cluster. You'd often have that config file defined in a `ConfigMap` and then mount it into the container. But kubernetes updates the mounted files whenever the ConfigMap is updated, which could conflict with flowctl's writes to it. And anything flowctl writes to that file will ultimately be lost, anyway, since the ConfigMap is authoritative. There's other environments where you can really only mount that type of file as read-only. The general takeaway being that if we want flowctl to run in as many different environments as possible, it'll be easier if that file can be read-only.

So I'd like to get to a place where we generally treat that config file as read-only unless the user specifically asks us to update it (by running `auth login` or something). For now, it's totally fine to keep storing the `access_token` there, as long as we're not requiring that it be re-written on every invocation.

psFried (approved):
I found one other thing to nit-pick, but should be good to merge after.

psFried:
One other thing that I noticed is that pre-existing config files cause flowctl to exit with this error:

```
$ ./target/x86_64-unknown-linux-musl/release/flowctl auth token --token foo
Error: parsing config

Caused by:
    missing field `user_id` at line 8 column 3
```

We should make sure that the upgrade path for existing users is smooth. Making `user_id` an option would be one way to do it.

crates/flowctl/src/controlplane.rs
mdibaiee: It's actually not that straightforward, since we can't borrow config of `CliContext` as both mutable and immutable. What you are suggesting would mean that we take a `&mut CliContext` in `new_client`, and then match on `&context.config().api` (immutable borrow), and if conditions make it necessary, we use `context.config_mut().api` to mutate the config (mutable borrow), but compiler will complain about that, unless we want to introduce some cell.

On the other hand, I think this is a hard constraint to maintain when put that way: that we must not modify our config struct throughout the program if the program is non-interactive / running in a specific environment. `flowctl` can only know so much about its environment.

The best I can suggest is introducing an environment variable that prevents writes to config file, and we can use that env in read-only environments if it's really necessary. If it isn't, then I would suggest we just go on as-is.

psFried: Yeah, you'd need to break it down into two separate borrows. One immutable borrow to check if there's existing credentials and whether they're still valid, and then a separate, mutable borrow to actually update the config if need be.

With regard to not writing to the config file more generally, I think I'd phrase it as more of an axiom about config files rather than a constraint about environments. It's generally bad form to store application managed "state" in a config file because it makes it harder for users to actually use it as a config file, mainly because it breaks lots of assumptions people have about "config" files, especially when considering other environments.

In this case, I think it's totally fine to merge as it is, and we can always revisit the dual borrow later. Perhaps at the same time as we remove the access token from the config file :shrug:

PR #956, opened 2023-03-01
fix: Specify a 30s timeout for discover operations.

### LINKED ISSUE #911: Add Reasonable in-App Timeouts
**Problem**
Operations in the UI can take a long time.  When this happens on discover, the user can not cancel the operation and needs to re-enter all their information.

**Proposal**
1. Lower global discovery timeout to 20 seconds
2. Ensure SSH timeouts happen at a reasonable timeframe.  This theoretically could also be the 20s above, bit might be good to make it tighter so that users don't have to always wait 20s.  Ex. 10s could make more sense.

jshearer:
**Description:**

There appeared to already be a timeout in place in `cmd-api-discover.go`, but the codepath that introduced the timeout was never reached, and was instead called from `cmd-discover.go`.

Fixes #911 

**Note:** This doesn't change any SSH timeouts, as I believe those are in the specific connectors, or possibly the SQL connector boilerplate or something like that. That can come next if the 30s timeout is still not enough, unless we think it's critical to do it now

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/956)

### DIFF
--- a/go/flowctl-go/cmd-api-discover.go
+++ b/go/flowctl-go/cmd-api-discover.go
@@ -67,7 +67,7 @@ func (cmd apiDiscover) execute(ctx context.Context) (*pc.DiscoverResponse, error
 func (cmd apiDiscover) Execute(_ []string) error {
 	defer mbp.InitDiagnosticsAndRecover(cmd.Diagnostics)()
 	mbp.InitLog(cmd.Log)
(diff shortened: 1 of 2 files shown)


jgraettinger (approved):
LGTM

PR #965, opened 2023-03-13
docs: materialize-redshift

williamhbaker:
**Description:**

Docs for new `materialize-redshift` connector, see https://github.com/estuary/connectors/pull/580

**Workflow steps:**

N/A

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/965)

### DIFF
--- a/site/docs/reference/Connectors/materialization-connectors/amazon-redshift.md
+++ b/site/docs/reference/Connectors/materialization-connectors/amazon-redshift.md
@@ -1,0 +1,122 @@
+
+# Amazon Redshift
+
+This connector materializes Flow collections into tables in an Amazon Redshift database.
+
+The connector uses your AWS account to materialize to Redshift tables by way of files in an S3
+bucket. The files in the bucket as as a temporary staging area for data storage and retrieval.
+
+[`ghcr.io/estuary/materialize-redshift:dev`](https://ghcr.io/estuary/materialize-redshift:dev)
+provides the latest connector image. You can also follow the link in your browser to see past image
+versions.
+
+## Prerequisites
+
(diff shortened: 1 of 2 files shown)


oliviamiannone (approved):
lgtm!

site/docs/reference/Connectors/materialization-connectors/amazon-redshift.md
```diff
@@ -0,0 +1,122 @@
...
+case-insensitive basis, as well as field names of the source collection. If any names are not unique
+on a case-insensitive basis (ex: `myField` vs. `MyField`) the materialization will fail to apply.
+
+If necessary, you can add [projections](../../../concepts/advanced/projections.md) to your
+collection specification to change field names.
+
+## Delta updates
+
+This connector supports both standard (merge) and [delta updates](../../../concepts/materialization.md#delta-updates).
+The default is to use standard updates.
```
mdibaiee, line 122: Should we include a note to this here?

> 	// If there are multiple materializations operating on different tables within the same database
	// using the same metadata table path, they will need to concurrently update the checkpoints
	// table. Aquiring a table-level lock prevents "serializable isolation violation" errors in this
	// case (they still happen even though the queries update different _rows_ in the same table),
	// although it means each materialization will have to take turns updating the checkpoints
	// table. For best performance, a single materialization per database should be used, or
	// separate materializations within the same database should use a different schema for their
	// metadata.

mdibaiee: I would also suggest including a note regarding maximum length of strings / text fields and how we work it

PR #971, opened 2023-03-14
flowctl-go: unassign failed shards when activating a publication

williamhbaker:
**Description:**

Closes https://github.com/estuary/flow/issues/970

This change will trigger an unassignment of failed shards when a publication is activated with `flowctl-go api activate`.

This is helpful for situations where a task has stopped working and a user publishes a change to fix it. Currently they will have to wait a while until the shard supervisor restarts the failed shard to see if the fix worked. With this, the shard will immediately be re-assigned so they will know right away if they task is working.

**Workflow steps:**

N/A

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/971)

### DIFF
--- a/go/flowctl-go/cmd-api-activate.go
+++ b/go/flowctl-go/cmd-api-activate.go
@@ -176,6 +176,23 @@ func (cmd apiActivate) execute(ctx context.Context) error {
 		return err
 	}
 
+	// Unassign any failed shards as a part of activating this publication.
+	if !cmd.DryRun {
+		ids := []pc.ShardID{}
+		for _, shard := range shards {
+			if shard.Upsert != nil {
+				ids = append(ids, shard.Upsert.Id)
+			}
+		}
+
+		if _, err := sc.Unassign(ctx, &pc.UnassignRequest{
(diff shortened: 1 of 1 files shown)


psFried (changes requested):
I think the code you added looks good, but I'm questioning whether it's in the right place. It seems like if any shards were already failed prior to activation, then it would still get caught in the loop above waiting for shards to become ready. It seems to me that the code you added could just be moved up to line 178 and be all good, though. LMK if I'm just missing something.

williamhbaker:
> I think the code you added looks good, but I'm questioning whether it's in the right place. It seems like if any shards were already failed prior to activation, then it would still get caught in the loop above waiting for shards to become ready. It seems to me that the code you added could just be moved up to line 178 and be all good, though. LMK if I'm just missing something.

Yeah I think that is a better place for it, good point. That way we wait for the restarted shards to be ready/fail again and don't restart them again.

I pushed a change to move it up to line 178 - Thanks!

psFried (approved):
LGTM

Issue #967, opened 2023-03-13
Connector docs link to "base" connector rather than specific connector docs for identical variants


williamhbaker:
With https://github.com/estuary/connectors/pull/479, we got the ability to easily re-tag a connector that is built from the same dockerfile for identical variants of the connector, for example `source-mysql` and `source-mariadb`. We even have separate docs for many of these variants which contain important information for them.

The connector code itself is exactly the same, even down to its reported `documentation_url` from the `spec` command. This means that the `spec` command for `source-mariadb` shows a `documentation_url` of `https://go.estuary.dev/source-mysql` instead of `https://go.estuary.dev/source-mariadb`

A consequence of this is that the `DOCS` link on the `"Create Capture"` page for these variants links to the "base" connector docs rather than the specific variant docs. And since we are now showing docs in the sidebar, this follows through to showing the docs for the "base" connector instead of the variant.

williamhbaker:
I'm thinking a way to address this may be to add a column to the `connector_tags` table called `docs_override`. If `docs_override` has a value, the agent would use that instead of the link from the connector spec for populating the `documentation_url` column.

PR #972, opened 2023-03-14
agent: add alternate documentation url column for connectors

williamhbaker:
**Description:**

Adds a new column `documentation_url_override` to the `connector_tags` table, and `agent` logic to process it.  This will allow us to set the correct documentation URL for connectors that are built as re-tags of other connectors, like `source-mariadb` vs . `source-mysql`. In these cases we would like the doc links to be for the variant rather than the connector it is re-tagged from.

Closes https://github.com/estuary/flow/issues/967

**Workflow steps:**

N/A

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/972)


williamhbaker:
Probably going in a different direction with this, so I'll close it for now.

PR #980, opened 2023-03-22
Phil/data plane auth

psFried:
**Description:**

This PR rolls up a few minor fixes related to estuary/data-plane-gateway#30. 

The main change introduced here is to use the https endpoint of data-plane-gateway when running flow locally. **This introduces a breaking change to start-flow.sh** (see below).

In addition, I'd noticed that our `gateway_auth_token` function had some weird behavior when you request scopes that you're not authorized to, so the second commit addresses that. There's still a number of other issues with how we generate these auth tokens, and I think it's best to leave those aspects unaddressed until #688 is addressed.

**Workflow steps:**

In order to get local flow installations working properly after this commit:

- Pull the latest changes from the flow, ui, and data-plane-gateway repos.
- Update the gateway URL to the HTTPS version, using one of two options:
    - Run `supabase db reset` or `supabase stop && supabase start`, and the new URL will be setup by the database seed.
    - Run the sql statement `update internal.gateway_endpoints set url = 'https://localhost:28318/';`
- Start flow using `./local/start-flow.sh` and wait for the data-plane-gateway to be up and running.
- Open your browser and navigate to (https://localhost:28318/healthz) and click through all the warnings and trust the self-signed certificate for data-plane-gateway, and you should be all good.

Trusting the TLS certificate will be a requirement from now on. Assuming you don't delete the certificates, though, you should only need to do that once per year, since that's how long the cert is valid for.

no user-facing changes here

**Notes for reviewers:** n/a

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/980)

### DIFF
--- a/local/README.md
+++ b/local/README.md
@@ -83,6 +83,14 @@ The Flow UI should open in your browser automatically.
 
 To login locally, you need to use the "magic link" method. Oauth does not work locally. You can enter any email address you like, and all emails will be sent to [http://localhost:5434/monitor]. After it says "email sent", go to that address and click the link in the email to complete the login process.
 
+### Data-plane-gateway TLS certificates
+
+Data-plane-gateway requires TLS, even locally. `start-component.sh` will generate a self-signed certificate automatically and store it in the root of the data-plane-gateway local repository.
+In order for the UI to work correctly, you'll need to tell your browser to trust that certificate. To do that, navigate to (https://localhost:28318/) and click through the warnings to trust the certificate.
+As long as you don't delete and regenerate the certificates, you should only need to do this once per year (the certificate is valid for a year).
+
+Once you trust the certificate, you should start seeing shard statuses in the UI.
+
 ### Stopping Flow
 
 - Kill the tmux session (`crtl+b` `:` then type `kill-session` and hit enter)
--- a/supabase/migrations/14_gateway_auth.sql
+++ b/supabase/migrations/14_gateway_auth.sql
@@ -48,29 +48,40 @@ returns text as $$
 
 $$ language sql stable security definer;
 
-
-
 create function gateway_auth_token(variadic prefixes text[])
 returns table (token text, gateway_url text) as $$
+declare
+  -- The number of distinct prefixes (i.e. scopes) that were requested.
+  requested_prefixes int := (select count(distinct p) from unnest(prefixes) p);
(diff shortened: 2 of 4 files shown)


local/README.md
```diff
@@ -83,6 +83,14 @@ The Flow UI should open in your browser automatically.
 
 To login locally, you need to use the "magic link" method. Oauth does not work locally. You can enter any email address you like, and all emails will be sent to [http://localhost:5434/monitor]. After it says "email sent", go to that address and click the link in the email to complete the login process.
 
+### Data-plane-gateway TLS certificates
```
jshearer, line 86: I read about this tool recently that manages creating certs and installing them in local browser stores https://github.com/FiloSottile/mkcert. Might be worth using this?

supabase/migrations/14_gateway_auth.sql
```diff
@@ -48,29 +48,38 @@ returns text as $$
...
+begin
+  
+  select array_agg(distinct p) into authorized_prefixes
+    from 
+      unnest(prefixes) as p
+      join auth_roles() as r on starts_with(p, r.role_prefix);
+
+  -- authorized_prefixes will be null when _none_ of the requested prefixes are authorized.
+  -- In that case the array_length comparison won't work, so we need an explicit null check.
+  if authorized_prefixes is null or array_length(authorized_prefixes, 1) != requested_prefixes then
```
jshearer, line 65: I'm not entirely positive I'm reading this right, but is this checking array length instead of array contents? Like, if you're authorized for X but request Y, then this won't throw so long as the number of requests and authorizations match?

jshearer (commented):
I have one question about how the authorization logic is implemented, but other than that this LGTM

supabase/migrations/14_gateway_auth.sql
psFried: Yeah, we're only checking the length of the arrays here. This works because `authorized_prefixes` is assigned by doing an inner join to the auth roles, so there's no way for it to contain prefixes that aren't a part of `prefixes`.

local/README.md
jgraettinger: The implemented solution seems very satisfactory, though, and has the benefits of simplicity and not pulling in yet another dependency to learn about and maintain.

jshearer (approved):
LGTM

PR #982, opened 2023-03-22
agent: quiet logging for gsutil

williamhbaker:
**Description:**

Runs `gsutil` with the -q option, which suppresses logs from normal operations completing successfully and instead only outputs errors if they occur.

Notably this will get rid of the `Copying file:///...` types of logs that are included at the end of the log stream when publishing tasks which should simplify interpretation of them, particularly in the case of errors with the publication.

**Workflow steps:**

Publish a task and observe the log output.

For example, prior to this change the expanded log output looks like this:

image

With this change in place the log output looks like this:

image

N/A

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/982)

### DIFF
--- a/crates/agent/src/publications/builds.rs
+++ b/crates/agent/src/publications/builds.rs
@@ -71,6 +71,7 @@ pub async fn build_catalog(
         &logs_tx,
         logs_token,
         tokio::process::Command::new("gsutil")
+            .arg("-q")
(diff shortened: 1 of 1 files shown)


psFried (approved):
LGTM

PR #984, opened 2023-03-24
Add longform connector description

jshearer:
This is to show on the marketing site connector detail pages

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/984)

### DIFF
--- a/diff.sql
+++ b/diff.sql
@@ -1,0 +1,5 @@
+ALTER TABLE IF EXISTS public.connectors
+    ADD COLUMN long_description jsonb_internationalized_value;
+
+COMMENT ON COLUMN public.connectors.long_description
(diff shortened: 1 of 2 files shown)


jgraettinger:
Is this still needed? Was it applied and this is a post-migration review?

Having yet another column for this is a bit smelly. Another option could be using a paragraph break or something, but if this is already in use then :shrug:

diff.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE IF EXISTS public.connectors
```
jgraettinger, line 1: This should be renamed, and under `supabase/pending/`, if it's an unapplied migration.

PR #989, opened 2023-03-29
Add Tilt for running Flow locally

psFried:
**Description:**

This adds a new way of running flow locally, as an alternative to `start-flow.sh`. This new method uses [Tilt](https://docs.tilt.dev/), which provides a way to orchestrate the startup of the various components. With the exception of supabase, the components are all run as direct binary invocations, not as docker images.

The motivation for this was to provide a way to have data-plane state persist across restarts when runnign locally. Unlike `start-flow.sh`, the Tiltfile does not use `temp-data-plane`, and instead runs etcd, gazette, and reactor as separate components.

This initial implementation probably still has some issues, especially for mac users. I left `start-flow.sh` in place for the time being, to ensure we can get the bugs worked out and give people time to transition.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/989)

### DIFF
--- a/Tiltfile
+++ b/Tiltfile
@@ -1,0 +1,112 @@
+# This file is interpreted by `tilt`, and describes how to get a local flow environment running.
+DATABASE_URL="postgresql://postgres:postgres@localhost:5432/postgres"
+os.putenv("DATABASE_URL", DATABASE_URL)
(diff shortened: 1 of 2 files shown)


jgraettinger (approved):
LGTM  - excited to try it

PR #990, opened 2023-03-30
validation: better error message when names collide

mdibaiee:
**Description:**

- Question: is it fair to suggest renaming `lhs_name` in the case of `NameCollision`? Is the left-hand side always the "new" thing? In case of prefixes it is, but I'm not sure if in caes of complete collision it also is?

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/990)

### DIFF
--- a/crates/validation/src/errors.rs
+++ b/crates/validation/src/errors.rs
@@ -14,10 +14,19 @@ pub enum Error {
         unmatched: String,
     },
     #[error(
-        "{lhs_entity} {lhs_name} {error_class} {rhs_entity} {rhs_name}, defined at {rhs_scope}"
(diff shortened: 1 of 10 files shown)


jgraettinger (commented):
Here's the rub: at this point, we don't have _any idea_ if renaming or adding a suffix to the collection is actually something that the user can even do. We don't know whether the LHS or RHS entities have already been published, or drafted, or what.

I understand and share the desire to make the error actionable, but there's a large potential to given the user an "action" that they can't actually do, and for that reason I think it's better to leave as-is.

PR #994, opened 2023-04-06
flowctl: improve `preview` command

jgraettinger:
**Description:**

Add `--interval `for a configurable flush interval.

Add `--infer-schema` to have preview infer the collection's JSON schema
and update it upon exit.

Make the `--collection` flag optional, and don't require it if there's only one collection in the `--source` specs.

Not specific to preview, but: stop filtering out "Not Found" errors.
We now resolve these dynamically from the control plane, so if it's not found, it really doesn't exist.

**Workflow steps:**

Pass new arguments to `flowctl preview`

None yet.

**Notes for reviewers:**

Tested manually.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/994)


jgraettinger:
I'm merging now to have this available for later today. Please do a post-merge review.

PR #997, opened 2023-04-10
use local flowctl profile with ops-publication.sh

psFried:
**Description:**

Explicitly uses the local profile in `ops-publication.sh`, since the `FLOWCTL_PROFILE` variable is not set when running with `tilt`. The is important because flowctl requires a valid login in order to run `raw bundle`, and it's super confusing to have it try to login to production in order to bundle specs for a local instance.

**Workflow steps:**

This is for devs, not end users. Start flow locally using `tilt up`, then run the `ops-publication.sh` script.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/997)

### DIFF
--- a/local/ops-publication.sh
+++ b/local/ops-publication.sh
@@ -4,7 +4,7 @@ set -o errexit
 set -o pipefail
 set -o nounset
 
-bundled_ops_catalog="$(.build/package/bin/flowctl raw bundle --source ops-catalog/template-local.flow.yaml | sed "s/'/''/g")"
+bundled_ops_catalog="$(.build/package/bin/flowctl --profile local raw bundle --source ops-catalog/template-local.flow.yaml | sed "s/'/''/g")"
 
 cat << EOF
 begin;


kiahna-tucker (approved):
This change looks innocuous but I do not have the ability to evaluate its effects given I cannot spin up a local control plane.

jgraettinger:
I'm uncertain about this one.

What might be better, is to defer constructing the client until we're _certain_ we actually need to call out to the control plane. We don't actually need to, in this case, since all collections are provided inline.

Otherwise there's a bit of a chicken / egg dance, where you're trying to bootstrap a local environment using a local environment. It definitely _can_ work but is a little surprising and not clearly "better" than "just log in to the production dashboard"

psFried:
@jgraettinger I'm on board with the idea of changing `flowctl` to create the client lazily.

> Otherwise there's a bit of a chicken / egg dance, where you're trying to bootstrap a local environment using a local environment. It definitely can work but is a little surprising and not clearly "better" than "just log in to the production dashboard"

We already have (had?) this situation in `start-flow.sh`. The script exports `FLOWCTL_PROFILE=local`, so it was already using the local profile, and running it before we startup the agent. I'm honestly not sure why this _hasn't_ been causing a bunch of failed startups :thinking: I think it may just be the fact that an expired/invalid token still "works" for the purposes of being able to run `raw bundle`, and perhaps everyone's just always had a pre-existing `local.json` file. :shrug: 

It's less of an issue with Tilt, since there a failure in `ops-catalog.sh` doesn't prevent startup of any other components like it does in `start-flow.sh`. I'd go either way on whether to merge this PR, but leaning toward merging. Either way I feel like it's gonna be a little weird unless/until we change flowctl to create the client lazily.

PR #1011, opened 2023-04-20
docs: rework documentation for derivations

jgraettinger:
Update and refresh the acmeBank examples used within derivation documentation.

Include documentation acmeBank examples in our catalog tests.

Largely re-write the "Derivations" documentation page.

**Workflow steps:**

None.

Significant updates to `/concepts/derivations` page. 

**Notes for reviewers:**

Lemme know if it's useful to VC and discuss structure and explanations, etc.
At this point I believe everything is technically accurate and complete.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1011)


site/docs/concepts/derivations.md
```diff
@@ -7,700 +7,769 @@ import Mermaid from '@theme/Mermaid';
...
-* In many cases, **lambdas**, functions defined in accompanying TypeScript modules
-that allow more complex transformations.
+## Tutorial
 
-![](<derivations-new.svg>)
+### Introducing AcmeBank
 
-## Creating derivations
+The following tutorial sections will refer to the following common example
+to illustrate concepts.
```
oliviamiannone, line 38: suggested reword (to set context and ensure readers don't over-agonize about understanding EVERYTHING from examples alone): 
The following tutorial sections use an illustrative example to introduce you to derivations, how you might use them, and their common components. We'll discuss each component in depth in subsequent sections of this page, but we recommend you start here to get your bearings.

site/docs/concepts/derivations.md
```diff
@@ -7,700 +7,769 @@ import Mermaid from '@theme/Mermaid';
...
+### State
+
+The abstract `IDerivation` class generated within the interfaces module
+includes additional, experimental methods which can be used for
+persisting and recovering internal state of the connector.
+
+Consult the generated implementation and feel free to reach out to support
+if you'd like more information on building stateful TypeScript derivations.
+
+## Specification
```
oliviamiannone, line 546: I think this section could be more effective *before* the **Connectors** heading. Then it’s available just after the tutorial as reference, and gives us a high-level overview before we go into detail in the next few sections.

site/docs/concepts/derivations.md
```diff
@@ -7,700 +7,769 @@ import Mermaid from '@theme/Mermaid';
...
 
-To create a new TypeScript module for the lambdas of your derivation,
-you can use `flowctl typescript generate` to generate it.
-In the derivation specification, choose the name for the new module and
-run `flowctl typescript generate`.
-Flow creates a module with the name you specified, stubs of the required interfaces,
-and TypeScript types that match your schemas.
-Update the module with your lambda function bodies,
-and proceed to test and deploy your catalog.
+Collectively, your migrations serve as the [internal state](#internal-state) of your derivation.
```
oliviamiannone, line 477: On first read, I got a little hung up on the fact that these are called 'migrations.' I see how they're different than any old SQL queries, and they're also separate from the concept of internal state more generally (because they're not in TS derivations). But still, once I got down to the "Internal State" section, I was like: wait, I just got used to this new term, "Migration," and now it's not being used anymore. So a bit of cognitive friction. 

Unfortunately I don't have a solution to offer, but I wanted to throw this in anyway.

site/docs/concepts/derivations.md
```diff
@@ -7,700 +7,769 @@ import Mermaid from '@theme/Mermaid';
...
-Register schemas may also have
-[reduction](schemas.md#reductions) annotations,
-and documents returned by update lambdas
-are _reduced_ into the current register value.
+[See "Where to Accumulate?" for more discussion](#where-to-accumulate).
 
-The compliance department reached out again,
-and this time they need you to identify transfers where
-the sender's account had insufficient funds.
+## Connectors
```
oliviamiannone, line 329: I think it would be better to call this section "Supported Languages" or something similar. Although from a Flow architecture viewpoint these *are* connectors, from a user viewpoint, I'm used to connectors as how I integrate with my source and destination. And here what I'm wondering about is *what languages can I use to write transformations?*

You could keep the next paragraph as-is (still explain that they are technically connectors and work as such) but then the next sentence could change to "At present, Flow supports transformations in SQL using SQLite, and TypeScript."

oliviamiannone (approved):
In general, this looks really great. I specifically really like how you relate throughout to how things work in different streaming systems the reader might be familar with. I added a few thoughts on how concepts are introduced and how items were named, but nothing I feel extremely strongly about. A few other general notes:

* I follow all the material here to the extent that I personally expect to as a non-technical person. I’m *assuming* that with the background of most users who’d be attempting this, enough context is given for them to get started. However, I still think someone else who’s more technical should put on their new-user hat and give this a read. 
* Should be a separate PR, but wanted to call out that the [tutorial to create a derivation](https://docs.estuary.dev/guides/flowctl/create-derivation/) will also need to be updated to include these changes + the addition of GitPod. That’s a great chance to dive into the simplified, practical steps of adding a derivation to an existing collection, which isn’t explicitly covered here.

site/docs/concepts/derivations.md
jgraettinger: Done 👍

site/docs/concepts/derivations.md
jgraettinger: Done 👍 

I was unsure if this was a bit too reference-y, which is why I moved it down, but if you think it fits here then I'm definitely sold.

site/docs/concepts/derivations.md
jgraettinger: I'd been using `migrations` as a kind-of short hand for "The SQLite database schema you create that is the internal state of your derivation". I took a pass and tried to fully unpack it throughout, explicitly connecting that the schema _created by_ migrations _is_ the internal task state.

site/docs/concepts/derivations.md
jgraettinger: Done :+1:

Agreed that heading better characterizes what the user is after.
Architecturally they _are_ connectors, and I didn't want to shy away from that, because in the future we may offer packaged-up use cases which are deployed as docker images / configured in the UI in the same way we do with captures and materializations. But we're not there yet, so no reason to sweat it, and even if we were "Supported Languages" is pretty reasonable.

jgraettinger:
Thanks for the review! Your feedback was not controversial. I took a pass and found some other cleanups / improvements to apply. Gonna go ahead and merge once it passes CI.

PR #1013, opened 2023-04-20
Feature/stripe backend

jshearer:
**Description:**

Implements the backend for billing details using stripe's API:

* Create `SetupIntent`s that validate and store a payment method for future use
  * When a tenant first tries to create a `SetupIntent`, we also create a `Customer` in stripe for that tenant
* List payment methods
* Delete a payment method
* Switch your primary payment method

There is also basic authorization to ensure that the user making the network request has `admin` on the tenant they're making the request for. In the future we should probably have a separate `billing` role, or something like that

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1013)


supabase/functions/billing/index.ts
```diff
@@ -0,0 +1,74 @@
...
+import { getTenantPaymentMethods } from "./get_tenant_payment_methods.ts";
+import { deleteTenantPaymentMethod } from "./delete_tenant_payment_method.ts";
+import { setTenantPrimaryPaymentMethod } from "./set_tenant_primary_payment_method.ts";
+import { createClient } from "https://esm.sh/@supabase/supabase-js@2.0.5";
+
+serve(async (req) => {
+    let res: ConstructorParameters<typeof Response> = [null, {}];
+    try {
+        if (req.method === "OPTIONS") {
+            res = ["ok", { status: 200 }];
```
psFried, line 13: This cors handling threw me for a bit of a loop because I'd expected to see the cors headers set on this line. Is there a reason that we need to add the cors headers to _every_ response down on line 69?

supabase/functions/billing/index.ts
```diff
@@ -0,0 +1,74 @@
...
+    } catch (e) {
+        res = [JSON.stringify({ error: e.message }), {
+            headers: { "Content-Type": "application/json" },
+            status: 400,
+        }];
+    }
+
+    res[1] = { ...res[1], headers: { ...res[1]?.headers || {}, ...corsHeaders } };
+
+    console.log(JSON.stringify(res, null, 4));
```
psFried, line 71: was this vestigial, or still useful?

psFried (approved):
LTGM

supabase/functions/billing/index.ts
jshearer: Vestigial, removed!

supabase/functions/billing/index.ts
jshearer: Looks like the headers need to be on both the preflight response and all actual responses: https://stackoverflow.com/questions/24264574/cors-headers-present-only-on-preflight-or-every-request

PR #1014, opened 2023-04-21
flowctl: use basic stdin readline for auth tokens

psFried:
**Description:**

Removes the use of `rpassword` crate for reading auth tokens from stdin. Turns out that OSX has a feature called "secure input", which can prevent users from pasting into terminals. This post seems to describe more or less what's going on:
https://fig.io/docs/support/secure-keyboard-input#conflicts-with-password-managers-1password-bitwarden It's of course still possible that users may still have trouble pasting into iterm in some scenarios due to secure input. But removing `rpassword` seems likely to help in most cases.

**Workflow steps:**

run `flowctl auth login`. Previously, a secure input was used for the token, so it didn't show up in plain text. Now, it just shows up in plain text.

**Notes for reviewers:** n/a

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1014)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1434,7 +1434,6 @@ dependencies = [
  "proto-flow",
  "proto-gazette",
  "reqwest",
- "rpassword",
  "runtime",
  "schema-inference",
  "serde",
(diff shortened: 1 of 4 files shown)


jgraettinger (approved):
LGTM

PR #1015, opened 2023-04-21
supabase: add billing_report RPC

jgraettinger:
Add a `billing_report` function / RPC which is intended as the canonical place where we compute billing, both for month-to-date figures as well as for final billing at the close of a month.

The RPC is called with arguments like:
```json
  {"billed_prefix": "acmeCo/", "billed_month": "2023-04-01T00:00:00Z"}
```

In response, you get a report like:
```json
  {
    "subtotal": 2484,
    "line_items": [
      {
        "rate": 0,
        "count": 2,
        "subtotal": 0,
        "description": "Included task shards (up to 2)"
      },
      {
        "rate": 2000,
        "count": 1,
        "subtotal": 2000,
        "description": "Additional task shards minimum (assessed at 2022-08-29 13:00:00+00)"
      },
      {
        "rate": 0,
        "count": 10,
        "subtotal": 0,
        "description": "Included data processing (in GB, up to 10.0GB)"
      },
      {
        "rate": 75,
        "count": 33.125,
        "subtotal": 2484,
        "description": "Additional data processing (in GB)"
      },
      {
        "description": "Subtotal is greater of task shards minimum, or data processing volume"
      }
    ],
    "billed_month": "2022-08-01T00:00:00+00:00",
    "billed_prefix": "aliceCo/"
  }
  ```

`rate` and `subtotal` fields are in cents.

It's expected that the _how_ of this function will evolve over time. The _hope_ is that its arguments, and the generalized shape of the billing report should not.

Issue #983

**Workflow steps:**

Call the `billing_report` RPC with suitable arguments.

None.

**Notes for reviewers:**

Please especially consider whether the arguments, or the shape of the billing report seems sound to you.
Some of the details here are certain to change. For one, I'm not happy with the way in which the maximum number of task shards is calculated, as it's a false-negative undercount.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1015)

### CODE REVIEW

**Comment by kiahna-tucker** [comment]
After working with the RPC, a few questions come to mind:

1. What was the motivation behind having `line_items` be an array (as opposed to an object with keys providing context for each line item)?

1. Could all pricing tier constraints be accessible via key-value pairs in a _line item_ object? Ideally, the UI would not surface any hardcoded constraints and, at the moment, the included task and GB count are not supplied in a consumable fashion.

1. As of today, this RPC will drive views which supply multiple months' worth of billing records (i.e., the billing history table and data volume by month graph). Consequently, I was curious if it would be possible for the RPC to return all the records within a specified interval.

### References

**Sample Response | Tenant with task count and total data volume beneath free threshold**

```json
{
    "subtotal": 0,
    "line_items": [
        {
            "rate": 0,
            "count": 1,
            "subtotal": 0,
            "description": "Included task shards (up to 2)"
        },
        {
            "rate": 2000,
            "count": 0,
            "subtotal": 0,
            "description": "Additional task shards minimum (assessed at 2023-04-24 20:00:00+00)"
        },
        {
            "rate": 0,
            "count": 0.0017,
            "subtotal": 0,
            "description": "Included data processing (in GB, up to 10.0GB)"
        },
        {
            "rate": 75,
            "count": 0,
            "subtotal": 0,
            "description": "Additional data processing (in GB)"
        },
        {
            "description": "Subtotal is greater of task shards minimum, or data processing volume"
        }
    ],
    "billed_month": "2023-04-01T00:00:00+00:00",
    "billed_prefix": "melk/"
}
```

kiahna-tucker (commented):
See detailed comment directly above.

jgraettinger:
> 1. What was the motivation behind having `line_items` be an array (as opposed to an object with keys providing context for each line item)?

Because line items are ordered, and express a narrative structure as to how the bill was calculated.

> 2. Could all pricing tier constraints be accessible via key-value pairs in a _line item_ object? Ideally, the UI would not surface any hardcoded constraints and, at the moment, the included task and GB count are not supplied in a consumable fashion.

Can you expand on exactly how these would be used in the UI? The intention is that the UI wouldn't need it. That's probably naive but I'd definitely like to use this to highlight and talk about any places where the UI needs more info, and how that affects presentation.

> 3. As of today, this RPC will drive views which supply multiple months' worth of billing records (i.e., the billing history table and data volume by month graph). Consequently, I was curious if it would be possible for the RPC to return all the records within a specified interval.

The plan, not implemented yet with this PR, is that we'll have a new table which is the snapshot history of finalized billing at the close of each month. _That's_ what the UI should query and present for prior-month history. This RPC would only be used for current-month.

kiahna-tucker:
> The plan, not implemented yet with this PR, is that we'll have a new table which is the snapshot history of finalized billing at the close of each month. That's what the UI should query and present for prior-month history. This RPC would only be used for current-month.

I have always anticipated a table, but was under the impression that the RPC was a temporary stand-in for that work, aiding the UI in calculating total data volume and cost. Interesting. If the intent of the RPC is only to provide details for the current month, I do not see its use in the UI as it stands today. Where were you expecting this to be consumed? Was the request to introduce something new to the UI that would show a breakdown for the current month?

jgraettinger:
Portions of the UI that are currently calculating anything money related, should _not_ live in the UI, and should instead present the results of this function without any modification. This RPC is the **one and only** place where the business logic will live, that maps task / GB usage to actual dollar amounts shown to users.

We talked through this most recently last Wednesday. And I've been pretty consistent before that, that the UI is not the place to calculate anything regarding cost, which is the whole rationale for this function.

kiahna-tucker:
> Portions of the UI that are currently calculating anything money related, should not live in the UI, and should instead present the results of this function without any modification. This RPC is the one and only place where the business logic will live, that maps task / GB usage to actual dollar amounts shown to users.
> 
> We talked through this most recently last Wednesday. And I've been pretty consistent before that, that the UI is not the place to calculate anything regarding cost, which is the whole rationale for this function.
> 

Not once did I contradict anything that you’ve said in this message. Consider the RPC working as intended.

kiahna-tucker (approved):

jgraettinger:
Apologies if I came off as sharp. Part of the confusion is (I think?) how the mocks are structured. As currently mocked, the table would be stitched from current-month totals (coming from this RPC) and prior month reconciled billing from a table that doesn't exist yet.

I missed that in my own read of the mocks, and believed that table to be a current-month-only detail, with historical billing in a separate section.

Either way, the calculation of cost should be done in this RPC and surfaced in the UI.

We could use more detail here, though. Lemme have a conversation with @dyaffe and see if there are adjustments to be made to the mocks.

PR #1016, opened 2023-04-22
oauth: fix default content-type not being overriden sometimes

mdibaiee:
**Description:**

- if a connector specified the content-type header using a different casing than `content-type`, the default content-type would not be overriden and cause issues.
- with this fix, we check for case-insensitive presence of that header and set the default only if the header is not present in the connector's oauth2 spec

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1016)


williamhbaker (approved):
LGTM

PR #1019, opened 2023-04-24
oauth: fix type-check error in access-token edge function

jgraettinger:
Fix a type-check that currently fails on a local stack (and would fail if applied in prod?).

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1019)

### DIFF
--- a/supabase/functions/oauth/access-token.ts
+++ b/supabase/functions/oauth/access-token.ts
@@ -1,5 +1,3 @@
-import { serve } from 'https://deno.land/std@0.131.0/http/server.ts';
-import { createClient } from 'https://esm.sh/@supabase/supabase-js';
 import jsonpointer from 'https://esm.sh/jsonpointer.js';
 import { returnPostgresError, compileTemplate } from '../_shared/helpers.ts';
 import { corsHeaders } from '../_shared/cors.ts';
@@ -81,7 +79,8 @@ export async function accessToken(req: Record<string, any>) {
         );
     }
 
-    const defaultContentType = Object.keys(headers).some(h => h.toLowerCase() == 'content-type') ? {} : { 'content-type': 'application/json' };
+    const defaultContentType: Record<string, string> = Object.keys(headers).some(
+        h => h.toLowerCase() == 'content-type') ? {} : { 'content-type': 'application/json' };
 
     const response = await fetch(url, {
         method: 'POST',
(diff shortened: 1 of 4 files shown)


jgraettinger:
To give more backround, here's the specific error (encountered on a local stack with Tilt):

```
Running cmd: cd /home/johnny/estuary/flow/../flow && supabase functions serve oauth
Starting supabase/functions/oauth
Serving supabase/functions/oauth
Watcher Process started.
Check file:///home/deno/functions/oauth/index.ts
error: TS2322 [ERROR]: Type '{ "Access-Control-Allow-Origin": string; "Access-Control-Allow-Headers": string; 'content-type'?: undefined; accept: string; } | { "Access-Control-Allow-Origin": string; "Access-Control-Allow-Headers": string; 'content-type': string; accept: string; }' is not assignable to type 'HeadersInit | undefined'.
  Type '{ "Access-Control-Allow-Origin": string; "Access-Control-Allow-Headers": string; 'content-type'?: undefined; accept: string; }' is not assignable to type 'HeadersInit | undefined'.
    Type '{ "Access-Control-Allow-Origin": string; "Access-Control-Allow-Headers": string; 'content-type'?: undefined; accept: string; }' is not assignable to type 'undefined'.
        headers: {
        ~~~~~~~
    at file:///home/deno/functions/oauth/access-token.ts:89:9
 ```

File: supabase/functions/oauth/access-token.ts [hunk 1]
```diff
@@ -81,7 +79,7 @@ export async function accessToken(req: Record<string, any>) {
         );
     }
 
-    const defaultContentType = Object.keys(headers).some(h => h.toLowerCase() == 'content-type') ? {} : { 'content-type': 'application/json' };
+    const defaultContentType = Object.keys(headers).every(h => h.toLowerCase() != 'content-type') && { 'content-type': 'application/json' };
```
mdibaiee, line 82: Just because this reads more like the sentence: "If there is already a `content-type` header, we don't have a default, otherwise we have a default.

Instead of this sentence: If all of the keys in the headers are _not_ `content-type`, then we add a default `content-type`.

I find the former sentence easier to follow.

```suggestion
    const defaultContentType: Record<string, string> = Object.keys(headers).some(h => h.toLowerCase() == 'content-type') ? {} : { 'content-type': 'application/json' };
```

jgraettinger: Done

jgraettinger:
I've also added a related fix for the Tilt setup, that causes it to use the local config-encryption service (with development keys). Tested locally 👍

mdibaiee (approved):

PR #1023, opened 2023-05-01
Fix `kill_on_drop`

jshearer:
**Description:**

Ultimately, the problem was that `Child::wait()` had to take ownership of `Child::inner`, because [both `wait()` and `kill()` take `&mut self`](https://stackoverflow.com/q/35093869), meaning that if you've called `wait()`, you can't also call `kill()`. 

The solution here was to introduce [`shared_child`](https://docs.rs/shared_child/latest/shared_child/), which lifts that limitation, allowing the `impl Drop` handler to successfully `SIGKILL` the running process, and clean up properly.

**Note:** Because this is in a drop handler, we're `SIGKILL`ing the `temp-data-plane` instead of `SIGTERM`ing it. On linux, we set a flag to instruct the kernel to kill all of `temp-data-plane`'s child processes when it is killed, but we either can't or don't set that flag on MacOS. As a result we appear to still leak a bunch of `flowctl-go serve consumer` processes on Mac. This isn't a problem on Linux, so we decided to leave it be for the moment.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1023)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -194,7 +194,9 @@ dependencies = [
 name = "async-process"
 version = "0.0.0"
 dependencies = [
+ "shared_child",
  "tokio",
+ "tracing",
 ]
 
 [[package]]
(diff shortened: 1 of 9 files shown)


jgraettinger (approved):
LGTM

PR #1026, opened 2023-05-02
docs: don't recommend 'for all tables' publication for source-postgres

williamhbaker:
**Description:**

Doc updates for https://github.com/estuary/connectors/pull/690. See that PR description for the technical background.

For manual setups these updated docs will instruct the user to create a targeted publication including only the tables they want to capture from rather than a "FOR ALL TABLES" publication which may have unwanted side effects.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1026)

### DIFF
--- a/site/docs/reference/Connectors/capture-connectors/PostgreSQL.md
+++ b/site/docs/reference/Connectors/capture-connectors/PostgreSQL.md
@@ -64,11 +64,11 @@ CREATE USER flow_capture WITH PASSWORD 'secret' REPLICATION;
 
     ```sql
     ALTER DEFAULT PRIVILEGES IN SCHEMA public GRANT SELECT ON TABLES to flow_capture;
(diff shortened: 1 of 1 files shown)


willdonnelly (approved):
LGTM

I think that it might still be worth noting _somewhere_ on the page that the user can `CREATE PUBLICATION ... FOR ALL TABLES` in some circumstances, because adding tables individually can be a huge permissioning-related pain in some setups (AFAIK the query `ALTER PUBLICATION ... ADD TABLE ...` must be run either by a user with ownership of both the table and the publication, or by the database superuser, and transferring ownership after a table has been created is just a mess).

But on reflection I agree with Johnny's point about automatically breaking your database -- for the recommended sequences of copy-pastable setup commands here it's probably better to recommend the less dangerous approach, since either will work perfectly fine if you're doing this setup as the database superuser.

oliviamiannone (approved):
lgtm from the writing side!

PR #1028, opened 2023-05-04
plumbing improvements for connector errors

jgraettinger:
**Description:**

 * Previously, `flow-connector-init` would collect a ring-buffer of recent stderr output and build a terminal error out of that unmodified buffer. This does pass through a comprehensive detailing of what went wrong, but it's hard to understand and it's also duplicative with ops logs and the control-plane `internal.log_lines` table. So, stop doing that and instead:
   * Build an error from the message of the very last ops::Log, and
   * Attach the entire final ops::Log to the gRPC error response. We plan to use this in the future but not yet.
 
 * Update discovery within the `agent` to properly populate the `draft_errors` table with a draft error for display when discovery fails.
 
 * Generally audit, tweak, and make errors more succinct for users.
 
 See screenshots posted in Slack for the end-to-end UX of errors in the UI.
 This depends on work within the UI as well, to better display errors.
 
**Workflow steps:**

No new steps. Just observe better structured draft errors in the UI.

None I'm aware of.

**Notes for reviewers:**

Probably best to review commit by commit.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1028)

### DIFF
--- a/crates/ops/src/lib.rs
+++ b/crates/ops/src/lib.rs
@@ -67,7 +67,7 @@ impl<'de> Deserialize<'de> for HexU32 {
 
 /// stderr_log_handler is a log handler that writes canonical
 /// JSON log serializations to stderr.
-pub fn stderr_log_handler(log: Log) {
+pub fn stderr_log_handler(log: &Log) {
     let mut buf = serde_json::to_vec(&log).expect("Log always serializes");
     buf.push(b'\n');
     _ = std::io::stderr().write_all(&buf); // Best-effort.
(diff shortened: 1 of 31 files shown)


crates/ops/src/lib.rs
```diff
@@ -67,7 +67,7 @@ impl<'de> Deserialize<'de> for HexU32 {
 
 /// stderr_log_handler is a log handler that writes canonical
 /// JSON log serializations to stderr.
-pub fn stderr_log_handler(log: Log) {
+pub fn stderr_log_handler(log: &Log) {
     let mut buf = serde_json::to_vec(&log).expect("Log always serializes");
```
williamhbaker, line 71: Nit: I don't think we need the `&` here anymore

williamhbaker (approved):
LGTM

PR #1032, opened 2023-05-09
Phil/re create collections

psFried:
**Description:**

Allows users to re-create Flow collections with new names, automatically updating materializations to read from the new collections. Individual commit messages offer more detail, but the main pieces in play here are:

- Publish now includes an `incompatible_collections` array suggesting to re-create any collections which have been rejected by a materialization validation response.
- A new `evolutions` async job that updates a draft to re-name a set of collections, including updates to any connected captures or materializations. Derivations are excluded from this for now.
- The UI will now look for the presence of `incompatible_collections` and offer to update them by running an `evolutions` job. This is implemented in estuary/ui#597

**Workflow steps:**

- Edit a capture, and update one or more collections to have a schema that is incompatible with an existing materialization.
- Click "Save and Publish". You should see a build error, along with the (new) dialog offering to re-create the incompatible collections.
- Click "Create new collection versions" and wait for the job to finish
- Now, "Save and Publish" again, and it should succeed. Any materializations of affected collections will re-start from the beginning for just those bindings.

We don't yet have docs on this, but I'm thinking that it'd be good to add a "Schema evolution in Flow" doc at some point in the near future.

**Notes for reviewers:**

For now, we rely only on materialization validation responses to identify incompatible schema changes. This is intentionally pretty minimal. A potential future improvement would be to validate whether any fields have had their JSON schema `type` change to a different and non-overlapping set of types. I intentionally left that out of this PR to limit scope.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1032)

### DIFF
--- a/crates/agent-sql/src/evolutions.rs
+++ b/crates/agent-sql/src/evolutions.rs
@@ -1,0 +1,133 @@
+use crate::{CatalogType, Id, TextJson as Json};
+use chrono::prelude::*;
+use serde::Serialize;
+use serde_json::value::RawValue;
+use sqlx::types::Uuid;
+
+// Row is the dequeued task of an evolution operation.
+#[derive(Debug)]
+pub struct Row {
+    pub id: Id,
+    pub created_at: DateTime<Utc>,
+    pub detail: Option<String>,
+    pub draft_id: Id,
+    pub logs_token: Uuid,
+    pub updated_at: DateTime<Utc>,
+    pub user_id: Uuid,
+    pub collections: Json<Box<RawValue>>,
+}
+
+pub async fn dequeue(txn: &mut sqlx::Transaction<'_, sqlx::Postgres>) -> sqlx::Result<Option<Row>> {
+    sqlx::query_as!(
+        Row,
+        r#"select
+            id as "id: Id",
+            created_at,
+            detail,
+            draft_id as "draft_id: Id",
+            logs_token,
+            updated_at,
+            user_id,
+            collections as "collections: Json<Box<RawValue>>"
+        from evolutions where job_status->>'type' = 'queued'
+        order by id asc
+        limit 1
+        for update of evolutions skip locked;
+        "#
+    )
+    .fetch_optional(txn)
+    .await
+}
+
+pub async fn resolve<S>(
+    id: Id,
+    status: &S,
+    txn: &mut sqlx::Transaction<'_, sqlx::Postgres>,
+) -> sqlx::Result<()>
+where
+    S: Serialize + Send + Sync,
+{
+    sqlx::query!(
+        r#"update evolutions set
+            job_status = $2,
+            updated_at = clock_timestamp()
+        where id = $1
+        returning 1 as "must_exist";
+        "#,
+        id as Id,
+        Json(status) as Json<&S>,
+    )
+    .fetch_one(txn)
+    .await?;
+
+    Ok(())
+}
+
+pub struct DraftSpecRow {
+    pub draft_spec_id: Id,
+    pub live_spec_id: Option<Id>,
+    pub catalog_name: String,
+    pub expect_pub_id: Option<Id>,
+    pub last_pub_id: Option<Id>,
+    pub draft_spec: Option<Json<Box<RawValue>>>,
+    pub draft_type: Option<CatalogType>,
+    pub live_type: Option<CatalogType>,
+}
+
+pub async fn fetch_draft_specs(
+    draft_id: Id,
+    user_id: Uuid,
+    txn: &mut sqlx::Transaction<'_, sqlx::Postgres>,
+) -> sqlx::Result<Vec<DraftSpecRow>> {
+    sqlx::query_as!(
+        DraftSpecRow,
+        r#"
+        select
+            draft_specs.id as "draft_spec_id: Id",
+            draft_specs.catalog_name,
+            draft_specs.expect_pub_id as "expect_pub_id: Id",
+            draft_specs.spec as "draft_spec: Json<Box<RawValue>>",
+            draft_specs.spec_type as "draft_type: CatalogType",
+            live_specs.spec_type as "live_type: CatalogType",
+            live_specs.last_pub_id as "last_pub_id: Option<Id>",
+            live_specs.id as "live_spec_id: Option<Id>"
+        from drafts
+        left join draft_specs on drafts.id = draft_specs.draft_id
+        left join live_specs
+            on draft_specs.catalog_name = live_specs.catalog_name
+            -- Ensure that `live_spec_id` is `None` if the user does not have admin access to the spec.
+            and exists (select 1 from internal.user_roles($2, 'admin') r where live_specs.catalog_name ^@ r.role_prefix)
+        where drafts.id = $1 and drafts.user_id = $2
+        "#,
+        draft_id as Id,
+        user_id as Uuid,
+    )
+    .fetch_all(txn)
+    .await
+}
+
+pub async fn fetch_resource_spec_schema(
+    image_name: String,
+    image_tag: String,
+    txn: &mut sqlx::Transaction<'_, sqlx::Postgres>,
+) -> sqlx::Result<Option<Json<Box<RawValue>>>> {
+    struct ResourceSpecSchema {
+        resource_spec_schema: Option<Json<Box<RawValue>>>,
+    }
+    let res = sqlx::query_as!(
+        ResourceSpecSchema,
+        r#"
+        select connector_tags.resource_spec_schema as "resource_spec_schema: Json<Box<RawValue>>"
+        from connectors
+            join connector_tags on connectors.id = connector_tags.connector_id
+        where connectors.image_name = $1
+            and connector_tags.image_tag = $2
+        "#,
+        image_name,
+        image_tag
+    )
+    .fetch_optional(txn)
+    .await?;
+
+    Ok(res.and_then(|r| r.resource_spec_schema))
+}
--- a/crates/agent/src/evolution.rs
+++ b/crates/agent/src/evolution.rs
@@ -1,0 +1,517 @@
+use super::{draft, Handler, HandlerStatus, Id};
+use agent_sql::{
+    evolutions::{DraftSpecRow, Row},
+    Capability,
+};
+use anyhow::Context;
+use itertools::Itertools;
+use lazy_static::lazy_static;
+use regex::Regex;
+use serde::{Deserialize, Serialize};
+use serde_json::Value;
+use std::collections::{BTreeMap, BTreeSet, HashMap};
+
+#[cfg(test)]
+mod test;
+
+pub struct EvolutionHandler;
+
+/// Rust struct corresponding to each array element of the `collections` JSON
+/// input of an `evolutions` row.
+#[derive(Serialize, Deserialize, Clone, PartialEq)]
+struct RenameRequest {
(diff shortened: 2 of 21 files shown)


crates/agent/src/evolution.rs
```diff
@@ -0,0 +1,529 @@
...
+        if row.live_spec_id.is_none() {
+            anyhow::bail!(
+                "cannot re-create collection '{old_name}' because it has never been published"
+            )
+        }
+
+        // Technically, these checks duplicate validations that would already be performed if the
+        // user tries to publish the draft after re-creating the collections. But this allows us
+        // to surface the error more quickly.
+        match row.live_type {
```
jgraettinger, line 241: noting this as a caution area, where we could inadvertently leak information about a spec that the user isn't entitled to see. especially if the "you're a draft admin" check is relaxed above. I would probably seek to filter joined live specs at query time.

crates/agent-sql/src/evolutions.rs
```diff
@@ -0,0 +1,183 @@
...
+            draft_specs.spec_type as "draft_type: CatalogType",
+            live_specs.spec_type as "live_type: CatalogType",
+            live_specs.last_pub_id as "last_pub_id: Option<Id>",
+            live_specs.id as "live_spec_id: Option<Id>",
+            (
+                select max(capability) from internal.user_roles($2) r
+                where starts_with(draft_specs.catalog_name, r.role_prefix)
+            ) as "user_capability: Capability"
+        from draft_specs
+        left join live_specs
```
jgraettinger, line 148: I'd recommend making this a left outer join and filtering `live_specs` to those the user has a `read` capability for. If the user shouldn't see it, why even bring it into the application?

crates/agent-sql/src/evolutions.rs
```diff
@@ -0,0 +1,183 @@
...
+    pub updated_at: DateTime<Utc>,
+    pub last_pub_id: Id,
+    pub catalog_name: String,
+    pub spec_type: Option<CatalogType>,
+    pub spec: Option<Json<Box<RawValue>>>,
+    pub spec_capabilities: Json<Vec<RoleGrant>>,
+    pub user_capability: Option<Capability>,
+}
+
+pub async fn fetch_live_specs(
```
jgraettinger, line 78: Adding this comment later, after having otherwise finished the review:

I completely missed where this is called. You're fetching draft specs, sure. You use the expand-graph function to fetch connected captures / materializations. Where does this fit in?

jgraettinger (commented):
Thanks! Sorry for delay, this was a bigger one. Comments below.

psFried:
@jgraettinger I'm still working through the rest of your comments, but one thing I'd like to clarify is that I think it's important that we only add specs to the draft that the user has `admin` capability to. Sure, _technically_ you're allowed to draft an item with only `read` access, but in this case I would argue that there's A, no value in doing so, and B, it pushes a lot more complexity onto the user and/or frontend (which now has to separately filter the draft specs based on access level).

A more principled approach might be for this handler to return an error if there's any connected materializations that the user doesn't have `admin` capability to. But IMO it'd be more reasonable in this scenario to allow the user to still evolve their collection and update any materializations they _do_ have access to, and then separately notify the owners of any other materializations that the collection they were materializing has been re-created.

Ultimately, I could see wanting to give users a choice in how we handle that, which is part of why I read in all the connected materializations along with the `user_capability` (as opposed to filtering out materializations based on `admin` capability).

crates/agent-sql/src/evolutions.rs
psFried: > If the user shouldn't see it, why even bring it into the application?

I don't have a _great_ answer to this. Originally, I was thinking we might want to take some other action to notify the owner of that other materialization that one of its bound collections is changing, but that doesn't actually make tons of sense given that we're only operating on a draft. At the moment, the only purpose is so we can log that we're not updating it, which admittedly doesn't seem worth it. I'll just change this to filter them out here.

crates/doc/src/inference.rs
```diff
@@ -899,6 +909,9 @@ impl Shape {
         let default = union_option(lhs.default, rhs.default);
         let secret = union_option(lhs.secret, rhs.secret);
 
+        let mut annotations = lhs.annotations;
+        annotations.extend(rhs.annotations.into_iter());
```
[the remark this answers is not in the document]
psFried: Yeah, that's actually what I started out doing, but I realized it doesn't actually work for annotations. Concretely, the problem is due to [this code](https://github.com/estuary/flow/blob/6183db1396e68e595af2f4d71779a1f6b42ea0ad/crates/doc/src/inference.rs#L873), which intersects the shapes of `oneOf` and `allOf` with the base shape. Consider a schema like:

```
x-outer-annotation: 1
oneOf:
  - x-some-annotation: true
```

If you intersect the top-level annotations with the `oneOf`, then the result will always be empty.

Less concretely, I think semantically the `annotations` of a shape are all the annotations that _might_ apply at a given location. If you want only the annotations that _do_ apply, then you need to actually validate a document. So like in [this example](https://github.com/estuary/flow/blob/6183db1396e68e595af2f4d71779a1f6b42ea0ad/crates/doc/src/inference.rs#LL2448C28-L2448C28) we should expect the Shape to include [all four annotations](https://github.com/estuary/flow/blob/6183db1396e68e595af2f4d71779a1f6b42ea0ad/crates/doc/src/snapshots/doc__inference__test__annotation_collection.snap#L44)

crates/doc/src/inference.rs
```diff
@@ -958,6 +972,12 @@ impl Shape {
         let default = lhs.default.or(rhs.default);
         let secret = lhs.secret.or(rhs.secret);
 
+        // Intersection of annotations is actually a _union_ of the sets of annotations.
+        // Though counterintuitive, this results in `annotations` being the full set of
+        // of annotations that might be applied at a given location.
+        let mut annotations = lhs.annotations;
```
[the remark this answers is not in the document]
psFried: If you accept the framing of `annotations` being all of the annotations that _might_ apply at a given location, then I think the most correct behavior here would actually be to use a multi-map and collect all the values from both `lhs` and `rhs`. I'd considered that a weird enough edge case that we didn't need to support it right now ([see comment here](https://github.com/estuary/flow/blob/6183db1396e68e595af2f4d71779a1f6b42ea0ad/crates/doc/src/inference.rs#L2463)). In truth, that was partly motivated by a bit of decision paralysis over which multimap implementation to use, and how to handle duplicate values.

crates/doc/src/inference.rs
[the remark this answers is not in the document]
jgraettinger: I'm not sure I'm following the problem here.

`one_of`, in the code you linked, will be the single sub-schema. It will be intersected with the parent schema. That intersection actually becomes `child.annotations.extend(parent.annotations)`, which is what you want, right?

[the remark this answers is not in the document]
jgraettinger: > I think semantically the annotations of a shape are all the annotations that might apply at a given location. 

That's not how `Shape` works for other annotations, like `title`, `format`, and so on. 
`Shape` tells us only what we know is _always_ true about a document location.

Let's talk through over VC?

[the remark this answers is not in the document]
psFried: Caught up on VC about this, and actually the annotations in a Shape should only include the annotations that are _guaranteed_ to apply at a particular location. Given that, I'm changing this back to an intersection.

crates/agent-sql/src/evolutions.rs
psFried: whoops, yea this was vestigial

crates/agent-sql/src/evolutions.rs
psFried: Caught up about this on VC, and the resolution is that there's no longer a need to pull in `user_capability` here since the user has already added the spec to their draft. So they must already have `read` access at a minimum, and authZ issues can get caught when they try to publish.

crates/agent/src/publications.rs
```diff
@@ -330,21 +374,80 @@ impl PublishHandler {
...
+    pub fields: Vec<RejectedField>,
+}
+
+#[derive(Debug, Deserialize, Serialize, Clone, PartialEq)]
+pub struct RejectedField {
+    pub field: String,
+    pub reason: String,
+}
+
+fn get_incompatible_collections(build_output: &builds::BuildOutput) -> Vec<IncompatibleCollection> {
```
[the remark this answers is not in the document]
psFried: I went ahead and moved it there for now, but TBH I'm kinda :shrug: on where it should live. A desirable future enhancement is to have this function also return incompatible collections based on a comparison of the live and draft json schemas, and it's not entirely clear to me how that change might impact all this.

psFried:
@jgraettinger One call out with that framing is that `evolutions` _requires_ that every collection given as input must exist in the given draft.

So in D, there's two possible scenarios:

- The collection exists in the draft as `new_name`, but not as `old_name`: an error is returned
- The collection exists in the draft as both `old_name` and `new_name`: the `new_name` collection spec is overwritten in the draft.

Regarding E and F, the handler does absolutely no checking of permissions for the `new_name`, so in both of those scenarios we'd just add the new collections to the draft and let publication return the errors. You might already have another live spec having `new_name`, and there's actually no guardrails around any of that right now. Another future enhancement of this _might_ be to check the `new_name`s to ensure they don't conflict with any live specs, although I held off on that for now because it's especially tricky to avoid leaking information about unauthorized specs, and doing so correctly would require explicit permission checking.

Overall, I'm fine with taking an approach where we skip a lot of authZ checks during evolution and instead rely on publish for those, at least for the time being.

I pushed up changes addressing all the feedback. PTAL

crates/doc/src/inference.rs
```diff
@@ -899,6 +909,11 @@ impl Shape {
         let default = union_option(lhs.default, rhs.default);
         let secret = union_option(lhs.secret, rhs.secret);
 
+        // Union of annotations is actually an _intersection_, which yields only
+        // the annotations that are guaranteed to apply at a given location.
+        let mut annotations = lhs.annotations;
+        annotations.retain(|k, _| rhs.annotations.contains_key(k));
```
jgraettinger, line 915: nit: technically this should assert that the values are also equal.

jgraettinger (approved):
Just a couple comments then LGTM

crates/doc/src/inference.rs
psFried: :+1: done

crates/agent-sql/src/evolutions.rs
```diff
@@ -0,0 +1,136 @@
...
+            draft_specs.spec_type as "draft_type: CatalogType",
+            live_specs.spec_type as "live_type: CatalogType",
+            live_specs.last_pub_id as "last_pub_id: Option<Id>",
+            live_specs.id as "live_spec_id: Option<Id>"
+        from drafts
+        left join draft_specs on drafts.id = draft_specs.draft_id
+        left join live_specs
+            on draft_specs.catalog_name = live_specs.catalog_name
+            -- Ensure that `live_spec_id` is `None` if the user does not have admin access to the spec.
+            and exists (select 1 from internal.user_roles($2, 'admin') r where live_specs.catalog_name ^@ r.role_prefix)
```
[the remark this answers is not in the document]
psFried: whoops, yea, I'd found that and I feel like it's clearer without the cte. fixed.

PR #1033, opened 2023-05-10
reactor: add reactor instance details to ops logs

psFried:
**Description:**

Adds information about the reactor instance to the ops logs. We had an incident recently where a single reactor instance was faulting due to a problem with the k8s node. While it was easy to see which tasks were assigned to that instance at the time, it was quite difficult to determine that after the fact. So the goal of this commit is to make it easier to troubleshoot this type of thing in the future.

The details of the reactor instance are not added to _all_ logs in order to prevent unnecessary bloat. Instead they are only logged on term initialization and shard failure.

Example log messages:

```
{
  "_meta": {
    "uuid": "90f10600-ef51-11ed-8400-5382b2069d86"
  },
  "shard": {
    "kind": "materialization",
    "name": "aliceCo/matfail",
    "keyBegin": "00000000",
    "rClockBegin": "00000000"
  },
  "ts": "2023-05-10T16:41:52.145687112Z",
  "level": "info",
  "message": "initialized catalog task term",
  "fields": {
    "labels": {
      "build": "09bf8de556810000",
      "log_level": 3,
      "ports": null,
      "range": {
        "key_end": 4294967295,
        "r_clock_end": 4294967295
      },
      "task_name": "aliceCo/matfail",
      "task_type": 3
    },
    "lastLabels": {
      "ports": null,
      "range": {}
    },
    "reactorInstance": {
      "Zone": "local",
      "ID": "",
      "Host": "localhost",
      "Port": "9000"
    }
  }
}
```

```
{
  "_meta": {
    "uuid": "923430f3-ef51-11ed-8400-5382b2069d86"
  },
  "shard": {
    "kind": "materialization",
    "name": "aliceCo/matfail",
    "keyBegin": "00000000",
    "rClockBegin": "00000000"
  },
  "ts": "2023-05-10T16:41:54.263609149Z",
  "level": "error",
  "message": "shard failed",
  "fields": {
    "error": "txnStartCommit: app.FinalizeTxn: combine.Finish: document failed validation...",
    "reactorInstance": {
      "Zone": "local",
      "ID": "",
      "Host": "localhost",
      "Port": "9000"
    }
  }
}
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1033)

### DIFF
--- a/go/runtime/task_term.go
+++ b/go/runtime/task_term.go
@@ -18,12 +18,14 @@ import (
 	"go.gazette.dev/core/consumer"
 	pc "go.gazette.dev/core/consumer/protocol"
 	"go.gazette.dev/core/keyspace"
+	"go.gazette.dev/core/mainboilerplate"
 	"go.gazette.dev/core/message"
 )
 
 // taskTerm holds task state used by Capture, Derive and Materialize runtimes,
 // which is re-initialized with each revision of the associated catalog task.
 type taskTerm struct {
+	consumerConfig *mainboilerplate.ServiceConfig
 	// Current ShardSpec under which the task term is running.
 	shardSpec *pf.ShardSpec
 	// The taskTerm Context wraps the Shard Context, and is further cancelled
@@ -35,6 +37,12 @@ type taskTerm struct {
 	build *flow.Build
 	// ops.Publisher of ops.Logs and ops.Stats.
 	opsPublisher *OpsPublisher
+}
+
+func newUninitializedTerm(consumerConfig *mainboilerplate.ServiceConfig) taskTerm {
+	return taskTerm{
+		consumerConfig: consumerConfig,
+	}
 }
 
 func (t *taskTerm) initTerm(shard consumer.Shard, host *FlowConsumer) error {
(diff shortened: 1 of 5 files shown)


go/runtime/task_term.go
```diff
@@ -93,6 +101,7 @@ func (t *taskTerm) initTerm(shard consumer.Shard, host *FlowConsumer) error {
 		"initialized catalog task term",
 		"labels", t.labels,
 		"lastLabels", lastLabels,
+		"reactorInstance", t.consumerConfig,
```
jgraettinger, line 104: Rather than threading through the service config, you can instead use the [Shard.Assignment](https://github.com/gazette/core/blob/master/consumer/interfaces.go#L47-L50) accessor to get the [allocator.Assignment](https://github.com/gazette/core/blob/master/allocator/allocator_key_space.go#L72), which has this info and a bit more:

```
// The cast isn't even technically necessary if you're just logging `Decoded`.
"assignment": shard.Assignment().Decoded.(allocator.Assignment)
```

psFried: Looks like that works even better, since hostname in consumer config is just a pod IP, and the `MemberSuffix` on the assignment is the k8s pod name. Easiest thing was to just re-implement as a new commit, so I'm just closing this in favor of #1039

PR #1034, opened 2023-05-10
flowctl: use rustyline for reading login token

### LINKED ISSUE #1031: Unable to paste full token with `flowctl auth login`  on mac
For some reason, our login tokens seem to have gotten a bit longer, which puts them over the 1024 character limit for pasting into iTerm2. I have no idea why there's a 1024 character limit, or why that limit doesn't seem to apply when pasting at the end of a `flowctl auth token --token ` command.

The goal here is to make `flowctl` logins work correctly on all common mac setups. One avenue might be to change `flowctl auth login` so that instead of trying to read the token from stdin directly it would print a message like "run `flowctl auth token --token <your-token-here>` with the token you get from the web page".

psFried:
**Description:**

Fixes #1031 

On some terminals, `flowctl auth login` would fail to read JWTs longer than 1024 characters. This bug was introduced when we removed `rpassword` in favor of `stdin.readline(_)`, which meant that we left the terminal in "cooked" instead of "raw" mode. This commit introduces the `rustyline` library (a pure-rust version of readline) to read the token input. This handles pastes of arbitrarily long strings.

This fix is a little different from what we were previously thinking. We'd previously talked about having the `auth login` command instead print a message telling people to follow up by running `flowctl auth token`. But having found the root cause and a simple solution, it seemed worthwhile to keep the current UX unchanged.

**Workflow steps:** no change

**Documentation links affected:** no change

**Notes for reviewers:**

I tested this on my m1 mac with both iTerm2 and the built-in terminal. Also tested on linux using both Alacritty and gnome terminal. Prior to this patch, I was able to reproduce the bug with either terminal on mac, but neither of the linux terminals. The explanation for this is that Alacritty (and presumably gnome terminal, too) has special behavior for this that copies pasted text to the input buffer in small chunks with a delay between each chunk.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1034)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -737,6 +737,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "clipboard-win"
+version = "4.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7191c27c2357d9b7ef96baac1773290d4ca63b24205b82a3fd8a0637afcf0362"
+dependencies = [
+ "error-code",
+ "str-buf",
+ "winapi",
+]
+
+[[package]]
 name = "codespan-reporting"
 version = "0.11.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


psFried:
The flow-web job failed due to the GH outtage, and I just didn't see any point in running it again.

jgraettinger (approved):
LGTM

PR #1039, opened 2023-05-15
reactor: add reactor instance details to ops logs

psFried:
**Description:**

Adds information about the reactor instance to the ops logs. We had an incident recently where a single reactor instance was faulting due to a problem with the k8s node. While it was easy to see which tasks were assigned to that instance at the time, it was quite difficult to determine that after the fact. So the goal of this commit is to make it easier to troubleshoot this type of thing in the future.

The details of the reactor instance are not added to _all_ logs in order to prevent unnecessary bloat. Instead they are only logged on term initialization and shard failure.

This is a new iteration on #1033, which will be closed.

Example of a logged field from a local instance:

```
{
  "ItemID": "materialize/aliceCo/mat1/00000000-00000000",
  "MemberZone": "local",
  "MemberSuffix": "notable-eft",
  "Slot": 0,
  "AssignmentValue": {}
}
```

In production, the `MemberSuffix` will be the k8s pod name, IIUC.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1039)

### DIFF
--- a/go/runtime/capture.go
+++ b/go/runtime/capture.go
@@ -414,8 +414,8 @@ func (c *Capture) FinalizeTxn(_ consumer.Shard, pub *message.Publisher) error {
 }
 
 // FinishedTxn logs if an error occurred.
-func (c *Capture) FinishedTxn(_ consumer.Shard, op consumer.OpFuture) {
-	logTxnFinished(c.opsPublisher, op)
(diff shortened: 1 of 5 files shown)


jgraettinger (approved):
LGTM

PR #1052, opened 2023-05-19
agent: better errors when key pointers change

psFried:
**Description:**

When publishing collection specs, the agent will now return `incompatible_collections` when either the `key` or partition fields of a collection have changed. This addresses the scenario where a user re-runs discovery on their endpoint and the connector outputs collections with new keys. In this case, the UI will now suggest re-creating the affected collections with new names.

**Workflow steps:**

- Create a SQL capture of table `foo`
- In the source database, change the primary key of `foo` to a different column
- edit a capture
- click "refresh" in the bindings component to discover the collection with a different key
- click save and publish
- With this PR, you'll now see an error like this one, with the suggestion to re-create the collections:

Screenshot from 2023-05-19 15-33-02

**Documentation links affected:** I don't think we currently document this

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1052)

### DIFF
--- a/crates/agent/src/publications/builds.rs
+++ b/crates/agent/src/publications/builds.rs
@@ -21,6 +21,7 @@ pub struct BuildOutput {
 #[derive(Debug, Deserialize, Serialize, Clone, PartialEq)]
 pub struct IncompatibleCollection {
     pub collection: String,
+    #[serde(skip_serializing_if = "Vec::is_empty")]
     pub affected_materializations: Vec<AffectedConsumer>,
 }
 
--- a/crates/agent/src/publications.rs
+++ b/crates/agent/src/publications.rs
@@ -167,10 +167,16 @@ impl PublishHandler {
             return stop_with_errors(errors, JobStatus::build_failed(Vec::new()), row, txn).await;
         }
 
-        let errors =
-            specs::validate_transition(&draft_catalog, &live_catalog, row.pub_id, &spec_rows);
-        if !errors.is_empty() {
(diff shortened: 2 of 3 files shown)


crates/agent/src/publications/builds.rs
```diff
@@ -21,6 +21,7 @@ pub struct BuildOutput {
 #[derive(Debug, Deserialize, Serialize, Clone, PartialEq)]
 pub struct IncompatibleCollection {
     pub collection: String,
+    #[serde(skip_serializing_if = "Vec::is_empty")]
```
psFried, line 24: There's no good way to populate `affected_materializations` in this scenario because we haven't yet expanded the catalog when we call `validate_transition`. So I'm thinking that "affected_" is not a great name here, and perhaps we can consider renaming it to `rejected_by_materializations`. Nothing actually reads this value, though, so I'm thinking we deal with renaming it in a subsequent PR so as not to delay this one.

jgraettinger:
>  the connector outputs collections with new keys

My recollection is the discovery deep merge behavior won't apply an updated key from a refresh. Is that wrong?

psFried:
> > the connector outputs collections with new keys
> 
> My recollection is the discovery deep merge behavior won't apply an updated key from a refresh. Is that wrong?

Yeah, it [looks like](https://github.com/estuary/flow/blob/734dfd2f234c6fcb51082a79d5cfbf3c93869df1/crates/agent/src/discovers/specs.rs#L146) we just overwrite the key as long as it's not empty

PR #1058, opened 2023-05-25
runtime: cleanup uses of `errors` package

jgraettinger:
Use the standard library `errors` package everywhere.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1058)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -11,6 +11,7 @@ require (
 	github.com/gogo/protobuf v1.3.2
 	github.com/golang/protobuf v1.5.2
 	github.com/gorilla/websocket v1.4.2
+	github.com/grpc-ecosystem/go-grpc-prometheus v1.2.0
 	github.com/invopop/jsonschema v0.5.0
 	github.com/jackc/pgx/v4 v4.10.1
 	github.com/jessevdk/go-flags v1.5.0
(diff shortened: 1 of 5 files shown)


jshearer (approved):
Sure looks like what we talked about on the open-hours call! After reading the docs, I admit to still not actually understanding the difference between `errors.Cause(err) == client.ErrOffsetNotYetAvailable` and `errors.Is(err, client.ErrOffsetNotYetAvailable)`, but asking ChatGPT for clarification claims that:

1. `errors.Is(err, client.ErrOffsetNotYetAvailable)`:
   - The `errors.Is` function performs a recursive search through the error chain, checking if any of the errors in the chain match the target error (`client.ErrOffsetNotYetAvailable` in this case).
   - It considers both the error value itself and its underlying causes during the search.

2. `errors.Cause(err) == client.ErrOffsetNotYetAvailable`:
   - The `errors.Cause` function extracts the underlying cause of the error, which is typically the original error that occurred before any wrapping or wrapping done by error types like `fmt.Errorf` or `errors.Wrap`.
   - It does not perform a recursive search; it only retrieves the immediate cause of the error.

If that's true, then it seems like the bug here is that we were only checking the "root cause" of the error, whereas in reality the `ErrOffsetNotYetAvailable` cause was somewhere in the middle of the chain of wrapped errors? If so, LGTM!

jgraettinger:
The actual problem is that the `errors` package is different. This PR switches from `github.com/pkg/errors`, which is an older third-party library for error management and chaining, to the stdlib `errors` package.

The core problem is the standard library `errors` and `fmt.Errorf` (which is used everywhere now) have a different interface for representing error chains, meaning the older package doesn't understand chained errors produced by `fmt.Errorf`.

jshearer:
When you say used everywhere, do you mean that it was used everywhere before, [except in this one place](https://github.com/estuary/flow/pull/1058/files#diff-0f7d73cb1ff98f4b131bc1b6010d98b99f9dc4ea33e6c1c3b434bd9a5c9e9e01R206)?

PR #1059, opened 2023-05-26
More Airbyte connectors

jman0119:
Docs for more Airbyte connectors:

YouTube Analytics
Braintree
Braze
NetSuite
Harvest
Jira
1 more: Instagram

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1059)


site/docs/reference/Connectors/capture-connectors/twilio.md
```diff
@@ -1,3 +1,8 @@
+
```
oliviamiannone, line 1: Whenever you add sidebar_position, make sure you start on line 1 of the file. Otherwise, docusaurus doesn't see it, so it doesn't work AND it prevents the title from rendering properly in the TOC (you can also tell by the grey text color when it's not working).

site/docs/reference/Connectors/capture-connectors/youtube-analytics.md
```diff
@@ -0,0 +1,80 @@
...
+* [playlist_device_os_a1](https://developers.google.com/youtube/reporting/v1/reports/channel_reports#playlist-device-type-and-operating-system)
+* [playlist_playback_location_a1](https://developers.google.com/youtube/reporting/v1/reports/channel_reports#playlist-playback-locations)
+* [playlist_province_a1](https://developers.google.com/youtube/reporting/v1/reports/channel_reports#playlist-province)
+* [playlist_traffic_source_a1](https://developers.google.com/youtube/reporting/v1/reports/channel_reports#playlist-traffic-sources)
+
+By default, each resource is mapped to a Flow collection through a separate binding.
+
+## Prerequisites
+
+* YouTube does not start to generate a report until you create a [reporting job](https://developers.google.com/youtube/reporting/v1/reports#step-3:-create-a-reporting-job) for that report.
```
oliviamiannone, line 41: Seems like you copied this from Airbyte's docs verbatim. This isn't advisable for a few reasons:
1. Plagiarism - we should be rephrasing everything we get from Airbyte.
2. It mentions "Airbyte" and we are "Flow."
3. Often, Airbyte's docs are structured in a different way than ours are. They're also pretty poorly written. This is a great example of that: not only is this paragraph difficult to read and doesn't actually _say much_ that's relevant to a user, it's not really a prerequisite. It's just telling us what the connector will do. If you look at Airbyte's docs page for this connector, you'll notice that what we list as "prerequisites" are baked into their "setup" sections. The convention we've been following in our docs thus far is to provide a bulleted list of Prerequisites, with an optional Setup section if meeting those prerequisites isn't very straightforward.

site/docs/reference/Connectors/capture-connectors/youtube-analytics.md
```diff
@@ -0,0 +1,80 @@
...
+The report for September 2, 2015, will be posted on September 4, 2015, and so forth.
+Youtube also generates historical data reports covering the 30-day period prior to when you created the job. Airbyte syncs all available historical data too.
+
+## Configuration
+
+You configure connectors either in the Flow web app, or by directly editing the catalog specification file.
+See [connectors](../../../concepts/connectors.md#using-connectors) to learn more about using connectors. The values and specification sample below provide configuration details specific to the YouTube Analytics source connector.
+
+### Properties
+
```
oliviamiannone, line 54: Even if there's no endpoint config beyond oauth, we should still provide it. People will still be working with their specs directly via the CLI, and if we ever make the path to self-hosting easier, we want to have our bases covered at that point.

site/docs/reference/Connectors/capture-connectors/jira.md
```diff
@@ -0,0 +1,118 @@
...
+* [Board issues](https://developer.atlassian.com/cloud/jira/software/rest/api-group-board/#api-rest-agile-1-0-board-boardid-issue-get)
+* [Issue comments](https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-issue-comments/#api-rest-api-3-issue-issueidorkey-comment-get)
+* [Issue worklogs](https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-issue-worklogs/#api-rest-api-3-issue-issueidorkey-worklog-get)
+* [Issues](https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-issue-search/#api-rest-api-3-search-get)
+* [Sprint issues](https://developer.atlassian.com/cloud/jira/software/rest/api-group-sprint/#api-rest-agile-1-0-sprint-sprintid-issue-get)
+By default, each resource is mapped to a Flow collection through a separate binding.
+
+## Prerequisites
+
+- API Token
```
oliviamiannone, line 73: Explain more on how to generate API token, at minimum link to Jira's docs. You'll notice in the actual Endpoint Config, we have a shortlink intended to bring people from the UI to this docs page. So we want to have that information here for folks when they follow that link, otherwise it's a bad experience.

site/docs/reference/Connectors/capture-connectors/jira.md
```diff
@@ -0,0 +1,118 @@
...
+You configure connectors either in the Flow web app, or by directly editing the catalog specification file.
+See [connectors](../../../concepts/connectors.md#using-connectors) to learn more about using connectors. The values and specification sample below provide configuration details specific to the Jira source connector.
+
+### Properties
+
+#### Endpoint
+
+| Property | Title | Description | Type | Required/Default |
+|---|---|---|---|---|
+| `/api_token` | API Token | ira API Token. See the docs for more information on how to generate this key. API Token is used for Authorization to your account by BasicAuth: https://go.estuary.dev/cVlN5P | string | Required |
```
oliviamiannone, line 88: Goes along with comment above... this description is intended to populate the UI, so it doesn't really make sense here -- get rid of the bit that says "See the docs..." and remove the shortlink. Also, "jira" is missing a "J."

site/docs/reference/Connectors/capture-connectors/netsuite.md
```diff
@@ -0,0 +1,68 @@
...
+
+It is available for use in the Flow web application. For local development or open-source workflows, [`ghcr.io/estuary/source-netsuite:dev`](https://ghcr.io/estuary/source-netsuite:dev) provides the latest version of the connector as a Docker image. You can also follow the link in your browser to see past image versions.
+
+This connector is based on an open-source connector from a third party, with modifications for performance in the Flow system.
+You can find their documentation [here](https://docs.airbyte.com/integrations/sources/netsuite/),
+but keep in mind that the two versions may be significantly different.
+
+## Supported data resources
+
+- Streams are generated based on `ROLE` and `USER` access to them as well as `Account` settings, make sure you're using the correct role assigned in our case `airbyte-integration-role` or any other custom `ROLE` granted to the Access Token, having the access to the NetSuite objects for data sync.
```
oliviamiannone, line 16: "streams" is airbyte verbiage we want to avoid (except in the resource config, where we have less control) - we say "resources." Also, `airbyte-integrations-role` should probably be changed. I'm not actually very clear based on reading this what the role does, ad if there's more prerequisites that might need to be done to get this to actually work?

oliviamiannone (changes requested):
Thanks for doing this Jenny -- the main pieces are all here, but I have some requested changes.

* Please add all the new connectors to the README page within capture connectors (acts as the capture connectors landing page). It’s worth considering removing this at some point as it’s getting lengthy, but while we have it it should be up-to-date.

* Please roll back changes to `sidebar_position`; the TOC is out of order. Some more context on how this works: by default, Docusaurus orders the TOC by filename. But this doesn’t always correspond to the title. So without adding `sidebar_position`, the TOC would be out of order. Rather than explicitly ordering _everything_, I’ve batched `sidebar_position` as needed to get the control I want. For example, everything within `sidebar_position_1` will be alphabetized amongst themselves, followed by everything with `sidebar_position_2` and so on. Things without explicit positioning come last. All this to say: it’s hard to get the correct behavior without [rendering a preview](https://docs.google.com/document/d/1SRC9VS9zyCzWl3n4HXHbc4wPB1eLxJHkA2rtu9ZNokM/edit#heading=h.db3qlsbnkjaf) and playing around with it. If you’re not able to render a preview to check your docs work (which I highly recommend you get set up at some point), please omit `sidebar_position` from everything and allow me to take care of it, since I am able to render a preview.

* Be really careful about copying things from Airbyte. I have a comment on a few of the specific files, but in general, we want to go BEYOND copying their stuff and changing the word “Airbyte” to “Flow” or “Estuary.” We should paraphrase,  and probably IMPROVE, on their docs (their writing is very erratic and often hard to understand.) We also should adhere to our own [docs style](https://docs.google.com/document/d/1JwnTP7zvv9x0IHjvKb4oOOA-0f_ukhflGDI-i7jaFW4/edit). Most importantly, though their connectors are open-source, copying their docs and publishing sections as if they’re our own is sketchy territory.

site/docs/reference/Connectors/capture-connectors/jira.md
jman0119: Corrected this! sorry for the typo

site/docs/reference/Connectors/capture-connectors/jira.md
jman0119: Added Jira link with instructions on how to generate the API token

site/docs/reference/Connectors/capture-connectors/youtube-analytics.md
jman0119: Added endpoint config section!

site/docs/reference/Connectors/capture-connectors/youtube-analytics.md
jman0119: I just tried the youtube connector and didn't need to prep anything else other than the oauth authentication. I think we can remove this prerequisite.

site/docs/reference/Connectors/capture-connectors/twilio.md
jman0119: Thanks for explaining how the `sidebar_position` works! During my last commit, there were a few conflicts suggesting the sidebar_position changes, I think it might be because I used a previously used codespace. I have rolled back the changes to sidebar_position now, those weren't intentional, I'll let you handle this. Also, will get with you to see if I can also render previews on my side, thanks!

site/docs/reference/Connectors/capture-connectors/netsuite.md
oliviamiannone: I just took a look at the Airbyte docs for this connector. It looks like you create a custom role as part of the setup (in their case, `airbyte-integrations-role`, so we probably wanna tell people to create `estuary-role`). And any objects to which you grant that role access are captured. 

So two action items here:
* Revise this section to say something like "Flow captures collections from any NetSuite object to which you grant access during setup[Link to Setup section], including `Transactions`, `Reports`, `Lists`, and `Setup`."  Don't worry about linking Netsuite docs to object types, as the NetSuite docs are incredibly confusing. 
* Expand the prerequisites section and add a Setup section based off Airbyte's prerequisites. Something worth noting: Airbyte talks about prerequisites as what's needed _before_  you begin setup; whereas we talk about the prerequisites in terms of what needs to be ready immediately before using Flow. So the Setup steps help users _meet_ the prerequisites. Hope that makes sense -- for an example see https://docs.estuary.dev/reference/Connectors/capture-connectors/PostgreSQL/#prerequisites. For Netsuite, looks like the actual prereqs are: new integration with token-based auth; a custom role with access to objects you want to capture; a new user assigned that role; access token generated for the role.

jman0119: Hi @oliviamiannone I've made those changes, let me know if they look fine to you, thanks!

oliviamiannone (approved):

site/docs/reference/Connectors/capture-connectors/netsuite.md
oliviamiannone: Hey Jenny -- This looks fine for the time being, considering where we're at with the connector! In the future we can add more details / links.

Also, want to clarify my comment about the Netsuite docs -- I don't think they're helpful to link for the _suported resources_ but I do think they'd be helpful to link for some of the prerequisites, since we don't want to write out a detailed step by step like airbyte. EG, the docs for a new integration are https://docs.oracle.com/en/cloud/saas/netsuite/ns-online-help/section_157771733782.html But again it's fine to hold off and just get this merged.

PR #1062, opened 2023-05-26
Introduce task failures derivation to the ops log

jshearer:
**Note:** this already applied and running as `ops.us-central1.v1/failure-notifications/v1`

The goal is to create a collection of task failures by scanning through the ops logs for messages that we know indicate a failure. Specifically, the string `shard failed` [should always be logged](https://github.com/estuary/flow/blob/d4cd662499c53d62d7f425dd043bd885e830b970/go/runtime/flow_consumer.go#L115-L125) when a shard fails.

In addition to logging `shard failed`, the log message also contains an estimated "cause" for the error under the `/fields/error` key, as well as any other associated metadata fields.

Since we're already processing log messages, it was easy to enhance each failure notification event with data about the most recent `warning` and `error` level log messages.

The derivation outputs both "raw" data in the form of a JSON document describing the failure, as well as data specifically formatted for the Slack API. The final notifications in slack look like this:
Screen Shot 2023-05-26 at 17 20 17

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1062)

### DIFF
--- a/ops-catalog/notify-failure-logs.sql
+++ b/ops-catalog/notify-failure-logs.sql
@@ -23,28 +23,7 @@ WHERE (shard, ts) IN
         LIMIT -1 OFFSET 10
     );
 
-WITH events_since_previous_notification as (
-    select
-        count(*) as event_count
-    from
-        failures
-    where
-        failures.shard = $name
-        and julianday(failures.ts) > COALESCE(
-            (
-                SELECT
-                    ts
-                from
-                    failure_notifications
-                where
-                    failure_notifications.shard = $name
-                    AND julianday(failure_notifications.ts) < julianday($ts)
-                order by
-                    julianday(failure_notifications.ts) desc
-            ), 0
-        )
-),
-return_values as (
+WITH return_values as (
     select
     $message,
     substr(COALESCE($fields->>'error','Missing `error` field, failure reason unknown.'),0,2500) as reason,
@@ -60,7 +39,6 @@ return_values as (
         'message',message,
         'formatted_message',SUBSTR('[ts='||ts||' level='||level||']: '||message,0,2500)
     )) from related_log_lines WHERE shard=$name) as log_lines,
-    (SELECT event_count from events_since_previous_notification) AS events_since_last_notification,
     (SELECT json_group_object(key, value) from json_each($fields) where key != 'error') as filtered_metadata
 )
 INSERT INTO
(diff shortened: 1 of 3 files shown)


ops-catalog/notify-failure-logs.sql
```diff
@@ -0,0 +1,223 @@
...
+DELETE FROM related_log_lines
+WHERE (shard, ts) IN
+    (
+        SELECT shard, ts from related_log_lines
+        WHERE shard = $name
+        ORDER BY ts DESC
+        LIMIT -1 OFFSET 10
+    );
+
+WITH events_since_previous_notification as (
```
jgraettinger, line 26: Not requesting changes, given this is already deployed and clearly works, but this is hard to follow and I'm curious for your thoughts on how it could be improved for readability.

One pattern I've been playing with myself is a `globals` table that holds any and all global-scope variables. I'll make its columns nullable and then have a single `INSERT INTO globals DEFAULT VALUES;` in the migration. Lambdas then `UPDATE` its one row as needed.

Pragmatically, I've found that it reduces my use of CTEs and lets me express logic a bit more linearly, in smaller chunks. Another benefit is that you can define common views which work off of `globals`, such as building up a complex output, and then trivially re-use them across multiple lambdas.

Going further down the rabbit hole: you can also model functions via a table like `create table do_the_thing(arg1 text, arg2 boolean)`, and then define an after-insert trigger which contains one or more statements, including a statement that deletes the inserted `rowid`. Multiple triggers can be defined, each having a `WHEN` clause, which allows for modeling conditionals. Triggers can also be called recursively, up to a limit.

Clearly this ☝️ can get a bit nuts, but it was nice to discover that it's not _too_ crazy to model a complex, procedural, branching control flow using only SQLite.

jgraettinger (approved):
LGTM

ops-catalog/notify-failure-logs.sql
jshearer: Yeah, I thought this would be more interesting/useful than it was. I ended up not sending it to slack but forgot to remove the calculation for it. 

Wrt the procedural, branching control flow in SQL(ite), yeah it took me a second to wrap my head around how to do what I wanted. For example, I needed to build up an array of objects to send to Slack and conditionally include certain fields/objects inside it. What I ended up with was this stuff:

```sql
        select JSON_GROUP_ARRAY(value)
        from JSON_EACH(
            JSON_ARRAY(
                CASE (select count(*) from json_each(return_values.log_lines))
                    WHEN 0 THEN NULL
                    ELSE JSON_OBJECT('type', 'divider')
                END,
                CASE (select count(*) from json_each(return_values.log_lines))
                    WHEN 0 THEN NULL
                    ELSE JSON_OBJECT(
                        'type','section',
                        'text',JSON_OBJECT(
                            'type', 'mrkdwn',
                            'text', '*Adjacent Log Messages:*'
                        )
                    )
                END
            )
        )
        WHERE type != 'null'
```

Issue #1063, opened 2023-05-30
Automated schema evolution 


psFried:
Discussed in #1042 

A common use case for Flow users is point to point data integrations. In these cases, users often want to have changes to source schemas automatically propagate to their destination. This issue tracks all of the work related to automated handling of schema evolution.

- [x] Enhance the `models::Catalog` to include automatic discovery options
- [x] Update evolutions handler to only update materialization bindings unless explicitly requested to re-create the collection
- [x] Add ability for discover, evolution, and publish jobs to automatically spawn follow-on jobs when they complete     
    - [x] discover can create a publication after a successful discover
    - [x] publish can create an evolution when publish fails due to incompatible collection
    - [x] evolution can create a publication after a successful evolution
- [x] Add pg_cron job to create periodic discovers for captures with automatic discovery enabled

Continuous schema inference is not yet represented here, but we can add it once that plan solidifies.

PR #1065, opened 2023-05-31
agent: evolutions handling inferred schemas

psFried:
**Description:**

This addresses a portion of #1063, but it seemed useful on its own and separable from the rest of the work related to it.

There's a big difference between collections that use inferred schemas and those that don't, and the overall aim of this commit is to have the `evolutions` handler do the "right thing" in both cases. This means introducing some special behavior for collections that bear the `x-infer-schema` annotation in their schema.

If a collection is using an inferred schema, then re-creating it is unlikely to help, and can be quite costly and inefficient. So generally in that case we'd prefer to _only_ update the materialization bindings to materialize the existing collection to a new table. But there are still cases where re-creating the collection is necessary. For example, if the key is updated.

So the following updates to the evolutions behavior will only apply when the collection schema has `x-infer-schema` and the `collections` input does _not_ explicitly request re-creating the collection (by specifying a `new_name`).

In that case, evolution will _not_ re-create the collection, but will instead only set a new resource name in any affected materialization bindings. For example, if you'd materialized `acmeCo/foo` to table `my_table`, then it would be updated to materialize to `my_table_v2`. The `new_name` in the `job_status` will then be the same as the `old_name`.

**Workflow steps:**

When you try to publish a change to a collection that uses `x-infer-schema` and it gets rejected, you'll still see the same message that recommends re-creating the collections. Now, when you click that button it will no longer re-create the collection, but will instead only update the materialization binding to materialize the same collection to a new resource.

For collections that _don't_ use `x-infer-schema`, the behavior is unchanged.

no docs on this _yet_, but I'm working on it.

**Notes for reviewers:**

- The `x-infer-schema` annotation isn't a _great_ way to identify these collections, especially now that the behavior is diverging based on the annotation. It seems desirable in the future to hoist `x-infer-schema` into a top-level property of a `models::CollectionDef`. But I left that out of this PR for the time being.
- @dyaffe @travjenkins @kiahna-tucker This changes the semantics of `evolutions` to no longer be synonymous with "re-creating" collections. So we should probably also coordinate a change to the wording in our UI.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1065)

### DIFF
--- a/crates/agent/src/evolution.rs
+++ b/crates/agent/src/evolution.rs
@@ -19,7 +19,7 @@ pub struct EvolutionHandler;
 /// Rust struct corresponding to each array element of the `collections` JSON
 /// input of an `evolutions` row.
 #[derive(Serialize, Deserialize, Clone, PartialEq)]
-struct RenameRequest {
+struct EvolveRequest {
     old_name: String,
     new_name: Option<String>,
 }
@@ -37,9 +37,13 @@ pub enum JobStatus {
 
 #[derive(Serialize, Deserialize, PartialEq)]
 pub struct EvolvedCollection {
+    /// Original name of the collection
     pub old_name: String,
+    /// The new name of the collection, which may be the same as the original name if only materialization bindings were updated
     pub new_name: String,
+    /// The names of any materializations that were updated as a result of evolving this collection
     pub updated_materializations: Vec<String>,
+    /// The names of any captures that were updated as a result of evolving this collection
     pub updated_captures: Vec<String>,
 }
 
@@ -83,7 +87,7 @@ async fn process_row(
         collections,
         ..
     } = row;
-    let collections_requests: Vec<RenameRequest> =
+    let collections_requests: Vec<EvolveRequest> =
         serde_json::from_str(collections.get()).context("invalid 'collections' input")?;
 
     if collections_requests.is_empty() {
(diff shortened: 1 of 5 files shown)


crates/agent/src/evolution.rs
```diff
@@ -455,35 +509,40 @@ impl ResourceSpecUpdater {
...
+            };
+
+            let new_val = new_collection_name
+                .map(|n| {
+                    n.rsplit('/')
+                        .next()
+                        .expect("collection name must contain a slash")
+                        .to_owned()
+                })
+                .unwrap_or_else(|| next_name(existing).to_owned());
```
jgraettinger, line 539: We can get into trouble if there's a bump of _just_ the binding, followed later by a bump of the source collection, right?

Not saying we should do anything about it, just confirming my understanding.

One possible option would be to differentiate the way in which we increment (just) a resource config name, vs a collection name?

jgraettinger (approved):
LGTM (see comment)

crates/agent/src/evolution.rs
psFried: Just documenting what we'd discussed on VC:
Yeah, this is weird. It seems like a more robust behavior would be to always only add the version suffix to the resource name, and ignore the new collection name completely. This doesn't seem like a change that's needed in this PR, but maybe something to consider for the future.

PR #1070, opened 2023-06-05
Feature/UUID timestamp extraction

jshearer:
**Description:**

Customers were asking for the ability to include a timestamp column in their materializations that represents the point at which Flow saw that piece of data, effectively a "last modified" or "valid as of" field. We actually always had this data encoded as part of each document's UUID, and this PR allows you to get access to it via a "virtual projection" at the location `/_meta/uuid/timestamp`. 

**Workflow steps:**

In order to materialize this timestamp, you have to:
1. Create a projection referencing `/_meta/uuid/timestamp`
1. Include that projection in your materialization's field selection

Unfortunately the UI does not support projections yet, so there is still some CLI work involved. It _does_ technically support field selection by way of the "Advanced Specification Editor". 

Given that, the workflow to materialize the meta uuid timestamp is:
1. Pull down an existing capture: `flowctl catalog pull-specs`
2. Add the projection:
    ```yaml
    projections:
      meta_uuid_timestamp: /_meta/uuid/timestamp
    ```
3. Publish your change: `flowctl catalog publish`
4. Create a materialization for that collection
5. Include your projection in the field selection inside the advanced spec editor: 
    ```json
    "bindings": [
        {
          "source": "my/collection/name",
          "resource": {
            "foo": "bar"
          },
          "fields": {
            "recommended": true,
            "include": {
              "meta_uuid_timestamp": {}
            }
          }
        }
    ],
    ```

If everything works, you should that field materialized! Example of materializing a hello world capture to postgres with the meta timestamp projection:

Screen Shot 2023-06-05 at 11 51 16

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1070)


jshearer:
@jgraettinger see here for my explanation of `uuid_ptr` and `inject_uuid_placeholder`:

https://github.com/estuary/flow/blob/93f88555ad841c62797f30fac5402f82f760179d/crates/derive/src/combine_api.rs#L65-L70

jgraettinger:
FYI you can update Go snapshots by running `UPDATE_SNAPSHOTS=1 make go-test-fast`

jshearer:
Superseded by https://github.com/estuary/flow/pull/1098

PR #1074, opened 2023-06-06
fix: Tuple snapshot tests to work on go 1.20

jshearer:
**Description:**

Go 1.20's [changelog](https://tip.golang.org/doc/go1.20) says:

> The [math/rand](https://tip.golang.org/pkg/math/rand/) package now automatically seeds the global random number generator (used by top-level functions like Float64 and Int) with a random value, and the top-level [Seed](https://tip.golang.org/pkg/math/rand/#Seed) function has been deprecated. Programs that need a reproducible sequence of random numbers should prefer to allocate their own random source, using rand.New(rand.NewSource(seed)).

These tuple packing tests originally came from the [FoundationDB](https://github.com/apple/foundationdb/blob/main/bindings/go/src/fdb/tuple/tuple_test.go) code, and relied on the behavior of `math/rand` initializing with the same seed every time. I ran into this test failure on my local machine as I had installed go 1.20, whereas we test with 1.19 in CI.

I could have opted to just set `GODEBUG=randautoseed=0`, which would have kept the previous fixed-seed behavior, but it felt more fragile to do that than just change the tests to use the original fixed seed value of `1`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1074)

### DIFF
--- a/go/protocols/fdb/tuple/tuple_test.go
+++ b/go/protocols/fdb/tuple/tuple_test.go
@@ -12,6 +12,8 @@ import (
 
 var update = flag.Bool("update", false, "update .golden files")
 
+var src = rand.New(rand.NewSource(1))
+
 func loadGolden(t *testing.T) (golden map[string][]byte) {
 	f, err := os.Open("testdata/tuples.golden")
 	if err != nil {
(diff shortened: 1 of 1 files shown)


jgraettinger (approved):
LGTM

PR #1075, opened 2023-06-08
parser: parse multiple files out of a zip archive as separate files

williamhbaker:
**Description:**

Currently we concatenate the uncompressed bytes of all files from a zip archive for processing. This actually works in some cases, but for most cases it leads to undesirable results. The present motivation is for parsing zip archives containing multiple CSV files where each CSV file has headers, and each file in the archive needs to be evaluated separately.

This changes the parser to handle zip archives specifically in this way. It preserves the ability for a parser type to opt-in to receiving the compressed bytes-as is, although this is not currently used (example: future processing of excel files that are encoded as zip archives). I had originally tried to also make it so that a format could be inferred from files within the archive, but this proved to be more effort than it's probably worth at the moment while still preserving that opt-in ability for the future.

Closes https://github.com/estuary/connectors/issues/754

**Workflow steps:**

Capture files using a `filesource`-based capture where the source contains zip archives. See that, for example, a zip archive containing CSV files each with headers does _not_ have the headers from all non-first files parsed as data rows, and each one has the headers row evaluated separately.

N/A

**Notes for reviewers:**


psFried (approved):
LGTM

PR #1084, opened 2023-06-14
parser: log error details in the message when exiting with an error

williamhbaker:
**Description:**

The actual `message` is what gets picked up as the shard failure reason when a connector fails, and with https://github.com/estuary/connectors/pull/762 this will be logged from the parser when a connector fails due to a parsing error. This changes the logged `message` to include the error details.

**Workflow steps:**

N/A

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1084)


willdonnelly (approved):
LGTM

PR #1085, opened 2023-06-19
agent: add built spec and validate response to control plane db

williamhbaker:
**Description:**

As part of a publication the agent will add the built spec to the `live_specs` table.

If the publication is a dry run, the agent will add the built spec and the validated response to the `draft_specs` table for further use in the publication authoring process.

I tested this manually by:
- Start up a local stack on the current main branch (without this change)
- Create a few tasks - they don't have built spec/validated data saved
- Apply the pending migrations in this PR to the local stack supabase
- Re-build the agent from this PR and re-start it
- Create new tasks - now they have built spec & validated data saved to the local stack supabase for their drafts and live specs

**Workflow steps:**

Create publications and the additional columns are populated in the control plane database

N/A

**Notes for reviewers:**

I experimented a little bit with automated tests for this. The `PublishHandler` currently has a `test_run` flag that is used for other automated tests, but that [ends the processing](https://github.com/estuary/flow/blob/6c131c6830d4a848cd67155ea8e187a6c6a2a51f/crates/agent/src/publications.rs#L269-L271) prior to actually running the build so we wouldn't be able to test anything happening in this PR with that. Maybe this could be re-worked to take some kind of mock build output instead but I did not pursue that very far since it seemed like it would add more complexity than it was worth.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1085)

### DIFF
--- a/crates/agent/src/publications.rs
+++ b/crates/agent/src/publications.rs
@@ -338,8 +338,19 @@ impl PublishHandler {
                 .await
                 .context("rolling back to savepoint")?;
 
+            // Add built specs to the draft for dry runs after rolling back other changes that do
+            // not apply to dry runs.
+            specs::add_built_specs_to_draft_specs(&spec_rows, &build_output, txn)
+                .await
+                .context("adding built specs to draft")?;
+
             return Ok((row.pub_id, JobStatus::Success));
         }
+
+        // Add built specs to the live spec when publishing a build.
+        specs::add_built_specs_to_live_specs(&spec_rows, &build_output, txn)
+            .await
+            .context("adding built specs to live specs")?;
 
         let errors = builds::deploy_build(
             &self.bindir,
--- a/crates/agent-sql/src/drafts.rs
+++ b/crates/agent-sql/src/drafts.rs
@@ -34,6 +34,32 @@ where
     Ok(())
 }
 
+pub async fn add_built_spec<S, V>(
+    draft_spec_id: Id,
+    built_spec: S,
+    validated: Option<V>,
+    txn: &mut sqlx::Transaction<'_, sqlx::Postgres>,
+) -> sqlx::Result<()>
+where
+    S: serde::Serialize + Send + Sync,
+    V: serde::Serialize + Send + Sync,
+{
+    sqlx::query!(
+        r#"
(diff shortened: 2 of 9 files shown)


crates/agent/src/publications.rs
```diff
@@ -338,9 +338,20 @@ impl PublishHandler {
...
+            // not apply to dry runs.
+            specs::add_built_specs_to_draft_specs(&spec_rows, &build_output, txn)
+                .await
+                .context("adding built specs to draft")?;
+
             return Ok((row.pub_id, JobStatus::Success));
         }
 
+        // Add built specs to the live spec when publishing a build.
+        specs::add_built_specs_to_live_specs(&spec_rows, &build_output, txn)
```
jgraettinger, line 351: I'm going back and forth on whether this should be integrated with `update_expanded_live_specs()` above.

The broader thought process is that, if a collection is updated, a materialization which sources from the collection (and which inlines its CollectionSpec) should _also_ be updated even if that materialization wasn't a draft spec of the publication. For these kinds of direct adjacencies, we _do_ want built_specs of adjacent specs to be updated.

But, currently `expanded_rows` is too over-broad. What I'd like to do, as future work, is:
 * Ensure a `built_spec` is populated for every `live_spec` row.
 * Implement a DB RPC that can be used to dynamically resolve `CollectionSpec`'s using`built_spec`, implementing the [ControlPlane::resolve_collections](https://github.com/estuary/flow/blob/master/crates/validation/src/lib.rs#L45-L85) trait.
  * _Then_ relax the graph expansion logic to only consider direct adjacencies of captures and materializations, rather than also expanding through it to all _other_ bound collections. This will significantly narrow the set of "expanded" specs.
  
Assuming we've done this ☝️ then I believe we'd _definitely_ want to update the `built_spec` of every expanded spec, in the same way that we do update their build ID today.

I'm also not aware of a hard reason that we _couldn't_ update `built_spec` of all current `expanded_rows` today, but it could result in more churn given the degree of graph (over)-expansion that we do today, and thus I 🧇.

jgraettinger (approved):
LGTM to land as-is, but I do have a comment for broader discussion below

PR #1086, opened 2023-06-19
source-sqlserver: Tweak setup instructions in documentation

willdonnelly:
**Description:**

I just noticed the SQL Server setup commands could use a bit of improvement:

* Replace the manual inserting of a row into the `flow_watermarks` table and instead just add `INSERT` to the permission grants on that table. This is the setup we currently test in CI builds, so that's what we should recommend users do.
* Remove the explicit `@capture_instance` parameter when enabling CDC on tables. We want the default `${schema}_${table}` naming here anyway so specifying it is just redundant and one more thing a user could get wrong.
* Consistent `--` commenting style instead of a mix of `--` and `/* */` comments.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1086)


oliviamiannone (approved):
lgtm, thank you!

PR #1087, opened 2023-06-19
materialize: txn size check fix

williamhbaker:
**Description:**

Fixes the materialization client so that the length of the flighted map for the binding that the document is being added to is checked for its current size vs. the maximum allowed.

This still does allow for cases where a materialization has more than 1 binding to exceed the overall txn size limit. At this point I'm not sure that is a big enough problem to worry about since we intend this to be a temporary measure.

Tested on a local stack with an artificially low limit to see that individual transactions are concluded when the limit is reached.

**Workflow steps:**

N/A

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1087)


jgraettinger (approved):
LGTM thank you!

PR #1088, opened 2023-06-20
fix: Adjacent log lines updates

jshearer:
* Remove unhelpful unique constraint on log line that was preventing faithful log reproduction
* Show all log lines, not just `error` or `warn`
* Sort adjacent log lines by timestamp so they're presented in order

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1088)


jgraettinger (approved):
LGTM

PR #1090, opened 2023-06-21
refactor flow.ShuffleRequest into proto `runtime` package and update contract

jgraettinger:
Drop flow.Shuffle and friends, in favor of runtime.ShuffleRequest & runtime.ShuffleResponse.

Remove the `Shuffle` message type abstraction, which held abstracted details of how shuffling should occur, in favor of explicitly plumbing a derivation CollectionSpec or MaterializationSpec into runtime.ShuffleRequest.

Add a `shuffle` type to the `shuffle` Go package which encapsulates these details, but the difference is that we can *change* the details without impacting the RPC contract, and the server can access any aspect of the derivation or materialization spec it may need now or in the future.

No changes to semantics or function, this is just a refactor.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1090)


go/shuffle/shuffle.go
```diff
@@ -0,0 +1,150 @@
...
+	usesLambda bool
+	// Shuffle key is the same as the source collection key.
+	usesSourceKey bool
+	// If non-nil, validate the schema on read.
+	validateSchema json.RawMessage
+}
+
+func derivationShuffles(task *pf.CollectionSpec) []shuffle {
+	if task.Derivation == nil {
+		return nil
```
psFried, line 45: Is there any reason we shouldn't panic here (and in `materializationShuffles`)? Seems like this should never be nil IRL. Perhaps just removing the check could work too.

psFried (approved):
Solid improvement 👍 LGTM

PR #1091, opened 2023-06-22
materialize: export load and store iterator total field

williamhbaker:
**Description:**

Export the `Total` field on `StoreIterator` and `LoadIterator`.  From `StoreIterator` this is useful for materializations implementing a transaction delay where the delay may not be applied if a large number of documents were stored in this transaction, which would be indicative of an on-going backfill (see https://github.com/estuary/connectors/pull/774). `LoadIterator` doesn't have an immediate use but it seems like it might be handy to have exported too.

**Workflow steps:**

N/A

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1091)


jgraettinger (approved):
LGTM

PR #1092, opened 2023-06-23
agent: fix publish bug in checking for incompatible collections

psFried:
**Description:**

I discovered a fairly major bug in the agent that causes it to crash when failing to publish due to incompatbile collections. Additionally, it was misidentifying collections that need to be re-created due to lacking the `x-infer-schema` annotation. This fixes both issues. There's an unfortunate lack of automated tests for this, but I'm thinking that perhaps a future commit could move the `BuildOutput` into the `validations` crate and be covered by the scenario tests there.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1092)

### DIFF
--- a/crates/agent/src/publications/builds.rs
+++ b/crates/agent/src/publications/builds.rs
@@ -109,7 +109,15 @@ impl BuildOutput {
                 affected_materializations,
                 requires_recreation: Vec::new(),
             };
-            if uses_schema_inference(&built.spec) {
+            // If the collection does _not_ use schema inference, then we assume
+            // that schema returned by the capture connector is authoritative.
+            // While it _might_ still be correct to handle this by only re-
+            // creating the materialization binding, it's much more common that
+            // you'd need to re-create the entire collection. For example, an
+            // ALTER TABLE statement that's run in the source database might
+            // change rows that have already been captured, and we have no way
+            // of knowing whether that's the case.
+            if !uses_schema_inference(&built.spec) {
                 ic.requires_recreation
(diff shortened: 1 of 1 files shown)


jshearer (approved):
After convo w/ @psFried, this seems correct in a bit of an indirect way. The real thing we want to do is say: "Whenever a publication changes a collection's write schema, we must start over with a fresh collection, as well as any captures and materializations that reference that collection".

The reason is that the write schema is descriptive of data in an external system that's outside of our control, so if that changes then we need to start over and look at all of it. Whereas the read schema is purely a narrowing of the write schema based on what we _think_ is likely true. 

Therefore, expanding the read schema (to something still purely a subset of the write schema) does not mean that the data has fundamentally changed shape, just that what we guess is true of its shape has changed. 

Conversely, changing the write schema means that the data may have fundamentally changed shape in the upstream system and any assumptions we've made about it so far (schema inference, columns in materializations, etc) cannot hold any longer, and must be re-evaluated from the beginning. This means a re-creation of the capture and materialization.

psFried:
@jshearer yep, that sounds right. The only (perhaps overly pedantic) correction would be that read schemas are not required to be a subset of the write schema, and in practice write schemas tend to be extremely minimal when read schemas with x-infer-schema are in use.

For those following along: we'd discussed that there's actually a further simplification we could make here. It doesn't actually matter whether a collection is using `x-infer-schema`, but only that the publication failed due to a change to the read schema, regardless of whether that change was due to schema inference or something else. I don't want to hold up this PR on that subtle refinement, so I'll plan on updating again in a subsequent PR.

PR #1095, opened 2023-06-27
parser: include file name in error message when failing to infer format

williamhbaker:
**Description:**

Includes the file name in the error message when failing to infer a format.

**Workflow steps:**

N/A

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1095)

### DIFF
--- a/crates/parser/src/format/mod.rs
+++ b/crates/parser/src/format/mod.rs
@@ -22,9 +22,9 @@ pub enum ParseError {
     MissingFormat,
 
     #[error(
-        "unable to automatically determine the format. explicit format configuration required"
+        "unable to automatically determine the format for file '{0}'. explicit format configuration required"
     )]
-    CannotInferFormat,
+    CannotInferFormat(String),
 
     #[error("unsupported format: '{0}'")]
     UnsupportedFormat(String),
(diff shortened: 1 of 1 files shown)


crates/parser/src/format/mod.rs
```diff
@@ -63,7 +63,15 @@ pub fn resolve_config(
         tracing::debug!("using user-provided format: {:?}", f);
         f.clone()
     } else {
-        let inferred = determine_format(&config).ok_or_else(|| ParseError::CannotInferFormat)?;
+        let inferred = determine_format(&config).ok_or_else(|| {
+            ParseError::CannotInferFormat(
+                config
+                    .filename
+                    .as_deref()
+                    .expect("file name must be known")
```
psFried, line 71: As originally written, the `filename` was intentionally not a required field in the config. It _might_ be the case that filename is always provided in practice, but I'm not sure. I think if we do want to require that `filename` is always defined, then we should do so by making it non-optional in the config so we can fail fast if it's not provided.

Otherwise, I'd suggest defaulting the name to `stdin` or something instead of panicking.

psFried (changes requested):
One comment I think we should address before merging

PR #1097, opened 2023-06-27
docs: fix materialization partitions example

psFried:
We recently discovered an incorrect example in the materialization docs, and this just fixes that.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1097)


jgraettinger (approved):
LGTM

PR #1099, opened 2023-07-05
feature: Update storage mappings to support `azure-ad`.

jshearer:
An `azure-ad` storage mapping looks like this:

```json
{
  "stores": [
    {
      "provider": "AZURE",
      "storage_account_name": "josephtesting",
      "container_name": "container_name",
      "prefix": "this_is_a_prefix",
      "account_tenant_id": "1212121212-1212-1212-1212-1212121219"
    }
  ]
}
```

- [x] TODO: Once https://github.com/gazette/core/pull/341 is merged, this will need to update `go.mod` to point to the new version of Gazette

Also include an update to `go.sh` that specifies how to find `CC` and `CXX` under homebrew on macos, otherwise it might use the built-in one shipped by Apple that has broken headers.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1099)


crates/models/src/journals.rs
```diff
@@ -21,8 +21,14 @@ pub struct BucketAndPrefix {
...
-    fn bucket_and_prefix(&self) -> (&str, &str) {
-        (self.bucket.as_str(), self.prefix.as_deref().unwrap_or(""))
+    fn as_url(&self, scheme: &str) -> url::Url {
+        url::Url::parse(&format!(
+            "{}://{}/{}",
+            scheme,
+            self.bucket,
+            self.prefix.as_deref().unwrap_or("")
+        ))
+        .expect("parsing as URL should never fail")
```
psFried, line 31: I don't think the `bucket` is actually guaranteed to be valid at the point when this is called. If the bucket validates against `BUCKET_RE`, then parsing should never fail. But it seems that our [validation logic](https://github.com/estuary/flow/blob/db2cdd86825132ee7e0bcac8b432712ab5866c83/crates/validation/src/storage_mapping.rs#L5) never actually validates the bucket name at all. Can you update that to add a validation?

IMO there ought to also be a comment here saying that the function may panic if the bucket and prefix have not been validated.

crates/models/src/journals.rs
```diff
@@ -33,6 +39,47 @@ impl BucketAndPrefix {
...
+    container_name: String,
+
+    /// Optional prefix of keys written to the bucket.
+    #[validate]
+    #[serde(default)]
+    prefix: Option<Prefix>,
+}
+
+impl AzureStorageConfig {
+    fn as_url(&self) -> url::Url {
```
psFried, line 62: same comment about validation applies here

psFried (approved):
LGTM % comment about storage validation. That was actually already missing before this PR, but seems worth adding here since we noticed it.

crates/models/src/journals.rs
jshearer: Good point, done!

crates/models/src/journals.rs
jshearer: Done

PR #1100, opened 2023-07-07
models: disable bindings using a boolean

psFried:
**Description:**

This changes the representation of disabled capture and materialization bindings in Flow specs.  We previously used `target: null` and `resource: null` to represent disabled capture and materialization bindings respctively.  That change was introduced quite recently and has not seen any use yet.  This commit changes both capture and materialization bindings to use `disabled: true`.

Ultimately, the reason for this is to make it easier to programatically manipulate Flow specs.  Say a user runs a discover of an existing capture via the UI. Afterward, the draft could contain a bunch of disabled bindings with `target: null` along with a corresponding collection for each disabled binding. The UI needs those collections so that it can make sure they're included in the draft if the user re-enables any of the disabled bindings. But the UI would have no idea how to match a disabled binding with its corresponding collection spec because the `target` is null. We'd have a similar problem in Flowctl. The output of a discover needs to somehow communicate the relationship between each binding and the collection that goes along with it. Rather than introduce a separate data structure containing these relationships, it seems preferable to just keep them within the specs themselves.

This ends up being a simpler, it works the same for both captures and materializations, and seems a bit more obvious and clear. It also means that disabled capture bindings would be required to have a non-null `target`.  The validation logic is skipped for bindings that are disabled, so disabled bindings can have _any_ collection name. It's a little awkward, but it seems preferable to the alternative of having a separate data structure containing the mappings from resources to collections.

**Workflow steps:**

To disable a capture binding, use:

```yaml
captures:
  acmeCo/foo:
    bindings:
    - resource: { stream: foo }
      target: acmeCo/some/collection
      disable: true
```

and for materializations:

```yaml
materializations:
  acmeCo/foo:
    bindings:
    - resource: { table: foo }
      source: acmeCo/some/collection
      disable: true
```

None yet.

**Notes for reviewers:**

I want to explicitly call out that the premise of this PR is open for debate. There's tradeoffs here, where we're using a somewhat less elegant representation (due to having a required `target` for disabled bindings), in order to make it easier to programmatically manipulate the specs. We don't _have_ to do this.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1100)

### DIFF
--- a/crates/agent/src/publications/specs.rs
+++ b/crates/agent/src/publications/specs.rs
@@ -628,7 +628,9 @@ fn extract_spec_metadata<'a>(
 
             if let Some(derivation) = &collection.derive {
                 for tdef in &derivation.transforms {
-                    reads_from.push(tdef.source.collection().as_ref());
+                    if !tdef.disable {
+                        reads_from.push(tdef.source.collection().as_ref());
+                    }
                 }
                 reads_from.reserve(1);
             }
@@ -642,7 +644,9 @@ fn extract_spec_metadata<'a>(
                 image_parts = Some(split_tag(&config.image));
             }
             for binding in &materialization.bindings {
-                reads_from.push(binding.source.collection().as_ref());
+                if !binding.disable {
+                    reads_from.push(binding.source.collection().as_ref());
+                }
             }
             reads_from.reserve(1);
         }
--- a/crates/agent/src/publications/specs.rs
+++ b/crates/agent/src/publications/specs.rs
@@ -616,8 +616,8 @@ fn extract_spec_metadata<'a>(
             image_parts = Some(split_tag(&config.image));
 
             for binding in &capture.bindings {
-                if let Some(target) = binding.target.as_ref() {
-                    writes_to.push(target.as_ref());
+                if !binding.disable {
+                    writes_to.push(binding.target.as_ref());
                 }
             }
             writes_to.reserve(1);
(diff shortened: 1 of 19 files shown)


crates/agent/src/publications/specs.rs
```diff
@@ -616,8 +616,8 @@ fn extract_spec_metadata<'a>(
             image_parts = Some(split_tag(&config.image));
 
             for binding in &capture.bindings {
-                if let Some(target) = binding.target.as_ref() {
-                    writes_to.push(target.as_ref());
+                if !binding.disable {
```
jgraettinger, line 619: do this for `reads_from` below, too? Both derivations and materializations need it.

psFried: :+1: good catch!

crates/models/src/materializations.rs
```diff
@@ -39,14 +39,18 @@ pub struct MaterializationBinding {
...
+    /// # Whether to disable the binding
+    /// Disabled bindings are inactive, and not validated.
+    #[serde(default, skip_serializing_if = "is_false")]
+    pub disable: bool,
     /// # Selected projections for this materialization.
     #[serde(default)]
     pub fields: MaterializationFields,
 }
 
 impl MaterializationBinding {
```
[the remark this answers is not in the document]
psFried: yeah, this was vestigial, and it helped readability to remove it.

PR #1101, opened 2023-07-07
refactor: support pagination in flowctl

jshearer:
**Description:**

Introduce a new `api_exec_paginated()` method that automatically reads all possible rows from a `postgrest::Builder` by using Postgrest's built-in pagination support. Previously requests that returned multiple rows would potentially get cut off if the total row count exceeded the per-request limit set in Supabase of 1000.

In order to support pagination I had to make a change to [`postgrest-rs`](https://github.com/supabase-community/postgrest-rs) to allow cloning `Builder`s. I realized we already have a [fork](https://github.com/estuary/postgrest-rs), but I didn't have permission to write to it so I just made my own. I rebased my change plus our fork ontop of upstream, and made two PRs: 
* https://github.com/supabase-community/postgrest-rs/pull/52 for a change from a while ago to add support for building but not sending a request
* https://github.com/supabase-community/postgrest-rs/pull/51 which is my change to support cloning `Builder`s.

Until those are merged, this PR changes the target repo/branch of `postgrest-rs` in `Cargo.toml` to my fork. It could just as easily be our existing fork, I just don't have permission to push to that repo and wanted to get this PR up ASAP to support a customer who needs flowctl to support >1000 row requests.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1101)

### DIFF
--- a/crates/flowctl/src/lib.rs
+++ b/crates/flowctl/src/lib.rs
@@ -23,8 +23,7 @@ mod raw;
 use output::{Output, OutputType};
 use poll::poll_while_queued;
 
-use crate::pagination::PaginationClient;
-use crate::pagination::PaginationRequest;
+use crate::pagination::into_items;
 
 /// A command-line tool for working with Estuary Flow.
 #[derive(Debug, Parser)]
@@ -220,17 +219,11 @@ where
 /// this will issue as many paginated requests as neccesary to fetch every row.
 async fn api_exec_paginated<T>(b: postgrest::Builder) -> anyhow::Result<Vec<T>>
 where
-    for<'de> T: serde::Deserialize<'de> + Send + Sync,
+    T: serde::de::DeserializeOwned + Send + Sync + 'static,
 {
     use futures::TryStreamExt;
-    use page_turner::PageTurner;
 
-    let pages = PaginationClient::<T>::new()
-        .pages(PaginationRequest::new(b))
-        .items()
-        .try_collect()
-        .await
-        .unwrap();
+    let pages = into_items(b).try_collect().await.unwrap();
 
     Ok(pages)
 }
--- a/crates/flowctl/src/lib.rs
+++ b/crates/flowctl/src/lib.rs
@@ -15,12 +15,16 @@ mod generate;
 mod local_specs;
 mod ops;
 mod output;
+mod pagination;
 mod poll;
 mod preview;
 mod raw;
 
 use output::{Output, OutputType};
 use poll::poll_while_queued;
+
+use crate::pagination::PaginationClient;
+use crate::pagination::PaginationRequest;
 
 /// A command-line tool for working with Estuary Flow.
 #[derive(Debug, Parser)]
@@ -192,7 +196,7 @@ impl Cli {
 }
 
 // api_exec runs a PostgREST request, debug-logs its request, and turns non-success status into an anyhow::Error.
-async fn api_exec<T>(b: postgrest::Builder<'_>) -> anyhow::Result<T>
+async fn api_exec<T>(b: postgrest::Builder) -> anyhow::Result<T>
 where
     for<'de> T: serde::Deserialize<'de>,
 {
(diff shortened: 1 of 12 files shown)


crates/flowctl/src/lib.rs
```diff
@@ -211,6 +215,26 @@ where
     }
 }
 
+/// Execute a [`postgrest::Builder`] request returning multiple rows. Unlike [`api_exec`]
+/// which is limited to however many rows Postgrest is configured to return in a single response,
+/// this will issue as many paginated requests as neccesary to fetch every row.
+async fn api_exec_paginated<T>(b: postgrest::Builder) -> anyhow::Result<Vec<T>>
+where
+    for<'de> T: serde::Deserialize<'de> + Send + Sync,
```
psFried, line 223: optional nit: The trait [DeserializeOwned](https://docs.rs/serde/latest/serde/de/trait.DeserializeOwned.html) exists as a more readable alternative to higher ranked trait bounds.

crates/flowctl/src/pagination.rs
```diff
@@ -0,0 +1,80 @@
...
+    type PageItem = Item;
+    type PageError = anyhow::Error;
+
+    async fn turn_page(
+        &self,
+        request: PaginationRequest,
+    ) -> PageTurnerOutput<Self, PaginationRequest> {
+        let resp: Vec<Item> = api_exec::<Vec<Item>>(request.builder.clone()).await?;
+
+        if resp.len() == request.page_size {
```
psFried, line 73: Are you familiar with the `Content-Range` header? [PostgREST seems to use it](https://postgrest.org/en/stable/references/api/tables_views.html#limits-and-pagination), although there's some caveats about the total number of items that gets reported, so maybe there's a reason to prefer the heuristic you have here.

Can you verify the behavior of the API when there's a page size of _n_ and exactly _n_ items? We'd send a last request that we'd expect to return a successful response with no items, and I just want to make sure that's what it does.

crates/flowctl/src/pagination.rs
```diff
@@ -0,0 +1,80 @@
+use std::marker::PhantomData;
+
+use page_turner::PageTurner;
```
psFried, line 3: Feel free to ignore this comment since I'm not 100% confident here, but it _feels_ like this library may not be pulling its weight vs a hand written implementation. Up to you whether that's really the case, but just thought I'd mention it in case you were on the fence.

psFried (approved):
I like the approach of having everything just use `api_exec_paginated` for now, while also adding the ability to opt in to a lazy iterator as we see the need 👍

crates/flowctl/src/pagination.rs
jshearer: `Content-Range`: Yeeeeah, I saw that, but then also saw that it seems to sometimes return an unknown total count i.e `0-19/*` instead of `0-19/50`, in which case we'd need to fall back to checking the response length anyway, so I figured it wasn't worth complicating things.

Validation:

```
├────────────────────────────────────┼────────────┼─────────────────────────────┼─────────────────────────────┼────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ 0a:1c:78:2a:b7:65:cc:00            │ 28         │ 2023-06-15T18:43:40.514676Z │ 2023-06-15T18:43:48.635952Z │ estuary/marketing/hubspot/source-hubspot                                                           │
├────────────────────────────────────┼────────────┼─────────────────────────────┼─────────────────────────────┼────────────────────────────────────────────────────────────────────────────────────────────────────┤
```

```
page_size: 28
```

Debug output:

```
$ flowctl draft develop
DEBUG flowctl::pagination: Got back a full response, progressing to the next page current_page=0 row_count=28
DEBUG flowctl::pagination: Got back a non-full response or we reached the builder's original limit so we're done current_page=1 row_count=0
Wrote 28 specifications under file:///Users/js/Documents/estuary/flow.yaml.
```

Works as expected!

crates/flowctl/src/pagination.rs
jshearer: Yeah, I did think about writing my own here. But looking at the implementation of PageTurner, it's actually doing a lot for us that I wouldn't want to re-implement. I think ideally the way this would work is we'd implement `PageTurner` for `postgrest::Builder`, but we can't do that so we have this somewhat awkward `PaginationClient` instead.

I just added `into_items()` so we don't have to deal with any of the pagination stuff directly, instead just get a stream of rows.

dyaffe:
Do we need a release to get this one out?  Would love to get it into the partners hands since they've hit a few issues and I'd like to get a win

PR #1105, opened 2023-07-12
docs: materialize-pinecone docs for new embedding input generation

williamhbaker:
**Description:**

Doc updates for `materialize-pinecone` for https://github.com/estuary/connectors/pull/786

**Workflow steps:**

These are the doc updates

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1105)

### DIFF
--- a/site/docs/reference/Connectors/materialization-connectors/pinecone.md
+++ b/site/docs/reference/Connectors/materialization-connectors/pinecone.md
@@ -25,30 +25,32 @@ To use this connector, you'll need:
 
 ## Embedding Input
 
-The materialization creates vector embeddings based on a text input from your collection. A field
-with type `string` must exist in your collection and it must be `required` in the collection schema.
+The materialization creates a vector embedding for each collection document. Its structure is based
+on the collection fields.
 
-By default the materialization connector will look for a field named `"input"` in your collection
-and use that value to create the embeddings. If you are using a
-[derivation](../../../concepts/derivations.md) to transform your source data before materializing it
-into Pinecone it may be convenient to create the derived collection with this field. Alternatively a
-[projection](../../../concepts/advanced/projections.md) can be configured for the source collection
-with the name of `"input"`. There is also an optional configuration for each binding to set an
-alternate name of the collection projection to use as the embedding input (see Bindings below).
+By default, fields of a single scalar type are including in the embedding: strings, integers,
+numbers, and booleans. You can include additional array or object type fields using [projected
+fields](../../../concepts/materialization.md#projected-fields).
+
+The text generated for the embedding has this structure, with field names and their values separated
+by newlines:
+```
+stringField: stringValue
+intField: 3
+numberField: 1.2
+boolField: false
+```
 
 ## Pinecone Record Metadata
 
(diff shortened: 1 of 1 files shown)


site/docs/reference/Connectors/materialization-connectors/pinecone.md
```diff
@@ -25,30 +25,32 @@ To use this connector, you'll need:
 
 ## Embedding Input
 
-The materialization creates vector embeddings based on a text input from your collection. A field
-with type `string` must exist in your collection and it must be `required` in the collection schema.
+The materialization creates vector embeddings based on fields in your collection.
```
oliviamiannone, line 28: this first sentence kinda makes it sound like each field gets its own embedding -- if I'm reading the rest correctly, in reality, each collection document results in one embedding. Assuming that's right, I suggest saying: "The materialization creates a vector embedding for each collection document. Its structure is based on the collection fields."

site/docs/reference/Connectors/materialization-connectors/pinecone.md
```diff
@@ -25,30 +25,32 @@ To use this connector, you'll need:
...
 
-By default the materialization connector will look for a field named `"input"` in your collection
-and use that value to create the embeddings. If you are using a
-[derivation](../../../concepts/derivations.md) to transform your source data before materializing it
-into Pinecone it may be convenient to create the derived collection with this field. Alternatively a
-[projection](../../../concepts/advanced/projections.md) can be configured for the source collection
-with the name of `"input"`. There is also an optional configuration for each binding to set an
-alternate name of the collection projection to use as the embedding input (see Bindings below).
+By default, fields of a single scalar type are including in the embedding: strings, integers,
+numbers, and booleans. Additional array or object type fields can optionally be included using
```
oliviamiannone, line 31: Stylistic nit, but FYI, it's a good idea to use active voice and refer to user as "you" whenever it feels natural. So here, saying "You can include additional array or object type fields using projected fields."

site/docs/reference/Connectors/materialization-connectors/pinecone.md
```diff
@@ -25,30 +25,32 @@ To use this connector, you'll need:
...
+[vector queries](https://www.pinecone.io/learn/vector-search-filtering/). This materialization will
+include the materialized document as a JSON string in the metadata field `flow_document` to enable
+retrieval of the document from vectors returned by Pinecone queries.
 
-Compatible fields have the following types, and do not have to be `required`:
-- `integer`
-- `number`
-- `string`
-- `boolean`
+Pinecone indexes all metadata fields by default, and it is recommended that [selective metadata
```
oliviamiannone, line 50: Another recommended reword. We can just say: "Pinecone indexes all metadata fields by default. To manage memory usage of the index, use selective metadata index to exclude the `flow_document` metadata field."

This eliminates passive voice as well as prevents us from saying "we recommend," which is something docs style guides advise against.

oliviamiannone (approved):
Left a few comments -- I feel more strongly about the first one and the other two are suggested stylistic voice changes. Other than that lgtm!

PR #1119, opened 2023-07-18
docs: Customize materialized fields

oliviamiannone:
**Description:**

Adds a new guide to the docs. This answers the common user question: "how do I get my output table fields to look how I want?" 

**Workflow steps:**

Follow the guide; determine if has the information a user would need to get the desired result.

**Notes for reviewers:**

I was having trouble testing schema inference in the UI (wasn't seeing it as an option when editing a Firestore capture) -- I wrote this given what I understand as the intended behavior.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1119)

### DIFF
--- a/site/docs/guides/customize-materialization-fields.md
+++ b/site/docs/guides/customize-materialization-fields.md
@@ -1,0 +1,125 @@
+# Customize materialized fields
+
+When you first materialize a collection to an endpoint like a database or data warehouse,
+the resulting table columns might not be formatted how you want.
+You might notice missing columns, extra columns, or columns with names you don't like.
+This happens when the default collection schema doesn't map to a table schema appropriate for your use case.
+
+You can control the shape and appearance of materialized tables using a two-step process.
+
+First, you modify the source collection **schema**.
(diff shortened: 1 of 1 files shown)


site/docs/guides/customize-materialization-fields.md
```diff
@@ -0,0 +1,135 @@
+# Customize materialized fields
+
+When you first materialize a collection to an endpoint like a database or data warehouse,
+the resulting table columns might not be formatted how you want.
+You might notice missing columns, extra columns, or columns with names you don't like.
+This happens when the default collection schema doesn't map to a table schema appropriate for your use case.
+
+You can control the shape and appearance of materialized tables using a two-step process.
```
psFried, line 8: I'm not sure exactly where this should go, but I feel like there should be something to the effect of "if you just want to add a field that isn't being included by default, then skip ahead to 'include desired fields in your materialization'(linky)"

psFried (approved):
Thanks! I had a few minor nit picks, but overall this looks great and will be something we can just update as our process improves.

PR #1121, opened 2023-07-19
docs: Add setup instructions for Azure AD storage mappings

jshearer:
**Description:**

Adds detailed setup instructions for gathering the required data to set up an Azure storage mapping.

* `/concepts/storage-mappings/azure-blob`

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1121)


oliviamiannone (commented):
Thanks for the detailed docs on this! The component where you input your tenant is particularly cool. A couple things before we can get this merged...
* The page should probably be moved. Right now, information on configuring storage mappings lives [here](https://docs.estuary.dev/getting-started/installation/#configuring-your-cloud-storage-bucket-for-use-with-flow). I think with this addition, it makes sense to break that section out into a separate page, but it should definitely be under the Get Started heading. 
* The level of detail could in _some_ areas be reduced. A good rule of thumb when describing a third party system (since we don't have an large docs team!) is: if they have docs that describe a workflow, link out to it. That way when they make changes we don't end up with incorrect docs we have to update. So for example here, for the authorization piece, you could link to [these docs](https://learn.microsoft.com/en-us/azure/role-based-access-control/role-assignments-portal#step-2-open-the-add-role-assignment-page). You could link on the word "grant" and instead of the bullets just leave the sentence "When you select members, search for `Estuary Storage Mappings Prod`. Be sure to search for the exact name...."
* LIkewise, the screenshots are super helpful, we'll just need to keep an eye on them as Azure could change something when we're not looking.
* I definitely love the "Give us a ring" section at the end-- when this page is combined with the other two types of storage buckets we should definitely keep that!

oliviamiannone (approved):
@jshearer Actually, comments above still hold but I want to change this to approved -- I imagine you're juggling several projects so this way, if you want to just merge it, I can take care of the edits after the fact

jshearer:
@oliviamiannone Awesome, thank you! I moved the page, and I'll do like you suggest and merge this. If you wouldn't mind making whatever edits you think should be made, that would be fantastic.

PR #1125, opened 2023-07-25
Phil/flowctl suggest schema fix

psFried:
**Description:**

This rolls up a couple of bug fixes for the `flowctl raw suggest-schema` subcommand.
Additionally, the second commit just runs rustfmt on the `flowctl` crate, since my editor runs it automatically and otherwise it was annoying to ignore those change.

I've tested the specific case that was failing, and it's now working with these changes.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1125)

### DIFF
--- a/crates/flowctl/src/raw/suggest_schema.rs
+++ b/crates/flowctl/src/raw/suggest_schema.rs
@@ -58,10 +58,7 @@ pub struct SuggestSchema {
 
 pub async fn do_suggest_schema(
     ctx: &mut crate::CliContext,
-    SuggestSchema {
-        args,
-        task,
-    }: &SuggestSchema,
+    SuggestSchema { args, task }: &SuggestSchema,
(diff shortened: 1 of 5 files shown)


crates/flowctl/src/raw/suggest_schema.rs
```diff
@@ -149,7 +149,11 @@ pub async fn do_suggest_schema(
         };
 
     // The original collection schema to be used as the starting point of schema-inference
-    let schema_model = collection_def.schema.unwrap();
+    let schema_model = collection_def
+        .schema
+        .as_ref()
+        .or(collection_def.read_schema.as_ref())
```
mdibaiee, line 155: If we are looking at a materializatio ntask, this should be a `write_schema` though, right?

mdibaiee (approved):

PR #1126, opened 2023-07-26
Schema inference: `Shape::widen()`

jshearer:
Implement `Shape::widen` which enables widening a `Shape` to fit a provided `AsNode`. This sets the groundwork for permanently keeping track of a running inferred schema in the combiner.

Of note, the schemas "inferred" by `widen()` are maximally strict:
* By default, newly inferred objects have have `additionalProperties: false`.
* Object fields initially have `required: true` until we encounter a document missing that field, at which point it will be downgraded to `required: false`. **Note:** This isn't required AFAICT. Should we really do this?

The next piece of work is implementing the stubbed-out `enforce_field_count_limits`. With that, we should have everything we need to implement the running inferred schema and emit it to the ops logs. Also of note is that the `reduce: flow-inferred-schema-merge` reduction annotation implementation can and should also use `enforce_field_count_limits`, as it's possible for itra-transaction documents to not exceed the limits while inter-transaction documents do, and we care about limiting both of those cases.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1126)

### DIFF
--- a/crates/doc/Cargo.toml
+++ b/crates/doc/Cargo.toml
@@ -43,6 +43,7 @@ rand_distr = { workspace = true }
 serde-transcode = { workspace = true }
 serde_json = { workspace = true }
 serde_yaml = { workspace = true }
+pretty_assertions = { workspace = true }
 
 [features]
 default = ["combine"]
--- a/crates/doc/Cargo.toml
+++ b/crates/doc/Cargo.toml
@@ -36,6 +36,7 @@ allocator = { path = "../allocator" }
 criterion = { workspace = true }
 hexdump = { workspace = true }
 insta = { workspace = true }
+pretty_assertions = { workspace = true }
 quickcheck = { workspace = true }
 quickcheck_macros = { workspace = true }
 rand = { workspace = true }
@@ -43,7 +44,6 @@ rand_distr = { workspace = true }
 serde-transcode = { workspace = true }
 serde_json = { workspace = true }
 serde_yaml = { workspace = true }
-pretty_assertions = { workspace = true }
 
 [features]
 default = ["combine"]
--- a/crates/doc/src/inference.rs
+++ b/crates/doc/src/inference.rs
@@ -1745,14 +1745,14 @@ impl Shape {
                 // As part of squashing all known property shapes together into
                 // additionalProperties, we need to also remove those explicit properties.
                 .drain(..)
-                .fold(existing_additional_properties, |accum, prop| {
+                .fold(existing_additional_properties, |accum, mut prop| {
+                    // Recur here to avoid excessively large `additionalProperties` shapes
+                    Shape::enforce_field_count_limits(&mut prop.shape, loc.push_prop(&prop.name));
                     Shape::union(accum, prop.shape)
                 });
 
             self.object.additional = Some(Box::new(merged_additional_properties));
         } else {
-            // We don't need to recur above because we already wiped out all
-            // existing properties, including nested ones.
             for prop in self.object.properties.iter_mut() {
                 prop.shape
                     .enforce_field_count_limits(loc.push_prop(&prop.name))
@@ -1834,6 +1834,38 @@ mod test {
                 type: integer
             "#,
             dynamic_keys.as_slice(),
+            true,
+        );
+    }
+
+    #[test]
+    fn test_field_count_nested() {
+        // Create an object like
+        // {
+        //    "big_key": {
+        //        ...751 properties...
+        //    },
+        //    ...750 more properties...
+        // }
+        let mut root = Map::default();
+        for id in 0..800 {
+            root.insert(format!("key-{id}"), json!(id * 5));
+        }
+
+        root.insert("big_key".to_string(), json!(root.clone()));
+
+        widening_snapshot_helper(
+            None,
+            r#"
+            type: object
+            additionalProperties:
+                anyOf:
+                    - type: integer
+                    - type: object
+                      additionalProperties:
+                        type: integer
+            "#,
+            &[ser::to_string(&root).unwrap()],
             true,
         );
     }
--- a/crates/doc/src/inference.rs
+++ b/crates/doc/src/inference.rs
@@ -7,7 +7,7 @@ use json::{
     LocatedProperty, Location,
 };
 use serde_json::Value;
-use std::collections::BTreeMap;
+use std::{collections::BTreeMap, ops::Deref};
 use url::Url;
 
 #[derive(Clone, Debug, Eq, PartialEq)]
@@ -1655,7 +1655,56 @@ impl Shape {
 
     // Prune any locations in this shape that have more than the allowed fields,
     // squashing those fields into the `additionalProperties` subschema for that location.
-    fn enforce_field_count_limits(&mut self) {}
+    fn enforce_field_count_limits(&mut self, loc: Location) {
+        // TODO: If we implement inference/widening of array tuple shapes
+        // then we'll need to also check that those aren't excessively large.
+        if !self.type_.overlaps(types::OBJECT) {
+            return;
+        }
+
+        let limit = match loc {
+            Location::Root => MAX_ROOT_FIELDS,
+            _ => MAX_NESTED_FIELDS,
+        };
+
+        if self.object.properties.len() > limit {
+            // Take all of the properties' shapes and
+            // union them into additionalProperties
+
+            let existing_additional_properties = self
+                .object
+                .additional
+                // `Shape::union` takes owned Shapes which is why we
+                // have to take ownership here.
+                .take()
+                .map(|boxed| *boxed)
+                .unwrap_or(Shape {
+                    // Start out maximally constrained
+                    type_: types::INVALID,
+                    provenance: Provenance::Inline,
+                    ..Default::default()
+                });
+
+            let merged_additional_properties = self
+                .object
+                .properties
+                // As part of squashing all known property shapes together into
+                // additionalProperties, we need to also remove those explicit properties.
+                .drain(..)
+                .fold(existing_additional_properties, |accum, prop| {
+                    Shape::union(accum, prop.shape)
+                });
+
+            self.object.additional = Some(Box::new(merged_additional_properties));
+        } else {
+            // We don't need to recur above because we already wiped out all
+            // existing properties, including nested ones.
+            for prop in self.object.properties.iter_mut() {
+                prop.shape
+                    .enforce_field_count_limits(loc.push_prop(&prop.name))
+            }
+        }
+    }
 }
 
 /// Returns true if the text is a match for the given regex. This function exists primarily so we
@@ -1677,14 +1726,15 @@ mod test {
     use json::schema::{self, index::IndexBuilder};
     #[cfg(test)]
     use pretty_assertions::assert_eq;
-    use serde_json::{de, json, Value};
+    use serde_json::{de, json, ser, Map, Value};
     use serde_yaml;
 
-    fn widening_snapshot_helper(
+    fn widening_snapshot_helper<T: AsRef<str>>(
         initial_schema: Option<&str>,
         expected_schema: &str,
-        docs: &[&str],
-    ) {
+        docs: &[T],
+        enforce_limits: bool,
+    ) -> Shape {
         let mut schema = match initial_schema {
             Some(initial) => shape_from(initial),
             None => Shape {
@@ -1695,13 +1745,119 @@ mod test {
         };
 
         for doc in docs {
-            let val: Value = de::from_str(doc).unwrap();
+            let val: Value = de::from_str(doc.as_ref()).unwrap();
             schema.widen(&val, Location::Root);
         }
 
         let expected = shape_from(expected_schema);
 
+        if enforce_limits {
+            schema.enforce_field_count_limits(Location::Root);
+        }
+
         assert_eq!(expected, schema);
+
+        schema
+    }
+
+    #[test]
+    fn test_field_count_limits() {
+        let dynamic_keys = (0..800)
+            .map(|id| {
+                ser::to_string(&json!({
+                    "known_key": id,
+                    format!("key-{id}"): id*5
+                }))
+                .unwrap()
+            })
+            .collect_vec();
+
+        widening_snapshot_helper(
+            None,
+            r#"
+            type: object
+            additionalProperties:
+                type: [number]
+            "#,
+            dynamic_keys.as_slice(),
+            true,
+        );
+    }
+
+    #[test]
+    fn test_field_count_limits_noop() {
+        let dynamic_keys = (0..1)
+            .map(|id| {
+                ser::to_string(&json!({
+                    "known_key": id,
+                    format!("key-{id}"): id*5
+                }))
+                .unwrap()
+            })
+            .collect_vec();
+
+        widening_snapshot_helper(
+            None,
+            r#"
+            type: object
+            additionalProperties: false
+            required: [known_key, key-0]
+            properties:
+                known_key:
+                    type: number
+                key-0:
+                    type: number
+            "#,
+            dynamic_keys.as_slice(),
+            true,
+        );
+    }
+
+    #[test]
+    fn test_field_count_limits_nested() {
+        let mut nested = Map::default();
+        for id in 0..1 {
+            nested.insert(format!("key-{id}"), json!(id * 5));
+        }
+
+        widening_snapshot_helper(
+            None,
+            r#"
+            type: object
+            additionalProperties: false
+            required: [container]
+            properties:
+                container:
+                    type: object
+                    additionalProperties: false
+                    required: [key-0]
+                    properties:
+                        key-0:
+                            type: number
+            "#,
+            &[&ser::to_string(&json!({ "container": nested })).unwrap()],
+            true,
+        );
+
+        for id in 0..300 {
+            nested.insert(format!("key-{id}"), json!(id * 5));
+        }
+
+        widening_snapshot_helper(
+            None,
+            r#"
+            type: object
+            additionalProperties: false
+            required: [container]
+            properties:
+                container:
+                    type: object
+                    additionalProperties:
+                        type: [number]
+            "#,
+            &[&ser::to_string(&json!({ "container": nested })).unwrap()],
+            true,
+        );
     }
 
     #[test]
@@ -1722,6 +1878,7 @@ mod test {
                             type: string
             "#,
             &[r#"{"test_key": {"test_nested": "pizza"}}"#],
+            false,
         );
     }
 
@@ -1739,6 +1896,7 @@ mod test {
                     type: string
             "#,
             &[r#"{"first_key": "hello"}"#],
+            false,
         );
         // Fields encountered after the first should not be required
         // AND required fields should stay required, so long as they
@@ -1765,6 +1923,7 @@ mod test {
                     type: string
             "#,
             &[r#"{"first_key": "hello", "second_key": "goodbye"}"#],
+            false,
         );
         // Required fields get demoted once we encounter a document
         // where they are not present
@@ -1791,6 +1950,7 @@ mod test {
                     type: string
             "#,
             &[r#"{"second_key": "goodbye"}"#],
+            false,
         );
     }
 
@@ -1814,6 +1974,7 @@ mod test {
             Some(schema),
             schema,
             &[r#"{"test_key": {"test_nested": "pizza"}}"#],
+            false,
         );
     }
 
@@ -1850,6 +2011,7 @@ mod test {
                                 type: number
                 "#,
             &[r#"{"test_key": {"nested_second": 68}}"#],
+            false,
         );
     }
 
@@ -1868,6 +2030,7 @@ mod test {
                 r#"{"test_key": "a_string"}"#,
                 r#"{"toast_key": "another_string"}"#,
             ],
+            false,
         );
     }
 
@@ -1886,6 +2049,7 @@ mod test {
                 type: [string, number]
             "#,
             &[r#"{"test_key": "a_string"}"#, r#"{"toast_key": 5}"#],
+            false,
         );
     }
 
@@ -1916,6 +2080,7 @@ mod test {
                                 type: [string, number]
                 "#,
             &[r#"{"test_key": {"test_nested": 68}}"#],
+            false,
         );
     }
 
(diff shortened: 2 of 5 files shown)


crates/doc/src/inference.rs
```diff
@@ -1484,6 +1486,201 @@ impl Shape {
...
+    // NOTE: If a particular location defines `additionalProperties` as a subschema, don't
+    // add the field from `AsNode`, instead jump straight to squashing the the field into `additionalProperties`
+    fn widen<'n, N>(&mut self, node: &'n N) -> bool
+    where
+        N: AsNode,
+    {
+        self.widen_inner(node, Location::Root)
+    }
+
+    fn widen_inner<'n, N>(&mut self, node: &'n N, loc: Location) -> bool
```
jgraettinger, line 1504: I'm thinking that this should be `widen`, and the current `widen` removed? Passing `Location::Root` as a caller is easy enough

crates/doc/src/inference.rs
```diff
@@ -1484,6 +1486,201 @@ impl Shape {
...
+    where
+        N: AsNode,
+    {
+        use crate::{Field, Fields};
+
+        match node.as_node() {
+            crate::Node::Object(obj) => {
+                self.type_ = self.type_ | types::OBJECT;
```
jgraettinger, line 1512: should this be an impl of ObjShape ? There's a lot happening / a lot of cyclomatic nesting in this routine

crates/doc/src/inference.rs
```diff
@@ -1484,6 +1486,201 @@ impl Shape {
...
+                        self.object.additional = Some(Box::new(Shape {
+                            type_: types::INVALID,
+                            ..Default::default()
+                        }))
+                    }
+                }
+
+                // If a particular location defines `additionalProperties` as a subschema
+                // don't add the field from `AsNode`, instead jump straight to squashing
+                let hint = match self.object.additional.as_mut() {
```
jgraettinger, line 1526: nit: if this were a separate routine, you could early-return here and reduce the following indentation a couple of levels. Have pity on us that code-review on a 13-inch screen 😀

crates/doc/src/inference.rs
```diff
@@ -1484,6 +1486,201 @@ impl Shape {
...
+                    // We only want to squash if your `additionalProperties`
+                    // is set to something other than `false`.
+                    Some(addl) if !addl.type_.eq(&types::INVALID) => {
+                        obj.iter().enumerate().fold(false, |accum, (idx, node)| {
+                            accum || addl.widen_inner(node.value(), loc.push_item(idx))
```
jgraettinger, line 1531: this should be `loc.push_property`. `push_item` is only for arrays.

crates/doc/src/inference.rs
```diff
@@ -1484,6 +1486,201 @@ impl Shape {
...
+                        })
+                    }
+                    _ => {
+                        let (hint, new_obj_shape) = itertools::merge_join_by(
+                            self.object.properties.iter_mut(),
+                            obj.iter(),
+                            |prop, field| prop.name.cmp(&field.property().to_string()),
+                        )
+                        .fold(
```
jgraettinger, line 1540: How about, instead of this fold, you instead `filter_map()` and `collect()` a vector of entirely-new properties which should be _added_ to `object.properties` after the iteration completes?

This does a couple things, the biggest of which is that it avoids _any_ allocation in the common case where no new properties are added, because this vector is empty:

* EoB::Left would mark !required in-place and return None
* EoB::Right would return Some((property, property-shape))
* EoB::Both would widen the child in-place and return None.

Collect the iteration result into a Vec, which releases the mutable borrow, and then `extend()` the LHS map with it's `into_iter()`. If there are no new properties, this is zero-cost.

crates/doc/src/inference.rs
```diff
@@ -1484,6 +1486,201 @@ impl Shape {
...
+                                EitherOrBoth::Both(lhs, rhs) => (
+                                    accum
+                                        || lhs.shape.widen_inner(
+                                            rhs.value(),
+                                            loc.push_prop(rhs.property()),
+                                        ),
+                                    new_obj_shape,
+                                ),
+                                // Shape has a field that the node is missing, so let's make sure it's not marked as required
+                                // Is this the right behavior here?
```
jgraettinger, line 1566: Yes, this is right.

crates/doc/src/inference.rs
```diff
@@ -1484,6 +1486,201 @@ impl Shape {
...
+                                EitherOrBoth::Left(mut lhs) => {
+                                    lhs.is_required = false;
+                                    (accum, new_obj_shape)
+                                }
+                                // The Node has a field that the shape doesn't, let's add it
+                                EitherOrBoth::Right(rhs) => {
+                                    let mut prop = ObjProperty {
+                                        name: rhs.property().to_owned(),
+                                        // Should we really do this?
```
jgraettinger, line 1575: no, it's not required with one exception, which is when we're merging in the very first instance, which needs to be accounted for. Also do you have a test on this case?

crates/doc/src/inference.rs
```diff
@@ -1484,6 +1486,201 @@ impl Shape {
...
+                };
+
+                match (hint, loc) {
+                    (true, _) => true,
+                    (false, Location::Root) => self.object.properties.len() > MAX_ROOT_FIELDS,
+                    (false, _) => self.object.properties.len() > MAX_NESTED_FIELDS,
+                }
+            }
+
+            crate::Node::Array(arr) => {
```
jgraettinger, line 1624: marking that this is as far as i got

psFried:
What's the plan for determining whether a shape was actually modified by calls to `widen`? Are we keeping two copies and doing an equality comparison? It'd be nice if `widen` could return a `bool` indicating whether or not it was actually mutated by the operation, though I'm also recognizing that that approach might add to an already considerable level of complexity in that function.

We need this in order to determine whether or not to actually emit the new schema after draining the combiner. Totally fine to implement it in a subsequent PR, IMO, as long as we have a plan for it.

crates/doc/src/inference.rs
jshearer: Yep, good point. I had originally had more setup stuff in `widen()`. Removed it

jshearer:
> What's the plan for determining whether a shape was actually modified by calls to widen? 

I'm inclined to keep it simple for the moment and give each transaction a copy of the running shape, which we can then compare after the transaction commits. Yes it's less efficient than keeping track of whether the shape was modified or not, but since it only happens once every transaction vs once for every document, I don't think it'd be that painful performance-wise.

crates/doc/src/inference.rs
jshearer: Yup, way better. Moved 👍

crates/doc/src/inference.rs
jshearer: This is in the `Node::Array` match arm, so I figured `push_item` was correct since it took a numeric index. But since we're not actually widening tuple shapes, I'm realizing now this index should probably always be `0`... or something? It's not a huge deal because the Location is only used at present to detect if we're at `Location::Root` or not, but if we have it I'd like it to be as correct as possible

crates/doc/src/inference.rs
jshearer: This is a great tip, it's much cleaner. Thanks!

crates/doc/src/inference.rs
jshearer: We talked, I understand what this means now: by definition, only properties that we have encountered on _every_ object should be marked as required. Therefore, only properties we encounter the _very first time_ we see an object at a particular location should get `is_required: true`, since any other time the implication is that we have not seen this property before, i.e we've seen documents without it, and therefore it should not be marked as required.

crates/doc/src/inference.rs
jshearer: Factored out to `ObjShape::widen` 👍

crates/doc/src/inference.rs
```diff
@@ -444,6 +445,82 @@ impl ObjShape {
...
+        N: AsNode,
+    {
+        use crate::{Field, Fields};
+        // If a particular location defines `additionalProperties` as a subschema
+        // don't add the field from `AsNode`, instead jump straight to squashing
+        match self.additional.as_mut() {
+            // We only want to squash if your `additionalProperties`
+            // is set to something other than `false`.
+            Some(addl) if !addl.type_.eq(&types::INVALID) => {
+                fields.iter().enumerate().fold(false, |accum, (idx, node)| {
```
jgraettinger, line 460: nit: how about returning early here, ending the `match` block, so that what's currently the `_ =>` match arm can be de-indented 2x:

```
// If `additionalProperties` is anything other than `false`, then widen it with all `fields`.
if let Some(addl) = self.additional.as_mut() {
  if addl.type_ != types::INVALID {
    return fields.iter().enumerate().fold(false, |accum, (idx, node)| {
        accum || addl.widen(node.value(), loc.push_item(idx))
    })
  }
}

// Handle fields individually ...

```

crates/doc/src/inference.rs
```diff
@@ -444,6 +445,82 @@ impl ObjShape {
...
+                    accum || addl.widen(node.value(), loc.push_item(idx))
```
jgraettinger, line 461: this is still incorrect usage -- it should be `push_property(node.property)`, as we're traversing RHS object properties, not RHS array indices.

crates/doc/src/inference.rs
```diff
@@ -444,6 +445,82 @@ impl ObjShape {
...
+                })
+            }
+            _ => {
+                let mut hint = false;
+
+                let shapes: Vec<_> = itertools::merge_join_by(
```
jgraettinger, line 467: nit: `shapes` => `new_properties` or `new_fields` ?

crates/doc/src/inference.rs
```diff
@@ -444,6 +445,82 @@ impl ObjShape {
...
+                            // A field can only be required if every single document we've seen
+                            // has that field present. This means that ONLY fields that exist
+                            // on the very first object we encounter for a particular location should
+                            // get marked as required, as any subsequent "new" fields
+                            // by definition did not exist on previous objects (and so cannot be required)
+                            // otherwise they would already be in the Shape
+                            // (and we would be in the `EoB::Both` branch).
+                            is_required: is_first_time,
+                            // Leave shape blank here, we're going to recur and expand it right below
+                            // Note: Shape starts out totally unconstrained (types::ANY) by default,
```
jgraettinger, line 496: no action: We should probably remove / refactor the `Default` and `new` impls -- they don't make logical sense, and are used only for tests.

crates/doc/src/inference.rs
```diff
@@ -1484,6 +1561,152 @@ impl Shape {
...
+                    self.object.additional = Some(Box::new(Shape {
+                        type_: types::INVALID,
+                        provenance: Provenance::Inline,
+                        ..Default::default()
+                    }))
+                }
+
+                let hint = self.object.widen::<N>(fields, loc, is_first_time);
+
+                match (hint, loc) {
```
jgraettinger, line 1594: nit: does this belong in ObjShape::widen ?

crates/doc/src/inference.rs
```diff
@@ -1484,6 +1561,152 @@ impl Shape {
...
+                    provenance: Provenance::Inline,
+                    ..Default::default()
+                }));
+
+                // Look at each element in the observed array and widen the shape to accept it
+                let hint = arr.iter().enumerate().fold(false, |accum, (idx, node)| {
+                    accum || shape.widen(node, loc.push_item(idx))
+                });
+
+                self.array.additional = Some(shape);
```
jgraettinger, line 1614: should we update `min` / `max` here?

crates/doc/src/inference.rs
```diff
@@ -1484,6 +1561,152 @@ impl Shape {
...
+                };
+
+                self.string = StringShape::union(self.string.clone(), partial_stringshape);
+                false
+            }
+            crate::Node::Null => {
+                self.type_ = self.type_ | types::NULL;
+                false
+            }
+            crate::Node::Number(_num) => {
```
jgraettinger, line 1638: you need to match over the _kind_ of number and set type_ appropriately.

See [Set::for_value](https://github.com/estuary/flow/blob/master/crates/json/src/schema/types.rs#L184) as currently used by [schema-inference](https://github.com/estuary/flow/blob/108864c33849e689da48a81f8ff7b65e02372f08/crates/schema-inference/src/inference.rs#L8) (and please double-check there isn't other handling from that implementation, which parallels this one, that we're accidentally missing).

crates/doc/src/inference.rs
```diff
@@ -1484,6 +1561,152 @@ impl Shape {
...
+                self.type_ = self.type_ | types::INT_OR_FRAC;
+                false
+            }
+            crate::Node::String(_s) => {
+                self.type_ = self.type_ | types::STRING;
+                let partial_stringshape = StringShape {
+                    // TODO: implement format widening
```
jgraettinger, line 1645: this needs doing to before we release for real. see [infer_string_shape](https://github.com/estuary/flow/blob/108864c33849e689da48a81f8ff7b65e02372f08/crates/schema-inference/src/inference.rs#L23C5-L37).

Though I am concerned about the performance aspect of these checks and we need to be careful.
Perhaps we always match over format on the very first string, and thereafter we match on format if and only if the LHS format is still set. Put differently, as soon as LHS `format` is unconstrained we stop checking the RHS.

We need test on this also -- perhaps ripped from `schema-inference` crate.


Also update min_length / max_length?

crates/doc/src/inference.rs
```diff
@@ -1484,6 +1561,152 @@ impl Shape {
...
+            }
+        }
+    }
+
+    // Prune any locations in this shape that have more than the allowed fields,
+    // squashing those fields into the `additionalProperties` subschema for that location.
+    pub fn enforce_field_count_limits(&mut self, loc: Location) {
+        // TODO: If we implement inference/widening of array tuple shapes
+        // then we'll need to also check that those aren't excessively large.
+        if !self.type_.overlaps(types::OBJECT) {
```
jgraettinger, line 1661: you still need to recurse for cases like `{"arr": [{"so":1, "many":2, "props":3}]}`

crates/doc/src/inference.rs
```diff
@@ -1484,6 +1561,152 @@ impl Shape {
...
+                // As part of squashing all known property shapes together into
+                // additionalProperties, we need to also remove those explicit properties.
+                .drain(..)
+                .fold(existing_additional_properties, |accum, prop| {
+                    Shape::union(accum, prop.shape)
+                });
+
+            self.object.additional = Some(Box::new(merged_additional_properties));
+        } else {
+            // We don't need to recur above because we already wiped out all
```
jgraettinger, line 1700: yes we do. i could have a lot of properties which are all objects, also having a lot of properties.

jgraettinger (commented):
More feedback for you

crates/doc/Cargo.toml
```diff
@@ -43,8 +43,9 @@ rand_distr = { workspace = true }
 serde-transcode = { workspace = true }
 serde_json = { workspace = true }
 serde_yaml = { workspace = true }
+pretty_assertions = { workspace = true }
```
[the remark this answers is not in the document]
jshearer: Oh ffs. I sorted it in the root, not here. thx

crates/doc/src/inference.rs
jshearer: Fair enough. I had tried this originally and ran into https://github.com/rust-lang/rust/issues/53667 and thought the nested if statements were weird, but looking at this again it's still cleaner because of the early return. Changed 👍

crates/doc/src/inference.rs
jshearer: Ah yep. On this topic though, is my comment [here](https://github.com/estuary/flow/pull/1126#discussion_r1276654169) right? I.e, if it really _is_ an array index, I should be using `push_item(index)`, right?

crates/doc/src/inference.rs
jshearer: Yeah, I think it does

crates/doc/src/inference.rs
jshearer: Yep, done

crates/doc/src/inference.rs
jshearer: Done! I'm not sure what you mean by checking for "other handling from that implementation"?

jshearer: Done! I'm not sure what you mean by checking for "other handling from that implementation"? It looks like `schema-inference` just calls `Set::for_value` directly

crates/doc/src/inference.rs
jshearer: This makes sense. I ended up constraining it a little bit more: only first-time strings get the full inference, then subsequent ones get validated against the format and if they don't match, lose the format.

crates/doc/src/inference.rs
jshearer: Good point!

crates/doc/src/inference.rs
jshearer: Neat, didn't think of this. Added a failing test and fixed it 👍

crates/doc/src/inference.rs
jgraettinger: That was also the Node::Object match arm.

but yes if it is an array push_item

crates/doc/src/inference.rs
jgraettinger: i meant 👀  for other things that `schema-inference` may be doing in it's own "widen"-equivalent operation, that aren't reflected here and should be.

crates/doc/src/inference.rs
jgraettinger: No, because there can be more nuanced handling for `format` in StringShape::union.

One current example is widening `format: integer` => `format: number`.

jshearer: Ah yep, good point. Changed and tested 👍

crates/doc/src/inference.rs
jshearer: I took a look over at `schema-inference` and didn't see it doing anything except calling `Set::for_value`, so I think this is it 👍

crates/doc/src/inference.rs
```diff
@@ -444,6 +460,84 @@ impl ObjShape {
             None
         }
     }
+
+    fn widen<'n, N>(&mut self, fields: &'n N::Fields, loc: Location, is_first_time: bool) -> bool
+    where
+        N: AsNode,
+    {
+        use crate::{Field, Fields};
+        // If a particular location defines `additionalProperties` as a subschema
```
jgraettinger, line 469: ~nit:~ this comment could be a bit crisper. For example:

```
Two possible cases:
* This schema has individual `properties` as well as `additionalProperties: false`,
   meaning no other properties are possible. We recursively widen each of the input
  `fields` into their respective `properties`, adding new ones as required.
* This schema has `additionalProperties` _other_ than `false`.
   In this case, we apply each of `fields` to widen matched `properties`,
   and otherwise widen `additionalProperties` where not matched.
```

Right, and this is the subtle power of good comments. Writing this out made me realize that our behavior is incorrect in the general case, where we're widening an `ObjShape` that may have started as an actual schema.

To be fully correct, we _would_ need to look for matched `properties` and even `patternProperties` to widen _before_ we resort to widening `additionalProperties`.

In your shoes I would correct the implementation now -- because I'll otherwise forget about it -- but I'm okay punting given the time pressure for this feature. If we do punt, please add a good comment here and issue on this defect.

crates/doc/src/inference.rs
```diff
@@ -1501,9 +1771,657 @@ fn regex_matches(re: &fancy_regex::Regex, text: &str) -> bool {
 mod test {
     use super::{super::Annotation, *};
     use json::schema::{self, index::IndexBuilder};
-    use serde_json::{json, Value};
+    #[cfg(test)]
```
jgraettinger, line 1774: nit: cfg(test) is redundant

crates/doc/src/inference.rs
```diff
@@ -1501,9 +1771,657 @@ fn regex_matches(re: &fancy_regex::Regex, text: &str) -> bool {
...
+    use serde_json::{de, json, ser, Map, Value};
     use serde_yaml;
 
+    fn widening_snapshot_helper<T: AsRef<str>>(
+        initial_schema: Option<&str>,
+        expected_schema: &str,
+        docs: &[T],
+        enforce_limits: bool,
+    ) -> Shape {
+        let mut schema = match initial_schema {
```
jgraettinger, line 1785: nit: it's a smell that we're repeating this all over the place. This should be a function.

jgraettinger (approved):
LGTM % comments below

We can land this, but the next big question is, what does this do to performance? This is adding more overhead in an area that's already critical path, so we'll have to accept some regression, but do need to quantify "how much?".

[There's a combiner benchmark](https://github.com/estuary/flow/blob/master/crates/doc/tests/combiner_perf.rs) that models a parameterized mix of real-world documents and can be further extended to drive `widen` and quantify some of this impact.

crates/doc/src/inference.rs
jshearer: Fixed and tested that:
* We widen explicitly named properties even if `additionalProperties` is non-false
* We widen matching `patternProperties` before widening `additionalProperties`

crates/doc/src/inference.rs
```diff
@@ -1484,6 +1578,182 @@ impl Shape {
     }
 }
 
+const MAX_ROOT_FIELDS: usize = 750;
+const MAX_NESTED_FIELDS: usize = 200;
+
+impl Shape {
+    // Widen a Shape to make the provided AsNode value fit.
+    // Returns a hint if some locations might exceed their maximum allowable size.
+    // NOTE: If a particular location defines `additionalProperties` as a subschema, don't
```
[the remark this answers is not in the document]
jshearer: I actually decided to move the docstring from `ObjShape::widen` here instead since this is a public function so it'll be more visible and explains what's going on better.

crates/doc/src/inference.rs
jshearer: Oops, yup my bad

crates/doc/src/inference.rs
jshearer: Yuuuup, thanks for pointing this out. Extracted into `Shape::invalid()`.

PR #1127, opened 2023-07-27
connector-init: start listening on tcp port before check_protocol

mdibaiee:
**Description:**

- With this change, we start listening on the tcp port in `connector-init` as soon as possible, but we don't process the requests while we check the protocol. After we have checked the protocol, we use the incoming stream to serve requests.
- This allows clients to connect to the server earlier and write to it, in a sense buffering some request(s)
- This is motivated by our code in `flowctl`'s `docker_run` not being able to clearly distinguish between a port that it can connect to, and a port that is ready for data to be written to it [here](https://github.com/estuary/flow/blob/master/crates/flowctl/src/connector.rs#L82). It's still a mystery to me how the go implementation of the same connection works fine and waits until the grpc server actually starts processing before it considers the connection as ready. The relevant go code is [here](https://github.com/grpc/grpc-go/blob/5ce5686d5e7111d8336f4a0d8b3036c1439e82b1/clientconn.go#L275-L300) and Tonic seems to conceptually check for readiness [here](https://github.com/estuary/flow/blob/master/crates/proto-grpc/src/capture.rs#L142), however it seems it has a different formulation of readiness.

I still don’t love this, because I would like to understand why the tonic client considers the connection ready and the go client doesn’t, and ideally we would have the tonic client behave similar to the go client…

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1127)

### DIFF
--- a/crates/flowctl/src/connector.rs
+++ b/crates/flowctl/src/connector.rs
@@ -49,20 +49,19 @@ pub fn docker_spawn(image: &str, args: &[&str]) -> anyhow::Result<(Child, TempDi
             [
                 &[
                     "run",
-                    "--interactive",
-                    "--init",
                     "--rm",
-                    "--log-driver=none",
-                    "--mount",
-                    &format!("type=bind,source={host_inspect_str},target={target_inspect}"),
                     "--mount",
                     &format!(
                         "type=bind,source={host_connector_init_str},target={target_connector_init}"
                     ),
-                    "--publish",
-                    &format!("0.0.0.0:{}:{}/tcp", port, CONNECTOR_INIT_PORT),
                     "--entrypoint",
                     target_connector_init,
+                    "--env",
+                    "LOG_LEVEL=trace",
+                    "--publish",
+                    &format!("0.0.0.0:{}:{}/tcp", port, CONNECTOR_INIT_PORT),
+                    "--mount",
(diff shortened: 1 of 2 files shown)


crates/flowctl/src/connector.rs
```diff
@@ -40,22 +40,16 @@ pub fn docker_spawn(image: &str, args: &[&str]) -> anyhow::Result<(Child, TempDi
...
+              "--entrypoint", target_connector_init,
+              "--env", "LOG_LEVEL=trace",
 		      "--publish", &format!("0.0.0.0:{}:{}/tcp", port, CONNECTOR_INIT_PORT),
-              "--entrypoint",
-              target_connector_init,
+              "--mount", &format!("type=bind,source={host_inspect_str},target={target_inspect}"),
               image,
               &format!("--image-inspect-json-path={target_inspect}"),
               &format!("--port={CONNECTOR_INIT_PORT}"),
         ], args].concat())
```
mdibaiee, line 52: Updated the arguments here to match the ones in `container.go`

jgraettinger: rust fmt? perhaps sort --flags ?

Also remove LOG_LEVEL=trace ? that's going to be _very_ verbose.

jgraettinger (approved):
LGTM % comments

Issue #1115, opened 2023-07-18
Implement continuous schema inference in the capture and derive runtimes


jshearer:
- [x] Update schema inference to handle dynamic fields correctly
- [x] Captures and derivations would startup using an initial empty doc::Shape as inferredShape associated with each binding
- [x] Every document that's drained from the combiner would be used to expand the inferredShape
- [x] The inferredShape is logged out whenever it changes, but at most once per transaction

PR #1128, opened 2023-07-28
feature: Teach the combiner to keep track of a running inferred schema and log whenever it changes.

jshearer:
This is the second step of https://github.com/estuary/flow/issues/1115.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1128)


jshearer:
I'm also done force-pushing here too, sorry 🤦

crates/derive/src/combine_api.rs
```diff
@@ -177,9 +194,27 @@ impl cgo::Service for API {
...
                 if !more {
+                    if state.shape.ne(&state.scratch_shape) {
+                        // Update the "true" shape with the newly widened shape
+                        state.shape = state.scratch_shape.clone();
+
+                        match serde_json::to_value(
+                            &SchemaBuilder::new(state.scratch_shape.clone()).root_schema(),
+                        ) {
+                            Ok(value) => tracing::info!(
+                                inferred_schema = ?DebugJson(value),
```
jgraettinger, line 209: this needs to include the collection, for the derivation to then be able to re-map it. That's going to require adding it to the protobuf configuration message.

nit: `schema` as the field name, instead of `inferred_schema`?

jgraettinger (commented):
Two major notes:

* We want to selectively enable this widening / logging. Capture tasks definitely should do it, but materializations don't need to and also don't want to pay the overhead. How about a new boolean to enable the behavior in the protobuf config API?

* The logs must include the collection name, which also needs to be plumbed through.

Other than that and comments below LGTM 👍

crates/ops/src/tracing.rs
```diff
@@ -1,3 +1,5 @@
+use std::string;
```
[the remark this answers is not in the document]
jshearer: ... what is going on with these. Is there some rust-analyzer setting I don't have set or something? Every single one of these has been from an auto-import

jshearer:
Ok, I added `collection_name` and `enable_schema_inference` to the `Config` message. Thoughts?

crates/derive/src/combine_api.rs
jshearer: Plumbed through, and renamed to `schema` 👍

crates/ops/src/tracing.rs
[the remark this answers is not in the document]
jshearer: Fixed, and [threw up](https://github.com/estuary/flow/pull/1136) to hopefully fix this forever

crates/derive/src/combine_api.rs
```diff
@@ -177,9 +194,25 @@ impl cgo::Service for API {
...
+                        // Update the "true" shape with the newly widened shape
+                        state.shape = state.scratch_shape.clone();
+
+                        match state.scratch_shape.clone().to_serde() {
+                            Ok(value) => tracing::info!(
+                                inferred_schema = ?DebugJson(value),
+                                "inferred schema updated"
+                            ),
+                            Err(err) => {
+                                tracing::error!("failed to serialize inferred schema: {err:?}")
```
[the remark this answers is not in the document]
jshearer: Yep, that makes sense. Changed to `expect()`.

crates/derive/src/combine_api.rs
```diff
@@ -54,6 +56,13 @@ impl StatsAccumulator for CombineStats {
 struct State {
     // Combiner which is doing the heavy lifting.
     combiner: doc::Combiner,
+    // Inferred shape of drained documents
+    shape: doc::inference::Shape,
+    // We need to keep track of the potentially-modified shape
```
jgraettinger, line 61: `doc::inference::Shape` => `doc::shape::Shape` or just `doc::Shape`.

i think we can drop `scratch_shape`, because you can instead check whether any `widen()` operation returned true to track whether it's changed?

crates/derive/src/combine_api.rs
```diff
@@ -177,9 +202,28 @@ impl cgo::Service for API {
                     arena,
                     out,
                     &mut state.stats.out,
+                    &mut state
+                        .enable_schema_inference
```
jgraettinger, line 206: could combine `enable_schema_inference` by making the inferred `Shape` an Option and then passing `self.inferred_shape.as_mut()`

crates/proto-flow/src/flow.rs
```diff
@@ -896,6 +896,12 @@ pub mod combine_api {
         /// TODO(johnny): This is a kludge as we seek to remove this API.
         #[prost(message, repeated, tag = "5")]
         pub projections: ::prost::alloc::vec::Vec<super::Projection>,
+        /// The name of the collection that's being written to.
+        #[prost(string, tag = "6")]
+        pub collection_name: ::prost::alloc::string::String,
+        /// Whether to emit maximally-constrained schemas as documents are combined.
```
jgraettinger, line 902: 🤔 what about `infer_schema_json: string`, which is a json-encoded string of JSON to start inferring from, or empty if inference shouldn't happen?

this is a generalization of the boolean which allows the caller to send `false` as the starting schema, but _also_ allows for some other schema as the starting point.

jgraettinger: background: i think we'll be plumbing the inferred schema through into the collection spec, which means we'd be able to start from the schema that was inferred as-of the last publication of the collection, which would make the inferred schema changes more meaningful in the task logs.

crates/derive/src/combine_api.rs
```diff
@@ -225,11 +269,18 @@ pub fn drain_chunk(
...
     stats: &mut DocCounter,
+    shape: &mut Option<&mut doc::inference::Shape>,
 ) -> Result<bool, doc::combine::Error> {
     // Convert target from a delta to an absolute target length of the arena.
     let target_length = target_length + arena.len();
 
     drainer.drain_while(|doc, fully_reduced| {
+        if let Some(shape) = shape.as_mut() {
+            match &doc {
+                doc::LazyNode::Node(n) => shape.widen(*n, Location::Root),
```
jgraettinger, line 280: you need to call the limit function if this returns true -- a single combiner can have a _lot_ of documents.

crates/derive/src/combine_api.rs
jshearer: Yup, changed `shape` to `Option<Shape>`, and added a `shape_changed` flag on State.

crates/derive/src/combine_api.rs
jshearer: 🤦 yup

crates/proto-flow/src/flow.rs
jshearer: Done 👍

crates/derive/src/combine_api.rs
jshearer: Done

crates/derive/src/combine_api.rs
```diff
@@ -177,9 +213,24 @@ impl cgo::Service for API {
                     arena,
                     out,
                     &mut state.stats.out,
+                    &mut state.shape,
+                    &mut state.shape_changed,
                 )?;
 
-                if !more {
+                if let Some(ref shape) = state.shape {
```
jgraettinger, line 220: this logic breaks combiner draining, I believe. Did tests not catch it, or just haven't been run yet?

The outer test needs to be `if !more`, which maps into an Accumulator, and otherwise is a Drainer.

schema inference is secondary.

crates/doc/src/lib.rs
```diff
@@ -114,6 +114,7 @@ pub mod tuple_pack;
 // It's similar to (and built from) a JSON Schema, but includes only
 // those inferences which can be statically proven for all documents.
 pub mod shape;
+pub use shape::schema::to_schema;
```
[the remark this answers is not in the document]
jshearer: blah sorry. My `rust-analyzer` bugged out because of the upgrade in Rust versions (apparently, a `rustup upgrade` fixed it..) and it kept on trying to make me do this refactor. Removed 👍

crates/derive/src/combine_api.rs
jshearer: Whoops! Tests did catch it, but they hadn't been run until recently

crates/derive/src/combine_api.rs
```diff
@@ -313,6 +378,8 @@ pub mod test {
                         ..Default::default()
                     },
                 ],
+                collection_name: "test".to_string(),
+                infer_schema_json: "".to_string(),
```
jgraettinger, line 382: nit: send in a schema, to exercise inference logic with this test case.

jgraettinger (approved):
LGTM

PR #1129, opened 2023-08-01
proto-flow: add support for materialization binding priority

jgraettinger:
Materialization priority has the same semantics as derivations. All documents of a higher-priority binding are processed before any documents of lower-priority bindings.

This allows a busy materialization to focus effort on specific high-priority bindings, while back-filling other bindings in the background with available spare capacity.

**Workflow steps:**

Updated specification of materialization concepts page.

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1129)


psFried (approved):
LGTM

Issue #1103, opened 2023-07-11
Introduce continuous schema inference


psFried:
Discussed in #1042 

- [x] #1131
- [x] #1140
    - use the collection name as the key, and the `flow-inferred-schema-merge` reduction
    - Materialize the ops/inferred-schemas collection to the `inferred_schemas` table in the control plane database
      - The table should have two additional columns, which contain the hash of the current inferred schema, and the hash of the most recently published inferred schema
- [x] #1115
    - Captures and derivations would startup using an initial empty `doc::Shape` as `inferredShape` associated with each binding
    - Every document that's drained from the combiner would be used to expand the `inferredShape`
    - The `inferredShape` is logged out whenever it changes, but at most once per transaction
- [x] #1142
    - The inferred schema should be composed with the `writeSchema` by putting them both into an `allOf` to ensure that we retain any reduce annotations that were given on the writeSchema.

### Inference with dynamic fields

We've seen a number of examples of documents that include large numbers of distinct dynamic fields. For example, objects with a seemingly unbounded number of fields like `custom_field_12345`. These could potentially make the overall size of the inferred schema pretty unbearable, and it's probably wise to implement some form of system protection limit on the number of fields in an inferred object shape. To that end, I'll propose the following rules be added to our inference:

- Limit the top-level object shape to at most 750 named fields. After this number is reached, put any other properties into `additionalProperties`, while retaining all of the existing named fields.
- Limit any nested Object shapes to at most 200 named fields. After that threshold, collapse all existing properties into `additionalProperties`, leaving the schema without naming individual fields.
- If an object shape defines `additionalProperties`, then never add explicit named fields, and instead update the `additionalProperties` shape
- Apply these rules when merging shapes

### Determining the `readSchema`

The `discovers` handler will determine the `readSchema` by composing the `writeSchema` and the inferred schema using an `allOf`. There's a few reasons for this:

- It ensures that any `reduce` annotations from the `writeSchema` are propagated to the `readSchema`, so that reduction behavior is the same for both reading and writing.
- It ensures that the `readSchema` will always cover any key/partition components that happen to be a location that schema inference determines to be a dynamic object as described above. Otherwise, we might not generate projections for key/partition fields after a group of named fields is replaced by an `additionalProperties` schema.

psFried:
This is now "Fully Implemented"™️

PR #1132, opened 2023-08-03
Introduce the `jsonSchemaMerge` reduction strategy

### LINKED ISSUE #1131: Add support for `reduce: flow-inferred-schema-merge` in reductions

jshearer:
**Description:**

Introduces the `jsonSchemaMerge` reduction strategy. I still need to add some more tests, but we should really be relying on the tests for `Shape::union()` (which are extensive) and `doc::schema::to_schema()` (which very much are not) to ensure correctness in all cases here.

This closes https://github.com/estuary/flow/issues/1131, and is part of the work laid out in https://github.com/estuary/flow/issues/1103.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1132)

### DIFF
--- a/crates/doc/src/inference.rs
+++ b/crates/doc/src/inference.rs
@@ -98,6 +98,7 @@ pub enum Reduction {
 
     Append,
     FirstWriteWins,
+    JsonSchemaMerge,
     LastWriteWins,
     Maximize,
     Merge,
@@ -107,8 +108,6 @@ pub enum Reduction {
 
     // Multiple concrete strategies may apply at the location.
     Multiple,
-
-    JsonSchemaMerge,
 }
 
 impl Reduction {
--- a/crates/doc/src/inference.rs
+++ b/crates/doc/src/inference.rs
@@ -107,6 +107,8 @@ pub enum Reduction {
 
     // Multiple concrete strategies may apply at the location.
     Multiple,
+
+    JsonSchemaMerge,
 }
 
 impl Reduction {
@@ -177,6 +179,7 @@ impl From<&reduce::Strategy> for Reduction {
             Strategy::Set(_) => Reduction::Set,
             Strategy::Sum => Reduction::Sum,
             Strategy::Merge(_) => Reduction::Merge,
+            Strategy::JsonSchemaMerge => Reduction::JsonSchemaMerge,
         }
     }
 }
--- a/crates/doc/src/reduce/strategy.rs
+++ b/crates/doc/src/reduce/strategy.rs
@@ -1,9 +1,13 @@
+use json::schema::index::IndexBuilder;
+
 use super::{
     compare_key_lazy, count_nodes, count_nodes_generic, count_nodes_heap, reduce_item, reduce_prop,
     Cursor, Error, Result,
 };
 use crate::{
+    inference::Shape,
     lazy::{LazyArray, LazyDestructured, LazyObject},
+    schema::SchemaBuilder,
     AsNode, BumpVec, HeapNode, Node, Pointer,
 };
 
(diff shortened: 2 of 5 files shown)


jshearer:
I'm done force-pushing now, sorry 🤦

crates/doc/src/inference.rs
```diff
@@ -107,6 +107,8 @@ pub enum Reduction {
 
     // Multiple concrete strategies may apply at the location.
     Multiple,
+
+    JsonSchemaMerge,
```
jgraettinger, line 111: nit: order / sort with the block of others. This isn't special, in the way that `Unset` and `Multiple` are.

crates/doc/src/reduce/strategy.rs
```diff
@@ -1,9 +1,13 @@
+use json::schema::index::IndexBuilder;
```
jgraettinger, line 1: nit: newline.

crates/doc/src/reduce/strategy.rs
```diff
@@ -301,6 +309,54 @@ impl Strategy {
...
+        let Cursor {
+            tape,
+            loc,
+            full: _,
+            lhs,
+            rhs,
+            alloc,
+        } = cur;
+
+        let (lhs, rhs) = (lhs.into_heap_node(alloc), rhs.into_heap_node(alloc));
```
jgraettinger, line 324: nit: i believe you could alternately work directly with `Node` and avoid mapping into HeapNode.

This is so minor, though, to not even be worth the effort if this already works.

crates/doc/src/reduce/strategy.rs
```diff
@@ -301,6 +309,54 @@ impl Strategy {
...
+
+        // Precompute this up here because we're going to consume `rhs` later on
+        let rhs_tape_len = count_nodes_heap(&rhs);
```
jgraettinger, line 327: nit: i think you can just update *tape immediately. You don't have any nested reduction annotations to worry about.

crates/doc/src/reduce/strategy.rs
```diff
@@ -301,6 +309,54 @@ impl Strategy {
         }
     }
 
+    fn json_schema_merge<'alloc, L: AsNode, R: AsNode>(
```
jgraettinger, line 312: nit: please place in a different file, similar to how the set reduction lives in `set.rs`.

crates/doc/src/reduce/strategy.rs
```diff
@@ -301,6 +309,54 @@ impl Strategy {
...
+        ) = (lhs, rhs) else {
+            return Err(Error::with_location(Error::JsonSchemaMergeWrongType { detail: None }, loc) )
+        };
+
+        let left = shape_from_node(lhs).map_err(|e| Error::with_location(e, loc))?;
+        let right = shape_from_node(rhs).map_err(|e| Error::with_location(e, loc))?;
+
+        // Union together the LHS and RHS, and convert back from `Shape` into `HeapNode`.
+        let merged_doc =
+            serde_json::to_value(&SchemaBuilder::new(Shape::union(left, right)).root_schema())
```
jgraettinger, line 344: do you also need to call the routine to trim excessive numbers of properties?

jgraettinger (approved):
LGTM % comments below

crates/doc/src/reduce/strategy.rs
jshearer: I tried to do this at first and it ended up being a bigger change than I wanted so I think I'm gonna leave it be -- `lhs` and `rhs` are `LazyNode`s which don't directly implement `AsNode` and I wasn't able to easily implement it because of the generic in `LazyNode::Node`. I feel like it should be possible since `lazy_node.as_node()` would to have a clearly defined type in all scenarios: whatever `N` is when you have `LazyNode::Node`, and `HeapNode` if you have `LazyNode::Heap`.

crates/doc/src/reduce/strategy.rs
jshearer: Wow, yes.

PR #1133, opened 2023-08-04
source-postgres-batch: Add very minimal documentation

willdonnelly:
**Description:**

This should be improved some time, but since we auto-open the page in a sidebar in our dashboard there needs to be *something* there immediately, and this is the bare minimum IMO.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1133)


oliviamiannone (approved):
From that "bare minimum" perspective, this looks good to me. The one thing I think should definitely be added sooner than later would be some version of prerequisites, which I imagine is the same as the CDC connector, less the CDC-specific items. 

in the future it'd be great to flesh out the last few paragraphs with some examples etc.

PR #1135, opened 2023-08-04
Refactor/schema builder

jshearer:
Move SchemaBuilder from schema-inference to doc since doc needs to learn how to turn a Shape into the correct blob of JSON representing that Shape as a JSON schema. This logic used to live in schema-inference, but not only do we want to get rid of that crate eventually, it also already depends on doc indirectly, so introducing the doc->schema-inference dependency caused a circular dependency issue.

In addition to moving SchemaBuilder into doc, I also realized that it was missing quite a few translations between Shape and Schema. I added the ones I could find, but it's entirely possible that there are more missing. I will add some more tests, but it's likely that there are still options we aren't supporting correctly.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1135)

### DIFF
--- a/crates/doc/src/inference.rs
+++ b/crates/doc/src/inference.rs
@@ -1,5 +1,5 @@
 use super::{compare, ptr::Token, reduce, Annotation, Pointer, Schema, SchemaIndex};
-use crate::{schema::SchemaBuilder, AsNode};
+use crate::AsNode;
 use fancy_regex::Regex;
 use itertools::{self, EitherOrBoth, Itertools};
 use json::{
--- a/crates/doc/src/inference.rs
+++ b/crates/doc/src/inference.rs
@@ -1489,13 +1489,6 @@ impl Shape {
             );
         };
     }
-
-    pub fn to_serde(self) -> Result<Value, serde_json::Error> {
-        let builder = SchemaBuilder::new(self);
-        let root_schema = builder.root_schema();
-
-        serde_json::to_value(root_schema)
-    }
 }
 
 // Sentinel Shape returned by locate(), which make take any value.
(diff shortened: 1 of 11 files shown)


crates/doc/src/inference.rs
```diff
@@ -1489,6 +1489,13 @@ impl Shape {
             );
         };
     }
+
+    pub fn to_serde(self) -> Result<Value, serde_json::Error> {
```
jgraettinger, line 1493: I'd prefer we don't add this to the public interface.
Callers can do `serde_json::to_value(&SchemaBuilder::new(shape).root_schema())` for the time being.

I'd like to do further refactors of the `SchemaBuilder` stuff.

crates/doc/src/schema.rs
```diff
@@ -1,11 +1,16 @@
 use std::collections::{BTreeMap, BTreeSet};
```
jgraettinger, line 2: nit: trim extra remaining spaces

jgraettinger (approved):
LGTM -- let's land this, and I may do some fast-follows with `doc::schema::SchemaBuilder`

crates/doc/src/inference.rs
jshearer: kk. I was trying to follow [`HeapNode::from_serde()`](https://github.com/estuary/flow/blob/master/crates/doc/src/heap_de.rs#L18-L26), but also happy to not add to the public interface 👍

crates/doc/src/schema.rs
jshearer: done!

crates/doc/src/inference.rs
jshearer: done!

PR #1137, opened 2023-08-05
crates/doc: refactor `inference` module into `shape` module and friends

jgraettinger:
See individual commits. No (intentional) behavior changes in this PR, it's just a refactor.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1137)


crates/doc/src/shape/limits.rs
```diff
@@ -0,0 +1,285 @@
+// This module defines limits which are used to simplify complex,
+// typically inferred schema Shapes.
+use super::*;
+use json::Location;
+
+// Prune any locations in this shape that have more than the allowed fields,
+// squashing those fields into the `additionalProperties` subschema for that location.
+pub fn enforce_field_count_limits(slf: &mut Shape, loc: Location) {
```
jshearer, line 8: Not an issue, but I am curious why this isn't in an `impl Shape` block, or rather what heuristic you're using to determine when to use a free function vs an impl. I had been using `impl`s mainly for discoverability: we have a pretty big codebase, and being able to write `my_shape.` and have my IDE give me a list of functions I can call on Shape feels useful. Is it that we're not confident in this API and don't want to "expose" it?

jshearer (approved):
LGTM -- I didn't check and make sure that each moved chunk works exactly the same as it did before, but it looks right and is way better broken up like this. `inference.rs` being >2k lines long was too much

crates/doc/src/shape/limits.rs
jgraettinger: >  Is it that we're not confident in this API and don't want to "expose" it?

That's some of it, yes.

It's also orthogonal to what `Shape` is and does, and isn't core to it's function. Especially once you re-interpret `widen` to return "did this widen change the Shape anywhere?". `widen` is a core capability of `Shape`, while `enforce_field_count_limits` is business logic applied _to_ a Shape. And a big distinction in my mind, is that we can define _lots_ of different heuristics and trade-offs for when to prune stuff out of a `Shape`, but there's only one sensible `widen` operation.

Also, as a general strategy, I definitely recommend starting out with free functions and only attaching them to structs when it's really obvious they belong. It makes refactoring way easier, especially in the early stages of a subsystem when you're unsure how everything fits together.

As a discipline, I find a lot of engineers will jump too quickly into an object-plus-behaviors-oriented mindset and then get stuck when their initial conceptual modeling doesn't match evolving requirements. Free functions that you can refactor and move as-needed avoids that mental trap, because it defers bestowing "object-ness" onto what are, at the end of the day, just a pile of bytes.

PR #1141, opened 2023-08-08
Add schema inference `L1` and `L2` collections, as well as materializing them to the database

### LINKED ISSUE #1140: Introduce ops/inferred-schemas collection

jshearer:
Per convo with @jgraettinger, I added `L1` under `ops.us-central1.v1/private` and `L2` under `estuary/global`

Closes #1140. It works!

Screen Shot 2023-08-08 at 14 21 43

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1141)

### DIFF
--- a/ops-catalog/schema-inference.flow.yaml
+++ b/ops-catalog/schema-inference.flow.yaml
@@ -1,0 +1,35 @@
+collections:
+  ops.us-central1.v1/inferred-schemas/L1:
+    derive:
+      using:
+        sqlite: {}
+      transforms:
+        - name: logs
+          source:
+            name: ops.us-central1.v1/logs
+          shuffle:
+            key: [/shard/name] # Use partition-based shuffle.
+          lambda: |
+            select
+              $fields->>'collection_name' as collection_name,
+              $fields->'schema' as schema
+            where $message = 'inferred schema updated';
+    key: [/collection_name]
+    schema: schema-inference.schema.yaml
+
+  # A direct-copy of L1, but enables us to add more L1s
+  # as we create more data planes.
+  ops.us-central1.v1/inferred-schemas/L2:
+    schema: schema-inference.schema.yaml
+    key: [/collection_name]
+    derive:
+      using:
+        sqlite: {}
+      # The idea with this L2 collection is to merge
+      # all data planes' L1 collections together
+      transforms:
+        - name: from-ops.us-central1.v1
+          source: ops.us-central1.v1/inferred-schemas/L1
+          shuffle:
+            key: [/collection_name]
+          lambda: select json($flow_document);
--- a/ops-catalog/template-local.flow.yaml
+++ b/ops-catalog/template-local.flow.yaml
@@ -9,6 +9,15 @@ materializations:
         config: stats-local-endpoint.sops.yaml
 
     bindings:
+      - source: ops.us-central1.v1/inferred-schemas/L2
+        resource:
+          table: inferred_schemas
+        fields:
+          recommended: false
+          include:
+            collection_name: {}
+            schema: {}
+            flow_document: {}
       - source: ops.us-central1.v1/catalog-stats-L2
         resource:
           table: catalog_stats
--- a/ops-catalog/template-prod.flow.yaml
+++ b/ops-catalog/template-prod.flow.yaml
@@ -9,6 +9,15 @@ materializations:
         config: stats-production-endpoint.sops.yaml
 
     bindings:
+      - source: ops.us-central1.v1/inferred-schemas/L2
+        resource:
+          table: inferred_schemas
+        fields:
+          recommended: false
+          include:
+            collection_name: {}
+            schema: {}
+            flow_document: {}
       - source: ops.us-central1.v1/catalog-stats-L2
         resource:
           table: catalog_stats
--- a/ops-catalog/schema-inference.schema.yaml
+++ b/ops-catalog/schema-inference.schema.yaml
@@ -1,0 +1,15 @@
+type: object
+reduce:
+  strategy: merge
+required:
+  - collection_name
+  - schema
+properties:
+  collection_name:
+    type: string
+    description: The name of the collection that this schema was inferred for
+  schema:
+    type: object
+    description: The inferred schema
+    reduce:
+      strategy: jsonSchemaMerge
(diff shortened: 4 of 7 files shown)


ops-catalog/schema-inference.flow.yaml
```diff
@@ -0,0 +1,36 @@
...
+      using:
+        sqlite: {}
+      # The idea with this L2 collection is to merge
+      # all data planes' L1 collections together
+      transforms:
+        - name: from-ops.us-central1.v1
+          source: ops.us-central1.v1/private/schema-inference-L1
+          shuffle: any
+          lambda: select $collection_name, $schema;
+  ops.us-central1.v1/private/schema-inference-L1:
```
psFried, line 19: nit: IMO a better name would be `ops.us-central1.v1/inferred-schemas/L1`. `/private/` feels like a bit of a smell, since visibility is already handled by grants.

ops-catalog/template-local.flow.yaml
```diff
@@ -2,6 +2,22 @@ import:
   - template-common.flow.yaml
 
 materializations:
+  estuary/global/inferred-schemas:
```
psFried, line 5: Is there a reason this should be a separate materialization vs an additional binding of the exiting one? What are the considerations there?

ops-catalog/template-local.flow.yaml
```diff
@@ -2,6 +2,22 @@ import:
...
+    endpoint:
+      connector:
+        image: ghcr.io/estuary/materialize-postgres:dev
+        config: stats-local-endpoint.sops.yaml
+
+    bindings:
+      - source: estuary/global/schema-inference-L2
+        resource:
+          table: catalog_schemas
```
psFried, line 14: I'll suggest `inferred_schemas` for the table name, since they may or may not actually be used in the catalog. That would also avoid any potential confusion with [the Catalog schema](https://github.com/estuary/flow/blob/master/flow.schema.json), which is the schema of a Flow catalog spec.

ops-catalog/template-local.flow.yaml
```diff
@@ -2,6 +2,22 @@ import:
...
+        fields:
+          recommended: false
+          include:
+            ts: {}
```
psFried, line 18: The `ts` field doesn't appear in the schema, so I'm assuming this is vestigial?

ops-catalog/template-prod.flow.yaml
```diff
@@ -2,6 +2,22 @@ import:
...
+        config: stats-production-endpoint.sops.yaml
+
+    bindings:
+      - source: estuary/global/schema-inference-L2
+        resource:
+          table: catalog_schemas
+        fields:
+          recommended: false
+          include:
+            ts: {}
```
psFried, line 18: Same, `ts`

psFried (changes requested):
I left a few comments to discuss before merging. High-level looks good, though.

One other general question: How is the `catalog_schemas` table created? I was expected to see a new migration adding that to the control plane database. Is that covered in a different PR?

ops-catalog/schema-inference.flow.yaml
jgraettinger: @psFried FYI, I'm deeply regretting that we placed `logs` and `stats` at the tenant top-level...

ops-catalog/schema-inference.flow.yaml
```diff
@@ -0,0 +1,36 @@
+collections:
+  estuary/global/schema-inference-L2:
+    schema: schema-inference.schema.yaml
+    key: [/collection_name]
+    projections:
```
jgraettinger, line 5: nit: these projections are already implicit and can be removed.

ops-catalog/schema-inference.flow.yaml
```diff
@@ -0,0 +1,36 @@
...
+    derive:
+      using:
+        sqlite: {}
+      # The idea with this L2 collection is to merge
+      # all data planes' L1 collections together
+      transforms:
+        - name: from-ops.us-central1.v1
+          source: ops.us-central1.v1/private/schema-inference-L1
+          shuffle: any
+          lambda: select $collection_name, $schema;
```
jgraettinger, line 18: nit: add newline?

ops-catalog/schema-inference.flow.yaml
```diff
@@ -0,0 +1,36 @@
...
+      transforms:
+        - name: inferredSchemas
+          source: ops.us-central1.v1/logs
+          shuffle: any
+          lambda: |
+            select
+              $fields->'collection_name' as collection_name,
+              $fields->'schema' as schema
+            where $message = 'inferred schema updated';
+    projections:
```
jgraettinger, line 32: nit: same here -- you don't need these projections.

ops-catalog/template-local.flow.yaml
jshearer: Had a talk with Johnny about this, the thought is that the L2 materialization should live in the `estuary/` tenant, since its purpose is to roll up all data-plane-specific L1 inferred schemas collections.

ops-catalog/schema-inference.flow.yaml
psFried: @jgraettinger I'm curious why? Is it making it more difficult to manage access grants?

jgraettinger: If I want to grant `aliceCo` access to `bobCo/logs`, I have to give them a read grant to all of `bobCo/` since grants operate on prefixes.

ops-catalog/template-local.flow.yaml
jshearer: Yep. Thought I heard Johnny wanted it to be called `catalog_schemas` but this is better 👍

ops-catalog/template-local.flow.yaml
jshearer: Moved it to a binding on the existing materialization. I had it here because we were going to introduce `estuary/` tenant to hold global L2s, but also undid that per convo w/ Johnny

ops-catalog/template-local.flow.yaml
```diff
@@ -9,6 +9,14 @@ materializations:
         config: stats-local-endpoint.sops.yaml
 
     bindings:
+      - source: ops.us-central1.v1/inferred-schemas/L2
+        resource:
+          table: inferred_schema
+        fields:
+          recommended: false
+          include:
+            collection_name: {}
```
jgraettinger, line 18: it won't work / will refuse to work without `flow_document` as well.

jgraettinger (approved):
LGTM % including `flow_document` as well.

psFried (approved):
LGTM 🚀 🚀 🚀

PR #1143, opened 2023-08-10
doc: performance and inference enhancements

jgraettinger:
Please review commit by commit.

There's a fair amount here, but I did my best to tell a story through the commit messages.

### DIFF
--- a/crates/doc/src/reduce/mod.rs
+++ b/crates/doc/src/reduce/mod.rs
@@ -135,40 +135,36 @@ impl<'alloc, L: AsNode, R: AsNode> Cursor<'alloc, '_, '_, '_, '_, L, R> {
     }
 }
 
-fn count_nodes<N: AsNode>(v: &LazyNode<'_, '_, N>) -> usize {
+fn count_nodes_lazy<N: AsNode>(v: &LazyNode<'_, '_, N>) -> usize {
     match v {
-        LazyNode::Node(doc) => count_nodes_generic(&doc.as_node()),
-        LazyNode::Heap(doc) => count_nodes_heap(doc),
+        LazyNode::Node(doc) => count_nodes(*doc),
+        LazyNode::Heap(doc) => count_nodes(doc),
     }
 }
 
-fn count_nodes_generic<N: AsNode>(node: &Node<'_, N>) -> usize {
-    match node {
-        Node::Bool(_) | Node::Null | Node::String(_) | Node::Number(_) | Node::Bytes(_) => 1,
-        Node::Array(v) => v
-            .iter()
-            .fold(1, |c, vv| c + count_nodes_generic(&vv.as_node())),
-        Node::Object(v) => v.iter().fold(1, |c, field| {
-            c + count_nodes_generic(&field.value().as_node())
-        }),
+fn count_nodes<N: AsNode>(node: &N) -> usize {
+    match node.as_node() {
+        Node::Bool(_)
+        | Node::Bytes(_)
+        | Node::Float(_)
+        | Node::NegInt(_)
+        | Node::Null
+        | Node::PosInt(_)
+        | Node::String(_) => 1,
+
+        Node::Array(v) => count_nodes_items(v),
+        Node::Object(v) => count_nodes_fields::<N>(v),
     }
 }
 
-// A HeapNode can also be counted as an AsNode, but this is faster (it avoids Node<> conversion).
-fn count_nodes_heap(node: &HeapNode<'_>) -> usize {
-    match node {
-        HeapNode::Bool(_)
-        | HeapNode::Bytes(_)
-        | HeapNode::Float(_)
-        | HeapNode::NegInt(_)
-        | HeapNode::Null
-        | HeapNode::PosInt(_)
-        | HeapNode::String(_) => 1,
-        HeapNode::Array(v) => v.iter().fold(1, |c, vv| c + count_nodes_heap(vv)),
-        HeapNode::Object(v) => v
-            .iter()
-            .fold(1, |c, field| c + count_nodes_heap(&field.value)),
-    }
+fn count_nodes_items<N: AsNode>(items: &[N]) -> usize {
+    items.iter().fold(1, |c, vv| c + count_nodes(vv))
+}
+
+fn count_nodes_fields<N: AsNode>(fields: &N::Fields) -> usize {
+    fields
+        .iter()
+        .fold(1, |c, field| c + count_nodes(field.value()))
 }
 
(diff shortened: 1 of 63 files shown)


crates/doc/src/shape/widen.rs
```diff
@@ -2,232 +2,393 @@
...
+        let mut lhs_it = self.properties.iter_mut();
+        let mut rhs_it = fields.iter();
+        let mut maybe_lhs = lhs_it.next();
+        let mut maybe_rhs = rhs_it.next();
+
+        // Perform an ordered merge over `lhs_it` and `rhs_it`.
+        // This loop structure is much faster than Itertools::merge_join_by.
+        loop {
+            match (&mut maybe_lhs, &maybe_rhs) {
+                (Some(lhs), Some(rhs)) => match lhs.name.as_ref().cmp(rhs.property()) {
```
jshearer, line 117: This is relying deeply on the assumption that lhs and rhs are both sorted, and anything that results in them not being sorted would cause subtly wrong behavior. I wonder if it would be worth adding a check here to ensure that `lhs < lhs_it.next()` and fail fast if not. 

I'm guessing you don't want to because of performance though, so... 🤞 we don't accidentally create incorrectly-sorted properties/fields.

crates/doc/src/shape/widen.rs
```diff
@@ -2,232 +2,393 @@
...
+}
 
-                self.string = StringShape::union(self.string.clone(), partial_stringshape);
-                false
-            }
-            crate::Node::Null => {
-                self.type_ = self.type_ | types::NULL;
-                false
+impl Shape {
+    pub fn widen<'n, N>(&mut self, node: &'n N) -> bool
```
jshearer, line 233: No more docstring?

crates/doc/src/shape/widen.rs
```diff
@@ -2,232 +2,393 @@
...
-
-                // Similar to the nuance around `is_required`, string format
-                // should only "become detected" the very first time. We still
-                // need to run `detect_format` on subsequent strings because
-                // `StringShape::union()` can sometimes widen a string format,
-                // e.g from `integer` -> `number`
-                let format = if is_first_time || self.string.format.is_some() {
-                    Format::detect(s)
+            Node::Bool(_) => apply_type(types::BOOLEAN),
+            Node::Bytes(_) => panic!("not implemented"),
```
jshearer, line 260: This was implemented previously, as a string with `content_encoding: "base64"`. (why) do we not care about this case?

crates/doc/src/shape/widen.rs
```diff
@@ -2,232 +2,393 @@
...
+                self.object.widen::<N>(fields, is_first)
+            }
+            Node::String(s) => self.string.widen(s, apply_type(types::STRING)),
+        }
+    }
 
-                false
+    #[cold]
+    #[inline(never)]
+    fn widen_enum<'n, N>(&mut self, node: &'n N) -> bool
```
jshearer, line 289: We should limit the number of enum variants. I could see this one in particular resulting in huge schemas if a field is accidentally defined as an enum, or was previously defined as an enum but then becomes a regular old string field.

crates/doc/src/shape/schema.rs
```diff
@@ -1,112 +1,204 @@
...
+        schema: to_sub_schema(shape).into_object(),
+        meta_schema: schemars::gen::SchemaSettings::draft2019_09().meta_schema,
+        ..Default::default()
+    }
 }
 
-impl SchemaBuilder {
-    pub fn new(shape: Shape) -> Self {
-        Self { shape }
+fn to_sub_schema(xyz: Shape) -> Schema {
```
jshearer, line 16: nit: variable name `shape`?

crates/validation/src/errors.rs
```diff
@@ -86,7 +86,7 @@ pub enum Error {
     },
     #[error("location {ptr} is unknown in schema {schema}")]
     PtrIsImplicit { ptr: String, schema: Url },
-    #[error("location {ptr} has a reduction strategy {strategy:?}, which is disallowed because the location is used as a key")]
+    #[error("location {ptr} has a reduction strategy, which is disallowed because the location is used as a key")]
```
jshearer, line 89: Why remove useful information from errors?

jshearer (approved):
Had some questions, but in general LGTM. Especially happy to see the compiler-verified Shape -> Schema round-tripping and `widen` perf improvements 👍

crates/doc/src/reduce/mod.rs
```diff
@@ -135,40 +135,36 @@ impl<'alloc, L: AsNode, R: AsNode> Cursor<'alloc, '_, '_, '_, '_, L, R> {
...
-        | HeapNode::Null
-        | HeapNode::PosInt(_)
-        | HeapNode::String(_) => 1,
-        HeapNode::Array(v) => v.iter().fold(1, |c, vv| c + count_nodes_heap(vv)),
-        HeapNode::Object(v) => v
-            .iter()
-            .fold(1, |c, field| c + count_nodes_heap(&field.value)),
-    }
+fn count_nodes_items<N: AsNode>(items: &[N]) -> usize {
+    items.iter().fold(1, |c, vv| c + count_nodes(vv))
```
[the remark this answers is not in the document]
jgraettinger: yep, the array / object itself is also a node.

crates/doc/src/shape/widen.rs
jgraettinger: whoops, re-adding

crates/doc/src/shape/widen.rs
jgraettinger: Node::Bytes isn't used today -- it exists somewhat speculatively, as a way to speed up not-yet-implemented reduction annotations that _may_ work over bytes.

Potentially it ought to be removed, except for a gut feel that it will be important one day. So, panic because it can never actually be encountered.

crates/doc/src/shape/widen.rs
jgraettinger: I think it's okay not too right now.

Enums and tuples would never be produced by schema inference starting from `Shape::nothing()`. Continuously-inferred schemas would never have them.

They _are_ useful for things like `flowctl preview` of a derivation, or suggesting a change of an existing schema given exemplar documents. A particular use case in my mind is SaaS sources that have enum'd schemas for various event types, where a new event type is introduced by the vendor (we've seen a bunch of these).

A commonality is these are non-production, human-in-the-loop usages of widen(), not something we're doing automatically.

crates/validation/src/errors.rs
jgraettinger: the debug representation is now pretty gnarly.

PR #1144, opened 2023-08-10
validation: rename canonical projection of /flow_document

psFried:
**Description:**

This addresses a rough UX when the collection schema contains a top-level property called `flow_document`. Previously, this would result in a confusing error, because the canonical projection for `/flow_document` would conflict with the automatic projection of the root document. This changes the generation of canonical projections so that it no longer generates a canonical projection for the `/flow_document` location.

**Workflow steps:**

Publish a collection where the schema has a top-level `flow_document` property, like:

```yaml
schema:
  properties:
    id: {type: string}
    flow_document: {type: object}
  required: [id]
key: [id]
```

Previously, this would result in an error. Now, there will not be a projection created for the inner `flow_document` property, and the field name `flow_document` will unambiguously refer to the root document.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1144)


jgraettinger (approved):
LGTM

PR #1146, opened 2023-08-12
supabase: add `inferred_schemas` migration and final fixes

jgraettinger:
New `inferred_schemas` table is much like `catalog_stats`, and is loaded by the same stats_loader task.

Also add a partition selector to the inferred-schemas L1 derivation, to read only capture logs (for now: in the future we'll add derivations) and to exclude it's own schema inferrences.

Finally, improve `flowctl preview` error when a selector is not matched to explicitly tell the user that no journals were found, and what the evaluated selector was.

**Workflow steps:**

New table / API `inferred_schemas` is available.

**Notes for reviewers:**

This migration has already been applied in production.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1146)

### DIFF
--- a/ops-catalog/schema-inference.flow.yaml
+++ b/ops-catalog/schema-inference.flow.yaml
@@ -1,5 +1,7 @@
 collections:
   ops.us-central1.v1/inferred-schemas/L1:
+    schema: schema-inference.schema.yaml
+    key: [/collection_name]
     derive:
       using:
         sqlite: {}
@@ -7,6 +9,20 @@ collections:
         - name: logs
           source:
             name: ops.us-central1.v1/logs
+            partitions:
+              include:
+                kind:
+                  - capture
+              exclude:
+                name:
+                  # Don't read our own inferences.
+                  - ops.us-central1.v1/inferred-schemas/L1
+                  - ops.us-central1.v1/inferred-schemas/L2
+
+                  # NOTE(johnny): In production, I've temporarily held back
+                  # a number of other task logs for $reasons. Do not publish
+                  # directly from ./ops-catalog/ into production quite yet.
+
           shuffle:
             key: [/shard/name] # Use partition-based shuffle.
           lambda: |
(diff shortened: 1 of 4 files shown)


supabase/migrations/22_inferred_schemas.sql
```diff
@@ -0,0 +1,28 @@
...
+    "schema"         json not null,
+    flow_document    json not null,
+    primary key (collection_name)
+);
+alter table inferred_schemas enable row level security;
+
+create policy "Users must be authorized to the collection name"
+  on inferred_schemas as permissive for select
+  using (exists(
+    select 1 from auth_roles('read') r where collection_name ^@ r.role_prefix
```
jshearer, line 14: Just to make sure, `auth_roles` is  the function that unrolls all of your direct and indirect grants, right? As in, this isn't checking just the collections you have direct grants to, but also the ones that your tenant has grants to, right?

ops-catalog/schema-inference.flow.yaml
```diff
@@ -1,21 +1,35 @@
...
+              include:
+                kind:
+                  - capture
+              exclude:
+                name:
+                  # Don't read our own inferences.
+                  - ops.us-central1.v1/inferred-schemas/L1
+                  - ops.us-central1.v1/inferred-schemas/L2
+
+                  # NOTE(johnny): In production, I've temporarily held back
```
jshearer, line 22: You've held back some captures with lots of logs I'm guessing?

jshearer (approved):
LGTM 👍

supabase/migrations/22_inferred_schemas.sql
jgraettinger: Correct. This returns anything you can access with a `read` capability (or above)

ops-catalog/schema-inference.flow.yaml
jgraettinger: Exactly, lots of logs that go back months.

PR #1147, opened 2023-08-14
flowctl: fix preview hanging on error

psFried:
**Description:**

`flowctl preview` would hang indefinitely when an error was encountered running the derivation. This fixes that to return immediately. In order to do that, we needed to explicitly shutdown the runtime, because the reason it was hanging was due to the runtime blocking on a future (reading stdin) that would never complete.

**Workflow steps:**

Run `flowctl preview` on a spec that results in a validation error. Previously, it would hang indefinitely. Now it'll return the error immediately.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1147)

### DIFF
--- a/crates/flowctl/src/main.rs
+++ b/crates/flowctl/src/main.rs
@@ -1,12 +1,15 @@
 use clap::Parser;
 
-#[tokio::main(flavor = "current_thread")]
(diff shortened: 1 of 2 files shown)


jgraettinger (approved):
LGTM

PR #1148, opened 2023-08-15
agent: fix next_auto_discovers view

psFried:
**Description:**

The `next_auto_discovers` view was incorrectly returning the entire `endpoint` to use as the `endpoint_config` of the discover. This resulted in a _deeply_ nested config after many periodic discovers, since the connector it was tested with would accept the broken config. Serves me right for not having a test for that view. I added a test here.

Also fixes a bug where disabled captures would still get discovered.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1148)


jgraettinger (approved):
LGTM

PR #1149, opened 2023-08-16
control-plane: exclude recently updated specs from next_auto_discovers

psFried:
**Description:**

Fixes another bug in the next_auto_discovers view, where it would not respect the `auto_discover_interval` and would publish much more frequently if the publications were successful. This fixes it by considering the `updated_at` timestamp of the `live_specs` table when checking to see if a capture is due for another auto discover.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1149)

### DIFF
--- a/supabase/migrations/20_auto_discovers.sql
+++ b/supabase/migrations/20_auto_discovers.sql
@@ -9,10 +9,13 @@ select
   coalesce((live_specs.spec->'autoDiscover'->>'addNewBindings')::boolean, false) as add_new_bindings,
   coalesce((live_specs.spec->'autoDiscover'->>'evolveIncompatibleCollections')::boolean, false) as evolve_incompatible_collections,
   connector_tags.id as connector_tags_id,
-  -- If there's not been any discovers, then we use the capture creation time as the starting point, so that we don't auto-discover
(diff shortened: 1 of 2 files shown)


supabase/migrations/20_auto_discovers.sql
```diff
@@ -31,7 +34,7 @@ where
   and coalesce(json_typeof(live_specs.spec->'autoDiscover'), 'null') != 'null'
 group by live_specs.id, connector_tags.id
 -- See comment on overdue_interval above
-having now() - coalesce(max(discovers.updated_at), live_specs.created_at) > connector_tags.auto_discover_interval
+having now() - greatest(max(discovers.updated_at), live_specs.updated_at) > connector_tags.auto_discover_interval
 -- This ordering isn't strictly necessary, but it
```
jshearer, line 38: Not changed in this PR, but this comment is cut off

jshearer (approved):
LGTM. Didn't realize/think about that a successful publication deletes the discover (though of course it makes sense to do so). Using the greater of latest discover and live spec modified time looks like the right thing to do in that case 👍

psFried:
FYI CI failed on a flaky test after I'd updated a comment, so I'm just gonna merge this anyway.

PR #1150, opened 2023-08-16
support `notBefore` & `notAfter` in derivations and materializations

jgraettinger:
Extend a derivation or materialization source specification to accept optional `notBefore` and `notAfter` timestamps.
When set, these date-times lower/upper bound the publication times of documents which are processed by a derivation or materialization.

This is done in two ways:
* By plumbing `notBefore` through as a journal `ReadRequest.BeginModTime` constraint, which will cause the broker to skip over any persisted fragments having a modification time < `notBefore`.
* By filtering read non-ACK documents which have a UUID Clock falling outside of the constrained time range.

**Workflow steps:**

Add `notBefore` and/or `notAfter` to the `source` stanza of your derivation or materialization.

I've updated the "Specification" block of the derivation and materialization concepts pages.

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1150)


jgraettinger:
I manually tested by hacking up a local stack to use `file:///` stores for collection & recovery fragments, with a 1m flush interval. This coerced the local broker to "persist" fragments to a directory, making them remote fragments with a ModTime. I verified that:

* `flowctl-go journals read --from` skips over data as-expected.
* A materialization with notBefore / notAfter materialized only the expected documents of a hello-world capture.
* The live ReadRequest issued by that materialization plumbed through a `BeginModTime` constraint.

psFried (approved):
LGTM

PR #1151, opened 2023-08-17
feature: Introduce `quickcheck`-based fuzzing for `Shape::widen()`, and fix two bugs that it found

jshearer:
1. We were widening strings using `.len()`, which returns different results than `.chars().count()`. [JSON Schema](https://json-schema.org/draft/2020-12/json-schema-validation.html#section-6.3) says:
   > The length of a string instance is defined as the number of its characters [..]
2. Fixed a bug inferring from `integer` -> `number`. The problem is that `NUMBER` is really a composite of `INTEGER | FRACTIONAL`, so if it’s already `INTEGER`, then the test of `type_ & INTEGER | FRACTIONAL` is `INTEGER`, not `INVALID`.

I'm somewhat conflicted on code organization here: I would kind of like to put all of the widen tests in one place inside `widen.rs`, but I also want to put the quickcheck stuff together, and don't want to smoosh all of it inside of `widen.rs`. So for now I left it where it is

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1151)

### DIFF
--- a/crates/doc/src/shape/widen.rs
+++ b/crates/doc/src/shape/widen.rs
@@ -8,7 +8,7 @@ use std::cmp::Ordering;
 impl StringShape {
     fn widen(&mut self, val: &str, is_first: bool) -> bool {
         if is_first {
-            let (min, max) = length_bounds(val.len());
+            let (min, max) = length_bounds(val.chars().count());
             self.format = Format::detect(val);
             self.max_length = Some(max);
             self.min_length = min;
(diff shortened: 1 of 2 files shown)


crates/doc/tests/shape_fuzz.rs
```diff
@@ -0,0 +1,165 @@
...
+                return false;
+            }
+        } else {
+            return false;
+        }
+    }
+    return true;
+}
+
+// Cases detected by quickcheck:
```
jgraettinger, line 138: nit: add or update these as regular unit tests in widen.rs ? we can simply verify the Shape that comes out, as other tests do -- it doesn't need to round-trip.

crates/doc/tests/shape_fuzz.rs
```diff
@@ -0,0 +1,165 @@
+use doc::{shape::schema::to_schema, Shape, Validator};
+use itertools::Itertools;
+use quickcheck::*;
```
jgraettinger, line 3: nit: can this `*` be narrowed ? Is there a trait that's required ?

jgraettinger (approved):
LGTM % nits

crates/doc/src/shape/widen.rs
```diff
@@ -32,14 +32,14 @@ impl StringShape {
             }
         }
 
-        if self.min_length as usize > val.len() {
-            self.min_length = length_bounds(val.len()).0;
+        if self.min_length as usize > val.chars().count() {
```
jgraettinger, line 35: nit: `val.chars().count()` is called multiple times (min and max cases). Hoist into variable?

PR #1152, opened 2023-08-18
Phil/cleanup examples

psFried:
**Description:**

Cleans up the `examples/` by fixing/updating some and deleting others.
Also updates some READMEs to better reflect the current state of Flow, particularly with regard to the removal of derivation registers and the introduction of sqlite derivations. The language of "accumulate in the database" has become ambiguous, and I tried to make that more clear.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1152)


examples/derive-patterns/join-inner.flow.yaml
```diff
@@ -9,26 +9,50 @@ collections:
...
+        - name: fromInts
+          source: patterns/ints
           shuffle: { key: [/Key] }
-          update: { lambda: typescript }
-          publish: { lambda: typescript }
+          lambda: |
+            insert into join_state (key, lhs) values ($Key, $Int)
+            on conflict (key) do update set lhs = lhs + $Int;
+            -- now emit the joined result
+            SELECT JSON_OBJECT(
```
jgraettinger, line 33: nit / not required: a pattern I've found quite nice when there are multiple lambdas (and would recommend to users), is to define a view in terms of join state which is the desired output of the derivation.

Then, across the multiple lambdas, do an upsert and then a `select * from output_view where key = $Key;`

examples/derive-patterns/join-one-sided.flow.yaml
```diff
@@ -9,6 +9,42 @@ collections:
       required: [Key]
     key: [/Key]
 
+    derive:
+      using:
+        sqlite:
+          migrations:
+            - |
+              create table join_state (
```
jgraettinger, line 17: nit: these are potentially close enough that you could reference a common migration SQL?

jgraettinger (approved):
LGTM!

Some nits below but they're entirely optional / bonus. happy to merge as-is.

PR #1154, opened 2023-08-18
docs: draft updates to guides and tutorial

oliviamiannone:
**Description:**

**Workflow steps:**

**Notes for reviewers:**


site/docs/getting-started/tutorials/continuous-materialized-view.md
```diff
@@ -13,9 +16,9 @@ a materialized view that updates continuously based on a real-time data feed.
 
 * An Estuary Flow account. If you don't have one, visit the [Flow web app](https://dashboard.estuary.dev) to register for free.
 
-* The [flowctl CLI installed](../installation.mdx#get-started-with-the-flow-cli) (for the optional section).
+* A GitLab, GitHub, or BitBucket account. You'll use this to log into [GitPod](https://www.gitpod.io/), the cloud development environment integrated with Flow.
```
psFried, line 19: My 2 cents is that we should still at least mention that installing flowctl locally is an option. Gitpod is probably still easiest for most people, but it feels pretty weird for that to be the _only_ option.

If you did that, then we'd probably also need another note somewhere near the derivation YAML that says something like "if you're using flowctl locally, just paste this into a flow.yaml file and change 'yourprefix' to your tenant name".

I'm not super strongly attached to this, but it's definitely what I'd prefer if I was following the tutorial myself.

site/docs/getting-started/tutorials/continuous-materialized-view.md
```diff
@@ -37,84 +40,194 @@ First, you'll scope the raw data down to a small fact table with a derivation.
...
 
-1. On the [CLI-API tab](https://dashboard.estuary.dev/admin/api) in the Flow web app, copy the access token.
+2. Search for `demo/wikipedia/recentchange` and click on its name.
 
-2. Authorize Flow in your local development environment:
+3. On the **Collection Details** page, click the **Spec** tab.
 
-   ```console
-   flowctl auth token --token <your-token-here>
+   The collection is keyed on its metadata, so every new change event is seen as unique. Its schema has many fields.
```
psFried, line 51: _Technically_, the key of this collection is not really correct, in that it does not uniquely identify each change. The incorrectness doesn't matter in this case, because we don't reduce documents before feeding them into derivations. But in either case, I think it'd be better to just say that this collection is _quite_ large due to handling so many wikipedia edits.

It might also be a good idea to offer users the option of using the `-sampled` version of this collection, just so it goes faster and costs less. AFAIK they could just add `-sampled` to the end of the collection name and wouldn't have to change anything else.

psFried (approved):
Looks great!

PR #1155, opened 2023-08-19
Phil/fix agent authz

psFried:
**Description:**

Fixes an authorization bug that was introduced in #1138 that would result in a publication being rejected incorrectly. Also fixes a sql migration file conflict.

Turns out the reason our tests didn't catch it was that our test setup only had a single role grant. Adding another role grant to the test setup was able to easily reproduce the failure seen in production, so I left that extra role grant in the setup.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1155)


jgraettinger (approved):
LGTM

Oof...

PR #1157, opened 2023-08-22
Improve schema limits

jshearer:
**Description:**

* Replace per-subschema field limit with a global complexity limit, currently set at 1000 total fields.
* Add a `--flow.enable-schema-inference` flag to `flowctl-go serve consumer` to enable schema inference for capture tasks

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1157)

### DIFF
--- a/crates/doc/src/shape/limits.rs
+++ b/crates/doc/src/shape/limits.rs
@@ -1,63 +1,164 @@
 // This module defines limits which are used to simplify complex,
 // typically inferred schema Shapes.
 use super::*;
-use json::Location;
+use crate::{ptr::Token, Pointer};
+use itertools::Itertools;
+use std::cmp::Ordering;
 
-// Prune any locations in this shape that have more than the allowed fields,
-// squashing those fields into the `additionalProperties` subschema for that location.
-pub fn enforce_field_count_limits(slf: &mut Shape, loc: Location) {
-    // TODO: If we implement inference/widening of array tuple shapes
-    // then we'll need to also check that those aren't excessively large.
-    if slf.type_.overlaps(types::ARRAY) {
-        if let Some(array_shape) = slf.array.additional_items.as_mut() {
-            enforce_field_count_limits(array_shape, loc.push_item(0));
+fn resolve_shape_mut(shape: &mut Shape, field: Token) -> Option<&mut Shape> {
+    match field {
+        Token::Index(idx) => shape.array.tuple.get_mut(idx),
+        Token::Property(prop_name) if prop_name == "*" => {
+            shape.object.additional_properties.as_deref_mut()
         }
-    }
-
-    if !slf.type_.overlaps(types::OBJECT) {
-        return;
-    }
-
-    let limit = match loc {
-        Location::Root => MAX_ROOT_FIELDS,
-        _ => MAX_NESTED_FIELDS,
-    };
-
-    if slf.object.properties.len() > limit {
-        // Take all of the properties' shapes and
-        // union them into additionalProperties
-
-        let existing_additional_properties = slf
-            .object
-            .additional_properties
-            // `Shape::union` takes owned Shapes which is why we
-            // have to take ownership here.
-            .take()
-            .map(|boxed| *boxed)
-            .unwrap_or(Shape::nothing());
-
-        let merged_additional_properties = slf
+        Token::Property(prop_name) => shape
             .object
             .properties
-            // As part of squashing all known property shapes together into
-            // additionalProperties, we need to also remove those explicit properties.
-            .drain(..)
-            .fold(existing_additional_properties, |accum, mut prop| {
-                // Recur here to avoid excessively large `additionalProperties` shapes
-                enforce_field_count_limits(&mut prop.shape, loc.push_prop(&prop.name));
-                Shape::union(accum, prop.shape)
-            });
+            .iter_mut()
+            .find(|prop| *prop.name == *prop_name)
+            .map(|inner| &mut inner.shape),
+        Token::NextIndex => shape.array.additional_items.as_deref_mut(),
+    }
+}
 
-        slf.object.additional_properties = Some(Box::new(merged_additional_properties));
-    } else {
-        for prop in slf.object.properties.iter_mut() {
-            enforce_field_count_limits(&mut prop.shape, loc.push_prop(&prop.name))
+fn squash_location_inner(shape: &mut Shape, name: Token) {
+    match name {
+        Token::Index(idx) => {
+            // Remove location from parent properties
+            let shape_to_squash = shape.array.tuple.remove(idx);
+
+            if let Some(addl_items) = &shape.array.additional_items {
+                shape.array.additional_items =
+                    Some(Box::new(Shape::union(*addl_items.clone(), shape_to_squash)));
+            } else {
+                shape.array.additional_items = Some(Box::new(shape_to_squash));
+            }
+        }
+        Token::Property(prop_name) => {
+            let prop_position = shape
+                .object
+                .properties
+                .iter()
+                .position(|prop| *prop.name == *prop_name);
+
+            if let Some(prop_position) = prop_position {
+                // Remove location from parent properties
+                let ObjProperty {
+                    shape: shape_to_squash,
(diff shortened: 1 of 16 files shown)


jgraettinger (commented):
A few comments, looking good!

crates/doc/src/shape/limits.rs
```diff
@@ -1,63 +1,164 @@
...
+
+                // First check to see if it matches a pattern
+                // and if so squash into that pattern's shape
+                if let Some(pattern) = shape
+                    .object
+                    .pattern_properties
+                    .iter_mut()
+                    .find(|pattern| regex_matches(&pattern.re, &prop_name))
+                {
+                    pattern.shape = Shape::union(pattern.shape.clone(), shape_to_squash)
```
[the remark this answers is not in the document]
jshearer: I went down this path and found out that there's no `replace_with` (see [this RFC](https://github.com/rust-lang/rfcs/pull/1736)), which is what I want here: take ownership of `pattern.shape`, compute something new using that owned value, and put that mapped value back. Instead it looks like we must replace with _something_.. so I'll change this to replacing with `Shape::nothing()`, only to throw that value away once `union()` completes. This is still significantly better than cloning.

crates/doc/src/shape/limits.rs
```diff
@@ -1,63 +1,164 @@
...
-            enforce_field_count_limits(&mut prop.shape, loc.push_prop(&prop.name))
+/// Reduce the size/complexity of a shape while making sure that all
+/// objects that used to pass validation still do.
+pub fn enforce_shape_complexity_limit(shape: &mut Shape, limit: usize) {
+    let mut locations = shape
+        .locations()
+        .into_iter()
+        .filter_map(|(ptr, _, shape, _)| {
+            if ptr.0.len() > 0 {
+                Some((ptr, shape.clone()))
```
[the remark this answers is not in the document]
jshearer: Oops, leftover. removed 👍

crates/doc/src/shape/limits.rs
```diff
@@ -1,63 +1,164 @@
...
+            .pop()
+            .expect("locations vec was just checked to be non-empty");
+
+        squash_location(shape, location_ptr);
     }
 }
 
-pub const MAX_ROOT_FIELDS: usize = 750;
-pub const MAX_NESTED_FIELDS: usize = 200;
+pub const DEFAULT_SCHEMA_COMPLEXITY_LIMIT: usize = 2000;
```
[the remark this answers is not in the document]
jshearer: @dyaffe asked for 2k because redshift has a 1,600 column limit. Happy to do either one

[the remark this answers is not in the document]
jgraettinger: we discussed, 1k was the resolution. Thanks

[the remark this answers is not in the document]
jshearer: ok!

crates/doc/src/shape/limits.rs
```diff
@@ -1,63 +1,180 @@
...
+use itertools::Itertools;
+use std::cmp::Ordering;
+
+// This logic somewhat overlaps with [`Shape::locate_token`], but
+// here we don't care about recursion, we don't care about how the location
+// may or may not exist, and most importantly we need an &mut Shape.
+fn resolve_shape_mut(shape: &mut Shape, field: Token) -> Option<&mut Shape> {
+    match field {
+        Token::Index(idx) => shape.array.tuple.get_mut(idx),
+        Token::Property(prop_name) if prop_name == "*" => {
```
jgraettinger, line 14: nit: the `Token::Property` and `Token::NextIndex` cases are no longer possible, right? Can the become `unreachable!` ?

crates/doc/src/shape/limits.rs
```diff
@@ -1,63 +1,180 @@
...
-                enforce_field_count_limits(&mut prop.shape, loc.push_prop(&prop.name));
-                Shape::union(accum, prop.shape)
-            });
-
-        slf.object.additional_properties = Some(Box::new(merged_additional_properties));
-    } else {
-        for prop in slf.object.properties.iter_mut() {
-            enforce_field_count_limits(&mut prop.shape, loc.push_prop(&prop.name))
+            shape_to_squash.array.additional_items =
+                squash_addl_properties(shape_to_squash.array.additional_items);
```
jgraettinger, line 59: can't you just set `additional_items` and `additional_properties` to `None` ?

nit: if you can't and must keep this, then `squash_addl_properties` => `squash_addl` ? sometimes item, sometimes property

crates/doc/src/shape/limits.rs
```diff
@@ -1,63 +1,180 @@
...
+
+    if pointers.len() < limit {
+        return;
+    }
+
+    pointers.sort_by(|a_ptr, b_ptr| {
+        // order by depth, then lexicographically
+        match a_ptr.0.len().cmp(&b_ptr.0.len()) {
+            // Same depth, stably sort lexicographically
+            Ordering::Equal => a_ptr.to_string().cmp(&b_ptr.to_string()),
```
jgraettinger, line 163: optional overly-pedantic improvement: It'd make sense to me, to implement a total ordering for `doc::Pointer`.

There's still a small bit of weirdness where `/foo/9/stuff` will get pruned before `/foo/10/stuff`, which won't actually come up in practice because we don't infer tuples. It's also mildly more efficient (avoids string serialization).

jgraettinger (approved):
LGTM % final comments!

crates/doc/src/shape/limits.rs
jshearer: Ah, this was as simple as deriving `PartialOrd, Ord` for `doc::Pointer`, neat. Done!

crates/doc/src/shape/limits.rs
```diff
@@ -1,63 +1,180 @@
...
 }
 
-pub const MAX_ROOT_FIELDS: usize = 750;
-pub const MAX_NESTED_FIELDS: usize = 200;
+fn squash_location(shape: &mut Shape, location: &[Token]) {
+    match location {
+        [] => unreachable!(),
+        [first] => squash_location_inner(shape, first),
+        [first, more @ ..] => {
+            let inner = resolve_shape_mut(shape, first.to_owned()).expect(&format!(
```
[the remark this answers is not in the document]
jshearer: I inlined it 👍

crates/doc/src/shape/limits.rs
jshearer: Yup, `unreachable!`'d them

crates/doc/src/shape/limits.rs
```diff
@@ -1,63 +1,180 @@
...
+/// Reduce the size/complexity of a shape while making sure that all
+/// objects that used to pass validation still do.
+pub fn enforce_shape_complexity_limit(shape: &mut Shape, limit: usize) {
+    let mut pointers = shape
+        .locations()
+        .into_iter()
+        .filter_map(|(ptr, _, _, _)| {
+            let last = ptr.0.last();
+            // Checking that the pointer has at least one item in it
+            if let Some(last) = last {
```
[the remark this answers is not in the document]
jshearer: Totally thought I was using `is_additionalx_field` elsewhere, but I'm not. Inlined 👍

crates/doc/src/shape/limits.rs
jgraettinger: Oh, that's perfect. :heart: Rust.

crates/doc/src/shape/limits.rs
jshearer: I take that back. Consider
```yaml
type: object
additionalProperties:
  type: object
  properties:
    foo: string
    bar: integer
```
We need to be able to squash `/*/foo` and `/*/bar` into `/*`. I'm looking at what we need to do to make this work..

jshearer: Ok, I think the solution is: filter out locations like `/foo/bar/*` where the `*` or `-` is a leaf node. Then these can still be `unreachable!`, and we reduce the number of locations we walk

PR #1159, opened 2023-08-23
Fix task failures sending over-large blocks to Slack

jshearer:
Limit message blocks generated by `ops.us-central1.v1/failure-notifications/v1` to be under the 3000 character limit.

This was deployed, but never committed so got rolled back. I deployed this again just now, so this is already live

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1159)


jgraettinger (approved):
LGTM

PR #1160, opened 2023-08-23
Phil/update gazette

psFried:
**Description:**

Rolls up a couple of miscellaneous fixes.
- Updates the rust protobuf-generated code to include the `notBefore` changes
- Updates the gazette dependency to the lastest tip of master, to pull in gazette/core#347

@skord, this ought to fix the `shards prune` errors, so we should be good to try that job again after this merges

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1160)


jgraettinger (approved):
LGTM

PR #1163, opened 2023-08-25
feature: Implement `billing_historicals` table

jshearer:
Also adds `internal.freeze_billing_month(billed_month timestamptz)` to populate `billing_historicals`

Also update `billing_report_202308()` to allow callers with `BYPASSRLS` set to generate the billing report for any tenant. This is so that a call to `freeze_billing_month()` can generate billing data for every tenant, without the caller (`postgres`) having admin grants.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1163)


supabase/migrations/21_billing_v2.sql
```diff
@@ -174,7 +177,13 @@ begin
...
   -- Verify that the user has an admin grant for the requested `billed_prefix`.
   perform 1 from auth_roles('admin') as r where billed_prefix ^@ r.role_prefix;
-  if not found then
+  has_admin_grant = found;
+
+  -- Check whether user has bypassrls flag
+  perform 1 from pg_roles where rolname = session_user and rolbypassrls = true;
+  has_bypassrls = found;
+
+  if not has_bypassrls and not found then
```
jgraettinger, line 186: `if not has_bypassrls and not has_admin_grant then`

supabase/pending/billing_historicals.sql
```diff
@@ -0,0 +1,234 @@
+create table "public"."billing_historicals" (
```
jgraettinger, line 1: we're not doing `pending` migrations anymore. Instead, add append-only new migrations to `supabase/migrations`.

supabase/migrations/24_billing_historicals.sql
```diff
@@ -0,0 +1,87 @@
+
+-- Always use a transaction, y'all.
+begin;
+
+-- Historical record of tenant billing statements. This structure
+-- comes from the return value of `billing_report_202308`.
+create table billing_historicals (
+    tenant              catalog_tenant  not null,
+    billed_month        timestamptz     not null,
+    line_items          jsonb           not null,
```
jgraettinger, line 10: I don't have confidence that these metrics are 100% right and durable, and I think we want to avoid codifying them in the structure of this table. Instead let's keep it really lean and just have:

* tenant
* billed_month
* report jsonb <- the complete jsonb output of the billing report

i suspect this also streamlines the UI, which is then just working with billing report JSON objects -- sometimes from this table, and sometimes from calling the API directly.

supabase/pending/billing_historicals.sql
jshearer: Okay, so this means that I'll largely be duplicating `billing_report_202308` in the new migration

supabase/migrations/21_billing_v2.sql
jshearer: 🤦 thanks

supabase/migrations/24_billing_historicals.sql
jshearer: Yup sounds good, I was on the fence about whether not we should be extracting these fields or not. Changed to capturing and JSON, and yes agreed it'll make the UI work easier in the short term, though it will make having inconsistent data in here easier in the long term which is why I split it out in the first place

supabase/migrations/24_billing_historicals.sql
```diff
@@ -0,0 +1,230 @@
...
+create table billing_historicals (
+    billed_prefix       catalog_tenant  not null,
+    billed_month        timestamptz     not null,
+    report              jsonb           not null,
+
+    check (date_trunc('month', billed_month) = billed_month),
+    unique (billed_prefix, billed_month)
+);
+alter table billing_historicals enable row level security;
+grant all on billing_historicals to postgres;
```
jgraettinger, line 16: nit: this grant to postgres is redundant (postres owns it).

supabase/migrations/24_billing_historicals.sql
```diff
@@ -0,0 +1,230 @@
...
+returns integer as $$
+declare
+    tenant_row record;
+    tenant_count integer = 0;
+begin
+    for tenant_row in select tenant as tenant_name from tenants loop
+        tenant_count = tenant_count + 1;
+        insert into billing_historicals
+        select
+            tenant_row.tenant_name as billed_prefix,
```
jgraettinger, line 37: nit: can this be `report->>billed_prefix, report->>billed_month` or does that not work for some reason?

supabase/migrations/24_billing_historicals.sql
```diff
@@ -0,0 +1,230 @@
+
+-- Always use a transaction, y'all.
+begin;
+
+-- Historical record of tenant billing statements. This structure
+-- comes from the return value of `billing_report_202308`.
+create table billing_historicals (
+    billed_prefix       catalog_tenant  not null,
```
jgraettinger, line 8: optional: should this be `tenant` ? I'm cool with `billed_prefix` instead, though, and am on the fence myself.

supabase/migrations/24_billing_historicals.sql
```diff
@@ -0,0 +1,230 @@
...
+    tenant_count integer = 0;
+begin
+    for tenant_row in select tenant as tenant_name from tenants loop
+        tenant_count = tenant_count + 1;
+        insert into billing_historicals
+        select
+            tenant_row.tenant_name as billed_prefix,
+            date_trunc('month', billed_month) as billed_month,
+            report
+        from billing_report_202308(tenant_row.tenant_name, date_trunc('month', billed_month)) as report;
```
jgraettinger, line 40: optional nit: you can skip this date_trunc and let the billing report worry about it :shrug:

supabase/migrations/24_billing_historicals.sql
```diff
@@ -0,0 +1,230 @@
...
+$$ language plpgsql volatile;
+
+comment on table billing_historicals is
+    'Historical billing statements frozen from `billing_report_202308()`.';
+comment on column billing_historicals.billed_prefix is
+    'The tenant for this statement';
+comment on column billing_historicals.billed_month is
+    'The month for this statement';
+comment on column billing_historicals.report is
+    'The report generated by billing_report_202308()';
```
jgraettinger, line 53: nit: The historical billing report.

jgraettinger (approved):
LGTM! % some final small nits

supabase/migrations/24_billing_historicals.sql
jshearer: So I actually had it as `tenant` originally, then changed it to `billed_prefix` to match the response coming back from `billing_report_202308()` so the UI could treat them interchangeably... but now I'm realizing that I could have just aliased the field when querying since postgrest lets you do that. I do think that `tenant` is slightly more correct -- we want to bill based on tenants, and we populate this table by iterating over tenants. I'm very slightly leaning towards changing it

jgraettinger: if the UI just fetches out the `report` field, that will contain `billed_prefix` in the same way that the RPC report does, if that's helpful for consolidation?

jshearer: Yeah that's a good point. I don't really love this duplication but at least we talked about it and decided that it's worth it to leave the exact shape of `report` up to the report generator instead of making it explicit here.

PR #1164, opened 2023-08-26
examples: add derive pattern k/v pairs example

psFried:
Adds an 'entity-attribute-values' example to derive-patterns, which shows how to normalize these values by rolling them up by entity id.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1164)


jgraettinger (approved):
LGTM

PR #1166, opened 2023-08-30
Implement `tenants.trial_start` and associated calculations in billing report

jshearer:
Update billing to support free trials. Fundamentally, the idea is to generate a list of all dates in the specified month along with their usage and subtotal. If there is a free trial, then sum up the subtotals for the days that are included in the free trial period, and provide a credit for that sum. 

One wrinkle here is that originally, we were dealing with monthly stats. This meant that we could comfortably multiply usage by rate and `round()` the result. Now that we're potentially dealing with summing up daily stats, the `round()`ing _must_ happen after summing, since `round(sum([list_of_real_numbers]))` != `sum(round([list_of_real_numbers]))`. Mechanically this is all fine, but it's a subtlety worth mentioning, and explains why `internal.incremental_usage_report()` must return `{"subtotal_frac": 1.234, ..}` and not `subtotal`. 

I ended up refactoring this into a couple different functions to ~preserve my sanity~ improve readability:

* **internal.compute_incremental_line_items()** deals with converting usage and usage tiers to line items incrementally. That is, rather than generating line items all at once, we can compute only the line items for the difference in usage (for a particular day, for example) while still keeping track of the total usage so far in order to determine which tier we're currently in.
* **internal.incremental_usage_report** is responsible for computing all incremental line items over a specified range, essentially driving `internal.compute_incremental_line_items()` with data from `catalog_stats`. **Note** that because we've defined `data_tiers` and `usage_tiers` as covering a specific month and then resetting, this function is inherently limited to a range of at most one month.
* **billing_report_202308** is no longer responsible for computing line items for usage. Instead, in addition to keeping its existing logic around adjustments and recurring fixed costs, it calls out to `internal.incremental_usage_report()` using the `monthly` granularity to generate the usage-based line items, and uses the `daily` granularity to calculate the free trial credit. It also now includes the list of incremental usage reports generated by `internal.incremental_usage_report` for each day of the month, with an empty fallback if no stats were recorded on that day. 

**Question:** Currently, the output of `billing_report_202308` includes line items in its per-day usage list. I'm visualizing these as mapping to different sections of a bar in a bar chart of daily usage, like this: Screen Shot 2023-09-05 at 17 16 19

That being said, the mapping isn't 1:1. For example, I would not expect to care about different billing _tiers_ when displaying that bar chart, which implies that maybe line items aren't actually what we want here and they should be left out in favor of some unspecified future work to break down billing by category (SKU? SKU grouping?) rather than line item. 

Additionally, getting rid of it would be nice because I couldn't think of a way to expose rounded subtotals in the daily line items that all add up to the final subtotal, because of the above-mentioned "round-before-sum != sum-before-round" problem. Thoughts?

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1166)

### DIFF
--- a/supabase/tests/billing.test.sql
+++ b/supabase/tests/billing.test.sql
@@ -193,19 +193,19 @@ begin
         "rate": 30,
         "count": 4,
         "subtotal": 120,
-        "description": "Data processing (first 4GB at $0.30/GB)"
+        "description": "Data processing (first 4 GBs at $0.30/GB)"
       },
       {
         "rate": 25,
         "count": 6,
         "subtotal": 150,
-        "description": "Data processing (next 6GB at $0.25/GB)"
+        "description": "Data processing (next 6 GBs at $0.25/GB)"
       },
       {
         "rate": 20,
         "count": 20,
         "subtotal": 400,
-        "description": "Data processing (next 20GB at $0.20/GB)"
+        "description": "Data processing (next 20 GBs at $0.20/GB)"
       },
       {
         "rate": 15,
@@ -253,7 +253,9 @@ begin
     "processed_data_gb": 43.125,
     "recurring_fee": 10000,
     "subtotal": 22044,
-    "task_usage_hours": 738.375
+    "task_usage_hours": 738.375,
+    "trial_credit": 0,
+    "trial_start": null
   }'::jsonb);
 
   set role postgres;
@@ -277,7 +279,7 @@ begin
         "rate": 50,
         "count": 21.125,
         "subtotal": 1056,
-        "description": "Data processing (first 30GB at $0.50/GB)"
+        "description": "Data processing (first 30 GBs at $0.50/GB)"
       },
       {
         "rate": 20,
@@ -295,7 +297,9 @@ begin
     "processed_data_gb": 21.125,
     "recurring_fee": 0,
     "subtotal": 11856,
-    "task_usage_hours": 720
+    "task_usage_hours": 720,
+    "trial_credit": 0,
+    "trial_start": null
   }'::jsonb);
 
   -- We're authorized as Bob.
@@ -311,7 +315,7 @@ begin
   -- Switch tiers so usage spills over
   -- and set trial so half of August is covered
   update tenants set
-    free_trial_start='2022-08-15',
+    trial_start='2022-08-15',
     data_tiers = '{50, 5, 20}'
     where tenant = 'aliceCo/';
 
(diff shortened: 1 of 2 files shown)


jgraettinger (commented):
This feels ... formidable to follow along with. As a meta-comment, this is billing logic and our goal should be that, for readers, it's very obvious, very obviously correct, and boring.

I don't have a fully-formed recommendation for how it can be improved, so this is going to be a hand-wavy articulation, but: the only actual requirement here is that we produce an array of daily additive subtotals. We don't need daily line-items, we don't even need daily categorized spend. And my spidy-sense is that we should avoid fine-grain daily categorizations, even if it might make for pretty plots, because it's a commitment to a particular billing framing that I don't think we're ready to make or stick to.

So, from a response shape standpoint, I'd really like this just to be the _current_ billing report plus an array of integers (cents), each representing the day's additive amount to the subtotal. I don't think we've even quite defined whether billing adjustments and recurring fees should be rolled into the daily numbers. For the moment I'd say "no" but we may need to revisit in the future.

In terms of implementation: we have some current logic that takes a `processed_data_gb` value and some tiers, and then returns an array of line-items. Ditto for `task_usage_hours`. Can we start by hoisting those into two functions?

We can trivially update the existing implementation to call those two functions with the current month-grain values, leaving everything else as-is.

Then, for computing this per-day array: I could imagine a new routine that walks daily stats using a SQL window query, rolling up cumulative `processed_data_gb` and `task_usage_hours` by-day. Then, in a loop for each day:

 * Call those two functions and sum up the line-item subtotals into a total-so-far (discard the line-items).
 * Subtract the previous cumulative subtotal -- initially zero -- from the total-so-far. That's the current day's incremental cost. Push it onto the array.

One other consideration i'm still thinking about: there's no guarantee that the added daily quantities exactly match the month, given rounding, which means our credit could be off by a cent or so...

jshearer:
I agree that it's daunting, and I'm honestly not sure if SQL is the right language to be writing this in. But fundamentally, I actually think that I'm already doing what you suggested:

> we have some current logic that takes a `processed_data_gb` value and some tiers, and then returns an array of line-items. Ditto for `task_usage_hours`. Can we start by hoisting those into two functions?

Yup, that was the first thing I did. 
 `internal.compute_incremental_line_items()` is that function. Since the logic is the same for data and hours (even if the tiers are different), this ended up as just one function that takes a couple string parameters in order to generate the correct line items (`"hours"` vs `"GB"`, etc).

I waffled back and forth on whether the incremental generator should return line items or not, and clearly I should have just thrown them away since they're confusing, but fundamentally the problem that's solved by generating a list of per-day line items (even if we throw them away immediately) is that in order to correctly compute the incremental cost, we have to deal with crossing multiple usage tiers in a single day. It's trivial to modify this PR to just sum up and then throw away the daily line items instead of keeping them.

jshearer:
The other reason I kept in the line-items is that process of generating line items is intrinsically linked to calculating cost. So rather than implement that cost/line-item logic twice, once for daily usage and once for monthly, it seemed more DRY to just parameterize `internal.incremental_usage_report()` over the granularity: `monthly` to generate a single whole-month cost/line-item report, and `daily` to generate a list of per-day costs (and line items that clearly should be thrown away because they're confusing).

jshearer:
>there's no guarantee that the added daily quantities exactly match the month, given rounding, which means our credit could be off by a cent or so...
I haven't thoroughly modeled this out but in practice I saw errors of up to 5 cents between the sum of daily usage and monthly usage.

jshearer:
>I don't think we've even quite defined whether billing adjustments and recurring fees should be rolled into the daily numbers. For the moment I'd say "no" but we may need to revisit in the future.

Agreed, this PR does not roll them into the daily numbers. The consequence is that the daily numbers won't line up with the monthly invoice subtotal... which I think is fine for now.

jgraettinger:
> I haven't thoroughly modeled this out but in practice I saw errors of up to 5 cents between the sum of daily usage and monthly usage.

Yikes. Perhaps this can be addressed by initializing with the whole-month subtotal (the true value) and walking backwards to subtract daily deltas.

jshearer:
Ah sorry, I think I miscommunicated: this PR fixed the problem that caused those discrepancies: rounding daily values before summing. Changing it to tracking fractional amounts per-day, and summing at the end to calculate the free trial credit fixed that, and this PR outputs exact free trial credits.

jgraettinger:
FYI I'm spending some time putting my money where my mouth is, that this can be simplified. I'll put up a sketch PR that we can look at and talk about.

Also, I think I got the requirement that the daily array should have one entry per date wrong: graphing libraries (d3 et all) work directly with `Date` instances for chart layout, etc, so this is probably just making work for the UI, as compared to returning an array of `[{"ts": "2023-0X-YZT00:00:00Z", "subtotal": 1234}, ...]`. The latter is also more extend-able in the future if it turns out we do want to break out categories of spend on a per-day basis.

jgraettinger:
Okay, [here's what I came up with](https://gist.github.com/jgraettinger/ad3a660ce1c51c893ea21fd7cf2972be). It runs and I believe does the write things, though I haven't fully updated tests.

It eschews all loops except for the factored-out calculation of tiers, and frames as much as possible as a progression of declarative CTEs. Window functions are used to map from deltas metrics to cumulative ones, and back again. I did end up including separate data and task subtotals, because it felt pretty clean to do so.

jshearer:
I pulled in your gist and updated the tests. I made a couple changes:
* The free trial credit was showing `Free trial credit (2022-08-15 - 2022-09-15 00:00:00)` so I changed it to a date
* Changed to always show the free trial credit, even if it's 0 so long as you're in the free trial range (to make it clear that you're under the free trial, even if you have no usage)

supabase/tests/billing.test.sql
```diff
@@ -159,8 +159,12 @@ begin
     catalog_name, grain, ts, flow_document, bytes_written_by_me, bytes_read_by_me, usage_seconds
   ) values
     ('aliceCo/aa/hello', 'monthly', '2022-08-01T00:00:00Z', '{}', 5.125 * 1024 * 1024 * 1024, 0, 3600 * 720),
```
jgraettinger, line 161: nit: these monthly grains can be removed altogether, right?

jgraettinger (approved):
LGTM

supabase/tests/billing.test.sql
jshearer: Yep! Removed

PR #1174, opened 2023-08-31
flowctl: change to new --partitions argument

psFried:
**Description:**

Replaces the --include-partition and --exclude-partition arguments of `flowctl collections read|list-journals|list-fragments` with a new `--partitions` argument. The new argument accepts a JSON object, which is deserialized into a `models::PartitionSelector`, and thus matches the shape of the selector that's used in catalog specs.

The motivation for this is that the old arguments didn't allow for reading from partitions where the value was a json `null`. When it was written, we didn't allow partitioning on nullable fields. Now that we do, the json syntax seemed desirable as it allows distinguishing between json types, so that `null` can be differentiated from the string `"null"`.

**Workflow steps:**

If you'd previously used `--include-partition foo=bar`, then you'll now have to use `--partitions '{"include": {"foo": ["bar"]}}'`

This isn't documented, except in the CLI help.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1174)

### DIFF
--- a/crates/flowctl/src/collection/mod.rs
+++ b/crates/flowctl/src/collection/mod.rs
@@ -1,7 +1,7 @@
 pub mod read;
 
 use crate::Timestamp;
-use assemble::percent_encode_partition_value;
+use anyhow::Context;
 use journal_client::{fragments, list};
 use proto_gazette::broker;
 use time::OffsetDateTime;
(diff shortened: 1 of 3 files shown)


crates/flowctl/src/collection/read/mod.rs
```diff
@@ -22,10 +22,6 @@ pub async fn get_collection_inferred_schema(
     ctx: &mut crate::CliContext,
     args: &SchemaInferenceArgs,
 ) -> anyhow::Result<()> {
-    if args.selector.exclude_partitions.len() > 0 || args.selector.include_partitions.len() > 0 {
```
psFried, line 25: This was incorrect before this PR, so I just deleted it. The infer-schema subcommand is due for a bit of re-work anyway, since it should probably just fetch them from the control plane.

crates/flowctl/src/collection/mod.rs
```diff
@@ -17,75 +17,39 @@ pub struct CollectionJournalSelector {
...
+    /// in Flow catalog specs. For example:
+    /// '{"include": {"myField1":["value1", "value2"]}}'
+    #[clap(
+        long,
+        value_parser(parse_partition_selector),
+        conflicts_with_all(&["include-partition", "exclude-partition"])
+    )]
+    pub partitions: Option<models::PartitionSelector>,
+
+    /// Deprecated, use --partitions instead
```
jgraettinger, line 31: If we're just gonna error out anyway, seems like we could just remove it altogether?

jgraettinger (approved):
LGTM

crates/flowctl/src/collection/mod.rs
psFried: 🤷 I know there's at least one person using it, so I figured why not give folks a better error message. We can always remove it the next time we're in there

PR #1175, opened 2023-09-01
Fix/billing report security

jshearer:
**Description:**

The only actual change to the billing report is to use `current_setting('role')` instead of `session_role` to check the role of the caller outside of `security definer`.

This also updates `run_sql_tests.sh` to return a non-zero exit code on test failure so we can fail the CI run when we break SQL tests in the future.

~~**NOTE:** not to be merged quite yet, I put an artificial failure in there to make sure the CI run fails as expected.~~
CI failure [works as expected](https://github.com/estuary/flow/actions/runs/6051532364/job/16423186539#step:9:30)!

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1175)

### DIFF
--- a/supabase/migrations/25_fix_billing_security.sql
+++ b/supabase/migrations/25_fix_billing_security.sql
@@ -1,0 +1,166 @@
+begin;
+
+-- Billing report which is effective August 2023.
+create or replace function billing_report_202308(billed_prefix catalog_prefix, billed_month timestamptz)
+returns jsonb as $$
+#variable_conflict use_variable
+declare
+  -- Auth checks
+  has_admin_grant boolean;
+  has_bypassrls boolean;
+  -- Retrieved from tenants table.
+  data_tiers  integer[];
+  usage_tiers integer[];
+  recurring_usd_cents integer;
+
+  -- Calculating tiered usage.
+  tier_rate    integer;
+  tier_pivot   integer;
+  tier_count   numeric;
+  remainder    numeric;
+
+  -- Calculating adjustments.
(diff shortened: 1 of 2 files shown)


supabase/run_sql_tests.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-set -o errexit
+# set -o errexit
```
psFried, line 3: nit: can just be deleted

supabase/migrations/25_fix_billing_security.sql
```diff
@@ -0,0 +1,166 @@
+begin;
```
psFried, line 1: I don't think we need to explicitly begin/commit transactions in migrations, since the supabase cli would do that already. Unless I'm missing something?

psFried (approved):
LGTM

supabase/migrations/25_fix_billing_security.sql
jshearer: Oh, I was just copying the other migrations that seemed to all say 
```sql
-- Always use a transaction, y'all.
begin;
```
at the top, though I am seeing some that don't have it... do you feel strongly that it should be removed? If not I think it's probably marginally safer to keep it

jgraettinger: I usually apply them with  psql, so it's good practice imo. Postgres doesn't mind if they're nested. Also I'm 70% that supabase doesn't wrap in a txn.

PR #1177, opened 2023-09-05
Phil/misc agent updates

psFried:
**Description:**

This rolls up a few different related updates to our agent. The general thrust is to address issues with `autoDiscover` and disabled bindings that have become clear now that those are in production. The main user-facing feature here is in the final commit, which defaults `autoDiscover` to enabled for new captures. The other commits are all necessary in order to enable that. The individual commits have more details, but the high level summary is:

- We need to start deleting old drafts, since `autoDiscover`s are creating a lot more of them
- Have `autoDiscover`s skip creating publications if none of the specs have changed, so that publication histories don't get too big/noisy
- Have the publications handler automatically trim out collection specs that are not bound to anything, so that we don't have to implement this behavior in both the UI and the discovers handler
- Finally, default `autoDiscover` to enabled

**Workflow steps:**

Creating any discover via the UI will default `autoDiscover` to enabled.

If you publish a catalog with "extra" collection specs (usually from disabled bindings), they will be automatically pruned from the draft catalog, and will not count against the tenant quota.

None yet

**Notes for reviewers:**

Best to review commit by commit

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1177)


psFried:
One other thing to note is that the UI currently prunes unbound collection specs when you click "Save and Publish" (but not when you click "Test"). This means that, in order to test this new behavior around pruning unbound collections, you need to publish via the CLI. The plan is to update the UI to stop pruning collections after this merges.

crates/agent/src/discovers.rs
```diff
@@ -166,38 +168,63 @@ impl DiscoverHandler {
...
             Err(errors) => {
                 draft::insert_errors(row.draft_id, errors, txn).await?;
-
-                Ok((row.id, JobStatus::MergeFailed))
+                return Ok((row.id, JobStatus::MergeFailed));
             }
+        };
+
+        // Prune out any unchanged specs, but only if we're automatically publishing.
+        // If a user is interactively discovering, then the UI will need all the specs,
```
jgraettinger, line 180: Does it? I'm missing something. Can't the UI use the existing spec? That's what it appears to do today, when you edit an existing capture and review a bound collection.

I'm thinking a) it would be nice to consolidate these branches, and b) publication history does get cluttered with no-op publications for collections that are touched but unchanged by a manual discover, so might as well prune those too.

crates/agent/src/discovers.rs
```diff
@@ -166,38 +168,63 @@ impl DiscoverHandler {
...
+            Some(id)
+        } else {
+            None
+        };
+
+        Ok((
+            row.id,
+            JobStatus::Success {
+                publication_id,
+                specs_unchanged: false,
```
jshearer, line 225: This isn't changed in this PR, but I feel like `specs_changed` would probably be better here. Now the reader has to figure out what `unchanged: false` means (it... wasn't not changed, so there _were_ changes?)

crates/flowctl/src/catalog/publish.rs
```diff
@@ -41,59 +41,23 @@ pub async fn remove_unchanged(
...
-
-        let new_spec_md5 = format!("{:x}", md5::compute(buf));
-
-        return *existing_spec_md5 != new_spec_md5;
-    }
+    sources::remove_unchanged_specs(&spec_checksums, &mut input_catalog);
 
-    // Catalog name does not yet exist in live specs.
-    true
+    // TODO(phil): This used to sneakily removed any storage mappings from the catalog. Need to re-add
```
jshearer, line 57: Reading commit by commit: does this need to get resolved before merging?

crates/agent-sql/src/publications.rs
```diff
@@ -287,7 +307,7 @@ pub async fn find_tenant_quotas(
                     where
                         live_specs.spec_type = 'capture' or
                         live_specs.spec_type = 'materialization' or
-                        live_specs.spec_type = 'collection' and live_specs.spec->'derivation' is not null
```
jshearer, line 290: I seem to remember something about how we still technically support `derivation` for backwards compatibility, right? Does that matter here (or have we since actually gotten rid of it?)

crates/agent/src/publications.rs
```diff
@@ -196,9 +201,54 @@ impl PublishHandler {
...
+        );
+        if !errors.is_empty() {
+            anyhow::bail!("unexpected errors from expanded specs: {errors:?}");
+        }
+
+        // It's important that we prune unbound collection specs _after_ the catalog is expanded,
+        // and before checking resource quotas
+        let pruned_collections =
+            specs::prune_unbound_collections(&mut draft_catalog, &mut spec_rows);
+        // `resolve_specifications` will have created `live_specs` for each of the `draft_specs` in the publication.
```
jshearer, line 221: Interesting... writing to `live_specs` ought to be one of the last steps of publishing, no? Does this mean that we temporarily leak incorrect `live_specs`? Does that not matter because everything here is happening inside a transaction?

crates/agent/src/publications.rs
```diff
@@ -196,9 +201,54 @@ impl PublishHandler {
...
+        // We need to delete any of those that correspond
```
jshearer, line 222: Unfinished comment?

jshearer (approved):
Had a couple questions but LGTM as far as I can tell

crates/agent/src/publications.rs
```diff
@@ -196,9 +201,54 @@ impl PublishHandler {
...
+                .iter()
+                .map(|r| (r.live_type, r.catalog_name.as_str(), r.live_spec.0.as_ref())),
+        );
+        if !errors.is_empty() {
+            anyhow::bail!("unexpected errors from expanded specs: {errors:?}");
+        }
+
+        // It's important that we prune unbound collection specs _after_ the catalog is expanded,
+        // and before checking resource quotas
+        let pruned_collections =
```
jgraettinger, line 219: I think we should be pruning in the DB, rather than via the `models::Catalog` and application code.

The reason why is that this change places new expectations on what, exactly, `expanded_specifications` is supposed to do which will make it difficult to refactor / update it's semantics. IMO `expanded_specifications` should be doing less, pulling fewer things into any given build. I think _probably_ it would all still be fine, but it also seems straightforward to identify specs to prune via a DB query instead.

The way this can work instead, is to first restore the previous control flow so that `apply_updates_for_row` is first called for each spec.

Next, immediately after `apply_updates_for_row`, you can query `live_specs` outer joined with `live_spec_flows`, and look for brand new rows having `spec_type = 'collection'` with no incoming or outgoing flows. That would give you the same result without any dependence on `expanded_specifications`.

ETA: OH! It's also not correct to call `expanded_specifications` _before_ applying updates for each row, because we need those updated `live_spec_flows` from `apply_updates_for_row` in order to power the expansion.

jgraettinger (changes requested):
I have some concerns about how pruning is implemented below.

Consolidating removal of unchanged specs to _all_ discover code paths is a 'would be nice' from me but isn't a deal-breaker.

Thanks!

crates/agent/src/discovers.rs
psFried: That's a good point. I guess I just assmumed they'd need to be a part of the draft in order for the UI to work, but if that's not the case, then I agree it'd be cleaner to just always prune them. I'll verify the UI behavior and make it unconditional if I can.

crates/agent/src/discovers.rs
psFried: This field is only serialized if the value is `true`:

```
        #[serde(skip_serializing_if = "std::ops::Not::not")]
        specs_unchanged: bool,
```

The intent is only to have something on the status to indicate when it's expected that the draft may still be empty after a successful discover.

jgraettinger (commented):
I took a quick pass focused on the changes and it looks great. I'll do a final review in the AM when I'm fresh.

crates/agent/src/publications.rs
psFried: this was superceded by the new approach / fixed

crates/agent/src/publications.rs
psFried: Yeah, we're inside a transaction, so nothing else will observe these changes unless it commits. We do write out the new live specs fairly early during the publication, as it simplifies some things. The new approach for pruning unbound collections will actually update all the live specs and then delete the ones that are unbound.

crates/agent-sql/src/publications.rs
psFried: yeah, we no longer have any specs in prod that use `derivation`.

crates/flowctl/src/catalog/publish.rs
psFried: I did hoist it up in a subsequent commit

crates/agent-sql/src/discovers.rs
```diff
@@ -161,6 +163,32 @@ where
     Ok(())
 }
 
+/// Returns a map of catalog_name to md5 hash of the live spec. The map will only
+/// include entities that exist and have a non-null md5 hash.
+pub async fn fetch_spec_md5_hashes(
```
jgraettinger, line 168: future refactor / not blocking: seems like this could be pretty naturally folded into `build_merged_catalog` and it's delegate calls to `agent_sql::discovers::resolve_merge_target_specs`, rather than being a separate pass ?

jgraettinger (approved):
LGTM!

jgraettinger:
Let's get this 🚢 'd!

PR #1178, opened 2023-09-06
end-to-end continuous schema inference

jgraettinger:
Introduce new soon-to-be-well-known URIs `flow://write-schema` and `flow://inferred-schema` which are available to collection read schemas.

These URIs may be `$ref`'d from within a read schema, and having done
so, their definitions are inlined into the effective read schema of the
built collection with every publication.
    
Extend the `validation::ControlPlane` trait to enable resolution of
collections to their current inferred schemas. Implement within
`flowctl` through our API, and within `agent` via out-of-transaction SQL
lookups done using the agent's Postgres pool.
   
Update the `agent` discovers handler to map the existing
`x-infer-schema: true` annotation into a synthesized initial read
schema, that composes the write & inferred schemas.

**Workflow steps:**

Use a connector that has the `x-infer-schema: true` annotation.
Then, every time its resulting collection is published (through a capture or materialization change), it's effective read schema is updated to the most-recent inferred schema of the collection.

If the nature of data changes, then the inferred schema also updates and a downstream use case (such as a materialization) fails with a schema violation. However, by next re-publishing the materialization (really, the collection, but we don't have an easy way to do this in the UI right now) the materialization is automatically fixed, and new columns are first added to the bound table _before_ the first usage of that column is materialized.

It doesn't appear we've got current docs on our janky schema inference service (which will be removed)? So it's not clear we have existing docs to update, but will need new docs on this feature.

**Notes for reviewers:**

Needs to be rebased on the "rust connectors: phase one" branch.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1178)


crates/validation/src/collection.rs
```diff
@@ -400,7 +426,123 @@ pub fn walk_selector(
...
+        read_defs.insert(
+            REF_WRITE_SCHEMA_URL.to_string(),
+            to_raw_value(&write_schema).unwrap().into(),
+        );
+    }
+
+    // Add a definition for an inferred schema if it's provided.
+    // Note that we previously filtered the set of retrieved schemas to those
+    // having a read schema matching super::REF_INFERRED_SCHEMA_PATTERN.
+    if let Some(inferred_bundle) = inferred_bundle {
```
psFried, line 469: What happens if there is no inferred schema yet because this is a new collection? For example, what would happen if I published:

```yaml
collections:
  acmeCo/some/new/collection:
    writeSchema:
      properties:
        id: { type: string}
      required: [id]
    readSchema:
      allOf:
        - $ref: flow://write-schema
        - $ref: flow://inferred-schema
    key: [/id]
```

It seems like it'd fail to resolve the inferred schema ref. I'm wondering if we should use some default placeholder (`true`?) if there isn't an inferred schema yet?

crates/validation/src/lib.rs
```diff
@@ -230,3 +246,15 @@ pub async fn validate(
...
 }
+
+// This pattern lets us cheaply detect if a read schema references the inferred
+// of it's collection. Assuming an otherwise well-formed JSON schema,
+// it can neither false-positive nor false-negative:
+// * It must detect an actual property because a the same pattern within a JSON
+//   string would be quote-escaped.
+// * It must be a schema keyword ($ref cannot be, say, a property) because
+//   "flow://inferred-schema" is not a valid JSON schema and would error at build time.
+const REF_INFERRED_SCHEMA_PATTERN: &str = "\"$ref\":\"flow://inferred-schema\"";
```
psFried, line 257: This is sensitive to whitespace around the `:`, and it took me a while to prove to myself that there aren't any code paths that could result in a `"$ref" : "flow://inferred-schema"`. I _think_ this works because `Loader::load_resource_content` will always re-serialize the schema without whitespace. Assuming that's correct, I think it at least warrants a comment.

I'll suggest going just a little further and factoring this out into `fn uses_*_schema_ref(schema: &str) -> bool` instead of using `schema.contains(REF_*)`, which can carry a comment about schema serialization.

crates/validation/src/lib.rs
```diff
@@ -105,9 +82,63 @@ pub async fn validate(
...
+    };
+
+    let remote_collections = remote_collections
+        .into_iter()
+        .map(|mut spec| {
+            tracing::debug!(collection=%spec.name, "resolved referenced remote collection");
+
+            // Clear a derivation (if there is one), as we do not need it
+            // when embedding a referenced collection.
+            spec.derivation = None;
```
psFried, line 119: This should be `derive` now. I checked, and there's no longer any `live_specs` that use `derivation`, so I think we're good to remove it altogether once we get around to it.

psFried (changes requested):
Overall, this is looking pretty good to me. I left a few minor questions and comments which would be good to resolve prior to merging.

crates/validation/src/lib.rs
```diff
@@ -151,14 +116,26 @@ pub async fn validate(
...
-                collection: models::Collection::new(&spec.name),
-                scope: url::Url::parse("flow://control-plane").unwrap(),
-                spec,
-                validated: None,
+            .map(|mut spec| {
+                tracing::debug!(collection=%spec.name, "resolved referenced remote collection");
+
+                // Clear a derivation (if there is one), as we do not need it
+                // when embedding a referenced collection.
+                spec.derivation = None;
```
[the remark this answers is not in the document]
jgraettinger: This is actually the [`derivation` field of the protobuf CollectionSpec type ](https://github.com/estuary/flow/blob/master/go/protocols/flow/flow.proto#L245), not the deprecated `models::CollectionSpec::derivation`. I know, confusing!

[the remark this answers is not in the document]
jgraettinger: Thanks for checking. I'm adding a commit that kills the deprecated field once and for all.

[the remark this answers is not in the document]
psFried: ah, right 🤦 this is `built_spec`

crates/validation/src/lib.rs
psFried: somehow this comment got duplicated

crates/validation/src/collection.rs
jgraettinger: The intent is definitely to use a placeholder `{}` (equivalent to `true`). I definitely had it wired up that way, and confirmed it works as intended in end-to-end testing, but now I'm doubting myself and wondering if i screwed up a refactor. I'll double-check

crates/agent/src/publications/builds.rs
```diff
@@ -62,7 +62,7 @@ impl BuildOutput {
             .iter()
             .map(|e| Error {
                 scope: Some(e.scope.to_string()),
-                detail: e.error.to_string(),
+                detail: format!("{:#}", e.error),
```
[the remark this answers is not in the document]
jgraettinger: Added a comment too.

crates/validation/src/lib.rs
jgraettinger: You're right, it felt brittle. I've moved this into members of `models::Schema`, using whitespace-invarant regex's under the hood.

jgraettinger:
I'm concurrently doing another round of end-to-end testing, but I think this is right. PTAL

jgraettinger:
Completed another pass of E2E testing using the source-postgres-batch connector, and it worked as expected (creating all inferred columns from the data captured thus-far).

PR #1183, opened 2023-09-13
docs: add delete permission to example s3 iam policy

psFried:
Tiny docs update to bring our example IAM policy in-line with our expectations, now that we're pruning recovery logs.


jshearer (approved):
LGTM

PR #1184, opened 2023-09-13
fix: Update free trial migration number

jshearer:
**Description:**
I forgot that these need to be in order, and we already got 2 more since I started. Updated to 28.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1184)


psFried (approved):
LGTM

PR #1185, opened 2023-09-14
flowctl: atomically persist config files

psFried:
**Description:**

We've observed an infrequent but _super_ annoying bug where flowctl can occasionally truncate its own config file. This seems to be related to concurrent invocations of flowctl, where one instance will observe the truncated config file as it's being overwritten by another instance.  In some scenarios, we've also observed the config file being left in a truncated state.

This commit updates the `Config::write` function to write the config to a temp file, which is subsequently renamed to the target config file. This guarantees that another instance will either see the first version of the config file, or the new one, but never anything in between.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1185)

### DIFF
--- a/crates/flowctl/src/config.rs
+++ b/crates/flowctl/src/config.rs
@@ -52,20 +52,35 @@ impl Config {
     /// Write the config to the file corresponding to the given named `profile`.
     /// The file path is determined as documented in `load`.
     pub fn write(&self, profile: &str) -> anyhow::Result<()> {
-        let config_file = Config::file_path(profile)?;
-        if let Some(parent) = config_file.parent() {
-            std::fs::create_dir_all(parent).context("couldn't create user config directory")?;
-        }
         let ser = serde_json::to_vec_pretty(self)?;
-        std::fs::write(&config_file, &ser).context("writing config")?;
+        let dir = Config::config_dir()?;
+        std::fs::create_dir_all(&dir).context("creating config dir")?;
+
+        // It's important that we persist the config file atomically, so that
+        // concurrent invocations of flowctl don't accidentially observe the
+        // truncated file before we overwrite it. The `persist` function will
+        // use a rename in order to ensure that the operation is atomic. In
+        // order for that to work, we must ensure that `temp` and `real` are on
+        // the same filesystem, which is why we use the config directory as the
+        // temp dir.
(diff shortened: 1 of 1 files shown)


crates/flowctl/src/config.rs
```diff
@@ -52,20 +52,35 @@ impl Config {
...
+        // truncated file before we overwrite it. The `persist` function will
+        // use a rename in order to ensure that the operation is atomic. In
+        // order for that to work, we must ensure that `temp` and `real` are on
+        // the same filesystem, which is why we use the config directory as the
+        // temp dir.
+        let temp = tempfile::NamedTempFile::new_in(dir)?;
+        std::fs::write(&temp, &ser).context("writing config")?;
+
+        let real = Config::file_path(profile)?;
+        temp.persist(real).context("persisting config file")?;
```
jshearer, line 70: The `persist()` docs say

> neither the file contents nor the containing directory are synchronized, so the update may not yet have reached the disk when persist returns.

I think that's fine since the relevant part here is that it's doing an atomic replace, but are there any assumptions we're making that are based on the config file for sure being updated by the time this function returns?

jshearer (approved):
LGTM, just had one sanity-check question

crates/flowctl/src/config.rs
psFried: Yeah, I'm pretty sure that's the case, but honestly not 100%. I found [this](https://users.rust-lang.org/t/how-to-write-replace-files-atomically/42821/13) and [this](https://users.rust-lang.org/t/correct-way-to-save-a-file-atomically-but-without-interferring-with-performance/89223) both interesting threads about different aspects of this. My conclusion was that it seems "probably"™️ safe to skip the fsync as long as we're not worried about scenarios involving a system crash.

If we really need to, there's the [atomicwrites crate](https://crates.io/crates/atomicwrites), which wraps up the whole "guaranteed" approach under a simple api. But I figured it'd be best to avoid the extra dependency and two (turns out 1 isn't enough!) fsync calls if we can.

jgraettinger: Agreed, I think this implementation is juuust write

jgraettinger (approved):

PR #1190, opened 2023-09-15
Update `billing_report_202308` to round quantities.

jshearer:
Rounded quantities end up adding to the correct usage ± 1 in the end:

`daily_stats`: `0.1, 0.1, 0.1, 0.1, 0.1, 0.1, 0.1, 0.1, 0.1`
`daily_line_items`: `1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 2, 2`
`daily_deltas`: `1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 0, 0`
Trial credit is computed from daily deltas
Line items come from the last `daily_line_items` of the month, which has everything rolled into it

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1190)

### DIFF
--- a/supabase/migrations/29_round_free_trial.sql
+++ b/supabase/migrations/29_round_free_trial.sql
@@ -3,7 +3,7 @@ begin;
 -- Compute a JSONB array of line-items detailing usage under a tenant's effective tiers.
 create or replace function tier_line_items(
   -- Ammount of usage we're breaking out.
-  amount numeric,
+  amount integer,
   -- Effective tenant tiers as ordered pairs of (quantity, cents), followed
   -- by a final unpaired cents for unbounded usage beyond the final quantity.
   tiers integer[],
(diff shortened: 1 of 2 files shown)


supabase/migrations/29_round_free_trial.sql
```diff
@@ -0,0 +1,256 @@
+begin;
+
+-- Compute a JSONB array of line-items detailing usage under a tenant's effective tiers.
+create or replace function tier_line_items(
+  -- Ammount of usage we're breaking out.
+  amount numeric,
```
jgraettinger, line 6: optional: 🤔 looking at this, perhaps it would be slightly cleaner to have the `amount` parameter be an integer and then call `ceil()` when passing the argument from the `daily_line_items` CTE below? I'm good either way.

jgraettinger (approved):
LGTM

supabase/migrations/29_round_free_trial.sql
jshearer: Yep, done

PR #1191, opened 2023-09-18
Implement manual bills with defined start/end dates, and enter them into Stripe

jshearer:
* Introduce the `manual_bills` table
* Update Stripe invoice generator to understand the new `InvoiceType::Manual` type, and track the invoice type as a metadata field in order to identify previously-generated invoices and update them instead of duplicating.
* Update Stripe invoice generator to look at `manual_bills` as well as `billing_historicals`, and generate/update invoices for both
* Attach the billing period start and end to each line item in Stripe. This should come out when exported, and allow for proper income recognition

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1191)


jgraettinger (commented):
Looking good, a couple comments as you look at the discussed invoices view.

crates/billing-integrations/Cargo.toml
```diff
@@ -14,11 +14,11 @@ license.workspace = true
 anyhow = { workspace = true }
 async-stripe = { workspace = true }
 clap = { workspace = true }
+chrono = { workspace = true }
```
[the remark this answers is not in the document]
jshearer: Oh huh, interesting. I spent quite a while struggling with the `time` crate, specifically I could not for the life of me figure out how to get a unix timestamp in seconds from the various date/datetime structs they had, whereas it was basically trivial to do what I wanted with `chrono`. But if `time` is where it's at then I'm happy to look at this again 👍

supabase/migrations/30_manual_billing.sql
```diff
@@ -0,0 +1,23 @@
...
+  tenant       catalog_tenant not null references tenants(tenant),
+  usd_cents    integer        not null,
+  description  text           not null,
+  date_start   date           not null,
+  date_end     date           not null
+);
+
+alter table manual_bills enable row level security;
+
+create policy "Users must be authorized to their catalog tenant"
```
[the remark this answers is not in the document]
jshearer: Yes.

jshearer:
Alright @jgraettinger, can I get a review on the SQL part? The rust script to update Stripe invoices still needs to be updated to work with `invoices_ext`, but I want to make sure the SQL side is good before I go too far down that path

supabase/migrations/30_manual_billing.sql
```diff
@@ -0,0 +1,354 @@
+begin;
+
+create table manual_bills (
```
jgraettinger, line 3: move to schema `internal` ? Not part of the public API right?

supabase/migrations/30_manual_billing.sql
```diff
@@ -0,0 +1,354 @@
...
+$$ language plpgsql;
+
+
+-- Billing report which is effective August 2023.
+-- Removed authorization logic as it's now going to be handled in invoices_ext
+create or replace function internal.billing_report_202308(billed_prefix catalog_prefix, billed_month timestamptz)
+returns jsonb as $$
+declare
+  -- Auth checks
+  has_admin_grant boolean;
```
jgraettinger, line 78: these params should be removed?

supabase/migrations/30_manual_billing.sql
```diff
@@ -0,0 +1,354 @@
...
+-- Calculate the specified month's billing report for every tenant
+-- and save those reports to billing_historicals.
+create or replace function internal.freeze_billing_month(billed_month timestamptz)
+returns integer as $$
+declare
+    tenant_row record;
+    tenant_count integer = 0;
+begin
+    for tenant_row in select tenant as tenant_name from tenants loop
+        insert into billing_historicals
```
jgraettinger, line 267: move `billing_historicals` into `internal` schema too? and we can drop RLS etc and revoke query access for `authenticated`.

supabase/migrations/30_manual_billing.sql
```diff
@@ -0,0 +1,354 @@
...
+  union all
+  select
+    m.date_start, m.date_end, m.billed_prefix, m.line_items, m.subtotal, m.extra, 'manual' as invoice_type
+  from manual_bills m
+  union all
+  select
+    c.date_start, c.date_end, c.billed_prefix, c.line_items, c.subtotal, c.extra, 'current_month' as invoice_type
+  from current_month c
+)
+select * from combined
```
jgraettinger, line 349: i find it's general good practice not to put ordering constraints on views. Let the user of the view do that.

supabase/migrations/30_manual_billing.sql
```diff
@@ -0,0 +1,354 @@
...
+    ) as line_items,
+    to_jsonb(usd_cents) as subtotal,
+    'null'::jsonb as extra
+  from manual_bills
+  inner join authorized_tenants on manual_bills.tenant ^@ authorized_tenants.tenant
+),
+current_month as (
+  select
+    date_trunc('month', (report->>'billed_month')::timestamptz) as date_start,
+    date_trunc('month', (report->>'billed_month')::timestamptz) + interval '1 month' - interval '1 day' as date_end,
```
jgraettinger, line 329: this is pedantic, but technically these `date-time`'s don't cover the whole month -- the last date is missing. This stands out to me in the test fixtures below.

It's a little unclear if these response date-end's are inclusive or exclusive. That's true of `manual_bills` as well -- is date_end included in the billing period, or not?

Particularly as these are rendered as full date-times in the API response, my knee-jerk inclination is that they should be an [inclusive, exclusive) range like `2023-07-01T00:00:00Z` to `2023-08-01T00:00:00Z`.

jgraettinger (commented):
A few comments, looking good.

supabase/migrations/30_manual_billing.sql
jshearer: This is a good thing to be specific about. My intent was for all dates to only end up being represented as `YYYY-mm-dd` with every date being inclusive. I used the `date` type for `manual_bills.date_start|date_end`, but incorrectly sent back `timestamptz`-formatted values. I _think_ the right thing to do here is change `invoices_ext` to return `date_start`/`date_end` as `YYYY-mm-dd`. That way when we send `"date_start": "2023-08-01", "date_end": "2023-08-31"` it'll be clear that that should cover the entire month. Thoughts on this?

supabase/migrations/30_manual_billing.sql
```diff
@@ -0,0 +1,352 @@
...
+-- Drop the public functions
+drop function billing_report_202308(catalog_prefix, timestamptz);
+drop function tier_line_items(integer, integer[], text, text);
+
+-- Move `billing_historicals` to internal and revoke access
+alter table billing_historicals disable row level security;
+drop policy "Users must be authorized to their catalog tenant" on billing_historicals;
+
+alter table billing_historicals set schema internal;
+
```
jgraettinger, line 27: do you need a `revoke select on billing_historicals from authenticated`?

supabase/migrations/30_manual_billing.sql
```diff
@@ -0,0 +1,352 @@
...
+),
+current_month as (
+  select
+    date_trunc('month', (report->>'billed_month')::timestamptz)::date as date_start,
+    (date_trunc('month', (report->>'billed_month')::timestamptz) + interval '1 month' - interval '1 day')::date as date_end,
+    report->>'billed_prefix' as billed_prefix,
+    report->'line_items' as line_items,
+    report->'subtotal' as subtotal,
+    report as extra
+  from authorized_tenants, internal.billing_report_202308(authorized_tenants.tenant, now()) as report
```
jgraettinger, line 336: Right, this does still have the discussed problem of invoices "disappearing" when the month rolls over but we haven't yet frozen the prior month's bills.

To do this comprehensively we'd need generate the series of months between the largest `internal.billing_historicals` month and `now()`, and generate reports for each of them.

This also makes it ambugious for the caller whether an invoice is finalized or not. Should we add a boolean to distinguish?

jgraettinger (approved):
LGTM % comments but this does have a magic disappearing bill problem.

I'm okay with merging and fixing it later but it will definitely come up again

supabase/migrations/30_manual_billing.sql
jshearer: Do I? I thought that `authenticated` doesn't have access to the `internal` schema at all, and that moving to `internal` would implicitly remove that grant (or make it impossible to address because `internal` "doesn't exist" if you aren't granted to it). But it doesn't hurt to be explicit, I'll add it.

supabase/migrations/30_manual_billing.sql
jshearer: Riiiiiiiiight, I forgot about that. So yeah, I think the right thing to do is what you say: fill in the missing months by generating the invoice. The view already have an `invoice_type` column, so I'm thinking that the right way to identify these invoices is to set their `invoice_type` to something like `preview` or `pending` or something like that.

jshearer: We'll also need to set a cutoff month for how far back we generate these. Probably something like the greater of August 2023 and the tenant's creation month

crates/billing-integrations/src/stripe.rs
```diff
@@ -41,32 +41,23 @@ pub struct PublishInvoice {
...
 
-#[derive(Serialize, Deserialize, Debug)]
-struct LineItem {
-    count: f64,
-    subtotal: i64,
-    description: Option<String>,
+#[derive(
+    Debug, Clone, PartialEq, Eq, Hash, sqlx::Type, Deserialize_enum_str, Serialize_enum_str,
+)]
+#[sqlx(rename_all = "snake_case")]
```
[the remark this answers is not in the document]
jshearer: WRT `serde_enum_str`: the string values for `invoice_type` live in quite a few different places:
* They come from the `invoices_ext` view
* They need to be parsed by sqlx into the right `InvoiceType` enum variant
* Sqlx uses serde, so Serde also needs to know the correct strings to parse/serialize for each variant
* They get set in the Stripe invoice metadata to allow looking up existing invoices instead of always creating new ones
* The front-end also uses them to do some different logic depending on which type of invoice it's rendering

I originally had a manually implemented `to_string` impl, but it felt smelly to have yet another place to manually keep these strings in sync. [Down below I use](https://github.com/estuary/flow/pull/1191/files#diff-4fe0b4ecd5665aaa642a6ced5a704e070710504713aba04f1502804c8aecab0fR174) the `Serialize_enum_str` `to_string()` impl to get me the correct string representation of the invoice type. 

For the same reason, I would like to keep the explicit renames. Reason being, when someone inevitably wants to go look and find all references to those enums, they'll be able to find them easily throughout all the repos, instead of having them be implicit by way of `rename_all="snake_case"`. I feel less strongly about this though, so I'm fine with doing the rename_all if you feel strongly that it should be that way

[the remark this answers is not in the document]
jgraettinger: > Sqlx uses serde, so Serde also needs to know the correct strings to parse/serialize for each variant

This is why I made the comment and the source of my confusion: as far as I am aware, if you add a plain Serialize to an enum with empty variants (not structs or tuples), the serde representation _is_ the simple string. For example, here's [CompressionCodec](https://github.com/estuary/flow/blob/master/crates/models/src/journals.rs#L220-L226), which serde's as snake-case variant names ([example here](https://github.com/estuary/flow/blob/master/crates/sources/src/scenarios/snapshots/sources__scenarios__test__collections.snap#L73)).

I'm pretty sure you could just use `serde_json::to_value(my_enum).as_str().unwrap()` if you want the string itself.

> For the same reason, I would like to keep the explicit renames. 

:+1:

supabase/migrations/30_manual_billing.sql
jgraettinger: that's why i asked: I don't know what happens. I know it's currently granted and i don't know why moving it to a different schema would necessarily remove the grant.

crates/billing-integrations/src/stripe.rs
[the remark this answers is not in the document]
jshearer: I'm going to test this right now, but I assumed that would give you the string `"enum_variant"`, with the quotes, since a bare `enum_variant` token isn't valid JSON. But let me confirm real quick..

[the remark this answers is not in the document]
jgraettinger: Yep it will, but serde_json::Value::as_str() would unwrap that. Or `let Value::String(my_enum) = serde_json::to_value(&my_enum).unwrap() else { unreachable!() };`.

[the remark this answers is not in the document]
jshearer: Ahhhhhh, okay

```
>> serde_json::to_string(&InvoiceType::Usage).unwrap().as_str()
"\"usage\""
>> serde_json::to_value(InvoiceType::Usage).unwrap().as_str().unwrap()
"usage"
```

So `serde_json::Value::as_str()` gives the bare string. Neat!

supabase/migrations/30_manual_billing.sql
jgraettinger: can we consolidate `current_month` and `generated_prior_months` into a single CTE? Would it then makes sense to have InvoiceTypes of "final", "preview", and "manual"? Or is there something special about current-month that makes it different from a prior month that hasn't been finalized?

jshearer: Nope I think that grouping makes sense: the UI needs to know whether the invoice is "real" or not in order to know whether to show stripe invoice buttons, which this `final`/`preview` distinction covers. I'll make this change 👍

supabase/migrations/30_manual_billing.sql
jshearer: Turns out, the answer is "it would have worked correctly"

```sql
create table test(col integer primary key);
grant select on test to authenticated;
alter table test set schema internal;
set role authenticated;
select * from internal.test;
```
=>
```
ERROR:  permission denied for schema internal
LINE 1: select * from internal.test;
```

jgraettinger (approved):
still LGTM

PR #1198, opened 2023-09-25
control-plane: fix deletion of old drafts

psFried:
This commit fixes the migration for deleting old drafts, so that it can be applied.  It was broken, and was thankfully never applied in production.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1198)


supabase/migrations/26_delete_old_rows.sql
```diff
@@ -0,0 +1,41 @@
...
+    delete from public.drafts where updated_at < (now() - '30 days'::interval) returning *
+  )
+  select into n_drafts count(*) as n from d;
+
+  with l as (
+    delete from internal.log_lines where logged_at < (now() - '30 days'::interval) returning *
+  )
+  select into n_logs count(*) as n from l;
+
+  return json_build_object(
```
jgraettinger, line 27: for reference, here's the maintenance routine I've been running manually ([slack thread where it's discussed](https://estuaryworkspace.slack.com/archives/C03QBN83GQ4/p1695408708786299?thread_ts=1695401611.797799&cid=C03QBN83GQ4)):

```
delete from internal.log_lines where logged_at < (now() - '2 days'::interval);
vacuum analyze internal.log_lines;

delete from catalog_stats_hourly where grain = 'hourly' and ts < (now() - '30 days'::interval);
vacuum analyze catalog_stats_hourly;

delete from drafts where updated_at < (now() - '10 days'::interval);
vacuum analyze drafts;
vacuum analyze draft_specs;
vacuum analyze discovers;
```

A few comments:

* I think we need much tighter bounds over `log_lines` retention. We pile up a lot of these every day and just in the couple days since I ran manual maintenance, we're back up to 4GB of them (I just now deleted to 2 days now and released 900MB of table size). IMO 2 days is reasonable.
* We should also clean `catalog_stats_hourly` because that's our second-largest table behind logs.
* _Probably_ these manual `vacuum analyze` statements aren't required, given the auto vacuum daemon, but i'm unclear if `catalog_stats_hourly` is covered by it ([docs note](https://www.postgresql.org/docs/current/routine-vacuuming.html#AUTOVACUUM) that partition tables aren't covered by auto vacuum, but i'm unclear if that's exempting `catalog_stats` -- the partitioned table -- but not `catalog_stats_hourly` which is not itself partitioned). So ... also `vacuum analyze catalog_stats_hourly` ?

psFried: I ran this query:

```
select *
from pg_stat_all_tables
where relname in ('log_lines', 'drafts', 'catalog_stats_hourly', 'draft_specs', 'discovers');
```

and is shows that all those tables are getting auto-vacuumed regularly. So I just adjusted the retention windows to match what you pasted, and added in pruning of `catalog_stats_hourly`.

jgraettinger (approved):
LGTM thanks!

PR #1199, opened 2023-09-26
docs/source-postgres: Set `publish_via_partition_root` in setup

willdonnelly:
**Description:**

Edits all examples of `CREATE PUBLICATION` queries to be followed by an `ALTER PUBLICATION p SET (publish_via_partition_root = true)` query.

This is done in a separate `ALTER PUBLICATION` query because it will fail on versions of PostgreSQL <13 which predate the setting, and if the user blindly executes the commands we want the create portion to succeed anyway.

Also adds some text explaining that the flag is recommended but not required, with a brief explanation of why it's recommended / what it does.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1199)


jgraettinger (approved):
LGTM

PR #1200, opened 2023-09-26
fix: Differentiate between an object that has a literal `*` field, and a pointer to an object's `additionalProperties` shape.

jshearer:
This flake was caught by quickcheck (only sometimes) and was causing Flow CI to intermittently fail.

The one thing I'm not confident on is calling this variant `AdditionalProperties`, since the variant for array's `additionalItems` is called `Token::NextIndex`, not `Token::AdditionalItems`. Thoughts @jgraettinger?

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1200)

### DIFF
--- a/crates/doc/src/ptr.rs
+++ b/crates/doc/src/ptr.rs
@@ -215,8 +215,7 @@ impl Pointer {
                         arr.last_mut().unwrap()
                     }
                     // Cannot match (attempt to query property of an array).
-                    Token::Property(_) => return None,
-                    Token::NextProperty => return None,
+                    Token::Property(_) | Token::NextProperty => return None,
                 },
                 Value::Number(_) | Value::Bool(_) | Value::String(_) => {
                     return None; // Cannot match (attempt to take child of scalar).
@@ -275,8 +274,7 @@ impl Pointer {
                         arr.last_mut().unwrap()
                     }
                     // Cannot match (attempt to query property of an array).
-                    Token::Property(_) => return None,
-                    Token::NextProperty => return None,
+                    Token::Property(_) | Token::NextProperty => return None,
                 },
                 HeapNode::Bool(_)
                 | HeapNode::Bytes(_)
@@ -517,6 +515,7 @@ mod test {
         for case in [
             "/foo/2/a/3", // Attempt to index string scalar.
             "/foo/bar",   // Attempt to take property of array.
+            "/foo/-",     // Attempt to take property of array
         ]
         .iter()
         {
--- a/crates/doc/src/ptr.rs
+++ b/crates/doc/src/ptr.rs
@@ -9,8 +9,8 @@ pub enum Token {
     Index(usize),
     /// JSON object property name without escaping. Never an integer.
     Property(String),
-    // Schema representing an object's additionalProperties
-    AdditionalProperties,
+    // Represents the concept of "the next property" to be added
+    NextProperty,
     /// Next JSON index which is one beyond the current array extent.
     /// If applied to a JSON object, the property literal "-" is used.
     NextIndex,
@@ -18,9 +18,7 @@ pub enum Token {
 
 impl Token {
     pub fn from_str(s: &str) -> Self {
-        if s == "-" {
-            Token::NextIndex
-        } else if s.starts_with('+') || (s.starts_with('0') && s.len() > 1) {
+        if s.starts_with('+') || (s.starts_with('0') && s.len() > 1) {
             Token::Property(s.to_string())
         } else if let Ok(ind) = usize::from_str(&s) {
             Token::Index(ind)
@@ -35,7 +33,7 @@ impl<'t> std::fmt::Display for Token {
         match self {
             Token::Index(ind) => write!(f, "{ind}"),
             Token::Property(prop) => write!(f, "{prop}"),
-            Token::AdditionalProperties => write!(f, "*"),
+            Token::NextProperty => write!(f, "*"),
             Token::NextIndex => write!(f, "-"),
         }
     }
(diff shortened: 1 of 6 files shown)


crates/doc/src/ptr.rs
```diff
@@ -9,6 +9,8 @@ pub enum Token {
     Index(usize),
     /// JSON object property name without escaping. Never an integer.
     Property(String),
+    // Schema representing an object's additionalProperties
```
jgraettinger, line 12: I think this should be `NextProperty`, paralleling `NextIndex`. That's because it's a conceptual placeholder for the next property to be added.

It might happen to sometimes be schematized through a `additionalProperties` JSON schema keyword, but does not have any literal association. This is a JSON Pointer, which has zero dependence on the JSON Schema spec.

crates/doc/src/ptr.rs
```diff
@@ -140,12 +146,13 @@ impl Pointer {
                 Node::Object(fields) => match token {
                     Token::Index(ind) => fields.get(&ind.to_string()),
                     Token::Property(property) => fields.get(&property),
+                    Token::AdditionalProperties => None,
```
jgraettinger, line 149: 🤔 it looks like `Token::from_str` does parse `-` into `NextIndex`. I'm thinking it shouldn't, which matches how `NextProperty` is being implemented. Then it should return `None` here as well, because you can never really query `NextIndex` or `NextProperty`.

crates/doc/src/ptr.rs
```diff
@@ -196,6 +203,7 @@ impl Pointer {
                     Token::Index(ind) => map.entry(ind.to_string()).or_insert(Value::Null),
                     Token::Property(prop) => map.entry(prop).or_insert(Value::Null),
                     Token::NextIndex => map.entry("-").or_insert(Value::Null),
```
jgraettinger, line 205: both `NextIndex` and `NextProperty` should return None here. They can't be created as properties.

(Technically `NextIndex` currently needs to create a "-" property, but won't once we no longer parse "-" into NextIndex).

crates/doc/src/ptr.rs
```diff
@@ -236,7 +245,7 @@ impl Pointer {
             // which we'll create the next child location.
             if let HeapNode::Null = v {
                 match token {
-                    Token::Property(_) => {
+                    Token::Property(_) | Token::AdditionalProperties => {
```
jgraettinger, line 248: i think the `NextIndex` and `NextProperty` cases should just return `None` here -- these locations cannot be created.

crates/doc/src/ptr.rs
```diff
@@ -251,6 +260,7 @@ impl Pointer {
                     Token::Index(ind) => fields.insert_property(&ind.to_string(), alloc),
                     Token::Property(property) => fields.insert_property(property, alloc),
                     Token::NextIndex => fields.insert_property("-", alloc),
```
jgraettinger, line 262: here too: return None. The current `insert_property` handling is only required because we currently parse "-" into NextIndex.

crates/json/src/lib.rs
```diff
@@ -87,6 +88,11 @@ impl<'a> Location<'a> {
         Location::EndOfArray(self)
     }
 
+    // Returns a new Location that extends this one with a pointer to the object's additionalProperties ("*").
+    pub fn push_additional_properties(&'a self) -> Location<'a> {
```
jgraettinger, line 92: nit: push_next_property

jgraettinger (commented):
Looks good, some detail comments on specific case handling, and i think we should use `NextProperty` for the variant.

jshearer:
Ok @jgraettinger I think this is ready to be looked at again. Changing `Token::from_str` and adjacent logic to no longer parse `-` broke a couple tests which I updated to match the new logic, but anything else that depended on that functionality will be broken.. I _think_ the tests should have caught most of those? I also did run a quick e2e test of starting up a local stack, publishing ops catalogs, running a hello-world capture, etc. which all appeared to work

crates/doc/src/ptr.rs
```diff
@@ -213,6 +216,7 @@ impl Pointer {
                     }
                     // Cannot match (attempt to query property of an array).
                     Token::Property(_) => return None,
```
jgraettinger, line 218: nit: collapse into a single line as is done elsewhere

crates/doc/src/ptr.rs
```diff
@@ -271,6 +276,7 @@ impl Pointer {
...
```
jgraettinger, line 278: nit; collapse into a single line

crates/doc/src/ptr.rs
```diff
@@ -502,7 +506,7 @@ mod test {
         }
```
jgraettinger, line 507: add back the `/foo/-` case here by manually building up the tokens? You can't implicitly build them by parsing a pointer as it used to do, but you can still build it manually.

Also add a case covering creation with NextProperty. It'd return None, right?

crates/doc/src/shape/limits.rs
```diff
@@ -28,7 +28,7 @@ fn squash_location_inner(shape: &mut Shape, name: &Token) {
         // Squashing of `additional*` fields is not possible here because we don't
         // have access to the parent shape
         Token::NextIndex => unreachable!(),
-        Token::Property(prop) if prop == "*" => unreachable!(),
+        Token::NextProperty => unreachable!(),
```
jgraettinger, line 31: nit: collapse cases. i'm harping on this only for consistency: why are we doing it in some places and not others?

jgraettinger (approved):
LGTM % some minor nits and adding a bit of removed test coverage back

crates/doc/src/ptr.rs
jshearer: Got totally side-tracked and forgot to finished this. What do you mean by adding `/foo/-` back? That pointer doesn't make any sense. `/foo` is an array, and `-` is interpreted as an object field called `-`, which you can't use to index an array

jgraettinger: I meant build it up from tokens, rather than parsing, so that it's the NextIndex variant.

It's possible to build such a pointer and pass it to create, right? Let's cover that in the test.

jgraettinger:
still lgtm!

PR #1204, opened 2023-09-27
control-plane: rename migration to fix conflict

psFried:
There were 2 migrations with number 30, so I just renamed one to 31

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1204)


jshearer (approved):
LGTM

PR #1205, opened 2023-09-27
fix: Dramatically improve `invoices_ext` perf when you have many tenant grants

jshearer:
**Description:**

The fix is to change from using CTEs to subqueries. CTEs, as it turns out, do not support predicate push-down, so the query planner was calculating the entire `invoices_ext` table, _then_ applying filtering. This was very slow when you have admin grants to many tenants, such as users with the `estuary_support/` role.

The other change I made was to select `authorized_tenants.tenant as billed_prefix` instead of `report->>'billed_prefix'`. This allows for a filter on `invoices_ext.billed_prefix` to get pushed all the way to the `authorized_tenants` CTE (not into the CTE, but its output is good enough).

This change is already live, since the billing page was broken for `estuary_support/` accounts otherwise.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1205)


jgraettinger (approved):
LGTM

PR #1206, opened 2023-09-27
fix: stripe invoice generator using the wrong `invoice_type` values

jshearer:
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1206)


jgraettinger (approved):
LGTM

PR #1211, opened 2023-09-28
Disable use of `realtime`

travjenkins:
**Description:**

Since we now using polling for discovers we no longer rely on the Supabase realtime capabilities. I have been keeping an eye on the project but I have a feeling they will [never guarantee messages](https://github.com/supabase/realtime#does-this-server-guarantee-message-delivery). This means all solutions we think about in the UI are based around using polling and I believe we can disable this functionality as we are not using it.

**Workflow steps:**

N/A

N/A

**Notes for reviewers:**

We can manually disable these settings in Supabase dashboard to cause these changes right now. This will keep that setting from getting flipped back on later.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1211)

### DIFF
--- a/supabase/migrations/06_connectors.sql
+++ b/supabase/migrations/06_connectors.sql
@@ -71,7 +71,6 @@ create table connector_tags (
     check (image_tag like ':%' or image_tag like '@sha256:')
 );
 -- Public, no RLS.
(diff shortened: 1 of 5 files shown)


jgraettinger (approved):
LGTM

PR #1215, opened 2023-09-29
Implement logic to move tenants into the free trial

jshearer:
**Description:**

Rather than writing `internal.set_new_free_trials()` with a CTE, I thought it would be nice to be able to take a look and see which tenants are slated to be kicked into the free trial next time the cron job runs, so I made `internal.new_free_trial_tenants` a view. Any reason this would be a bad idea?

The criteria I came up with from coversations is: a tenant should be moved into the trial if
* They are not already a paying customer (i.e they don't have a free trial set) AND
* They have at least one active, not-disabled capture or materialization
* One of the following is true
  * Daily task usage time exceeds 2*24h
  * Monthly usage time exceeds 2*30d
  * Monthly data used exceeds 10gb

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1215)

### DIFF
--- a/crates/billing-integrations/src/stripe.rs
+++ b/crates/billing-integrations/src/stripe.rs
@@ -142,6 +142,13 @@ impl Invoice {
             }
         }
 
+        // The minimum chargable amount of USD in Stripe is $0.50.
+        // https://stripe.com/docs/currencies#minimum-and-maximum-charge-amounts
+        if self.subtotal < 50 {
+            tracing::info!("Skipping invoice for less than the minimum chargable amount ($0.50)");
+            return Ok(());
+        }
+
         // Anything before 12:00:00 renders as the previous day in Stripe
         let date_start_secs = self
             .date_start
@@ -219,7 +226,7 @@ impl Invoice {
                     Some(invoice)
                 }
                 Some(stripe::InvoiceStatus::Open) => {
-                    bail!("Found finalized invoice {id}. Pass --recreate-finalized to delete and recreate this invoice.", id = invoice.id.to_string())
+                    bail!("Found open invoice {id}. Pass --recreate-finalized to delete and recreate this invoice.", id = invoice.id.to_string())
                 }
                 Some(status) => {
                     bail!(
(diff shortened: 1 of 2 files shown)


crates/billing-integrations/src/stripe.rs
```diff
@@ -243,47 +256,42 @@ impl Invoice {
...
-                        value: date_start_human.to_owned(),
+                    client,
+                    stripe::CreateInvoice {
+                        customer: Some(customer.id.to_owned()),
+                        // Stripe timestamps are measured in _seconds_ since epoch
+                        // Due date must be in the future. Bill net-30, so 30 days from today
+                        due_date: Some((Utc::now() + Duration::days(30)).timestamp()),
+                        description: Some(
+                            format!(
+                                "Your Flow bill for the billing preiod between {date_start_human} - {date_end_human}. Tenant: {tenant}",
```
jgraettinger, line 267: nit: preiod

supabase/migrations/33_automate_trial_start.sql
```diff
@@ -0,0 +1,88 @@
+begin;
+
+create or replace view internal.new_free_trial_tenants as
+with hrs_by_day as (
```
jgraettinger, line 4: nit: `hrs` => `hours`. It's a needless abbreviation that saves two characters and can confuse a reader. Prefer to spell things out. [See also](https://google.github.io/styleguide/cppguide.html#General_Naming_Rules)

supabase/migrations/33_automate_trial_start.sql
```diff
@@ -0,0 +1,88 @@
...
+  select
+    tenants.tenant as tenant,
+    ts,
+    sum(catalog_stats_daily.usage_seconds / (60.0 * 60)) as daily_usage_hours
+  from catalog_stats_daily
+  join tenants on tenants.tenant = split_part(catalog_stats_daily.catalog_name,'/',1)||'/'
```
jgraettinger, line 10: always always use the `^@` operator for prefix comparisons. Compare to how we're doing it elsewhere for authorization checks, etc.

The reason why is because this usage of split_part forces it to walk the table actually doing string splits for every row. The optimizer almost certainly can't "see into" what this is doing. Whereas postgres is smart enough to push down `^@` into an evaluation against the ordered table index.

supabase/migrations/33_automate_trial_start.sql
```diff
@@ -0,0 +1,88 @@
...
+    count(distinct live_specs.id) filter (where live_specs.spec_type = 'materialization') as today_materializations
+from tenants
+left join hrs_by_day on hrs_by_day.tenant = tenants.tenant
+left join hrs_by_month on hrs_by_month.tenant = tenants.tenant
+left join gbs_by_month on gbs_by_month.tenant = tenants.tenant
+join live_specs on (split_part(live_specs.catalog_name,'/',1)||'/' = tenants.tenant and (live_specs.spec #>> '{shards,disable}')::boolean is not true)
+where tenants.trial_start is null
+group by tenants.tenant
+having (
+  (
```
jgraettinger, line 52: why is this repeated? don't the individual views already have these conditions?

supabase/migrations/33_automate_trial_start.sql
```diff
@@ -0,0 +1,88 @@
...
+    max(hrs_by_day.daily_usage_hours) > (2 * 24) or
+    max(hrs_by_month.monthly_usage_hours) > (24 * 31 * 2) or
+    max(gbs_by_month.monthly_usage_gbs) > 10
+  ) and (
+    count(distinct live_specs.id) filter (where live_specs.spec_type = 'capture') > 0 or
```
jgraettinger, line 57: 🤔 why would we _not_ want to transition a tenant, just because they don't *currently* have tasks?

They clearly did have tasks which exceeded the threshold, right?

jshearer: This was a request by @dyaffe, but it turns out he just wanted it because it's pointless from a business perspective since we're unlikely to successfully bill old/inactive tenants. We probably shouldn't leak that sort of business logic into here though, and should independently figure out the right way to disable inactive/past due tenants

supabase/migrations/33_automate_trial_start.sql
jshearer: This exact check is a bit unclear and leftover from before I pushed the filters into the CTEs, but we do need a filter here to make sure that the tenant actually _has_ at least one non-null criterion pushing them over the free tier. Changed it to `having count(hours_by_month) > 0 or count(hours_by_day) > 0 or count(gbs_by_month) > 0`

crates/billing-integrations/src/stripe.rs
jshearer: Period at the end after the tenant name? Or you don't want the period after the billing period and tenant?

jgraettinger: Lol I was point out that period is misspelled.

supabase/migrations/33_automate_trial_start.sql
jshearer: So yeah I must be doing something wrong then. I originally used `^@`, but switched to this much less legible `split_part` filtering because it's _dramatically_ more performant. Just switching the 3 join conditions in the CTEs to `join tenants on catalog_stats.catalog_name ^@ tenants.tenant` instead of `split_parts` causes the query to take 117s, instead of 500ms. Here's just one of the CTEs split out:

```sql
 select
    tenants.tenant as tenant,
    ts,
    sum(catalog_stats_daily.usage_seconds / (60.0 * 60)) as daily_usage_hours
  from catalog_stats_daily
  join tenants on catalog_stats_daily.catalog_name ^@ tenants.tenant
  -- join tenants on tenants.tenant = split_part(catalog_stats_daily.catalog_name,'/',1)||'/'
  where tenants.trial_start is null
  group by tenants.tenant, ts
  having sum(catalog_stats_daily.usage_seconds / (60.0 * 60)) > (2 * 24)
```
* With `split_part` it runs in 312ms. `EXPLAIN` output: https://explain.dalibo.com/plan/4051a76d37ag4b2h
* With `^@`, it runs in **1m35s**. `EXPLAIN` output: https://explain.dalibo.com/plan/he33e57c8a07702f

crates/billing-integrations/src/stripe.rs
jshearer: d'oh! 🤦

supabase/migrations/33_automate_trial_start.sql
jshearer: Thx for your help (on slack, not being sassy here lolol). Turns out postgres is finicky and won't (or can't) use the regular btree index that we had on `catalog_stats.catalog_name`. The following index fixed it, and is included in this migration (though I already applied it to the DB):

```sql
create index catalog_stats_catalog_index_spgist on catalog_stats using spgist ((catalog_name::text));
```

jgraettinger (approved):
LGTM!

PR #1216, opened 2023-09-29
docs/mysql-batch: Stub so the UI won't look too awful

willdonnelly:
**Description:**

Adds a stub documentation page for the new batch MySQL connector so that the UI sidebar won't be empty.

This (and the corresponding Postgres page) really ought to be fleshed out more once I get back next week, but this should suffice for now.

This creates https://docs.estuary.dev/reference/Connectors/capture-connectors/mysql-batch/

I've already set up https://go.estuary.dev/source-mysql-batch pointing there as well.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1216)


williamhbaker (approved):
LGTM

PR #1229, opened 2023-10-10
Adding in what tenant was being fetched with the payment

travjenkins:
**Description:**

To make life easy on the UI we are not passing back what tenant was used when getting the payment methods.

**Workflow steps:**
N/A

N/A

**Notes for reviewers:**
N/A

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1229)


jshearer (approved):
We talked about this on a VC -- ideally we shouldn't need to pass back a parameter that we pass in, but this feels innocuous enough and it makes the frontend's job easier (and is currently causing a pretty alarming error message in the UI), so I say LGTM 👍

PR #1231, opened 2023-10-10
sources: remove newlines from inferredSchemaIsNotAvailable

psFried:
**Description:**

The sentinel schema that's used when the inferred schema is not available contained newline characters, which make their way into the connector protocols. This breaks some connectors, which expect each message to be on its own line, since the schema value in, for example, a `Validate` request could contain newlines. This changes the handling of that sentinel schema to use plain `serde_json::Value`s instead of `RawValue`s, so that re-encoding it as JSON will cause the newlines to be removed.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1231)


jgraettinger (approved):
LGTM

PR #1232, opened 2023-10-11
Add `pays_externally` flag to tenants

jshearer:
**Description:**

This allows us to flag tenants that pay outside of Stripe (or whatever payment system we might use in the future). We shouldn't show these tenants billing alerts in the UI because it's expected that they won't have billing information entered.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1232)


jgraettinger (approved):
LGTM

PR #1233, opened 2023-10-11
Adding the CORS headers to the billing functions

travjenkins:
**Description:**
Adding the CORS headers to the `billing` functions.

**Workflow steps:**
N/A

N/A

**Notes for reviewers:**

This is just to be super safe. Everything is working today but we should probably be including these. This especially helps locally when there can be some wonky CORS issues.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1233)

### DIFF
--- a/supabase/functions/billing/delete_tenant_payment_method.ts
+++ b/supabase/functions/billing/delete_tenant_payment_method.ts
@@ -1,4 +1,4 @@
-import { customerQuery, StripeClient } from "./shared.ts";
+import { billingResponseHeaders, customerQuery, StripeClient } from "./shared.ts";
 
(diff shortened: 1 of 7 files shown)


jshearer (approved):
These changes all LGTM. One question (that I couldn't create inline for some reason) is whether we need to reply with the correct CORS headers in the `OPTIONS` response as well: https://github.com/estuary/flow/pull/1233/files#diff-8f0dc4f8a45981d5117d609e501af8cc97cdbe842a0418dda9b535326f892b08L17

As it is right now, it appears that Supabase is giving us some CORS headers already, so maybe we don't need to do anything in that case:

Screen Shot 2023-10-16 at 10 47 48

PR #1240, opened 2023-10-13
Add `flowctl raw spec` to invoke the Spec RPC

jshearer:
**Description:**

This allows for calling both local and container connectors' spec RPC manually, and inspecting the output. It's also useful for manually invoking the RPC to enable debugging connectors.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1240)

### DIFF
--- a/crates/flowctl/src/raw/spec.rs
+++ b/crates/flowctl/src/raw/spec.rs
@@ -1,0 +1,94 @@
+use crate::local_specs;
+use anyhow::Context;
+use proto_flow::{capture, flow};
+
+#[derive(Debug, clap::Args)]
+#[clap(rename_all = "kebab-case")]
+pub struct Spec {
+    /// Path or URL to a Flow specification file.
+    #[clap(long)]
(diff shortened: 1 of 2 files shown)


crates/flowctl/src/raw/spec.rs
```diff
@@ -0,0 +1,94 @@
...
+        source,
+        capture,
+        network,
+    }: &Spec,
+) -> anyhow::Result<()> {
+    let source = build::arg_source_to_url(source, false)?;
+    let mut sources = local_specs::surface_errors(local_specs::load(&source).await.into_result())?;
+
+    // Identify the capture to inspect.
+    let needle = if let Some(needle) = capture {
```
jgraettinger, line 32: for later: this should be expanded to cover derivations and materializations.

There's also probably opportunity to combine with the "flowctl generate" command.

jgraettinger (approved):
LGTM

PR #1247, opened 2023-10-19
agent: fix panic on collection deletion

psFried:
**Description:**

When deleting a collection, the `add_inferred_schema_md5` function is still called. This panics because that function included a condition on `spec_type = 'collection'`, which is no longer true for collections being deleted. This results in panic because we explicitly expect one row to always be updated. This removes the condition, so that the live spec is updated by id only.

**Workflow steps:**

This bug can be reproduced by deleting any collection spec.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1247)

### DIFF
--- a/crates/agent-sql/src/publications.rs
+++ b/crates/agent-sql/src/publications.rs
@@ -160,7 +160,7 @@ pub async fn add_inferred_schema_md5(
     sqlx::query!(
         r#"
         update live_specs set inferred_schema_md5 = $1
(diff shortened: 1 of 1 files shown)


travjenkins (approved):
lgtm

psFried:
Merging because I'm confident that this definitely doesn't make anything _worse_, though I'm increasingly skeptical that this was the root of the issue.

PR #1248, opened 2023-10-19
create notification system foundation

kiahna-tucker:
**Description:**

Create the foundation for a notification system. The inaugural alert that will be supported by the system has the following requirements:

* Provide a means for a user to subscribe to any alert defined under a specific prefix (of which they admin). The user will only be presented with the option to receive alerts via email at this time, but Slack channels are a notification delivery method that will need to be supported in the future.
  
* Provide a means for the user to select a period of time over which a task's data processing should be evaluated from a predefined set of options. Accepted time periods are: two hours, four hours, eight hours, 12 hours, 24 hours, and two days.

* Send one alert email when the condition for the aforementioned alert is valid.

* Send a confirmation email when the condition for the aforementioned alert is no longer valid.

* Allow a user to enter an email that is not associated with an account in the system to receive alerts.

Additional context for this PR can be found in this [discussion](https://github.com/estuary/flow/discussions/1255). It should be noted that there are two, behavioral quirks of the implementation that are considered acceptable:

* Disabling or deleting a task for which an alert is firing (i.e., there is a row in the `alert_history` table for the task where `resolved_at` is `null`) will result in a confirmation email sent.

* Unsetting the `evaluation_interval` of a task for which an alert is firing (i.e., there is a row in the `alert_history` table for the task where `resolved_at` is `null` but the row in the `alert_data_processing` table for the task has been deleted) will result in a confirmation email sent.

**Workflow steps:**

N/A

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1248)


supabase/migrations/37_alerts.sql
```diff
@@ -0,0 +1,136 @@
...
+
+create policy "Users access alert history for admin-authorized tasks"
+  on alert_history as permissive
+  using (exists(
+    select 1 from auth_roles('admin') r where catalog_name ^@ r.role_prefix
+  ));
+
+grant select on alert_history to authenticated;
+
+create view internal.alert_data_processing_firing as
```
psFried, line 51: I think this view could use some test cases. See [this example](https://github.com/estuary/flow/blob/master/supabase/tests/live_specs_ext.test.sql) for how to write tests using pgTap, and LMK if you have any questions.

supabase/migrations/37_alerts.sql
```diff
@@ -0,0 +1,136 @@
...
+  'data_not_processed_in_interval' as alert_type,
+  alert_subscriptions.email,
+  live_specs.spec_type,
+  coalesce(sum(catalog_stats_hourly.bytes_written_by_me + catalog_stats_hourly.bytes_written_to_me + catalog_stats_hourly.bytes_read_by_me), 0)::bigint as bytes_processed
+from alert_data_processing
+  left join live_specs on alert_data_processing.catalog_name = live_specs.catalog_name and live_specs.spec is not null and (live_specs.spec->'shards'->>'disable')::boolean is not true
+  left join catalog_stats_hourly on alert_data_processing.catalog_name = catalog_stats_hourly.catalog_name
+  left join alert_subscriptions on alert_data_processing.catalog_name ^@ alert_subscriptions.catalog_prefix
+where live_specs.created_at <= date_trunc('hour', now() - alert_data_processing.evaluation_interval)
+      and catalog_stats_hourly.ts >= date_trunc('hour', now() - alert_data_processing.evaluation_interval)
```
psFried, line 63: I think this `catalog_stats_hourly` condition needs to be a part of the `join` condition instead of the `where`. If this were an inner join, there would be no logical difference. But we need this to be an outer join so that a row is returned even for tasks that have no corresponding `catalog_stats_hourly` rows. Ideally, this could be covered by a test case where there's a `live_specs` row but no corresponding `catalog_stats_hourly` rows within the time window.

supabase/migrations/37_alerts.sql
```diff
@@ -0,0 +1,136 @@
...
+select
+  alert_data_processing.*,
+  'data_not_processed_in_interval' as alert_type,
+  alert_subscriptions.email,
+  live_specs.spec_type,
+  coalesce(sum(catalog_stats_hourly.bytes_written_by_me + catalog_stats_hourly.bytes_written_to_me + catalog_stats_hourly.bytes_read_by_me), 0)::bigint as bytes_processed
+from alert_data_processing
+  left join live_specs on alert_data_processing.catalog_name = live_specs.catalog_name and live_specs.spec is not null and (live_specs.spec->'shards'->>'disable')::boolean is not true
+  left join catalog_stats_hourly on alert_data_processing.catalog_name = catalog_stats_hourly.catalog_name
+  left join alert_subscriptions on alert_data_processing.catalog_name ^@ alert_subscriptions.catalog_prefix
```
psFried, line 61: My reading of this is that we'd get an `alert_data_processing_firing` row with a null `email` if there's no subscribers. Is that intentional/important? Might be another good case to cover in the tests.

supabase/migrations/37_alerts.sql
```diff
@@ -0,0 +1,136 @@
...
+from internal.alert_data_processing_firing
+group by
+  internal.alert_data_processing_firing.catalog_name,
+  internal.alert_data_processing_firing.alert_type,
+  internal.alert_data_processing_firing.bytes_processed,
+  internal.alert_data_processing_firing.evaluation_interval,
+  internal.alert_data_processing_firing.spec_type
+order by catalog_name asc;
+
+create or replace function internal.evaluate_alert_events()
```
psFried, line 90: Would be good to have test coverage of this function, too.

psFried (changes requested):
This is coming along nicely! I left a few comments, not all of which require changes. IMO the most important thing is to add SQL test coverage.

supabase/functions/alert-data-processing/index.ts
```diff
@@ -0,0 +1,186 @@
...
+                catalog_name,
+            }) => {
+                let formattedEvaluationInterval = evaluation_interval;
+
+                // A postgresql interval in hour increments has the following format: 'HH:00:00'.
+                if (evaluation_interval.includes(":")) {
+                    const timeOffset = evaluation_interval.split(":");
+                    const hours = Number(timeOffset[0]);
+
+                    formattedEvaluationInterval = isFinite(hours) ? `${hours} hours` : `${timeOffset[0]} hours`;
```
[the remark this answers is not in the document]
kiahna-tucker: `timeOffset` corresponds to the hour substring of the evaluation interval, taking the format `HH` (e.g., `'02'`, `'04'`, `'12'`). Ideally, an hour-based interval that is less than ten would be represented by a single digit (e.g., a two-hour evaluation interval, `'02:00:00'`, should be translated to the following substring in the email: `2 hours`). To accomplish this, the hour segment of the evaluation interval is selected (i.e., `timeOffset[0]`) and that substring is attempted to be converted to a number. This conditional is a failsafe, in the event the aforementioned conversion fails which would result in the display of two digits for the hour (e.g., `02 hours` instead of `2 hours`).

supabase/functions/alert-data-processing/index.ts
```diff
@@ -0,0 +1,186 @@
...
+                }
+
+                const subject = `Estuary Flow: Alert for ${spec_type} ${catalog_name}`;
+
+                const content =
+                    `<p>You are receiving this alert because your task, ${spec_type} ${catalog_name} hasn't seen new data in ${formattedEvaluationInterval}.  You can locate your task <a href="https://dashboard.estuary.dev/captures/details/overview?catalogName=${catalog_name}" target="_blank" rel="noopener">here</a> to make changes or update its alerting settings.</p>`;
```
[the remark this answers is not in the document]
kiahna-tucker: This was the content directly specified by Dave in the PRD. Any content changes require his approval.

supabase/functions/alert-data-processing/index.ts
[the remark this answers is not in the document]
psFried: Makes sense, thanks for the explanation. Might be good to leave a comment as to the purpose, in case others get confused by that too.

supabase/functions/alert-data-processing/index.ts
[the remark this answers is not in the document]
psFried: ah, please disregard, then.

supabase/migrations/37_alerts.sql
kiahna-tucker: The existence of an email can be a mandatory condition while it is the only alert method available. The UI does not allow `alert_subscriptions.email` to be `null` but there is no harm in being explicit in this join condition.

supabase/functions/alert-data-processing/index.ts
```diff
@@ -0,0 +1,186 @@
...
+        ({ emails, content, subject }) =>
+            fetch("https://api.resend.com/emails", {
+                method: "POST",
+                headers: {
+                    ...corsHeaders,
+                    "Content-Type": "application/json",
+                    "Authorization": `Bearer ${RESEND_API_KEY}`,
+                },
+                body: JSON.stringify({
+                    from: "Estuary <onboarding@resend.dev>",
```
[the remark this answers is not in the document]
kiahna-tucker: Yes, I moved it over to an environment variable as well.

supabase/functions/alert-data-processing/index.ts
```diff
@@ -0,0 +1,218 @@
...
+                }),
+            }),
+    );
+
+    await Promise.all(notificationPromises);
+};
+
+serve(async (_request: Request): Promise<Response> => {
+    const startTimestamp = new Date();
+    const minuteOffset = startTimestamp.getUTCMinutes() - 5;
```
psFried, line 89: One other thing I just noticed: This can result in `minuteOffset` being negative, and the behavior of `setUTCMinutes` seems pretty surprising in that case.

kiahna-tucker: This should not be a problem given the only use of `minuteOffset` is when it is passed to `setUTCMinutes()`, which handles positive and negative integers as expected (e.g., `date.setUTCMinutes(-3)` results in the minute value of the `date` being `57`).

psFried (approved):
The tests look good. Thanks.

supabase/functions/alert-data-processing/index.ts
kiahna-tucker: It should not impact the correctness of the edge function since `minuteOffset` will only be negative when the current time is within five minutes of the hour.

But, to your point, I do believe many people would be surprised by the behavior of `setUTCMinutes`.

PR #1251, opened 2023-10-20
introduce `SerPolicy` and use it to truncate strings in some cases

jgraettinger:
SerPolicy is a serialization policy which is used for document extraction and serialization.

In most cases we just want "default" behavior, which is to continue serializing the same way we have been.

For specific materialization connectors, however, apply a policy which truncates strings larger than 64KB. This is to help these connectors work around fundamental limitations in these materialization destinations.

In the future, we should make SerPolicy an explicitly negotiated portion of the connector protocols.

Testing:
 * I included `/materialize-sqlite` in the list of truncated connectors.
 * I spun up a local stack and configured a hello-world capture, using a custom message which is about 150KB.
 * I materialized it using the sqlite connector and confirmed that messages were being truncated as expected:
 
```
sqlite> select length(message), length(flow_document) from events limit 1;
65536|65644
```

**Workflow steps:**

Immediately applies to existing redshift and snowflake materializations which are named as such.

**Notes for reviewers:**

Most of this is pretty mechanical fallout from removing `impl Serialize for Node` and friends, which guarantees they can't be serialized without a `SerPolicy`.

This PR also depends on the reduce-by-ref PR #1250 which I pulled out of a different work stream, and was fortunately ready to go. It's required because `Node` is no longer directly serializable (to make sure a `SerPolicy` is always used) and as-is on master, there's a very awkward spot where it needs to be that this 👈 PR resolves.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1251)


go/runtime/materialize.go
```diff
@@ -130,6 +131,16 @@ func (m *Materialize) RestoreCheckpoint(shard consumer.Shard) (cp pf.Checkpoint,
 		if err != nil {
 			return nil, err
 		}
+
+		// TODO(johnny): Hack to address string truncation for these common materialization connectors
```
jgraettinger, line 135: Very hack-y mechanism to have this take effect immediately for materialization connectors that need it. I added sqlite as an aide for end-to-end testing.

crates/runtime/src/derive/combine.rs
```diff
@@ -303,12 +305,15 @@ impl Opened {
             ..
         } = derivation.as_ref().context("missing derivation")?;
 
+        // TODO(johnny): Expose to connector protocol and extract from Open/Opened.
+        let ser_policy = doc::SerPolicy::default();
```
jshearer, line 309: Is this where the Rust runtime would configure the `ser_policy`? I.e, if we were actually running container-based connectors using the Rust runtime, the hack to enable this for snowflake and redshift would be here?

jshearer (approved):
The actual `SerPolicy` stuff looks to be straightforward, it just ends up effecting a lot of different locations. The actual configuration appears to be on a per-combiner basis which is good as it allows us to configure this per task if we want to, even though for the time being we want it configured per connector.

One question I have is: this should only be set for materialization connectors, right? As in, we would never want to truncate, for example, a captured document being written to a collection, just the document that gets materialized to the destination. Do we just control that here just by not setting a restrictive `SerPolicy` for capture connectors? Should we make that constraint more explicit to prevent a footgun? 

LGTM 👍

crates/runtime/src/derive/combine.rs
jgraettinger: Almost. This file is specifically about combining for derivations, where I'm not currently aware of a use case for `SerPolicy`, but I also mostly think I lack the required imagination.

There's a corresponding combine.rs i've been working on for captures in a different branch, and materializations will be after that, and yes a hack would live there (once it's written)

jgraettinger:
> One question I have is: this should only be set for materialization connectors, right?

I don't know. Having a way to inject new behaviors, that's consistent for `flow_document` as well as extracted fields, is a pretty broad capability. Example features we're not building but _maybe_ could using `SerPolicy`:

* Stripping out or redacting certain identifies in outputs.
* Making over-large captured documents "smaller" (we do have a 64MB system limit right now, that can impact captures too).
* Changing rendering of certain types (for example, hex vs base64 byte encoding??). We could add further variants to our own `doc::Node` type, using `SerPolicy` to understand how they map back out into JSON.

At this point I have no idea how it should be used beyond this case, but I see the utility in the joint, and expect we'll probably have more ideas in the future.

dyaffe:
I would love to implement the 64MB limit on captures somehow.  We have a binding that we can't capture for a client right now which that would help.

PR #1252, opened 2023-10-20
Phil/sneaky spec expansion

psFried:
**Description:**

This rolls up a few related bug fixes that were identified in the course of troubleshooting a panic in the agent:

- Stop adding built specs to `live_specs` rows when the corresponding `draft_specs` row represents a deletion
- Migration to remove all the `built_specs` that aren't supposed to be there
- Stop resolving remote collections for bindings that are disabled

The individual commit messages have move details. [This slack thread](https://estuaryworkspace.slack.com/archives/C012S5U0WP2/p1697719643581899) contains this summary:

The steps to reproduce are simple:
- Create hello-world capture
- Disable hello-world and publish
- Delete the events collections and publish, which triggers the panic due to trying to add the built_spec to live_specs after the spec_type is set to null :boom:

So why is there a built_spec that it would be adding to collections?
- The draft starts out with only the events collection (with null spec/type)
- Spec expansion expands to include the hello-world capture. This is not actually a bug, because we still want to show the connected collections for tasks that are disabled.
- The draft catalog is built, containing only the source-hello-world capture
- When building, we again [don't check for disabled tasks or bindings](https://github.com/estuary/flow/blob/3a1700b7dbeae3448b91d182720f31d0335a1192/crates/validation/src/reference.rs#L4) when we resolve referenced collections, so the collection spec (which we're currently deleting) gets pulled into the build, and subsequently output in `built_collections`

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1252)

### DIFF
--- a/crates/agent/src/publications/specs.rs
+++ b/crates/agent/src/publications/specs.rs
@@ -478,6 +478,12 @@ pub async fn add_build_output_to_live_specs(
     build_output: &builds::BuildOutput,
     txn: &mut sqlx::Transaction<'_, sqlx::Postgres>,
 ) -> Result<(), sqlx::Error> {
+    // We use the `draft_spec.is_some()` check throughout in order to avoid
+    // adding built specs to live_specs that are being deleted by this
+    // publication. These can be present in `built_collections` due to being
+    // referenced by other tasks in the build. Technically, only collections are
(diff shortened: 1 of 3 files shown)


jgraettinger (approved):
LGTM !

PR #1257, opened 2023-10-24
agent: sanitize null bytes in log lines

psFried:
**Description:**

Sanitizes log lines before persisting them to `internal.log_lines`. We observed a failure in production due to a connector logging a null byte. Postgres does not allow nulls in `TEXT` columns, even though they are valid UTF-8, so this removes them before persisting to postgres. Log lines are left unchanged if they're not being saved to postgres.

**Notes for reviewers:**

I've tested that this doesn't completely break anything, but not that this actually allows us to work end to end with a connector that logs out null bytes.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1257)

### DIFF
--- a/crates/agent/src/logs.rs
+++ b/crates/agent/src/logs.rs
@@ -47,6 +47,19 @@ where
     Ok(())
 }
 
+/// NULL is a perfectly valid thing to include in UTF8 bytes, but not according
+/// to postgres. It rejects `TEXT` values containing nulls. This replaces all
+/// null bytes with a space character. The space was chosen somewhat arbitrarily
+/// as a goodenuf replacement in this rare edge case.
+fn sanitize_null_bytes(line: String) -> String {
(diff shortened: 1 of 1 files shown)


crates/agent/src/logs.rs
```diff
@@ -102,7 +115,7 @@ pub async fn serve_sink(
             trace!(%token, %stream, %line, "rx (cont)");
             tokens.push(token);
             streams.push(stream);
-            lines.push(line);
+            lines.push(sanitize_null_bytes(line));
```
jgraettinger, line 118: also need to get the line that's pop'd in the select above.

jgraettinger (approved):
LGTM % fixing the additional location

crates/agent/src/logs.rs
psFried: Whoops, good catch!

PR #1259, opened 2023-10-25
docs: First pass at documenting schema evolution features

psFried:
Updates some existing concepts pages to add docs on new schema evolution features. Also adds a new concepts page on evolutions, and a new guide outlining how to handle some common schema evolution scenarios.

Keeping this as a draft for now, so we can look at the preview and iterate.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1259)


site/docs/guides/schema-evolution.md
```diff
@@ -0,0 +1,244 @@
...
+When recommended fields are enabled, new fields are added automatically if they meet the criteria for the particular materialization connector. For example, scalar fields (strings, numbers, and booleans) are considered "recommended" fields when materializing to database tables.
+
+If your materialization binding is set to `recommended: false`, or if the new field is not recommended, you can manually add it to the materialization.
+
+To manually add a field:
+
+* **In the Flow web app,** [edit the materialization](./edit-data-flows.md#edit-a-materialization), find the affected binding, and click **Show Fields**.
+* **Using flowctl,** add the field to `fields.include` in the materialization specification as shown [here](../concepts/materialization.md#projected-fields).
+
+### A field's data type has changed
```
oliviamiannone, line 162: @psFried would the UI show that schema evolution dialog option (like with the changed key & projections) in this case?

site/docs/concepts/advanced/evolutions.md
```diff
@@ -18,44 +27,64 @@ import Mermaid from '@theme/Mermaid';
     Materialization-->Dest[Destination System];
 `}/>
 
-Collection specs serve as a formal contract between producers (captures) and consumers (derivations and materializations) of data. 
-This contract encompasses the `id`, `schema` (or `readSchema` and `writeSchema` if defined separately), and logical partitioning of the collection. When any part of the contract changes, both the producers and consumers must approve of the change. If any party rejects the proposed change, Flow will fail the publication with an error. Evolutions are a feature that updates your draft to allow such a publication to proceed.
+Each collection and its data are defined by a **collection specification**, or spec.
+The spec serves as a formal contract between the capture and the materialization, ensuring that data is correctly shaped and moves through the Data Flow without error.
 
-Collection specs may change for a huge variety of reasons, such as:
+The spec includes the collection's `id`, its `schema`, and [logical partitions](./projections.md#logical-partitions) of the collection, if any.
```
psFried, line 33:
```suggestion
The spec includes the collection's `key`, its `schema`, and [logical partitions](./projections.md#logical-partitions) of the collection, if any.
```

site/docs/concepts/captures.md
```diff
@@ -57,21 +57,23 @@ You may then modify the generated configuration as needed before publishing the
 Discovers can also be run when editing an existing capture. This is commonly done in order to add new bindings, or update the collection specs and schemas associated with existing bindings.
 :::
 
-## AutoDiscover
+## Automatically update captures
 
-Capture specs may optionally include an `autoDiscover` property to opt in to periodic discovers that are run in the background. For captures that opt in, the Flow control plane will periodically run a discover operation, and publish the resulting draft.
+You can choose to run periodic discovers in the background by adding the n `autoDiscover` property to the capture. Flow will periodically check for changes to the source and re-publish the capture to reflect those changes.
```
psFried, line 62:
```suggestion
You can choose to run periodic discovers in the background by adding the `autoDiscover` property to the capture. Flow will periodically check for changes to the source and re-publish the capture to reflect those changes.
```

site/docs/guides/schema-evolution.md
psFried: Technically, it depends on whether or not the connector permits the change. If the connector says it's OK, then we won't show any errors or an evolution dialog. But if the connector rejects the change, then we will. It can be a little nuanced, but some general guidelines are:

- materializations to S3 or webhooks will generally tolerate most any change
- database materializations will generally reject most changes to materialized columns, except for dropping `NOT NULL` constraints
- If a given field isn't part of the field selection, then any change will be allowed by the connector (even if it's not a good idea for other reasons)

But yeah, probably the most common scenario would be that changing a field type would result in the UI showing the schema evolution dialog.

oliviamiannone (approved):

PR #1264, opened 2023-10-27
flowctl: don't require sops unless we actually need it

psFried:
**Description:**

The `runtime::unseal` module was a little aggressive about checking to make sure `sops` and `jq` are both on the path.  This would cause errors, even in cases where the connector configuration wasn't actually sops-enctypted.  This moves those checks to be performed only after it's determined that the config needs to be decrypted.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1264)


jgraettinger (approved):
LGTM

PR #1266, opened 2023-10-30
combiner & reduction improvements

jgraettinger:
This is a series of low-level improvements to reduction strategies and the combiner. Major changes:

### Reductions / combiner now support non-associative reductions

These come up in a few places:
  * The JSON Merge Patch algorithm, which is used for connector states, is not always associative (you cannot associatively reduce the sequence (`{"a":{"b":"c"}}`,`{"a":42}`,`{"a":{"d":"e"}}`). I found it _deeply_ unsatisfying that a JSON schema could not previously represent a correct merge-patch algorithm.
  * We've identified a desire for CDC captures to be able to "disable" associative reductions to preserve fine-grain log history, while still offering full reductions for a standard materialization.

### New support for `delete` keywords attached to `merge` and `lastWriteWins` strategies

This keyword marks a location as a tombstone which is carried through in partial reductions, and removed in full reductions. This includes nested locations (such as properties of an object or items of an array), as well as surfacing deletions of the entire document through a `deleted()` accessor of the new `combine::Meta` type.

Supporting `delete` was also a requirement for implementing merge patch as a JSON-Schema.

### Relaxing "loaded" concept into a combiner "front" vs "back" concept.

We previously allowed only a single "loaded" document, which was treated as a fully-reduced base document into which all others are reduced.

The combiner semantics have been tweaked / corrected to have each binding instead declare whether it seeks a partial or full reduction of documents. Then, in full reduction contexts, the left-most document acts as the "base" document into which all others are reduced (if there is only one document we do not perform a reduction).

So, a separate "reduced" concept no longer makes sense, but what _does_ make sense is modeling a separate "front" and "back" of the combiner. The purpose of the combiner "front" is to allow a caller to add document in the most efficient order, and then later add documents at the front which order _before_ those already in the combiner. It's a generalization of the single "loaded" document we had previously.

One practical outcome of this relaxation is it now makes it sensible to convert from a delta-updates materialization to a standard materialization, because the combiner can now handle a series of associative delta-updates which are loaded from the endpoint.

**Workflow steps:**

~~For the moment, most of these features are effectively disabled and we continue to perform "full" reductions for captures and materializations, as we have been. There are no immediate user-facing changes. 
This will become more important as forthcoming capture / materialization runtime refactors begin to use these features.~~

No long accurate: captures will immediately use partial reduction while materializations use full reduction.

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1266)


crates/doc/src/combine/mod.rs
```diff
@@ -210,52 +225,92 @@ impl Combiner {
...
-                .map_err(Error::FailedValidation)?;
-
-            // TODO(johnny): handle _delete.
-            let (node, _delete) = reduce::reduce(lhs, rhs, rhs_valid, &alloc, lhs_reduced)
-                .map_err(Error::Reduction)?;
-
-            Ok((node, lhs_reduced))
+impl Meta {
+    fn new(mut binding: u32, front: bool) -> Self {
+        if front {
```
psFried, line 230: nit: maybe worth an `assert!(binding <= META_BINDING_MASK)`

psFried (approved):
Looks good. I'm a little uncertain about the `associative` keyword on the `LastWriteWins` strategy, particularly about the behavior when there's different permutations of those. It might help to have test cases covering some different permutations of `delete` and `associative`, but I don't feel strongly about it, particularly in light of it being marked as experimental.

jgraettinger:
Thanks. I've since realized I got this slightly wrong and need to update further: it's not correct to associatively reduce down into a single document when writing into collections.

That's because we can never be sure if there's another left-hand document for that key previously written to the collection or not.
 * If there is, it's okay to reduce down to one document.
 * But if there isn't, it's not, because we now no longer have a LHS into which we can perform a correct full reduction during a future materialization.

But of a bummer because it means associative reductions only become helpful when there are > 2 documents of a key in a single transaction.

jgraettinger:
Added one more commit which holds back the initial document of an associatively-reduced group. PTAL

PR #1274, opened 2023-11-02
capture protocol: add resource path to discovered response

psFried:
**Description:**

Discussed in #1268 

Adds a new `resource_path` field to the discovered response message.  This is to support a more advanced merge behavior of dicovered bindings in the agent.

**Notes for reviewers:**

The value of `resource_path` is intentionally ignored by the agent. The plan is to first update connectors to start returning it, and then update the discover merge behavior in the agent.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1274)


jgraettinger (approved):
LGTM

PR #1281, opened 2023-11-09
Implement `flowctl raw oauth` to test connector oauth flow end-to-end

jshearer:
**Description:**

I was working on python connector oauth and realized I didn't have a good way to test the e2e flow, so I threw together this `flowctl raw` command to drive it.

**Note:** I had to modify the `oauth` edge function to allow passing in custom connector configs instead of using the ones from the `connectors` table. This _shouldn't_ be a security risk because we're not leaking any data, just allowing custom configs. The only risk I could see is that an authenticated user could hypothetically use this edge function to obscure a DoS attack through supabase... thoughts?

Output:
Screen Shot 2023-11-09 at 10 25 30

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1281)


supabase/functions/oauth/access-token.ts
```diff
@@ -11,35 +11,42 @@ interface OauthSettings {
 }
 
 export async function accessToken(req: Record<string, any>) {
-    const { state, code_verifier, config, redirect_uri, ...params } = req;
+    const { state, code_verifier, config, redirect_uri, connector_config, ...params } = req;
```
psFried, line 14: I _think_ this is probably OK, but I want to discuss a particular danger so that we can collectively give this a little more scrutiny.

In this context, `connector_config` really only refers to the oauth-related parts. You can either pass your own oauth configs, or use whatever we have in the database for that connector. One of the most important configurations is the connector `endpoint_spec_schema`, which determines which fields are sensitive and need encrypted. For example, if we allowed clients to pass a custom `endpoint_spec_schema` while using "secret" configuration from the database, then the secrets could be leaked. Again, I don't see how that would be possible given the current code. But it might be good to identify and maybe comment the specific invariants that prevent that from happening.

AFAICT, the main thing is really just that we cannot allow any request to provide a custom `endpoint_spec_schema`, but LMK if you've thought of anything else.

crates/flowctl/src/raw/oauth.rs
```diff
@@ -0,0 +1,273 @@
...
+        source,
+        task,
+        network,
+        client_id,
+        client_secret,
+        endpoint_config,
+        injected_values,
+    }: &Oauth,
+) -> anyhow::Result<()> {
+    // TESTING
```
psFried, line 56: The `tracing` stuff should already be setup in `main.rs`, so I don't think we'd need it here. Guessing this was intended to be removed?

crates/flowctl/src/raw/oauth.rs
```diff
@@ -0,0 +1,273 @@
...
+        Ok(index) => &mut sources.captures[index],
+        Err(_) => anyhow::bail!("could not find the capture {needle}"),
+    };
+
+    tracing::info!(
+        task_name = capture.capture.as_str(),
+        "Performing oauth flow on task"
+    );
+
+    let spec_req = match &capture.spec.endpoint {
```
psFried, line 99: No action for now, but I'll point out that we'd like to eventually switch the `connector_tags` job over to using the `runtime` crate to pull connector specs like this. I imagine this section could be easily factored out and exposed as a `async fn spec(...)`, which could also be used by `agent`.

crates/flowctl/src/raw/oauth.rs
```diff
@@ -0,0 +1,273 @@
...
+    .await
+    .map_err(crate::status_to_anyhow)?
+    .spec
+    .context("connector didn't send expected Spec response")?;
+
+    let oauth_spec = spec_response
+        .oauth2
+        .expect("Connector did not return an oauth config");
+
+    // TODO (jshearer): Validate endpoint config against spec.config_schema_json
```
psFried, line 142: Should this be done now, or just something that might be nice?

crates/flowctl/src/raw/oauth.rs
```diff
@@ -0,0 +1,273 @@
...
+                if let Some(tx) = maybe_sender.take() {
+                    tx.send(query).unwrap();
+                    Response::builder().body("You may close this window now.")
+                } else {
+                    Response::builder().body("Something went wrong")
+                }
+            }
+        });
+
+    tokio::select! {
```
psFried, line 265: This seems like maybe a good place for a timeout?

crates/flowctl/src/raw/oauth.rs
```diff
@@ -0,0 +1,273 @@
...
+/// then shuts down and returns that requet's query params.
+pub async fn get_single_request_query<S>(listen_addr: S) -> anyhow::Result<HashMap<String, String>>
+where
+    std::net::SocketAddr: From<S>,
+    S: Debug,
+{
+    let (tx, rx) = tokio::sync::oneshot::channel::<HashMap<String, String>>();
+    // All this because a oneshot Sender's `.send()` method consumes `self`
+    let wrapped = Arc::new(Mutex::new(Some(tx)));
+    let service = warp::get()
```
psFried, line 250: Another optional/informational nit pick, but I've elsewhere found that `axum` is easier to use and has a nicer ecosystem. Kind of a non-issue since we no longer have other rust-based web servers in this repo.

psFried (approved):
I had a few thoughts on this, but nothing blocking. The `tracing` one is the only thing I think really needs to be addressed before merging

supabase/functions/oauth/access-token.ts
jshearer: This sounds right, I can add a comment to `encrypt-config.ts`. FWIW, this PR doesn't actually modify `encrypt-config.ts` at all, so the only `endpoint_spec_schema` we can load must come from the database.

jshearer: FWIW we do also have some basic sanity checking that makes sure we don't expose the plaintext client ID or secret

crates/flowctl/src/raw/oauth.rs
jshearer: So I actually wonder what your opinion is here: currently it seems that we only really use `tracing` for debug logging, and we just do regular old `println!` for normal/informative output. That makes sense when we need to do things like print out tables or capture input, but I would love to be able to use e.g `tracing::info!` to log out informative messages. The problem is, we set up tracing to only log messages that are warn-level and higher, and it didn't seem straightforward to refactor that to be configurable on a per-command basis. Though as I write that, I'm doing exactly that here so I could just extract this `tracing_subscriber` override logic to a function that any command handler could call. 

Am I barking up the wrong tree by trying to use tracing here?

crates/flowctl/src/raw/oauth.rs
jshearer: Done

crates/flowctl/src/raw/oauth.rs
jshearer: Done

crates/flowctl/src/raw/oauth.rs
jshearer: Yeah, I tried to find our best-practice and realized we didn't really have one. Unless you think it's a big deal I'm inclined to leave it

crates/flowctl/src/raw/oauth.rs
psFried: The current state of affairs is pretty messy, and I don't think we follow any principle especially closely. So I'm open to other ideas, but here's how I've been thinking about it:

There's a difference between logging that we're putting in for debugging purposes, and that which is informational and meant to communicate to the user. IMO the main thing is that I want the code to differentiate those somehow, so that it's clear which one it is when you're reading it. For example, if we used `tracing` directly to output user-facing messages, then if you saw `tracing::info!("added specs to draft");`, it may not be clear whether that was meant as a user-facing message or not. IMO this isn't an especially strong argument, but 🤷 it's where we're at. So we've been just using `println!` for user-facing messages.

One possible improvement might be to define a `CliContext::info_message(&self, msg: &str)` function, which could delegate to either `println!` or `tracing::info!`. One other consideration is just that `tracing` outputs to stderr, while `println!` goes to stdout, but I'm not really sure how significant that is.

I think for this PR, we should just stick to the `println!` pattern, but I'd be open to trying to improve on this situation in a separate PR.

supabase/functions/oauth/access-token.ts
psFried: Sounds good, thanks.

PR #1283, opened 2023-11-10
doc: fix panic in `SerPolicy`

psFried:
The `SerPolicy` was panicking due to attempting to split a string in the middle of a utf8 character. Updates the policy to instead truncate strings at the nearest character boundary.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1283)

### DIFF
--- a/crates/doc/src/ser.rs
+++ b/crates/doc/src/ser.rs
@@ -38,6 +38,20 @@ impl SerPolicy {
             str_truncate_after: 512,
         }
     }
+
+    fn apply_to_str<'a>(&self, raw: &'a str) -> &'a str {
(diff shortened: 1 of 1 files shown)


mdibaiee (approved):

mdibaiee (changes requested):
`TuplePack::pack` has the same flawed indexing logic, about 20 lines below your fix, I think that would probably also trip if we don’t fix

psFried:
Good catch @mdibaiee, thanks! I pushed up a fix for that one, too. PTAL

mdibaiee (approved):
thanks!

PR #1284, opened 2023-11-10
build: return all tables after build failure

psFried:
**Description:**

Introduces the `build::BuildOutput` type, which represents the results of a build.  This new type includes all tables, even when there are build errors. It replaces the binary `Result<_,_>` type with one that is able to simultaneously represent both the errors that occurred during the build, as well as the outputs from the successful stages of the build.

The immediate effect of this, and motivation for doing it, is that if fixes a bug that prevented the agent from notifying users about collections with incompatible schema changes.  The agent relies on the presence of `built_materializations` in the output in order to populate the `incompatible_collections` array in the publication status.  The prior `Result` representation didn't propagate `built_materializations` if there were any `errors`, so this feature could not work.  The new, more complete representation allows that to work.

**Workflow steps:**

- Create a postgres capture and materialize it back to another schema
- Run an `alter table alter column ...` statement to change the type of a materialized column
- Edit the capture and hit the "refresh" button to perform a discover
- Click either "test" or "save and publish"
    - Previously, this would _only_ show the error message
    - Now, it once again shows the suggestion to evolve incompatible collections

**Notes for reviewers:**

The `BuildOutput` type may be kind of a temporary facade. I'm confident that we need _something_ here that can represent the complete output of a build, though I'm less confident about what the shape of it should be.

The context for this is that I'm considering other cases where we may want to access different parts of the build output even when there's errors. One example would be populating `built_spec`s into `draft_specs`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1284)

### DIFF
--- a/crates/build/src/lib.rs
+++ b/crates/build/src/lib.rs
@@ -161,6 +161,53 @@ where
     (sources, validations)
 }
 
+/// The output of a build, which can be either successful, failed, or anything
+/// in between. The "in between" may seem silly, but may be important for
+/// some use cases. For example, you may be executing a build for the purpose
+/// of getting the collection projections, in which case you may not want to
+/// consider errors from materialization validations to be terminal.
+pub struct Output {
+    sources: tables::Sources,
+    validations: tables::Validations,
+}
+
+impl Output {
+    pub fn new(sources: tables::Sources, validations: tables::Validations) -> Self {
+        Output {
+            sources,
+            validations,
+        }
+    }
+
+    pub fn into_parts(self) -> (tables::Sources, tables::Validations) {
+        (self.sources, self.validations)
+    }
+
+    /// Returns an iterator of all errors that have occurred during any phase of the build.
+    pub fn errors(&self) -> impl Iterator<Item = &tables::Error> {
+        self.sources
+            .errors
+            .iter()
+            .chain(self.validations.errors.iter())
+    }
+
+    pub fn built_materializations(&self) -> &tables::BuiltMaterializations {
+        &self.validations.built_materializations
+    }
+
+    pub fn built_captures(&self) -> &tables::BuiltCaptures {
+        &self.validations.built_captures
+    }
+
+    pub fn built_collections(&self) -> &tables::BuiltCollections {
+        &self.validations.built_collections
+    }
+
+    pub fn built_tests(&self) -> &tables::BuiltTests {
+        &self.validations.built_tests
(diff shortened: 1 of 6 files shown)


crates/flowctl/src/local_specs.rs
```diff
@@ -119,11 +119,11 @@ async fn validate(
             .as_secs();
 
         let db_path = std::env::temp_dir().join(format!("flowctl_{seconds}.sqlite"));
-        build::persist(Default::default(), &db_path, Ok(&out)).expect("failed to write build DB");
+        build::persist(Default::default(), &db_path, &out).expect("failed to write build DB");
         tracing::debug!(db_path=%db_path.to_string_lossy(), "wrote debugging database");
     }
 
-    out
+    out.into_parts()
```
psFried, line 126: I'm thinking that in the future it seems likely we may want to refactor the flowctl code paths to also use `BuildOutput` directly. But for now, I left it using the old tuple in order to keep the initial scope of changes small and quicker to review and test.

crates/build/src/lib.rs
```diff
@@ -161,6 +161,53 @@ where
     (sources, validations)
 }
 
+/// The output of a build, which can be either successful, failed, or anything
+/// in between. The "in between" may seem silly, but may be important for
+/// some use cases. For example, you may be executing a build for the purpose
+/// of getting the collection projections, in which case you may not want to
+/// consider errors from materialization validations to be build failures.
+pub struct BuildOutput {
```
jgraettinger, line 169: nit: should we call this Output so it doesn't stutter when used as `build::Output` ?

crates/build/src/lib.rs
```diff
@@ -161,6 +161,53 @@ where
     (sources, validations)
 }
 
+/// The output of a build, which can be either successful, failed, or anything
+/// in between. The "in between" may seem silly, but may be important for
+/// some use cases. For example, you may be executing a build for the purpose
+/// of getting the collection projections, in which case you may not want to
+/// consider errors from materialization validations to be build failures.
```
jgraettinger, line 168: supernit: "to be ~build~ terminal failures" ?

jgraettinger (approved):
LGTM much better! one comment on naming

crates/build/src/lib.rs
jgraettinger: `build::Build` could also work

PR #1295, opened 2023-11-27
Correct alert trigger function logic and update edge function

kiahna-tucker:
**Description:**

The following features are included in this PR:

* Include the newly inserted/updated row in the POST request body of the `alert-data-processing` edge function invocation.

* Use a shared secret to enhance the authorization check for the `alert-data-processing` edge function. This required the major version of PostgreSQL used by Supabase be updated to 15 to support the `supabase_vault` extension. Consequently, Johnny made a handful of small tweaks to the following migrations that will be rolled up into this PR: _10_specs_ext.sql_ and _11_stats.sql_. 

* Update the error handling performed in the `alert-data-processing` edge function.

**Workflow steps:**

1. Insert a row into the `alert_data_processing` table for the task(s) of choice. To test receiving an alert, each task must have existed for the desired `evaluation_interval`. Accepted time periods for production alerts are: two hours, four hours, eight hours, 12 hours, 24 hours, and two days.

1. Ensure that the `catalog_stats_hourly` table reflects that there is no processed data by the focal task(s) over the desired `evaluation_interval` to add an entry into the `alert_history` table. A cron job is responsible for inserting into and updating the `alert_history` table. In production, this job runs every ten minutes.

1. Add a non-zero value to the appropriate column of the `catalog_stats_hourly` table row within the desired `evaluation_interval` for the focal task(s) to resolve a firing alert.

Testing this functionality locally will require the following changes:

* Replace the production URL used to invoke the `alert-data-processing` edge function with its local counterpart (see _supabase/migrations/39_alerts.sql_).

* Update the following local, Supabase environment variables: `RESEND_API_KEY` and `ALERT_EMAIL_FUNCTION_SECRET` (see _supabase/env.local_). If you would like to test sending an email, you will need a valid Resend API key which can be obtained by creating a [Resend](https://resend.com/) account yourself or contacting me. The shared secret used in this edge function can be any value.

* Add a secret to the Supabase vault by running the command in the local, Supabase SQL editor: `select vault.create_secret(<shared-secret-in-env-var>, 'alert-email-fn-shared-secret', 'This is an optional description');` (ref: [documentation](https://supabase.com/docs/guides/database/vault#adding-secrets)).

**NOTE:** The cron job responsible for initiating an evaluation of alert events is scheduled to run every ten minutes. This can be a relatively long wait when testing this feature locally, so feel free to reduce that interval as you see fit (see _supabase/migrations/39_alerts.sql_).

N/A

**Notes for reviewers:**

N/A

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1295)


supabase/functions/alert-data-processing/index.ts
```diff
@@ -207,12 +152,42 @@ serve(async (_request: Request): Promise<Response> => {
...
+    const responses = await emailNotifications(
+        pendingEmail,
+        resendToken,
+        senderAddress
+    );
+
+    const errors = responses.filter((response) => response.status >= 400);
+
+    if (errors.length > 0) {
+        console.info(`${responses.length - errors.length} emails sent.`);
```
psFried, line 164: It might be nice to log the `catalog_name` here, to make it easier to correlate errors with catalog items. I could see also consolidating the separate logs into a single statement, like:

```
console.log("finished sending emails", {"catalog_name": catalog_name, "attempted": responses.length, "errors": errors });
```

psFried (changes requested):
I left a few small comments inline, but overall this looks like it ought to work.

There's a few things that I think we may yet want to tweak, but I think can be deferred until later. The main one is that I imagine the edge function could format emails for _all_ alert types. Again, I think it's OK to defer that until we're ready to add another alert type, but just wanted to share how I see this evolving in the future.

supabase/functions/alert-data-processing/index.ts
```diff
@@ -26,164 +26,109 @@ interface EmailConfig {
...
-            headers: { ...corsHeaders, "Content-Type": "application/json" },
-            status: 200,
-        });
-    }
+    const pendingEmail: EmailConfig =
+        alertRecord.resolved_at === null
+            ? formatAlertEmail(alertRecord)
+            : formatConfirmationEmail(alertRecord);
 
     const resendToken = Deno.env.get('RESEND_API_KEY');
```
jgraettinger, line 133: We do need an additional authorization check in this edge function, because it can be used to send (limited) email content to arbitrary email addresses.

There are many ways to do this, and I'm pretty ambivalent on exactly how, but I'll give a suggested approach to make it concrete:
 * Add an additional passed-in Deno.env `SHARED_SECRET`, which corresponds to a `alert-email-fn-shared-secret` secret I just created within the database using the Supabase vault extension: https://supabase.com/dashboard/project/eyrcnmuzzyriypdajwdk/settings/vault/secrets
 * When building requests, the value can then be accessed via view `vault.decrypted_secrets` and passed as an `Authorization` header.
 * Replace the existing public API key Authorization header with this ☝️ by using the `--no-verify-jwt` flag as discussed below.

supabase/migrations/39_alerts.sql
```diff
@@ -124,19 +124,23 @@ create or replace function internal.send_alerts()
...
 end;
 $trigger$ LANGUAGE plpgsql;
 
-create trigger "Send alerts" after insert or update on alert_history
-  for each statement execute procedure internal.send_alerts();
+create trigger "Send email after alert fired" after insert on alert_history
+  for each row execute procedure internal.send_alerts();
+
+create trigger "Send email after alert resolved" after update on alert_history
+  for each row when (old.* is distinct from new.*) execute procedure internal.send_alerts();
```
jgraettinger, line 144: nit: this can be tightened to:

`when (old.resolved_at is null and new.resolved_at is not null)`

It _shouldn't_ matter but i can easily imagine a UPDATE of historical alert rows (by an operator, or a future migration) inadvertently blasting out a ton of emails.

supabase/functions/alert-data-processing/index.ts
jgraettinger: Another option, which might be a bit simpler overall: We can add the Resend API key as a vaulted secret in supabase, and pass it into the edge function.

This makes the edge function itself unprivileged: it's only useful if you already have a valid Resend API key to call it with.

kiahna-tucker: I have not been able to get `supabase/vault` working locally. I am running into the same issue reported in this [discussion](https://github.com/orgs/supabase/discussions/9022) when I attempt to create the `supabase_vault` extension (ref: [`supabase/vault` installation instructions](https://github.com/supabase/vault#installation)).

Error reported by the trigger function when attempting to fetch the secret:
```
ERROR:  relation "vault.decrypted_secrets" does not exist
LINE 1: select decrypted_secret            from vault.decrypted_secr...
```

Error reported when attempting to create the extension:
```
Error running SQL: could not open extension control file "/usr/share/postgresql/14/extension/supabase_vault.control": No such file or directory
```

jgraettinger: Taking a look myself. If we can't get it working, I think it'd be fine to use a regular table to store & plumb through the credential -- it just can't be in our migrations. `vault` looked nice and easy with built-in UI so :shrug:

kiahna-tucker: **NOTE:** Updating the major version of PostgreSQL used by Supabase to 15 enabled support for the `supabase_vault` extension. Johnny made a handful of small tweaks to the following migrations that will be rolled up into this PR: _10_specs_ext.sql_ and _11_stats.sql_. The details of this note will be called out in the PR description explicitly for general awareness.

supabase/functions/alert-data-processing/index.ts
```diff
@@ -26,164 +26,109 @@ interface EmailConfig {
...
-    const pendingEmails = [...pendingAlertEmails, ...pendingConfirmationEmails];
-
-    if (pendingEmails.length === 0) {
-        return new Response(null, {
-            headers: { ...corsHeaders, "Content-Type": "application/json" },
-            status: 200,
-        });
-    }
+    const pendingEmail: EmailConfig =
+        alertRecord.resolved_at === null
```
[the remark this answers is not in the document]
kiahna-tucker: I created a shallow type guard to check for the existence and type of high-level, `AlertRecord` properties, with the exclusion of a type check for the `argument` property` given its complexity.

supabase/migrations/39_alerts.sql
```diff
@@ -124,19 +124,23 @@ create or replace function internal.send_alerts()
...
-    url:='https://eyrcnmuzzyriypdajwdk.supabase.co/functions/v1/alert-data-processing',
-    headers:='{"Content-Type": "application/json", "Authorization": "Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJzdXBhYmFzZSIsInJlZiI6ImV5cmNubXV6enlyaXlwZGFqd2RrIiwicm9sZSI6ImFub24iLCJpYXQiOjE2NDg3NTA1NzksImV4cCI6MTk2NDMyNjU3OX0.y1OyXD3-DYMz10eGxzo1eeamVMMUwIIeOoMryTRAoco"}'::jsonb,
-    body:=concat('{"time": "', now(), '"}')::jsonb
-  );
+if new.alert_type = 'data_not_processed_in_interval' then
+  perform
+    net.http_post(
+      'https://eyrcnmuzzyriypdajwdk.supabase.co/functions/v1/alert-data-processing',
+      to_jsonb(new.*),
+      headers:='{"Content-Type": "application/json", "Authorization": "Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJzdXBhYmFzZSIsInJlZiI6ImV5cmNubXV6enlyaXlwZGFqd2RrIiwicm9sZSI6ImFub24iLCJpYXQiOjE2NDg3NTA1NzksImV4cCI6MTk2NDMyNjU3OX0.y1OyXD3-DYMz10eGxzo1eeamVMMUwIIeOoMryTRAoco"}'::jsonb
```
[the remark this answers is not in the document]
kiahna-tucker: Instead of deploying the edge function with the `--no-verify-jwt` flag, I set the `verify_jwt` property in the configuration file to `false` so that this setting is applied locally as well (ref: [Supabase CLI Configuration](https://supabase.com/docs/guides/cli/config#functions.function_name.verify_jwt)). Let me know if you see any issue with that.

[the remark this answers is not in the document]
jgraettinger: Even better

psFried (approved):
The new migration looks good to me

PR #1299, opened 2023-11-30
runtime: fix schema inference for derivations

psFried:
**Description:**

A user reported a problem with continuous schema inference not working for their derivation. This seems to have been caused by the change to the new derive runtime, which was missing a line to set the `updated_inference` flag when the shape was widened.

I tested this locally by running a derivation. Prior to this change, there were no `inferred schema updated` messages in the task logs. After, those logs appear as expected.


jgraettinger (approved):
LGTM thanks!

PR #1301, opened 2023-12-04
refactor: Update our OAuth flow to be more secure

jshearer:
**Description:**

The impetus for this change is that while I was implementing Airtable OAuth (before I realized that it uses expiring refresh tokens and we had a whole slack conversation concluding that we shouldn't do it for now), I realized that we were doing a naughty. The OAuth authorize flow uses a concept of [state](https://auth0.com/docs/secure/attack-protection/state-parameters) to represent a particular request for authorization. You're supposed to generate a secret "challenge" (we do this), store it internally along with the state key (we _don't_ do this), send a hash of that secret to the provider (we do this), and then use the returned state value and challenge hash to look up the secret and verify that it matches (also don't do this), thereby validating that you haven't been MITM'd. 

The state key is supposed to be a random string, but in practice any string is usually fine. As it turns out, we were using a base64-encoded JSON object containing some metadata about the request (the connector ID, specifically). Airtable has a particularly strict implementation of OAuth, and they intentionally [only allow a few ranges of characters](https://airtable.com/developers/web/api/oauth-reference#authorization-parameter-rules), specifically excluding the `=` character, which usually shows up at the end of a base64-encoded string.

This change has us do the "right thing" by keeping track of ongoing OAuth flows, their state keys, challenge secrets, and connector IDs, and actually doing the challenge validation. That being said, it's not strictly necessary since we are still holding off on Airtable OAuth. I pushed it up because I did the work to implement it and I think it's a good security precaution to have.

* One thought I just had is that we really ought to clear out `oauth_flows` after the flow is finished, but we could also just have a cron job clear out old flows every day.
* The other annoying thing is that in order for the supabase SDK to be able to read the table, it _must_ be in the `public` schema, per Supabase [limitations](https://github.com/orgs/supabase/discussions/1222). I would have preferred to put it in `internal`, but so long as we don't grant `authenticated` to read it, it should be fine...?

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1301)

### DIFF
--- a/crates/flowctl/src/raw/oauth.rs
+++ b/crates/flowctl/src/raw/oauth.rs
@@ -175,7 +175,10 @@ pub async fn do_oauth(
 
     let port = 16963;
 
-    tracing::warn!("Make sure that your application has http://localhost:{port} set as an allowed redirect URL");
+    let redirect_uri = format!("http://localhost:{port}/");
+    tracing::warn!(
+        "Make sure that your application has {redirect_uri} set as an allowed redirect URL"
+    );
     let api = ctx
         .config
         .api
@@ -185,11 +188,10 @@ pub async fn do_oauth(
     let mut oauth_endpoint = api.endpoint.clone();
     oauth_endpoint.set_path("functions/v1/oauth");
 
-    #[derive(serde::Deserialize)]
+    #[derive(serde::Deserialize, serde::Serialize)]
     struct AuthorizeResponse {
         url: String,
         state: String,
-        code_verifier: String,
     }
 
     let authorize_response_bytes = reqwest::Client::new()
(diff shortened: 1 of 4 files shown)


jgraettinger:
I'd like to think about this a bit more. For similar problems in past companies, we had good success using compact encryption / HMAC initialized with the state to avoid MITM, and it seems like a similar thing could be done here, which would obviate the need to store each and every challenge.

jshearer:
How do we solve the problem of needing to attach metadata to the request then? Currently we're just sticking them inside the state key, but that's not recommended since a third party could theoretically change them. 

I guess you're saying that we could have the challenge be the result of signing the key with some secret that we then validate later to ensure that it wasn't modified? That still doesn't solve the problem of character restrictions for the state key, though I suppose we could just use base-24 or something to get purely `a-z` states..

PR #1304, opened 2023-12-06
Phil/discover merge updates

psFried:
**Description:**

Rolls up a number of updates and fixes to the agent. Best to review commit by commit.

**Workflow steps:**

- Tested creating captures in the ui and then trying to delete the collection spec. Prior to this change, I was able to do so. After, it returns an error complaining about the existing capture binding that references the collection.
- Tested to ensure that the `built_spec` is properly updated as specs are published, and nulled out when specs are deleted.
- Manually update `resource_path_pointers` and ensure that a subsequent `connector_tags` job fails due to the change.
- Ensure that `connector_tags` jobs work when `resource_path_pointers` haven't changed.
- Ensure that discovers output the same results before and after this change.

**Notes for reviewers:**

This should not get merged/deployed until after estuary/connectors#1110 is merged, so that it doesn't prevent the `resource_path_pointers` from being updated.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1304)

### DIFF
--- a/crates/agent/src/discovers/specs.rs
+++ b/crates/agent/src/discovers/specs.rs
@@ -21,9 +21,9 @@ pub fn parse_response(
     for binding in &mut bindings {
         binding.recommended_name = normalize_recommended_name(&binding.recommended_name);
     }
-    if bindings.iter().any(|b| b.resource_path.is_empty()) {
+    if bindings.iter().any(|b| !b.resource_path.is_empty()) {
         tracing::warn!(%image_name, %image_tag,
-            "connector discovered response omits resource_path, this is OK for now but will become an error in a future release");
+            "connector discovered response includes deprecated field 'resource_path'");
     }
 
     Ok((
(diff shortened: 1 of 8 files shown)


crates/agent/src/discovers/specs.rs
```diff
@@ -33,40 +35,74 @@ pub fn parse_response(
     ))
 }
 
+#[derive(Clone, Copy, Serialize, Deserialize, Debug, PartialEq)]
+pub enum BindingType {
```
jgraettinger, line 39: nit: this feels pretty heavy-weight as compared to anyhow::bail with `.context()`
It's not intended to be machine accessible, right?

jgraettinger (approved):
LGTM

crates/agent/src/discovers/specs.rs
psFried: right. This is just to make it easier to test error cases of `merge_capture`

PR #1306, opened 2023-12-07
Fix ubuntu references in workflow yaml

psFried:
**Description:**

I'm pretty sure this ought to fix the flowctl linux builds.

Doesn't do anything for the mac builds, because it didn't look like there was a problem with our workflow yaml for that one. I'm hopeful that the mac build may work if we just try it again, as that seemed to be due to a signing error.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1306)


jgraettinger (approved):
LGTM thanks

PR #1314, opened 2023-12-14
ops-catalog: make stats properties optional

psFried:
The runtime started outputting stats documents without any `materialize` property in certain situations. This updates the stats schema so that the `materialize`, `capture`, etc properties are no longer required.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1314)


jgraettinger (approved):
LGTM

PR #1315, opened 2023-12-14
runtime: limit serialized document complexity for materializations

### LINKED ISSUE #1296: More limitations on serialized document complexity
When materializing, we need to respect the limits imposed by destination systems, which may be more restrictive than Flow's own limits. For example, some systems have limits on the number of items in a JSON array, or the number of properties in an object. The ultimate goal is to respect those limits on a system-by-system (i.e. connector-by-connector) basis.

Some immediate questions that come up:

- What types of limits do we wish to put in place?
    - Array length?
    - Object properties?
    - String length? Truncate, or replace the string with a sentinel?
- It might be nice to avoid adding this to the materialize protocol until we have more confidence about how to model these limits. Is there a way to do that without unintentionally applying limits to systems that don't need them?
- What kind of logging would we like to see when these limits modify documents? Or should we try to leave sentinel values?
    - Or we could try adding something under `_meta` that describes which limits were applied, though that could also be tricky if we're trying to limit the overall size of the document, since it would then add to it.

psFried:
**Description:**

resolves #1296

Enhances the `SerPolicy` to include limits on object and array sizes, including a separate limit on the size of nested objects. The overall goal is to automatically truncate very large documents to get them to fit within the limits of systems like redshift, which don't support very large objects.

The current policy is based on the testing results from #1296. It's still being applied based on the connector image name, which is a temporary workaround until we figure out a better way to configure different policies for different materializations.

Note that this PR does _not_ include any way to tell whether a document has been pruned. There's no logging, because it could easily result in very high log volume. We'd previously discussed having a synthetic projection that can indicate whether a value was trimmed, but I'm still not sure how best to implement that. I'm thinking it'd be OK to kick the can down the road on that issue, but would be happy to work on it now if others think that's motivated.

**Workflow steps:**

- Capture some huge documents
- Materialize them to snowflake, redshift, or pinecone
- See that it now (probably, depending on the structure of the documents) works

We don't currently document this, but it'd probably be a good idea to do so. Should I add a blurb to the snowflake, redshift, and pinecone connector docs?

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1315)


crates/doc/src/ser.rs
```diff
@@ -75,16 +92,21 @@ pub struct SerOwned<'p> {
...
     {
         match self.node.as_node() {
-            Node::Array(arr) => serializer.collect_seq(arr.iter().map(|d| Self {
-                node: d,
-                policy: self.policy,
-            })),
+            Node::Array(arr) => {
+                let child_policy = self.policy.for_child();
+                serializer.collect_seq(arr.iter().take(self.policy.array_truncate_after).map(|d| {
+                    SerNode {
```
jgraettinger, line 104: I'm somewhat concerned about performance here, as per profiling we do spend a good amount of time serializing as we move data into Store requests.

Rather than pushing a new cloned policy onto the stack at each level, could you instead track depth on SerNode? What about making SerNode a const generic over a boolean which indicates if it's the root or not, which would let you do a compile time branch when evaluating the policy? This would be zero cost -- no extra stack allocation or runtime branches.

Or, slightly fancier, a const generic integer that does a bounded degree of compile-time depth tracking, by having branches for depths 0, 1, 2, .. X, where depth isn't incremented beyond X. Right now we don't need more than 0 (root) and 1 (node).

crates/doc/src/ser.rs
```diff
@@ -7,15 +7,23 @@ use std::io;
...
     pub str_truncate_after: usize,
+    pub array_truncate_after: usize,
+    pub obj_truncate_after: usize,
+    pub nested_obj_truncate_after: usize,
 }
 
 impl SerPolicy {
-    pub fn new(str_truncate_after: usize) -> Self {
-        Self { str_truncate_after }
+    pub fn truncate_strings(str_truncate_after: usize) -> Self {
```
jgraettinger, line 16: what about a builder pattern? `new()` returns a SerPolicy that does no pruning, and then you call `with_str_truncate_after` or `with_array_truncate_after` etc.

I'm guessing `obj_truncate_after` is in the policy just to make root => node transitions easier to track, right?

crates/extractors/src/lib.rs
```diff
@@ -17,7 +17,7 @@ type Result<T> = std::result::Result<T, Error>;
 pub fn map_policy(policy: &flow::SerPolicy) -> doc::SerPolicy {
```
jgraettinger, line 17: please also add these fields to `flow::SerPolicy` so it's in sync with `doc::SerPolicy` and implement the policy mapping here.

crates/doc/src/extractor.rs
```diff
@@ -93,12 +121,41 @@ impl Extractor {
         doc: &N,
```
jgraettinger, line 121: drive by nit on your WIP: how about not changing the interface for `extract_all`, and having a new function `extract_all_track_pruned` that takes a &AtomicBool (non-optional), and then have `extract_all` delegate to it with a stack-local bool? `sentinel` -> `pruned` (i'm not sure what sentinel means in this context)?

There are many _other_ places we do extraction where we simply don't care about this pruning bit.

crates/doc/src/ser.rs
jgraettinger: a concrete refinement after thinking on it a bit more: it could be generic over const `Depth: usize`, which is instantiated today only with values `0` and `usize::MAX`. In the future, if needed, we could introduce depths `1`, `2`, etc.

crates/derive-sqlite/src/lambda.rs
```diff
@@ -186,11 +186,11 @@ fn bind_parameter_node<N: doc::AsNode>(
         Node::PosInt(u) => stmt.raw_bind_parameter(index + 1, u),
         Node::Array(_) => stmt.raw_bind_parameter(
             index + 1,
-            &serde_json::to_string(&doc::SerPolicy::default().on(node)).unwrap(),
+            &serde_json::to_string(&doc::SerPolicy::default().on(node, None)).unwrap(),
```
jgraettinger, line 189: similar to my comment on extract_all -- this function is used a lot, and most call sites don't care about pruning. Rather than making them, how about updating SerPolicy to add additional ways of calling this which take a non-optional bool reference?

Then, this function can be updated to pass a no-op instance without changing its signature.

crates/doc/src/extractor.rs
```diff
@@ -63,25 +81,35 @@ impl Extractor {
...
-                )
-                .ok()
-            }) {
-                return Err(Cow::Owned(serde_json::Value::String(
-                    date_time
-                        .format(&time::format_description::well_known::Rfc3339)
-                        .expect("rfc3339 format always succeeds"),
-                )));
+            Some(Magic::TruncationSentinel) => {
+                // The real magic is behind some _other_ curtain.
```
jgraettinger, line 108: nit: reference where this actually happens?

crates/doc/src/ser.rs
jgraettinger: I see you added PolicyHelper but IMO it's more verbose and indirected as compared to ☝️ , and also depends on an `obj_truncate_after` concept which we don't have an expected use case for and IMO ought to remove, lest some future reader is lead to believe it _is_ required and must thus be supported indefinitely. I continue to think a Depth const generic makes more sense.

crates/validation/src/collection.rs
```diff
@@ -441,8 +460,13 @@ pub fn walk_selector(
 const FLOW_DOCUMENT: &str = "flow_document";
 /// The default field name for the document publication time.
 const FLOW_PUBLISHED_AT: &str = "flow_published_at";
+/// The default field name for the truncation sentinel.
+const FLOW_TRUNCATED: &str = "flow_truncated";
```
jgraettinger, line 464: this should be named `_meta/flow_truncated` -- the implicit projection of `/_meta/flow_truncated` -- as materializations use the field name to determine whether it's recommended or not.

Also, and this is half baked, but I suspect the validation logic might get easier to reason about if we _don't_ think of `/_meta/flow_truncated` as the projection pointer. Conceptually, this is a value that exists outside of the document, and simply can't be referenced via pointer.

Perhaps it's just tacking on a Projection like:

Projection {
  field: '_meta/flow_truncated',
  ptr: '',
  contentEncoding: 'flow_truncated'
}

after first verifying there isn't an implicit or explicit projection of that name, so we can provide a good error message.

crates/runtime/src/materialize/protocol.rs
```diff
@@ -305,20 +307,42 @@ pub fn send_connector_store(
...
+        buf,
+        Some(&truncation_sentinel),
+    );
 
     // Accumulate metrics over reads for our transforms.
     let stats = &mut txn.stats.entry(binding_index as u32).or_default();
     stats.2.docs_total += 1;
     stats.2.bytes_total += doc_json.len() as u64;
 
-    if !binding.store_document {
```
jgraettinger, line 318: accidental?

crates/derive-sqlite/src/lambda.rs
psFried: I don't find that objectionable at all, and can update to make "prune-tracking" and "non-tracking" versions. But I'll point out that most of those call sites also use a hard coded `SerPolicy::default().on...`. It feels like trying to improve ergonomics in situations where we don't want to be using a `SerPolicy` in the first place. `default` always returns an unlimited policy, and it doesn't seem like a good idea to change that given the large variety of different call sites (I actually tried and broke things).

In other words, I'm questioning why we're using `SerPolicy` at all if we don't want to actually truncate anything, and also don't care about knowing whether we did or not. It feels to me like we ought to only be truncating data in situations where we're able to report on whether or not we actually did. Some usages, like for extracting keys, seem like they would be better off without a `SerPolicy` at all. Others, like for capture combiners, seem like we'd really want to track whether the policy has actually truncated anything.

crates/doc/src/extractor.rs
```diff
@@ -10,7 +13,13 @@ pub struct Extractor {
...
     policy: SerPolicy,
     default: serde_json::Value,
-    is_uuid_v1_date_time: bool,
+    magic: Option<Magic>,
+}
+
+#[derive(Debug, Copy, Clone, PartialEq)]
+enum Magic {
+    UuidV1DateTime,
+    TruncationSentinel,
```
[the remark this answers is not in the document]
psFried: changed to "indicator"

crates/extractors/src/lib.rs
psFried: done

crates/runtime/src/materialize/protocol.rs
psFried: This was intentional. `doc_json` is being created above, and I just skip serializing it if it's not needed.

crates/doc/src/ser.rs
psFried: I could see adding a non-generic `depth: usize` property to the struct, but I don't think it'd work to do something like `struct SerNode<const Depth: usize, ...>` (assuming I'm understanding your suggestion right). The problem I see with that is that the depth can't be `const`, since we only know it by recursively adding one to the previous value. Do you know of a way to make that work?

psFried: Just to clarify, I'm reasonably sure that stable rust won't permit us to use const generics in this way, but am open to being proven wrong. Adding a `depth: usize` field to `SerNode` seems like it could work, though. TBH this seems to me like an implementation detail that'd be easy to change, so I'm inclined to leave it as is until we can see what the next requirements are. But can refactor to use the `depth: usize` approach if you feel strongly about it.

jgraettinger: > since we only know it by recursively adding one to the previous value. Do you know of a way to make that work?

Yea, you can't do recursive Depth + 1 -- the recursive addition breaks -- but you _can_ do:

```
if Depth == 0 {
   SerPolicy::<1>::something()
} else if Depth == 1 {
   SerPolicy::<2>::something()
 } else if Depth == 2 {
   SerPolicy::<3>::something() // and so on...
} else {
   SerPolicy::<usize::MAX>::something()
}
```

And just unroll it only for the actual levels of depth where you plan to have differentiated behavior (which is just Depth = 0 vs Depth = usize::MAX here, so just one branch for now)

crates/derive-sqlite/src/lambda.rs
jgraettinger: >  It feels like trying to improve ergonomics in situations where we don't want to be using a SerPolicy in the first place.

That's accurate. When I implemented SerPolicy, I took away the Serialize implementation for the doc::Node trait deliberately, to force each serialization site to think about whether SerPolicy was applicable or not. So, yea, the goal was to have it be as succinct as possible but to force the writer and reader to think about it every time.

crates/validation/src/collection.rs
psFried: The field name is updated.

> Also, and this is half baked, but I suspect the validation logic might get easier to reason about if we don't think of /_meta/flow_truncated as the projection pointer. Conceptually, this is a value that exists outside of the document, and simply can't be referenced via pointer.

I don't think I see what this is buying us in terms of simplicity. I'm open to discussing more if there's something I'm missing. But the way I see it now it's that we can either match against a phony JSON pointer or a phony content encoding, and I see them as more or less equivalently distasteful. So I'm inclined to leave it as is.

psFried:
I wanted to circle back and try to tie together a few loose threads:

I don't think it makes sense to `impl Default for SerPolicy`, because I don't think there's a reasonable default that's the same in all contexts. For example, the default policy for captures
should almost certainly be different than the default for materializations. I think `SerPolicy::default()` should be removed, and most current call sites should instead transition to `SerPolicy::unrestricted()`, which just makes the current default more explicit and allows us to remove the `Default` impl.
`SerPolicy::unrestricted().on_ignore_truncation(&json)` is what we're left with instead of `SerPolicy::default().on(&json, None)`. It's longer and more obviously silly, but it's also more clear that it's just a placeholder until we decide on a different policy (in which case I assume we probably want to detect truncation).

Instead of a builder pattern, we can use something like:
```rust
let policy = SerPolicy {
  str_truncate_after: 64000,
  ..SerPolicy::unrestricted()
};
```

Optionally, we can add a phantom private field to `SerPolicy` in order to require that users of the API must include the `..SerPolicy::unrestricted()`, to ensure forward compatibility. IMO it'd probably be overkill at the moment, though, as would a builder pattern.

We remove `root_obj_truncate_after` from the `SerPolicy`, and just never truncate root objects.

This is all implemented in the last few commits.

jgraettinger (approved):
LGTM

PR #1317, opened 2023-12-15
Updates to the stripe invoice generator

jshearer:
**Description:**
* For tenants that do not have a credit card:
   * Skip generating invoices for tenants that have no data flowing in the billed month
   * Skip generating invoices for tenants that didn't have both an active capture _and_ an active materialization in the billed month
* Add stub stripe customers table to allow sqlx to properly infer query shapes
* Automatically remove existing draft invoices if a new rule has been introduced that would have skipped that invoice

@dyaffe Is the logic for active dataflow correct here? I interpreted it as being "had >0 active task hours for both capture and materialization task types during the billed month". But if you want it to be more granular, we could change it to check whether there are any capture and materialization task hours on the last day of the billed month?

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1317)


jgraettinger (approved):
I reviewed somewhat quickly but nothing stood out. LGTM

PR #1321, opened 2023-12-20
fix: Make `runtime::Binding` public

jshearer:
This fixes a build error: `can't leak private type`

```
error[E0446]: private type `combine::Binding` in public interface
  --> crates/runtime/src/combine/protocol.rs:6:1
   |
6  | / pub fn recv_client_open(
7  | |     open: Request,
8  | | ) -> anyhow::Result<(doc::combine::Accumulator, Vec<Binding>)...
   | |______________________________________________________________^ can't leak private type
   |
  ::: crates/runtime/src/combine/mod.rs:13:1
   |
13 |   struct Binding {
   |   -------------- `combine::Binding` declared as private
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1321)


jgraettinger (approved):
LGTM

what rust version?

PR #1322, opened 2023-12-20
Updated error message for parser

jonwihl:
**Description:**

- Changed the short link found in the parsers error message
- Added .DS_Store to .gitignore

[**Documentation links affected:**]

(https://go.estuary.dev/Pgy3nf)

Updated the short link for advanced parser config. In this case it points to the docs for GCS, but we use the same docs for all connectors that this option.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1322)


psFried (approved):
LGTM thanks

PR #1333, opened 2024-01-10
parser: error logs when parsing failes and limit is exceeded

mdibaiee:
**Description:**

- This fixes an issue where if during `prefill_buffer` we would already hit the error limit, we would not log any of the errors.
- Moreover, makes the error limit exceeded exception include the first error in the buffer to make it more informative
- One thing that I find missing though is having the file name that is failing as part of the errors, that makes it a lot more useful for users. Looking at how the plumbing works, I think we need to make that change in `filesource` library in connectors repo to include the file name, but it's a bit tricky given that the parser logs the error and `filesource` abstains from logging anything to leave the parser error as the last log line

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1333)

### DIFF
--- a/crates/parser/src/format/character_separated/error_buffer.rs
+++ b/crates/parser/src/format/character_separated/error_buffer.rs
@@ -70,7 +70,8 @@ impl<I: Iterator<Item = ParseResult>> ParseErrorBuffer<I> {
     /// Returns true if we successfully added another item to the buffer.
     fn buffer_next(&mut self) -> bool {
         if let Some(item) = self.inner.next() {
-            if item.is_err() {
+            if let Err(err) = item.as_ref() {
+                tracing::warn!(error=err.to_string(), "failed to parse row");
                 self.errors_in_buffer += 1;
             }
             self.total_records += 1;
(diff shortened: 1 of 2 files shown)


crates/parser/src/format/character_separated/error_buffer.rs
```diff
@@ -70,7 +70,8 @@ impl<I: Iterator<Item = ParseResult>> ParseErrorBuffer<I> {
     /// Returns true if we successfully added another item to the buffer.
     fn buffer_next(&mut self) -> bool {
         if let Some(item) = self.inner.next() {
-            if item.is_err() {
+            if let Err(err) = item.as_ref() {
+                tracing::warn!(error=err.to_string(), "failed to parse row");
```
psFried, line 74: I'm a little mixed on whether we should log these at this point, because it's possible it could result in a lot of log spam. I think if we keep it, the log should maybe indicate that the error is being ignored.

edit: Now that I think of it, perhaps it'd be best to log these at a debug or info level.

psFried (approved):
LGTM % one nit.

> One thing that I find missing though is having the file name that is failing as part of the errors, that makes it a lot more useful for users

Agreed that this shouldn't block this PR. Though I believe we do have the filename available in the `config`, since that's how the `/_meta/file` gets populated. My suggestion would be to add the filename from the config to the tracing `span`, so that it'll automatically get included as part as any of the logs within the span. We can even do the same for the zip archive members, so we can more easily tell which file within a zip archive causes an error.

crates/parser/src/format/character_separated/error_buffer.rs
mdibaiee: Even if the error threshold is set, and we ignore some rows if they fail, is this not a fair warning to the user to say that these rows failed parsing? I think it is

mdibaiee: Updated the message to explicitly mention that we are ignoring because we are below the threshold

psFried: > Even if the error threshold is set, and we ignore some rows if they fail, is this not a fair warning to the user to say that these rows failed parsing? I think it is

Yeah, I think that's totally fair ... but I also think that users might set a 50% threshold (I've actually seen this) and then try to parse CSVs with millions of rows and have a bunch of them fail (I've not actually seen this). If we log it at the info level, they'd be shown by default, but at least it'd be easy to turn it off if it became overwhelming. TBH, I'm not super concerned here, so I'm fine with merging as is and re-visiting if we need to.

PR #1338, opened 2024-01-16
Phil/discover timeouts

psFried:
**Description:**

Adds another exception to our discover timeout for the `source-salesforce` connector, which is routinely timing out in production.

Allows flowctl-go to read the discover timeout from an env variable, so that we can change it without needing to re-build. The env var is `FLOW_DISCOVER_TIMEOUT`, and it's value will be used as the default timeout, which is overridden by the exception cases.

**Workflow steps:**

Do a discover with the source-saleforce connector, and it is now allowed to take up to 5 minutes before timing out.

**Notes for reviewers:**

`FLOW_DISCOVER_TIMEOUT` is not necessarily meant to be a permanent solution, though it may be fine for us to keep using that env variable. But there's a desire to change the default timeout based on observed failures in production, and seems likely that we'll tweak it more than once, so it seemed worthwhile to just use an env var.

I tested this by manually running some `flowctl-go api discover` commands and observing the logged timeout value.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1338)

### DIFF
--- a/go/flowctl-go/cmd-api-discover.go
+++ b/go/flowctl-go/cmd-api-discover.go
@@ -86,6 +86,13 @@ func (cmd apiDiscover) Execute(_ []string) error {
 	mbp.InitLog(cmd.Log)
 
 	var timeout = time.Second * 30
+	if timeoutStr := os.Getenv("FLOW_DISCOVER_TIMEOUT"); timeoutStr != "" {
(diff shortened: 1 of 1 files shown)


jgraettinger (approved):
LGTM

PR #1346, opened 2024-01-18
Phil/evolutions backfill

psFried:
**Description:**

This updates the evolutions handler to increment the backfill counter instead of re-naming resources. Best to review commit by commit. The individual commit messages have more details.

To summarize the new behavior of the evolutions handler:

- It will never add a version suffix to materialization resource names
- It will always increment the backfill counter for affected materializations
- It can optionally limit the evolution to only affect a subset of materializations
- It now sets the `expect_pub_id` column on `draft_specs` that it modifies
- It will increment the backfill counter of affected captures when re-creating collections (but not otherwise)

**Workflow steps:**

Some test cases:

- Changing primary key of a captured table when published via the UI
- Incompatible change to a database column (I just dropped and re-created the source table)
- Incompatible column change published via auto-discover
- Inferred schema changing when publishing a capture via the UI
- Inferred schema changing when publishing a materialization via the UI
- Inferred schema change published due to auto-discover

I tested each of these in an end-to-end fashion by starting with a postgres or webhook capture and a postgres materialization, then publishing a change that gets rejected, and subsequently triggering an evolution.

I updated two docs pages in the last commit.

**Notes for reviewers:**

I'm working on wrapping up a related UI PR, and will link that once it's ready. The scenario where an incompatible inferred schema change is observed when publishing a materialization will break the current UI, and that'll be fixed in the UI PR.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1346)

### DIFF
--- a/crates/agent/src/evolution.rs
+++ b/crates/agent/src/evolution.rs
@@ -423,6 +423,12 @@ fn evolve_collection(
             for binding in new_spec.bindings.iter_mut() {
                 if &binding.target == &old_collection {
                     binding.target = new_name.clone();
+                    // When re-creating collections, it's quite likely that
+                    // users will also want to trigger a new backfill. Unlike
+                    // materializations, capture connectors will only backfill
+                    // when the counter is incremented, not when only the
+                    // collection name is changed.
+                    binding.backfill += 1;
                 }
             }
         }
--- a/crates/agent/src/evolution.rs
+++ b/crates/agent/src/evolution.rs
@@ -5,8 +5,7 @@ use agent_sql::{
 };
 use anyhow::Context;
 use serde::{Deserialize, Serialize};
-use serde_json::Value;
-use std::collections::{BTreeSet, HashMap};
+use std::collections::BTreeSet;
 
 #[cfg(test)]
 mod test;
@@ -169,17 +168,6 @@ async fn process_row(
         anyhow::bail!("unexpected errors from drafted specs: {errors:?}");
     }
 
-    // Create our helper for updating resource specs of affected materialization
-    // bindings. This needs to fetch all of the resource_spec_schemas for each
-    // of the  materialization connectors involved
-    let update_helper = match ResourceSpecUpdater::for_catalog(txn, &before_catalog).await {
-        Ok(help) => help,
-        Err(err) => {
-            tracing::warn!(error=%err, "failed to create ResourceSpecUpdater during evolution");
-            return error_status(err.to_string());
-        }
-    };
-
     let mut new_catalog = models::Catalog::default();
     let mut changed_collections = Vec::new();
     for req in collections_requests.iter() {
@@ -188,7 +176,6 @@ async fn process_row(
             &before_catalog,
             req.current_name.as_str(),
             req.new_name.as_deref(),
-            &update_helper,
         );
         match result {
             Ok(summary) => {
@@ -311,13 +298,12 @@ fn validate_evolving_collections(
     Ok(())
 }
 
-#[tracing::instrument(skip(new_catalog, prev_catalog, update_helper))]
+#[tracing::instrument(skip(new_catalog, prev_catalog))]
 fn evolve_collection(
     new_catalog: &mut models::Catalog,
     prev_catalog: &models::Catalog,
     old_collection_name: &str,
     new_collection_name: Option<&str>,
-    update_helper: &ResourceSpecUpdater,
 ) -> anyhow::Result<EvolvedCollection> {
     let old_collection = models::Collection::new(old_collection_name);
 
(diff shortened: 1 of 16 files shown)


site/docs/concepts/advanced/evolutions.md
```diff
@@ -66,17 +66,17 @@ Alternatively, you could manually update all the specs to agree to your edit, bu
...
 
    This is a simpler change, and how evolutions work in most cases.
 
-* **Re-create the Flow collection with a new name**: The evolution creates a completely new collection with numerical suffix, such as `_v2`. This collection starts out empty and backfills from the source. The evolution also updates all captures and materializations that reference the old collection to instead reference the new collection. This also updates any materializations to materialize the new collection into a new resource.
+* **Re-create the Flow collection with a new name**: The evolution creates a completely new collection with numerical suffix, such as `_v2`. This collection starts out empty and backfills from the source. The evolution also updates all captures and materializations that reference the old collection to instead reference the new collection, and increments their `backfill` counters.
 
    This is a more complicated change, and evolutions only work this way when necessary: when the collection key or logical partition changes, or when a schema change would cause the endpoint system to reject the materialized data.
```
jgraettinger, line 75: > or when a schema change would cause the endpoint system to reject the materialized data.

This is no longer the case, correct?

site/docs/concepts/advanced/evolutions.md
```diff
@@ -66,17 +66,17 @@ Alternatively, you could manually update all the specs to agree to your edit, bu
...
 
-:::info
-Evolutions will soon support the re-creation of materialization resources, such as tables, while keeping the same names.
-:::
+In either case, the names of the destination resources will remain the same. For example, a materialization to Postgres would drop and re-create the affected tables with the same names they had previously. (Previously, it used to add version suffixes, but not any longer)
```
jgraettinger, line 77: >  (Previously, it used to add version suffixes, but not any longer)

I think we can drop this, that it would just be confusing for readers.

I lol'd at the thought of documenting all the behavior we _used_ to have and no longer do... yesh.

site/docs/guides/schema-evolution.md
```diff
@@ -125,20 +126,15 @@ materializations:
...
-
-:::warning
-Currently, changing the `target` collection in the capture spec will _not_ cause the capture to perform another backfill. This means that the `anvils_v2` table will get all of the _new_ data going forward, but will not contain the existing data from `anvils`.
-
-We will soon release updates that make it much easier to keep your destination tables fully in sync without needing to change the names. In the meantime, feel free to [reach out on Slack](https://join.slack.com/t/gazette-dev/shared_invite/enQtNjQxMzgyNTEzNzk1LTU0ZjZlZmY5ODdkOTEzZDQzZWU5OTk3ZTgyNjY1ZDE1M2U1ZTViMWQxMThiMjU1N2MwOTlhMmVjYjEzMjEwMGQ) for help.
-:::
+ Also note the addition of the `backfill` property. If the `backfill` property already exists, just increment it's value. For the materialization, this will ensure that the destination table in Snowflake gets dropped and re-created, and that the materialization will backfill it from the beginning. In the capture, it similarly causes it to start over from the beginning, writing the captured data into the new collection.
```
jgraettinger, line 137: nit: it's value => its value

jgraettinger (approved):
LGTM!

some minor factoring comments to take or leave, and grammar stuff.

crates/agent/src/evolution.rs
```diff
@@ -232,92 +253,65 @@ async fn process_row(
...
     let mut seen = BTreeSet::new();
-    for row in spec_rows
-        .iter()
-        .filter(|r| evolving_collections.contains_key(r.catalog_name.as_str()))
-    {
-        let old_name = row.catalog_name.as_str();
-        seen.insert(old_name.to_owned());
+    for req in reqs {
+        // Make sure there's only one request per collection
+        let old_name = req.current_name.as_str();
```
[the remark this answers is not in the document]
psFried: yeah, that's fair, but it seemed preferable to keep all the validations in one place, since we need the `spec_rows` for at least some of the validations. This probably would make sense in conjunction with your other suggestion about consolidating the requests and `spec_rows`, though.

site/docs/concepts/advanced/evolutions.md
psFried: lol fair. I only thought to add that because of how infamous the `_v2` tables were, but I'm sure they'll be forgotten quickly

psFried:
Thanks, I did a quick pass to at least address all the comments on the docs.
I've been doing some more testing, and unfortunately discovered a problem due to evolutions setting `expect_pub_id`. I'm still running down exactly what the problem is, but decided to pull that commit out of this PR, since it didn't seem worthwhile for it to hold up the rest of this.

PR #1351, opened 2024-01-23
agent: evolutions set expect_pub_id in draft_specs

psFried:
**Description:**

Updates the evolutions handler to have it set the `expect_pub_id` column when creating draft specs. This clears up a minor bit of tech debt, and helps prevent changes from an auto-discover from clobbering changes from a concurrent manual publication.

**Workflow steps:**

I've tested this both manually and via auto-discovers, and with multiple materializations of the same collection. Also tested the interaction with linked materializations, and I'm confident that we won't see any issues there since we only create linked materialization publications after successfully publishing the evolved specs.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1351)

### DIFF
--- a/crates/agent-sql/src/evolutions.rs
+++ b/crates/agent-sql/src/evolutions.rs
@@ -90,6 +90,7 @@ where
     Ok(())
 }
 
+#[derive(Debug)]
 pub struct SpecRow {
     pub catalog_name: String,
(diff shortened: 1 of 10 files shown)


jgraettinger (approved):
LGTM

PR #1353, opened 2024-01-24
feature: Generalized email sending

### LINKED ISSUE #1337: Billing Updates: Email when someone goes to paid tier & Free connectors

jshearer:
We want to support sending multiple different types of emails. This refactors the alert sending logic to support arbitrary `alert_type`s. Also introduced is much nicer email generation using [mjml](https://mjml.io/).

#### Introducing the following alert types:
- [x] `free_tier_exceeded`: This fires once a tenant is transitioned into the free trial. As I write this, maybe it should be called `free_trial_started`...
- [x] `free_trial_ending`: This fires 5 days before a tenant's trial is set to end (as defined by `trial_start + interval '1 month'`)
- [x] `free_trial_ended`: This fires once a tenant's trial ends
- [x] `free_trial_grace_period_over`: This fires 5 days after a tenant's trial ends. Currently it just sends an emails to `dave@estuary.dev` indicating whether that tenant has entered a CC or not
- [x] `data_not_processed_in_interval_v2`: This is the same as `data_not_processed_in_interval_v2`, except we include the recipient's name if we have it, so we can send the nice `Dear so and so` line
- [x] `provided_payment_method`: This fires whenever a tenant enters CC. It provides as a parameter:
   * Whether or not the tenant is currently within their trial
   * Whether or not the tenant has a trial set
   * With these two data points, we will send 3 different messages:
     * You added a payment method before getting free trial
     * You added a payment method during your free trial
     * You added a payment method after your free trial

#### Questions/concerns:
* It [this](https://github.com/estuary/flow/pull/1353/files#diff-f7948db77679e75917ec3e1c7de83922636d04d500e7fb5570a2237faad9470cR334-R337) the right way to do this migration so we don't generate a bunch of these emails for all tenants that are covered, but didn't transition to the state just now?
* Is [this](https://github.com/estuary/flow/pull/1353/files#diff-f7948db77679e75917ec3e1c7de83922636d04d500e7fb5570a2237faad9470cR146-R148) the right way to do the migration to a new parameter shape? This way we leave the old `internal.alert_data_processing_firing` and email sender, and just transition to the new one by switching the edge function that gets called [here](https://github.com/estuary/flow/pull/1353/files#diff-f7948db77679e75917ec3e1c7de83922636d04d500e7fb5570a2237faad9470cR327) 

- [x] TODO: Add SQL migration to calculate when billing events fire
- [ ] TODO: Test that all of these emails look correct
- [x] TODO: Write a more descriptive PR description
- [ ] TODO: Fall back to email if name is not in supabase

Screen Shot 2024-01-23 at 19 33 03

_Billing email example_

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1353)

Fixes #1337


jshearer:
Yep, really good point. We should probably workshop _all_ the text in these, I only had copy from @dyaffe for one of them.

jgraettinger:
Just from reading the description, and not having gotten into the content, it seems like there's a mismatch of understanding on how alerts are modeled and what they are.

In the "Introducing the following alert types" description above, the text uses event-driven language in reference to all of these event types ("fires once a tenant is transitioned into the free trial" or "This fires whenever a tenant enters CC").

But alerts are fundamentally **_conditions_**, not events. They're conditions that start to become true at some point, and eventually stop being true. As we've implemented it, "starts being true" means appearing as a row in a queried view, and "stops being true" means no longer appearing in that view.

Then, the the job of the existing alerting infrastructure is to figure out the bounds of when that condition starts to fire, and when it stops firing (resolves). That's not something that an alert author is supposed to be concerned with.

Given that -- and again, not having engaged with the content yet -- I'm pretty confused what a `provided_payment_method` alert is supposed to mean, for example, or how it "fires whenever a tenant enters CC".

tl;dr: how can the use cases be adapted to think of the implementing alerts as conditions, and not events?

jshearer:
There may very well be a different way I should be thinking about these, but as far as I can see it right now the functional requirement is "send an email at a particular point in the lifecycle of a tenant". These are one-and-done alerts, you should only _ever_ get one of each type during the lifetime of your tenant (unless we fiddle with the database). That being said, I do understand the design of alerts as "conditions that can become true or false", and have implemented these emails with that in mind:

* `free_tier_exceeded` begins firing as soon as your tenant gets a trial start date, and never stops firing
* `free_trial_ending` begins firing once your tenant is < 5 days before the end of its trial, and stops firing once it's < 4 days from ending. I added the second condition to filter out tenants that have already passed this window, as it's not valid to say that they are "ending" any more
* `free_trial_ended` begins firing once the trial has ended and never stops firing
* `provided_payment_method` begins firing for a tenant as soon as they enter a credit card, and never stops

We only send emails if an alert has started firing, we ignore events that have a non-null `resolved_at`. This leaves the idea of alerts firing/resolving alone for future alert types that may want that functionality

I understand that this is kind of... working _around_ the design rather than _with_ the design, but it seemed like the best option that I could think of. Do you have a better framing that's more aligned with the design?

jgraettinger:
> I understand that this is kind of... working around the design rather than with the design

Yes. And that's what I'm reacting to. For example we'll eventually show UI with the `alert_history` table to users (and they can already access it via API). What are they to think of a `free_trial_ended` alert that shows up and never stops firing?

> Do you have a better framing that's more aligned with the design?

For one, make a `free_trial` alert **condition** that's true when the tenant is within their free trial range. Send an email when it begins to fire. Send another email when it resolves, letting them know that their free trial is over.

Add a `missing_payment_method` condition that starts when a tenant is within N days of their free trial end date AND doesn't have a credit card. If they enter a credit card, it resolves.

Perhaps a `free_trial_almost_done` condition that's true when the tenant is within the last N days of free trial range.

Perhaps a `nonpaying_tenant` condition that fires when the free trial is over by N days and they _still_ haven't entered a CC.

I'm pretty confident this can all map elegantly to condition-based framing.

jshearer:
Hmm... alright, that's helpful. I'll make some changes to try and frame these as meaningful conditions relating to tenants, rather than events.

jgraettinger:
Thanks. Would you please update with the specific set of conditions you're thinking of, to make sure we're all aligned? Those ☝️ were off the cuff and I'm not as privy to the specific user stories that @dyaffe had in mind.

jshearer:
These are _also_ off the cuff, but matching with what Dave said he wants:

* `internal.free_trial_tenants`: selects for tenants currently in their free trial, starting at `trial_start` and ending `interval '1 month'` thereafter. Joins in Stripe info
  * When this starts firing, and the user has a CC
    * This is abnormal, but can happen if we want to give a user the free trial even after they entered the cc. Copy is mine: `Your Estuary account '/tenant' has started its 30-day free trial. This trial will end on 2024-01-01. Billing will begin accruing then.`
  * When this starts firing, and the user has not entered their CC
    * Copy is from Dave: `Your Estuary account '/tenant' has gone over its free tier limit, and is advancing to a 30 day free trial that ends on 2024-01-01. Please enter a credit card before then to continue using the platform`
  * When this stops firing and the user has entered their CC
    * Copy is mine: `Your free trial for '/tenant' has ended. Since you have already added a payment method, no action is required.`
  * When this stops firing and the user has not entered their CC
    * Copy is mine: `Your free trial for '/tenant' ended on 2024-01-01, and your account is now accruing usage. Please enter a payment method within 5 days to continue using the platform`
* `internal.free_trial_ending` selects for tenants with exactly 5 days (ignoring time, so it selects for a 24-hour period) left in their trial. 
  * Starts firing, has cc
    * Copy mine: `Just so you know, your free trial for '/tenant' will be ending on 2024-01-01. Since you have already added a payment method, no action is required.`
  * Starts firing, no cc
    * Copy mine: `Your free trial for '/tenant' is ending on 2024-01-01, at which point your account will begin accruing usage. Please enter a payment method in order to continue using the platform after your trial ends. We'd be sad to see you go!`
  * Stops firing: I don't think it makes sense to send an email here
* `internal.paid_tenants` Simply selects for all tenants that have entered their payment method. This email will always end with an FAQ that currently has "Where is my data stored?" in it
  * Starts firing, within free trial
    * Copy mine: `After your free trial ends on 2024-01-01, your account will begin accruing usage. Your dataflows will not be interrupted.`
  * Starts firing, no free trial set (so, they added CC straight from the free tier)
    * Copy mine: `Your account will now begin accruing usage. Your dataflows will not be interrupted`
  * Starts firing, after free trial
    * Copy mine: `Your account is now active.`
  * I don't think it makes sense to send an email when this stops firing
* `internal.delinquent_tenants` selects for tenants that are >=5 days after their trial end date and have not entered a cc
  * Starts firing
    * We email `dave@estuary.dev` with some data about the tenant
    * Maybe in the future we'll notify the tenant that their account is now cut off until they pay
  * Stops firing
    * Do nothing? Email dave that a delinquent tenant actually gave us a cc? We don't need to email the tenant here because we already cover this case in the `internal.paid_tenants` handler.

dyaffe:
- Free tier exceeded:
Subject: Estuary Free Trial
Body:
{name},

We hope you're enjoying Estuary Flow.  Our free tier includes 10 GB / month and 2 connectors and you've now exceeded that which means your account has been transitioned to a 30 day free trial.

Please add payment information [here](https://dashboard.estuary.dev/admin/billing) in the next 30 days to continue using the platform.  If you have any questions, feel free to reach out to support@estuary.dev.

Thanks,
Estuary Team


 - Free trial ending
Subject: Estuary Paid Tier
Body:
{name},

We hope you are enjoying Estuary Flow.  Your free trial will end on {date}.  

Please add payment information [here](https://dashboard.estuary.dev/admin/billing) in the before that date to continue using the platform.  If you have any questions, feel free to reach out to support@estuary.dev.

Thanks, 
Estuary Team


 - Free trial ended
Subject: Estuary Paid Tier: Enter Payment Info to Continue Access
Body:
{name},

We hope you are enjoying Estuary Flow.  Your free trial will is officially over.

Please add payment information [here](https://dashboard.estuary.dev/admin/billing) immediately to continue using the platform.  If you have any questions, feel free to reach out to support@estuary.dev.

Thanks, 
Estuary Team


 - Provided Payment Method
Subject: Estuary: Thanks for Adding a Payment Method
Body:
{name},

We hope you are enjoying Estuary Flow.  We have received your payment method and your account will be transitioned to the paid tier.

If you have any questions, feel free to reach out to support@estuary.dev.

Thanks, 
Estuary Team

jgraettinger:
`free_trial`, `free_trial_ending`, `delinquent_tenants`: :+1: these all make sense to me as alerts, though maybe the last should be `free_trial_expired` , with the same meaning (free trial is over and _still_ no payment).

`internal.paid_tenants` -- I'm thinking this should be inverted to `missing_payment_method`, which makes more sense as an open alert condition.

When it fires, we could email nothing, or we could email a welcome of some kind, since everybody starts w/o payment.
When it resolves, we email a notification which is further templated on where they are in the free trial cycle.

jshearer:
>though maybe the last should be free_trial_expired
This because we're going to be showing it to users, and calling them delinquent might be a bit mean? :P

> `internal.paid_tenants`
Ah, this is because the default condition for an alert ought to be not firing, right?

jgraettinger:
> > though maybe the last should be free_trial_expired
> This because we're going to be showing it to users, and calling them delinquent might be a bit mean? :P

Well, it's also a very specific kind of not paying us (free trial expired while failing to convert), as opposed to the generalized form of delinquency (account not in good standing). `free_trial_stalled` or `free_trial_failed` would be other options, perhaps.

> > internal.paid_tenants
> Ah, this is because the default condition for an alert ought to be not firing, right?

That's right. If we're looking through open alerts, or users are, the default state for an account in good standing should be no alerts.

`missing_payment_method`, on the other hand, would make sense as a new user who's poking through the open alerts on their account.

jshearer:
Managed to set up resend to test locally, here are how these emails are looking:
Screen Shot 2024-01-30 at 16 43 46
Screen Shot 2024-01-30 at 16 36 47
Screen Shot 2024-01-30 at 16 33 22

I'll update to invert the condition to `missing_payment_method` and use `free_trial_stalled`, then this should be ready for review again 👍

supabase/migrations/43_payment_emails.sql
```diff
@@ -0,0 +1,301 @@
+begin;
```
jgraettinger, line 1: I know you're busy elsewhere, but dropping this feedback for when you're ready. A few things:

* Per our conversation, `paid_tenants` should be inverted to `missing_payment_method`.

* I'd like all of these granular alert views to produce the same table shape: (catalog_name, alert_type, JSON arguments). Then, `alert_all_firing` is the simplest possible UNION ALL of the constituent sub-views.

* Similarly, IIRC we were going for a convention of naming these views `internal.alert_foobar`, where `foobar` is the alert type. This lets views be naturally ordered and grouped in DB resource listings and makes it more obvious what they are.

* Rather than having the edge function collect _all_ alerts whenever it runs, instead have the trigger that _invokes_ the edge function send along the arguments / tokens of a single alert to evaluate. With this design, I don't believe it needs to call back to the DB at all. I'm aware that's not how the current edge function is factored. I've been ... shouting this from the rooftop for some time now.

* nit: rather than `data_not_processed_in_interval_v2` can we call it `data_delay` or something similar ? (Note all of our other discussed alert types are present-tense.)

supabase/migrations/43_payment_emails.sql
```diff
@@ -0,0 +1,292 @@
...
+    'trial_end', (tenants.trial_start + interval '1 month')::date,
+    'has_credit_card', stripe.customers."invoice_settings/default_payment_method" is not null
+  ) as arguments,
+  -- Since we don't need to communicate post-alert arguments, we can instead
+  -- simply omit tenants that are no longer in their free trial, and mark the
+  -- those that are as firing.
+  true as firing
+from tenants
+  left join alert_subscriptions on alert_subscriptions.catalog_prefix ^@ tenants.tenant and email is not null
+  left join stripe.customers on stripe.customers."name" = tenants.tenant
```
jgraettinger, line 42: no action, i think: just noting we'll need to account for google cloud marketplace here in the future.

More generally, there can be payment methods other than stripe.

supabase/migrations/43_payment_emails.sql
```diff
@@ -0,0 +1,292 @@
...
+  'free_trial_stalled',
+  'missing_payment_method',
+  'data_movement_stalled',
+  'data_not_processed_in_interval' -- Old alert type
+);
+
+-- In order to allow alerts to contain arguments after they're done firing
+-- we need to refactor alerts to contain a `firing` boolean, rather than
+-- simply omitting no-longer-firing alerts from the view.
+create type alert_snapshot as (
```
jgraettinger, line 15: I don't actually see this used.

If it makes sense to have views return it, that's great.
Or, I'm perfectly happy with them returning the same consistent shape as they are right now, and this could simply be removed.

jgraettinger (approved):
LGTM

I think this is on a really solid footing now :+1: Thanks for working through the back-and-forth.

jshearer:
## Emails:

### `free_trial`
##### starts firing, no payment method
Screen Shot 2024-02-10 at 18 05 56

##### starts firing, has payment method
**This should never happen unless we manually give someone a trial**
Screen Shot 2024-02-10 at 18 07 21

##### stops firing, no payment method
Screen Shot 2024-02-10 at 18 12 57

##### stops firing, has payment method
Screen Shot 2024-02-10 at 18 10 16

---

### `free_trial_ending`
##### starts firing, has payment method
Screen Shot 2024-02-10 at 18 22 22

##### starts firing, no payment method
Screen Shot 2024-02-10 at 18 23 10

---

### `free_trial_stalled`
Fires when >=5 days past end of free trial and no payment. Stops firing when payment method entered, so the "💳✅" email only sends when they _were_ >=5 days past with no cc, _then_ entered a CC.

**All emails are sent to `dave@estuary.dev` for now.** In the future, we should notify users that they're getting cut off and then actually cut them off.

##### starts firing
Screen Shot 2024-02-10 at 18 33 02

##### stops firing
Screen Shot 2024-02-10 at 18 33 21

---

### `missing_payment_method`
Fires when a tenant has no payment method in Stripe, stops firing when a method is added. We don't notify when it starts firing, since every tenant would get an alert when they sign up.

##### stops firing when tenant is in free tier
Screen Shot 2024-02-10 at 18 39 51

##### stops firing when tenant is in free trial
Screen Shot 2024-02-10 at 18 41 49

##### stops firing when after free trial is over
Screen Shot 2024-02-10 at 18 42 58

---

### `data_processing_stalled`

##### Starts firing
Screen Shot 2024-02-10 at 18 45 50

##### Stops firing
Screen Shot 2024-02-10 at 18 47 03

dyaffe:
Two quick updates if possible @jshearer:

1. For the first email, can we add a new sentence after the words "2 connectors"?
2. Can we update instances of "dataflows" to "data flows"?

PR #1355, opened 2024-01-24
flowctl-go: discover timeout exception for redshift

psFried:
Adds an exception to our growing list of discover timeout exceptions. I'm sure eventually we'll thread through some configuration of this from the `connector_tags` table 😉 but not today

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1355)


jgraettinger (approved):
LGTM

PR #1357, opened 2024-01-29
Phil/csv detection

### LINKED ISSUE #1134: parser: `detect_delimiter` should look for consistent counts across lines
We have a `detect_delimiter` routine which peeks at a prefix of parsed file data and attempts to select a delimiter by looking for the most-frequent byte instance among a common set of delimiter bytes, such as ` `, `,` `\t`, and so on.

It frequently doesn't work. A great example of this, is typical CSV files that are comma-delimited but happen to have text cells containing multiple spaces. In this garden-path case, we frequently pick space ` ` as the delimiter (it's frequent!) and then fail to parse correctly.

Instead, we should look for _consistency_ of the various possible delimiters across a sample of newline-delimited rows. If we see a delimiter type which is most-consistent in its number of instances appearing _across_ rows, that is likely a better choice.

psFried:
**Description:**

Fixes #1134 

The previous implementation of CSV dialect detection was pretty poor, and often resulted in incorrect delimiters.  This introduces a new detection module that works much better. For one thing, the delimiter search space is much more reasonable, in that it no longer includes space characters. Also, the detection now accounts for the whole dialect instead of detecting delimiters and quote characters using two separate functions. Now, we look at the whole dialect (delimiter, quote, escape character, line ending) and assign it an overall
score, and return the dialect with the greatest score.

Scoring is based on two main factors: the number of rows that were able to be successfuly parsed, and the consistency of columns between rows. The idea behind the number of rows is that incorrect dialects will result in parse errors, and thus a lower count of succesfully parsed rows. A dialect that fails to parse the entire sample can pretty trivially be ranked worse than one that is able to parse all the content.  The idea behind consistency is that the correct dialect will result in rows having a relatively consistent number of columns. We use the standard deviation of the column counts to measure this. The score comparison also accounts for the average number of columns, and prefers dialects resulting in a greater number of columns, as long as the standard deviation isn't too much worse. These two factors (standard deviation and average column count) are combined into an overall [coefficient of
variation](https://en.wikipedia.org/wiki/Coefficient_of_variation), and we prefer scores having a lower coefficient.

**Workflow steps:**

No change to workflow. Just use a filesource connector to capture CSV files, and you should see an improved ability to parse files without explicit configuration.

**Notes for reviewers:**

Given the ubiquity of CSV and the endless sprawl of different dialects, you'd think that this would be more of a solved problem. I did find [this paper](https://gertjanvandenburg.com/papers/VandenBurg_Nazabal_Sutton_-_Wrangling_Messy_CSV_Files_by_Detecting_Row_and_Type_Patterns_2019.pdf), which uses a slightly different measure of "consistency" to score dialects, and the [clevercsv](https://clevercsv.readthedocs.io/) python package that implements the algorithm described there. But I opted not to use the exact process described there for a few reasons:

- It's computationally expensive, as computing the "type score" requires a second pass to parse each column value for each candidate dialect.
- I found some test cases in the clevercsv repo that seemed just plain wrong, which degraded by trust a bit.
- I found the explanation in the paper to be extremely obtuse for what amounts to "count the delimiters in each row".
- I compared their method of scoring (without the "type score") to the one implemented here, and found it to perform worse on the corpus of test files I've accumulated so far.

In any case, that paper may still be worth keeping in mind if we find a need to revist dialect detection again in the future, as the "type score" seems to help accuracy quite a bit.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1357)


crates/parser/tests/happy_path_test.rs
```diff
@@ -95,6 +95,9 @@ fn assert_file_is_parsed_with_config(file: PathBuf, mut config: ParseConfig) {
     config
         .add_values
         .insert(JsonPointer::from(canary_ptr), json!(true));
+    config
```
psFried, line 98: FYI: I added this so that the filename will get included in the error message if it fails to parse.

williamhbaker (approved):
LGTM

PR #1358, opened 2024-01-30
parser: prefer double quotes in dialect detection

psFried:
Updates the parser CSV dialect detection to explicitly prefer double quotes when there's a tie between two dialects.  Also improves the logging of dialect detection to log the top few runners up, to aid in debugging.

I'm doubtful that this will fix the issue with the specific file that's failing in production, but I do think it will at least help debugging the issue.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1358)


williamhbaker (approved):
LGTM

PR #1364, opened 2024-02-02
crates/json: don't validate strings with underscores as integers or numbers

williamhbaker:
**Description:**

Previously strings like `"1_234"`, `"123_"`, `"1__2"`, `" _123.3"` (etc.) positively validated as having a format of `integer` or `number`. This is problematic for many systems that will process these downstream that do not have such permissive handling for strings as integers/numbers. It is also a little questionable whether we should consider these strings as numeric values in the first place.

This changes those types of strings to not validate as `integer` or `number`, and will instead cause schema inference to consider them to be strings with no format annotation.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1364)

### DIFF
--- a/crates/json/src/schema/formats.rs
+++ b/crates/json/src/schema/formats.rs
@@ -175,10 +175,11 @@ impl Format {
             Self::Integer => ValidationResult::from(
                 BigDecimal::from_str(val)
                     .map(|d| d.is_integer())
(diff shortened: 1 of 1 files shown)


jgraettinger (approved):
LGTM

PR #1365, opened 2024-02-05
Phil/discover name validation

psFried:
**Description:**

Addresses a bug I found while troubleshooting a production issue, and adds better logging in case we try to insert another draft spec with an invalid name. The lack of normalization was not the cause of the issue we saw last week, but it certainly would present similarly if a connector returned a recommended collection name that wasn't normalized, since nfkc normalization is required by our postgres constraints.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1365)

### DIFF
--- a/crates/agent-sql/src/drafts.rs
+++ b/crates/agent-sql/src/drafts.rs
@@ -20,6 +20,7 @@ pub async fn create(
     Ok(row.id)
 }
 
+#[tracing::instrument(err, skip(spec, txn))]
 pub async fn upsert_spec<S>(
     draft_id: Id,
     catalog_name: &str,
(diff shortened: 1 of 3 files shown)


jgraettinger (approved):
LGTM, but if a connector produces an empty recommended name, that would still cause the agent to fail, right?

Should we a) normalize "" to "_", or b) make this a draft error (though I'm not sure what the User is supposed to do about it)

psFried:
Yeah, the agent will still fail. I was thinking that it's essentially a connector bug, and that for now just having better visibility is helpful. Happy to add another commit to either make this a draft error or normalize the empty string to something else. IMO, making it a draft error is probably the better option at this stage.

PR #1367, opened 2024-02-06
json: fix ordering of integers greater than i64::MAX

psFried:
Fixes a bug in the ordering of JSON numbers that are greater than the maximum signed 64 bit integer.  The previous ordering logic would convert the unsigned integer to signed, which would overflow if the value was larger than `i64::MAX`, leading to an incorrect ordering result.  This fixes that by adding in special cases for comparisons between negative signed integers and unsigned integers. The casts from `u64 -> i64` were replaced with casts from `i64 ->
u64`, relying on the previous checks to ensure that the `i64` value is not negative.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1367)

### DIFF
--- a/.devcontainer/release.Dockerfile
+++ b/.devcontainer/release.Dockerfile
@@ -28,7 +28,7 @@ RUN apt update -y \
      jq \
      nodejs \
      docker-ce-cli \
-     google-cloud-sdk \
+     google-cloud-cli \
      && rm -rf /var/lib/apt/lists/*
 
 # Create a non-privileged "flow" user.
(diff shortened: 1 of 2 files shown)


psFried:
Finally found [an explanation](https://cloud.google.com/blog/products/application-development/redesigning-the-cloud-sdk-cli-for-easier-development) of the change to the package name. Sounds like it's just a re-name of the package, and no other changes, though that information is over a year old now, so I'm not entirely trusting of it.

PR #1371, opened 2024-02-08
Inserting `type` and `main` into `package.json` for flow-web

travjenkins:
**Description:**
To import `flow-web` into vitest we need to set the `type` and `main` in the `package.json` file. Right now `wasm-pack` is not including these but there is a PR has merged that should fix this. 

Also hard cording the name for the output to `flow_web` to ensure these new props are set correctly.

**Workflow steps:**

N/A

N/A

**Notes for reviewers:**

As called out in the comment in the YAML - this is kind of hacky but is a known hack that folks have been doing to work around the issue. We should probably keep an eye on `wasm-pack` to see when it releases a new version (hopefully with the fix for this)

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1371)

### DIFF
--- a/.github/workflows/flow-web.yaml
+++ b/.github/workflows/flow-web.yaml
@@ -39,7 +39,15 @@ jobs:
         run: curl https://rustwasm.github.io/wasm-pack/installer/init.sh -sSf | sh
 
       - name: Build flow-web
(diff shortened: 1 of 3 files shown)


psFried (approved):
LGTM

PR #1374, opened 2024-02-12
thread through `--allow-local` argument when running locally

psFried:
**Description:**

When running locally using `tilt up` on a mac, I noticed I was unable to run any connectors.  This turned out to be caused by a recent change to have the runtime expose ports based on the presence of that flag, instead of being based on the host OS type.  When running on mac, exposing the ports is required, or else the runtime itself can't connect to the containers.  So this carries on the previous work, and threads through the `--allow-local` argument in the couple of places where it was missing.  The end result is that running locally once again works on mac.

**Workflow steps:**

Run `tilt up` ;)

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1374)

### DIFF
--- a/crates/agent/src/discovers.rs
+++ b/crates/agent/src/discovers.rs
@@ -32,14 +32,21 @@ pub struct DiscoverHandler {
     connector_network: String,
     bindir: String,
     logs_tx: logs::Tx,
+    allow_local: bool,
 }
 
 impl DiscoverHandler {
-    pub fn new(connector_network: &str, bindir: &str, logs_tx: &logs::Tx) -> Self {
+    pub fn new(
+        connector_network: &str,
+        bindir: &str,
+        logs_tx: &logs::Tx,
+        allow_local: bool,
+    ) -> Self {
         Self {
             connector_network: connector_network.to_string(),
             bindir: bindir.to_string(),
             logs_tx: logs_tx.clone(),
+            allow_local,
         }
     }
 }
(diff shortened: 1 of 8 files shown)


crates/agent/src/discovers.rs
```diff
@@ -124,31 +131,42 @@ impl DiscoverHandler {
...
-                .arg("--output=json")
-                .arg("--log.level=warn")
-                .arg("--log.format=color"),
+            &mut cmd,
         )
         .await?;
 
         if !discover.success() {
+            let detail = if discover_output.is_empty() {
+                format!(
```
psFried, line 159: I noticed that there was nothing in `log_lines` when this happened, and that was pretty disheartening. So this was just an unrelated improvement for that case.

crates/runtime/src/container.rs
```diff
@@ -48,7 +48,7 @@ pub async fn start(
     network: &str,
     task_name: &str,
     task_type: ops::TaskType,
-    allow_local: bool,
+    publish_ports: bool,
```
psFried, line 51: `allow_local` was a confusing name in this context, so I updated it to reflect what it actually does.

crates/agent/src/discovers.rs
```diff
@@ -124,31 +131,42 @@ impl DiscoverHandler {
...
+            .arg("--config=/dev/stdin")
+            .arg("--image")
+            .arg(&image_composed)
+            .arg("--network")
+            .arg(&self.connector_network)
+            .arg("--output=json")
+            .arg("--log.level=warn")
+            .arg("--log.format=color");
+        if self.allow_local {
+            cmd.arg("--allow-local");
```
jshearer, line 146: `flowctl-go` supports running local connectors? I thought it didn't, and that only the Rust runtime did.. 🤔

go/flowctl-go/cmd-api-discover.go
```diff
@@ -30,6 +30,7 @@ type apiDiscover struct {
 	Name        string                `long:"name" description:"The Docker container name."`
 	Config      string                `long:"config" description:"Path to the connector endpoint configuration"`
 	Output      string                `long:"output" choice:"json" choice:"proto" default:"json"`
+	AllowLocal  bool                  `long:"allow-local" description:"basically dev mode"`
```
jshearer, line 33: Ah nevermind, I see what you did here

jshearer (approved):
LGTM 👍

PR #1375, opened 2024-02-12
validation: allow unsatisfiable constraints on excluded fields

psFried:
**Description:**

Updates our valiation logic to only consider unsatisfiable constraints an error in the case that the field isn't explicitly excluded. This allows users to change the type of a field without needing to backfill, as long as they exclude the changed field.

**Workflow steps:**

- materialize a collection that has `fieldA: { type: string }`, using one of the database connectors
- Update the source collection schema to change to `fieldA: { type: integer }`
- The publication should fail due to the unsatisfiable constraint
- Explicitly exclude `fieldA` from the field selection
- Now, the publication should succeed

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1375)

### DIFF
--- a/crates/validation/src/materialization.rs
+++ b/crates/validation/src/materialization.rs
@@ -536,7 +536,8 @@ fn walk_materialization_response(
         ) {
             // Selector / connector constraints conflict internally:
             (true, true, _) => panic!("included and excluded (should have been filtered)"),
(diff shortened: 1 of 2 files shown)


williamhbaker (approved):
LGTM

PR #1377, opened 2024-02-13
Draft collections eligible for deletion

kiahna-tucker:
**Description:**

Facilitate the deletion of a capture and its associated collections in the UI by populating a specified draft with the collections eligible for deletion. A collection is eligible for deletion if it is not consumed by an active task.

**Workflow steps:**

1. Initialize a draft with a capture pending deletion.

1. Execute the `draft_collections_eligible_for_deletion` procedure, supplying the live spec ID of the capture of interest and the ID of the draft created in the previous step.

1. Observe that the collections that are not consumed by another active task in the system are added to the draft and slated for deletion.

N/A

**Notes for reviewers:**

This is a dependency of `ui` [issue 939](https://github.com/estuary/ui/issues/939). Integration of the procedure can be found in [PR 978](https://github.com/estuary/ui/pull/978).

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1377)

### DIFF
--- a/supabase/migrations/44_draft_associated_collections.sql
+++ b/supabase/migrations/44_draft_associated_collections.sql
@@ -1,0 +1,66 @@
+begin;
+
+-- Update permissions for live_spec_flows to allow users to directly access it as long as they
+-- have access to at least one side. This may allow users to know the id of specifications that
+-- they are not authorized to, but nothing else.
+alter policy "Users must be authorized to referenced specifications" on live_spec_flows
+  rename to "Users must be authorized to one referenced specification";
+
+-- This policy is tecnically a little more permissive than we truly want, though that seems
+-- acceptable since it only allows clients to get the ids of connected specs, not anything else.
+-- If you have read-only access to a spec, then you can see the ids of all connected specs.
+-- Ideally, you'd only be able to select rows for which you have admin capability to at least
+-- one side, _or_ at least read capability to _both_ sides. Such a policy seems computationally
+-- expensive and complicated to write, though, so this is a compromise.
+alter policy "Users must be authorized to one referenced specification" on live_spec_flows
+  using (
+    source_id in (select id from live_specs) or
+    target_id in (select id from live_specs)
+  );
+grant select on live_spec_flows to authenticated;
+
+create or replace function draft_collections_eligible_for_deletion(capture_id flowid, draft_id flowid)
+returns void as $$
+begin
+
+  insert into draft_specs (draft_id, catalog_name, expect_pub_id, spec, spec_type)
+  with target_collections as (
+    select target_id from live_spec_flows
+      where source_id = capture_id
+  ),
+  collections_read as (
+    select target_collections.target_id from target_collections
+      join live_spec_flows lsf on target_collections.target_id = lsf.source_id
+  ),
+  collections_written as (
+    select target_collections.target_id from target_collections
+      join live_spec_flows lsf on target_collections.target_id = lsf.target_id and lsf.source_id <> capture_id
+  ),
+  ineligible_collections as (
+    select target_id from collections_read
+      union select target_id from collections_written
+  ),
+  eligible_collection_ids as (
+    select target_id from target_collections
+      except select target_id from ineligible_collections
+  ),
+  eligible_collections as (
+    select
+    ls.id,
+    ls.catalog_name,
+    ls.last_pub_id
+    from eligible_collection_ids
+    join live_specs ls on eligible_collection_ids.target_id = ls.id
+  )
+  select draft_id, catalog_name, last_pub_id, null, null from eligible_collections;
+
+end;
+$$ language plpgsql security invoker;
+
+comment on function draft_collections_eligible_for_deletion is '
+draft_collections_eligible_for_deletion facilitates the deletion of a capture and its associated collections
+in the same publication by populating the specified draft with the collections eligible for deletion.
+The specified draft should contain the capture pending deletion.
+';
+
+commit;
--- a/supabase/tests/draft_collections_eligible_for_deletion.test.sql
+++ b/supabase/tests/draft_collections_eligible_for_deletion.test.sql
@@ -1,0 +1,168 @@
+create function tests.test_draft_collections_eligible_for_deletion()
+returns setof text as $$
+declare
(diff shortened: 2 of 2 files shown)


psFried (changes requested):
Sorry it took me so long to get to this. The migration looks good to me, though I have a few asks before merging:

- add comments (postgres `comment on function ...`) to these functions. I think the comment should mention the fact that the capture identified by the parameter must not have been deleted already.
- Add a test in `supabase/tests/`. [This test](https://github.com/estuary/flow/blob/master/supabase/tests/prune_unchanged_draft_specs.test.sql) is an example of testing a function that modifies a draft.

supabase/migrations/44_draft_associated_collections.sql
```diff
@@ -0,0 +1,46 @@
+create or replace function internal.get_collections_eligible_for_deletion(live_spec_id flowid)
```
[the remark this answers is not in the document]
kiahna-tucker: The same thought crossed my mind. Will rename.

supabase/migrations/44_draft_associated_collections.sql
```diff
@@ -0,0 +1,67 @@
...
+)
+select
+  ls.id,
+  ls.catalog_name,
+  ls.last_pub_id
+from eligible_collections
+  join live_specs ls on eligible_collections.target_id = ls.id;
+
+end;
+$$ language plpgsql security definer;
```
psFried, line 42: Now that the RLS policy on `live_spec_flows` is updated, shouldn't this be `security invoker`?

supabase/tests/draft_collections_eligible_for_deletion.test.sql
```diff
@@ -0,0 +1,147 @@
...
+  collection_e_id flowid;
+
+  capture_consumer_id flowid;
+  materialization_consumer_id flowid;
+  derivation_consumer_c_id flowid;
+  derivation_consumer_d_id flowid;
+begin
+
+  insert into user_grants (user_id, object_role, capability) values
+    ('33333333-3333-3333-3333-333333333333', 'carolCo/', 'admin');
```
psFried, line 20: I think this test setup should include a materialization/derivation that the user doesn't have access to and reads from a collection that would otherwise be eligible for deletion. This would help validate that the RLS policy changes are all working right.

psFried (changes requested):
I noticed a few more things to update before merging. LMK if you want to talk through the `security definer` issue. My understanding is that we'd either need it to be `invoker`, or else we'd need to manually check permissions in the function.

supabase/tests/draft_collections_eligible_for_deletion.test.sql
psFried: I think this still could use an assertion that does not populate the draft with specs you don't have access to, for example when calling `draft_collections_eligible_for_deletion` with the id of a capture belonging to another user. It seems like if you did:

```plpgsql
perform set_authenticated_context('22222222-2222-2222-2222-222222222222');
perform draft_collections_eligible_for_deletion(target_capture_id, target_draft_id);
```

then it should either return an error or result in no specs added to the draft. Except that `draft_collections_eligible_for_deletion` still has `security definer`, so I'd think it might actually push the `carolCo` specs into the `bobCo` draft.

supabase/migrations/44_draft_associated_collections.sql
```diff
@@ -0,0 +1,71 @@
...
+begin
+
+with eligible_collections as (
+  select * from internal.get_collections_eligible_for_deletion(capture_id)
+)
+insert into draft_specs (draft_id, catalog_name, expect_pub_id, spec, spec_type)
+  select draft_id, catalog_name, last_pub_id, null, null from eligible_collections;
+
+end;
+$$ language plpgsql security definer;
```
psFried, line 63: I think this one needs to be `invoker` too (see other comment)

psFried (changes requested):
Seems like the authZ still isn't quite right, but LMK if I'm missing something.

supabase/migrations/44_draft_associated_collections.sql
```diff
@@ -0,0 +1,74 @@
+begin;
+
+-- Update permissions for live_spec_flows to allow users to directly access it as long as they
+-- have access to at least one side. This may allow users to know the id of specifications that
+-- they are not authorized to, but nothing else.
+alter policy "Users must be authorized to referenced specifications" on live_spec_flows
+  rename to "Users must be authorized to one referenced specification";
+
+alter policy "Users must be authorized to one referenced specification" on live_spec_flows
```
jgraettinger, line 9: This policy differs from what I thought we landed at:

* I'm able to see an edge if I have **admin** to one side.
* I'm able to see an edge if I have read of **both** sides.
* I should **not** see all other tasks B, C, D that use spec A just because I, too, have read to A.

It's not the biggest deal because you can't actually get anything but the ID's.

But it **would** be a bigger deal if this auth logic were copy/pasted into a future `live_spec_flows_ext` view.

Please at least comment how this is wrong and what it should be instead.

supabase/migrations/44_draft_associated_collections.sql
```diff
@@ -0,0 +1,74 @@
...
+  using (
+    source_id in (select id from live_specs) or
+    target_id in (select id from live_specs)
+  );
+grant select on live_spec_flows to authenticated;
+
+create or replace function get_collections_eligible_for_deletion(capture_id flowid)
```
jgraettinger, line 16: Can this be inlined directly into `draft_collections_eligible_for_deletion()`, which seems like a pretty reasonable API? Then we don't have to take a position on whether this too should be public API and can refactor it later.

Or is it actively being used by the UI in the proposed feature?

psFried: It's not used. I think it came from a desire to show what _would_ be deleted, but that's not a thing that the UI does currently. I removed it.

supabase/migrations/44_draft_associated_collections.sql
psFried: :+1: commented

psFried (approved):

jgraettinger (approved):
LGTM

PR #1381, opened 2024-02-15
Send all emails to their intended recipients

jshearer:
**Description:**

This removes the `overrideEmail` change I made to redirect all non-data processing emails to myself to make sure we don't spam people. I'm leaving in the `bcc` to myself just to monitor emails for a while longer, though plan to remove that soon as well. 

Also removes no-longer used `alert-data-processing` edge function, as that logic is handled by the more general-purpose `alerts` function.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1381)


psFried (approved):
LGTM

PR #1388, opened 2024-02-20
Phil/agent logs

psFried:
**Description:**

estuary/flow#1379 changed our `tracing_subscriber` configuration to log at the close of each span. That required updating some `#[instrument]` attributes to include `level = "debug"`, since the default level of spans is `INFO`. Apparently I missed a couple, so here we are. Production is currently logging pretty noisily, and this should help tone it down.

Also, I added `time_queued` log field to each async job handler, so we can get a sense of end-to-end processing latency. In retrospect, I wish I would have done this _before_ making all these improvements to it, but 🤷

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1388)

### DIFF
--- a/crates/agent-sql/src/drafts.rs
+++ b/crates/agent-sql/src/drafts.rs
@@ -20,7 +20,7 @@ pub async fn create(
     Ok(row.id)
 }
 
-#[tracing::instrument(err, skip(spec, txn))]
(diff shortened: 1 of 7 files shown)


jgraettinger (approved):
LGTM

PR #1389, opened 2024-02-21
update a number of dependencies, including RocksDB (to 8.10)

jgraettinger:
Run a general `cargo update`.

Update our rust-rocksdb to rebase on their most-recent release. All but one of changes have been merged upstream.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1389)

### DIFF
--- a/crates/runtime/src/container.rs
+++ b/crates/runtime/src/container.rs
@@ -535,10 +535,14 @@ mod test {
             }]
         );
 
-        #[cfg(target_os = "linux")]
-        assert_eq!(container.mapped_host_ports, super::BTreeMap::new());
-        #[cfg(not(target_os = "linux"))]
-        assert_ne!(container.mapped_host_ports, super::BTreeMap::new());
+        assert_eq!(
+            container
+                .mapped_host_ports
+                .keys()
+                .copied()
+                .collect::<Vec<_>>(),
+            vec![8080, 49092]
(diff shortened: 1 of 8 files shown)


crates/runtime/src/container.rs
```diff
@@ -535,10 +535,14 @@ mod test {
             }]
         );
 
-        #[cfg(target_os = "linux")]
-        assert_eq!(container.mapped_host_ports, super::BTreeMap::new());
-        #[cfg(not(target_os = "linux"))]
-        assert_ne!(container.mapped_host_ports, super::BTreeMap::new());
+        assert_eq!(
```
psFried, line 538: What's this about? Seems like we used to assert that `mapped_host_ports` was empty on linux, and now we'd assert that it's non-empty regardless of OS. Why?

psFried (approved):
LGTM % one question about a change that seems maybe unintentional

crates/runtime/src/container.rs
jgraettinger: It was broken by the recent change to make publish_ports a thing. This test doesn't typically run in CI because `flowctl` isn't built, but does run locally.

This used to correspond to the behavior of port mapping when we used target_os.

PR #1398, opened 2024-02-28
Updating SQL Server wording to correctly talk about which databases we support

dyaffe:
**Description:**

Updating SQL Server wording to correctly talk about which databases we support since we support earlier than 2017.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1398)


jonwihl (commented):
LGTM

jonwihl (approved):

PR #1403, opened 2024-03-05
runtime: publish uptime stats using `usageRate` from image label

psFried:
**Description:**

The immediate goal is to allow certain connectors to be free, in the sense that they do not incur any charges based on uptime. The TLDR is that you can now add `LABEL dev.estuary.usage-rate=0` on any connector `Dockerfile` in order to make it free.

The uptime charges are based on the number of seconds that the container is running, multiplied by the `usageRate`. Previously, the `usageRate` was determined based on the task type. Derivations had a `usageRate` of 0 (making them free), and it was 1 for all other task types. That's is still the default after this commit. But now the `usageRate` can be overridden by a `dev.estuary.usage-rate` label on the connector container. This intentionally allows for values other than 0 or 1, so that connectors can have basically any pricing.

This required a bit of refactoring of how we start emitting uptime stats. Previously, we started to emit uptime stats before actually starting the container. This has been changed so that we now wait until the container is started and the `usageRate` is determined before starting to emit uptime stats.

I also did a bit of refactoring in `container.rs` in order to bring consistency to the behavior of the `start` and `flow_runtime_protocol` functions. When we `start` a container, we first inspect the image and parse the output in order to determine which network ports to expose and (now) what the `usageRate` is. The `flow_runtime_protocol` function is used by the `connector_tags` job when installing a connector. Those two functions now use the same logic for parsing the image inspection. This allows us to fail fast on `connector_tags` jobs for images with missing or invalid labels. As part of this, I also removed the backward compatibility logic for determining the runtime protocol, which means that the `FLOW_RUNTIME_PROTOCOL` label is now required to be present on all images. I've confirmed that the label is present on all 124 `connector_tags` in the system, so it seems safe to do so.

**Workflow steps:**

No user-facing changes.
For connector developers, add `LABEL dev.estuary.usage-rate=0` to the `Dockerfile` in order to make it free.

**Notes for reviewers:**

I chose to stop short of renaming `usageSeconds` in `ops-catalog-stats-schema.json`, since our only immediate need is to set `usageRate` to 0. But it does seems like we may want to do something to indicate that the value is more like an abstract "credits" rather than a literal number of seconds, since that will be the case as soon as we determine a `usageRate` that isn't 0 or 1.

The related connectors PRs are estuary/connectors#1321 and estuary/airbyte#293

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1403)


psFried:
One other thing I wanted to call out explicitly is that the `usageRate` can only be taken from image labels for image connectors. For all other types, like sqlite derivations, or `local` connectors, the `usageRate` defaults to 0. This makes no difference as things are today, since derivations default to 0 `usageRate` anyway, and we don't run `local` connectors in production. But something to keep in mind as things change.

go/runtime/task.go
```diff
@@ -111,6 +108,19 @@ func newTaskBase[TaskSpec pf.Task](
...
 
+// StartTaskHeartbeatLoop begins a long-lived task which writes stats at regular intervals,
+// and then logs the final exit status of this shard. The `usageRate` is the number of
+// "credits" used per second by the connector.
+func (t *taskBase[TaskSpec]) StartTaskHeartbeatLoop(shard consumer.Shard, container *pr.Container) {
+	var usageRate float32 = 0.0
+	if container != nil {
+		usageRate = container.UsageRate
+	} else {
+		logrus.WithField("shard", shard.Spec().Id).Info("usageRate will be 0 because there is no container")
```
williamhbaker, line 119: When can `container` be `nil`?

williamhbaker (approved):
LGTM

go/runtime/task.go
psFried: Ah, that's what I was trying to get at with [this comment](https://github.com/estuary/flow/pull/1403#issuecomment-1981746981), which wasn't very clear. If the connector type is anything but `Image`, then `container` will be nil. We may need to revisit this if we ever want to charge for other connector types, but for now those are just defaulted to a 0 `usageRate`.

PR #1410, opened 2024-03-11
models/journals: region configuration for S3 storage mappings

williamhbaker:
**Description:**

Allows setting the region for S3 storage mappings, for buckets that are not in the `us-east-1` region.

See also: https://github.com/gazette/core/pull/365. We will need to be running on this version of Gazette prior to deploying the changes here.

Closes https://github.com/estuary/flow/issues/1409

**Workflow steps:**

Create a storage mapping like this, and it will now work:

```json
{
  "stores": [
    {
      "provider": "S3",
      "bucket": "my-bucket-in-us-west-2",
      "region": "us-west-2",
      "prefix": "collection-data/"
    }
  ]
}
```

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1410)

### DIFF
--- a/crates/models/src/journals.rs
+++ b/crates/models/src/journals.rs
@@ -41,8 +41,8 @@ impl GcsBucketAndPrefix {
 }
 
 #[derive(Serialize, Deserialize, Clone, Debug, JsonSchema, Validate)]
-#[schemars(example = "S3BucketAndPrefix::example")]
-pub struct S3BucketAndPrefix {
+#[schemars(example = "S3StorageConfig::example")]
+pub struct S3StorageConfig {
     /// Bucket into which Flow will store data.
     #[validate(regex = "S3_BUCKET_RE")]
     pub bucket: String,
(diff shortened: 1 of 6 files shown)


crates/models/src/journals.rs
```diff
@@ -51,24 +51,35 @@ pub struct S3BucketAndPrefix {
...
+        let mut u = url::Url::parse(&format!(
             "s3://{}/{}",
             self.bucket,
             self.prefix.as_deref().unwrap_or("")
         ))
-        .expect("parsing as URL should never fail")
+        .expect("parsing as URL should never fail");
+
+        u.query_pairs_mut()
+            .append_pair("region", self.region.as_deref().unwrap_or("us-east-1"));
```
psFried, line 73: I'm wondering if we can avoid hard coding `us-east-1` as the default. What happens if we leave it unspecified? Would that cause the default to be whatever is in the aws config file on the brokers? If so, then it kinda seems preferable.

psFried (commented):
LGTM % question about the default

crates/models/src/journals.rs
williamhbaker: Yep it would use the default from the AWS config of the brokers. I don't have a compelling reason not to just do that, so I've made the update.

psFried (approved):
Thanks, LGTM

PR #1430, opened 2024-04-02
Fix BumpVec test after change to u128 alignment

psFried:
**Description:**

There was a recent change in the rust compiler that changed the aligmnent of u128 values. See: https://blog.rust-lang.org/2024/03/30/i128-layout-update.html This updates a BumpVec test that had asserted the old values, so that the test now hopefully passes.

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1430)


jgraettinger (approved):
LGTM

PR #1432, opened 2024-04-02
flowctl: resurrect stats subcommand under raw

psFried:
`flowctl stats` was removed a few months ago due to concerns about UX. I've really missed having it, though, because the process for finding and reading from the stats journal using `flowctl-go` and port-forwarding is a pain. So this adds the subcommand back in, but moves it under `flowctl raw` to address the UX issue.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1432)


jgraettinger (approved):
LGTM

PR #1438, opened 2024-04-09
agent: backend support for updating storage mappings

psFried:
**Description:**

Introduces two different pieces of backend functionality in support of a UI for updating storage mappings. The first is a directive for updating the actual storage mappings, including validating the new mapping by ensuring we have appropriate permissions to the bucket. The second is a postgres function that creates a publication of all the existing specs under a given prefix.

Only top-level tenant prefixes are supported at this time. It shouldn't be too hard to support adding new prefixes in the future, but it seemed wise to avoid the complexity for now.

The directive is meant to validate our access to the users storage buckets. It only supports GCS and S3 at this time, but it shouldn't be too hard to add support for azure and even custom storage endpoints. The checks are done by putting a file, listing it, and deleting it. This is somewhat more crude than checking our users permissions, but it works for any cloud provider and ensures that we're only requiring a minimal set of permissions.

The `republish_prefix` function is needed in order to apply the newly updated storage mappings. Stitching this together with the directive is being left as a UI responsibility for now.

**Workflow steps:**

The basic steps are (assume `--profile local` if running locally):

- `flowctl raw rpc --function exchange_directive_token --body '{"bearer_token": "<directive-token>"}'
- `update applied_directives set user_claims = '{"addStore": {"provider": "S3", "bucket": "estuary-test"}, "catalogPrefix": "phil/"}' where id = '<id-from-rpc>'`
- Observe applied directive status and storage mappings
- If directive was successful, `flowctl raw rpc --function republish_prefix --body '{"prefix": "phil/"}'` to publish everything under the tenant

I ran through these steps and tested a few different buckets and scenarios:

- add GCS bucket with and without a prefix
- add an S3 bucket with and without a prefix
- add an S3 bucket with an explicit region
- add a mapping that's already in the historical mappings, and ensure that it's not duplicated (should only be at index 0)

Some sad-path scenarios:

- catalog prefix in user claims is one that the user doens't admin results in `invalidClaims`
- catalog prefix that is not a top-level tenant prefix results in `invalidClaims`
- republish_prefix called with prefix that user doesn't have read access to (results in publication of empty draft)

None yet, will document once we have UI done.

**Notes for reviewers:** n/a

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1438)

### DIFF
--- a/crates/agent/src/directives/storage_mappings.rs
+++ b/crates/agent/src/directives/storage_mappings.rs
@@ -1,0 +1,294 @@
+use std::io::Write;
+
+use crate::{directives::JobStatus, logs};
+use agent_sql::directives::{
+    storage_mappings::{
+        fetch_storage_mappings, upsert_storage_mapping, user_has_admin_capability, StorageMapping,
+    },
+    Row,
+};
+use anyhow::Context;
+use serde::{Deserialize, Serialize};
+use sqlx::types::Uuid;
+use validator::Validate;
+
+#[derive(Debug, Deserialize, Serialize, Validate, schemars::JsonSchema)]
+#[serde(rename_all = "camelCase", deny_unknown_fields)]
+pub struct Directive {}
+
+#[derive(Debug, Deserialize, Serialize, Validate, schemars::JsonSchema)]
+#[serde(rename_all = "camelCase", deny_unknown_fields)]
+pub struct Claims {
+    add_store: models::Store,
+    catalog_prefix: models::Prefix,
+}
+
+#[tracing::instrument(skip_all, ret, err, fields(row.user_claims))]
+pub async fn apply(
+    _: Directive,
+    row: Row,
+    logs_tx: &logs::Tx,
+    txn: &mut sqlx::Transaction<'_, sqlx::Postgres>,
+) -> anyhow::Result<JobStatus> {
+    let detail = format!(
+        "updated by user {} via applied directive {}",
+        row.user_id, row.apply_id
+    );
+    let (collection_data, recovery) = match validate(txn, logs_tx, row).await {
+        Ok(c) => c,
+        Err(err) => {
+            return Ok(JobStatus::invalid_claims(err));
+        }
+    };
+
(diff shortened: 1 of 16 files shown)


crates/agent/src/directives/storage_mappings.rs
```diff
@@ -0,0 +1,294 @@
...
+        anyhow::bail!("invalid prefix, must end with '/'");
+    }
+
+    // Storage mappings can only be updated for an entire tenant. We may one day wish to support
+    // updates to narrower prefixes, but are trying to keep it simple for now.
+    if claims.catalog_prefix.matches('/').count() > 1 {
+        anyhow::bail!("catalog prefix contains too many path segments. Only top-level tenant prefixes can have storage mappings altered");
+    }
+
+    // Note: we must assert that user has admin capability for the _entire tenant_, even if in the
```
jgraettinger, line 88: This constraint is fine for now, but I suspect it's too strict and that it is desired to let a user who admins `a/b/` to be able to update the storage mapping for that prefix.

crates/agent/src/directives/storage_mappings.rs
```diff
@@ -0,0 +1,294 @@
...
+                "put object",
+                check_command!(
+                    "aws",
+                    "s3",
+                    "cp",
+                    test_file_path,
+                    without_query(conf.as_url()).join(TEST_FILENAME)?.to_string()
+                    ; region conf.region.as_ref()),
+            ),
+            // List comes after put, so the prefix, if configured, is guaranteed to exist.
```
jgraettinger, line 166: Could we add a `get` as well here? IIRC we've actually seen this misconfiguration in the past, where a user had granted list but not get.

jgraettinger (approved):
LGTM % get access

crates/agent/src/directives/storage_mappings.rs
psFried: yep, done

PR #1439, opened 2024-04-12
supabase: add `registered_avro_schemas` table

jgraettinger:
This table provides a global namespace for u32 schema IDs which appear within encoded Kafka keys and values.

IDs are integrated into our catalog authorization model, so one is still only allowed to inspect a schema to which one is read authorized, but the table still facilitates deduplication of common schema shapes.

To register or retrieve a schema, a user must only be read-authorized to its corresponding catalog name. This means that all readers of a shared collection are able to re-use the same schema ID.

Testing:
 * I've performed end-to-end testing on a local stack with `dekaf`, which uses the table for registering content-addressed AVRO schemas. PR for that forthcoming.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1439)

### DIFF
--- a/supabase/migrations/47_registered_avro_schemas.sql
+++ b/supabase/migrations/47_registered_avro_schemas.sql
@@ -1,0 +1,32 @@
+
+create table registered_avro_schemas (
+  like internal._model including all,
+
+  avro_schema      json not null,
+  avro_schema_md5  text generated always as (md5(trim(avro_schema::text))) stored,
+  catalog_name     catalog_name not null,
+  registry_id      serial unique not null
+);
+
+create index idx_registered_avro_schemas_avro_schema_md5 on registered_avro_schemas (avro_schema_md5);
+
+comment on table registered_avro_schemas is '
+Avro schemas registered with a globally unique, stable registery ID.
+
+This is used to emulate the behavior of Confluent Schema Registry when
+transcoding collection documents into Avro for use with Dekaf,
+which must encode each message with an Avro schema ID (registry_id).
+';
+
+alter table registered_avro_schemas enable row level security;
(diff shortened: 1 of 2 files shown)


supabase/migrations/47_registered_avro_schemas.sql
```diff
@@ -0,0 +1,32 @@
...
+
+create policy "Users must be read-authorized to the schema catalog name"
+  on registered_avro_schemas as permissive
+  using (exists(
+    select 1 from auth_roles('read') r where catalog_name ^@ r.role_prefix
+  ));
+
+grant select on registered_avro_schemas to authenticated;
+grant insert (catalog_name, avro_schema) on registered_avro_schemas to authenticated;
+grant update (updated_at) on registered_avro_schemas to authenticated;
```
psFried, line 31: This seems weird. Why would we want users to update only the `updated_at` column? Seems like it should include the schema, too?

psFried (approved):
LGTM % one question

supabase/migrations/47_registered_avro_schemas.sql
jgraettinger: Registered schemas need to be immutable -- once a registry ID is encoded on the wire, data has been coded to the schema and the recipient needs to be able to retrieve the exact same schema in order to reliably decode it.

However, it would be really nice to have a measure of "which of these schemas aren't actually used anymore?" so that we could expunge them / reclaim portions of the registry_id space, and that was the role I figured `updated_at` could play.

jgraettinger:
Thanks!

PR #1446, opened 2024-04-19
examples: add an example of outer join using a sql derivation

psFried:
**Description:**

Adds an example of doing an outer join using a sql derivation. We've been asked about this several times, so I thought it would help to have an example we can link to.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1446)


jonwihl (approved):
LGTM

PR #1453, opened 2024-04-29
docs: Sort connector pages alphabetically

danthelion:
**Description:**

This PR adds a custom sort function to Docusaurus which enables the sorting of "category" and "doc" type objects on the same level. The default sorting mechanism sorts only docs (including the ones nested in categories) which causes the order to be not truly alphabetical.

[Both capture & materialization connector](https://docs.estuary.dev/reference/Connectors/) lists are affected

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1453)


jonwihl (approved):
LGTM

PR #1459, opened 2024-05-06
docs: Add CloudSQL CDC to Snowflake tutorial

danthelion:
**Description:**

Add new tutorial

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1459)


jgraettinger:
This looks good, but two things I'm noting:

* I think the source database setup section should be more of an appendix, and Step 1 should be setting up the capture from your DB. That step 1 can say something to the effect of "if you'd like a walkthrough of setting up a test database, jump to here". The reason is that it's an intimidating wall of setup steps, and we want to make it abundantly clear that that's not inherent complexity of using Flow and is just prototyping setup.

* The CREATED_AT dates in the table screen shot are all over the place and look wrong, like there's a data correctness error. What's going on there?

danthelion:
Closing and moving this to the blog, rather than the docs page

PR #1473, opened 2024-05-13
docs: Add spreadsheet formatting to source-google-sheets

Luishfs:
**Description:**

Added some formatting rules for google sheets


jonwihl (changes requested):
Can we change the language around 3.1? I think we can remove it all together since nothing will happen if theres an empty sheet

PR #1489, opened 2024-06-12
Phil/cpv2

psFried:
**Description:**

Rolls up whole bunch of work from both @jgraettinger and myself related to:

- publications
    - builds
    - validation
    - activation
- inferred schema handling
- materialization schema evolution (`onIncompatibleSchemaChange`)
- controllers :sparkles: 

I recommend reviewing this commit by commit, and tried to write helpful commit messages. Here's a high-level summary:

Reframe validation to be an operation over a `tables::LiveCatalog` (a snapshot from the control plane) and a `tables::DraftCatalog`, which is able to represent deletions.

Updates the publications handler to:
- work with the new `validation` crate changes
- use optimistic locking instead of holding row-level locks
- setup for a future `publish` function that can be invoked by controllers
- use a more minimal spec expansion
- trigger controller runs after a successful publication

Introduces controllers, which are background automations that are tied to specific live specs. Controllers are responsible for:

- updating shard and journal specs in the data plane, in response to a publication
- keeping specs up-to-date with respect to their dependencies
- updating collection inferred schemas (inferred schemas are now added directly to the collection spec)
- updating materailizations in response to incompatible schema changes
- updated materializations in response to changes of their `sourceCapture`
- maintaining a status that can provide some basic information about their operation and any errors that occur

**Workflow steps:**

The control plane should work more or less as it has in terms of user-facing operations, with a few exceptions:

- You can now delete and re-create collections and tasks with the same name :tada: 
- Inferred schemas now work with derivations
- Inferred schemas now get baked into the collection model. This means that you can technically now overwrite the published inferred schema with flowctl. This might be easy to do, since it's hard to know if the inferred schema has changed, and your local spec is now stale.
- Materializations now obey the `onIncompatibleSchemaChange` property (top-level, or can be overridden on a per-binding basis)
- Users will need to update flowctl in order for things to work with the new control plane. One reason for this is that the `spec_type` column on `live|draft_specs` is now non-nullable.
- Activations now happen asynchronously, after the publication itself has already succeeded. Thus there is not a possibility of a slight delay after publication before the new shard status can be shown on the "details" page. This means a user can publish a task that was failing, and then get redirected to the details page and possibly be confused by the shard status still showing as failed. One way to address this might be to have the details page query shard status a little more frequently. We also have the information available in `controller_jobs.status->activation->last_activated`, which can be compared agains `live_specs.last_pub_id` to see if the activation has been completed yet.
- We no longer return `JobStatus::TestsFailed` when a test run fails. Instead it just returns `BuildFailed` (the can be fixed but I wasn't sure if it was important).

**Notes for reviewers:**

I'd appreciate any time you have to test this yourself locally, and report any issues. I'd like to continue testing, myself, particularly on the transition from the old version, and I expect I'll need to push some updates before this is ready to deploy. I'd like to keep this PR open until we feel this is in a deployable state.

Please reach out by any means with any questions.

### Deployment plan

The basic steps for deploying this are the same as any other control plane PR involving a database migration, except that there's an additional migration to be run _after_ the deployment finishes. This second migration will actually need to be run multiple times, because it will incrementally enable controllers for pre-existing live specs, in batches of 1000 at a time. This gives us a chance to bail out if things aren't working right, instead of just unleashing controllers on all of our precious live specs. The end-to-end high level plan is thus:

- Run the `50_controllers.sql` migration, previous agent versions will continue to run against the new db schema
- Deploy new agents, just using the normal ops repo
- Verify that we can publish a new task and edit an existing one through the UI, verify controller runs in `controller_jobs`
- Run the `51_backfill_controllers.sql` migration, and spot check the controller runs to make sure things look good
- Repeat the previous step until we've backfilled all of them
- Profit

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1489)


jgraettinger:
>  This means that you can technically now overwrite the published inferred schema with flowctl. This might be easy to do, since it's hard to know if the inferred schema has changed, and your local spec is now stale.

`flowctl` now adds `expectPubId` to any live specs it pulls down from the control plane. `expectPubId` of zero means "must not exist".

This is to prevent a user from inadvertently overwritting an updated live spec by publishing a stale spec from their checkout. They'll get an error because expectPubId will disagree.

They'd have to explicitly remove the expectPubId field to denote that they _want_ their local version to take precedence.

`flowctl` tries not to publish a spec which isn't actually changing, so they should be good to leave unmodified specs as-in in a local checkout while publishing a managed spec over-and-over.

crates/tables/src/built.rs
```diff
@@ -232,3 +234,88 @@ impl BuiltRow for crate::BuiltTest {
...
+            )
+            .chain(
+                self.built_materializations
+                    .iter()
+                    .map(|r| r.catalog_name().as_str()),
+            )
+            .chain(self.built_tests.iter().map(|r| r.catalog_name().as_str()))
+    }
+
+    pub fn spec_count(&self) -> usize {
```
jgraettinger, line 266: `all_spec_names().count()` ?

crates/tables/src/draft.rs
```diff
@@ -1,6 +1,73 @@
...
+            .chain(self.collections.iter().map(|c| c.catalog_name().as_str()))
+            .chain(
+                self.materializations
+                    .iter()
+                    .map(|m| m.catalog_name().as_str()),
+            )
+            .chain(self.tests.iter().map(|t| t.catalog_name().as_str()))
+    }
+
+    pub fn to_models_catalog(&self) -> models::Catalog {
```
jgraettinger, line 45: this isn't called that I can find

jgraettinger (commented):
Some initial feedback. I've gotten up to (but haven't started) agent: publications redux

crates/models/src/lib.rs
```diff
@@ -42,18 +44,245 @@ pub use references::{
...
+    fn catalog_type(&self) -> CatalogType {
+        match self {
+            AnySpec::Capture(_) => CatalogType::Capture,
+            AnySpec::Collection(_) => CatalogType::Collection,
+            AnySpec::Materialization(_) => CatalogType::Materialization,
+            AnySpec::Test(_) => CatalogType::Test,
+        }
+    }
+
+    fn sources(&self) -> impl Iterator<Item = &Source> {
```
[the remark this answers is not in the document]
psFried: ah, that needs a comment, and probably should be replaced with a `panic!`. The reason for this, and for `reads_from/writes_to`, is that this union type doesn't work well with `impl Iterator<Item = &_>`. `AnySpec` can't easily abstract over the different concrete types of the iterators returned by the various `models::_Def`. It can be done, but didn't seem worth it just yet, because nothing actually calls those functions on this type.

Ideally, we'd unify `sources` and `reads_from`, and so I was also thinking perhaps the borrowed iterator version might go away. I think even that unification can wait a bit until we get a better idea of what else we might do with controllers, though. If, for example, we start adding more types of dependencies, then it might be better still to go with something like `fn dependencies(&self) -> Vec<(String, FlowType)>`.

So I'll propose adding an explicit `unimplemented!` note there for now and kick the can a little.

crates/agent-sql/src/live_specs.rs
```diff
@@ -0,0 +1,162 @@
...
+            "#,
+        collections as &[&str],
+    )
+    .fetch_all(pool)
+    .await
+}
+
+/// Queries for all non-deleted `live_specs` that are connected to the given `collection_names` via
+/// `live_spec_flows`.
+pub async fn fetch_expanded_live_specs(
```
jgraettinger, line 114: This routine is run only on user-initiated publications, as I understand :+1:

One note is we still need the recursive resolution behavior of resolve_expanded_rows() as well, but now we only need it to pull in live and unchanged derivations which are attached to any drafted tests.

We still need to recursively walk such derivations to pull out the full data flow cycle being put under test.

crates/agent/src/publications/status.rs
```diff
@@ -0,0 +1,193 @@
...
+
+/// Reasons why a draft collection spec would need to be published under a new name.
+#[derive(Serialize, Deserialize, Clone, Copy, PartialEq, Debug)]
+#[serde(rename_all = "camelCase")]
+pub enum ReCreateReason {
+    /// The collection key in the draft differs from that of the live spec.
+    KeyChange,
+    /// One or more collection partition fields in the draft differs from that of the live spec.
+    PartitionChange,
+    /// A live spec with the same name has already been created and was subsequently deleted.
```
jgraettinger, line 93: nit: remove?

supabase/migrations/50_controllers.sql
```diff
@@ -0,0 +1,127 @@
+begin;
```
jgraettinger, line 1: (meta: stopping here for now)

jgraettinger (commented):
More comments! thanks

crates/agent-sql/src/live_specs.rs
psFried: I wasn't thinking we needed to do that. The expansion logic here applies to the draft, which will subsequently go though another round of resolution to pull in all the collections referenced by tasks/tests in the draft. So this function expands from published collections to any tests/tasks that depend on them, and then the subsequent `live_specs` resolution will pull in collection specs that those tests need in order to run. I really should have more explicit tests for this, which might help to illustrate that better. But I'm not 100% sure I've got your point. Are you saying that we should try to expand through an arbitrary number of derivations until we get to the captures/materializations "edges"? _That_ would be pretty different from what's here.

crates/agent-sql/src/publications.rs
```diff
@@ -1,10 +1,165 @@
...
+                built_spec,
+                $1::flowid,
+                $1::flowid,
+                now(),
+                case when json_typeof(reads_from) is null then
+                    null
+                else
+                    array(select json_array_elements_text(reads_from))
+                end,
+                case when json_typeof(reads_from) is null then
```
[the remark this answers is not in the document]
psFried: :grimacing: geeze good catch. I'm ... gonna try adding a test that covers this

crates/agent-sql/src/live_specs.rs
psFried: BTW the best spec expansion test we have right now is [`user_publications.rs`](https://github.com/estuary/flow/blob/2570da4f8b7b028da9f45f058159189f2b06377e/crates/agent/src/integration_tests/user_publications.rs#L167).

supabase/migrations/50_controllers.sql
```diff
@@ -0,0 +1,127 @@
...
+alter table live_specs alter column spec_type set not null;
+
+-- Allow spec_type to remain non-null for deleted specs
+alter table publication_specs drop constraint "spec and spec_type must be consistent";
+
+
+create table controller_jobs (
+    -- The name of the live spec that this pertains to
+    live_spec_id flowid not null references live_specs (id) on delete cascade,
+    -- The version of the controller that last updated this row. Used to identify controllers to run
```
jgraettinger, line 59: 🤔 how does this impact query performance of the controller? If we're querying for rows where the controller version is less than our current version, that would need to be indexed or it'll require a full table scan (i haven't gotten to this code yet; or is it an output-only column?).

If the goal is to be able to trigger controllers on software releases, it seems like we could as-easily do that through a migration that tickles `controller_next_run` on `live_specs` -- that might even be better because we could set up a schedule that staggers / slow-rolls the controller runs.

crates/activate/src/lib.rs
```diff
@@ -524,24 +536,41 @@ mod test {
         ),)
     }
 
-    #[tokio::test]
-    async fn fixture_subtests() {
-        let source = build::arg_source_to_url("./src/test.flow.yaml", false).unwrap();
+    async fn managed_build(source: url::Url) -> build::Output {
```
jgraettinger, line 541: this was to get rid of build::managed_build ? Was it the last remaining usage?

This test introduced a pattern we've had in Go for a while, of doing an actual catalog build in service of establishing a test case fixture. We do want to keep that path as easy to walk as possible, perhaps by adjusting managed_build's interface if there are no other non-test usages.

jgraettinger (commented):
A few more comments, about to dig into `agent: introduce controllers`.

crates/agent-sql/src/controllers.rs
```diff
@@ -0,0 +1,172 @@
...
+            select id as live_spec_id
+            from live_specs
+            where
+                -- This condition is required in order to for this query to use the sparse index
+                controller_next_run is not null
+                and controller_next_run <= now()
+            order by controller_next_run asc
+        ),
+        needs_upgrade(live_spec_id) as (
+            select live_spec_id
```
jgraettinger, line 42: As noted in a prior comment -- this is gonna do a full table scan every time.

I'm coming to believe that `controller_version` isn't buying us much and should be removed.
* We can always have migrations tweak controller_next_run if we want to do a mass-run of controllers because of a software release.
* The remaining use case for `controller_version` is understanding how to interpret a status (is it the old or new version), but we have lots of experience doing incremental updates of JSON specifications and we've never before needed a version to do so.

It seems like `controller_version`'s utility can be gotten at through other mechanisms and so it should be dropped. If that's wrong and we do find we need it, it's also easy to add it as a future migration with a default.

crates/models/src/lib.rs
```diff
@@ -42,18 +44,245 @@ pub use references::{
...
+    pub fn catalog_type(&self) -> CatalogType {
+        match self {
+            AnySpec::Capture(_) => CatalogType::Capture,
+            AnySpec::Collection(_) => CatalogType::Collection,
+            AnySpec::Materialization(_) => CatalogType::Materialization,
+            AnySpec::Test(_) => CatalogType::Test,
+        }
+    }
+
+    pub fn as_capture(&self) -> Option<&CaptureDef> {
```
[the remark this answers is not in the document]
psFried: The problem is that there's just _a lot_ of this type of "trivial" conversion that needs to be done, and it adds up to a ton of "noise" if we do all of them at each call site. I'm open to changing or removing these, but would ask that we bikeshed for now and take it up in a subsequent PR.

crates/models/src/lib.rs
```diff
@@ -42,18 +44,245 @@ pub use references::{
...
+    /// Source collections read by this specification.
+    /// Disabled bindings must be excluded from the iterator.
     fn sources(&self) -> impl Iterator<Item = &Source>;
-    // Target collections written to by this specification.
+    /// Target collections written to by this specification.
+    /// Disabled bindings must be excluded from the iterator.
     fn targets(&self) -> impl Iterator<Item = &Collection>;
+
+    /// Returns the `sources` of this spec as an owned set.
+    fn reads_from(&self) -> BTreeSet<Collection> {
```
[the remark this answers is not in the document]
psFried: I got rid of the separate impls, which were a holdover from before consolidating `SpecExt`. Basically, `SpecExt` and `ModelDef` were developed concurrently, and one used owned `Collection`s while the other used borrowed `&Source`s. Ideally, we'd consolidate into a single function that works for everyone, but I don't yet have a clear picture of what that should look like. See [this comment](https://github.com/estuary/flow/pull/1489#discussion_r1643403689) for more deets.

crates/agent-sql/src/publications.rs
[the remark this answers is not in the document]
psFried: Fixed and added some test assertions in [this commit](https://github.com/estuary/flow/pull/1489/commits/b998cd3796035f7077cf5f8a53dafd21312b83cb)

crates/agent-sql/src/id.rs
```diff
@@ -5,6 +5,17 @@ use sqlx::{postgres, Decode, Encode, Type, TypeInfo};
 #[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub struct Id([u8; 8]);
```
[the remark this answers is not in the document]
psFried: As it is now, we still need separate copies of `Id` and `CatalogType` because `models` doesn't depend on `sqlx` and thus can't `impl sqlx::Type`. I'm half convinced that we should add an optional `sqlx` dependency (and feature) to `models` in order to avoid this duplication and having conversions sprinkled everywhere, but figured we could address it in a subsequent PR.

crates/tables/src/draft.rs
psFried: removed

crates/tables/src/built.rs
psFried: done

supabase/migrations/50_controllers.sql
```diff
@@ -0,0 +1,127 @@
...
+  strict superset of catalog_spec_type, for historical reasons.';
+
+-- This works because the `flow_type` enum is a superset of the `catalog_spec_type` enum.
+-- This approach was taken from:
+-- https://www.munderwood.ca/index.php/2015/05/28/altering-postgresql-columns-from-one-enum-to-another/
+alter table live_spec_flows
+alter column flow_type set data type flow_type
+using flow_type::text::flow_type;
+
+-- We now allow live specs to be deleted even though they're still referenced by other specs.
```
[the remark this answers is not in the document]
psFried: Originally this was here because controllers would hard delete `live_specs` and then call `notify_dependents`, so the `live_spec_flows` were still needed in order to identify the dependent specs to notify. But now, the deletion of live specs happens _after_ the call to `notify_dependents`, and so I would think that we can keep these constraints and add `on delete cascade`. I'll give it a go

crates/agent-sql/src/controllers.rs
psFried: I think `controller_version` is still needed in order to allow for agent rollouts. When we deploy a new agent version (that has an incremented `controller_version`), there will still potentially be running agent processes having an older `controller_version`. Part of the purpose of `controller_version` is to allow agents to filter controller jobs so that they don't run for rows that have already been upgraded.

TBH I was thinking that a full table scan might not be a problem in the short term. In practice, most full table scans on `live_specs` seemed to take somewhere in the ~45ms range, and I'm assuming that `controller_jobs` will be roughly similar or maybe slightly faster (fewer columns). But I haven't tested the `deque` performance under realistic conditions.

If we want to avoid the full table scans, I can think of a few approaches:
- Add an index on `controller_version`
- execute a single query when the agent starts up, which sets `controller_next_run = now()` for all rows having `controller_version < $agent_controller_version`

From my perspective, I'd _really_ like to avoid needing to do a database migration when updating the controller version. And I don't think it'd really work, given the possibility for multiple agent versions to be running simultaneously during rollouts.

crates/agent/src/publications.rs
```diff
@@ -1,623 +1,365 @@
...
+    pub completed_at: DateTime<Utc>,
+    pub draft: tables::DraftCatalog,
+    /// The state of any related live_specs, prior to the draft being published
+    pub live: tables::LiveCatalog,
+    /// The build specifications that were output from a successful publication.
+    /// Will be empty if the publication was not successful.
+    pub built: tables::Validations,
+    /// The final status of the publication. Note that this is not neccessarily `Success`,
+    /// even if there are no `errors`.
+    pub publication_status: JobStatus,
```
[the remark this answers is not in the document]
psFried: done

crates/agent-sql/src/publications.rs
```diff
@@ -1,10 +1,165 @@
...
 use serde_json::value::RawValue;
 use sqlx::types::Uuid;
 
+#[derive(Debug)]
+pub struct LiveRevision {
+    pub catalog_name: String,
+    pub last_pub_id: Id,
+}
+
+/// Locks the given live specs rows and returns their current revisions. This is used for verifying the revisions
```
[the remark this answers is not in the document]
psFried: done

crates/activate/src/lib.rs
psFried: Yeah, this was to get rid of `build::managed_build`, as this was the last remaining usage. We're agreed that we want to keep an easy way to do builds of test fixtures. Perhaps it'll make sense to pull this back over into `build::managed_build` with a `#[cfg(test)]`, but there's also the option of using `Publisher::build` in tests. Both would have a few minor peculiarities, and I feel like I'd have a stronger opinion after seeing the needs of the tests, which is why I stuffed this in here for now.

crates/agent/src/draft.rs
```diff
@@ -1,16 +1,49 @@
...
+            // use alternate to print chained contexts
+            detail: format!("{:#}", err.error),
+        }
+    }
+}
+
+impl From<LockFailure> for Error {
+    fn from(err: LockFailure) -> Self {
+        let detail = format!(
+            "the expectVersionId of spec {:?} {:?} did not match that of the live spec {:?}",
```
jgraettinger, line 36: nit: expectPubId?

crates/agent/src/handlers.rs
```diff
@@ -113,16 +116,39 @@ enum Status {
     Idle,
 }
 
+const PUBLICATIONS_TABLE: &str = "publications";
+const CONTROLLER_JOBS_TABLE: &str = "controller_jobs";
+
+/// The purpose of this is to ensure that the controllers handler gets polled
```
jgraettinger, line 122: I understand our intention is to rip all of this out when we tackle the updated lease-based job queue model.

We'll still want a mechanism to immediately wake a job dequeuer from a polling backoff sleep though :thinking:

crates/agent/src/publications/status.rs
psFried: I'd been leaving all the existing publication status fields, just in case we needed to deserialize an old status for some reason. That's now seeming overly cautious, though, so I just removed this and the `linked_materializations` stuff.

jgraettinger (commented):
Made it through "agent: cleanups from code review"

A lot of my comments are unedited and stream of conscious and calling it as I see it. Perhaps we sync up to review them and discuss next steps.

This looks awesome though -- clearly a lot of work and I'm excited about what this factoring / architecture buys us.

crates/agent/src/controllers/collection.rs
```diff
@@ -0,0 +1,255 @@
...
+        // the last time the inferred schema was updated. Then clamp that
+        // duration into a reasonable bounds. Collections that see very
+        // frequent inferred schema updates will be checked much more
+        // frequently, while those that are updated infrequently will be
+        // checked somewhat less frequently.
+        let start_time = self.schema_last_updated.unwrap_or(last_update);
+        let after_minutes = now
+            .signed_duration_since(start_time)
+            .num_minutes()
+            .max(min_backoff_minutes)
```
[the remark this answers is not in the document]
psFried: I think this is actually right?
Starting with the duration `d` since the previous update of the inferred schema, if `d < min_backoff_minutes` then we want to use `min_backoff_minutes`. But if `d > 120 minutes`, then we take the minimum value of `120`.

crates/agent/src/controllers/publication_status.rs
```diff
@@ -0,0 +1,317 @@
...
+            .map(|p| p.id)
+            .unwrap_or(Id::zero())
+            .max(state.last_pub_id);
+        if self.target_pub_id > last_success_pub_id {
+            let fail_count = self.history.iter().take_while(|p| !p.is_success()).count();
+            let after_seconds = match fail_count {
+                0..=1 => 60,
+                2 => 180,
+                3 => 300,
+                _ => 600,
```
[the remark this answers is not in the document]
psFried: No, if it has `abort` then the controller returns an error and it uses the logic [defined here](https://github.com/estuary/flow/pull/1489/files#diff-1ae46c5f49923434028707845bdf48fca06dc2f5b877dac192641a7f4893422bR122), which uses a _much_ steeper backoff. I'll admit that this separation does feel a bit half-baked, and it seems like we should be able to consolidate these cases. I'll look into it and report back.

crates/agent/src/handlers.rs
psFried: Yeah, one thing I realized is that our existing notifications mechanism doesn't really work for `controller_jobs` because they don't have a `{"type":"queued"}` status, so they rely on the periodic triggers from within the agent, as well as this hack. But I think that a periodic trigger seems totally fine, as long as we can also trigger controllers promptly in response to a publication. The main improvements I'd like to see here are to make the periodic trigger more explicit and targeted to controllers, and to make this publications hack somewhat less "hacky" (perhaps :wave: by passing a `ControllersWaker` into the `Publisher`).

crates/agent-sql/src/controllers.rs
psFried: Following up after a VC, we decided to keep `controller_version`, but just not use it to dequeue jobs needing "upgraded". This actually allows the initial rollout of controllers for existing `live_specs` to be done incrementally. The `dequeue` function will only pick up rows that have _both_ a `controller_next_run` that's due and a corresponding `controller_jobs` row. So we can remove the initial backfill of the `controller_jobs` table from the controllers database migration, and move it into a separate migration, which is run _after_ the rollout of new agents completes. This backfilling of `controller_jobs` rows can then be done incrementally, which limits the blast radius if anything goes wrong.

crates/agent/src/draft.rs
psFried: done

supabase/migrations/50_controllers.sql
psFried: We talked about this on VC and [this commit](https://github.com/estuary/flow/pull/1489/commits/c33c583b4242cfdc3afa6870052aafe9d4bd4dea) introduces a better approach for the migration. Also edited the pr description to add an overall deployment plan.

crates/agent/src/controllers/publication_status.rs
```diff
@@ -0,0 +1,317 @@
...
+        let pending = self.pending.as_mut().unwrap();
+        pending.details.push(add_detail);
+        pending
+    }
+
+    pub fn has_pending(&self) -> bool {
+        self.pending.is_some()
+    }
+
+    pub async fn start_spec_update<'a>(
```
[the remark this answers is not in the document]
psFried: We spoke on VC and decided to kick the can on actually doing this, but agreed that it would be a good refactor.

supabase/migrations/50_controllers.sql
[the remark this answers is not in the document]
psFried: Following up: the constraints have been re-added along with `on delete cascade`

crates/agent-sql/src/live_specs.rs
psFried: We discussed this on VC and I created #1494 to track the issue

jgraettinger (approved):
LGTM :tada:

PR #1495, opened 2024-06-20
crates/flowctl: improve handling for unauthenticated clients (for CI)

jgraettinger:
Ensure we're surfacing an error from LiveCatalog resolution. Remove LiveCatalog as a result from local_specs::validate() because no callers actually care about it.

Extend Client to track whether it's authenticated or not, and then update the flowctl Resolver to no-op resolutions for unauthenticated clients instead of returning an error.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1495)

### DIFF
--- a/crates/flowctl/src/controlplane.rs
+++ b/crates/flowctl/src/controlplane.rs
@@ -12,7 +12,14 @@ use std::sync::Arc;
 /// This client implements `Deref<Target=Postgrest>`, so you can use it
 /// just like you would the normal `Postgrest` client.
 #[derive(Clone)]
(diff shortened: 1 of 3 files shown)


psFried (approved):
LGTM
Just to double check: Do we have any usages of `flowctl raw build` that rely on the behavior of _not_ resolving live specs? Wondering if we'd need a `--no-resolve` flag to avoid breaking things in case `flowctl` _is_ authenticated.

jgraettinger:
I'm not aware of any. I've confirmed catalog tests continue to pass with this change when authenticated.

Issue #1494, opened 2024-06-19
agent: handle recusive spec expansion when publishing tests covering multiple derivations


psFried:
Spec expansion currently just pulls in authorized tasks that read from or write to one of the drafted collections. This works fine for most cases, but there's one case that it doesn't cover:

Say there's a series of derivations that source from one another, like `derive/a -> derive/b -> derive/c` and a test that inserts into `a` and verifies the output of `c` _without_ inserting or verifying `b`. Then spec expansion would only pull in `a` and `c` if you draft just the test.

Of course the chain of derivations can be arbitrarily long, and so our spec expansion logic would need to be recursive in order to handle that.

PR #1496, opened 2024-06-20
agent: fix running tests

psFried:
**Description:**

Previously when we ran tests, the single build database in the temp directory would always contain all the specs needed in order to perform the test run. This is no longer the case, now that builds only ever modify the specs that were actually drafted. So the build ids of unmodified specs would not have a corresponding build database in the temp directory.

This updates the agent to override the build id of each shard that's activated for testing to be the current build id, thus satisfying the expectations of `api test` and `temp-data-plane`.

As part of making that work, the activation of shards in the temp data plane is now done using the `activate` crate instead of shelling out to `api activate`. This allows us to modify the shard specs in memory instead of needing to write out a whole separate build database. It's also another step toward eliminating `api activate` altogether.

**Workflow steps:**

**Notes for reviewers:**

There's at least one other problem case that isn't handled here:
Say a user has `derive/a`, `derive/b`, and `test/a`, where `test/a` requires both derivations to be present. If a user publishes only `derive/a`, then spec expansion will pull in `test/a`, but not necessarily `derive/b`. Then when we try to run the test, it won't have all the required derivations. This is kind of a variation on #1494, which we've decided isn't super urgent, so I'm assuming this isn't super urgent either.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1496)


crates/agent/src/publications.rs
```diff
@@ -263,31 +263,35 @@ impl Publisher {
...
-            tokio::pin!(test_jobs);
-            let errors: Vec<tables::Error> = tokio::select! {
-                r = data_plane_job => {
-                    tracing::error!(?r, "test data-plane exited unexpectedly");
-                    test_jobs.await // Wait for test jobs to finish.
-                }
-                r = &mut test_jobs => r,
-            }?;
+            tracing::info!(%build_id, %publication_id, tmpdir = %tmpdir.display(), "running tests");
+            let errors = {
```
jgraettinger, line 267: nit: i'm not spotting why this extra nesting is required, or what's actually changed in here

jgraettinger (approved):
LGTM! % one nit

crates/agent/src/publications.rs
psFried: Ah, that's due to the fact that we now borrow the built output when running tests, and it conflicted with the mutable borrow when adding test errors to the built output.

I'm now realizing that it may be best to keep test errors as a separate field in `UncommittedBuild`, which would eliminate the need for this nesting and more importantly would allow returning the `TestsFailed` status, which was left as a todo. I'll put up a separate pr for that tomorrow morning.

PR #1498, opened 2024-06-21
validation: more intuitive error message for expected pub

mdibaiee:
**Description:**

- the original error is too technical for most users to understand what it means

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1498)

### DIFF
--- a/crates/agent/src/integration_tests/locking_retries.rs
+++ b/crates/agent/src/integration_tests/locking_retries.rs
@@ -46,7 +46,7 @@ async fn test_publication_optimistic_locking_failures() {
         vec![
             (
             "flow://collection/mice/does-not-exist".to_string(),
-            "expected publication ID 0102030405060708 was not matched (it's actually 0000000000000000)".to_string()
+            "expected publication ID 0102030405060708 was not matched (it's actually 0000000000000000): your changes have already been published or another publication has modified this spec; please try again with a fresh copy of the spec.".to_string()
             )
         ],
         errors);
--- a/crates/validation/src/errors.rs
+++ b/crates/validation/src/errors.rs
@@ -200,7 +200,7 @@ pub enum Error {
         pub_id: models::Id,
         larger_id: models::Id,
     },
-    #[error("expected publication ID {expect_id} was not matched (it's actually {actual_id}): another publication has modified this spec; to avoid conflicts, please try again with a fresh copy of the spec.")]
+    #[error("expected publication ID {expect_id} was not matched (it's actually {actual_id}): your changes have already been published or another publication has modified this spec; please try again with a fresh copy of the spec.")]
     ExpectPubIdNotMatched {
         expect_id: models::Id,
         actual_id: models::Id,
--- a/crates/validation/src/errors.rs
+++ b/crates/validation/src/errors.rs
@@ -200,7 +200,7 @@ pub enum Error {
         pub_id: models::Id,
         larger_id: models::Id,
     },
-    #[error("expected publication ID {expect_id} was not matched (it's actually {actual_id})")]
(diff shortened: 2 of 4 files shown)


psFried (changes requested):
Thanks for bringing this up. I'd like to pile on and suggest a further improvment:

~this usually means~ another publication has modified ~the resources included in this publication~ this spec; to avoid conflicts, please refresh the resources and try again

I feel like "refresh the resources" is also rather vague, though I don't have a ready suggestion. If the user was publishing via flowctl, then they need to run `pull-specs` with `--overwrite`, and then make their edits again and re-publish. If they were publishing via the UI, then they need to cancel their current edit and start over.

The "refresh" terminology might be a bit misleading in the UI, because I _think_ (not sure, looking to @travjenkins or @kiahna-tucker to confirm) that if a user is on the edit page and they re-load the page, then it will preserve their previous draft so they would just get the same error again. My understanding is that in the UI they'd need to cancel their edit entirely and start over from the beginning. Technically, they might be able to just re-fresh the bindings if the error was for a collection spec when editing a capture, but that feels like too much nuance to try to communicate in an error message.

dyaffe (changes requested):
I would probably switch to this since the common path is that your changes were already made:

expected publication ID {expect_id} was not matched (it's actually {actual_id}): this usually means your changes have already been made; to avoid conflicts, please refresh the resources and confirm.

mdibaiee:
> another publication has modified this spec; to avoid conflicts, please refresh the resources and try again

I think "spec" is still a very technical word and not quite as obvious what it means for the "spec" to have been modified. I'm trying to think through the eyes of a user who has just landed on our dashboard for the first time and has created a capture and a materialization and went to update it for the first time and got this error. We do use the word in the UI though, we call the editor "Advanced Specification Editor" and in the details page of tasks and collections we have "SPEC" tab, so perhaps I'm being pessimistic and it is a familiar word for users.

> I feel like "refresh the resources" is also rather vague

I agree, it irks me as well... given that users need to start over in the UI and they need to fetch the remote spec again before continuing in flowctl, I guess we could say "please try the operation again on a fresh copy of specifications" or something of the sort?

kiahna-tucker:
>  if a user is on the edit page and they re-load the page, then it will preserve their previous draft so they would just get the same error again. My understanding is that in the UI they'd need to cancel their edit entirely and start over from the beginning.

Yes Phil, that is correct.

mdibaiee:
> expected publication ID {expect_id} was not matched (it's actually {actual_id}): this usually means your changes have already been made; to avoid conflicts, please refresh the resources and confirm.

@dyaffe this case of a change already having been made, to me is more of a bug in the UI that needs to be fixed to be honest. I've hit this issue a few times, where I hit Save, the publication is published but I'm not redirected / the page is not refreshed as usual, so I end up trying to save again only to hit this error. The error would be misleading in other scenarios though (when multiple tabs edit the resource, etc.)

dyaffe:
@mdibaiee everyone says that and we attempted to fix this but 100% of the time that a user has gotten this error message so far, it has been a double click in the UI.

mdibaiee:
How about this? I would personally prefer if the ending was a bit more clear (a fresh copy of the spec might not be as clear), so please suggest if you have a better idea 🤔 

> expected publication ID {expect_id} was not matched (it's actually {actual_id}): your changes have already been published or another publication has modified this spec; please try again with a fresh copy of the spec.

dyaffe (approved):
I'm not in love with the phrasing "with a fresh copy of the spec.", but don't have anything better.

PR #1499, opened 2024-06-21
agent: return TestFailed status when tests fail

psFried:
**Description:**

Updates the publisher to store test errors separately from other types of errors, and to set the `TestFailed` status if any test errors are present. Previously, test failures would result in a `BuildFailed` status, which was unhelpful.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1499)

### DIFF
--- a/crates/agent/src/publications.rs
+++ b/crates/agent/src/publications.rs
@@ -30,8 +30,11 @@ pub struct PublicationResult {
     /// The state of any related live_specs, prior to the draft being published
     pub live: tables::LiveCatalog,
     /// The build specifications that were output from a successful publication.
-    /// Will be empty if the publication was not successful.
+    /// May be empty if the build failed prior to validation. Contains any validation
+    /// errors.
     pub built: tables::Validations,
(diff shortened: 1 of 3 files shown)


crates/agent/src/publications.rs
```diff
@@ -30,8 +30,11 @@ pub struct PublicationResult {
     /// The state of any related live_specs, prior to the draft being published
     pub live: tables::LiveCatalog,
     /// The build specifications that were output from a successful publication.
-    /// Will be empty if the publication was not successful.
+    /// May be empty if the build failed prior to validation. Contains any validation
+    /// errors.
     pub built: tables::Validations,
+    /// Errors that occurred while running tests.
```
jgraettinger, line 36: supernit: tables::Errors instead of Vec ?

jgraettinger (approved):
LGTM

crates/agent/src/publications.rs
psFried: yeah, agreed. Done

PR #1500, opened 2024-06-21
agent: generate JSON schema for controller status

psFried:
**Description:**

Adds an automatically generated json schema for the agent controller status objects. The goal is just to make it easier to consume status information.

Also updates the default publication history length from 3 to 5, which has just felt like a better number to ease debugging schema evolution issues.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1500)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4428,9 +4428,9 @@ dependencies = [
 
 [[package]]
 name = "schemars"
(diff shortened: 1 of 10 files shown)


jgraettinger (approved):
LGTM

PR #1503, opened 2024-06-24
agent: add logging and errors around null built_specs

psFried:
Adds some logging and better errors when failures occur due to `built_spec` being null

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1503)


jgraettinger (approved):
LGTM

PR #1506, opened 2024-06-25
Phil/no key changes

psFried:
**Description:**

Re-introduces the prior behavior of rejecting any changes to the key or
logical partitioning of a collection spec. We'd eventually like to allow
these changes, but it seems our first go was a bit premature, since we
don't yet have another way to respect `evolveIncompatibleCollections:
false` if the key does change.

I tried to keep this change relatively localized and easy to rip out.

**Workflow steps:**

Publish an existing collection with a different key or logical partitioning.
It should now fail and return `incompatible_collections`.

n/a

**Notes for reviewers:**

Also includes a minor cleanup in agent-sql.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1506)

### DIFF
--- a/crates/agent-sql/src/live_specs.rs
+++ b/crates/agent-sql/src/live_specs.rs
@@ -31,7 +31,7 @@ pub struct LiveSpec {
     pub catalog_name: String,
     pub spec_type: Option<CatalogType>,
     pub spec: Option<TextJson<Box<RawValue>>>,
-    pub built_spec: Option<Json<Box<RawValue>>>,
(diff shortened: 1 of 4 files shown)


jgraettinger (approved):
LGTM

FYI I don't think it's _ever_ gonna be okay to change logical partitioning of an active collection, but we can change the key. I'll make a note that we should probably push these down into the `validation` crate.

psFried:
> I don't think it's ever gonna be okay to change logical partitioning of an active collection

Ok, that makes sense, and I'd agree that eventually pushing it down into `validation` also makes sense :+1:

PR #1510, opened 2024-06-28
Phil/agent fixes

psFried:
**Description:**

Rolls up a few different fixes for agent:

- Prevent null bytes from getting into live/built specs
- Handle errors in publications and controllers more gracefully
- Process controller jobs sooner

Individual commit messages have more details.

**Notes for reviewers:**

The default backoff for controller errors is pretty steep, and also uses a very large random jitter. This is to prevent thundering herd scenarios in case there's a widespread problem that causes a large proportion of controller jobs to fail.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1510)

### DIFF
--- a/crates/agent/src/publications/handler.rs
+++ b/crates/agent/src/publications/handler.rs
@@ -35,19 +35,34 @@ impl Handler for Publisher {
                 .context("clearing old errors")?;
 
             let time_queued = chrono::Utc::now().signed_duration_since(row.updated_at);
-            let result = self.process(row).await?;
 
-            draft::insert_errors(draft_id, result.draft_errors(), &mut txn).await?;
+            let (status, draft_errors) = match self.process(row).await {
+                Ok(result) => {
+                    let errors = result.draft_errors();
+                    (result.status, errors)
+                }
+                Err(error) => {
+                    tracing::warn!(?error, pub_id = %id, "build finished with error");
+                    let errors = vec![draft::Error {
+                        catalog_name: String::new(),
+                        scope: None,
+                        detail: format!("{error:#}"),
+                    }];
+                    (JobStatus::PublishFailed, errors)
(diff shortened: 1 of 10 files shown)


psFried:
Here's a scenario I've tested locally:

- Setup a webhook capture with schema inference, and a materialization to postgres
- Ingest a document with a `\u0000` in one of the property names, and observe that the inferred schema is updated to reflect that
- The next controller run for the collection fails with an error about the null byte in the spec, and is set to retry
- Pull and edit the collection spec to no longer use the inferred schema
- After publishing the updated collection spec, the controller error clears

crates/agent/src/publications/handler.rs
```diff
@@ -35,19 +35,34 @@ impl Handler for Publisher {
...
-            draft::insert_errors(draft_id, result.draft_errors(), &mut txn).await?;
+            let (status, draft_errors) = match self.process(row).await {
+                Ok(result) => {
+                    let errors = result.draft_errors();
+                    (result.status, errors)
+                }
+                Err(error) => {
+                    tracing::warn!(?error, pub_id = %id, "build finished with error");
+                    let errors = vec![draft::Error {
+                        catalog_name: String::new(),
```
jgraettinger, line 47: is it okay with the UI if the draft_error `scope` column is completely empty?

jgraettinger (approved):
LGTM % question

crates/agent/src/publications/handler.rs
psFried: Great question. It took me a bit to test this because apparently postgresT doesn't allow you to insert specs containing null bytes. I was actually pretty happy to find that out. I wasn't able to draft anything including a null byte either using the advanced spec editor or via flowctl.

PR #1513, opened 2024-07-02
agent: revert publication job status serde changes

psFried:
Recently, the publication `JobStatus` serde attributes changed to using camelCase instead of snake_case. This broke integration with the UI, which expects `requires_recreation` instead of `requiresRecreation` in order to determine whether a collection needs re-created. This reverts that change. The move to camelCase had been in vain, anyway, because we still had a mix of camel and snake case afterward. That seems like something we're kind of stuck with for now.


jshearer (approved):
I haven't been following all of the changes here particularly closely, but the problem makes sense and functionally this LGTM 👍

PR #1514, opened 2024-07-03
Phil/inferred schemas now

psFried:
**Description:**

Introduces _much_ faster updates to inferred schemas by adding a database trigger that schedules a controller run in response to each insert/update on the `inferred_schemas` table.

This replaces the periodic checks as a mechanism for updating inferred schemas. I did leave a very infrequent periodic check in place, though, as a fallback in case the triggers get bypassed or disabled somehow. Note that removing the frequent periodic check is not required for correctness, and serves mainly to reduce the load on our agents and database. Thus the main functionality should work just fine with the current agent version after applying the migration, and deploying the agent itself would just be to reduce the frequency of (now unnecessary) controller runs.

The trigger fires for each row, because the postgres materialization will execute a prepared statement for each insert/update, and thus there's no benefit to using a statement-level trigger.

One risk to call out is that the `stats_view` materialization might contend with publications over updates to `live_specs` rows. This doesn't seem likely because the materialization should use pretty short-lived transactions when performing the inserts/updates. If that does turn out to be an issue, then it might be possible to work around it by introducing a `controller_triggers` table, and staging the updates there and having a cron job that updates `live_specs.controller_next_run`.

**Testing:**

I've been testing this out locally using a mongodb capture with 200 collections, all using schema inference, and a postgres materialization. Then running a few variations on this script in `mongosh` to repeatedly insert documents with new properties into each of the collections:

```
for (let r = 2; true; r++) {
  let propName = "p_" + r; 
  for (let i = 0; i < 200; i++) {
    let name = "c_" + i;
    let d = {"cname": name};
    d[propName] = true;
    db.getCollection(name).insertOne(d);
    printjson(d);
    sleep(2000);
  }
}
```

As expected, the materialization will stop essentially all progress while that script is running, because it continuously hits schema validation errors. Once the script is stopped and all pending controllers have run, the materialization starts working again.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1514)

### DIFF
--- a/crates/agent/src/controllers/collection.rs
+++ b/crates/agent/src/controllers/collection.rs
@@ -201,35 +201,9 @@ impl InferredSchemaStatus {
             tracing::debug!(%collection_name, "No inferred schema available yet");
         }
 
-        let next_run = Some(self.next_run(state.updated_at, control_plane.current_time()));
-        Ok(next_run)
-    }
-
-    fn next_run(&self, last_update: DateTime<Utc>, now: DateTime<Utc>) -> NextRun {
-        // The idea here is to check frequently if there isn't an inferred schema at all yet,
-        // so we can quickly start materializing some data. But after it works at least once,
-        // then we can slow down a little.
-        let min_backoff_minutes = if self.schema_md5.is_none() {
-            1i64
-        } else {
-            10i64
-        };
-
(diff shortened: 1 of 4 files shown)


jgraettinger (approved):
LGTM!

PR #1516, opened 2024-07-08
Dekaf group management protocols

jshearer:
**Description:**

Implement KafkaApiClient, which is a slightly higher level wrapper around a connection to a particular Kafka broker.

This client will mainly be used to transparently proxy group management API requests from Dekaf users through to a real Kafka broker. In addition to proxying, `KafkaApiClient` also needs to have enough smarts to be able to open and authenticate (and keep authenticated) its connection to the broker.

Fixes https://github.com/estuary/flow/issues/1531

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1516)


crates/dekaf/src/api_client.rs
```diff
@@ -0,0 +1,273 @@
...
+                .auth_bytes(Bytes::from(state_buf.into_inner()?))
+                .build()?;
+
+            let auth_resp = self._send_request(authenticate_request, None, conn).await?;
+
+            if auth_resp.error_code > 0 {
+                let err = kafka_protocol::ResponseError::try_from_code(handshake_resp.error_code)
+                    .map(|code| format!("{code:?}"))
+                    .unwrap_or(format!("Unknown error {}", handshake_resp.error_code));
+                bail!(
```
jshearer, line 261: TODO: Lift this duplicated error representation

jshearer:
Alright, should have hit that "ready for review" button days ago. Let me know what you think!

crates/dekaf/src/session.rs
```diff
@@ -1,33 +1,55 @@
...
+    handle: tokio::task::JoinHandle<anyhow::Result<(Read, usize, bytes::Bytes)>>,
 }
 
 pub struct Session {
     app: Arc<App>,
     client: postgrest::Postgrest,
     reads: HashMap<(TopicName, i32), PendingRead>,
+    listed_offsets: HashMap<TopicName, HashMap<i32, Option<(i64, i64)>>>,
+    /// ID of the authenticated user
+    uid: Option<String>,
```
jshearer, line 37: `uid` was going to be for isolating groups across users, but I realized later that I'm like 70% sure that nobody actually needs `ListGroups`, as all of the other group management APIs are idempotent over the group existing -- if it doesn't exist, it'll get created. It's just observability tools that care about `ListGroups`.

So yes this is unused, except that we'll _probably_ want it later on so I left it

crates/dekaf/src/read.rs
```diff
@@ -59,10 +65,71 @@ impl Read {
             uuid_ptr: collection.uuid_ptr.clone(),
             value_schema: collection.value_schema.clone(),
             value_schema_id,
+
+            journal_name: partition.spec.name.clone(),
+            client,
+        }
+    }
+
+    async fn skip_to_next_doc(&mut self) -> anyhow::Result<()> {
```
jgraettinger, line 74: In [this commit](https://github.com/estuary/flow/pull/1557/commits/1a971b6775793e3304e864f95bb648daa7153e95), I update `gazette` to:

> Refactor streaming error handling within the crate, to remove back-off
timers and policy around which errors are terminal or may be retried.
Instead, surface all errors in the response Stream, and let the client
decide to retry by polling the Stream again, or cancel by dropping the
Stream. Add Error::is_transient() to simplify client checks of common
retry-able error conditions.

I'm looking at this handling, and thinking that this pattern should be extended all the way down into JSON document transcoding: if a JSON parse error weren't terminal for a ReadJsonLines stream, then you could simply handle / log the error and decide that _in this context_ you'd like to poll it again to get the document at the newline _after_ the current (invalid) document.

crates/dekaf/src/session.rs
```diff
@@ -331,22 +351,82 @@ impl Session {
...
+
+            let latest_topic_requested = self.listed_offsets.get(&key.0);
 
             for partition_request in &topic_request.partitions {
                 key.1 = partition_request.partition;
 
-                if matches!(self.reads.get(&key), Some(pending) if pending.offset == partition_request.fetch_offset)
-                {
+                let mut fetch_offset = partition_request.fetch_offset;
+                let record_limit = None;
```
jgraettinger, line 362: record_limit is never set

crates/dekaf/src/read.rs
```diff
@@ -59,10 +65,71 @@ impl Read {
...
         }
+        bail!("Unable to find next document")
     }
 
-    pub async fn next_batch(mut self, target_bytes: usize) -> anyhow::Result<(Self, bytes::Bytes)> {
+    #[tracing::instrument(skip_all,fields(journal_name=self.journal_name))]
+    pub async fn next_batch(
+        mut self,
+        target_bytes: usize,
+        target_docs: Option<usize>,
```
jgraettinger, line 131: target_docs isn't ever Some(). Remove?

crates/dekaf/src/session.rs
```diff
@@ -61,20 +83,20 @@ impl Session {
...
         let response = match self.app.authenticate(authcid, password).await {
-            Ok(client) => {
+            Ok((client, config, uid)) => {
                 self.client = client;
+                self.config.replace(config);
+                self.uid.replace(uid);
 
                 let mut response = messages::SaslAuthenticateResponse::default();
-                response.session_lifetime_ms = i64::MAX; // TODO(johnny): Access token expiry.
+                response.session_lifetime_ms = 60 * 60 * 60 * 24; // TODO(johnny): Access token expiry.
```
jgraettinger, line 92: this is 60 days

crates/dekaf/src/session.rs
```diff
@@ -1,33 +1,55 @@
...
 }
 
 pub struct Session {
     app: Arc<App>,
     client: postgrest::Postgrest,
     reads: HashMap<(TopicName, i32), PendingRead>,
+    listed_offsets: HashMap<TopicName, HashMap<i32, Option<(i64, i64)>>>,
+    /// ID of the authenticated user
+    uid: Option<String>,
+    config: Option<ConfigOptions>,
```
jgraettinger, line 38: nit: I'd make it just `config: ConfigOptions` initialized with defaults, which are overwritten, and then you can avoid Some() checks. Although this doesn't appear to be accessed?

crates/dekaf/src/session.rs
```diff
@@ -477,29 +548,347 @@ impl Session {
...
+                .try_get_i16()
+                .context("failed to parse consumer protocol message")?;
+
+            tracing::debug!(
+                version = consumer_protocol_subscription_version,
+                remaining_bytes = consumer_protocol_subscription_raw.len(),
+                "Got consumer protocol message version"
+            );
+
+            // TODO: validate acceptable version
```
jgraettinger, line 633: what's TODO, exactly?

crates/dekaf/src/session.rs
```diff
@@ -477,29 +548,347 @@ impl Session {
...
+        if let Some(ref topics) = mutated_req.topics {
+            self.app
+                .kafka_client
+                .ensure_topics(topics.iter().map(|t| t.name.to_owned()).collect())
+                .await?;
+        }
+        let mut resp = self
+            .app
+            .kafka_client
+            .connect_to_group_coordinator(req.group_id.as_str())
```
jgraettinger, line 882: Should we cache the group ID on the session, along with the routed connection / pool of the current coordinator?

Seems like we could do this once if the session has a long life-cycle for heartbeats and whatnot. The group ID shouldn't ever change for a single session.

crates/dekaf/src/session.rs
```diff
@@ -552,6 +949,13 @@ impl Session {
...
+        res.api_keys.insert(
+            ApiKey::ProduceKey as i16,
+            ApiVersion::default()
+                .with_min_version(3)
+                .with_max_version(9),
+        );
+
         // UNIMPLEMENTED.
         /*
         res.api_keys
```
jgraettinger, line 961: please update this. what remains unimplemented?

crates/dekaf/src/api_client.rs
```diff
@@ -0,0 +1,519 @@
...
+}
+
+#[tracing::instrument(skip_all)]
+async fn sasl_auth<S: StreamSink>(
+    conn: &mut S,
+    args: &KafkaConnectionParams,
+) -> anyhow::Result<()> {
+    let sasl = SASLClient::new(args.sasl_config.clone());
+
+    let mechanisms = get_supported_sasl_mechanisms(args).await?;
```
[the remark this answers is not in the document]
jshearer: In a comment in `get_supported_sasl_mechanisms`:

> In order to pick the best method to use, we need to know the options supported by the server.
   `SaslHandshakeResponse` contains this list, but you have to send a `SaslHandshakeRequest` to get it,
    and if you send an invalid mechanism, Kafka will close the connection. So we need to open a throw-away
    connection and send an invalid `SaslHandshakeRequest` all in order to discover the supported mechanisms.

crates/dekaf/src/api_client.rs
```diff
@@ -0,0 +1,519 @@
...
+use std::{io::BufWriter, pin::Pin, sync::Arc};
+use tokio::net::TcpStream;
+use tokio_rustls::{
+    rustls::{pki_types::ServerName, ClientConfig, RootCertStore},
+    TlsConnector,
+};
+use tracing::instrument;
+use url::Url;
+
+trait StreamSink<I = Bytes, O = BytesMut, E = std::io::Error>:
```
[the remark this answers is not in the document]
jshearer: ... Wow, that _actually_ just works. I spent so long figuring out how to define a dyn Stream/Sink this when the answer was so simple. Thanks!

crates/dekaf/src/api_client.rs
```diff
@@ -0,0 +1,519 @@
+use anyhow::{anyhow, bail, Context};
+use bytes::{Bytes, BytesMut};
+use deadpool::managed::{self, Manager, Metrics, PoolError, RecycleResult};
```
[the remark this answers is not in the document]
jshearer: Yup, this is something I wasn't sure about but makes a lot of sense to me. I almost wish there was a `rust-analyzer` feature like `autoimport`, but that automatically qualifies names with their parent module... except then you'd have a bunch of references to e.g `managed::PoolError` which is much less useful than `deadpool::managed:PoolError`, so this might just be more subjective than a rule can define... I'll go through and make this change though 👍

crates/dekaf/src/api_client.rs
```diff
@@ -0,0 +1,519 @@
...
+    args: &KafkaConnectionParams,
+) -> anyhow::Result<()> {
+    let sasl = SASLClient::new(args.sasl_config.clone());
+
+    let mechanisms = get_supported_sasl_mechanisms(args).await?;
+
+    let maybe_offered_mechanisms: Result<Vec<_>, _> = mechanisms
+        .iter()
+        .map(|m| Mechname::parse(m.as_str().as_bytes()))
+        .collect();
```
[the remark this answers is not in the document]
jshearer: Weird, I had tried that originally and it didn't work so I switched to this more awkward construction. It works now though so... changed!

crates/dekaf/src/api_client.rs
```diff
@@ -0,0 +1,519 @@
...
+    )?;
+
+    tracing::debug!(api_key_name=?req_api_key, api_key=Req::KEY, api_version=request_header.request_api_version, "Sending request");
+
+    req.encode(&mut req_buf, request_header.request_api_version)?;
+
+    // Then write the message
+    conn.send(req_buf.freeze()).await?;
+
+    // Now we can read the whole message. Let's not worry about streaming this
```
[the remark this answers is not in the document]
jshearer: Yup, true 👍

Cargo.toml
```diff
@@ -105,6 +106,16 @@ rocksdb = { version = "0.22", default-features = false, features = [
     "rtti",
 ] }
 rkyv = { version = "0.7", features = ["archive_le"] }
+# TODO: Switch back to crates.io version of rsasl when this PR is merged:
+# https://github.com/dequbed/rsasl/pull/48
```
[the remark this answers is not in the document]
jshearer: Yup

Cargo.toml
```diff
@@ -127,10 +140,13 @@ sqlx = { version = "0.6", features = [
     "uuid",
 ] }
 
+typestate = "0.8.0"
+tokio-rustls = "0.26.0"
+webpki-roots = "0.26"
```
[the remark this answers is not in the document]
jshearer: Great, this is much better. Thanks!

crates/dekaf/src/api_client.rs
```diff
@@ -0,0 +1,519 @@
...
+                        coord
+                    },
+                )
+            }
+            Some(coord) => Ok(coord.clone()),
+        }
+    }
+
+    #[instrument(skip(self))]
+    pub async fn connect_to_controller(&self) -> anyhow::Result<KafkaApiClient> {
```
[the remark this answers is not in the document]
jshearer: Added

crates/dekaf/src/api_client.rs
```diff
@@ -0,0 +1,519 @@
...
+}
+
+#[tracing::instrument(skip_all)]
+async fn async_connect(broker_url: &str) -> anyhow::Result<BoxedKafkaConnection> {
+    // Establish a TCP connection to the Kafka broker
+
+    let parsed_url = Url::parse(broker_url)?;
+
+    let mut root_cert_store = RootCertStore::empty();
+    root_cert_store.extend(webpki_roots::TLS_SERVER_ROOTS.iter().cloned());
```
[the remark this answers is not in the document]
jshearer: Done

crates/dekaf/src/read.rs
```diff
@@ -241,20 +241,22 @@ impl Read {
...
-                partition_leader_epoch: 1,
-                producer_epoch: 1,
-                producer_id: producer.as_i64(),
-                sequence: kafka_offset as i32,
-                timestamp: unix_seconds as i64 * 1000 + unix_nanos as i64 / 1_000_000, // Map into millis.
-                timestamp_type: TimestampType::LogAppend,
-                transactional: false,
-                value,
-            });
+            if !is_control {
```
jshearer, line 244: @jgraettinger This feels like a particularly high-impact change. After our talk on sprint planning today I thought more about this, and finally after a bunch of testing it _seems_ to work just fine and solve the problem of emitting a null document any time there's an ack... Am I introducing incorrectness here that you can tell?

jshearer: FWIW, I first tried just setting `control: is_control`, since it looks like that is supposed to make consumers skip over that document as if it weren't there, but that broke everything so I went with this approach instead

crates/dekaf/src/read.rs
jgraettinger: This is implemented here: https://github.com/estuary/flow/pull/1589

Want to pull it in and give it a try?

crates/dekaf/src/session.rs
jshearer: Accessed down in `encode_topic_name()`.

AFAICT, the only reason to do that (initialize `ConfigOptions` with defaults rather than have it be `Option`al) is to avoid the runtime overhead/messyness of an `.expect()` call. Considering the non-trivial complexity involved, i.e a `Session` _is_ valid and can respond to certain API requests before authentication (like `ApiVersions`, `SaslHandshake`, etc) but not others, I'd rather have a hard failure if I try to access something that hasn't been initialized yet (such as `config`), rather than silently and subtily do the wrong thing by using a default.

The core problem is that we're representing the internal state of a state machine without being clear about it -- accessing `config` simply doesn't make any sense before a `Session` is authenticated, not to mention calling all of the APIs that rely on you being authenticated... I should probably make it explicit which APIs can be called when, shouldn't I.

crates/dekaf/src/session.rs
jshearer: The thought was that in the future, newer consumer versions could introduce newer versions that this code doesn't support, and we should validate that the version is <= the max supported version

crates/dekaf/src/session.rs
jshearer: tl;dr we used to cache the group->coordinator map, and I witnessed the coordinator of a group change which broke that session until I restarted Dekaf. Granted, this happened when I was updating the MSK cluster config so probably not common, but it does happen. These requests are cheap and fast, and I updated to cache the broker uri -> pool map instead

crates/dekaf/src/session.rs
jshearer: No longer unused

crates/dekaf/src/read.rs
jshearer: Done!

```
2024-08-29T21:11:38.494853Z  WARN serve{addr=[::ffff:127.0.0.1]:51954}:handle_api{api_key=FetchKey v=11}: dekaf::session: discarding pending read due to offset jump topic="demo.2Fwikipedia.2Frecentchange.2Dsampled" partition=0 old_offset=14064267732 new_offset=14064264074
2024-08-29T21:11:38.578116Z DEBUG next_batch{journal_name="demo/wikipedia/recentchange-sampled/pivot=00"}: dekaf::read: Ignoring first parse error to skip past partial document err=failed to parse document at journal offset range 14064264074..14064264125
2024-08-29T21:11:38.578517Z DEBUG next_batch{journal_name="demo/wikipedia/recentchange-sampled/pivot=00"}: dekaf::read: returning records count=3 first_offset=14064265337 last_offset=14064267731 last_write_head=14064267801 ratio=0.4885906040268456 records_bytes=2234
```

crates/dekaf/src/session.rs
jshearer: Removed

crates/dekaf/src/read.rs
```diff
@@ -59,10 +65,71 @@ impl Read {
...
+        let req = broker::ReadRequest {
+            offset: self.offset,
+            block: true,
+            journal: self.journal_name.clone(),
+            begin_mod_time: not_before_sec as i64,
+            ..Default::default()
+        };
+
+        let read = self.client.clone().read(req);
+        pin_mut!(read);
```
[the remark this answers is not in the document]
jshearer: Code got deleted

jshearer:
Forgot to commit after getting it working locally: added `/prometheus` endpoint that looks like this for the moment

```
# HELP bytes_read bytes_read
# TYPE bytes_read counter
bytes_read{journal_name="demo/wikipedia/recentchange-sampled/pivot=00"} 15021766
# HELP documents_read documents_read
# TYPE documents_read counter
documents_read{journal_name="demo/wikipedia/recentchange-sampled/pivot=00"} 16788
# HELP total_connections total_connections
# TYPE total_connections gauge
total_connections 5
```

crates/dekaf/src/api_client.rs
```diff
@@ -0,0 +1,519 @@
...
+                anyhow::bail!(e)
+            }
+        };
+
+        send_request(conn.as_mut(), req, header).await
+    }
+
+    #[instrument(skip(self))]
+    pub async fn connect_to_group_coordinator(&self, key: &str) -> anyhow::Result<KafkaApiClient> {
+        let mut coordinators = self.coordinators.clone().lock_owned().await;
```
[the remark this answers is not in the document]
jshearer: This has been refactored now

.github/workflows/dekaf.yaml
```diff
@@ -0,0 +1,60 @@
...
+  pull_request:
+    branches: [master]
+    paths:
+      - ".github/workflows/dekaf.yaml"
+      - "crates/dekaf/**"
+      - "Cargo.lock"
+
+jobs:
+  build-and-test:
+    runs-on: ubuntu-20.04
```
jgraettinger, line 18: use ubuntu-latest

crates/dekaf/Dockerfile
```diff
@@ -0,0 +1,24 @@
+FROM lukemathwalker/cargo-chef:latest-rust-1 AS chef
```
jgraettinger, line 1: I understand you're probably using this to more easily develop on Mac, but all other CI pipelines build directly within the Github actions runner environment and then package the binary into a docker container.

The reason we do it this way is for across-the-board consistency of the build environment of our binaries.

crates/dekaf/src/read.rs
```diff
@@ -75,17 +83,53 @@ impl Read {
...
+                                "skipping past non-object node at offset {}: {:?}",
+                                self.offset,
+                                non_object.to_debug_json_value()
+                            );
+                            continue;
+                        }
+                    },
+                    Err(err) if err.is_transient() => {
+                        tracing::warn!(%err, "Retrying transient read error");
+                        // We can retry transient errors just by continuing to poll the stream
```
jgraettinger, line 116: add a delay here to not thundering-herd

crates/dekaf/src/read.rs
jgraettinger: My hazy recollection is that:

* Materialize will query all partitions for their high water-mark.
* It expects to read through those water-marks before it release the first "update" within it's timely data-flow machinery.
* If a collection is idle (no ongoing writes), that means the last message is an ACK. If we suppress ACKs then Materialize would basically hang forever and not return results because it never reads through its expected initial high water-mark.

Please confirm this issue isn't re-opened with this change.

jgraettinger (approved):
LGTM % a few bits to clean up in comments below.

(Also there's a current build failure in CI)

crates/dekaf/Cargo.toml
```diff
@@ -15,25 +15,49 @@ doc = { path = "../doc" }
 gazette = { path = "../gazette" }
 labels = { path = "../labels" }
 ops = { path = "../ops" }
+flowctl = { path = "../flowctl" }
```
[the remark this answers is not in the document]
jshearer: Out of curiosity, where do you think the referred code should live then? Specifically it's [`flowctl::pagination`](https://github.com/estuary/flow/blob/master/crates/flowctl/src/pagination.rs) which is used to read items from postgrest using the pagination API. We don't seem to have a `utils` crate or anything like that 🤔

crates/dekaf/Dockerfile
jshearer: Alright, I can make that change. It's a deviation from my understanding of the "normal" way to build an image with a `Dockerfile` and instead encoding more of the build logic in github actions's less portable/standard syntax... but consistency is valuable and I'm pretty sure it'll be faster to save/load build caches. Out of curiosity, why do we do it this way in the first place?

Cargo.toml
```diff
@@ -129,10 +144,15 @@ sqlx = { version = "0.6", features = [
...
+typestate = "0.8.0"
+tokio-rustls = "0.26.0"
+rustls = { version = "0.23.10", features = ["aws_lc_rs"] }
+rustls-pemfile = "2.1.3"
+webpki = "0.22.4"
 tempfile = "3.3"
 portpicker = "0.1.1"
 tempdir = "0.3"
-thiserror = "1.0"
+thiserror = "1.0.63"
```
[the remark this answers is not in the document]
jshearer: Yeah that makes sense. I think I had added this to try and fix a cargo incompatible-version error, but it works just fine without it so I'll leave it off 👍

.github/workflows/dekaf.yaml
```diff
@@ -0,0 +1,60 @@
...
+          echo "${{ secrets.GITHUB_TOKEN }}" | \
+            docker login --username ${{ github.actor }} --password-stdin ghcr.io
+      - name: Run git describe
+        id: git_describe
+        run: echo "describe_output=$(git describe --dirty --tags)" >> $GITHUB_OUTPUT
+      # This step uses [docker/metadata-action](https://github.com/docker/metadata-action#about) to extract tags and labels that will be applied to the specified image.
+      # The `id` "meta" allows the output of this step to be referenced in a subsequent step. The `images` value provides the base name for the tags and labels.
+      - name: Extract metadata (tags, labels) for Docker
+        id: meta
+        uses: docker/metadata-action@a64d0487d7069df33b279515d35d60fa80e2ea62
```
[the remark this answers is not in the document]
jshearer: This seems to be an open point of disagreement in the github actions community. From what I could find, the strongest argument for using SHAs instead of tags is security -- avoiding a maliciously re-published tag. Beyond that, it seems people still prefer SHAs as otherwise you're relying on the action owner to correctly implement semver, or if you specify down to the patch version then you have to manually update anyway.

Github themselves [say](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/using-pre-written-building-blocks-in-your-workflow#using-shas):
> If you need more reliable versioning, you should use the SHA value associated with the version of the action. SHAs are immutable and therefore more reliable than tags or branches. However, this approach means you will not automatically receive updates for an action, including important bug fixes and security updates. 

If you still think we should stick to tags instead of SHAs, I'm happy to. Just wanted to frame the reason I went with what I did

crates/dekaf/Dockerfile
jgraettinger: Quite literally right this moment, @psFried is debugging a subtle weirdness / breakage in our `source-http-ingest` connector which is currently suspected to be an issue in its... wait for it... dockerized build process :laughing: 

Build environments are a *long term* commitment. It's not enough that it builds today. The environment also needs to be continuously updated with version bumps, security fixes, and so on. It's never actually a static environment, as much as the idea of "hermetic builds" is appealing. One day `lukemathwalker` will move on, and the image will not be updated any more, and it will bit rot to an unacceptable degree, and there will be a scramble to get another build set up that usually requires first de-tangling bespoke bits of the legacy one.

In that framing, Github actions environments are **very much** a standardized build environment: an environment used by tons of people, that see continuous delivery of updated compilers, base OS images, and so on.

C.f. Gazette's make-within-a-docker-builder-called-by-make build system which I wrote, and which is "clever", but which is today quite bit-rotted and confounding even to me.

crates/dekaf/Cargo.toml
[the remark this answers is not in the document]
jgraettinger: A little duplication is better than a little dependency.

Or, live with the smell for now, but with an active plan to improve it later (we're intending to fetch singleton materialization specs after all; or, perhaps you have an opinion on a better factoring, or perhaps a better spot will become apparent later, or maybe the postgrest crate has better utilities for pagination now?).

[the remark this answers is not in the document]
jshearer: Just to fine-tune my "inner Johnny" a bit,  what's the problem with this dependency? Is it about build times, i.e pulling in all of flowctl (and its dependencies) just for one module? Or is it some more fundamental argument?

crates/dekaf/Dockerfile
jshearer: I'm happy to make this change and have most of this done and will push it up tmw, but I am curious about understanding your position on this more clearly.

> One day ’lukemathwalker’ will move on, and the image will not be updated any more

This is the point I most clearly understand and agree with. This base image isn't adding all that much ontop of its own base image so I should really just include it here and depend on e.g ’rust:latest’ instead. 

> Build environments are a long term commitment. It's not enough that it builds today. The environment also needs to be continuously updated with version bumps, security fixes, and so on.

💯. If you depend on e.g ’rust:latest’ or even ’rust:1-bullseye’, you should have an always up to date build environment so long as the Rust organization maintains their base images, much in the same way depending on ’ubuntu-latest’ + ’actions-rust-lang/setup-rust-toolchain@v1’ does in GH actions. 
 
> In that framing, Github actions environments are very much a standardized build environment: an environment used by tons of people, that see continuous delivery of updated compilers, base OS images, and so on.

Sure, but I guess I’m not seeing how they’re any better than Dockerfiles? They both let you pull in snippets written and maintained (or not) by other people, they both let you define some series of build steps to run with inputs and outputs, they both let you specify dependencies/environments in such a way that they’re “always up to date” for better (security updates) or worse (subtly breaking changes to things you depend on). 


I would argue that Dockerfiles are better because they’re not nearly as vendor-locked in. I can find all sorts of non-Docker options for building, testing and running Dockerfiles on my machine exactly as they are in CI, whereas with GitHub actions, your only option is to run them inside GH’s infrastructure against whatever set of largely opaque base images you’re offered. I dunno about you, but my build scripts be they Docker or GH actions _never_ work on the first (or second, or third) try and I’d much rather develop them locally than have to awkwardly force-push commits over and over until I get a green checkmark.

.github/workflows/dekaf.yaml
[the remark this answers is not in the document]
jgraettinger: > However, this approach means you will not automatically receive updates for an action, including important bug fixes and security updates.

Yep, this is exactly the reason not to do this. Using fixed versions would be appropriate if and only if we were actually auditing the release and were willing to pay the cost of more intensive dependency update and re-verification cycles.

The far greater practical risk is unknowingly perpetuating a bug or vulnerability that's already been fixed for two years because we're not regularly updating dependencies.

As with my comment below, dependencies are long-term commitments, and the author of a dependency matters more when evaluating risk. `docker/` is fine. `leetHaxor/` isn't. Pinning a hash is only a temporary salve that buys immediate reproducability in exchange for deferred maintenance.

Obviously I'm addressing this question in a broader context than just GitHub actions. AFAIK the best we can do, as a community, is Cargo.toml / go.mod for expressive permissive / evergreen semver, paired with Cargo.lock / go.sum for pinned and fully reproducible builds, and automated tooling for fetching updates.

crates/dekaf/Cargo.toml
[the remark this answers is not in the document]
jgraettinger: I'm a little baffled here. Presume a goal of a well-structured code base with an intuitive dependency graph. Turning the question around, finish the sentence "It's reasonable for `dekaf` to depend on our end-user CLI command because ___".

Cargo.toml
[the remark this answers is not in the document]
jshearer: @jgraettinger I'm curious what you think about [this](https://users.rust-lang.org/t/psa-please-specify-precise-dependency-versions-in-cargo-toml/71277):

> ```
> serde = "1.0.130"  # good!
> ```
> Cargo works well with versions specified this way. It is future-proof: the specified version is only the the minimum version, and still allows upgrades of minor versions. Cargo also unifies versions with the same semver-major number, so when two crates specify different versions, the higher version will be used for both.

From reading that (long...) thread, my take-away is that specifying at least minor versions (if major >= 1) is important as it prevents cargo from picking a minor version _lower_ than the one specified, while still allowing it to pick newer minor/patch (but not major) versions as needed to come up with a dependency solution. Downgrading a minor (or patch is major = 0) version is bad [because](https://semver.org/spec/v2.0.0.html#summary):

> increment the MINOR version when you add functionality in a backward compatible manner

So therefore, _decrementing_ the minor version would result in a loss of functionality that's potentially used by the code that depends on it.

crates/dekaf/Dockerfile
jgraettinger: I'm going to not wade in further to protect my time. Please junk the dockerized build. Make `dekaf` a target of the `linux-binaries` make rule. You can include it into the singular `flow` docker image we push today, or create a new minimal docker image that wraps the pre-built binary.

.github/workflows/dekaf.yaml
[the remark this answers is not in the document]
jshearer: Switched to `docker/metadata-action@v5` 👍

crates/dekaf/Cargo.toml
```diff
@@ -15,25 +15,49 @@ doc = { path = "../doc" }
...
 simd-doc = { path = "../simd-doc" }
 
 anyhow = { workspace = true }
 axum = { workspace = true }
+axum-server = { workspace = true }
 axum-extra = { workspace = true }
 base64 = { workspace = true }
 bytes = { workspace = true }
-clap = { workspace = true }
+# TODO(jshearer): Upgrade every other usage of clap to v4 and change this to:
```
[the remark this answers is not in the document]
jshearer: Upgraded all usage of `clap` to v4 -- I had wanted to avoid scope creep of an already fairly big PR, but I see your point about there being no better time than right now to pay down tech debt 👍

crates/dekaf/src/read.rs
```diff
@@ -119,7 +172,27 @@ impl Read {
 
             // Encode the key.
             let key = if is_control {
-                None
+                // From https://cwiki.apache.org/confluence/display/KAFKA/KIP-98+-+Exactly+Once+Delivery+and+Transactional+Messaging
```
jshearer, line 175: @jgraettinger I did more digging into how control messages are implemented, and came up with this. Materialize is able to read through the end of a slowly changing collection, and Tinybird doesn't complain about null documents. Anything using librdkafka will show debug messages like this:

```
%7|1726574941.686|MSG|rdkafka#consumer-2|: demo.2Fwikipedia.2Frecentchange.2Dsampled [0]: Skipping ctrl msg with unsupported version 9999 at offset 14536044024
```

But other than that, completely ignore them. I think this is the ideal solution.

crates/dekaf/src/read.rs
jshearer: Confirmed! See https://github.com/estuary/flow/pull/1516#discussion_r1763135832

Cargo.toml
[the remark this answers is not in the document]
jshearer: I left it at `1.0` for now

crates/dekaf/Cargo.toml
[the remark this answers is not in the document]
jshearer: I think I'd like to leave it for now, just to avoid creeping the scope of this PR even more. It's a pretty tight dependency on a single isolated module within `flowctl`. Prior to this PR, `flowctl` is the only crate that directly called out to PostgREST, so it made sense to include utilities such as pagination directly in it. In terms of well-structured code with an intuitive dependency graph, now that we have two "leaf" crates that need to share this functionality, it does make more sense to move it to somewhere shared by both.

PR #1518, opened 2024-07-11
Phil/db perf

psFried:
**Description:**

Rolls up a few different all-natural performance enhancements. The individual commit messages have more details. There's likely more PRs to come, but this seemed like a useful enough set of improvements to be worth deploying.

The `minTxnDuration` from the first commit was already applied manually, so this just brings the checked-in ops catalog in line with that.

I tested this by running the current agent while applying the migration, verifying I can still publish and use the UI with the current agent, and then upgrading to the new agent version and re-testing via the UI.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1518)

### DIFF
--- a/crates/agent-sql/src/publications.rs
+++ b/crates/agent-sql/src/publications.rs
@@ -692,25 +692,29 @@ pub struct StorageRow {
     pub spec: serde_json::Value,
 }
 
+/// Returns the storage mappings for the given set of tenants.
+/// Mappings for `recovery/{tenant}` will also be returned.
 pub async fn resolve_storage_mappings(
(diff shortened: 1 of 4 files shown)


psFried:
One note about the new indexes to support dequeuing jobs: The `controller_jobs` table already had a similar partial index, and the `mean_exec_time` for those is in the single-digit ms. So I just copied the same approach on the jobs that had slower queries for dequeuing. I didn't add indexes for `applied_directives` or `connector_tags` because those are already quite fast.

jgraettinger (approved):
LGTM

PR #1524, opened 2024-07-16
agent: prevent auto-discovers overwriting endpoint

### LINKED ISSUE #1521: Auto-discovers can revert changes to `endpoint`
Background discovers get prioritized lower than interactive publications. Thus it's possible for a background discover to be created, but not immediately processed. Then an interactive publication updates the `endpoint` of the capture before the discover is processed. When the background discover completes, it overwrites the `endpoint` with the now outdated value from the `discovers` row.

psFried:
Fixes #1521 

Background discovers get prioritized behind interactive publications, so it's possible for an interactive publication to update the endpoint config after the `discovers` row has already been created but before it's been processed. This detects that condition and bails, so we can try again on the next auto-discover.

I tested this manually, by creating `discovers` rows and modifying the endpoint config, and getting the expected error on the auto-discover. I also tested that regular auto-discovers still work when the endpoint isn't modified.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1524)

### DIFF
--- a/crates/agent/src/discovers.rs
+++ b/crates/agent/src/discovers.rs
@@ -3,6 +3,7 @@ use super::{
 };
 use agent_sql::discovers::Row;
 use anyhow::Context;
+use models::CaptureEndpoint;
 use serde::{Deserialize, Serialize};
 use sqlx::types::Uuid;
 
@@ -194,6 +195,7 @@ impl DiscoverHandler {
             &row.image_name,
             &row.image_tag,
             row.update_only,
+            row.background,
             row.user_id,
             txn,
         )
@@ -273,6 +275,7 @@ impl DiscoverHandler {
         image_name: &str,
         image_tag: &str,
         update_only: bool,
+        background: bool,
         user_id: Uuid,
         txn: &mut sqlx::Transaction<'_, sqlx::Postgres>,
     ) -> anyhow::Result<Result<models::Catalog, Vec<draft::Error>>> {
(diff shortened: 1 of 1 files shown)


crates/agent/src/discovers.rs
```diff
@@ -335,11 +338,28 @@ impl DiscoverHandler {
 
         // Deeply merge the capture and its bindings.
         let capture_name = models::Capture::new(capture_name);
+        let existing_capture = catalog.captures.remove(&capture_name);
+        // This is a hack to prevent autoDiscovers from overwriting changes to the capture endpoint.
```
jshearer, line 342: Why is it a hack? What's the non-hacky way to do this?

crates/agent/src/discovers.rs
```diff
@@ -335,11 +338,28 @@ impl DiscoverHandler {
...
+        // Background discovers get prioritized behind interactive publications, so it's possible
+        // for an interactive publication to update the endpoint config after the `discovers` row
+        // has already been created but before it's been processed. This detects that condition and
+        // bails, so we can try again on the next auto-discover.
+        if existing_capture
+            .as_ref()
+            .is_some_and(|spec| background && is_endpoint_changed(&spec.endpoint, &endpoint))
```
jshearer, line 349: Why wouldn't we want this to apply to all publications? That is: if the proverbial Alice starts a publication that changes the endpoint config, and Bob starts another one that doesn't before Alice's publication finishes, won't Alice's change get clobbered? Feels like bob's publication should fail with a retryable error in this case (and maybe even auto-retry in the future?)

jshearer (approved):
This LGTM, though I did have a question about potentially broadening its scope. 👍

crates/agent/src/discovers.rs
psFried: The non-hacky way is to re-implement auto-discovers in controllers. The hack is because that'll take a while

crates/agent/src/discovers.rs
psFried: If we did it for all publications, it would break when you edit a capture's endpoint config and then hit the "refresh bindings" button. The alice/bob scenario is legit a problem (though it's less of a problem with interactive publications because they generally don't sit in queued status anywhere near as long). But it's not new, and not exacerbated by this hack. Ultimately, we need discovers to set the `expect_pub_id` to protect against that sort of scenario. It doesn't today, partly because a `models::Catalog` historically had no way of representing it.

crates/agent/src/discovers.rs
jshearer: Ahhhh, okay that makes sense: auto-discovers aren't controllers yet, that's why this is a problem in the first place. Otherwise this would be taken care of by the controllers' new transaction-less updates mechanism

crates/agent/src/discovers.rs
jshearer: >If we did it for all publications, it would break when you edit a capture's endpoint config and then hit the "refresh bindings" button.

Oh... yeah good point.

>But it's not new, and not exacerbated by this hack.
Yep, sounds good 👍

jshearer (approved):

PR #1530, opened 2024-07-18
control-plane: add user_info_summary sql function

psFried:
**Description:**

Introduces the `user_info_summary` sql function, which returns a high level summary of user authorizations. Determining this information requires checking all of a users authorization grants, which can take ages for the UI because it has to paginate. The function can do this quickly, even for users with many roles. This is intended to be used by the UI.

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1530)

### DIFF
--- a/supabase/migrations/54_user_info_summary.sql
+++ b/supabase/migrations/54_user_info_summary.sql
@@ -1,0 +1,27 @@
+begin;
+
+create function user_info_summary()
+returns json as $$
+    with all_grants(role_prefix, capability) as (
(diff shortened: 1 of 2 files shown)


jshearer (approved):
Phil and I talked in a DM about this -- I would usually prefer to avoid adding a narrow/"custom" API like this, but in this case the whole point is that the general-purpose API we currently have (PostgREST) isn't capable of serving this request as efficiently as a custom endpoint is, so adding a once-off api is appropriate here.

PR #1532, opened 2024-07-23
agent: limit frequency of controller publications

### LINKED ISSUE #1520: inferred schema updates blocking interactive publications of captures and materializations
Say you have a capture and materialization with many bindings that use inferred schemas. When the control plane publishes the inferred schemas, it increases the `last_pub_id` of the collection. If a user goes into the UI to try to publish the capture or materialization, the collection gets added to the `LiveCatalog` with a `last_pub_id` that is potentially greater than the `id` of the current publication. This results in [this error being returned](https://github.com/estuary/flow/blob/8f5145eb5742328bfffc1bbbc82535735ce1180c/crates/validation/src/lib.rs#L211).

## Current plan (updated after discussion)

- [x] (#1623) Switch to using `last_build_id` instead of `last_pub_id` for optimistic locking
- [x] (#1623) Introduce the concept of a "touch" publication, which only updates the `built_spec`, leaving the `spec` and `last_pub_id` unchanged
- [x] (#1623) Controllers will hash the `last_pub_id`s of all their dependencies in order to determine whether they need to do a touch publication to update their own `built_spec`.
    - This prevents the `last_pub_id` from being incremented when the dependencies of a spec have changed, thus avoiding `ExpectPubIdNotMatched` errors caused by that.
- [ ] (PR tbd) Update the publications handler to always inject the current inferred schema into the models of drafted (non-touch) collections.
    - This allows the UI and `flowctl` to stop setting `expectPubId` on collection specs. They must set it currently to avoid clobbering inferred schema changes, but this will make clobbering inferred schemas impossible, so `expectPubId` will no longer be needed.
- [ ] (PR tbd) Update the UI to no longer set `expectPubId` for collection specs when editing captures.
    - It must still set it for the capture specs, though.
    - This avoids the possibility for `expectPubId` conflicts on the collections
- [ ] (PR tbd) Update the publications handler to generate a new `pub_id` for each publication instead of using the `id` of the `publications` row, and to retry `PublicationSuperseded` errors
    - Inventing a new `pub_id` is what allows us to retry these errors by generating a new (higher) `pub_id`

psFried:
**Description:**

Fixes #1520 

We've observed a scenario where there's a very large capture with hundreds of collections all using schema inference. Among all of the collections, publications due to inferred schema updates are quite frequent. Because each collection publication results in a subsequent publication of the capture and materialization, it results in extremely frequent publications of those tasks, which has two important consequences:

- It can prevent user-initiated publications from completing successfully due to `PublicationSuperceded` or `ExpectPubIdNotMatched` errors.
- It can prevent the materialization from making progress, because every time it's activated it must re-start any in-progress reads of source journals. There may also be a data-plane bug that's affecting this, but the frequent publications are certainly a factor.

So we've decided to slow down the rate of controller-initiated publications to try to alleviate those symptoms. This introduces two new fields to the publication status: `last_publication_time` and `min_publication_interval`. Controllers will now wait at least `min_publication_interval` since the `last_publication_time` before publishing a spec in response to publication of dependencies.

The default value of `min_publication_interval` is 30 minutes, which is somewhat arbitrary, but aligns with the default `update_delay` of materializations. Since this value is part of the controller status, we can update it on a per-task basis.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1532)

### DIFF
--- a/crates/agent/src/controllers/capture.rs
+++ b/crates/agent/src/controllers/capture.rs
@@ -1,9 +1,7 @@
 use super::{
-    backoff_data_plane_activate,
-    publication_status::{ActivationStatus, PendingPublication},
-    ControlPlane, ControllerErrorExt, ControllerState, NextRun,
+    backoff_data_plane_activate, ActivationStatus, ControlPlane, ControllerErrorExt,
+    ControllerState, NextRun, PendingPublication, PublicationStatus,
 };
-use crate::controllers::publication_status::PublicationStatus;
 use anyhow::Context;
 use itertools::Itertools;
 use schemars::JsonSchema;
@@ -83,7 +81,7 @@ impl CaptureStatus {
                 .context("failed to notify dependents")?;
         }
 
-        Ok(None)
+        Ok(dependencies.next_run)
     }
 }
 
--- a/crates/agent/src/controllers/catalog_test.rs
+++ b/crates/agent/src/controllers/catalog_test.rs
@@ -39,13 +39,13 @@ impl TestStatus {
             // return a terminal error if the publication failed
             result.error_for_status().do_not_retry()?;
             // TODO(phil): This would be a great place to trigger an alert if the publication failed
-        } else {
+        } else if dependencies.next_run.is_none() {
             // We're up-to-date with our dependencies, which means the test has been published successfully
             self.passing = true;
         }
 
         // Don't re-try when tests fail, because fixing them will likely require a change to either
         // the test itself or one of its dependencies.
-        Ok(None)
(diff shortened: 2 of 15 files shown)


jgraettinger (approved):
LGTM (I didn't review deeply but makes sense)

psFried:
After further e2e testing, it seems that this approach does not really help overall. It just shifts the symptoms so that materializations will stay in a failed state for a longer time. We've discussed another option, which is to ensure that inferred schema updates are published _more_ promptly, and I'm closing this PR in favor of that.

PR #1533, opened 2024-07-24
agent: fix retry of materialization publications

psFried:
**Description:**

The materialization controller would fail to retry publications if they failed for reasons other than unsatisfiable constraints. This plugs the hole, and adds a publication failure/retry to the `dependencies_and_activations` integration tests.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1533)

### DIFF
--- a/crates/agent/src/controllers/materialization.rs
+++ b/crates/agent/src/controllers/materialization.rs
@@ -80,7 +80,7 @@ impl MaterializationStatus {
         }
 
         if pending_pub.has_pending() {
-            let result = pending_pub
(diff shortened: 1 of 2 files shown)


jshearer (approved):
Took me a sec to read the diff on github, but I see the problem now: the call to `with_maybe_retry` was inside the `if result.status.has_incompatible_collections()` block so we'd only (maybe) retry if we had incompatible collections, instead of always.

LGTM, and I love the test harness infra here! :+1:

PR #1537, opened 2024-07-25
doc: limit maximum depth of schema inference

psFried:
**Description:**

Introduces a `depth_limit` to `enforce_shape_complexity_limit` to prevent inferred schemas from becoming too deeply nested. Deeply nested json can present problems, as essentially all parsers have a limit on the amount of nesting that they will accept. We've found that deeply nested documents will result in schemas with roughly double the levels of nesting because of things like `properties: { ... }`. And deeply nested JSON schemas also just aren't that useful in Flow, anyway, since if someone has 64 layers of nesting, they likely don't want us flattening out all their fields into separate columns. So this sets the limit to 63 levels of nesting, which was chosen as the most conservative value that results in schemas that can be parsed by `serde-json`.

**Also fixes CI**

Rust 1.80 was released recently, and CI builds are using it by default. Unfortunately, there's a regression (I'm assuming, anyway) with the `time` crate failing to compile using rust 1.80. Updating the `time` dependency by itself wouldn't work because it's also used by other dependencies transitively. So I did a `cargo update` and that seems to have worked.

One issue with that was that the `uuid` crate updated, and seems to have fixed a bug with the previous version where `uuid.get_timestamp()` now returns a timestamp _with_ the counter value. The test was expecting there to be no counter value.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1537)

### DIFF
--- a/crates/doc/src/shape/limits.rs
+++ b/crates/doc/src/shape/limits.rs
@@ -123,22 +123,30 @@ fn squash_location(shape: &mut Shape, location: &[Token]) {
 
 /// Reduce the size/complexity of a shape while making sure that all
 /// objects that used to pass validation still do.
-pub fn enforce_shape_complexity_limit(shape: &mut Shape, limit: usize) {
-    let mut pointers = shape
-        .locations()
-        .into_iter()
-        .filter_map(|(ptr, _, _, _)| match ptr.0.as_slice() {
+/// The `limit` limits the total number of properties, and the `depth_limit` removes any
+/// properties after more than `depth_limit` levels of nesting.
+pub fn enforce_shape_complexity_limit(shape: &mut Shape, limit: usize, depth_limit: usize) {
+    let locations = shape.locations();
+    let mut pointers = Vec::with_capacity(locations.len());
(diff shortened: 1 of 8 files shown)


crates/doc/src/shape/limits.rs
```diff
@@ -413,8 +430,28 @@ mod test {
...
+                    type: object
+                    additionalProperties:
+                        type: object
+                foo9:
+                    type: object
+                    additionalProperties:
+                        type: object
+            "#,
+            docs,
+            Some((99, 1)),
```
jgraettinger, line 454: Perhaps add `Some((99, 2))` here as well, as a case?

I'm wondering if there are other ways we can bang on this. I do have high confidence in the change, but a mistake here would be rather painful to recover from.

jgraettinger (approved):
LGTM % testing comment below

crates/doc/src/shape/limits.rs
psFried: Yeah, good call. I added a couple more test cases here, including some that exercise arrays.

PR #1538, opened 2024-07-25
Add SSO related `attribute mapping` files into Supabase directory

travjenkins:
**Description:**

These attribute mappings are updated via the CLI and stored in Supabase DB. However, an easy way to update them is with a CLI command that points to the file.

**Workflow steps:**

N/A

N/A

**Notes for reviewers:**

N/A

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1538)


psFried (approved):
Looks good, thanks for adding the readme :+1:

PR #1540, opened 2024-07-30
Phil/schema bundling

psFried:
**Description:**

This affects only the json schema bundling that's used when updating collection models to bundle in the inferred schemas. Previously, this function would bundle in the write schema, too. This makes that optional so that we no longer inline the write schema as part of the read schema `$defs`.

**Workflow steps:**

No change to what users _do_, only what they see as part of collection specs. Once this is deployed, controllers will start to remove bundled write schemas from every collection that uses schema inference. The periodic controller runs for such collections are every 3-4 hours, so after ~4 hours all collections should be updated.

**Notes for reviewers:**

The idea is to revert the entire second commit after we're satisfied that all the collections have been updated.

One other note is that I checked the existing `live_specs` and confirmed that all existing collection specs have either 0, 1, or 3 occurrences of the substring `flow://write-schema`. So that hacky check in `read_schema_bundles_write_schema` seems like it'll probably work OK.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1540)

### DIFF
--- a/crates/agent/src/controllers/collection.rs
+++ b/crates/agent/src/controllers/collection.rs
@@ -218,10 +218,7 @@ fn update_inferred_schema(
         let Some(read_schema) = model.read_schema.as_ref() else {
             anyhow::bail!("model is missing read schema");
         };
-        let Some(write_schema) = model.write_schema.as_ref() else {
-            anyhow::bail!("model is missing write schema");
-        };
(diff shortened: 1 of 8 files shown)


jgraettinger (approved):
LGTM

PR #1546, opened 2024-08-07
Phil/fix auto evolve

psFried:
**Description:**

Rolls up two agent fixes, best reviewed independently.
First is a fix for evolutions in auto-discovers.
Second is a fix for auto-discovers happening too frequently.


jgraettinger (approved):
LGTM -- i went fast and not deep here.

PR #1550, opened 2024-08-12
control-plane: delete old data with multiple jobs

psFried:
**Description:**

Previously we used the `internal.delete_old_rows()` function to cleanup data that was past its retention period. This function takes long enough that it frequently times out, leaving old data in place and exacerbating the timeout issue, leading to no old data being cleaned up.

Testing has shown that the solution is simply to run the delete commands much more frequently, as it runs significantly faster when there are fewer rows to delete. Additionally, I split up the `delete_old_rows` function into separate functions for each delete statement, in order to prevent timeouts/errors in one from affecting the others.

While doing this, I also came across an important detail in the pg_cron docs: it does not automatically clean up old `cron.job_run_details`. Given the significantly increased frequency of cron job runs introduced with this migration, I thought it wise to also add a job to regularly clean up old run details as recommended in the pg_cron docs.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1550)

### DIFF
--- a/supabase/migrations/56_delete_old_rows_faster.sql
+++ b/supabase/migrations/56_delete_old_rows_faster.sql
@@ -1,0 +1,79 @@
+begin;
+
+-- Delete old drafts, which will cascade to draft_specs, draft_errors, and discovers
+create function internal.delete_old_drafts()
+returns integer as $$
+    with d as (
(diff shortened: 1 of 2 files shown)


jshearer (approved):
LGTM 👍

PR #1554, opened 2024-08-14
protocols/flow: add numeric inference to protocol

williamhbaker:
**Description:**

Adds "numeric" information to projection inference, which will include the approximate minimum and maximum for the numeric values observed if available.

This change only adds the fields to the protocol, and does not actually add the values to the sent messages yet. The reason for this is not very satisfying: Currently CI tests are run using the `source-typescript:dev` image, and that derivation connector will fail if messages contain unknown fields. Since we are updating the protocol and structs that it uses that would normally be fine, but the only way to get a new `source-typescript:dev` image to publish is to have it pass CI tests, so it's kind of a catch-22. I couldn't figure out an easy way around this so the plan is to merge this change which adds the fields to the protocol only, and then follow-up with a change that starts populating the fields.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1554)


crates/proto-flow/tests/regression.rs
```diff
@@ -25,6 +25,7 @@ fn ex_projections() -> Vec<flow::Projection> {
                 max_length: 12345,
             }),
             types: vec!["integer".to_string(), "string".to_string()],
+            numeric: None,
```
jgraettinger, line 28: nit: you could use an actual fixture here, to ensure you don't forget to do so later.

go/protocols/flow/flow.proto
```diff
@@ -121,6 +121,17 @@ message Inference {
   bool secret = 7;
   // Existence of this document location.
   Exists exists = 8;
+  // Numeric type-specific inferences, or nil iff types doesn't include
+  // "integer" or "number".
+  message Numeric {
+    // Minimum value when the projection is of "integer" or "number" type. Zero
```
jgraettinger, line 127: Perhaps reconsider the use of zero as a "not set" sentinel, as it can be quite lossy.

Suppose minimum is zero and maximum is 1,000. That could _either_ be a non-zero integer, or an integer with an unlimited negative bound, and a materialization might want to make a wildly different choice in those two conditions (use of an unsigned value, for example).

One cute thought that's probably a bad idea is to use NaN, assuming that serializes through serde into JSON okay (not confident here). Another would be to use a max/min-valued double for maximum and minimum, respectively. Or booleans ? Or split these into separate null-able messages ?

Finally, there will also be pre-existing built specs that _do_ have integer / number but don't have this of course. I assume you're already thinking about migration implications but just saying it out loud.

williamhbaker: Yeah it's probably best to be more explicit here. Nothing that I'm currently trying to do with materialization connectors cares what the difference between a real 0 and unset is, but it could come up in the future.

I generally like the idea of NaN as a sentinel but that seems to break our own serde tests so it doesn't seem like that will work out. So I added a couple additional boolean fields of `has_minimum` and `has_maximum` to indicate if the values are really 0 or are just not set.

The note about pre-existing specs is good since it means that it's not actually true that the `Numeric` field will be set if the types include integer / number, which is a somewhat different assumption than can be made about the presence of the `String` field in the inference message. I updated the comment to be more reflective of that.

jgraettinger (approved):
LGTM

PR #1555, opened 2024-08-14
doc: fix numeric bounds in schema reductions

psFried:
**Description:**

This fix addresses the following scenario with inferred schemas:
- The captured data has a mix of integer and float values.
- The `minimum` and `maximum` values in the inferred schemas thus also have a mix of integer and float values, where the values themselves are equal except for some have a `.0` decimal and others do not.
- When we reduce the inferred schemas, the result will arbitrarily select either the decimal or integer value, as it considers them to be equal.

A consequence of that behavior is that it's possible for an inferred schema to flip back and forth between the decimal and integer representations, causing unnecessary publications of the inferred schema.

The fix is to adjust the behavior of `NumericShape::union`, which is used for schema reductions. This updates it to always prefer a decimal value over an integer value in cases where the two are equal. This makes it consistent with the widening behavior of considering `number` to be "wider" than `integer` types.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1555)

### DIFF
--- a/crates/doc/src/shape/union.rs
+++ b/crates/doc/src/shape/union.rs
@@ -205,9 +205,30 @@ impl NumericShape {
             },
         ) = (lhs, rhs);
 
-        // Take the least-restrictive bounds of both.
-        let min = lhs_min.and(rhs_min).and(lhs_min.min(rhs_min));
-        let max = lhs_max.and(rhs_max).and(lhs_max.max(rhs_max));
+        // Take the least-restrictive bounds of both. We must be careful here to always prefer
+        // a float value over an integer value if the two are otherwise equal (the `Ord` impl for
+        // `Number` ignores such differences). For example, we prefer 5.0 over 5. This ensures that
+        // the minimum/maximum values are consistent with the `type` in inferred schemas.
+        let min = if let Some((lmin, rmin)) = lhs_min.zip(rhs_min) {
+            match lmin.cmp(&rmin) {
+                std::cmp::Ordering::Less => Some(lmin),
+                std::cmp::Ordering::Equal if lmin.is_float() => Some(lmin),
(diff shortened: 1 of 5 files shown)


crates/doc/src/shape/union.rs
```diff
@@ -205,9 +205,30 @@ impl NumericShape {
...
 
-        // Take the least-restrictive bounds of both.
-        let min = lhs_min.and(rhs_min).and(lhs_min.min(rhs_min));
-        let max = lhs_max.and(rhs_max).and(lhs_max.max(rhs_max));
+        // Take the least-restrictive bounds of both. We must be careful here to always prefer
+        // a float value over an integer value if the two are otherwise equal (the `Ord` impl for
+        // `Number` ignores such differences). For example, we prefer 5.0 over 5. This ensures that
+        // the minimum/maximum values are consistent with the `type` in inferred schemas.
+        let min = if let Some((lmin, rmin)) = lhs_min.zip(rhs_min) {
+            match lmin.cmp(&rmin) {
```
jgraettinger, line 213: nit: lift into min and max helpers, and then also use for implementing intersection ?

jgraettinger (approved):
LGTM, but I do think we should give intersection the same treatment for consistency.

crates/doc/src/shape/union.rs
psFried: The intersection logic is the inverse of this, in that it prefers integers over floats. This is to be consistent with the tightening of constraints, since the intersection of `number & integer` is `integer`. So I didn't bother factoring out a separate function.

jgraettinger: :+1: fair!

jay-growflow:
💜

Issue #1552, opened 2024-08-12
Add new column to handle if a connector supports `interval` prop


travjenkins:
### Request
The UI wants a way to know if it should show the `interval` property while creating/editing a capture. In a Slack conversation we came up with adding a new column to handle this. 

There is an idea where we could store the `default value` for the field in that column. That way we can use it for more than just controlling the input. Also, this allows certain connectors to recommend an interval.

### Related
https://github.com/estuary/ui/issues/1214

PR #1556, opened 2024-08-14
Add two new columns to `connectors` for UI to show/hide properties

travjenkins:
**Description:**

Added two new columns to the `connectors` table. 

https://github.com/estuary/ui/issues/1194 / https://github.com/estuary/flow/issues/1552
`capture_interval`
when `null` the UI will hide the interval input on the Capture edit/create workflow
when a value the UI will show the interval input and default it to the provided value

https://github.com/estuary/ui/issues/1219
`disable_backfill`
when `false` the UI will show backfill
when `true` the UI will hide backfill

**Workflow steps:**

Run a migration on Supabase

n/a

**Notes for reviewers:**

Ran `supabase stop` and `supabase start` to make sure the script ran
```
...
Applying migration 48_storage_mapping_directive.sql...
Applying migration 49_tenant_hide_preview.sql...
Applying migration 50_controllers.sql...
Applying migration 51_backfill_controllers.sql...
Applying migration 52_extend_pg_net_timeout.sql...
Applying migration 53_rls_perf.sql...
Applying migration 54_user_info_summary.sql...
Applying migration 55_inferred_schema_triggers.sql...
Applying migration 56_connector_tags_control_flags.sql...
Seeding data supabase/seed.sql...
Started supabase local development setup.
...

```
image

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1556)


psFried (changes requested):
Unfortunately, I think you've been led astray by our out-of-date docs on migrations. Sorry. The way to do migrations now is to add a new migration and run `alter table` commands to add the columns. You might try using the `pgadmin` workflow from those docs in order to generate the migration, though I'm not sure whether it still works (which is why we switched workflows).

I think we might also want these columns on `connector_tags` instead of connectors. We have things like `autodiscover_interval` on `connector_tags`, and having it there allows us to set it automatically if we ever want to. Would that work from a front-end perspective?

psFried (approved):
LGTM, thanks!

PR #1559, opened 2024-08-15
agent: switch to jemalloc allocator

psFried:
**Description:**

Switches `agent` to use `jemalloc` from the allocator crate. This is being done in order to facilitate debugging of memory issues using jemalloc's built-in profiling features.

This enables the use of the `_RJEM_MALLOC_CONF` env variable for selectively enabling profiling. For example: `_RJEM_MALLOC_CONF=prof_leak:true,lg_prof_sample:0,prof_final:true`, which is recommended in the [jemalloc docs on leak detection](https://github.com/jemalloc/jemalloc/wiki/Use-Case%3A-Leak-Checking). (The `_RJEM_` prefix is added by the `jemalloc-sys` crate.)

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1559)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -62,6 +62,7 @@ version = "0.0.0"
 dependencies = [
  "activate",
  "agent-sql",
+ "allocator",
  "anyhow",
  "async-process",
  "async-trait",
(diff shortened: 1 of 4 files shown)


jshearer (approved):
LGTM!

PR #1561, opened 2024-08-15
control-plane: fix pruning of unchanged specs

psFried:
This is a quick fix for the pruning of unchanged draft specs. It stops trying to account for inferred schema changes, since those are now part of the spec itself, and the `inferred_schema_md5` column is no longer used. (We can remove it in a subsequent migration).

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1561)


jgraettinger (approved):
LGTM

PR #1562, opened 2024-08-15
doc: non-fractional floats which exceed integer ranges are numbers

jgraettinger:
Previously, we would map `32.0` or `100.0` or `3.14159e100` into `type: integer`, because they have a zero'd out fractional component. While _technically_ this is correct, in practice there's a big difference between the first examples and the last: `3.14159e100` cannot fit into a native 64-bit integer representation.

This is a problem downstream because materializations use `type: integer` as an indication to use a native integer representation, and then fail when attempting to fit the float into such a column.

Instead, infer such values to be `number` if they're too big to fit into a native integer.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1562)

### DIFF
--- a/crates/doc/src/shape/widen.rs
+++ b/crates/doc/src/shape/widen.rs
@@ -292,11 +292,13 @@ impl Shape {
             Node::Null => apply_type(types::NULL),
             Node::Float(f) => self.numeric.widen(
                 Number::Float(f),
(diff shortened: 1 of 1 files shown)


williamhbaker (approved):
LGTM

PR #1566, opened 2024-08-19
Doc updates for egress IP addresses

skord:
**Description:**

This just updates docs to reflect the three current possible egress IP's for Flow. 

**Workflow steps:**

It's all documentation, so every doc in the commit.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1566)

### DIFF
--- a/site/docs/concepts/connectors.md
+++ b/site/docs/concepts/connectors.md
@@ -54,7 +54,7 @@ or any other public image registry provider.
 To interface with a connector, the Flow runtime needs to know:
 
 1. The specific image to use, through an image name such as `ghcr.io/estuary/source-postgres:dev`.
-  Notice that the image name also conveys the specific image registry and version tag to use.
+   Notice that the image name also conveys the specific image registry and version tag to use.
 
 2. Endpoint configuration such as a database address and account, with meaning that is specific to the connector.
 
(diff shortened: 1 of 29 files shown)


site/docs/concepts/connectors.md
```diff
@@ -286,7 +288,7 @@ In some cases, your source or destination endpoint may be within a secure networ
 to allow direct access to its port due to your organization's security policy.
 
 :::tip
-If permitted by your organization, a quicker solution is to whitelist the Estuary IP address, `34.121.207.128`.
+If permitted by your organization, a quicker solution is to whitelist the Estuary IP addresses, `34.121.207.128, 35.226.75.135, 34.68.62.148`.
```
travjenkins, line 291: I think from a UI/UX aspect breaking these up into a small list that has one IP per line could make sense. However, that would make them a lot harder to copy/paste the list.

From an ops perspective would keeping these as a comma separated list make it easier to configure on the client's system?

image

skord: image

This is actually kind of tricky. I moved these into a code block for where it's in a tool tip and it looks like the attached. 

In other cases, I just added a `:` so it was clear a list was coming. Trying to make a UL in markdown for other cases was very visually jarring. The other option was to put something like 


`34.121.207.128, 35.226.75.135` and `34.68.62.148` 

for proper English, but the flow on that is gross too.

travjenkins: Looks good to me and follows the pattern up above... though that is kinda different given you _really_ should copy the value.

image

travjenkins (approved):
LGTM

Tested the main changes on pages with the new IPs (just searched for `35.226.75.135` and hit those pages). 

Hit a few other pages that changed due to linting and they all seem to match prod.

PR #1572, opened 2024-08-22
flow: x-schema-name annotation for 2nd-to-last resource path

mdibaiee:
**Description:**

- Sets the base for supporting threading through of the schema name from the second-to-last part of collection name
- Added tests for both cases (stub config) and sourceCapture

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1572)

### DIFF
--- a/crates/agent/src/controllers/materialization.rs
+++ b/crates/agent/src/controllers/materialization.rs
@@ -9,7 +9,7 @@ use crate::{
 };
 use anyhow::Context;
 use itertools::Itertools;
-use models::{ModelDef, OnIncompatibleSchemaChange, SourceCaptureDef};
+use models::{ModelDef, OnIncompatibleSchemaChange, SourceCapture};
 use proto_flow::materialize::response::validated::constraint::Type as ConstraintType;
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
@@ -66,10 +66,7 @@ impl MaterializationStatus {
                 dependencies.live.captures.get_by_key(&source_capture.capture_name())
             {
                 if self.source_capture.is_none() {
-                    self.source_capture = Some(SourceCaptureStatus {
-                        source_capture: source_capture.def(),
-                        ..Default::default()
-                    });
+                    self.source_capture = Some(SourceCaptureStatus::default())
                 }
                 let source_capture_status = self.source_capture.as_mut().unwrap();
                 // Source capture errors are terminal
@@ -293,7 +290,6 @@ fn is_false(b: &bool) -> bool {
 /// Status information about the `sourceCapture`
 #[derive(Debug, Clone, Serialize, Deserialize, Default, PartialEq, JsonSchema)]
 pub struct SourceCaptureStatus {
-    pub source_capture: SourceCaptureDef,
     /// Whether the materialization bindings are up-to-date with respect to
     /// the `sourceCapture` bindings. In normal operation, this should always
     /// be `true`. Otherwise, there will be a controller `error` and the
@@ -375,7 +371,7 @@ impl SourceCaptureStatus {
         draft_row.is_touch = false;
 
         update_linked_materialization(
-            &self.source_capture,
+            model.source_capture.as_ref().unwrap(),
             resource_spec_pointers,
             &self.add_bindings,
             draft_row.model.as_mut().unwrap(),
@@ -404,7 +400,7 @@ fn get_bindings_to_add(
 }
 
 fn update_linked_materialization(
-    source_capture: &SourceCaptureDef,
+    source_capture: &SourceCapture,
     resource_spec_pointers: ResourceSpecPointers,
     bindings_to_add: &BTreeSet<models::Collection>,
     materialization: &mut models::MaterializationDef,
--- a/crates/agent/src/controllers/materialization.rs
+++ b/crates/agent/src/controllers/materialization.rs
@@ -66,7 +66,7 @@ impl MaterializationStatus {
                 dependencies.live.captures.get_by_key(&source_capture.capture_name())
             {
                 if self.source_capture.is_none() {
-                    self.source_capture = Some(SourceCaptureStatus::default())
+                    self.source_capture = Some(SourceCaptureStatus::default());
                 }
                 let source_capture_status = self.source_capture.as_mut().unwrap();
                 // Source capture errors are terminal
--- a/crates/agent/src/controllers/materialization.rs
+++ b/crates/agent/src/controllers/materialization.rs
@@ -9,7 +9,7 @@ use crate::{
 };
 use anyhow::Context;
 use itertools::Itertools;
-use models::{ModelDef, OnIncompatibleSchemaChange};
+use models::{ModelDef, OnIncompatibleSchemaChange, SourceCaptureDef};
 use proto_flow::materialize::response::validated::constraint::Type as ConstraintType;
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
@@ -59,14 +59,17 @@ impl MaterializationStatus {
             }
         }
 
-        if let Some(source_capture_name) = &model.source_capture {
+        if let Some(source_capture) = &model.source_capture {
             // If the source capture has been deleted, we will have already handled that as a
             // part of `handle_deleted_dependencies`.
             if let Some(source_capture_model) =
-                dependencies.live.captures.get_by_key(source_capture_name)
+                dependencies.live.captures.get_by_key(&source_capture.capture_name())
             {
                 if self.source_capture.is_none() {
-                    self.source_capture = Some(SourceCaptureStatus::default());
+                    self.source_capture = Some(SourceCaptureStatus {
+                        source_capture: source_capture.def(),
+                        ..Default::default()
+                    });
                 }
                 let source_capture_status = self.source_capture.as_mut().unwrap();
                 // Source capture errors are terminal
@@ -265,10 +268,10 @@ fn handle_deleted_dependencies(
     if drafted
         .source_capture
         .as_ref()
-        .map(|sc| deleted.contains(sc.as_str()))
+        .map(|sc| deleted.contains(sc.capture_name().as_str()))
         .unwrap_or(false)
     {
-        let capture_name = drafted.source_capture.take().unwrap();
+        let capture_name = drafted.source_capture.take().unwrap().capture_name();
         source_capture.take();
         descriptions.push(format!(
             r#"removed sourceCapture: "{capture_name}" because the capture was deleted"#
@@ -290,6 +293,7 @@ fn is_false(b: &bool) -> bool {
 /// Status information about the `sourceCapture`
 #[derive(Debug, Clone, Serialize, Deserialize, Default, PartialEq, JsonSchema)]
 pub struct SourceCaptureStatus {
+    pub source_capture: SourceCaptureDef,
     /// Whether the materialization bindings are up-to-date with respect to
     /// the `sourceCapture` bindings. In normal operation, this should always
     /// be `true`. Otherwise, there will be a controller `error` and the
@@ -371,6 +375,7 @@ impl SourceCaptureStatus {
         draft_row.is_touch = false;
 
         update_linked_materialization(
+            &self.source_capture,
             resource_spec_pointers,
             &self.add_bindings,
             draft_row.model.as_mut().unwrap(),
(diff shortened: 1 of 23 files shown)


psFried (changes requested):
I think this is looking directionally right. I left some inline comments, and just wanted to add a general request to rename any `collection_name` or `task_name` variables that aren't valid catalog names, since I didn't leave separate comments on each one.

crates/flowctl/src/generate/mod.rs
```diff
@@ -393,7 +393,7 @@ fn stub_config(shape: &doc::Shape, collection: Option<&models::Collection>) -> s
         let mut properties = serde_json::Map::new();
 
         for p in &shape.object.properties {
-            if p.is_required {
+            if p.is_required || p.shape.annotations.get("x-schema-name").is_some() {
```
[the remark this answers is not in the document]
mdibaiee: @psFried `x-collection-name` is a requirement for materialization bindings to have, but schema is a new concept being introduced that won't be required for all bindings (in the absence of binding-specific schema we can have a global one). So far we have always had this requirement that the `x-collection-name` property must also be `required` in the binding's JSONSchema config.

crates/agent/src/controllers/materialization.rs
```diff
@@ -283,6 +287,7 @@ fn is_false(b: &bool) -> bool {
 /// Status information about the `sourceCapture`
 #[derive(Debug, Clone, Serialize, Deserialize, Default, PartialEq, JsonSchema)]
 pub struct SourceCaptureStatus {
+    pub source_capture: SourceCaptureDef,
```
psFried, line 290: It doesn't seem necessary or desirable to duplicate the `SourceCaptureDef` on the status. The live spec has it, and should be the source of truth.

crates/agent/src/resource_configs.rs
```diff
@@ -1,34 +1,69 @@
...
+            "cannot create location '{x_collection_name_ptr}' in resource spec '{resource_spec}'"
         );
     };
 
-    Ok(std::mem::replace(prev, resource_name.into()))
+    let _ = std::mem::replace(x_collection_name_prev, x_collection_name.into());
+
+    if source_capture.schema_mode == SourceCaptureSchemaMode::CollectionSchema {
+        if let Some(x_schema_name_ptr) = &resource_spec_pointers.x_schema_name {
+            if let Some(x_schema_name_prev) = x_schema_name_ptr.create_value(resource_spec) {
```
psFried, line 38: It's probably worth returning an error if `create_value` returns None, yeah?

crates/models/src/source_capture.rs
```diff
@@ -0,0 +1,64 @@
+use super::Capture;
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Serialize, Deserialize,  Clone, PartialEq, JsonSchema)]
+#[serde(deny_unknown_fields, rename_all = "camelCase")]
+pub enum SourceCaptureSchemaMode {
```
psFried, line 7: Can you explain your thinking on why this is an enum vs a boolean? Are there other values that you think might make sense in the future?

My knee jerk is just that the name `collectionSchema` is a bit confusing, since flow collections don't technically have a schema. But wondering if maybe bikeshedding the names could be irrelevant, if we're able to change it to something like `set_schema_name: bool`?

crates/models/src/source_capture.rs
```diff
@@ -0,0 +1,64 @@
...
+#[derive(Serialize, Deserialize, Clone, Debug, JsonSchema, PartialEq, Default)]
+#[serde(deny_unknown_fields, rename_all = "camelCase")]
+pub struct SourceCaptureDef {
+    /// Capture name
+    pub capture: Capture,
+
+    /// When adding new bindings from a source capture to a materialization, how should the schema
+    /// of the materialization binding be set
+    #[serde(default)]
+    pub schema_mode: SourceCaptureSchemaMode,
```
psFried, line 32: Should this also have a `skip_serializing_if`? Kinda seems like we should

psFried (changes requested):
Left a few more comments, but I think this is getting pretty close.

crates/models/src/source_capture.rs
mdibaiee: I can get on board with `collectionSchema` being ambiguous / confusing... but I don't think `set_schema_name` is any better, they both do not really convey what is going to happen through their name and they both would need documenting to make it clear.

I went with an enum to allow for future extensions to this system. We already are going to have fixed literals as schema for collections when added through the UI. We may want to have the same thing for the `sourceCapture`.

In either case I think we will need a more meaningful name

crates/agent/src/resource_configs.rs
```diff
@@ -1,34 +1,79 @@
...
         );
     };
 
-    Ok(std::mem::replace(prev, resource_name.into()))
+    let _ = std::mem::replace(x_collection_name_prev, x_collection_name.into());
+
+    let source_capture_def = source_capture.to_normalized_def();
+
+    if source_capture_def.target_schema == SourceCaptureSchemaMode::FromSourceName {
+        if let Some(x_schema_name_ptr) = &resource_spec_pointers.x_schema_name {
```
psFried, line 39: Now that we have `target_schema` as part of the source capture, it seems like maybe we ought to error or at least log something if a user sets `target_schema` but the connector doesn't have an `x-schema-name` annotation. We should perhaps consider checking this as part of publication/validation, to disallow creating materializations with `target_schema` when the target system doesn't have an analogous concept.

crates/agent/src/resource_configs.rs
```diff
@@ -1,34 +1,79 @@
...
+            } else {
+                anyhow::bail!(
+                    "cannot create location '{x_schema_name_ptr}' in resource spec '{resource_spec}'"
+                );
+            }
+        }
+    }
+
+    if source_capture_def.delta_updates {
+        if let Some(x_delta_updates_ptr) = &resource_spec_pointers.x_delta_updates {
```
psFried, line 51: Similar to `target_schema`, it seems like we should validate that the source capture configuration is consistent with the resource spec schema of the connector, or at least log something if it's not.

psFried (commented):
The remaining issue that I'm seeing is just how to handle the possibility of mismatch between the `target_schema` and `delta_updates` fields on the materialization spec, and the resource spec schema of the materialization _connector_. To me, it seems worth validating that the connector actually has the necessary annotations on the resource spec schema. But it might be worth discussing as others may disagree.

crates/agent/src/integration_tests/harness.rs
```diff
@@ -168,7 +168,28 @@ impl TestHarness {
                     'materialization',
                     'http://test.test/',
                     '{"type": "object"}',
-                    '{"type": "object", "properties": {"id": {"type": "string", "x-collection-name": true}}}',
+                    '{"type": "object", "properties": {"id": {"type": "string", "x-collection-name": true}, "schema": {"type": "string", "x-schema-name": true}, "delta": {"type": "boolean", "x-delta-updates": true}}}',
+                    '{/id}',
+                    '{"type": "success"}'
+                ) on conflict do nothing
+            ),
+            materialize_tag_no_annotations as (
```
psFried, line 176: nit: this could be added as just another value in the `materialize_tag` cte

crates/agent/src/publications/specs.rs
```diff
@@ -300,6 +300,41 @@ async fn update_live_specs(
...
+    for materialization in draft.materializations.iter() {
+        let Some(model) = materialization.model() else { return Ok(errors) };
+        let Some(image) = model.connector_image() else { return Ok(errors) };
+        let (image_name, image_tag) = split_image_tag(image);
+        let Some(connector_spec) = agent_sql::connector_tags::fetch_connector_spec(&image_name, &image_tag, pool).await? else { return Ok(errors) };
+        let resource_config_schema = connector_spec.resource_config_schema;
+
+        let resource_spec_pointers = crate::resource_configs::pointer_for_schema(resource_config_schema.0.get())?;
+
+        if let Some(SourceCapture::Configured(source_capture_def)) = &model.source_capture {
```
psFried, line 319: nit: fetching the connector spec could be moved into this block so that we skip fetching it if the materialization doesn't use the full source capture definition.

crates/agent/src/publications/specs.rs
```diff
@@ -300,6 +300,41 @@ async fn update_live_specs(
...
+    pool: &sqlx::PgPool,
+) -> anyhow::Result<tables::Errors> {
+    let mut by_image: BTreeMap<String, bool> = BTreeMap::new();
+    let mut errors = tables::Errors::default();
+
+    for materialization in draft.materializations.iter() {
+        let Some(model) = materialization.model() else { return Ok(errors) };
+        let Some(image) = model.connector_image() else { return Ok(errors) };
+        let (image_name, image_tag) = split_image_tag(image);
+        let Some(connector_spec) = agent_sql::connector_tags::fetch_connector_spec(&image_name, &image_tag, pool).await? else { return Ok(errors) };
```
psFried, line 314: If there's no `connector_tags` row, then we just silently skip this validation? That seems wrong, no? I could see how it would be pretty inconvenient to error in cases where we temporarily switch a task to use a tag from a PR branch or something. But IMO it seems pretty reasonable to temporarily remove the `sourceCapture` in that case. So I'm not really seeing a reason to skip validation. WDYT?

crates/agent/src/publications/specs.rs
```diff
@@ -300,6 +300,41 @@ async fn update_live_specs(
...
+        let resource_config_schema = connector_spec.resource_config_schema;
+
+        let resource_spec_pointers = crate::resource_configs::pointer_for_schema(resource_config_schema.0.get())?;
```
psFried, line 317: Thinking through what would happen if a resource config schema was malformed or missing `x-collection-name`, and I don't see anywhere else where we really validate that. I'm realizing that we're currently missing an opportunity to validate the resource config schema in the `connector_tags` handler, which would allow us to fail fast in that case. Would you mind adding a check to `connector_tags.rs`?

crates/agent/src/publications/specs.rs
```diff
@@ -300,6 +300,41 @@ async fn update_live_specs(
     Ok(lock_failures)
 }
 
+pub async fn check_source_capture_annotations(
+    draft: &tables::DraftCatalog,
+    pool: &sqlx::PgPool,
+) -> anyhow::Result<tables::Errors> {
+    let mut by_image: BTreeMap<String, bool> = BTreeMap::new();
```
psFried, line 307: This looks unused

psFried (commented):
Left a few more comments/questions

crates/agent/src/integration_tests/harness.rs
mdibaiee: the rows that are inserted are accessible by these assigned names in the rest of the script, and I wanted to keep both of these rows accessible with a name

crates/agent/src/publications/specs.rs
mdibaiee: @psFried there is already a check that checks for validity of the connector tag, and these two checks run together, so I've not included any more connector validations here, otherwise we will be validating connector tags twice (i.e. we will essentially run `check_connector_image` in this function as well to achieve validation of connector tags).

crates/agent/src/publications/specs.rs
mdibaiee: Good idea, added it and tested it manually by creating a local image of `materialize-sqlite` without the annotation to see the agent fail the connector_tags job:
image

crates/agent/src/publications/specs.rs
psFried: The issue is that `check_connector_image` only checks that the _image_ is in the `connectors` table. It doesn't check `connector_tags` at all. So it's possible that `fetch_connector_spec` to return `None`, and also for `check_connector_image` to not return any errors.

mdibaiee: @psFried and I think that's by design: we want to allow custom image tags to be specified in publications for testing purposes (I do that quite frequently). I don't think we should fail all publications with tags that are not in the connector_tags table

psFried: I agree that it's really useful to test with different tags. But if we skip the validation here, then it's only going to fail in the controller, which is far less obvious. The goal of this validation is to fail fast with a clear error message, and I'd think we especially want that in testing scenarios where you're already trying to debug some other issue. It also doesn't seem like too great a burden to remove the `sourceCapture` when you switch the connector image tag.

We're also able to add `:testing` tags to `connector_tags` now, since the UI will prefer a `:v1` tag.

mdibaiee: @psFried Resolved this 👍🏽

psFried (approved):
LGTM

PR #1592, opened 2024-08-28
crate(gazette): Include the original error message when we log a `Transport` error warning

jshearer:
**Description:**

Dekaf has been getting a bunch of `gazette::journal::read: broker transport error (will retry) err=transport error` errors recently. It doesn't seem to be causing any issues that I can tell --- all of my test dataflows are still flowing, but it is suspicious. It seems that this message indicates a `gazette::Error::Transport`, which wraps a `tonic::transport::Error` which gets mapped into by a whole bunch of things inside tonic.

Screen Shot 2024-08-28 at 12 34 51

I made this quick change to get some more details about the cause of the transport errors (of course they stopped as soon as I deployed the change, which is itself a hint) and wanted to get opinions before I do more. Should all of these cases include the source error message? Should it instead be an additional `tracing::debug!` message? I imagine the reason for not including it originally is to simplify the error messaging as it could be user facing. Thoughts?

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1592)

### DIFF
--- a/crates/gazette/src/journal/read.rs
+++ b/crates/gazette/src/journal/read.rs
@@ -29,7 +29,12 @@ impl Client {
                 match result {
                     Ok(()) => attempt = 0,
                     Err(Error::Transport(err)) => {
-                        tracing::warn!(%err, "broker transport error (will retry)");
+                        let source = std::error::Error::source(&err);
+                        if let Some(source) = source {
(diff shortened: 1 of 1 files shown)


jgraettinger:
These log statements are going away in the federated data-planes branch, and will be surfaced as part of polling the stream.

When logging probably easiest to wrap with anyhow::anyhow! which has full handling for the error chain.

PR #1600, opened 2024-08-30
crates/sources: extend_from_catalog fixes

### LINKED ISSUE #1558: flowctl catalog pull-specs sometimes removes specs
User reported that running `flowctl catalog pull-specs --overwrite` would sometimes remove a capture from their local `flow.yaml` file.

- Start with a directory containing a top-level `flow.yaml` file that imports from a subdirectory.
- The top-level `flow.yaml` defines a capture
- Run `flowctl catalog pull-specs --overwrite`
- See that the capture was removed
- Run `flowctl catalog pull-specs --overwrite` again and see that the capture was added again

### LINKED ISSUE #1590: Flowctl --overwrite does not overwrite changed specs
I found a bug where when I:

1. pull specs (using `flowctl catalog pull-specs --prefix "$source" --overwrite`)
2. change them through UI
3. pull again (using `flowctl catalog pull-specs --prefix "$source" --overwrite`)

They don't change locally, if I `rm` them locally and then pull again, they are correct.

jgraettinger:
Use the _local_ scope, NOT the remote one, to determine the existing local scope of a resource which is being kept or over-written.

We must also clear out imports of a local spec which is being replaced, because the replacement is a fully inline model, and we're typically going to indirect it before writing out resources, which would otherwise result in duplicated imports.

Fixes #1590
Fixes #1558

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1600)

### DIFF
--- a/crates/sources/src/merge.rs
+++ b/crates/sources/src/merge.rs
@@ -217,10 +217,16 @@ where
                         entity,
                         catalog_name.as_ref(),
                         &root,
-                        Some(rhs.scope()),
+                        Some(lhs.scope()),
                     );
 
                     if let Some(last) = chain.last() {
+                        // We're fully replacing the lhs resource with the already-inlined
+                        // `model`, which means that all of its imports are no longer used.
+                        draft
+                            .imports
+                            .retain(|r| !r.scope.as_str().starts_with(lhs.scope().as_str()));
+
                         add_imports(draft, &chain);
                         *count += 1;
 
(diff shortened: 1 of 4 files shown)


psFried (approved):
It took me a while to figure out how this fixed the problem, and I'm still not really sure I understand how it ever appeared to have worked. Because I'm pretty sure it's at least sort of worked for me in the past. :shrug: LGTM

PR #1610, opened 2024-09-05
runtime: hard code `materialize-bigquery` in `ser_policy`

williamhbaker:
**Description:**

BigQuery actually has a limit of 20MiB on an individual row that can be returned via its Query API. Although we can store a row with a `flow_document` that large, we then can't read it back out using the conventional query mechanism the materialization connector currently uses.

This adds `materialize-bigquery` to the list of connectors that use `ser_policy`, which should generally reduce how often we store rows that are excessively large.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1610)


jgraettinger (approved):
LGTM

PR #1618, opened 2024-09-09
parser: parquet format support

mdibaiee:
**Description:**

- Simple parquet parser
- Limit file size to 1GB, tested this manually with a large file
- Tested against two files from https://www.nyc.gov/site/tlc/about/tlc-trip-record-data.page and a file from a materialize-parquet connector

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1618)


crates/parser/src/format/parquet.rs
```diff
@@ -0,0 +1,103 @@
...
+use std::convert::TryFrom;
+use serde_json::Value;
+
+struct ParquetParser;
+
+pub fn new_parser() -> Box<dyn Parser> {
+    Box::new(ParquetParser)
+}
+
+const MAX_FILE_SIZE: i64 = 1024 * 1024 * 1024;
```
williamhbaker, line 15: I'd expect us to be able to read a file row group by row group, and if there was going to a be limit it would be based on the row group size and not the total file size.

I'm also wondering how this works with compressed files - is the metadata byte size for compressed data or uncompressed data? If `rg.total_byte_size()` is telling us about compressed size it may be dramatically different when uncompressed and loaded into memory. But I am not sure about this.

mdibaiee: @williamhbaker this is the uncompressed size. Good point about row groups, I'm going apply the limit on a row group level rather than the whole file size... that is after verifying that's how the data is read 🤔

williamhbaker (approved):
LGTM

PR #1626, opened 2024-09-13
fix: `flowctl catalog` does need pagination after all

jshearer:
**Description:**
I was having a weird issue where `flowctl catalog pull-specs --captures=true --collections=true` wasn't returning any captures, but `flowctl catalog pull-specs --captures=true --collections=false` was, and realized that the first command was returning 1001 specs, which is suspiciously close to PostgREST's max row limit. Turns out that even though we're paginating the inputs, we still need to paginate the queries themselves.

### Existing `flowctl`
```
$ flowctl catalog pull-specs --captures=true --collections=true --prefix <...>
Wrote 1001 specifications under file:///Users/js/Documents/estuary/automatic-backfills/test/flow.yaml.
```

### Updated `flowctl`
```
$ /usr/local/bin/flowctl catalog pull-specs --captures=true --collections=true --prefix <...>
Wrote 1732 specifications under file:///Users/js/Documents/estuary/automatic-backfills/test/flow.yaml.
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1626)

### DIFF
--- a/crates/flowctl/src/catalog/mod.rs
+++ b/crates/flowctl/src/catalog/mod.rs
@@ -6,9 +6,13 @@ mod test;
 use crate::{
     api_exec, api_exec_paginated, controlplane,
     output::{to_table_row, CliOutput, JsonCell},
+    pagination::into_items,
 };
 use anyhow::Context;
-use futures::stream::{FuturesUnordered, StreamExt};
+use futures::{
+    stream::{FuturesUnordered, StreamExt},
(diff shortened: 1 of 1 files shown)


jgraettinger (approved):
LGTM

jshearer:
Updated to remove incorrect comment:
> // No need for pagination because we're paginating the inputs.

PR #1636, opened 2024-09-17
flowctl: use max build_id for validating local specs

psFried:
The validation logic was changed in a prior commit to error if the `last_build_id` of a live spec is greater than the current build id. This change was supposed to go along with that, but got missed. Using the maximum id value ensures that we never get errors due to the build id being superseded.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1636)


jshearer (approved):
Confirmed that this change fixes the unexpected 
```
current build ID 0101010101010101 has been superseded by a larger build ID ...; please retry the operation
```
errors

PR #1637, opened 2024-09-18
flowctl: Update pagination to use 0-indexed inclusive ranges, and page by response length rather than fixed page size.

jshearer:
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1637)

### DIFF
--- a/crates/flowctl/src/pagination.rs
+++ b/crates/flowctl/src/pagination.rs
@@ -12,7 +12,7 @@ use crate::api_exec;
 /// it's currently on. Used in [`page_turner::PageTurner`] to implement pagination.
 pub struct PaginationRequest {
     builder: postgrest::Builder,
-    page: usize,
+    next_offset: usize,
     page_size: usize,
     /// The original Builder's range value, if set. This is used to make sure we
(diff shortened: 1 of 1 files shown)


crates/flowctl/src/pagination.rs
```diff
@@ -120,27 +122,31 @@ where
         request: PaginationRequest,
     ) -> PageTurnerOutput<Self, PaginationRequest> {
         let resp: Vec<Item> = api_exec::<Vec<Item>>(request.builder.clone()).await?;
+        let response_length = resp.len();
 
         // Sometimes, it seems, we can get back more than the requested page size.
         // So far I've only seen this on a request of 1,000 and a response of 1,001.
-        if resp.len() >= request.page_size
+        if response_length >= request.page_size
```
psFried, line 129: The comment above is explained by the range being inclusive, yeah? Might be good to remove or update it.

psFried (approved):
LGTM

crates/flowctl/src/pagination.rs
jshearer: Good call, removed!

PR #1638, opened 2024-09-18
dekaf: Gracefully handle invalid connection attempts without shutting down the whole server

jshearer:
**Description:**

Since Dekaf had to learn how to terminate its own TLS connections, we've been seeing some crashes when a non-TLS client tries to open a connection.

**Before:** (crashes)
```
Error: received corrupt message of type InvalidContentType |  

Stack backtrace: 
   0: anyhow::error::<impl core::convert::From<E> for anyhow::Error>::from |  
   1: dekaf::main::{{closure}} |  
   2: tokio::runtime::park::CachedParkThread::block_on |  
   3: tokio::runtime::context::runtime::enter_runtime |  
   4: tokio::runtime::runtime::Runtime::block_on |  
   5: dekaf::main |  
   6: std::sys::backtrace::__rust_begin_short_backtrace |  
   7: std::rt::lang_start::{{closure}} |  
   8: std::rt::lang_start_internal |  
   9: main |  
  10: <unknown> |  
  11: __libc_start_main |  
  12: _start
```

**After:**
```
ERROR dekaf: Failed to accept TLS connection e="received fatal alert: UnknownCA"
ERROR dekaf: Failed to accept TLS connection e="received corrupt message of type InvalidContentType"
ERROR dekaf: Failed to accept TLS connection e="received corrupt message of type InvalidContentType"
ERROR dekaf: Failed to accept TLS connection e="received corrupt message of type InvalidContentType"
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1638)

### DIFF
--- a/crates/dekaf/src/main.rs
+++ b/crates/dekaf/src/main.rs
@@ -161,7 +161,8 @@ async fn main() -> anyhow::Result<()> {
             tls_cfg.certificate_file.clone().unwrap(),
             tls_cfg.certificate_key_file.clone().unwrap(),
         )
-        .await?;
+        .await
+        .context("failed to open or read certificate or certificate key file")?;
 
         let schema_server_task = axum_server::bind_rustls(schema_addr, axum_rustls_config.clone())
             .serve(schema_router.into_make_service());
@@ -195,8 +196,12 @@ async fn main() -> anyhow::Result<()> {
             let acceptor = acceptor.clone();
             tokio::select! {
                 accept = kafka_listener.accept() => {
-                    let (socket, addr) = accept?;
-                    let socket = acceptor.accept(socket).await?;
+                    let Ok((socket, addr)) = accept else {
+                        continue
+                    };
+                    let Ok(socket) = acceptor.accept(socket).await else {
+                        continue
+                    };
 
                     tokio::spawn(serve(Session::new(app.clone(), cli.encryption_secret.to_owned()), socket, addr, stop.clone()));
                 }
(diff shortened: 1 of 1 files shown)


crates/dekaf/src/main.rs
```diff
@@ -195,8 +196,20 @@ async fn main() -> anyhow::Result<()> {
             let acceptor = acceptor.clone();
             tokio::select! {
                 accept = kafka_listener.accept() => {
-                    let (socket, addr) = accept?;
-                    let socket = acceptor.accept(socket).await?;
+                    let (socket, addr) = match accept {
+                        Ok(res) => res,
+                        Err(e) => {
+                            tracing::error!(e=e.to_string(), "Failed to accept TCP connection");
```
psFried, line 202: In general, my understanding is that it's best to avoid things like `e.to_string()` because it's eagerly evaluated. Tracing also has an easier shorthand for lazily evaluated `Display` rendering using for example `e=%e` or just `%e`.

Elsewhere, we've been using `error` in structured logs. IDK if there's already a standard in dekaf for `e`, but if not then I'd suggest switching to `error`.

crates/dekaf/src/main.rs
```diff
@@ -195,8 +196,20 @@ async fn main() -> anyhow::Result<()> {
...
+                            continue;
+                        }
+                    };
+                    let socket = match acceptor.accept(socket).await {
+                        Ok(res) => res,
+                        Err(e) => {
+                            tracing::error!(e=e.to_string(), "Failed to accept TLS connection");
```
psFried, line 209: IDK how common this has been, but I'll just put it out there that it could potentially result in a lot of log spam.

psFried (approved):
LGTM % some optional nit picks

crates/dekaf/src/main.rs
jshearer: Yeah, that's true... could conceivably be quite frequent if someone configures a connection using `SASL_PLAINTEXT` and forgets about it for a while. I think I'll just get rid of the logs

crates/dekaf/src/main.rs
jshearer: Ah, `%` is helpful thx. I went through a couple different iterations of trying to represent this without getting junk like `Some(Error{error: FatalAlert(UnknownCA)})`. Will keep in mind for the future

PR #1641, opened 2024-09-19
agent: use new pub_id when evolving materializations

psFried:
Ensure that the materialization controller will always use a new publication id when applying onIncompatibleSchemaChange actions. Previously, it would re-use the publication id of the one that had failed. That doesn't work if the failed publication was a "touch" publication. Applying `onIncompatibleSchemaChange` actions modifies the model, and so the publication id must be greater than `last_pub_id`. Generating a new id ensures that is always the case.

Also includes some minor cleanups that didn't make it into the previous pr (#1639) to eliminate the duplicated `is_touch` property in favor of just checking the draft rows.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1641)

### DIFF
--- a/crates/agent/src/controllers/publication_status.rs
+++ b/crates/agent/src/controllers/publication_status.rs
@@ -114,6 +114,13 @@ fn count_schema(_: &mut schemars::gen::SchemaGenerator) -> schemars::schema::Sch
     .unwrap()
 }
 
+fn is_touch_pub(draft: &tables::DraftCatalog) -> bool {
+    draft.tests.iter().all(|r| r.is_touch)
+        && draft.collections.iter().all(|r| r.is_touch)
+        && draft.captures.iter().all(|r| r.is_touch)
+        && draft.materializations.iter().all(|r| r.is_touch)
+}
+
 impl PublicationInfo {
     pub fn is_success(&self) -> bool {
         // TODO: should EmptyDraft be considered successful?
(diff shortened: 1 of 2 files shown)


crates/agent/src/controllers/publication_status.rs
```diff
@@ -189,8 +187,9 @@ impl PendingPublication {
         tracing::info!("starting touch");
         let new_hash = new_dependency_hash.unwrap_or("None");
         let old_hash = state.live_dependency_hash.as_deref().unwrap_or("None");
-        self.details.push(format!("in response to change in dependencies, prev hash: {old_hash}, new hash: {new_hash}"));
-
+        self.details.push(format!(
+            "in response to change in dependencies, prev hash: {old_hash}, new hash: {new_hash}"
```
jshearer, line 191: I don't suppose we could include the names of the dependencies that changed easily, could we? I'm just envisioning reading a publication that says this in the future and wishing it told me _which_ dependency change was causal. Out of scope for this PR for sure, just thinking.

jshearer (approved):
LGTM

crates/agent/src/controllers/publication_status.rs
psFried: To do so would require a data structure to store the `last_pub_id` of every dependency individually. It wouldn't be crazy to stuff them into the `built_spec`, and come to think of it that might have been a better approach than the one I took. But yeah, doesn't seem like the highest priority.

Issue #1622, opened 2024-09-10
dekaf: introduce new materialization endpoint type


jshearer:
The intent of this task is to flesh out the framing for how Dekaf is used. Currently, it works by exposing all of the collections you have read access to as Kafka topics. This isn't ideal for a few reasons:
* Stats/logs/billing
* Bindings/Projections
* Config
* UX

We already have a bunch of design and infrastructure built to support creating and configuring tasks that pull data out of Flow collections: materializations. So, the idea is to frame Dekaf usage as another kind of materialization. Currently, we support [two types of materializations](https://github.com/estuary/flow/blob/master/crates/models/src/materializations.rs#L47-L53): `Local` and `Connector`. The idea is to add a third type `Dekaf`. 

This stretches the abstraction a bit, as at present the intent of a materialization is to (describe how to) actively run a task that performs the work of materialization, whereas a `Dekaf`-type materialization won't directly start any work and instead just act as a vehicle for configuring actual Dekaf, but AFAICT it doesn't introduce any direct contradictions and does feel like it elegantly gets us all of the presently missing features we want.

In practice, this will look something like this:
* Add a new `MaterializationEndpoint::Dekaf` variant. This will contain whatever configuration is specific to this dekaf task such as the `strict_topic_names` flag. 
* When a materialization using this endpoint type gets activated, it will still create ops collections for logs and stats, but it won't start up any containers.
* Dekaf will be updated to take the name of a materialization as part of its config input provided in the username field. Then, instead of advertising all available collections, it will advertise all of the enabled bindings for that materialization instead.
* Dekaf will directly write logs and stats to the ops collections of its configured materialization task.

PR #1642, opened 2024-09-19
Introduce `MaterializationEndpoint::Dekaf`

jshearer:
**Description:**

This is the first part of #1622. It introduces a new materialization endpoint type called `dekaf`, as well as the necessary control-plane plumbing to give materializations of this type ops logs and stats journals when activated. As materializations  of this type are purely descriptive and are actually served by the stand-alone Dekaf service, no shards or recovery log journals should be created.

I added a new happy-path test to snapshot a new dekaf-type materialization, and am able to publish the following spec to my local stack, see it show up in the UI, and delete it successfully.

```yaml
materializations: 
  aliceCo/dekaf-test:
    bindings:
      - source: aliceCo/test-hello/events
        resource: {}
        fields:
          recommended: true
    endpoint:
      dekaf: {}
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1642)


crates/activate/src/lib.rs
```diff
@@ -253,15 +268,30 @@ async fn apply_changes(
     Ok(())
 }
 
+/// Describes the desired future state of a task.
+/// Virtual tasks get logs and stats journals,
+/// but are otherwise purely descriptive and
+/// do not get shards and recovery log journals
+/// created for them like real tasks do.
+#[derive(Clone, Copy, Debug)]
+enum TaskTemplate<'a> {
```
psFried, line 277: I don't _love_ the naming of real vs virtual tasks here, though I could live with it. I'll toss out a suggestion of `UpsertShard` and `UpsertOpsJournals` if that sounds any better to you.

crates/models/src/materializations.rs
```diff
@@ -50,6 +50,8 @@ pub enum MaterializationEndpoint {
     Connector(ConnectorConfig),
     /// # A local command (development only).
     Local(LocalConfig),
+    /// # A Dekaf connection
+    Dekaf(DekafConfig),
```
psFried, line 54: Just my opinion here, but "dekaf" doesn't seem like a term that we should make users learn. `kafka` or `kafkaRead` might be better, since "kafka" is the term that people are already familiar with. IDK if that's open for debate, but just my 2 cents.

psFried (approved):
Looks good!

crates/models/src/materializations.rs
jshearer: Yeah, I struggled with this a bit, but now that you bring it up I think you're right. I went with Dekaf originally because it made it clear that this was a hard-coded config specifically for our special snow-flake endpoint type. But `kafka` or `kafkaRead` both seem equally descriptive and less confusing 👍 I'll switch the wording around

crates/activate/src/lib.rs
jgraettinger: I think I actually prefer the Real / Virtual distinction, because UpsertShard also implies creation of ops journals, which isn't clear with UpsertopsJournals.

`UpsertShard` and `UpsertPlaceholder` is the other place my mind goes but I don't think that's better.

crates/flowctl/src/generate/mod.rs
```diff
@@ -288,6 +288,14 @@ async fn generate_missing_materialization_configs(
             },
             serde_json::from_str::<url::Url>(config.config.get()).ok(),
         ),
+        models::MaterializationEndpoint::Dekaf(config) => (
+            materialize::request::Spec {
+                connector_type: flow::materialization_spec::ConnectorType::Dekaf as i32,
+                config_json: serde_json::to_string(config).unwrap(),
+            },
+            // Dekaf isn't a pluggable connector, and so does not have dynamic config.
```
jgraettinger, line 296: nit: It does have a config, though -- it's just empty today (and probably we'll add to it in the future as we allow more customization of behavior). Ideally we'd be able to generate stubs here for it based on the JSON schema of `DekafConfig`, just as we generate stubs for the dynamic endpoint config schemas of connectors. I don't recall this code well enough to know how much of a lift that is.

crates/models/src/materializations.rs
jgraettinger: I'm on the other side of this. I think `dekaf` is better because there's far less potential for confusion: if a newcomer was looking at a flow.yaml spec and saw a `kafka` stanza they could easily have a **very** different assumption about what's happening.

"Dekaf" is also something we're explicitly building brand around, and these will be set up in the UI mostly so discover-ability isn't a real concern.

crates/sources/src/indirect.rs
```diff
@@ -498,6 +498,10 @@ fn indirect_materialization(
             resources,
             threshold,
         ),
+        // Dekaf isn't a pluggable connector, and so does not have dynamic config.
```
jgraettinger, line 501: It could be, though, especially if config grows to be large. There's no reason not to treat it like any other config that I'm aware of. In practice, it will remain inline unless it gets large.

jgraettinger (approved):
LGTM % nits around config handling

crates/models/src/materializations.rs
jshearer: ☝️ this is the thought process I originally had, will keep it as is then

crates/flowctl/src/generate/mod.rs
jshearer: Updated to support either a string (in which case it's assumed to be a file or url and gets loaded if valid), or an object in which case it's an inlined `DekafConfig`

crates/sources/src/indirect.rs
jshearer: Updated

crates/models/src/materializations.rs
```diff
@@ -41,6 +41,13 @@ pub struct MaterializationDef {
     pub delete: bool,
 }
 
+#[derive(Serialize, Deserialize, Clone, Debug, JsonSchema, PartialEq)]
+#[serde(untagged)]
+pub enum DekafConfigContainer {
```
jshearer, line 46: @jgraettinger This seem right to you? The rest of the config objects just use a `RawValue` for their config because it's dynamic, and that lets them inherently accept either a string (which is treated as an indirect reference), or an object (which is treated as an inline config).

jshearer:
Closing in favor of https://github.com/estuary/flow/pull/1665 which contain some substantial changes from this PR which I didn't want to push here and confuse things

PR #1645, opened 2024-09-20
agent: use touch publications for expanded draft specs

### LINKED ISSUE #1643: agent: touch expanded specs for user-initiated publications
For user-initiated publications, we expand the draft in order to re-validated connected specs along with those that are being modified. With the introduction of touch publications, we now have the option to only touch expanded specs instead of incrementing their `last_pub_id`. This would help avoid unnecessary entries in `publication_specs`, and also prevents unnecessary `ExpectPubIdNotMatched` errors in scenarios where multiple publications concurrently affect connected specs (e.g. publication A of a collection and publication B of a connected materialization).

psFried:
**Description:**

Touch publications are kind of a perfect fit for expanded specs in user-initiated publications. Expanded specs are never modified by the publication, and it's annoying to add publication specs for things that users haven't intended to publish. Using `is_touch: true` on expanded specs prevents the creation of unnecessary publication specs. It also avoids unnecessary modification of `last_pub_id`, so there will be less chance of `ExpectPubIdMismatch` errors. The touched specs are still validated, which serves the original purpose of spec expansion.

Resolves #1643

**Workflow steps:**

- Create a capture into one or more collections, and a materialization.
- Publish a change to only the collection using flowctl
- Observe that the `last_pub_id` of the capture and materialization have not changed, but the `last_build_id` has
- No `publication_specs` are added for the capture and materialization, only the collection

**Notes for reviewers:**

This is a follow up change from #1629 , which added the "touch" capability. I wanted to break this out as a separate PR just to limit the scope of changes in any one deployment.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1645)

### DIFF
--- a/crates/agent/src/integration_tests/harness.rs
+++ b/crates/agent/src/integration_tests/harness.rs
@@ -18,6 +18,10 @@ use tempfile::tempdir;
 
 const FIXED_DATABASE_URL: &str = "postgresql://postgres:postgres@localhost:5432/postgres";
 
+pub fn set_of(names: &[&str]) -> BTreeSet<String> {
+    names.into_iter().map(|n| n.to_string()).collect()
+}
+
 #[derive(Debug, Deserialize)]
 #[allow(dead_code)] // Suppress lints for fields used only in test snapshots
 pub struct LiveSpec {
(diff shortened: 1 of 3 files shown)


jgraettinger (approved):
LGTM

PR #1647, opened 2024-09-20
docs: update source-mongodb docs for batch mode

williamhbaker:
**Description:**

Doc updates for https://github.com/estuary/connectors/pull/1956

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1647)

### DIFF
--- a/site/docs/reference/Connectors/capture-connectors/mongodb.md
+++ b/site/docs/reference/Connectors/capture-connectors/mongodb.md
@@ -27,11 +27,57 @@ If you are using a user with access to all databases, then in your mongodb addre
 `?authSource=admin` parameter so that authentication is done through your admin database.
 :::
 
-- ReplicaSet enabled on your database, see [Deploy a Replica
-  Set](https://www.mongodb.com/docs/manual/tutorial/deploy-replica-set/).
-
 - If you are using MongoDB Atlas, or your MongoDB provider requires allowlisting of IPs, you need to
   [allowlist the Estuary IP addresses](/reference/allow-ip-addresses).
+
+## Capture Modes
+
+MongoDB [change streams](https://www.mongodb.com/docs/manual/changeStreams/) are
+the preferred way to capture on-going changes to collections. Change streams
+allow capturing real-time events representing new documents in your collections,
+updates to existing documents, and deletions of documents. If change streams are
+enabled on the MongoDB instance/deployment you are connecting to, they will be
+used preferentially for capturing changes.
+
+An alternate "batch" mode of capturing documents can be used for deployments
+that do not support change streams, and for MongoDB collection types that do not
+support change streams ([views](https://www.mongodb.com/docs/manual/core/views/)
+and [time
+series](https://www.mongodb.com/docs/manual/core/timeseries-collections/)
+collections). The capture mode is configured on a per-collection level in the
+**Bindings** configuration and can be one of the following:
+- **Change Stream Incremental**: This is the preferred mode and uses change streams to capture change events.
+- **Batch Snapshot**: Performs a "full refresh" by scanning the entire MongoDB
+  collection on a set schedule. A cursor field must be configured, which should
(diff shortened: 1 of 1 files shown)


site/docs/reference/Connectors/capture-connectors/mongodb.md
```diff
@@ -44,19 +90,25 @@ MongoDB source connector.
...
+| `/batchAndChangeStream` | Capture Batch Collections in Addition to Change Stream Collections | Discover collections that can only be batch captured if the deployment supports change streams. Check this box to capture views and time series collections as well as change streams. All collections will be captured in batch mode if the server does not support change streams regardless of this setting. | boolean |                  |
+| `/pollSchedule`         | Default Batch Collection Polling Schedule                          | When and how often to poll batch collections. Accepts a Go duration string like '5m' or '6h' for frequency-based polling or a string like 'daily at 12:34Z' to poll at a specific time (specified in UTC) every day. Defaults to '24h' if unset                                                                 | string  |                  |
```
Alex-Bair, line 100: I think the default `'24h'` should be placed in the "Required/Default" column instead of tucked inside "Description".

site/docs/reference/Connectors/capture-connectors/mongodb.md
```diff
@@ -44,19 +90,25 @@ MongoDB source connector.
...
+| **`/database`**   | Database         | Database name                                                                                                                                                                                                                                                                              | string | Required         |
+| **`/collection`** | Stream           | Collection name                                                                                                                                                                                                                                                                            | string | Required         |
+| `/captureMode`    | Capture Mode     | Either **Change Stream Incremental**, **Batch Snapshot**, or **Batch Incremental**                                                                                                                                                                                                         | string |                  |
```
Alex-Bair, line 108: I think **Change Stream Incremental** is the default if not set, right? Could that be represented in the "Required/Default" column?

Alex-Bair (approved):
LGTM % minor comments around clarifying default values.

site/docs/reference/Connectors/capture-connectors/mongodb.md
williamhbaker: It's not really a "schema default" in the sense that it would get auto-populated in the UI from a `default` annotation.  Probably a distinction without a difference thing here, but the default is actually an empty input which the connector will interpret as `24h`.

site/docs/reference/Connectors/capture-connectors/mongodb.md
williamhbaker: Like above, there is no default, although an empty input is handled as "Change Stream Incremental" by the connector.

site/docs/reference/Connectors/capture-connectors/mongodb.md
Alex-Bair: I see, I've been interpreting any values in the "Required/Default" column not as a "schema default" but rather as a more general "if you don't provide a value, the connector will use X". As long as the value that's used by the connector if the user doesn't provide one is mentioned somewhere, that sounds good.

site/docs/reference/Connectors/capture-connectors/mongodb.md
Alex-Bair: Would it be too verbose to add "Defaults to Change Stream Incremental if unset"?

williamhbaker: Practically speaking it is impossible to have it be unset going forward, since discovery will always populate it with something, and you won't be able to select an empty input in the UI. The change-stream-if-empty behavior is only there for legacy tasks which do not have anything set.

Alex-Bair: Gotcha, that makes sense! Thanks for explaining that for me!

PR #1648, opened 2024-09-20
dekaf: Introduce transient error retry limit

jshearer:
**Description:**

I found Dekaf getting stuck in a never-ending loop of `Retrying transient read error err=transport error`. Now we bail when that happens, but also log some slightly more useful data: 
```
Retrying transient read error error=Transport(tonic::transport::Error(Transport, ConnectError(ConnectError("tcp connect error", Custom { kind: TimedOut, error: Elapsed(()) }))))
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1648)


jgraettinger (approved):
LGTM

PR #1651, opened 2024-09-20
dekaf: Temporarily disable record LZ4 compression

jshearer:
**Description:**

So, this was crazy. @danthelion reported to me this morning that ClickHouse wasn't able to read from Dekaf. I spent a bunch of time digging into it, and found that long story short, Clickhouse's Kafka consumer was acting as if we were reporting bad data -- issuing a Fetch request, and then throwing the response away and re-issuing a Fetch request for the same offset, over and over.

I initially thought it was related to the change I had made to [treat Flow acks as Kafka control messages](https://github.com/estuary/flow/pull/1516/commits/e127a5c1794ce4935565f9004ea72d7adc19092e), but backing that out had no effect. I tried a bunch of other things unsuccessfully until I finally took a wild guess and disabled LZ4 compression... which immediately completely fixed the problem.

This _isn't_ the first time that client/server compression compatibility has broken Dekaf. I ran into a similar issue before which was fixed [here](https://github.com/tychedelia/kafka-protocol-rs/pull/59), but this time I'm not able to reproduce this issue in any local testing. 

I'd like to get this merged ASAP to unblock the Dekaf release (at least for Clickhouse) happening early next week, but ultimately I intend to make compression configurable as part of a Dekaf-type materialization's endpoint config.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1651)

### DIFF
--- a/crates/dekaf/src/read.rs
+++ b/crates/dekaf/src/read.rs
@@ -252,7 +252,7 @@ impl Read {
         }
 
         let opts = RecordEncodeOptions {
-            compression: Compression::Lz4,
+            compression: Compression::None,
             version: 2,
         };
         RecordBatchEncoder::encode(&mut buf, records.iter(), &opts)


jgraettinger:
This will have a marked impact on network egress fees and make Dekaf more expensive to operate.

What's they response about supporting LZ4 compression?

jshearer:
I sent an email earlier today about this, haven't heard back. I will follow up with more details, but it's Friday afternoon, I doubt we'll hear back before next week.

jshearer:
They got back to me and said they use https://github.com/twmb/franz-go instead of librdkafka, which is great -- if we don't figure it out from logs, I can come up with a quick consumer using that library to hopefully reproduce the issue locally and fix it.

jgraettinger (approved):
LGTM

jshearer:
Just for a paper trail, I spent a bunch of time today digging into this and ended up finding a related issue that fixes the problem, but unfortunately no smoking gun. 

The actual error (which is [unhelpfully swallowed](https://github.com/twmb/franz-go/blob/b77dd13e2bfaee7f5181df27b40ee4a4f6a73b09/pkg/kgo/source.go#L1478) by `franz-go`) is this:
```
lz4: invalid block checksum: got a1d59126; expected e48f521c
```

It _appears_ that somewhere between `kafka-protocol` which uses [`lz4-rs`](https://crates.io/crates/lz4) and `franz-go` which uses [`lz4`](https://github.com/pierrec/lz4), there's some disagreement about what a correct lz4 chunk looks like. I tried a whole bunch of different things:

* Updated `kafka-protocol` to [bleeding edge](https://github.com/tychedelia/kafka-protocol-rs/pull/81) in order to use a custom compressor and directly invoke lz4-rs
* Manually lz4'd a bunch of different test inputs and provided them directly to go's LZ4 library
* Extracted the compressed records from [`RecordBatchEncoder`](https://docs.rs/kafka-protocol/latest/kafka_protocol/records/struct.RecordBatchEncoder.html) and directly decompressed with go
* Turned on and off all of the LZ4 flags I could: content checksums, block checksums, tried various compression levels. This is where I found the related issue.

Turns out, `lz4-rs` forgot to actually propagate its encoder setting `block_checksum_flag`, leaving it to be always enabled no matter what you pass to `.block_checksum()`. Since the default is `BlockChecksum::NoBlockChecksum` (disabled), and the error is talking specifically about block checksums, I went and made the change to propagate this flag down to the underlying lz4 library. This fixed the problem by treating a symptom (disabling block checksums), rather than the core issue of why the block checksums were getting either incorrectly calculated or incorrectly rejected.

Either way, once https://github.com/10XGenomics/lz4-rs/pull/52 works its way through we should be able to turn lz4 back on with no more problems (though we should test it... 😬).

twmb:
> which is [unhelpfully swallowed](https://github.com/twmb/franz-go/blob/b77dd13e2bfaee7f5181df27b40ee4a4f6a73b09/pkg/kgo/source.go#L1478) by franz-go

My bad, but this is no longer entirely the case as of this commit!
https://github.com/twmb/franz-go/commit/1473778f73e803eb1d311c0bd972e9c938400379

PR #1653, opened 2024-09-23
dekaf: Swap to `lz4_flex` from `lz4`

jshearer:
While investigating the cause of LZ4 compression issues related to franz-go (https://github.com/estuary/flow/pull/1651#issuecomment-2369322939), I found [`lz4_flex`](https://github.com/PSeitz/lz4_flex) which is a pure-Rust lz4 implementation which appears to be safer and faster than `lz4`/`lz4-sys` that `kafka-protocol` is using. 

Now that https://github.com/tychedelia/kafka-protocol-rs/pull/81 allows us to use our own compression, and `lz4`'s configuration of block checksums is broken (fix here https://github.com/10XGenomics/lz4-rs/pull/52), I thought it would be a good time to swap to `lz4_flex`.

I've confirmed that this change _also_ fixes the issue with block checksums, mainly because `lz4_flex`'s block checksums are actually off by default (unfortunately, turning them on reproduces the issue).

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1653)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1831,6 +1831,7 @@ dependencies = [
  "itertools 0.10.5",
  "kafka-protocol",
  "labels",
+ "lz4_flex",
  "md5",
  "metrics",
  "metrics-prometheus",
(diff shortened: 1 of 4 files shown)


jgraettinger (approved):
LGTM

PR #1655, opened 2024-09-25
dekaf: Track and report additional metrics, and improve connection stability/freshness

jshearer:
I observed some concerning behavior when switching `dekaf.estuary.dev` to advertise `dekaf.estuary-data.com`, namely all API requests to the upstream MSK broker were returning with error codes. A restart of Dekaf "fixed" the problem, and I ended up realizing that those connections were probably sitting open and idle for days.

The theory is that even though we set TCP keepalive, the connections had gotten into a bad state. So I introduced a couple different mitigations:

* All idle connections (that is, connections that were opened but have since been returned to the pool) get closed after a short period of time (60s in this case)
* All idle connections that were opened over 30m ago get closed. The stateless usage pattern of connections should mean that this results in no connection lasting longer than roughly 30m.
* A connection is exercised every time it's returned to the pool to check whether it's still open and able to communicate with the broker.

New metrics:
* `pool_size` (by broker): The number of connections open to this broker
* `pool_available` (by broker): The number of open, free connections available for use
* `pool_waiting` (by broker): The number of requests blocked waiting for a connection to become available
* `pool_connection_avg_age` (by broker): the average age of a pool connection to this broker
* `pool_connection_avg_idle` (by broker): the average time a connection to this broker has been idle
* `api_call_time` (by api method): A histogram of the time it takes to serve a kafka protocol request

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1655)


jgraettinger (approved):
LGTM

crates/dekaf/src/api_client.rs
```diff
@@ -323,6 +349,44 @@ impl KafkaApiClient {
...
         .build()?;
 
+        // Close idle connections, and any free connection older than 30m.
+        // It seems that after running for a while, connections can get into
+        // a broken state where every response returns an error. This, plus
+        // the healthcheck when recycling a connection solves that problem.
+        let reap_interval = Duration::from_secs(30);
+        let max_age = Duration::from_secs(60 * 30);
+        let max_idle = Duration::from_secs(60);
+        let reaper = tokio_util::task::AbortOnDropHandle::new(tokio::spawn({
```
jgraettinger, line 359: :heart: thanks for the explicit syntax

jgraettinger (approved):

PR #1662, opened 2024-09-26
Phil/pub conflicts deux

### LINKED ISSUE #1634: agent: controller publications should bypass authorization checks
Automated publications that are performed by controllers should bypass authorization checks. We currently require controllers to assume the `estuary_support/` role, which must have an authorization grant to every prefix in the system. We'd like to instead have agents bypass authZ checks for publications that are triggered by our automation.

psFried:
**Description:**

Rolls up some refactors of the publish api and the remainder of the work for #1520. 

- Introduces automated retries of `Build/PublicationSuperseded` errors (#1520)
- Controller-initiated publications now bypass user authorization checks (resolves #1634)
- All non-touch publications of collection specs will now update the embedded inferred schema definition (#1520)
- Discovers will no longer set `expect_pub_id` on drafted collection specs (#1520)

This change should not really be visible to end users.
The one caveat is that now the `publications.id` column will no longer match up with `live_specs.last_pub_id` or `publication_specs.pub_id`, since the true publication id is now determined by the publisher. I added the `publications.pub_id` column in order to allow joining `publications` to those tables, since that's still useful for debugging.

Note that the `publication_specs_ext` view is unaffected by that change, since it does not join to `publications`. Thus `flowctl catalog history` still works without any changes.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1662)


psFried:
I moved this back to a draft. At this point, I think it'd be unwise to merge/deploy it until after I'm back from vacation, so I'm thinking I might as well fold in the inferred schema changes to finish out the rest of #1520 .

crates/agent/src/api/create_data_plane.rs
```diff
@@ -184,30 +187,23 @@ async fn do_create_data_plane(
...
-
-    _ = publisher
-        .commit(built)
+    let publication = DraftPublication {
+        user_id,
+        logs_token: insert.logs_token,
+        draft,
+        dry_run: false,
+        detail: Some(format!("publication for data-plane {base_name}")),
+        // TODO: Should we verify user authz for publications to create a data plane?
```
jgraettinger, line 196: that's currently handled at L69 at the routine entrypoint: we require the user is an admin of `ops/`.

crates/agent/src/api/update_l2_reporting.rs
```diff
@@ -174,41 +173,32 @@ export class Derivation extends Types.IDerivation {"#
...
-        for err in uncommitted.output.errors() {
-            tracing::error!(scope=%err.scope, err=format!("{:#}", err.error), "data-plane-template build error")
+    let publication = DraftPublication {
+        user_id,
+        logs_token,
+        draft,
+        dry_run,
+        detail: Some(format!("publication for updating L2 reporting")),
+        default_data_plane_name: Some(default_data_plane.clone()),
+        // TODO: should we verify user authz for updating L2 reporting?
```
jgraettinger, line 184: Same story, it's done first-thing by the routine, unless I'm not understanding the question.
I suppose a difference here is that the user could be an admin of `ops/` but  _not_ be an admin of `ops.us-central1/` and then it would fail? Meh :shrug:

crates/agent/src/publications/finalize.rs
```diff
@@ -1,12 +1,53 @@
...
+
+/// A `FinalizeBuild` that doesn't modify the build in any way.
+pub struct NoopFinalize;
+impl FinalizeBuild for NoopFinalize {
+    fn finalize(&self, _build: &mut UncommittedBuild) -> anyhow::Result<()> {
+        Ok(())
+    }
+}
+
+/// A `FinalizeBuild` that removes and built collection specs that are not referenced by other
```
jgraettinger, line 28: supernit: removes and built => removes any built

> Note that this must be used in conjuction with the `ExpandDraft`  `Initialize` impl, or else it can prune collections that are referenced by live specs that are not included in the draft.

Also, is it a problem that it might remove collections referenced only by pass-through live specs? I'm not seeing why.

crates/agent/src/publications/initialize.rs
```diff
@@ -0,0 +1,100 @@
...
+    }
+}
+
+/// An `Initialize` that expands the draft to touch live specs that read from or write to
+/// any drafted collections. This may optionally filter the specs based on whether the user
+/// has `admin` capability to them.
+pub struct ExpandDraft {
+    /// Whether to filter specs based on the user's capability. If true, then only specs for which
+    /// the user has `admin` capability will be added to the draft.
+    pub filter_user_has_admin: bool,
```
jgraettinger, line 34: I've been debating whether `admin` is too restrictive -- I could see an argument that Alice's update of C should be able to observe / touch-publish a materialization M for which they have read but not admin.

But ultimately I think you're right, and it should be admin, because we don't want to fail Alice's publish of C due to a materialization M that they're unable to change. That's the admin of M's problem, and they'll see a controller error about the issue.

This is a departure from my prior understanding of how spec expansion should work. In the past, I've thought that you _would_ see specs that make use of a read grant you provide for your collection, which you yourself have no grant. Controllers and the eventual consistency they provide have changed the narrative here.

crates/agent/src/publications/initialize.rs
```diff
@@ -0,0 +1,100 @@
...
+        let expanded_rows = agent_sql::live_specs::fetch_expanded_live_specs(
+            user_id,
+            &drafted_collections,
+            &all_drafted_specs,
+            db,
+        )
+        .await?;
+        let mut expanded_names = Vec::with_capacity(expanded_rows.len());
+        for exp in expanded_rows {
+            if self.filter_user_has_admin
```
jgraettinger, line 68: should this be pushed down into `fetch_expanded_live_specs` ? I'm betting that routine can be simplified a bunch given the refined / simpler authorization expectations noted above :point_up:

jgraettinger (commented):
This looks great so far! Really like the DraftPublication refactor.

crates/agent/src/publications/finalize.rs
psFried: Yeah, this comment was based on the faulty assumption that we might prune draft specs that already have corresponding live specs. But we only consider _new_ specs for pruning, so there's no chance that a pruned draft spec might have been referenced by an existing live spec. I updated the comment.

crates/agent/src/publications/initialize.rs
psFried: Yeah, good call out. Draft expansion is no longer guaranteed to be exhaustive, and we'll rely on controllers to report errors that appear in connected specs that aren't able to be included in the expanded draft.

FWIW, this was already the behavior that's used in prod. The `filter_user_has_admin` variable was added primarily in order to make that behavior more explicit. But I also thought it might be nice to allow disabling that check, so that `ExpandDraft` could at least theoretically be used with a publication having `verify_user_auth: false`. We don't currently have a use case for that, since controller-initiated publications never use spec expansion. But perhaps it could be useful if we ever want to do an exhaustive check to ensure that a given change won't break anything.

crates/agent/src/publications/initialize.rs
psFried: IMO it doesn't really seem worth it at this stage. Doing it in the db just saves us fetching specs that we'd just filter out. But at present there's only a handful of specs in the whole system that would actually be filtered out by this. Maybe worthwhile once we get more people sharing collections, though.

jgraettinger (approved):
LGTM

Issue #1661, opened 2024-09-26
Extend the `live_specs_ext` view to include basic information about matched data-planes


kiahna-tucker:
Extend the `live_specs_ext` view to include the following `data_planes` table columns: `data_plane_name`, `reactor_address`, `broker_address`. This is a dependency for a couple client features that aim to surface data-plane information on entity _Details_ pages (ref: [issue 1273](https://github.com/estuary/ui/issues/1273), [issue 1277](https://github.com/estuary/ui/issues/1277)).

PR #1664, opened 2024-09-27
Extend live_specs_ext to include basic data-plane information

kiahna-tucker:
**Description:**

Extend the `live_specs_ext` view to include the following `data_planes` table columns: `data_plane_name`, `reactor_address`, `broker_address`. This is a dependency for a couple client features that aim to surface data-plane information on entity _Details_ pages.

**Issues:**

The issues directly below are completely resolved by this PR:
https://github.com/estuary/flow/issues/1661

**Workflow steps:**

Execute a SQL query that selects the columns added to the `live_specs_ext` view; or run the migration locally and observe that the new columns appear in the `live_specs_ext` view in the Supabase UI.

N/A

**Notes for reviewers:**

N/A

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1664)


supabase/migrations/65_extend_live_specs_view.sql
```diff
@@ -0,0 +1,66 @@
...
+    l.data_plane_id,
+    d.broker_address,            -- Added column
+    d.data_plane_name,           -- Added column
+    d.reactor_address            -- Added column
+FROM live_specs l
+LEFT JOIN publication_specs p ON l.id::macaddr8 = p.live_spec_id::macaddr8 AND l.last_pub_id::macaddr8 = p.pub_id::macaddr8
+LEFT JOIN connectors c ON c.image_name = l.connector_image_name
+LEFT JOIN connector_tags t ON c.id::macaddr8 = t.connector_id::macaddr8 AND l.connector_image_tag = t.image_tag
+LEFT JOIN internal.user_profiles u ON u.user_id = p.user_id
+LEFT JOIN data_planes d ON d.id::macaddr8 = l.data_plane_id::macaddr8
```
kiahna-tucker, line 52: This join was introduced in this PR as well. Let me know if it should be called out by a comment in the file (and, obviously, if it needs to be altered).

supabase/migrations/65_extend_live_specs_view.sql
kiahna-tucker, line 1: For the life of me, I could not think of a pithy name for this migration that wasn't too ambiguous, verbose, or closely mirroring another file.

psFried: Goodenuf!

supabase/migrations/65_extend_live_specs_view.sql
```diff
@@ -0,0 +1,66 @@
...
+WHERE (
+    EXISTS (
+        SELECT 1
+        FROM pg_roles
+        WHERE pg_roles.rolname = CURRENT_ROLE AND pg_roles.rolbypassrls = TRUE
+    )
+) OR (
+    l.id::macaddr8 IN (
```
psFried, line 60: Optional nit: these macaddr8 casts doesn't seem necessary or clarifying

psFried (commented):
LGTM % one question
It seems possible for a user to have read access to a task, but not the data plane. Is that actually the case? If so, how should we handle it? Should we try to filter out the addresses here? Or just say that task read access is good enough to see the data plane addresses, regardless? Perhaps @jgraettinger would want to weigh in?

jgraettinger (approved):
Also LGTM.

Yep, it's possible that someone has read of task T, but does not have read of the data-plane D that T lives within.

I think it's totally okay for the user to see that D exists and its name based on their read capability to T.

The interpretation of "read" for a data-plane is a little weird -- it means that one is able to 1) independently discover the existence of the data-plane from the DB and 2) deploy new tasks or collections which use the data-plane. Neither of those abilities is conferred here.

It does mean that one can discover the existence of a private data-plane you otherwise wouldn't know about, because you observe that T is running within it. I think this is desired, because the user's ability to read T already implies a relationship between the data-plane owner and the user.

PR #1683, opened 2024-10-07
agent: fix pub status incompatible_collections

psFried:
The `incompatible_collections` field on the publication `job_status` was not correctly accounting for disabled bindings, which caused it to output incorrect collection names in some cases. This corrects that.

Additionally, the `resource_path` of affected bindings was added to the `AffectedConsumer` struct, so that we can identify specific bindings in cases where there are multiple bindings with the same source collection. In the future, we can enhance `evolutions` to identify bindings by resource path instead of collection name.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1683)

### DIFF
--- a/crates/agent/src/publications/status.rs
+++ b/crates/agent/src/publications/status.rs
@@ -110,6 +110,10 @@ pub struct IncompatibleCollection {
 pub struct AffectedConsumer {
     pub name: String,
     pub fields: Vec<RejectedField>,
+    /// Identifies the specific binding that is affected. This can be used to differentiate
+    /// in cases there are multiple bindings with the same source.
+    #[serde(default, skip_serializing_if = "Vec::is_empty")]
+    pub resource_path: Vec<String>,
 }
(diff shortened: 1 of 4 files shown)


crates/agent/src/publications/status.rs
```diff
@@ -141,13 +145,25 @@ pub fn get_incompatible_collections(output: &tables::Validations) -> Vec<Incompa
...
-                let collection_name = model.bindings[i].source.collection().to_string();
+                // We must skip over disabled bindings in order to translate the index of the
+                // validated binding to the index of the model binding.
+                let collection_name = model
+                    .bindings
+                    .iter()
+                    .filter(|b| !b.disable)
+                    .skip(i)
+                    .next()
+                    .unwrap() //
```
jshearer, line 156: nit: left over comment

jshearer (approved):
LGTM!

PR #1684, opened 2024-10-07
control-plane: add constraint for grant_capability

### LINKED ISSUE #1675: agent: add check constraint to `grant_capability` type
We had a production outage that was caused by a `role_grants` row having a `grant_capability` value outside of the named `read`, `write`, `admin` roles. We should add a check constraint to prevent adding or updating a capability to anything outside of those three roles.

psFried:
Ensures that no user or role grants can be created with the placeholder capabilities.

Fixes #1675

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1684)


jshearer (approved):
LGTM

PR #1688, opened 2024-10-08
agent: setup rustls CryptoProvider

psFried:
Fixes a crash in `agent` due to missing a `CryptoProvider` for `rustls` TLS connections.

I was able to reproduce this locally, and confirmed that the fix is working.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1688)


jgraettinger (approved):
LGTM

PR #1689, opened 2024-10-08
dekaf: Always make sure we're using fresh tokens

jshearer:
I was noticing a bunch of `ExpiredSignature` errors, and realized that if a client didn't close its connection on read errors (which it appears is sometimes the case), a new access token would never get created. This ensures that every time we get a `flow_client::Client`, we call `refresh_authorizations()` to make sure it's using valid tokens. This is cheap in the common case that the tokens aren't expired.

Also removes refresh token from `flow_client::Client`, as refreshing is a concern that should live outside of the Client (and indeed, `refresh_token` was no longer user anywhere).

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1689)

### DIFF
--- a/crates/dekaf/src/lib.rs
+++ b/crates/dekaf/src/lib.rs
@@ -50,8 +50,34 @@ pub struct DeprecatedConfigOptions {
 
 pub struct Authenticated {
     client: flow_client::Client,
+    refresh_token: RefreshToken,
+    access_token: String,
     task_config: DekafConfig,
     claims: models::authorizations::ControlClaims,
+}
+
+impl Authenticated {
+    pub async fn get_client(&mut self) -> anyhow::Result<&flow_client::Client> {
+        let (access, refresh) = refresh_authorizations(
+            &self.client,
+            Some(self.access_token.to_owned()),
+            Some(self.refresh_token.to_owned()),
+        )
+        .await?;
+
+        if access.ne(&self.access_token) {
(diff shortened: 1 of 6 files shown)


psFried (commented):
I don't see anything that's _incorrect_ here, but I do see a few things that seem maybe worth improving.

My understanding is that `get_client` needs to be called before pretty much every request, so that it can refresh the access token if necessary. That seems to go against the (I think?) commonly held assumption that a `Client` can be long lived and re-used indefinitely.

It also re-parses (and I'm assuming verifies?) the JWT each time. Probably not a big deal in terms of real perf impact, but it does seem like the expiration timestamp is something we could keep in memory alongside the token. This would also make the expiration check pretty trivial, so it could happen in `get_client` instead of in `refresh_token`. That seems beneficial to me since it feels a bit tricky for a function called `refresh_token` to sometimes _not_ refresh the token.

Putting it all together, I'm thinking it could look roughly like:

```rust
struct AccessToken {
  expiration: DateTime<Utc>,
  token: String,
}
```

and in `get_client`:

```rust
if self.access_token.expires_within(ONE_MINUTE) {
  let (access, refresh) = refresh_token(...).await?;
  ...
```

Let me know what you think. I don't think it's strictly _required_ to refactor, but my gut says it's probably worth it. I may be missing something, though.

psFried (approved):
LGTM

PR #1692, opened 2024-10-11
flowctl: initialize rustls crypto provider

psFried:
Add the required initialization for the `rustls` crypto provider to the beginning of the agent main function, so that TLS will work again.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1692)


jshearer (approved):
LGTM

PR #1693, opened 2024-10-11
dekaf: Fix `Fetch` encoding issue for some versions of librdkafka

jshearer:
I wrongly assumed that `kafka-protocol` would do the right thing when passing `None` to a fetch response's records. It appears that it encodes this with a length of `-1`, but I found that caused issues when testing locally:

```
Protocol parse failure for Fetch v11 ... invalid MessageSetSize -1
```

It still reads documents when the chunk returns some, but otherwise it logs that error frequently. This fixes it

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1693)

### DIFF
--- a/.github/workflows/dekaf.yaml
+++ b/.github/workflows/dekaf.yaml
@@ -14,8 +14,8 @@ on:
       - "Cargo.lock"
 
 jobs:
-  build-and-test:
-    runs-on: ubuntu-latest
+  build-dekaf:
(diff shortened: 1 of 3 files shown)


jshearer:
Confirmed that this fixes the errors, and also fixes the Dekaf build

psFried (approved):
LGTM

PR #1698, opened 2024-10-14
validation: add inferred schema placeholder if needed

psFried:
This fixes a bug in `flowctl`, where it would fail to load catalog specs that were missing an inlined `flow://inferred-schema` definition.

Commit 7190b42faf made the agent update collection inferred schemas as part of the initialization of a draft publication instead of updating the inferred schema as part of validation. Unfortunately, validation is called in other contexts as well, such as `flowctl::local_specs`, where it would be undesirable to have the caller need to inject the inferred schema placeholder. So this adds the inlining of the inferred schema back into the `validation` crate, but only in cases where there isn't already an existing `$defs` entry for it. In other words, it will only ever inject the placeholder, and only if necessary in order to resolve references within the schema.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1698)

### DIFF
--- a/crates/models/src/schemas.rs
+++ b/crates/models/src/schemas.rs
@@ -100,7 +100,7 @@ impl Schema {
         write_bundle: Option<&Self>,
         inferred_bundle: Option<&Self>,
     ) -> Self {
-        use serde_json::{value::to_raw_value, Value};
+        let mut defs = Vec::new();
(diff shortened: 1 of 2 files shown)


jgraettinger (approved):
LGTM

PR #1699, opened 2024-10-15
flowctl: build using clang instead of gcc

psFried:
Switches to using clang instead of GCC in order to work around a compilation error in one of our dependencies, which is due to a bug in GCC 9. See:
https://github.com/rustls/rustls/issues/1967

Unfortunately, I can't think of a way to test this locally, since there's no way to pull the github runner image. I think we'll just need to merge and try the release again and see whether it works :grimacing:

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1699)

### DIFF
--- a/.github/workflows/flowctl-release.yaml
+++ b/.github/workflows/flowctl-release.yaml
@@ -33,6 +33,9 @@ jobs:
       # Linux build steps:
       - name: Build Linux
         if: matrix.config.os == 'ubuntu-20.04'
+        env:
+          CC: clang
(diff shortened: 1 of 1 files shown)


mdibaiee (approved):

psFried:
Tested this by pushing a change to the workflow yaml to have it build the release binary during a PR build. It worked, so I just removed that commit and will merge without waiting for the current (unnecessary) builds to finish.

PR #1700, opened 2024-10-15
dekaf: apply a 30 second idle session timeout

jgraettinger:
Require that clients send a request within thirty seconds, or close the connection.

Tested with a shorter timeout using a local franz-go client.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1700)

### DIFF
--- a/crates/dekaf/src/main.rs
+++ b/crates/dekaf/src/main.rs
@@ -17,7 +17,7 @@ use std::{
     path::{Path, PathBuf},
     sync::Arc,
 };
-use tokio::io::{split, AsyncRead, AsyncWrite, AsyncWriteExt};
+use tokio::io::{AsyncRead, AsyncWrite, AsyncWriteExt};
 use tracing_subscriber::{filter::LevelFilter, EnvFilter};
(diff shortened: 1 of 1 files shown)


crates/dekaf/src/main.rs
```diff
@@ -261,41 +261,43 @@ where
...
             out.clear();
         }
-
-        res
     }
     .await;
 
     metrics::gauge!("dekaf_total_connections").decrement(1);
 
+    w.shutdown().await?;
```
jshearer, line 300: What's your thinking on doing this after vs before decrementing the gauge? This way the gauge gets decremented even if there's an error shutting down the socket, which... probably still results in a closed socket in the end, so makes the gauge more accurate?

jshearer (approved):
I don't think we should have any sessions normally hanging around with no activity in them, so by default I would expect this to not do much, _unless_ we're somehow leaking Sessions, which this would surface quickly. LGTM 👍

PR #1705, opened 2024-10-16
Fix `flowctl auth ...` not properly exchanging access token for refresh token

### LINKED ISSUE #1703: flowctl: auth not working
A user complained that all their `flowctl` invocations (like `catalog list`) were returning `Error: 401 Unauthorized` even though they had just recently run `flowctl auth login`.

I've been able to reproduce the issue by editing my flowctl config so that it only contains `user_access_token` and not a `refresh_token`. For example:

`~/.config/flowctl/default.json`:
```json
{
  "user_access_token": "eyJhbGc..."
}
```

Once I did that, running `flowctl catalog list` returns a `401` error, even though the `user_access_token` is still valid.

After a bit more playing around, I realized that running `flowctl auth login` when there's no existing config file (the exact scenario that all new users would be in) will also result in the config file containing only `user_access_token`, and all other commands result in the `401` error.

**annoying warning:**

There's one last issue, which isn't _technically_ related, but should probably be fixed along with this: running `flowctl auth login` will print a warning that tells me to run `flowctl auth login`:

```
WARN flowctl: You are not authenticated. Run `auth login` to login to Flow.
```

This makes it feel broken, even if was working correctly.

jshearer:
**Description:**

This bug was introduced in https://github.com/estuary/flow/pull/1665. I tested the `flowctl auth` flow as part of that PR, but apparently didn't exercise the branch used when we have an access token, but no refresh token (i.e authenticating for the first time). This fixes that branch by always ensuring the client used to create the refresh token is authenticated, and also makes sure to generate and store a refresh token on the `Config` as part of  the `auth` subcommand.

Also tested `auth token`, which was similarly broken and is similarly fixed.

Fixes https://github.com/estuary/flow/issues/1703

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1705)

### DIFF
--- a/crates/flow-client/src/client.rs
+++ b/crates/flow-client/src/client.rs
@@ -312,7 +312,7 @@ pub async fn refresh_authorizations(
         (Some(access), None) => {
             // We have an access token but no refresh token. Create one.
             let refresh_token = api_exec::<RefreshToken>(
-                client.rpc(
+                client.clone().with_creds(Some(access.to_owned())).rpc(
                     "create_refresh_token",
                     serde_json::json!({"multi_use": true, "valid_for": "90d", "detail": "Created by flowctl"})
                         .to_string(),
--- a/crates/flowctl/src/auth/mod.rs
+++ b/crates/flowctl/src/auth/mod.rs
@@ -1,6 +1,7 @@
 mod roles;
 
 use anyhow::Context;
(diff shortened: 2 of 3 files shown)


jgraettinger (approved):
LGTM

PR #1711, opened 2024-10-16
two fixes of observed production issues

jgraettinger:
**Description:**

See commits:
* A task term context cancellation shouldn't prevent us from creating physical partitions
* A task delete-then-recreate shouldn't invalidate the SNI cache used for connector networking

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1711)


jgraettinger:
I was able to get a reliable reproduction on a local stack by having a logically partitioned "hello world" collection, which I raced with updates to the capture spec from the control plane. The task would wedge upon the control plane publication while attempting to create a partition, which is what I observed in the production example.

With this patch, I could no longer reproduce even when spamming publications over and over, etc.

psFried (approved):
LGTM

PR #1716, opened 2024-10-18
dekaf: Remove throttling in `FetchResponse`, let clients poll if they want to

jshearer:
I originally added this polling request as a "couldn't hurt" change to slow down clients polling in a tight loop for new documents, but I realized two things:

* Polling like that is not very costly to serve. Reading is done asynchronusly, and buffered per journal and session. So, so long as a reader keeps using the same session, multiple polling requests to the same journal will all use the same reader and buffer.
* It looks like in "real" Kafka, a throttle request is indicative of abnormal operation, and I theorize that consumers are using it as a signal to back off much more aggressively than requested. This is bad in a situation where you want close to real-time data to flow.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1716)

### DIFF
--- a/crates/dekaf/src/session.rs
+++ b/crates/dekaf/src/session.rs
@@ -366,8 +366,6 @@ impl Session {
         let timeout_at =
             std::time::Instant::now() + std::time::Duration::from_millis(max_wait_ms as u64);
 
-        let mut hit_timeout = false;
-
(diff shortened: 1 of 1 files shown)


jshearer:
Rolled into https://github.com/estuary/flow/pull/1701

PR #1717, opened 2024-10-18
agent: add diagnostics for db connections

psFried:
- Adds periodic logging of connection pool size
- Sets the application name to the k8s pod name, to better correlate with pg_stat_activity
- Sets a connection acquisition timeout
- Configures logging of slow statements
- Pings the connection when returning it to the pool, so that we can detect and remove bad connections

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1717)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -124,6 +124,7 @@ dependencies = [
  "jsonwebtoken",
  "labels",
  "lazy_static",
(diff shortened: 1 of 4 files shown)


jshearer (approved):
LGTM

PR #1720, opened 2024-10-18
Johnny/dp timeout

psFried:
Wraps connector and gazette client dials in timeouts, since the `connect_timeout` grpc options are insufficient to cover all cases that might block indefinitely. This is intended to fix issues with agents that stop making progress.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1720)


jshearer (approved):
LGTM

Interesting, this smells very similar to the issues I was seeing with Dekaf where after some period of time gRPC requests would just hang, and my temporary work-around was to just dial a whole new TCP connection for each session to get around it.

jgraettinger (approved):
LGTM

PR #1724, opened 2024-10-21
Adding subcommand to list shard in flowctl.

### LINKED ISSUE #1710: flowctl: add subcommands for listing shards
With the introduction of federated data planes, our current debugging workflow of using `flowctl-go` has become more inconvenient and error prone. We should add a subcommand to `flowctl` for listing shards, which will determine the correct data plane and acquire an authorization token automatically.

n0tank3sh:
**Description:**

 Fixes  #1710

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1724)

### DIFF
--- a/crates/flow-client/src/client.rs
+++ b/crates/flow-client/src/client.rs
@@ -1,5 +1,9 @@
 use crate::{api_exec, parse_jwt_claims};
 use anyhow::Context;
+use gazette::{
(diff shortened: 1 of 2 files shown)


psFried:
Sorry for the slow response here, but we not currently accepting PRs from outside contributors unless it's something specific we've discussed.

PR #1725, opened 2024-10-21
flow-client: Make sure to remove any potentially expired JWT from the client used to exchange a refresh token for an access token in `refresh_authorizations()`

jshearer:
While running down some issues with Dekaf, namely frequent consumer group rebalances, I noticed that these rebalances correlated with Dekaf errors such as:

```
dekaf: error=failed to obtain access token

Caused by:
    Unauthorized: {"code":"PGRST301","details":null,"hint":null,"message":"JWT expired"}
```

I then realized that Dekaf was passing a client containing an expired access token to `refresh_authorizations`. Since `generate_access_token` RPC performs its own authentication internally by way of the provided refresh token, it shouldn't be called with an authorization header.

This isn't a critical bug as it should just disconnect the session, causing a new connection to be created that'll have a freshly authenticated client, but nevertheless it's good to fix.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1725)

### DIFF
--- a/crates/flow-client/src/client.rs
+++ b/crates/flow-client/src/client.rs
@@ -59,16 +59,9 @@ impl Client {
         }
     }
 
-    pub fn with_creds(self, user_access_token: Option<String>) -> Self {
+    pub fn with_user_access_token(self, user_access_token: Option<String>) -> Self {
         Self {
-            user_access_token: user_access_token.or(self.user_access_token),
-            ..self
-        }
-    }
-
-    pub fn as_anonymous(self) -> Self {
-        Self {
-            user_access_token: None,
+            user_access_token,
             ..self
         }
     }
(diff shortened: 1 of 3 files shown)


crates/flow-client/src/client.rs
```diff
@@ -331,10 +338,15 @@ pub async fn refresh_authorizations(
...
+            // We either never had an access token, or we had one and it expired,
+            // in which case the client may have an invalid access token configured.
+            // The `generate_access_token` RPC only needs the provided refresh token
+            // for authentication, so we should use an unauthenticated client to make
+            // the request.
             let Response {
                 access_token,
                 refresh_token: next_refresh_token,
-            } = api_exec::<Response>(client.rpc(
+            } = api_exec::<Response>(client.clone().as_anonymous().rpc(
```
jgraettinger, line 349: nit: couldn't this just be `with_creds(None)` ?

I also think that method should be renamed `with_user_access_token()` to match `new()` and the field itself...

jgraettinger (approved):
LGTM % comment

crates/flow-client/src/client.rs
jshearer: > I also think that method should be renamed with_user_access_token() to match new() and the field itself...

Yup, now that the client doesn't have a refresh token, that sounds right to me

> nit: couldn't this just be with_creds(None) ?

Yes. I don't remember why `with_creds` does `user_access_token: user_access_token.or(self.user_access_token)`, but none of the usages of it currently pass anything other than `Some(..)`, so I think it's safe to make it simply return 
```rust
        Self {
            user_access_token,
            ..self
        }
```

jgraettinger (approved):

PR #1726, opened 2024-10-21
dekaf: Refresh `Read`'s journal client auth token when it expires

jshearer:
**Description:**

This is a follow-up to https://github.com/estuary/flow/pull/1725. I was still seeing some JWT expired errors, this time they looked like this:

```
WARN serve{...}: dekaf: error=status: Unauthenticated, message: "parsing jwt: token is expired by 12m4.003303197s", details: [], metadata: MetadataMap { headers: {} }
```

After some more digging, I realized that a sufficiently long-lived session containing a sufficiently long-lived `Read` could run into this issue, as the Read's journal client is never refreshed. This fixes that.

**Note:** I did consider whether it would be cleaner to refactor Read to just take a `flow_client::Client` and wholly manage its own journal client, but it didn't seem better enough to justify the bigger refactor. I'm low conviction on that though, happy to do that refactor if someone else feels more strongly about it.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1726)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1880,6 +1880,7 @@ dependencies = [
  "tokio",
  "tokio-rustls 0.26.0",
  "tokio-util",
+ "tonic",
  "tower-http",
  "tracing",
  "tracing-subscriber",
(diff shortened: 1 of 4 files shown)


jshearer:
Closing as the current behavior of closing the connection on token timeout is actually desired, as it surfaces issues wrt connections closing/opening more frequently

PR #1730, opened 2024-10-22
dekaf: Add metrics to track fetch request counts by topic and partition

jshearer:
**Description:**

In order to debug what's happening with infrequent consumer lags, we want to keep track of how many read requests Dekaf is getting, broken down by collection and partition. This tracks that, as well as breaking down by read request type:

* `state => read_pending` means the read request came in after a previous request had already fetched the same offset for the topic and partition, and that request is pending.
* `state => collection_not_found` and `state => partition_not_found` are self-evident, though I believe a normal Kafka consumer would have figured out that these topics don't exist during the metadata discovery phase
* `state => new_data_preview_read` indicates a fetch request that we flagged as being for a data preview. These requests have special handling and poison the entire connection to only be usable for other data preview reads
* `state => new_regular_read` indicates a fetch request for a new topic/partition/offset combination.

Ex: 
```
# TYPE dekaf_fetch_requests counter
dekaf_fetch_requests{topic_name="joseph/dekaf-testing",partition_index="0",state="new_regular_read"} 19
dekaf_fetch_requests{topic_name="joseph/dekaf-testing",partition_index="0",state="read_pending"} 248
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1730)

### DIFF
--- a/crates/dekaf/src/session.rs
+++ b/crates/dekaf/src/session.rs
@@ -455,9 +455,23 @@ impl Session {
                 };
 
                 if matches!(self.reads.get(&key), Some(pending) if pending.offset == fetch_offset) {
+                    metrics::counter!(
+                        "dekaf_fetch_requests",
+                        "topic_name" => key.0.to_string(),
+                        "partition_index" => key.1.to_string(),
+                        "state" => "read_pending"
+                    )
+                    .increment(1);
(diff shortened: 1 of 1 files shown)


jshearer:
Rolled into https://github.com/estuary/flow/pull/1733

PR #1734, opened 2024-10-24
simd-doc: Add failing test for parsing incomplete rows

jshearer:
**Description:**

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1734)

### DIFF
--- a/crates/simd-doc/src/tests/fixtures.rs
+++ b/crates/simd-doc/src/tests/fixtures.rs
@@ -102,6 +102,127 @@ fn test_simd_and_fallback_results_are_equal() {
     assert!(!failed);
 }
 
+/// Make sure that we don't parse and return partial data from the middle of a row
+#[test]
+fn test_incomplete_row_parsing() {
+    let mut inputs = Vec::new();
+    inputs.push(r###"st": {"this": "is", "a": "doc"}}"###.as_bytes());
+    inputs.push(r###""test": {"this": "is", "a": "doc"}}"###.as_bytes());
+    inputs.push(r###"{"test": 5}, {"test": 6"}]"###.as_bytes());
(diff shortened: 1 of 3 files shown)


crates/simd-doc/src/tests/fixtures.rs
```diff
@@ -166,76 +167,91 @@ fn test_incomplete_row_parsing() {
...
+                    },
+                ),
+            ],
+        ),
+        (
+            "\"test\": {\"this\": \"is\", \"a\": \"doc\"}}\n{\"this\": \"is a real doc\"}\n",
+            [
+                (
+                    -1,
+                    String("trailing characters at line 1 column 7 @ 0..36"),
```
jshearer, line 197: Where does this message come from? It doesn't match the new errors you added to the parser: `missing trailing newline`

jshearer:
This LGTM btw -- I created the PR so I can't approve it, but would be good to merge this at some point, my question above isn't blocking (though I am curious)

jgraettinger (approved):
LGTM

crates/simd-doc/src/tests/fixtures.rs
jgraettinger: Yep. The Parser attempts to use simd-json first, which almost always works. If simd-json fails, though, we throw a c++ exception which gets turned into a Result by the `cxx` crate.

We handle this by using a fallback path which does the same job but using serde_json, which produces better errors and is able to handle documents of larger sizes (simd-json is constrained to 1MB documents).

So tl;dr these errors are produced by the serde_json fallback after the "missing excepted newline" exception is thrown / caught / handled.

PR #1736, opened 2024-10-24
dekaf: Expire `Read`s after 5m

jshearer:
**Description:**

We saw problems before where something about a `journal::Client` would stop responding after some amount of time in use. We were seeing issues where consumers would stop seeing updates to journals after a certain amount of time, and I just realized that it's possible for a `Read` (which contains a `journal::Client`) to live for a long time. This times out a `Read` after 5 minutes, so it get a new `journal::Client`, and hopefully bypasses the issue.

Tested locally with `kcat`, will test in `dekaf-dev` when build passes

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1736)

### DIFF
--- a/crates/dekaf/src/session.rs
+++ b/crates/dekaf/src/session.rs
@@ -41,7 +41,7 @@ enum SessionDataPreviewState {
 
 pub struct Session {
     app: Arc<App>,
-    reads: HashMap<(TopicName, i32), PendingRead>,
+    reads: HashMap<(TopicName, i32), (PendingRead, std::time::Instant)>,
     secret: String,
     auth: Option<Authenticated>,
     data_preview_state: SessionDataPreviewState,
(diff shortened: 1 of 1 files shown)


jgraettinger (approved):
LGTM

crates/dekaf/src/session.rs
```diff
@@ -454,16 +454,33 @@ impl Session {
...
+                            "topic_name" => key.0.to_string(),
+                            "partition_index" => key.1.to_string(),
+                            "state" => "read_pending"
+                        )
+                        .increment(1);
+                        continue; // Common case: fetch is at the pending offset.
+                    }
+                    _ => {}
                 }
+
```
jgraettinger, line 483: what's the mechanism by which this session will grab an entirely new gazette::Router (e.g. new underlying connection) for the next started read?

jshearer: I was thinking that the call to `authenticated_client()` always created a new gazette client, but it turns out it only does when refreshing the access token. I updated this to explicitly call `client.with_fresh_gazette_client()` upon timeout. We could also go nuclear and have `authenticated_client()` _always_ return a client with fresh router, but that seemed like it might be unnecessary. Thoughts?

PR #1738, opened 2024-10-25
dekaf: Implement `DeletionMode` to allow representing deletions as a Kafka header instead of a tombstone

jshearer:
This adds support for customizable deletion handling:

* `DeleteMode::Default`, which is what you get if you don't specify a deletion mode, operates the same was as before where deletes are mapped into Kafka records with a key, and null value.
* `DeleteMode::Header` changes this behavior to include an `_is_deleted` header on every returned Record. Non-deletes get `0`, deletes get `1`. In addition, the full deletion tombstone stored in the Flow collection is returned for the Record's value instead of null.

I tested this out in Tinybird and it looks like this works as expected: setting a username of `{"deletions": "header"}` causes Flow deletion tombstones (`op: d`) to cause the corresponding row to get deleted in Tinybird. 

**Question:** Should we just go all the way and name this `DeleteMode::Tinybird`? Or does someone have a better suggestion than `Header`? 

Also included is a small change to `Read` to make data preview UIs work reliably again. I realized they stopped working consistently after merging https://github.com/estuary/flow/pull/1733, and this fixes that.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1738)


jgraettinger (approved):
LGTM :+1:

crates/dekaf/src/read.rs
```diff
@@ -267,12 +270,24 @@ impl Read {
...
                     None
                 } else {
                     tmp.push(0);
                     tmp.extend(self.value_schema_id.to_be_bytes());
-                    () = avro::encode(&mut tmp, &self.value_schema, root.get())?;
+
+                    if matches!(self.deletes, DeletionMode::CDC) {
+                        let mut heap_node = HeapNode::from_node(root.get(), &self.alloc);
+                        let foo = DELETION_INDICATOR_PTR
+                            .create_heap_node(&mut heap_node, &self.alloc)
```
jgraettinger, line 282: there's a memory leak here -- you're never resetting the bump allocator.

I would recommend _not_ having it as a field of `self` and instead creating one on the stack. Use it, and then call reset() on it immediately after each avro::encode so that it's only retaining a single HeapNode.

jshearer: I was going to ask you to take a look at this as I wasn't sure I was doing it correctly. That makes sense to me, thanks 👍

crates/dekaf/src/topology.rs
```diff
@@ -87,7 +89,16 @@ impl Collection {
...
             &spec.read_schema_json
         };
-        let (key_schema, value_schema) = avro::json_schema_to_avro(json_schema, &key_ptr)?;
+
+        let json_schema = doc::validation::build_bundle(json_schema)?;
+        let validator = doc::Validator::new(json_schema)?;
+        let mut shape = doc::Shape::infer(&validator.schemas()[0], validator.schema_index());
+
+        if matches!(deletion_mode, DeletionMode::CDC) {
+            shape.widen(&serde_json::json!({"_meta":{"is_deleted":1}}));
```
jshearer, line 98: @jgraettinger does this seem like the best way to go about adding a new field to the schema? It was the simplest way I found, but do you think that somehow directly modifying the `Shape` would be better?

jgraettinger: seems reasonable to me

PR #1739, opened 2024-10-27
data-plane-controller for managing data-plane life cycles

### LINKED ISSUE #1727: automations for data-plane operations

jgraettinger:
**Description:**

The controller uses the recent `automations` crate to monitor structured
changes to a bound data_planes and drive the data-plane to convergence.
    
It's modeled as a CI/CD pipeline, where an indicated branch of our
dry-dock repo (containing Pulumi and Ansible infrastructure) is deployed
for the given data-plane.
    
The controller performs the full lifecycle required for rolling updates:
    
- `pulumi up` to create new resources or respond to replacements
- Awaiting DNS propagation
- Running the Ansible playbook to provision instances
- Running `pulumi up` to reflect readiness of started instances
- Awaiting DNS propagation (again)
    
It also periodically refreshes a stack from remote providers to detect
changed or deleted resources, such as EC2 instance replacements, and
responds accordingly to heal the infrastructure.

A number of sanity-checks are built in to verify that we're modifying data-plane configurations in allowed ways, and that we're not performing modifications while the controller is actively driving the data-plane towards convergence.

The controller also publishes a variety of exported Pulumi outputs which customers need to know about, such as IAM users and GCP service accounts, AWS private link bindings, and VPC CIDR blocks.

Fixes #1727 

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1739)

### DIFF
--- a/crates/data-plane-controller/src/controller.rs
+++ b/crates/data-plane-controller/src/controller.rs
@@ -1,0 +1,875 @@
+use super::{run_cmd, stack};
+use crate::repo;
+use anyhow::Context;
+use itertools::{EitherOrBoth, Itertools};
+use std::collections::VecDeque;
+
+pub struct Controller {
+    pub logs_tx: super::logs::Tx,
+    pub repo: super::repo::Repo,
+    pub secrets_provider: String,
+    pub state_backend: url::Url,
+}
+
+#[derive(Debug, serde::Deserialize, serde::Serialize)]
+#[serde(rename_all = "camelCase")]
+pub enum Message {
+    Start(models::Id),
+    Disable,
+    Enable,
+    Preview,
+    Refresh,
+    Converge,
+    // TODO(johnny): `Destroy` variant for managed tear-down.
+}
+
+#[derive(Debug, serde::Deserialize, serde::Serialize)]
+pub struct State {
+    // DataPlane which this controller manages.
+    data_plane_id: models::Id,
+    // Git branch of the dry-dock repo for this data-plane.
+    deploy_branch: String,
+    // DateTime of the last `pulumi up` for this data-plane.
+    last_pulumi_up: chrono::DateTime<chrono::Utc>,
+    // DateTime of the last `pulumi refresh` for this data-plane.
+    last_refresh: chrono::DateTime<chrono::Utc>,
+    // Token to which controller logs are directed.
+    logs_token: sqlx::types::Uuid,
+    // Pulumi configuration for this data-plane.
+    stack: stack::PulumiStack,
+    // Name of the data-plane "stack" within the Pulumi tooling.
+    stack_name: String,
+    // Status of this controller.
+    status: Status,
+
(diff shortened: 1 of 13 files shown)


jgraettinger:
## Testing

I performed extensive scenario testing against a private estuary_support/ data-plane in Vultr, including:
 - Scaling up and scale-down of gazette and reactor deployments.
 - Rolling upgrades by scaling up a new deployment, and then scaling down the old one for gazette and reactors.
 - Adding a new Etcd deployment and scaling it down again. **(note: etcd operations must be done one node at a time!)**
 - Deleting a gazette broker via the Vultr UI, and then allowing the controller to refresh and repair the data-plane.
 
From an operations perspective, `flowctl raw bearer-logs` is really nice for tailing the controller's actions in a mostly seamless way.

crates/data-plane-controller/src/repo.rs
```diff
@@ -0,0 +1,147 @@
...
+                .arg("checkout")
+                .arg("--detach")
+                .arg("--quiet")
+                .arg(format!("origin/{branch}"))
+                .current_dir(dir.path()),
+            "git-checkout",
+            logs_tx,
+            logs_token,
+        )
+        .await?;
```
mdibaiee, line 75: I think given that we are detaching the head, a sanity check that we are not working on a dirty directory may be useful, otherwise there is a possibility of conflict. Alternatively we can `git checkout --detach --force` to clean up the dirty state.

The sanity check can be done by running `git status --porcelain` and verifying the output is empty.

crates/data-plane-controller/src/controller.rs
```diff
@@ -0,0 +1,875 @@
...
+        // Now that we've completed Ansible, all deployments are current and we
+        // can prune empty deployments.
+        state
+            .stack
+            .config
+            .model
+            .deployments
+            .retain_mut(|deployment| {
+                deployment.current = deployment.desired;
+                deployment.current != 0
```
mdibaiee, line 639: nit: is this two lines for better readability? i.e. to show that now we have attained `desired` number of replicas before the retain condition

Asking since this could be simplified to:
```suggestion
                deployment.desired != 0
```

mdibaiee (approved):
LGTM, left a nit and one sanity check suggestion

crates/data-plane-controller/src/controller.rs
jgraettinger: No, it's important / semantic. If desired is `3`, we need to update `current` from (for example) `1 => 3` in cases where the deployment is retained. We're doing an interior state mutation which is persisted back to the task row.

mdibaiee: Ah I see, the `Outcome` returned updates the data_planes row after each run of the controller, makes sense now 👍🏽

PR #1741, opened 2024-10-28
validation: always overwrite inlined write schema

psFried:
Users occasionally get confusing error messages when they try to publish a collection spec where the `readSchema` contains an incorrect `flow://write-schema` definition. This updates our validation logic to always overwrite any inlined write schema definition, so we can ensure it is correct.

Normally, our response would probably be "just don't add that def", but the control plane previously used to do that inlining, and there's still some collections for which the inlined defs have not yet been removed.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1741)


jgraettinger (approved):
LGTM

PR #1745, opened 2024-10-28
Small dekaf fixes

jshearer:
**Description:**

This pulls a few fixes out of https://github.com/estuary/flow/pull/1738 in order to get them merged before we figure out the right way to do tinybird deletions

* Remove the 64mb readback behavior as it wasn't the solution to lagged consumers, and seemed to be triggering some unexpected behavior in librdkafka when it tries to read at the high-watermark and gets a messageset with _just_ a control doc in it (the ack for the transaction, which is almost always the last document available). This was causing intermittent issues with data preview UIs
* Bump `kafka-protocol` to 0.13.0 and upgrade the Fetch API to support flexible versions. This fixes the warning about `Message at offset xx might be too large to fetch, try increasing receive.message.max.bytes`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1745)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3364,8 +3364,9 @@ dependencies = [
 
 [[package]]
 name = "kafka-protocol"
-version = "0.12.0"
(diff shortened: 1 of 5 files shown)


jgraettinger (approved):
LGTM

PR #1746, opened 2024-10-29
billing: Switch payment intents to use Stripe's "automatic payment methods" feature

jshearer:
This lets you select and configure which payment methods are enabled from the Stripe web UI. The original goal was to enable ACH payments, but it seemed better to just switch us over to something that lets these be configured w/o any code changes required.

Payment input UI after this change (since card and ACH are turned on in Stripe):

Screenshot 2024-10-29 at 13 42 34

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1746)

### DIFF
--- a/supabase/functions/billing/setup_intent.ts
+++ b/supabase/functions/billing/setup_intent.ts
@@ -60,8 +60,9 @@ export async function setupIntent(
         customer: customer.id,
         description: "Store your payment details",
         usage: "off_session",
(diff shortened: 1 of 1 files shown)


jshearer:
FYI I tested this out myself and after talking w/ Dave, deployed this. So this has been live for a few days now, and we confirmed that it works as intended and people can add their bank accounts now

jgraettinger (approved):
LGTM

PR #1753, opened 2024-11-04
feat(firebolt): Update firebolt auth and sql

ptiurin:
**Description:**

New version of Firebolt has changed authentication. Now all programmatic access is done via [service accounts](https://docs.firebolt.io/Guides/managing-your-organization/service-accounts.html). Engine URL is no longer available, instead engine name should be used to connect.
This PR also fixes SQL syntax and tries to to address issues with internal keys added by Flow that don't have a defined type in the schema. 

**Workflow steps:**

No usage changes for Firebolt, only authentication.

Fields removed:
* Username
* Password
* Engine URL

Added fields:
* Client ID
* Client Secret
* Account Name
* Engine Name

Reflected changes in authentication in the relevant doc below, original in [Firebolt reference](https://docs.estuary.dev/reference/Connectors/materialization-connectors/Firebolt/). 

**Notes for reviewers:**

There's a discussion about this change in #estuary-firebolt-integration shared channel in Slack.
Related PR https://github.com/estuary/connectors/pull/2116

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1753)

### DIFF
--- a/crates/schemalate/src/firebolt/firebolt_schema_builder.rs
+++ b/crates/schemalate/src/firebolt/firebolt_schema_builder.rs
@@ -1,12 +1,15 @@
 use super::errors::*;
 use super::firebolt_queries::{CreateTable, DropTable, InsertFromTable};
 use super::firebolt_types::{Column, FireboltType, Table, TableSchema, TableType};
+use doc::shape::location::Exists::Implicit;
 use doc::shape::Shape;
 use doc::{Annotation, Pointer};
 use json::schema::{self, types};
 use proto_flow::flow::materialization_spec::Binding;
+use proto_flow::flow::Inference;
 use proto_flow::flow::MaterializationSpec;
 use serde::{Deserialize, Serialize};
+use tracing::warn;
 
 pub const FAKE_BUNDLE_URL: &str = "https://fake-bundle-schema.estuary.io";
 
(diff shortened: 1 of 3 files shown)


crates/schemalate/src/firebolt/firebolt_schema_builder.rs
```diff
@@ -166,6 +189,35 @@ pub fn build_drop_query(table: &Table) -> Result<String, Error> {
     Ok(DropTable { table }.to_string())
 }
 
+fn projection_implicit_type_to_firebolt_type(inference: &Inference) -> Option<FireboltType> {
+    let t_type = &inference.types;
+    if t_type.len() > 1 {
```
mdibaiee, line 194: I think we want to store values with multiple types as stringified JSONs in a Firebolt `text` field

crates/schemalate/src/firebolt/firebolt_schema_builder.rs
```diff
@@ -79,17 +82,37 @@ pub fn build_firebolt_schema(binding: &Binding) -> Result<TableSchema, Error> {
...
+                },
+            )?;
+        } else {
+            fb_type = projection_type_to_firebolt_type(shape).ok_or(Error::UnknownType {
+                r#type: shape.type_.to_string(),
+                field: field.clone(),
+            })?;
+
+            nullable = nullable || shape.type_.overlaps(types::NULL);
+        }
```
mdibaiee, line 107: I think this piece can be refactored so that we use a single function for converting projection type to firebolt type, it's just that in one scenario we first find the inferred type and pass it to the same function.

Is there a reason we need the two functions to have different behaviour?

mdibaiee (commented):
Thanks Petro, two comments

crates/schemalate/src/firebolt/firebolt_schema_builder.rs
ptiurin: Unfortunately we won't be able to parse those values with EXTERNAL TABLE. It's somewhat strict in this regard so if the encountered JSON type is not a quoted string then parsing into TEXT will fail.
There's a possibility to use [COPY FROM](https://docs.firebolt.io/sql_reference/commands/data-management/copy-from.html) command instead of external tables, since it's more flexible in terms of data types it can parse, but we'd need to consider any edge cases and issues that can arise from this. This would have to be a separate change.

crates/schemalate/src/firebolt/firebolt_schema_builder.rs
ptiurin: There were two reasons to create a separate function here:
* `projection_type_to_firebolt_type` operates on a `Shape` object, which contains more information than `Inference`, which only holds type names as a vector of strings.
* Since `Shape` is more complex we're able to get the inner array type, while `Inference` does not contain it (as far as I understand) and would only have "array" as a type.

mdibaiee (approved):

PR #1757, opened 2024-11-04
bound maximum serialization depth to < 128, similar to serde_json

jgraettinger:
Restrict the maximum serialization depth of documents to < 128, which guarantees they can be parsed by serde_json and other tools like `jq` which limit maximum recursion.

We want this to be universal, because we never want serde_json to be unable to parse collection documents or loaded `flow_document` columns, and its limit on maximum depth has well-founded motivations: it would be inappropriate to, for example, completely disable limits on recursion, so we follow serde_json's lead in this regard.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1757)

### DIFF
--- a/crates/doc/src/ser.rs
+++ b/crates/doc/src/ser.rs
@@ -22,6 +22,12 @@ pub struct SerPolicy {
 }
 
 impl SerPolicy {
+    /// Maximum node depth which SerPolicy will serialize.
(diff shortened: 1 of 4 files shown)


psFried (approved):
LGTM

PR #1763, opened 2024-11-07
docs: Add docs on dbt Cloud integration

danthelion:
**Description:**

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1763)

### DIFF
--- a/site/docs/guides/dbt-integration.md
+++ b/site/docs/guides/dbt-integration.md
@@ -1,0 +1,55 @@
+# dbt Cloud Integration
+
+Estuary Flow offers an integration with dbt Cloud, enabling users to trigger dbt jobs automatically when new data
+is available in a materialized view. This integration provides orchestration between the data ingestion and
+transformation layers, making real-time data workflows more efficient and automating data transformations.
+
+With the dbt Cloud Job Trigger feature in Estuary Flow, you can:
+
+- Automate transformations with dbt jobs as soon as new data is materialized, ensuring data freshness in your analytics.
+- Specify custom job behavior, like replacing or skipping jobs if a trigger is already in progress.
+- Define a custom "cause" message to add context for each triggered job.
+
+The integration can be configured when creating or editing a Materialization.
+
+## How to Configure dbt Cloud Integration
+
+Follow these steps to configure the dbt Cloud Job Trigger within an Estuary Flow materialization connector:
+
+### Required Parameters
+
+To configure the dbt Cloud Job Trigger, you’ll need the following information:
+
+- Job ID: The unique identifier for the dbt job you wish to trigger.
+- Account ID: Your dbt account identifier.
+- API Key: The dbt API key associated with your account. This allows Estuary Flow to authenticate with dbt Cloud and
+  trigger jobs.
+
+### Optional Parameters
+
+- Access URL: The dbt access URL can be found in your dbt Account Settings. Use this URL if your dbt account requires a
+  specific access endpoint. For more information, visit go.estuary.dev/dbt-cloud-trigger.
+- Cause Message: Set a custom message that will appear as the "cause" for each triggered job. This is useful for
+  tracking the context of each run, especially in complex workflows. If left empty, it defaults to "Estuary Flow."
+- Job Trigger Mode:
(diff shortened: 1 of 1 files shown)


site/docs/guides/dbt-integration.md
```diff
@@ -0,0 +1,55 @@
...
+
+Follow these steps to configure the dbt Cloud Job Trigger within an Estuary Flow materialization connector:
+
+### Required Parameters
+
+To configure the dbt Cloud Job Trigger, you’ll need the following information:
+
+- Job ID: The unique identifier for the dbt job you wish to trigger.
+- Account ID: Your dbt account identifier.
+- API Key: The dbt API key associated with your account. This allows Estuary Flow to authenticate with dbt Cloud and
```
mdibaiee, line 25: They also need an Access URL, it is mandatory, I know the connector marks it as non-required, but that's because we previously had Account Prefix and to be backward-compatible we had to keep the new one marked as non-required, but we will validate that one of the two exists every time

site/docs/guides/dbt-integration.md
```diff
@@ -0,0 +1,55 @@
...
+  trigger jobs.
+
+### Optional Parameters
+
+- Access URL: The dbt access URL can be found in your dbt Account Settings. Use this URL if your dbt account requires a
```
mdibaiee, line 30: I think it is worth adding here since a few customers have had this issue: If they can't find their Access URL in their dashboard, it is because they are old customers and have not yet migrated to the new API, in this case their Access URL is: `https://cloud.getdbt.com/`

site/docs/guides/dbt-integration.md
```diff
@@ -0,0 +1,55 @@
...
+
+### Regular Data Transformation on New Data
+
+Suppose you have a data pipeline that ingests data into a warehouse every 1 hour (configured via a Sync Frequency),
+and you want to run a dbt job on the same cadence to transform this data. Configure the `Run Interval` to `1h` to ensure
+that the dbt job is triggered automatically after every data ingestion cycle.
+
+### Job Management
+
+If you want to avoid triggering multiple overlapping dbt jobs, set Job Trigger Mode to skip. This way, if a job is
```
mdibaiee, line 51: I think it's worth mentioning this is the default behavior

site/docs/guides/dbt-integration.md
```diff
@@ -0,0 +1,55 @@
...
+    - replace: Cancels any currently running job and starts a new one.
+    - ignore: Initiates a new job regardless of any existing jobs.
+- Run Interval: Defines the interval at which the dbt job should run. This interval only triggers if new data has been
+  materialized. The default is 30m (30 minutes).
+
+## Use Cases
+
+### Regular Data Transformation on New Data
+
+Suppose you have a data pipeline that ingests data into a warehouse every 1 hour (configured via a Sync Frequency),
```
mdibaiee, line 45: The dbt cloud trigger starts the timer as soon as the first data arrives at the connector, and any subsequent timers are also started when data arrives.

If a connector has a delay of 1 hour, this is how it would look like:

Connector starts up -> runs a first dbt job trigger (this is to ensure consistency when connector restarts) -> materializes one small chunk -> starts timer to trigger dbt job in N minutes -> materializes the rest of chunks -> start 1 hour delay of connector of not backfilling -> trigger dbt job when N minutes have passed since the timer started (this includes during backfills)

So in that sense, it is best that their dbt job trigger interval is not very long. The default is 30 minutes which means 30 minutes after the first bulk of data is committed. It is not very short to avoid many jobs during backfills, but it means during non-backfill periods we will wait 30 minutes after commiting the first commit and then triggering a job. How much of a latency this creates between the final data point being materialized and the dbt job triggering depends on how long it takes for their data to be materialized to the destination

This is the current compromise we have to be able to set a minimum interval between dbt job triggers, support cases where connectors don't use Sync Interval, support use cases where data is arrival is very sparse (once a day for example)

danthelion: thanks for the detailed writeup, I tried to incorporate this as best as possible

mdibaiee (approved):

PR #1768, opened 2024-11-11
.github: deployment automation for data-plane-controller and agent-api

### LINKED ISSUE #1767: cloud run CD for data-plane-controller and agent-api

jgraettinger:
Introduce workflow_dispatch actions for building and deploying these control-plane services to Cloud Run, using workload identity federation.

Include Dockerfile infrastructure and entrypoints for placing secrets into exepected locations.

Minor tweaks to data-plane-controller, adjusting defaults to work better with Cloud Run and also to support IPv4 addresses for Ansible hosts, which is required due to our dependency on Cloud Run (which only supports IPv4, paired with Cloud NAT).

Fixes #1767 

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1768)

### DIFF
--- a/crates/agent/Dockerfile
+++ b/crates/agent/Dockerfile
@@ -1,0 +1,31 @@
+FROM ubuntu:noble
+
+# Install required packages.
+RUN apt update -y \
+    && apt install --no-install-recommends -y \
+        ca-certificates \
+        s3cmd \
+        curl \
+    && rm -rf /var/lib/apt/lists/*
+
+# Install the `sops` CLI.
+RUN curl -L -o /usr/local/bin/sops \
+      https://github.com/getsops/sops/releases/download/v3.9.1/sops-v3.9.1.linux.amd64 \
+   && chmod +x /usr/local/bin/sops
+
+# Copy in our local assets.
+COPY agent /usr/local/bin/
+COPY entrypoint.sh /usr/local/bin/
+
+ENV BIN_DIR /usr/local/bin/
+ENV RUST_LOG=info
+
+CMD ["/usr/local/bin/entrypoint.sh"]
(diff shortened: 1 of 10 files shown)


crates/agent/Dockerfile
```diff
@@ -0,0 +1,31 @@
+FROM ubuntu:noble
```
mdibaiee, line 1: Does the image here affect our costs in any noticable way? I'm wondering if we could save costs by using a slimmer image. I guess it depends on how frequently this image is going to be run by Cloud Run and how much time is taken by the image to boot up compared to other images 🤔

crates/agent/entrypoint.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+set -o errexit
+set -o pipefail
+set -o nounset
+
+# Place secrets into expected file locations.
+mkdir /root/.aws
```
mdibaiee, line 8: We are not placing anything here, is this actually necessary to create here manually?

mdibaiee (approved):
LGTM % one question and one small nit

crates/agent/Dockerfile
jgraettinger: Not especially, but this isn't a huge image (50MB iirc?). I do think it's a good idea to use the same OS base as our GitHub runner environment.

crates/agent/entrypoint.sh
jgraettinger: I left it because it's likely we will in the future, if we have agent APIs doing s3 operations (such as for updating storage mappings).

PR #1773, opened 2024-11-13
dekaf: Introduce `IDLE_SESSION_TIMEOUT_MS` to allow configuring the idle session timeout

jshearer:
Introducing this because I want to see if raising the session timeout resolves issues with static consumer group memberships.


crates/dekaf/Cargo.toml
```diff
@@ -27,6 +27,7 @@ axum = { workspace = true }
 axum-extra = { workspace = true }
 axum-server = { workspace = true }
 base64 = { workspace = true }
+humantime = { workspace = true }
```
mdibaiee, line 30: Given that we use Go's ParseDuration in all other places I know of, maybe we should use https://docs.rs/go-parse-duration/latest/go_parse_duration/

Thoughts? @williamhbaker @jgraettinger @psFried

jgraettinger: We've been using humantime for a while for this, I think we should stick with it

PR #1774, opened 2024-11-13
agent: tweak auto-discover logging, set timeout

psFried:
Tweaks the logging around auto-discovers to be more useful. ~~Also sets an explicit timeout on auto-discovers, which had been an oversight.~~

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1774)

### DIFF
--- a/crates/agent/src/controllers/capture.rs
+++ b/crates/agent/src/controllers/capture.rs
@@ -79,6 +79,7 @@ impl CaptureStatus {
                 .context("updating auto-discover")?;
             Some(next_auto_discover)
         } else {
+            self.auto_discover = None; // clear auto-discover status to avoid confusion
             None
         };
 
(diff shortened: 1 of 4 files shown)


crates/agent/src/main.rs
```diff
@@ -50,6 +50,13 @@ struct Args {
     /// Origin to allow in CORS contexts. May be specified multiple times.
     #[clap(long = "allow-origin")]
     allow_origin: Vec<String>,
+    #[clap(
```
jgraettinger, line 53: This seems to duplicate the function of the `CONNECTOR_TIMEOUT` static in proxy_connectors.
That mechanism is broader -- it covers discovers, and also validations (which also need a timeout).

Looking forward, if we're updating how timeouts work, we probably want to allow for _not_ timing out a user-initiated discover done through an future API so long as the user's POST request hasn't been cancelled, and just time out background operations... :thinking:

psFried: Yeah... I removed the separate discover timeout for now. But I kinda think that we might want the ability to specify a separate timeout per discover operation (and per publication for validations). Can take a stab at that at some point later, if/when we find we want to change from the default 5 minutes.

jgraettinger (approved):
LGTM

PR #1779, opened 2024-11-17
go/network: improve support for clients lacking ALPN

jgraettinger:
We've seen a few legacy clients that don't support HTTP/2 or ALPN negotiation. Improve the connector networking frontend to support HTTP/1.1 transports for clients that don't support HTTP/2, or don't support ALPN at all, while still preferring HTTP/2 where possible.

In particular, this lets connector networking work seamlessly with NodeJS clients that are using the old "https" NodeJS module instead of the newer "http2" module. This is less performant than HTTP/2, but metrics can indicate which tasks are using large numbers of user connections / TLS handshakes.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1779)

### DIFF
--- a/go/network/frontend.go
+++ b/go/network/frontend.go
@@ -21,7 +21,6 @@ import (
 	lru "github.com/hashicorp/golang-lru/v2"
 	pb "go.gazette.dev/core/broker/protocol"
 	pc "go.gazette.dev/core/consumer/protocol"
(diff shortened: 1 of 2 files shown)


psFried (approved):
LGTM

PR #1787, opened 2024-11-21
protocols/flow: add array inference to protocol

williamhbaker:
**Description:**

Adds "array" information to projection inference, which will include the types of items the array contains, as well as the minimum and maximum number of items it is allowed to contain.

This is only adding the (optional) field to the protocol. All users of the updated rust structs will need to be updated to use the new structs before the new field is populated.

Things that will need updated to use the updated rust structs include:
- `derive-typescript`
- Rust connectors: `source-kafka` and `source-http-ingest`
- dekaf
- flowctl, with gitpod implications
- ATF

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1787)


jgraettinger (approved):
LGTM

jgraettinger:
Noting we'll also want to update this pin: https://github.com/estuary/flow/blob/master/crates/runtime/src/container.rs#L21

(It's used for the flow-connector-init binary by flowctl if it can't otherwise locate one)

PR #1788, opened 2024-11-22
Phil/moar auto discover cleanup

psFried:
This addresses a few relatively minor issues with discovers.
Individual commit messages have more info.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1788)

### DIFF
--- a/crates/agent/src/controllers/capture.rs
+++ b/crates/agent/src/controllers/capture.rs
@@ -71,16 +71,19 @@ impl CaptureStatus {
             }
         }
 
-        let ad_next_run = if model.auto_discover.is_some() && !model.shards.disable {
-            let ad_status = self.auto_discover.get_or_insert_with(Default::default);
-            let next_auto_discover = ad_status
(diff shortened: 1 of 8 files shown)


jshearer (approved):
This LGTM from a ~1000 ft view. Love the new backoff logic for discover errors, exposing the next discover time, and logging for duplicate bindings.

One question: now that it's in the database instead of being purely computed, what happens if you manually update `next_at`? Does setting it far in the future effectively pause auto-discovers? If so, I wonder if that's something we might want to surface at least internally as a knob to turn if needed.. or maybe we explicitly _don't_ want to surface it 🤔

psFried:
> Does setting it far in the future effectively pause auto-discovers?

Yes, and setting it in the past would effectively trigger an auto-discover on the next controller run. I'm pretty hesitant about considering this a "feature", since some of this may change when we move to using the `automations` crate for controllers. But maybe?

PR #1792, opened 2024-11-25
agent: combine regular failed publications in status history

psFried:
This updates the publication history in the controller status. We already reduced consecutive touch publications in the history. This is important because it prevents touch publications, which can sometimes number in the hundreds per day, from quickly pushing out other relevant information from the publication history.  But it turns out there's another case where repeated publications can clog up the history. This happens from repeated build failures, since we retry them indefinitely.

Previously, the rule was that we'd never combine non-touch publications. The goal was to make sure that we always show distinct history entries for publications that modify the model.  This tweaks that rule so that we will combine _unsuccessful_ non-touch publications. The result is that the publications history in the status remains a lot more readable and useful when there are repeated auto-discover publication failures.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1792)

### DIFF
--- a/crates/agent/src/controllers/publication_status.rs
+++ b/crates/agent/src/controllers/publication_status.rs
@@ -128,7 +128,6 @@ fn is_touch_pub(draft: &tables::DraftCatalog) -> bool {
 
 impl PublicationInfo {
     pub fn is_success(&self) -> bool {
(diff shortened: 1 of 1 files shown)


jshearer (approved):
LGTM

PR #1797, opened 2024-11-27
agent: set updated_at when updating storage mappings

### LINKED ISSUE #1707: agent: set `updated_at` when modifying `storage_mappings`
We're not updating the `updated_at` timestamp on the `storage_mappings` table when users modify their storage mappings via directive. We should be setting it to the current time.

psFried:
Fixes #1707

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1797)


jshearer (approved):
LGTM

PR #1798, opened 2024-11-27
flowctl: add `raw spec` support for materializations

williamhbaker:
**Description:**

Adds support for the `flowctl raw spec` command to work with materializations too.

It changes the command from `flowctl raw spec --capture ...` to `flowctl raw spec --task ...`, which can work with both captures and materializations. This is is technically a breaking change, but since it is in `flowctl raw` I don't think it will have any practical impact.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1798)

### DIFF
--- a/crates/flowctl/src/raw/spec.rs
+++ b/crates/flowctl/src/raw/spec.rs
@@ -1,6 +1,7 @@
 use crate::local_specs;
 use anyhow::Context;
-use proto_flow::{capture, flow};
+use proto_flow::{capture, derive, flow, materialize};
+use tables::DraftCatalog;
 
 #[derive(Debug, clap::Args)]
 #[clap(rename_all = "kebab-case")]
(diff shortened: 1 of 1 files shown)


crates/flowctl/src/raw/spec.rs
```diff
@@ -1,17 +1,17 @@
...
 
 #[derive(Debug, clap::Args)]
 #[clap(rename_all = "kebab-case")]
 pub struct Spec {
     /// Path or URL to a Flow specification file.
     #[clap(long)]
     source: String,
-    /// Name of the capture to discover within the Flow specification file.
-    /// Capture is required if there are multiple captures in --source specifications.
+    /// Name of the task (capture or materialization) to request the spec for.
```
psFried, line 11: I kinda feel like this ought to support derivations, too, if that wouldn't be too difficult. Or can defer until we actually need it, but it might be worth noting that explicitly in the comment if we're not going to support them yet.

psFried (approved):
LGTM, unless you want to add support for derivations, too

crates/flowctl/src/raw/spec.rs
williamhbaker: Hmm yeah I didn't even realize that derivations had a spec request / response, but I guess it's in the protocol. I'll work on adding that when I get back.

williamhbaker: Also note to self: `flowctl preview` uses the flag `--name` to specify which capture/derivation/materialization should be run, so that it would probably be best to be consistent with that.

williamhbaker:
Updated the code to support derivations as well, in addition to materializations and captures. And it will now use the `--name` flag to identify the task to get the spec for. @psFried mind taking another look?

psFried (approved):
LGTM

PR #1799, opened 2024-11-27
dekaf: Log all topic names in use

jshearer:
**Description:**

We need to clear out a bunch of junk topics from MSK, and this is the best way we could think of to figure out which ones to keep

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1799)


jgraettinger (approved):
LGTM

PR #1800, opened 2024-12-02
fix(firebolt-schema-builder): better fallback to projection type resolution

ptiurin:
**Description:**

Previously a fresh setup of S3 source -> Firebolt sink would lead to an error by default due to the internal `_meta/flow_truncated` field's type not being resolved. 
In this PR I'm trying to ensure we try to resolve the type via the projection whenever we're not able to do so with the explicit mapping from the Shape object.

**Workflow steps:**

Now setting up an S3 source and Firebolt sink should work by default.

N/A

**Notes for reviewers:**

Added relevant tests to illustrate and test the issue.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1800)

### DIFF
--- a/crates/schemalate/src/firebolt/firebolt_schema_builder.rs
+++ b/crates/schemalate/src/firebolt/firebolt_schema_builder.rs
@@ -1,7 +1,6 @@
 use super::errors::*;
 use super::firebolt_queries::{CreateTable, DropTable, InsertFromTable};
 use super::firebolt_types::{Column, FireboltType, Table, TableSchema, TableType};
-use doc::shape::location::Exists::Implicit;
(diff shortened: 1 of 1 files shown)


ptiurin:
> FYI, inner array types are being included in the Inference message [by this PR](https://github.com/estuary/flow/pull/1787), which AFAIK will remove the need for this connector to use `schemalate` / extract directly from the JSON schema. This is a fine stop-gap.

Thanks for the heads-up! It would be great to have this connector more in line with the others! Though I'd imagine it would take some effort to refactor out `schemalate` so I think merging this PR in the meantime would let us proceed faster.

PR #1802, opened 2024-12-02
Phil/periodic publications

### LINKED ISSUE #1671: agent: periodically re-publish active tasks
Controllers should periodically re-publish all active specs. We'd like to do this for several reasons:

- It limits the staleness of built specs, ensuring that all built specs in active use were produced by a relatively recent version of the agent.
- It allows us to limit the size of the builds storage bucket by setting a lifecycle policy

The actual interval at which we re-publish should probably be configurable via an application argument. If nothing else, it makes it easier to test. The default interval is up for debate. I'll start the bidding at 30 days.

psFried:
**Description:**

Resolves #1671 

Adds a periodic publication of all live specs as part of controllers.
The much larger commit here is to refactor controllers. This was partly in order to simplify adding the periodic publication logic, and partly because it's something that needed done anyway. See the individual commit messages for more.

**Workflow steps:** No user-visible changes here

**Notes for reviewers:**

I started the refactor looking at the pattern established in the `data-plane-controller` crate, but I ended up going with something a bit different. There's a few differences between the two controllers that motivated this. Compared to data plane controllers, live specs controllers are run a lot more frequently, and commonly their various sub-statuses will just do some checks and not actually need to publish. So the pattern I used here seemed like a good compromise between the current factoring and that of the data plane controller. We go down the list of sub-statuses and as soon as one of them actually publishes we return and schedule another immediate run.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1802)


jgraettinger (approved):
LGTM

PR #1805, opened 2024-12-02
agent: read connector timeout from env

psFried:
Allows setting the agent's connector timeout via the `FLOW_CONNECTOR_TIMEOUT` env variable, so we can change it without needing to re-compile the whole application.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1805)


jshearer (approved):
LGTM! I initially thought that this should be an agent arg, but it looks like that has already been done and then backed out in favor of potential future per-request timeouts.. so this is a good middle-ground that's easy to change without a whole refactor to re/un-plumb the arg through again.

PR #1806, opened 2024-12-03
dekaf: Fix dropped topic name translation

jshearer:
This fixes a bug introduced in #1799 that dropped the topic name translation in `offset_fetch`

Confirmed that a consumer can join a group and fetch offsets correctly now.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1806)


psFried (approved):
LGTM

PR #1812, opened 2024-12-06
Add docs for default schema names and delta updates

danthelion:
Docs for feature that was recently shipped. Discussion here: https://estuaryworkspace.slack.com/archives/C02G25WJCNS/p1733421953163009

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1812)

### DIFF
--- a/site/docs/concepts/materialization.md
+++ b/site/docs/concepts/materialization.md
@@ -319,13 +319,13 @@ standard updates only unless otherwise noted in its documentation.
 
 ### Delta updates for every binding in a Materialization
 
-When configuring a materialization, you can set delta updates as the materialziation mechanism
+When configuring a materialization, you can set delta updates as the materialization mechanism
 for every binding. Afterwards, you are able to modify this setting individually for each binding
 for further customization.
 
 :::note
-This functionality is not supported for all Materialization connectors. Consult the individual connector
-documentation for details.
+This functionality is only supported for Materialization connectors that have the `x-delta-updates`
+field implemented. Consult the individual connector documentation for details.
 :::
 
 ## Using `sourceCapture` to synchronize capture and materialization bindings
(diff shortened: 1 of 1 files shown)


site/docs/concepts/materialization.md
```diff
@@ -317,6 +317,17 @@ A given endpoint may support standard updates, delta updates, or both.
 This depends on the [materialization connector](../reference/Connectors/materialization-connectors/README.md). Expect that a connector will use
 standard updates only unless otherwise noted in its documentation.
 
+### Delta updates for every binding in a Materialization
+
+When configuring a materialization, you can set delta updates as the materialziation mechanism
```
aeluce, line 322: Typo on "materialization" here (materialziation).

site/docs/concepts/materialization.md
```diff
@@ -317,6 +317,17 @@ A given endpoint may support standard updates, delta updates, or both.
...
+for every binding. Afterwards, you are able to modify this setting individually for each binding
+for further customization.
+
+:::note
+This functionality is not supported for all Materialization connectors. Consult the individual connector
+documentation for details.
```
aeluce, line 328: Can we be more specific in this note and in the schema name note on what to look for? Something like: "This functionality is only supported for materialization connectors with [xyz]. Consult the individual connector documentation for details."

site/docs/concepts/materialization.md
danthelion: good catch, thanks!

site/docs/concepts/materialization.md
danthelion: Clarified a bit, let me know if it makes sense

aeluce (approved):
Thanks for updating; LGTM!

PR #1813, opened 2024-12-06
dekaf: Remove unneccesary timeouts

jshearer:
**Description:**

I had added these timeouts when we were dealing with unexplained hangs. We have now narrowed the problem down to long-running data-plane reads, so all other timeouts not related to long-running reads should be removed.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1813)


jgraettinger (approved):
LGTM

PR #1814, opened 2024-12-06
close out a number of TODOs which have been resolved

jgraettinger:
* deploy-agent-api, deploy-data-plane-controller, and derive-typescript are merged.

* `name:prefix` support has been live for some time now.

* Continue to clear the `internal` field in flow-connector-init. There's no use case for sending it into connectors.

* No current plan to implement a Destroy data-plane-controller message.

* Current UI is generating tokens via /authorize/user/task

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1814)

### DIFF
--- a/.github/workflows/deploy-agent-api.yaml
+++ b/.github/workflows/deploy-agent-api.yaml
@@ -2,9 +2,6 @@ name: Deploy agent-api
 
 on:
   workflow_dispatch: {}
-  # TODO(johnny): Remove after merging.
-  push:
-    branches: [johnny/dpc-cd]
 
 env:
   CARGO_INCREMENTAL: 0 # Faster from-scratch builds.
--- a/.github/workflows/deploy-data-plane-controller.yaml
+++ b/.github/workflows/deploy-data-plane-controller.yaml
@@ -2,9 +2,6 @@ name: Deploy data-plane-controller
 
 on:
   workflow_dispatch: {}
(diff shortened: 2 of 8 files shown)


crates/assemble/src/lib.rs
```diff
@@ -384,14 +386,10 @@ pub fn journal_selector(
 ) -> broker::LabelSelector {
     let mut include = labels::build_set([
         (labels::COLLECTION, collection.name.as_ref()),
-        // TODO(johnny): Enable this as soon as the label prefix change has propagated
-        // to all connectors, such as derive-typescript.
-        /*
         (
             "name:prefix",
```
psFried, line 390: What's the purpose of this? Seems like this new selector is essentially equivalent to the old one. And aren't they implicitly `AND`ed together anyway? What am I missing?

psFried (approved):
LGTM

crates/assemble/src/lib.rs
jgraettinger: There's a performance optimization in gazette where it uses name:prefix to efficiently seek to the range of possible matches. Same with `id:prefix`. We use both `estuary.dev/collection` and `name:prefix` consistently with this change. I'd held this particular instance back because connectors need to be able to parse it and needed the protocol update.

PR #1825, opened 2024-12-12
Phil/agent jwt secret

psFried:
**Description:**

Updates the agent to read `CONTROL_PLANE_JWT_SECRET` from the env instead of querying it from the database. Supabase recently did a [rug pull](https://github.com/orgs/supabase/discussions/30606) on `app.settings.jwt_secret`, and agents were relying on the presence of that as part of their startup routine. This updates agent to read the jwt secret from an env variable instead of querying it from the database.

Also updates the `internal.access_token_jwt_secret` sql function to read the jwt secret from vault instead of relying on `app.settings.jwt_secret`.

I also hacked in a disablement of the periodic publications, since the fix to make those apply to tasks only hasn't merged yet.

The migration has already been run, and the agents have already been deployed.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1825)

### DIFF
--- a/Tiltfile
+++ b/Tiltfile
@@ -164,6 +164,7 @@ local_resource(
         "DATABASE_URL": DATABASE_URL,
         "RUST_LOG": "info",
         "SSL_CERT_FILE": CA_CERT_PATH,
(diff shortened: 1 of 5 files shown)


jshearer (approved):
LGTM

PR #1830, opened 2024-12-13
agent: limit periodic publications to tasks only

psFried:
**Description:**

Updates the periodic publication controller so that it applies only to tasks. Previously, we deployed periodic publications and then reverted it because the number of collection publications was overwhelming. The original reason for publishing collections was only to ensure that built specs were all reasonably up to date, but we've decided that no longer seems worth the effort of doing so many publications. So this excludes collections from the periodic publications. The main objective of letting us expire old build databases from the cloud storage bucket is still met by periodically publishing captures, derivations, and materializations. I couldn't think of a reason to exclude catalog tests from periodic publication, so I just left it.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1830)

### DIFF
--- a/crates/agent/src/controllers/collection.rs
+++ b/crates/agent/src/controllers/collection.rs
@@ -75,6 +75,7 @@ impl CollectionStatus {
         } else {
             None
         };
+        // This function will return None unless this collection is an enabled derivation.
         let periodic_next = periodic::next_periodic_publish(state);
         Ok(NextRun::earliest([inferred_schema_next, periodic_next]))
     }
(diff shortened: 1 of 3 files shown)


crates/agent/src/controllers/collection.rs
```diff
@@ -75,7 +75,11 @@ impl CollectionStatus {
         } else {
             None
         };
-        let periodic_next = periodic::next_periodic_publish(state);
+        let periodic_next = if model.derive.is_some() {
```
jshearer, line 78: Should this not also check whether the derivation is active, like you do down [here](https://github.com/estuary/flow/pull/1830/files#diff-941bb1e08a9eeb14ecb2d4063a56b2da43671068a6d0681794979aa6403e9b03R80)? Or is that not applicable here?

jshearer (approved):
LGTM % one question about running controllers on derived collections

crates/agent/src/controllers/collection.rs
psFried: Now that you mention it, I think I can just remove this whole conditional. `next_periodic_publish` already returns `None` if `!is_enabled_task(state)`

PR #1831, opened 2024-12-16
flowctl: add raw list-shards subcommand

### LINKED ISSUE #1710: flowctl: add subcommands for listing shards

psFried:
Resolves #1710
Adds `flowctl raw list-shards`, which fetches shard information for a given task. It's not yet clear whether this will have relevance to most users, since we plan to integrate shard status into the higher-level "status". But it's pretty inconvenient to use `flowctl-go` with legacy data planes, so this is just meant to give us a debugging tool that we can use in the meantime.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1831)

### DIFF
--- a/crates/flowctl/src/raw/mod.rs
+++ b/crates/flowctl/src/raw/mod.rs
@@ -14,6 +14,7 @@ use tables::CatalogResolver;
 mod discover;
 mod materialize_fixture;
 mod oauth;
+mod shards;
 mod spec;
 
 #[derive(Debug, clap::Args)]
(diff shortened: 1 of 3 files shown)


mdibaiee (approved):
LGTM, thanks Phil!

PR #1832, opened 2024-12-16
dekaf: Set `_meta/is_deleted` when `deletions: CDC` is set, even if collection schema has `additionalProperties: true`

jshearer:
**Description:**

Had to switch from `Shape::widen()` to manually adding the location as if the collection schema has `additionalProperties: true` (or unset), `Shape::widen()` will be a no-op.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1832)

### DIFF
--- a/crates/dekaf/src/topology.rs
+++ b/crates/dekaf/src/topology.rs
@@ -1,9 +1,8 @@
 use crate::connector::DeletionMode;
 use anyhow::Context;
-use futures::{StreamExt, TryFutureExt, TryStreamExt};
+use futures::{StreamExt, TryStreamExt};
 use gazette::{broker, journal, uuid};
 use proto_flow::flow;
-use std::time::Duration;
 
 /// Fetch the names of all collections which the current user may read.
 /// Each is mapped into a kafka topic.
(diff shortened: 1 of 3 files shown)


crates/dekaf/src/topology.rs
```diff
@@ -94,8 +93,45 @@ impl Collection {
         let validator = doc::Validator::new(json_schema)?;
         let mut shape = doc::Shape::infer(&validator.schemas()[0], validator.schema_index());
 
+        // In order for the `_meta/is_deleted` field to get serialized correctly into avro, we need an explicit
+        // field definition for it in the schema. Since some schemas can have `additionalProperties: true`
+        // (or unset, which defaults to true), trying to use e.g Shape::widen() or Shape::union() will
```
jgraettinger, line 98: nit: i know this comment is relevant to why you were just in here, but it's not super relevant to the reader. Why would they presume `Shape::widen()` is preferred in the first place?

jgraettinger (approved):
LGTM

crates/dekaf/src/topology.rs
jshearer: Heh, I keep thinking this code is smelly so I keep trying to justify its existence vs a much simpler call to e.g `widen()` or `union()`. I'll get rid of the comment

PR #1840, opened 2024-12-18
Dekaf materialization endpoint support

jshearer:
**Description:**

This adds support for the server side of Dekaf's support for materialization endpoints. At a high level, Dekaf is just another way to get data out of Flow. We already have a well fleshed out concept for these things: materializations. So back in https://github.com/estuary/flow/pull/1665 we introduced support for a new materialization "endpoint type": `dekaf`. This lives alongside `local` and `connector` as the third kind of materialization, and is configured like so:

```yaml
    endpoint:
      dekaf:
        variant: some-dekaf-variant
        config:
          token: "foo"
          ...other configuration for the behavior of your Dekaf task... 
``` 

The second part of this work is for Dekaf the server to support this mode of operation. Briefly, it needs to:
* Support authentication and authorization using the control plane. This means that conceptually, Dekaf is authenticating you based on the task name and token specified in the endpoint config, and authorizing you based on the access granted to that task
  * Specifically, this means that you now use the **task name** for your username, and the token specified in the endpoint config when connecting to Dekaf.
* Act more like a regular materialization. This means...
  * Support for field selection. Materializations let you specify which field(s) you want included in your destination using **field selection**. This fundamentally looks like a document transformation, where some fields may be removed and some fields may be **projected** from their original home to a new location in the output. 
  * Support for task logs. Just like captures, materialization log output is captured and presented to users for status monitoring and debugging. As Dekaf is multi-tenant at its core, presenting these logs requires identifying which log messages are associated with which task, and then capturing and writing them to the corresponding logs journal.
  * **[Fast-Follow]** Support for task stats. In order to monitor the status of your running tasks, as well as aggregate usage information for billing, all tasks need to periodically emit information about how much work they've been doing. While Dekaf was in beta, all usage was free so this was less of a priority, but now that there's a task to associate stats _with_, implementing stats will likely be one of the last things to do before going GA.

I still have a couple of things on my list before this is fully wrapped up:
- [x] Implement emitting stats
- [x] Implement routing `SessionAuthentication::Task` sessions to a new/different backing store for migration purposes 
- [x] Support journal suspension
- [ ] Implement CI task for Dekaf integration tests
- [ ] **Later**: out how to make shard statuses show green in the UI for Dekaf tasks

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1840)


crates/dekaf/src/log_journal.rs
```diff
@@ -0,0 +1,344 @@
...
+        client: &journal::Client,
+    ) -> anyhow::Result<()> {
+        let resp = client.append(
+            gazette::broker::AppendRequest {
+                journal: journal.to_owned(),
+                ..Default::default()
+            },
+            || {
+                futures::stream::once({
+                    let value = log_data.clone();
```
jshearer, line 105: Pretty sure this is a cheap clone since `log_data` is `Bytes` and never modified. It needs to be cloned because this is `Fn`, not `FnOnce` since it needs to be callable multiple times for retry purposes

crates/dekaf/src/log_journal.rs
```diff
@@ -0,0 +1,344 @@
...
+            }
+        }
+    }
+
+    fn serialize_log(&self, log: ops::Log) -> Vec<u8> {
+        let uuid = gazette::uuid::build(
+            self.producer,
+            gazette::uuid::Clock::from_time(std::time::SystemTime::now()),
+            uuid::Flags(
+                (message_flags::MASK | message_flags::OUTSIDE_TXN)
```
jshearer, line 140: @jgraettinger do these looks right to you? IIRC I was seeing some suspicious `Unexpected UUID flags` errors in my Gazette logs, but I'm not able to find/reproduce that

crates/dekaf/src/log_journal.rs
```diff
@@ -0,0 +1,344 @@
...
+
+#[derive(Clone)]
+struct SessionLogger {
+    tx: tokio::sync::mpsc::Sender<LoggingMessage>,
+    _handle: Arc<JoinHandle<()>>,
+}
+
+impl SessionLogger {
+    fn new(app: Arc<App>, producer: Producer) -> Self {
+        let (log_tx, log_rx) = tokio::sync::mpsc::channel(1000);
```
jshearer, line 195: This is probably not ideal.. I'd like to refactor so that setting the task name is a oneshot and this is more like a stack that only contains the N newest items, rather than erroring once this channel gets full.

crates/avro/src/lib.rs
```diff
@@ -43,14 +45,8 @@ pub enum Error {
...
 /// Map a [`doc::Shape`] and key pointers into its equivalent AVRO schema.
-pub fn shape_to_avro(
-    shape: doc::Shape,
-    key: &[doc::Pointer],
-) -> (apache_avro::Schema, apache_avro::Schema) {
-    (
-        schema::key_to_avro(key, shape.clone()),
-        schema::shape_to_avro(json::Location::Root, shape, true),
-    )
+pub fn shape_to_avro(shape: doc::Shape) -> apache_avro::Schema {
```
jshearer, line 48: See [here](https://github.com/estuary/flow/pull/1840/files#diff-8cb17d92f697c0e1519db60f893d21d41cc59c75e2d9872ff522072968357b22R265-R266) for why I made this change. I'm happy to try and find a different way if you feel like you'd rather keep the API of the `avro` crate the same, this was just the most direct path, and Dekaf is the only consumer of this code so far so it was pretty limited in impact

crates/dekaf/src/topology.rs
```diff
@@ -1,30 +1,104 @@
...
+    pub fn get_binding_for_topic(
+        &self,
+        topic_name: &str,
+    ) -> anyhow::Result<
+        Option<(
+            proto_flow::flow::materialization_spec::Binding,
+            DekafResourceConfig,
+        )>,
+    > {
+        Ok(self
```
jshearer, line 62: Either make this a hashmap or sort the vec and use binary search -- we've seen >3k bindings, this could potentially get slow

crates/dekaf/src/topology.rs
```diff
@@ -59,42 +135,99 @@ pub struct PartitionOffset {
 impl Collection {
     /// Build a Collection by fetching its spec, a authenticated data-plane access token, and its partitions.
```
jshearer, line 136: **an** authenticated data-plane access token

crates/dekaf/src/topology.rs
```diff
@@ -323,40 +469,239 @@ impl Collection {
...
+        return shape.clone();
+    }
+
+    tracing::debug!(?ptr, "Building shape");
+    let mut current_shape = doc::Shape::nothing();
+    let mut current = &mut current_shape;
+
+    // For each component in the pointer path except the last one,
+    // create the object structure
+    for token in ptr.iter().take(ptr.0.len() - 1) {
```
jshearer, line 615: This is hard to follow even just a few days after writing it, refactor to be a fold over `ptr.iter()` with `doc::Shape::nothing()` as the initial state

jshearer:
Alright, I could probably do another few rounds of self-review here, but I think it's at a good enough spot to mark this as ready for review. Let me know what y'all think!

jgraettinger (commented):
Sending comments from a partial review. I haven't finished.

jgraettinger (commented):
Okay, I've finished a pass.

As a meta comment, for big features like this please either:
a) Open multiple smaller PRs which can be reviewed independently, or

b) Maintain ever-green commits which are re-written for the reader: they should tell more bite-size stories about changes being made, stand on their own, and be feasible to review (or re-review) independently. This is the workflow @psFried and I use. In essence each commit is a mini-PR -- it requires that you regularly rebase/squash/fixup as you develop.

crates/dekaf/src/connector.rs
```diff
@@ -116,34 +118,53 @@ pub async fn unary_materialize(
...
+                        ))?;
+
+                        let collection = binding.collection.expect("collection must exist");
+
+                        let mut constraints = BTreeMap::new();
+
+                        for proj in collection.projections {
+                            constraints.insert(
+                                proj.field,
+                                validated::Constraint {
```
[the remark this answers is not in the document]
jshearer: Thanks for the link, I took a look and followed that example.

crates/dekaf/src/log_journal.rs
```diff
@@ -0,0 +1,654 @@
...
+
+        let mut attempts = 0;
+
+        loop {
+            match resp.try_next().await {
+                Err(err) if err.is_transient() && attempts < 3 => {
+                    attempts += 1;
+                }
+                Err(err) => {
+                    if err.is_transient() {
```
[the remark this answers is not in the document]
jshearer: 95% sure this is resolved by https://github.com/estuary/flow/pull/1897 -- anything else I should need to do here?

crates/dekaf/src/log_journal.rs
```diff
@@ -0,0 +1,654 @@
...
+            self.producer,
+            gazette::uuid::Clock::from_time(std::time::SystemTime::now()),
+            uuid::Flags(
+                (message_flags::MASK | message_flags::OUTSIDE_TXN)
+                    .try_into()
+                    .unwrap(),
+            ),
+        );
+
+        let mut val = serde_json::to_value(log).expect("Log always serializes");
```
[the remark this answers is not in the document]
jshearer: Great point, didn't realize `ops::Log` had a meta field already. Populating that works great :+1:

crates/dekaf/src/main.rs
```diff
@@ -100,45 +143,64 @@ struct TlsArgs {
...
+    // There's probably a neat bit-banging way to do this with i64 and masks, but I'm just not that fancy.
+    let mut producer_id = rand::thread_rng().gen::<[u8; 6]>();
+    producer_id[0] |= 0x01;
+
+    let registry = tracing_subscriber::registry()
+        .with(tracing_record_hierarchical::HierarchicalRecord::default())
+        // We attach `ops::tracing::Layer` so we can use the `Log` structs it attaches to spans
+        .with(ops::tracing::Layer::new(|_| {}, std::time::SystemTime::now))
+        .with(
+            dekaf::SessionSubscriberLayer::new(
```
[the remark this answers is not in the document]
jshearer: Am now! See https://github.com/estuary/flow/pull/1840/files#diff-422fb856376c95a1ba5c8fd23d9012c4f901c5b0cef1041c2592b4616875b03dR18

crates/dekaf/src/main.rs
```diff
@@ -257,6 +326,13 @@ async fn main() -> anyhow::Result<()> {
                             addr,
                             cli.idle_session_timeout,
                             stop.clone()
+                        ).instrument(
```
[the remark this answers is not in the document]
jshearer: Yup, ended up refactoring this using [`tokio::LocalKey`](https://docs.rs/tokio/latest/tokio/task/struct.LocalKey.html)

crates/dekaf/src/read.rs
```diff
@@ -118,7 +126,7 @@ impl Read {
             Compression, Record, RecordBatchEncoder, RecordEncodeOptions,
         };
 
-        let mut alloc = bumpalo::Bump::new();
+        let alloc = bumpalo::Bump::new();
```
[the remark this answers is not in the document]
jshearer: I'll do one better, now that reading is refactored to serialize iteratively, I don't even need to create HeapNodes at all, so `alloc` here can go away entirely

crates/dekaf/src/topology.rs
```diff
@@ -1,30 +1,103 @@
...
+            proto_flow::flow::materialization_spec::Binding,
+            DekafResourceConfig,
+        )>,
+    > {
+        Ok(self
+            .built_spec
+            .bindings
+            .iter()
+            .map(|b| {
+                serde_json::from_str::<crate::connector::DekafResourceConfig>(
```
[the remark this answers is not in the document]
jshearer: Updated to parse bindings into a map upfront and then just query out of that map later on

jshearer:
Alright, I did another pass taking into account your feedback and this is ready for you to take another look at. I refactored the field selection stuff to account for the fact that avro documents can be encoded sequentially. This has a really nice consequence of removing the need to build up a `HeapNode` of the data to serialize, which I believe should improve memory usage a good bit.

I had already been keeping evergreen commits and working on them by fixup/rebasing, but I cleaned them up even more to simplify the story. 

> Recently, it became possible to build catalog specs in Rust as part of tests, which would let you exercise your logic with "real" tasks.
I'm taking a look at this as I'd like to introduce more narrow-focused unit testing especially for things like field selection

jshearer:
Okay, I think this is ready for another look. I changed around the field selection to be closer to your suggested pattern of building up a list of extractors and their associated avro schemas. The realization that avro documents can be encoded sequentially allowed me to avoid building up the full `HeapNode` before encoding, which should reduce memory usage a good bit. 

FWIW though, I ended up generating a `(avro::Schema, Vec<doc::Extractor>)` in `build_field_extractors()`, mainly because I need the full schema in `Collection` in order to save it to the database and get a content-addressed ID. I then extract out the field schemas and zip them with the extractors in `Read`. I suppose I could instead avoid building the full schema in `build_field_extractors` and just build it when dealing with schema registry stuff... this feels like six of one, half dozen of the other IMO.

>Recently, it became possible to build catalog specs in Rust as part of tests, which would let you exercise your logic with "real" tasks.

I'll take a look at this and come up with a good way to unit test `build_field_extractors()`. As it stands now, are you still wanting more separation between "glue" and "business logic"? If so, could you try to articulate where to draw that imaginary line? Usually when I think about this sort of thing I've been thinking about what logic can be written in a more reusable/more broadly applicable way (i.e what belongs in an "externally" exposed interface vs what should remain internal, etc), but it sounds like you're talking more about distinguishing between things that can fail vs things that can't? 

For example, I extracted the log forwarding stuff into its own module which roughly exposes an interface for how to use it (with the rough edges getting refined over time, as the things it needs to do becomes more clear), so that's all you should have to think about when you want to forward logs -- this should probably eventually be moved to its own crate, especially if anything else wants to use it as well later on. But this doesn't sound to me like the same kind of separation you're asking for.

As far as your meta-comments, like I mentioned over Slack, I [certainly](https://github.com/estuary/flow/pull/1865) [have](https://github.com/estuary/flow/pull/1858) [been](https://github.com/estuary/flow/pull/1854) [opening](https://github.com/estuary/flow/pull/1848) [PRs along the way](https://github.com/estuary/flow/pull/1841), though granted I could have been doing even more of that. 

And as far as evergreen commits, that is indeed always my goal, and with this round of changes I've done some rewriting and condensing of the commits to make them more self-contained. I do have a question about that though: when you're at the point of pushing your work up to a PR, do you actually just squash everything into one commit and then chunk it out from there into separate "mini-PRs"? Because I've been trying to work with the evergreen commits the whole time, and if I need to make more-than-trivial changes, I find myself spending more time than I'd like on figuring out exactly which chunks of changes to send to which commit. Thoughts?

crates/dekaf/src/read.rs
```diff
@@ -269,27 +273,15 @@ impl Read {
...
-                    }
+                    self.extract_and_encode(root.get(), &mut tmp)?;
 
                     record_bytes += tmp.len();
                     buf.extend_from_slice(&tmp);
                     tmp.clear();
                     Some(buf.split().freeze())
                 };
 
+            input_bytes += next_offset - self.offset;
```
jshearer, line 284: This will count acks... should we skip those?

crates/dekaf/src/utils.rs
```diff
@@ -0,0 +1,154 @@
...
+
+#[derive(Debug, Clone)]
+pub enum CustomizableExtractor {
+    Extractor(doc::Extractor),
+    IsDeleted,
+}
+
+// This lets us add our own "virtual" fields to Dekaf without having to add them to
+// doc::Extractor and all of the other platform machinery.
+impl CustomizableExtractor {
```
jgraettinger, line 20: IMO it's very reasonable to add this in extractors.rs, but :man_shrugging:

jgraettinger (approved):
LGTM! :ship: 

I know you're still working on amortizing multiple log appends, but all of the big picture stuff looks solid, and I think this can land as-is.

Going forward (_not_ this PR), a refactor I'd urge you to look into is separating the pre-authorization session loop from the post-authorization loop. A refactored pre-auth session loop would _only_ expect authorization-related messages. It would build up all of the bits of session state & context, and then the moment the session authorization is complete, it would tail-call into the post-authorization handler.

The post-auth handler then gets the benefit of _assuming_ the session context is fully established, that the tokio::task_local! is set and Some, and it can also use tracing::instrument for contextual fields of the session (instead of using tracing-record-hierarchical to mutate the parent Span).

crates/dekaf/src/log_appender.rs
```diff
@@ -0,0 +1,804 @@
...
+                    let wait_ms = rand::thread_rng().gen_range(400..5_000);
+
+                    tracing::warn!(
+                        ?attempt,
+                        ?inner,
+                        ?wait_ms,
+                        "Got recoverable error trying to write logs, retrying"
+                    );
+
+                    tokio::time::sleep(Duration::from_millis(wait_ms)).await;
```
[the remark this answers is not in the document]
jshearer: removed 👍

PR #1841, opened 2024-12-18
dekaf: Fix endpoint config parsing

jshearer:
This got left out of #1793 and is needed for Dekaf materialization endpoints

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1841)

### DIFF
--- a/crates/dekaf/src/connector.rs
+++ b/crates/dekaf/src/connector.rs
@@ -1,5 +1,5 @@
 use anyhow::{bail, Context};
-use proto_flow::materialize;
+use proto_flow::{flow::materialization_spec, materialize};
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 use std::collections::BTreeMap;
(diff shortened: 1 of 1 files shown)


crates/dekaf/src/connector.rs
```diff
@@ -70,8 +71,22 @@ pub async fn unary_materialize(
 ) -> anyhow::Result<materialize::Response> {
     use proto_flow::materialize::response::validated;
     if let Some(mut validate) = request.validate {
-        serde_json::de::from_str::<DekafConfig>(&validate.config_json)
-            .context("validating endpoint config")?;
+        match materialization_spec::ConnectorType::try_from(validate.connector_type)? {
```
psFried, line 74: Just as an fyi, you could also use

```rust
let ConnectorType::Dekaf = ConnectorType::try_from(validate.connector_type)? else {
  bail!("...");
};
```

psFried (approved):
LGTM

crates/dekaf/src/connector.rs
jshearer: But then I wouldn't have the parsed `proto_flow::flow::materialization_spec::ConnectorType`, which can give me a human-readable`as_str_name()` instead of just `proto_flow::materialize::request::Validate:: connector_type: i32`, which is much less useful in an error

PR #1848, opened 2024-12-30
dekaf: Generate predictable shard IDs for dekaf materialization endpoints

jshearer:
For UX reasons, we only want to use the task name (`myTenant/materializations/dekaf`) when authenticating to Dekaf, as opposed to the full shard template ID (`/myTenant/materializations/dekaf/0fa42cafc58003fa`). 

OTOH, the agent API's `/authorize/task` endpoint expects to be passed the full shard template ID, so since Dekaf materializations never actually create any shards, we decided to use a fixed value for the pub ID component, allowing Dekaf to craft the correct shard template ID given only the task name.

Cherry-picked out of https://github.com/estuary/flow/pull/1840, as it's required to test Dekaf materialization functionality e2e

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1848)

### DIFF
--- a/.github/workflows/dekaf.yaml
+++ b/.github/workflows/dekaf.yaml
@@ -6,6 +6,16 @@ on:
     paths:
       - "crates/dekaf/**"
       - "crates/proto-**"
+      - "crates/allocator"
+      - "crates/avro"
+      - "crates/doc"
+      - "crates/flow-client"
+      - "crates/gazette"
+      - "crates/json"
+      - "crates/labels"
+      - "crates/models"
+      - "crates/ops"
+      - "crates/simd-doc"
       - "Cargo.lock"
   pull_request:
     branches: [master]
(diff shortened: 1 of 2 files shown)


.github/workflows/dekaf.yaml
```diff
@@ -6,6 +6,16 @@ on:
     paths:
       - "crates/dekaf/**"
       - "crates/proto-**"
+      - "crates/allocator"
```
jshearer, line 9: I think I'm coming around to the idea that Dekaf should just be another binary in the flow image, as opposed to its own image. But for the moment, this just has it rebuild when any of its local crate deps change

crates/validation/src/materialization.rs
```diff
@@ -327,7 +327,22 @@ async fn walk_materialization(
     {
         shard_template.id.clone()
     } else {
-        assemble::shard_id_prefix(pub_id, materialization, labels::TASK_TYPE_MATERIALIZATION)
+        assemble::shard_id_prefix(
+            match endpoint {
```
psFried, line 331: optional nit: feel free to disagree, but I found this inlined match expression to be a little hard on the eyes, and would I think prefer a variable.

psFried (approved):
LGTM

PR #1854, opened 2025-01-03
Dekaf runtime changes

jshearer:
**Description:**

Rolls up a couple more changes to the agent and runtime for Dekaf:

* Detect tasks intended for Dekaf inside `Runtime::unary_materialize()` and route them to `dekaf::connector::unary_materialize()`. Doing it here instead of in `build` allows more of Flow to interact with Dekaf tasks correctly:
  * The field selection UI uses a `dry_run` publication (i.e a `validate` call) to fetch which fields to show. This now intercepts that call and correctly passes it to Dekaf because we set `connector_type` based on the endpoint config in `walk_materialization()`: https://github.com/estuary/flow/blob/44dc0db762feaf1381c417d1893dc06ff0728380/crates/validation/src/materialization.rs#L126-L128
  * The `connector_tags` agent task calls a connector's `Spec` RPC, and we need to get Dekaf's endpoint/resource config schemas _somehow_ so I figured rather than writing a thing to do it manually, all we need to do is send `connector_type: ConnectorType::Dekaf` when we detect a task with `image_name` starting in `ghcr.io/estuary/dekaf-`.
* I also had to skip image pulling in `connector_tags` when processing a Dekaf row as obviously there's no image to pull.
* Move unseal out of runtime into its own crate. We now need `runtime` to depend on `dekaf`, and they both need `unseal` so it needed to be moved somewhere, and there didn't seem to be anywhere good to put it other than its own crate.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1854)


crates/agent/src/connector_tags.rs
```diff
@@ -117,15 +123,20 @@ impl TagHandler {
             }
         }
 
-        let proto_type = match runtime::flow_runtime_protocol(&image_composed).await {
-            Ok(ct) => ct,
-            Err(err) => {
-                tracing::warn!(image = %image_composed, error = %err, "failed to determine connector protocol");
-                return Ok((row.tag_id, JobStatus::SpecFailed));
+        let proto_type = if row.image_name.starts_with(DEKAF_IMAGE_NAME_PREFIX) {
```
psFried, line 126: I kinda feel like this conditional should be in `runtime::flow_runtime_protocol`. Is there a reason to prefer it to be in `agent`?

psFried (approved):
LGTM

crates/agent/src/connector_tags.rs
jshearer: Yeah, you're right. I read `runtime::flow_runtime_protocol` and `parse_image_inspection` briefly and saw that `docker inspect` call, but it's trivial to check the image first. :+1:

PR #1856, opened 2025-01-07
gazette: Implement `journal::append()`

jshearer:
**Description:**

This implements low-level appends to Gazette. It takes a function that generates a stream of `Bytes`, and attempts to drive an append RPC. If the RPC errors, you can retry by re-polling after recieving an Error.

Pulled out of my Dekaf work in https://github.com/estuary/flow/pull/1840 for better reviewability.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1856)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -186,6 +186,7 @@ tokio = { version = "1", features = [
     "time",
 ] }
 tokio-util = { version = "0.7", features = ["io", "compat", "rt"] }
+tokio-stream = { version = "0.1.17" }
 tonic = { version = "0.12", features = ["tls", "tls-roots"] }
 hyper-util = "0.1"
 tower = { version = "0.5", features = ["util"] }
--- a/crates/gazette/src/journal/append.rs
+++ b/crates/gazette/src/journal/append.rs
@@ -1,0 +1,107 @@
+use super::Client;
+use crate::{journal::check_ok, Error};
+use futures::{FutureExt, Stream, StreamExt};
+use proto_gazette::broker::{self, AppendResponse};
+use std::sync::Arc;
+
+impl Client {
+    /// Append the contents of a byte stream to the specified journal.
+    /// Returns a Stream of results which will yield either:
+    /// - An AppendResponse after all data is successfully appended
+    /// - Errors for any failures encountered.
+    /// If polled after an error, regenerates the input stream and
+    /// retries the request from the beginning.
+    pub fn append<'a, S>(
+        &'a self,
+        request: broker::AppendRequest,
+        source: impl Fn() -> S + Send + Sync + 'a,
+    ) -> impl Stream<Item = crate::Result<broker::AppendResponse>> + '_
+    where
+        S: Stream<Item = std::io::Result<bytes::Bytes>> + Send + 'static,
+    {
+        coroutines::coroutine(move |mut co| async move {
+            loop {
+                let input_stream = source();
+
+                match self.try_append(request.clone(), input_stream).await {
+                    Ok(resp) => {
+                        () = co.yield_(Ok(resp)).await;
+                        return;
+                    }
+                    Err(err) => {
+                        () = co.yield_(Err(err)).await;
+                        // Polling after an error indicates the caller would like to retry,
+                        // so continue the loop to re-generate the input stream and try again.
+                    }
+                }
+            }
+        })
+    }
+
+    async fn try_append<S>(
+        &self,
+        request: broker::AppendRequest,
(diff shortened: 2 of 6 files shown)


crates/gazette/src/journal/append.rs
```diff
@@ -0,0 +1,186 @@
...
+        // over an undefined period of time, that won't work. So instead we need to pass
+        // `JournalClient::append()` a stream of _just_ the `AppendRequest`s that come out
+        // of the above `request_generator`, while also promptly returning any errors if they
+        // crop up, and cancelling the append request.
+
+        let (req_tx, req_rx) = tokio::sync::mpsc::channel(100);
+
+        let mut client = self.into_sub(self.router.route(None, false, &self.default)?);
+
+        // Run `JournalClient::append` in a separate Tokio task, and feed it a steady diet of `AppendRequest`s
```
jgraettinger, line 136: We should avoid using tokio::spawn altogether. I don't believe it's necessary, and it significantly complicates the "drop to cancel" story.

Performing a concurrent poll of the next chunk of bytes AND polling to send an AppendRequest is a good idea, but that can also be achieved using `futures::join` instead.

crates/gazette/src/lib.rs
```diff
@@ -35,6 +35,8 @@ pub enum Error {
     },
     #[error("{0}")]
     Protocol(&'static str),
+    #[error("{0}")]
+    Append(String),
```
jgraettinger, line 39: use `Protocol` for fixed string constants, or a more-specific error type wrapping `std::io::Error` for other errors as needed (similar to `ReadFragment` for example)

jgraettinger (changes requested):

jshearer:
Alright @jgraettinger, want to take another look?

crates/gazette/src/journal/append.rs
```diff
@@ -0,0 +1,186 @@
...
+};
+
+// TODO: Tune this?
+const CHUNK_SIZE: usize = 1 << 14;
+
+impl Client {
+    /// Helper function to appends the contents of `source` to the specified journal.
+    /// This drives a single append RPC cycle and propagates any errors. If you need
+    /// streaming input or retries, use `[append_stream]` instead.
+    pub async fn append_once(
```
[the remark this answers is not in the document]
jshearer: Refactored to simplify it a bunch based on our convos in Slack

Cargo.toml
```diff
@@ -186,6 +186,7 @@ tokio = { version = "1", features = [
     "time",
 ] }
 tokio-util = { version = "0.7", features = ["io", "compat", "rt"] }
+tokio-stream = { version = "0.1.17" }
```
jgraettinger, line 189: I don't see tokio-stream as actually being used. Remove?

crates/gazette/src/journal/append.rs
```diff
@@ -0,0 +1,111 @@
...
+    /// retries the request from the beginning.
+    pub fn append<'a, S>(
+        &'a self,
+        request: broker::AppendRequest,
+        source: impl Fn() -> S + Send + Sync + 'a,
+    ) -> impl Stream<Item = crate::Result<broker::AppendResponse>> + '_
+    where
+        S: Stream<Item = std::io::Result<bytes::Bytes>> + Send + 'static,
+    {
+        let source = Arc::new(source);
```
jgraettinger, line 22: This Arc isn't needed and can be removed.

crates/gazette/src/journal/append.rs
```diff
@@ -0,0 +1,111 @@
...
+        coroutines::coroutine(move |mut co| async move {
+            loop {
+                let input_stream = source();
+
+                match self
+                    .append_all(request.clone(), Box::pin(input_stream))
```
jgraettinger, line 28: I believe you can remove the Unpin constraint on append_all, and then this Box::pin

crates/gazette/src/journal/append.rs
```diff
@@ -0,0 +1,111 @@
...
+                        () = co.yield_(Err(err)).await;
+                        // Polling after an error indicates the caller would like to retry,
+                        // so continue the loop to re-generate the input stream and try again.
+                    }
+                }
+            }
+        })
+    }
+
+    async fn append_all<S>(
```
jgraettinger, line 45: I'm not sure what "all" refers to in `append_all`. What about `try_append` or `append_once` ?

jgraettinger (approved):
LGTM % some nits below. Thanks

PR #1858, opened 2025-01-08
agent: Populate `live_specs.connector_image_name` and `live_specs.connector_image_tag` when publishing a Dekaf materialization

jshearer:
**Description:**

Populates `live_specs.connector_image_name` and `live_specs.connector_image_tag` when publishing a Dekaf materialization

The UI depends on `live_specs_ext` containing these values in order to display connector information, display task edit workflows/schema-driven forms, etc. Since we've already agreed on Dekaf "connectors" getting an image of `ghcr.io/estuary/dekaf-{variant}`, the only other convention is the tag, and I went with `:v1`.

The caveat is that since this value will get populated in `live_specs`, which `live_specs_ext` will then use to look up other information from `connector_tags`, we must ensure that those values match. Normally this isn't a problem because the full image+tag is stored in the task config, but since Dekaf doesn't use containers and so has no need to store an image tag, we instead have to explicitly validate that only the value `v1` is specified for the tag to avoid getting into a confusing/broken state.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1858)

### DIFF
--- a/crates/agent/src/connector_tags.rs
+++ b/crates/agent/src/connector_tags.rs
@@ -22,6 +22,7 @@ pub enum JobStatus {
 #[derive(Debug, Deserialize, Serialize)]
 pub enum ValidationError {
     ResourcePathPointersChanged { rejected: Vec<String> },
+    InvalidDekafTag,
 }
 
 impl JobStatus {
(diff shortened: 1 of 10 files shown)


psFried (approved):
LGTM

PR #1861, opened 2025-01-09
Phil/controller automations

psFried:
**Description:**

Transitions to using `automations` for live specs controllers. Individual commit messages have more details.

There will need to be another follow up PR to finish removing `live_specs.controller_next_run` after this is deployed.

**Workflow steps:** No user-facing changes

I've tested this by starting with the current agent version, publishing a bunch of different specs, and then running the first migration and starting a second agent with the new version.  The second migration enables automation-based controller runs for existing specs, and once that's converted all the specs, we can shut down the old agent version. I think when we deploy this to prod, we can migrate all the the existing specs pretty quickly, and can probably just remove the `limit 1000` after spot checking a few.

I've also tested rolling this back, just in case that's necessary. I haven't checked it in, but the rollback script is:

```sql
begin;

    update public.live_specs ls
    set controller_next_run = t.wake_at
    from (
        select task_id, wake_at
        from internal.tasks
        where task_type = 2
    ) t
    where ls.controller_task_id = t.task_id;

    alter table public.live_specs drop column controller_task_id;

    delete from internal.tasks where task_type = 2;
commit;
```

The existing agent controllers will then go back to handling all specs.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1861)


crates/agent-sql/src/publications.rs
```diff
@@ -88,14 +88,57 @@ pub async fn update_live_specs(
...
             from inputs
             left outer join live_specs ls on ls.catalog_name = inputs.catalog_name
         ),
+        create_controller_tasks as (
+            insert into internal.tasks (task_id, task_type, wake_at, inbox)
+            select
+                controller_task_id,
+                2,
+                now(),
+                array[json_build_array(
```
jgraettinger, line 117: This is replicating a lot of innards of `internal.send_to_task`.

Can you separate it into two CTE stages ? The first would create task rows which don't exist (on conflict do nothing), setting just the task type, task ID, and parent ID as zero.

The second CTE would then call `internal.send_to_task` across all now-existing task rows.

crates/agent/src/controllers/executor.rs
```diff
@@ -0,0 +1,204 @@
...
+    catalog_name = %controller_state.catalog_name
+))]
+async fn run_controller<C: ControlPlane>(
+    task_state: &mut State,
+    inbox: &mut VecDeque<(Id, Option<Event>)>,
+    controller_state: &ControllerState,
+    control_plane: &C,
+) -> (ControllerStatus, i32, Option<String>, Option<NextRun>) {
+    let mut next_status = controller_state.current_status.clone();
+    tracing::debug!(?inbox, "inbox events");
```
jgraettinger, line 170: confirming intent here:

The sent messages give really nice structured info about why the controller is being run.
Today, they're ignored (except for logging / metrics) because the existing controller factoring doesn't need them.

Tomorrow, we might want to toggle aspects of what the controller's doing based on the received message types. This would involve more refactoring of the controller than we want to do right now (plus we're still supporting legacy agents).

Is that accurate?

crates/automations/src/server.rs
```diff
@@ -31,9 +31,11 @@ impl Server {
         heartbeat_timeout: std::time::Duration,
         shutdown: impl std::future::Future<Output = ()>,
     ) {
+        let semaphore = Arc::new(tokio::sync::Semaphore::new(permits as usize));
```
jgraettinger, line 34: Let's talk through the changes to this source file -- I'm not understanding the motivation.

jgraettinger (commented):
Looks awesome!

One suggested cleanup on an agent SQL query below.
Otherwise, I'd like to understand the rationale behind the `automations` crate permit changes.

crates/automations/src/server.rs
jgraettinger: Okay, I get it now. It's about being able to finely control the execution of automations within a testing context.

crates/automations/src/server.rs
```diff
@@ -152,15 +153,12 @@ async fn ready_tasks_iter(
     semaphore: &Arc<tokio::sync::Semaphore>,
     task_types: &[i16],
 ) {
-    // Block until at least one permit is available.
-    if semaphore.available_permits() == 0 {
-        let _ = semaphore.clone().acquire_owned().await.unwrap();
-    }
-
     // Acquire all available permits, and then poll for up to that many tasks.
+    // If there's no currently available permits, then just acquire one.
```
jgraettinger, line 157: The downside of this change, is that it may be the case that _multiple_ permits become available before this async fn is polled again, and we would fail to pick up those multiple permits (and take only one).

Is this change required for testing use cases? It seems like it shouldn't be, so long as the test harness can control what permits are released to the semaphore?

crates/agent/src/controllers/executor.rs
psFried: Yeah, exactly.

crates/agent-sql/src/publications.rs
psFried: Yeah, I should have commented this. The reason I did this was because we frequently update hundreds of rows at a time, and I was concerned about the performance of repeatedly calling `send_to_task`, which IIUC would fetch each row separately vs doing a join. TBH I'd be quite willing to call that a 'premature optimization' and just use `send_to_task`. It's hard to do meaningful profiling outside of prod, so I was just going off vibes here and I don't really have useful intuitions. But if there's a perf issue with repeatedly calling `send_to_task`, then I'd expect it to show up pretty quickly in prod.

crates/automations/src/server.rs
psFried: The problem for tests was that `available_permits()` could return 0 because the harness holds the only available permit before this is called.

But here `n_permits` should be the larger value of `available_permits()` and `1`, so we'll still try to acquire multiple permits if they're available.

crates/automations/src/server.rs
```diff
@@ -226,11 +224,10 @@ async fn ready_tasks_iter(
 
     // If permits remain, there were not enough tasks to dequeue.
     // Sleep for up-to `dequeue_interval`, cancelling early if a task completes.
-    if permits.num_permits() != 0 {
+    if semaphore.available_permits() > 0 {
```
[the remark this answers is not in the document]
psFried: 🤔 I think you're right. I didn't realize that we're peeling off permits as we yield ready tasks, so I just thought that `num_permits` would return the number of permits that were actually acquired, which didn't seem consistent with the comment about sleeping if there were permits _remaining_. But I changed this after seeing it sleep during tests, so it seems like there might still be something off.

crates/agent-sql/src/publications.rs
jgraettinger: Fair. I would hope that it's not gonna be a problem, even if doing hundreds of updates, but I can't say that with any certainty. This is fine, I suppose, but it's a little eye watering.

crates/automations/src/server.rs
jgraettinger: the scenario i was thinking of is:
 - no permits are available. We set `n_permits` to max(1) = 1.
 - we call acquire_many_owned(1)
 - the executor is busy, and a number of other tasks complete releasing their permits before we're polled again.
 - we're polled again and acquire a single permit, even though more are available.

jgraettinger (approved):
LGTM % ReadyTask refactor discussed over VC

crates/automations/src/server.rs
psFried: just to close the loop on this, I ended up reverting this change along with the refactor to have `TestHarness` call `poll_task` directly.

PR #1865, opened 2025-01-13
agent: Switch `/authorize/dekaf` to fetch+return the task's built spec

jshearer:
**Description:**

We need the built spec as it contains the final output of field selection, which Dekaf needs in order to generate the correct schema and output documents. Only https://github.com/estuary/flow/pull/1840 uses this endpoint, and it's not live yet so there shouldn't be any risk in changing this.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1865)


psFried (approved):
LGTM

PR #1867, opened 2025-01-14
support for journal suspension

jgraettinger:
**Description:**

Flow runtime changes in support of journal hibernation:

- Skip or drain shuffled reads of suspended journals.
- Pass-through suspension configuration of JournalSpecs when activating catalog changes.
- Update local Tiltfile to use --broker.file-only flag and multiple gazette brokers
- Other minor bug fixes.

Issue #1837

**Workflow steps:**

No notable behavior differences. The reactor is now simply aware of suspended journals and how to (not) process them.

**Notes for reviewers:**

Quite a bit of local stack testing using Tiltfile, and I additional resurrected the `kustomize`-based Gazette soak tests.
I manually spammed a very busy local cluster / soak test with forced suspensions, and saw quick resumptions as new appends awoke the force-suspended journals.

I also tested various permutations of:
 - Partially and fully suspended journals
 - Materializations stopping and re-starting reads of fully-suspended journals.
 - Enabling tasks which write to suspended journals

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1867)


crates/activate/src/lib.rs
```diff
@@ -129,29 +154,23 @@ pub async fn activate_materialization(
...
-        Some(task_spec) => {
-            let shard_template = task_spec
-                .shard_template
-                .as_ref()
-                .context("MaterializationSpec missing shard_template")?;
+    let task_template = if let Some(task_spec) = task_spec {
+        let shard_template = task_spec
+            .shard_template
+            .as_ref()
+            .context("MaterializationSpec missing shard_template")?;
```
psFried, line 161: TIL that `Option` has a `Context` impl. Nice

go/shuffle/read.go
```diff
@@ -603,6 +603,8 @@ func walkReads(id pc.ShardID, shardSpecs []*pc.ShardSpec, shuffles []shuffle,
 			var start, stop int
 			// filtered indicates that the partition is excluded and not read by this shard.
 			var filtered = !shuffle.sourcePartitions.Matches(source.LabelSet)
+			// suspended indicates that the partition is fully suspended and has no content.
+			var suspended = source.Suspend.GetLevel() == pb.JournalSpec_Suspend_FULL
```
psFried, line 607: I'm not sure I fully understand why we're only skipping journals that are fully suspended here. I had expected that we'd also short-circuit reads for journals that are only partially suspended after the read reaches the current write head. Is that right? Or do we do a blocking read for each partially suspended journal after reaching the current head?

psFried (approved):
LGTM

go/shuffle/read.go
jgraettinger: > I had expected that we'd also short-circuit reads for journals that are only partially suspended after the read reaches the current write head.

We could certainly do this, but I didn't implement it. We'll instead just use the regular read path, and have a blocking read established with that singular broker which never does anything.

PR #1868, opened 2025-01-14
agent: discovers reject duplicate bindings

### LINKED ISSUE #1833: Discovers should fail if the discovered bindings contain duplicates
When processing discovers, error if the discover response includes multiple bindings that have the same resource path.

psFried:
**Description:**

Resolves #1833
Updates the discover merge logic to return an error if it detects multiple bindings with the same resource path.  Previously, we had just logged in this scenario, and those logs confirmed that only two captures were affected by this issue, and both were already failing.  So this just ensures that we fail fast with a better error message in case this shows up again.

**Workflow steps:**

Do a discover (or auto-discover) with a source connector that returns multiple bindings having the same resource path. This should fail the discover with an error about the duplicated bindings.

**Documentation links affected:**: n/a

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1868)


jshearer (approved):
LGTM

PR #1881, opened 2025-01-17
runtime: add materialize-kafka to ser_policy

williamhbaker:
**Description:**

Kafka has a default maximum message size of 1MiB.

This adds `materialize-kafka` to the list of connectors that use `ser_policy`, which should generally ensure that documents with individual extremely large string, array, or object fields don't exceed the message size limit for Kafka.

This simple heuristic may not always be 100% what users desire, but in general it seems better to have the connector at least sort-of-work rather than crash in a way that would require a more complicated fix (possibly a derivation) to best
support data exploration / PoC use cases.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1881)

### DIFF
--- a/crates/runtime/src/materialize/task.rs
+++ b/crates/runtime/src/materialize/task.rs
@@ -32,6 +32,7 @@ impl Task {
         // See go/runtime/materialize.go:135
         let ser_policy = if [
             "ghcr.io/estuary/materialize-bigquery",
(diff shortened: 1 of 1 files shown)


jgraettinger (approved):
LGTM

PR #1885, opened 2025-01-21
agent: sanitize draft error messages before persisting

psFried:
Postgres text columns cannot store strings containing a null byte. This sanitizes error messages by replacing null bytes with the unicode replacement character, so that they can be stored in the database.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1885)


mdibaiee (approved):
thanks Phil

PR #1895, opened 2025-01-23
Phil/fetch live chunks

psFried:
Tiny PR to update agent to fetch live specs in chunks, to avoid statement timeouts. We were seeing these queries fail when discover responses included a large number of bindings. Also removes the limit on the number of discovered bindings, since that's no longer necessary.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1895)


jgraettinger (approved):
LGTM

PR #1897, opened 2025-01-24
gazette: refactor retry, introduce backoffs, and fixing routing

jgraettinger:
Previously, gazette routing was previously broken in some subtle ways:

* The process ID of a prior response was sent in the next request, causing "ProcessID doesn't match our own" errors in production.

* Improperly assuming that endpoints present in response headers are reach-able from the client. We need to condition that assumption on the `do_not_proxy` flag.

* `append` routine was not leveraging a prior response header to route its next attempt.

Sleep-based backoffs are also introduced to retry-able client routines, rather than making them a responsibility of the caller. This standardizes handling and prevents accidental thundering-herds of retries if a caller forgets to put in a backoff. Now callers don't need to worry about it, and must only decide whether to re-poll, drop to cancel, log, or take some other action.

We additionally want to surface a notion of how many operation _attempts_ there have been since the last success. This is important both for logging (logged attempts in Go implementations are useful), and also for client determinations of when to give up.

Update simd_doc::Parser::chunk to be able to recover should a caller provide discontinuous partial documents. This was a missing re-entrancy case when retrying a ReadJsonLines stream.

Reserve a simd_doc::Parser::transcode_many() buffer based on the input size, to prevent frequent re-allocations in the common usage of ReadJsonLines.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1897)

### DIFF
--- a/crates/gazette/src/journal/list.rs
+++ b/crates/gazette/src/journal/list.rs
@@ -1,5 +1,5 @@
 use super::{check_ok, Client};
-use crate::Error;
+use crate::{router, Error};
 use futures::TryStreamExt;
 use proto_gazette::broker;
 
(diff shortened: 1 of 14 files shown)


jgraettinger:
This ended up bigger than I expected, but I started pulling on strings and this is where it landed.

crates/gazette/src/journal/list.rs
```diff
@@ -16,32 +16,42 @@ impl Client {
...
-                                // Server stopped an ongoing watch. Expected and not an error.
-                            } else {
-                                co.yield_(Err(err)).await;
-                            }
-                            break; // Start new stream on next poll.
-                        }
-                    }
+                if matches!(err, Error::UnexpectedEof if req.watch_resume.is_some()) {
+                    // Server stopped an ongoing watch. Expected and not an error.
+                    continue;
```
jshearer, line 48: Should this also get a backoff sleep?

jshearer (approved):
LGTM

crates/gazette/src/journal/list.rs
jgraettinger: No, because this can only be a clean server closure of a prior List RPC, which happens when the member is stopping. We can immediately retry with another member.

PR #1899, opened 2025-01-27
Phil/agent fails

psFried:
**Description:**

Prevents most of the remaining discover failure scenarios from bubbling up to a more widespread failure of the agent. Discussed in [this Slack thread](https://estuaryworkspace.slack.com/archives/C03QBN83GQ4/p1737689571320549). Returning an error from a `Handler` will cause the agent to crash, and it will also result in the job that was being handled remaining `queued` (so another agent will immediately try to handle it). Thus an error in one job could severely degrade our handling of other jobs. This fixes that by failing the discover instead of returning an error from the Handler.

I also reviewed the error handling in the handlers for publications, connector tags, and directives, didn't see any other failure scenarios that looked like they'd need similar treatment.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1899)

### DIFF
--- a/crates/agent/src/discovers/handler.rs
+++ b/crates/agent/src/discovers/handler.rs
@@ -129,7 +129,7 @@ impl<C: DiscoverConnectors> DiscoverHandler<C> {
         };
 
         let image_composed = format!("{}{}", row.image_name, row.image_tag);
-        let disco = prepare_discover(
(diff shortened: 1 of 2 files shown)


jgraettinger (approved):
LGTM

Issue #1876, opened 2025-01-16
dekaf: Improvements to handle higher scale


jshearer:
* https://github.com/estuary/flow/pull/1911
* similar to MSK, limit the _rate_ of new connection requests
* possibly also add a connection cap
* Limit things like `max.wait.ms`

PR #1911, opened 2025-01-31
dekaf: Upstream kafka fallback and per-connection-type routing

jshearer:
**Description:**

As part of #1876, we want to implement broker fallback so Dekaf can connect to any of the brokers in the cluster if one doesn't respond. This adds the ability to pass a list of broker addresses, which Dekaf will try one after another until one picks up.

An improvement here would be to periodically fetch the metadata from at least one of the responding brokers and update this list of addresses so that future sessions can know about/use any newly created members of the cluster. I don't anticipate changing the topology of our cluster that frequently, and if we do then updating Dekaf's deployment configs isn't that big of a deal. I may eat my hat on this, we'll see.

In addition, we want to move people over to the new MSK cluster, so this implements routing new-style connections to a separate cluster with separate credentials.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1911)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -353,36 +353,12 @@ checksum = "b3d1d046238990b9cf5bcde22a3fb3584ee5cf65fb2765f454ed428c7a0063da"
 
 [[package]]
 name = "apache-avro"
-version = "0.16.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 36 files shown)


jshearer:
Rolling into https://github.com/estuary/flow/pull/1840

PR #1923, opened 2025-02-04
dekaf: Implement capturing and appending stats

jshearer:
**Description:**

A couple things to note:

* I originally tried to create a single `journal::Client` responsible for appending both logs and stats, but I ended up realizing that `/authorize/task` only allows authoring a token for a single task/prefix at a time. So I took the simpler route of creating two clients, rather than teaching `/authorize/task` how to handle multiple tasks, which has some fairly delicate requirements.
* As it turns out, the stats rollups assume the presence of a `shard` field on both logs and stats. So I ended up needing to craft a `ShardRef` that just contains the Dekaf materialization name, and attach it to both the logs and stats documents that get emitted.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1923)


jshearer:
Temporarily switching base to master and marking this as non-draft so CI will run and I can get a dev build to test on `dekaf-dev`. This is actually ontop of https://github.com/estuary/flow/pull/1840

PR #1930, opened 2025-02-11
Update GitHub Workflows

skord:
**Description:**
This updates the GitHub workflows to use later versions of ubuntu as GitHub is deprecating the use of ubuntu-20.04 in the near future.

It does not touch the dev container, as I'm not sure any actually uses it.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1930)


travjenkins (approved):
lgtm - the version updates are fine and I am assuming yaml does not care about single vs double quotes so those are safe changes

PR #1947, opened 2025-02-19
dekaf: Improve log filtering

jshearer:
Since we're now piping `tracing` logs into the task logs which is user facing, we'd like to make the logs we send as meaningful as possible. In order to do that, we need to be able to more narrowly specify which logs should get sent. In order to do that, I've swapped the simple log level filter for a slightly more advanced `tracing_subscriber::filter::Targets` which allows us to specify log filters such as `debug,simple_crypt=warn,aws_config=warn,h2=warn`.

An alternative to Arc<RwLock<T>> would be something like the the [`arc-swap`](https://docs.rs/arc-swap/latest/arc_swap/) crate which provides a lock-free way to accomplish this shared mutable state, but since the usage pattern of `LOG_LEVEL` is pretty much entirely reads, I don't anticipate any actual lock contention to happen here and wanted to avoid yet another new dependency.

**This appears to have some locking issue, holding off for the moment.**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1947)

### DIFF
--- a/crates/dekaf/src/log_appender.rs
+++ b/crates/dekaf/src/log_appender.rs
@@ -494,17 +494,17 @@ impl<W: TaskWriter + 'static> TaskForwarder<W> {
     pub fn send_stats(&self, collection_name: String, stats: ops::stats::Binding) {
         if stats
             .left
-            .is_some_and(|s| s.bytes_total == 0 || s.docs_total == 0)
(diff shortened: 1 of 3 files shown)


jshearer:
Closing in favor of https://github.com/estuary/flow/pull/1965

PR #1950, opened 2025-02-19
doc: Fix intersection of shape defaults

### LINKED ISSUE #1944: Fix intersected shapes having possibly invalid `default`s
A customer found this when trying to use Dekaf on data coming from `source-hubspot-native`.

the connector schema says a location is `type: [integer, null]`  and `default: null` (for some reason). The inferred schema says that it's `type: integer` but not always present, i.e not `required`.  We turn that into this Shape:
```
ObjProperty {
    name: "my_name",
    is_required: false,
    shape: Shape {
        type_: "integer",
        default: Some(
            (
                Null,
                None,
            ),
        ),
        ... defaults
    },
},
```

When we extract `/my_name`, we get the default value of `null`.. but the `allOf` requires that the type be `integer`, not null. Now, we *do* get `Exists::May` for `/my_name`, so we pass `required = false` to `shape_to_avro()`  which is supposed to add a union with `NULL` if it's not required... but I believe that [this line](https://github.com/estuary/flow/blob/master/crates/avro/src/schema.rs#L13) is the culprit: changing it to also mark the field as nullable if its default is null fixes the encoding error:

```
diff --git a/crates/avro/src/schema.rs b/crates/avro/src/schema.rs
index 8237e15e2f..2f8a46cde5 100644
--- a/crates/avro/src/schema.rs
+++ b/crates/avro/src/schema.rs
@@ -10,7 +10,12 @@ pub fn shape_to_avro(loc: json::Location, shape: doc::Shape, required: bool) ->
     let mut type_ = shape.type_;

     // Is this location nullable ? NULL may union with any other schema.
-    let nullable = if shape.type_.overlaps(types::NULL) || (!required && shape.default.is_none()) {
+    let default_is_none_or_null = if let Some(default) = shape.default {
+        default.as_ref().0 == serde_json::Value::Null
+    } else {
+        true
+    };
+    let nullable = if shape.type_.overlaps(types::NULL) || (!required && default_is_none_or_null) {
         type_ = type_ - types::NULL;
         true
     } else {
```

jshearer:
**Description:**

We weren't correctly narrowing the default of intersected shapes. For example, if we intersected one shape like
```json
{
    "type": ["string", "null"],
    "default": "null"
}
```

with another shape like
```json
{
    "type": "string"
}
```

We'd end up with a shape of
```json
{
    "type": "string",
    "default": null
}
```

Which isn't correct, as when we try to actually emit a document for this shape that's missing the field in question and provide its default value of null, that will subsequently fail to validate against the schema.

Also adds a test to confirm Dekaf's proper handling of schemas like this, which is why this branch is ontop of that one.

Fixes #1944

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1950)

### DIFF
--- a/crates/dekaf/tests/dekaf_schema_test.rs
+++ b/crates/dekaf/tests/dekaf_schema_test.rs
@@ -238,6 +238,64 @@ async fn roundtrip(
 }
 
 #[tokio::test]
+async fn test_allof_with_null_default() -> anyhow::Result<()> {
+    for output in roundtrip(
+        connector::DekafConfig {
+            deletions: connector::DeletionMode::Kafka,
+            token: "1234".to_string(),
+            strict_topic_names: false,
+        },
+        json!({
+          "schema": {
+            "allOf": [
+              {
+                "properties": {
+                  "id": {
+                    "title": "Id",
+                    "type": "integer"
+                  },
+                  "conflicts": {
+                    "type": ["integer", "null"],
+                    "default": null,
+                    "title": "Updatedbyuserid"
+                  }
+                },
+                "required": ["id"],
+                "type": "object"
+              },
+              {
+                "properties": {
+                  "id": {
+                    "title": "Id",
+                    "type": "integer"
+                  },
+                  "conflicts": {
+                    "type": "integer"
+                  }
+                },
+                "required": ["id"],
+                "type": "object"
+              }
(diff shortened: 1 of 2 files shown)


psFried (approved):
LGTM

PR #1952, opened 2025-02-19
dekaf: Fix `_meta/is_deleted` for old-style connections

jshearer:
**Description:**

This adds back in the original logic around `/_meta/is_deleted` for refresh-token-based connections in order to avoid breaking them. 

I also realized that since people are going to have to create new dataflows when they upgrade to Dekaf materializations anyway, we don't have to keep doing the janky logic to actually _mutate_ the `/meta` location to include `is_deleted`, we can just add a new extractor for a location like `_is_deleted`.

Ideally I'd like for the connector to expose this field during field selection, but we currently have a (perfectly reasonable) assertion that the connector can't return constraints for fields that aren't projections.

Based ontop of https://github.com/estuary/flow/pull/1957 for testing purposes

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1952)


crates/dekaf/src/utils.rs
```diff
@@ -33,7 +36,26 @@ impl CustomizableExtractor {
...
+            CustomizableExtractor::RootExtractorWithIsDeleted => {
+                let deletion = match META_OP_PTR.query(doc) {
+                    Some(n) => match n.as_node() {
+                        doc::Node::String(s) if s == "d" => 1,
+                        _ => 0,
+                    },
+                    None => 0,
+                };
+
+                let mut full_doc = doc.to_debug_json_value();
```
jshearer, line 50: The alternative is to figure out how to plumb through an allocator here and turn `doc` into a `HeapNode`... but unless I'm missing something, this seems fine as a stop-gap -- this codepath goes away once everyone migrates over to materializations and we turn off support for that connection method.

PR #1953, opened 2025-02-19
docs/SQLServer: Remove mention of watermarks

willdonnelly:
**Description:**

Now that github.com/estuary/connectors/pull/2385 is merged the default capture behavior for new users will be read-only, which requires the `VIEW DATABASE STATE` permission but doesn't use a watermarks table. This updates the docs to reflect that.

Part of https://github.com/estuary/connectors/issues/2339

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1953)


dyaffe (approved):
LGTM

PR #1956, opened 2025-02-20
Dekaf: fix datetimes and stats

jshearer:
* Stop using `apache_avro::schema::Schema::canonical_form()` for output schemas. This transformation strips out important information from schemas such as `logicalType`, which clients were using to differentiate between e.g numbers and date-timestamps. In `0.17.0` it was [upgraded](https://github.com/apache/avro/pull/2976) to specifically strip `logicalType`.
* Fix emitting incorrect stats docs when we have an ack-only read. We were correctly ignoring the bytes of ack documents, but were still counting them as documents. This was resulting in recording stats like

  ```
  DocsAndBytes {
    docs_total: 1,
    bytes_total: 0,
  }
  ```

  This was subsequently encoded as
  ```
  {
    "docsTotal": 1
  } 
  ```

  Which doesn't match the schema for stats, and brought down the whole L1 rollup shard.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1956)

### DIFF
--- a/crates/dekaf/src/log_appender.rs
+++ b/crates/dekaf/src/log_appender.rs
@@ -492,7 +492,23 @@ impl<W: TaskWriter + 'static> TaskForwarder<W> {
     }
 
     pub fn send_stats(&self, collection_name: String, stats: ops::stats::Binding) {
-        self.send_message(TaskWriterMessage::Stats((collection_name, stats)))
+        if stats
+            .left
+            .is_some_and(|s| s.bytes_total == 0 || s.docs_total == 0)
+            || stats
+                .right
+                .is_some_and(|s| s.bytes_total == 0 || s.docs_total == 0)
+            || stats
+                .out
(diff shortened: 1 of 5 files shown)


psFried (approved):
LGTM
Just to double check, is there any other scenario you can think of that might result in either `docsTotal` or `bytesTotal` being 0? Would it be worth adding an assertion or filter? I'm guessing the answer to both is no, but being a little extra cautious.

jshearer:
I'll add an assertion just in case

jshearer:
Excellent, confirming that this fixes the problem -- I'm seeing dates as `"receivedAt": "2025-02-19T21:23:53.861Z"` on both dekaf-dev and dekaf prod, and also seeing tinybird inferring these columns as the correct `DateTime64` instead of `UInt64`
Screenshot 2025-02-20 at 15 40 23

PR #1965, opened 2025-02-24
Dekaf: Fixes for release

jshearer:
I decided to roll all of my remaining changes into one PR for easier reviewing. We have:

* Improvements to log filtering
  > Since we're now piping `tracing` logs into the task logs which is user facing, we'd like to make the logs we send as meaningful as possible. In order to do that, we need to be able to more narrowly specify which logs should get sent. In order to do that, I've swapped the simple log level filter for a slightly more advanced `tracing_subscriber::filter::Targets` which allows us to specify log filters such as `debug,simple_crypt=warn,aws_config=warn,h2=warn`.
* Add back in the original logic around `/_meta/is_deleted` for refresh-token-based connections in order to avoid breaking them. 
  > I also realized that since people are going to have to create new dataflows when they upgrade to Dekaf materializations anyway, we don't have to keep doing the janky logic to actually _mutate_ the `/meta` location to include `is_deleted`, we can just add a new extractor for a location like `_is_deleted`.
   >
   > Ideally I'd like for the connector to expose this field during field selection, but we currently have a (perfectly reasonable) assertion that the connector can't return constraints for fields that aren't projections.
* https://github.com/estuary/flow/pull/1950 which was already approved, but wasn't merged into `master` because it was ontop of the `validation-test` refactor

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1965)


crates/dekaf/src/log_appender.rs
```diff
@@ -494,17 +494,17 @@ impl<W: TaskWriter + 'static> TaskForwarder<W> {
     pub fn send_stats(&self, collection_name: String, stats: ops::stats::Binding) {
         if stats
             .left
-            .is_some_and(|s| s.bytes_total == 0 || s.docs_total == 0)
```
jshearer, line 497: I realized that this was too restrictive and was complaining too much -- it's fine to emit stats where both `bytes_total` and `docs_total` are 0 because of the filter [here](https://github.com/estuary/flow/blob/29e8bc6f744ff28e7e2322cb8c03c4ab58ae5dbc/crates/dekaf/src/log_appender.rs#L56-L60)

crates/dekaf/src/utils.rs
```diff
@@ -33,7 +36,26 @@ impl CustomizableExtractor {
...
+            CustomizableExtractor::RootExtractorWithIsDeleted => {
+                let deletion = match META_OP_PTR.query(doc) {
+                    Some(n) => match n.as_node() {
+                        doc::Node::String(s) if s == "d" => 1,
+                        _ => 0,
+                    },
+                    None => 0,
+                };
+
+                let mut full_doc = doc.to_debug_json_value();
```
jshearer, line 50: The alternative is to figure out how to plumb through an allocator here and turn doc into a HeapNode... but unless I'm missing something, this seems fine as a stop-gap -- this codepath goes away once everyone migrates over to materializations and we turn off support for that connection method.

jgraettinger:
>Move validation test infrastructure into shared validation-test crate in order to use it in dekaf. Specifically, I needed to be able to convert an input fixture spec into its corresponding built output, which is exactly what validation_test::run() does.

I'm going to push back on moving all of this, and introducing `validation-tests` at all.
Instead, look at the `activate` crate, it's `managed_build` routine, and how that's used in its tests.

jgraettinger:
I think this needs to be split up. What is 100% necessary to land for tomorrow? I'd like to see a focused PR just on those essential changes.

jshearer:
> look at the activate crate, it's managed_build routine, and how that's used in its tests.

Okay cool, will do. In the mean-time, I rewrote the commits here to omit all of the testing stuff and will land that in a separate PR that doesn't mess with `validation`

crates/dekaf/src/log_appender.rs
jgraettinger: Why is this filtering required in the first place? How does it come to be that bytes_total != 0 but docs_total == 0, or vice versa?

Can you instead make this an assert()-ion rather than a tracing::error?

crates/dekaf/src/utils.rs
jgraettinger: The conversion into / out of serde_json::Value is ... fine, especially given it's short expected life-cycle, but did you look at what `to_debug_json_value` does / why it's a thing? It's behavior is not appropriate here.

jgraettinger (approved):
LGTM % comments

jgraettinger:
What was the resolution for the date-time handling observed in the bump from avro 0.16 => 0.17? Did a PR for that land separately?

jshearer:
That's in the k8s pod spec, it'll be in the ops PR to deploy this

crates/dekaf/src/utils.rs
jshearer: Ah I see -- I looked at the function and saw it was serializing to a `serde_json::Value`, but didn't look into what `SerPolicy::debug()` does. I think I want `SerPolicy::noop()` instead to disable any truncation of the output

PR #1966, opened 2025-02-24
docs/backfilling-data: Recommend plain backfills

willdonnelly:
**Description:**

This bit of documentation could very easily be read as implying that users _ought to_ follow a more complicated process to pause writes and perform an only-changes backfill when upgrading PostgreSQL.

My perspective on the matter is that just hitting the "Backfill" button should always be our #1 recommendation because it's easy for users and is the most guaranteed-to-work-in-all-cases option. While many users probably will want the more complicated approach, I want us to be clear that this is an _optimization_ to reduce the time and data usage of a backfill, where the user is deliberately making a tradeoff of doing more setup work and manually making certain guarantees in order to avoid the full backfill.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1966)

### DIFF
--- a/site/docs/reference/backfilling-data.md
+++ b/site/docs/reference/backfilling-data.md
@@ -15,14 +15,16 @@ In this case, many connectors allow you to turn off backfilling on a per-stream 
 
 ### Preventing backfills during database upgrades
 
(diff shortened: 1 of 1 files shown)


Alex-Bair (approved):
LGTM

PR #1970, opened 2025-02-25
docs/source-postgres: Discuss read-only captures

willdonnelly:
**Description:**

Adds a section to our docs discussing the "Read-Only Capture" advanced option, the conditions for using it, and how it can be applied to capturing from a read-only standby (and the extra setup that requires).

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1970)


dyaffe (approved):
LGTM

PR #1971, opened 2025-02-26
dekaf: Prevent orphaned sessions when logging fails

jshearer:
We noticed that sometimes, an already-authenticated session would get a second `SaslAuthenticate` message. This would cause log forwarding to crash and stop emitting logs for that task, while leaving the task still alive.

- Ensure logging failures propagate up to terminate the affected session: if logging can't function properly, the session should not continue

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1971)

### DIFF
--- a/crates/dekaf/src/log_appender.rs
+++ b/crates/dekaf/src/log_appender.rs
@@ -1,7 +1,6 @@
 use crate::{dekaf_shard_template_id, topology::fetch_dekaf_task_auth, App};
 use anyhow::Context;
 use async_trait::async_trait;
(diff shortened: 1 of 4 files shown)


jshearer:
Found another cause of these: when the request to `/authorize/dekaf` fails (in this case due to `failed to fetch task spec: pool timed out while waiting for an open connection`) in the logging handler but not in the Session handler, so the session starts but its logging is broken

psFried (approved):
LGTM

PR #1974, opened 2025-02-27
dekaf: Prevent already authenticated sessions from handling auth requests

jshearer:
Explicitly reject re-authentication attempts: a session that's already authenticated can't be authenticated again.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1974)


jgraettinger (approved):
LGTM

PR #1975, opened 2025-02-27
flowctl: Add support for `flowctl collections read --offset`

jshearer:
I needed this to debug Dekaf, and figured it would be a useful addition to our arsenal.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1975)


mdibaiee (approved):

jgraettinger:
This isn't public API that we can support, and I think it needs to be reverted. There's a reason that gazctl doesn't offer a similar `offset` flag, and instead has a more complex offsets map. Collections can have lots of journals: which one is meant by the --offset ?

PR #1979, opened 2025-02-28
agent-sql: storage mapping query should pull all mappings prefixed by tenant

jgraettinger:
Rather than just exact matches. Verified (on a local stack) that his allows a tenant to be set up with sub-mappings (only), and that they're correctly used.

**Description:**

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1979)

### DIFF
--- a/crates/agent-sql/src/publications.rs
+++ b/crates/agent-sql/src/publications.rs
@@ -596,7 +596,7 @@ pub async fn resolve_storage_mappings(
             m.catalog_prefix,
             m.spec
         from prefixes p
-        join storage_mappings m on m.catalog_prefix = p.prefix;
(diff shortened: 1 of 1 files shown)


psFried (approved):
Just to double check, I think we could end up pulling in storage mappings for, e.g. `a/b/c/`, even if the publication only includes specs under `a/b/x/`. But I'm assuming that's no big deal because we'd just ignore the extra storage mappings during validation. Assuming that's right, then LGTM

jgraettinger:
Correct. I think this is actually ideal, because if there's a typo in the task name, it will helpfully suggest "did you mean a/b/c/ ?"

Issue #1984, opened 2025-03-03
user-facing OpenMetrics API endpoint


jgraettinger:
We'd like to offer a public OpenMetrics API endpoint, powered by Estuary refresh tokens, which allows a user to integrate our reporting with their preferred metrics service, such as DataDog, Cloud Grafana, or self-hosted Prometheus.

PR #1985, opened 2025-03-03
add user-facing OpenMetrics API

jgraettinger:
**Description:**

This PR offers a new public OpenMetrics API, which draws from monthly catalog stats and is powered by user refresh tokens.

It transforms current-month stats of the `catalog_stats` table into a number of metrics which are scrape-able by user metrics infrastructure.

This is a baseline / proof-of-concept set of metrics. We intend to add new measures around transaction count and last source published-at timestamps shortly.

Issue #1984

**Workflow steps:**

* Create a refresh token in the Estuary dashboard.
* Integrate our public API into a Prometheus or DataDog scrape job (instructions forthcoming)
* Profit.

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/1985)


jgraettinger:
DataDog integration:

image

Prometheus integration:

image

Prometheus types / help extraction: 

image

crates/agent/src/api/public/open_metrics.rs
```diff
@@ -0,0 +1,505 @@
...
+use crate::api::{public::ApiErrorExt, snapshot::Snapshot, App, ControlClaims};
+use axum::response::IntoResponse;
+use chrono::{Datelike, TimeZone};
+use futures::TryStreamExt;
+use ops::stats::DocsAndBytes;
+
+#[axum::debug_handler]
+pub async fn handle_get_metrics(
+    state: axum::extract::State<Arc<App>>,
+    axum::Extension(ControlClaims {
```
psFried, line 14: I'm noting that the function signature here only requires a `ControlClaims`, and doesn't require that they come from a refresh token. It would be nice to encode that requirement in the function signature, so that we can also surface it as part of the openapi spec. Currently the spec just hard codes the requirement to provide the access token, which would be misleading for anyone who's actually looking at our api docs. So IDK whether this is really worth doing right now, but I think one solution might be to define a `struct RefreshToken(ControlClaims);`, and then use it as a parameter of type `axum_extra::TypedHeader<RefreshToken>`, and `impl Header for RefreshToken ...`. 

Or maybe a simpler solution would be to just manually add a `RefreshToken` security scheme [to the openapi spec here](https://github.com/estuary/flow/blob/877db5668237a9518866ac16d825bb64775d6071/crates/agent/src/api/public/mod.rs#L123), and use [aide::ApiOverride](https://docs.rs/aide/latest/aide/trait.ApiOverride.html) to specify the requirement to provide a refresh token.

I've found security requirements to be the most complicated and annoying part of dealing with openapi specs, so I can also live with just having a description or something that says that this route is special and needs a refresh token.

crates/agent/src/api/mod.rs
```diff
@@ -234,34 +242,92 @@ where
...
 
-    req.extensions_mut().insert(token.claims);
+    req.extensions_mut().insert(claims);
     next.run(req).await
 }
 
+// Middleware which exchanges a refresh bearer token for an access token,
+// validates the generated access token before proceeding,
+// and attaches verified ControlClaims encoded within its JWT.
+async fn layer_refresh_claims(
```
psFried, line 269: Another possible solution to the openapi spec auth requirements might be to have a layer that accepts either an access token or a refresh token, and allow all of our routes to accept either.

psFried (commented):
I found a few things that may be worth talking though, though I don't know that any of them are necessarily blocking. The main question to resolve is just whether it's worth adding the prefix to the url path. But I could live with merging it as is.

crates/agent/src/api/mod.rs
jgraettinger: :+1: I think this is the way to go. There's no _requirement_ that his api take a refresh token, it's just not practical to configure grafana-agent / datadog-agent et all to do a token exchange flow.

I can't think of any downside to letting other APIs take a refresh token either.

psFried (approved):
LGTM

PR #2001, opened 2025-03-12
incremental refactors towards better validations

jgraettinger:
**Description:**

Incremental refactors towards better validations:
* Remove `Projection.write_inference`
* Add `reset` flag to the Collection model, which will support collection reset.
* Add `model_fixes` to `tables::BuiltCollection` and friends, into which validations will record details of model tweaks they undertake.
* Refactor how we invoke connectors in the control plane.
  * Instead of using unary calls, have validations start a connector RPC, issue Spec, do some thinking, and _then_ issue Validate.
* Remove materialization `resource_path_pointers`, and replace with a new validation routine for extracting annotations of a materialization's resource config schema.

Briefly tested this on a local stack. The last commit in particular adds a bunch of new validation checking of materialization resource config schemas. Spot-checking production, all connectors appear to already be in compliance. If they're not, landing this will provide early warning in the connectors repo.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2001)


jgraettinger:
Next planned steps (a different PR?) will be: 
* Aligning the bindings of the live model, the live built spec, and the draft model into a structure that makes introspection and tweaking easier, using resource path pointers.
* Implementing various validation tweaks which modify the model and record details as `model_fixes`.

psFried (approved):
LGTM
One consideration I want to just call out is that I feel a little uncertainty about handling model-update scenarios with disabled tasks. The potential risk I'm thinking of is that disabled tasks use a dummy `Spec` response, so it could be a problem if we updated a model based on that spec. It's hard to tell whether there's really anything to worry about there, since there's no model changes in this PR. But I just wanted to bring it up before you get into that aspect.

PR #2002, opened 2025-03-12
control-plane: bug fixes related to ops catalog changes

psFried:
Rolls up a few fixes related to the ops catalog changes:

- Adds not null constraints, which were removed from the original `ops_catalog_status` migration in order to allow legacy agent processes to create new data planes.  This just breaks out those constraints to a separate migration.
- Fixes the catalog stats L1 rollup. I'd made a last minute change to validate that the type of `/fields/error` is a string on events documents, and it broke the whole thing. This fixes it and adds a test, which really should have been there from the beginning.
- Fixes the `new_free_trial_tenants` view. The last line of the view definition got missed during the copypasta, which caused us to send out a bunch of free trial emails accidentally. This fixes it and adds a sql test.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2002)

### DIFF
--- a/supabase/migrations/20250312121737_data-planes-events-column-constraints.sql
+++ b/supabase/migrations/20250312121737_data-planes-events-column-constraints.sql
@@ -1,0 +1,19 @@
+begin;
+
+-- Populate these new columns. Note that the legacy data plane needs some
+-- special handling because it doesn't follow the normal naming convention.
+-- The substr call is to strip off the 'ops/dp/' prefix.
+update public.data_planes
+set
+    ops_l1_events_name = case when data_plane_name = 'ops/dp/public/gcp-us-central1-c1'
+        then 'ops/rollups/L1/public/gcp-us-central1-c1/events'
+        else concat('ops/rollups/L1/', substr(data_plane_name, 8), '/events')
+        end,
+    ops_l2_events_transform = concat('from.', data_plane_fqdn, '.events')
+where
+    ops_l1_events_name is null and ops_l2_events_transform is null;
+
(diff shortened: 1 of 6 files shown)


supabase/migrations/20250312121737_data-planes-events-column-constraints.sql
```diff
@@ -0,0 +1,19 @@
+begin;
+
+-- Populate these new columns. Note that the legacy data plane needs some
+-- special handling because it doesn't follow the normal naming convention.
+-- The substr call is to strip off the 'ops/dp/' prefix.
+update public.data_planes
+set
+    ops_l1_events_name = case when data_plane_name = 'ops/dp/public/gcp-us-central1-c1'
+        then 'ops/rollups/L1/public/gcp-us-central1-c1/events'
```
jgraettinger, line 9: why does cronut need a special case? you say it doesn't follow the naming convention, but that doesn't seem to matter here?

jgraettinger (approved):
LGTM

supabase/migrations/20250312121737_data-planes-events-column-constraints.sql
psFried: Yeah, I think I'd initially thought to stick with the prior existing naming convention for this data plane (which would have resulted in `ops.us-central1.v1/events-L1`), but then changed my mind. This portion of the migration has already been run, and has already established the cronut events collection name. It's copied into this migration so that it can update the couple of data planes that were created since running the initial migration.

PR #2008, opened 2025-03-13
Dekaf: Add proper field selection/extraction tests, and validate that field names are valid avro

### LINKED ISSUE #2004: dekaf: Reject field names that can't be represented in an AVRO schema
Customer tried a field with a `-` in it, and that blew up generating avro schemas. After talking about this, the most immediate thing we can and clearly should do is make Dekaf's validation logic reject fields like this with a nice error message indicating why and how to add a projection to rename the field

jshearer:
**Description:**
* Add field extraction tests. I had originally written these using the `validate` crate, but this time it's using the logic straight from `build` instead, and is much cleaner
* Add a condition to Dekaf's `validate` implementation that forbids fields with non-avro-compliant names, and returns a helpful error message with a link to our projection docs in case you really want that field.

Fixes https://github.com/estuary/flow/issues/2004

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2008)


jgraettinger (approved):
LGTM

PR #2018, opened 2025-03-19
runtime: Correct handling for calling `dekaf::connector::unary_materialize`

jshearer:
**Description:**

Originally the Dekaf special-case lived in `Runtime::unary_materialize`, but https://github.com/estuary/flow/pull/2001 changed the way unary RPCs are invoked in a way that bypassed that logic. Moving it into `Runtime::serve_materialize` fixes things.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2018)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2325,6 +2325,7 @@ dependencies = [
  "bumpalo",
  "bytes",
  "clap 4.5.21",
+ "coroutines",
  "crypto-common",
  "deadpool",
  "doc",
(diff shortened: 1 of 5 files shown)


jgraettinger:
You don't actually need to spawn it in a tokio task -- those are details that are specific to how sqlite needs to work.

You can instead model off of the [validation noop implementation](https://github.com/estuary/flow/blob/master/crates/validation/src/noop.rs#L110). It only needs to serve Spec and Validate, and can error on anything else.

jgraettinger:
I would get rid of dekaf::connector::unary_materialize() altogether, and implement a dekaf connector stream in terms of `coroutine` that services Spec and Validate, and errors on anything else.

jgraettinger (approved):
LGTM

PR #2022, opened 2025-03-24
agent: controllers activate despite publication failures

psFried:
This addresses a fairly common failure scenario, where controllers have a need to both publish a new version of the spec, and also activate the current version of the spec.  Controllers generally perform any publications before attempting any activations, in order to minimize the number of activations that need to be preformed.

Previously, this meant that a failure to publish would cause the controller to error before it could attempt the activation, meaning that persistent publication failures could completely prevent the activation of task shards.

This updates controllers to attempt activations after a publication failure, so that a failed publication can no longer prevent the activation of the existing build.  This is especially important now that controllers are responsible for restarting failed shards. There are tasks that can run and make progress, even though attempts to publish them will consistently fail.  As of this commit, agents will now restart those shards as expected.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2022)

### DIFF
--- a/crates/agent/src/controllers/capture.rs
+++ b/crates/agent/src/controllers/capture.rs
@@ -15,8 +15,46 @@ pub async fn update<C: ControlPlane>(
     control_plane: &C,
     model: &models::CaptureDef,
 ) -> anyhow::Result<Option<NextRun>> {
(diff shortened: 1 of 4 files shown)


jshearer (approved):
LGTM 👍

PR #2026, opened 2025-03-27
validation: allow missing x-collection-name annotation

psFried:
Eases the validation of materialization connector resource spec schemas, in the specific case where the materialization does not have a `sourceCapture`.

We validate that these annotations exist because they're required in order to automatically add bindings to materialization specs. But not all connectors have a reasonable way to support this. For example, the slack connector requires users to get a channel ID to use as part of the resource config. This has prevented any publications of specs for the materialize-slack connector.

So this commit changes the validation logic to only validate the resource spec schema annotations if the materialization actually defines a `sourceCapture`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2026)

### DIFF
--- a/crates/validation/src/materialization.rs
+++ b/crates/validation/src/materialization.rs
@@ -100,7 +100,7 @@ async fn walk_materialization<C: Connectors>(
     let scope = Scope::new(scope);
 
     let models::MaterializationDef {
-        source_capture: _,
+        source_capture,
         endpoint,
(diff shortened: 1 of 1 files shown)


jshearer (approved):
Like you said in Slack,
>There really isn’t any sensible way for our control plane to automatically add bindings to a slack capture in any case

I agree, this seems like the right thing to do. LGTM 👍

PR #2027, opened 2025-03-27
dekaf: Fix `connector_tags` for Dekaf

jshearer:
**Description:**

This got broken in https://github.com/estuary/flow/pull/2018. It was throwing errors like this:

```
agent::connector_tags: connector Spec RPC failed error=parsing local config

Caused by:
    missing field `variant` at line 1 column 56 image=ghcr.io/estuary/dekaf-generic:v1
```

Now that we don't bypass `Runtime::serve_materialize` for Dekaf anymore, we need to pass a valid `DekafConfig`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2027)


psFried (approved):
LGTM

PR #2038, opened 2025-04-02
Phil/controller backoffs

### LINKED ISSUE #2032: agent: controllers should backoff failed publications
Controllers currently retry any necessary publications and auto-discovers every time they are run. This interacts poorly with controllers being run every time there's a shard failure, particularly while we still have a periodic shard kicker.

A solution is to use a backoff for controller-initiated publications and discovers so that we don't re-attempt them too frequently.

psFried:
**Description:**

Resolves #2032
Adds backoffs to controller-initiated publications and discovers, so that they don't retry them too frequently when the controller is invoked frequently for various different reasons. This ought to dramatically reduce the rate of failed publications and discovers, which are being run due to frequent shard failures, which trigger controller runs.

In all of these, the goal is to still show the controller as being in error status when it's waiting on a backoff, so that failures are still obvious when looking at the overall status.

Also fixes a bug in the runtime, which caused us to fail to set `lastSourcePublishedAt` in materialization stats documents if no documents were read from the source.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2038)

### DIFF
--- a/crates/agent/src/controllers/capture/auto_discover.rs
+++ b/crates/agent/src/controllers/capture/auto_discover.rs
@@ -8,7 +8,7 @@ use models::status::{
 use crate::{
     controllers::{
         publication_status::{self, PendingPublication},
-        ControllerState, NextRun,
+        ControllerErrorExt, ControllerState, NextRun,
     },
     controlplane::ConnectorSpec,
     discovers::DiscoverOutput,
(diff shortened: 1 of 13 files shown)


jgraettinger (approved):
LGTM

PR #2039, opened 2025-04-02
avro: fix encoding of values like `0.0` into schemas expecting integers

jshearer:
**Description:**

JSON schema says [the following about integers](https://json-schema.org/understanding-json-schema/reference/numeric#integer):

> The presence or absence of a decimal point is not enough to distinguish between integers and non-integers. For example, `1` and `1.0` are two ways to represent the same value in JSON. JSON Schema considers that value an integer no matter which representation was used.

As such, we schematize `1.0` as an integer, so we also need to be able to encode it as one.

Originally this was a Dekaf bug, so I also refactored the Dekaf field extraction tests to use `simd-doc`, and add a Dekaf test for this case.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2039)


crates/avro/src/encode.rs
```diff
@@ -97,6 +97,10 @@ fn maybe_encode<'s, 'n, N: AsNode>(
             zig_zag(b, v as i64);
             Ok(true)
         }
+        (Schema::Long, Node::Float(v)) if v.fract() == 0.0 => {
+            zig_zag(b, v as i64);
```
jgraettinger, line 101: please also cover this case in `test_basic_types` below.

jgraettinger (approved):
LGTM % updating `test_basic_types`

crates/avro/src/snapshots/avro__encode__test__basic_types-2.snap
```diff
@@ -14,14 +14,14 @@ expression: to_hex(&b)
...
- |00000401 00000802 02000001 02020261| ...............a 000000e0
- |001e226c 65616469 6e672065 78747261| .."leading extra 000000f0
- |22000001 00010202 65001c22 6d696464| ".......e.."midd 00000100
- |6c652065 78747261 22000200 0102027a| le extra"......z 00000110
- |00202274 7261696c 696e6720 65787472| . "trailing extr 00000120
- |61220000 0002126e 6f742076 616c6964| a".....not valid 00000130
- |01000000 01020e7a 7a202d20 65780016| .......zz - ex.. 00000140
- |5b226578 74726122 2c335d00|          ["extra",3].     00000150
-                                                        0000015c
+ |30303030 30303030 30000214 22726177| 000000000..."raw 000000c0
```
jshearer, line 17: @jgraettinger im not entirely sure how to verify that this snapshot update is correct. The other 2 are exactly what I would expect to see, so I'm pretty sure this one is too, but figured I'd check before hitting merge.

crates/avro/src/encode.rs
```diff
@@ -397,6 +402,7 @@ mod test {
             "m2_no_addl": {"d": true},
             "m2_disallowed_field": {"not valid": true},
             "zz - ex": ["extra", 3],
+            "int_with_zero_frac": serde_json::Value::Number(serde_json::Number::from_f64(1.0).unwrap()),
```
jgraettinger, line 405: nit: i'm pretty sure this can just be 1.0

crates/avro/src/snapshots/avro__encode__test__basic_types-2.snap
jgraettinger: Yep I think it's good.

jgraettinger (approved):

crates/avro/src/encode.rs
jshearer: .. Huh, you're right. I could have sworn I tried that in the Dekaf tests and it didn't trigger the encoding error 🤔 
Anyway, changed!

PR #2042, opened 2025-04-03
dekaf: Skip to the correct offset when we get a `Meta` message

jshearer:
**Description:**

We noticed this when the stats for a Dekaf materialization reading from a collection with expired fragments was reporting dramatically higher stats than it should. The problem is that we were accidentally counting the entire missing fragment(s) as data read, instead of correctly resetting our offset tracker to the beginning of the next fragment.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2042)


jshearer:
Alright, tested and confirmed that this emits the expected stats values now when reading from collections that have missing/expired fragments

jgraettinger (approved):
LGTM

PR #2043, opened 2025-04-03
dekaf: Handle `Suspended` journals in `is_fetch_data_preview`

jshearer:
**Description:**

Even though the existing behavior is doing the right thing, this is currently spamming a bunch of errors in the logs so let's clean it up and handle suspended journals here as well.

**Edit:**

Confirmed that this fixes the logspam

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2043)


psFried (approved):
LGTM

PR #2047, opened 2025-04-04
dekaf: Fix creation of upstream Kafka topics for partitioned journals

jshearer:
**Description:**

We were correctly ensuring that the topics existed upstream, but were failing to create them with the correct number of partitions if the Flow collection had >1 partition.

This both adds logic to fetch and specify the number of partitions when creating topics, it also adds support for increasing the number of partitions if a journal is split while already being read by Dekaf.

**Note:**

<del>This PR also contains a fix for another subtle issue I discovered: log messages emitted immediately before Session closure (such as the log containing the error that caused a Session to crash) didn't seem to be getting delivered. I also hypothesize that some messages could have gotten dropped if multiple were emitted very quickly, but I didn't find evidence of that.

<del>It turns out there was a bug where we were passing the call to `append_logs_to_writer` as one of the futures in a `tokio::select!` call, causing it to get cancelled if another future resolved before it did, such as the message stream closing. This call didn't need to be in the `select` in the first place, so moving it out solved the problem.

Moving to a separate PR

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2047)

### DIFF
--- a/crates/dekaf/src/session.rs
+++ b/crates/dekaf/src/session.rs
@@ -19,6 +19,7 @@ use kafka_protocol::{
     },
     protocol::{buf::ByteBuf, Decodable, Encodable, Message, StrBytes},
 };
+use rustls::crypto::hash::Hash;
 use std::{cmp::max, sync::Arc};
 use std::{
     collections::{hash_map::Entry, HashMap},
@@ -1107,13 +1108,15 @@ impl Session {
         mut req: messages::OffsetCommitRequest,
         header: RequestHeader,
     ) -> anyhow::Result<messages::OffsetCommitResponse> {
-        let collection_partitions = self
-            .fetch_collection_partitions(req.topics.iter().map(|topic| &topic.name))
-            .await?
-            .into_iter()
-            .map(|(topic_name, partitions)| {
-                self.encrypt_topic_name(topic_name)
-                    .map(|encrypted_name| (encrypted_name, partitions))
+        let collections = self
+            .fetch_collections(req.topics.iter().map(|topic| &topic.name))
+            .await?;
+
+        let desired_topic_partitions = collections
+            .iter()
+            .map(|(topic_name, collection)| {
+                self.encrypt_topic_name(topic_name.clone())
+                    .map(|encrypted_name| (encrypted_name, collection.partitions.len()))
             })
             .collect::<Result<Vec<_>, _>>()?;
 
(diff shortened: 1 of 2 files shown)


crates/dekaf/src/session.rs
```diff
@@ -1142,48 +1143,68 @@ impl Session {
...
+                            request_partitions,
+                            collection_partitions
                         ))?
                         .committed_offset;
 
                     metrics::gauge!("dekaf_committed_offset", "group_id"=>req.group_id.to_string(),"journal_name"=>journal_name).set(committed_offset as f64);
                 }
             }
+
+            tracing::info!(topic_name = ?topic.name, partitions = ?topic.partitions, "Committed offset");
```
psFried, line 1207: Would it be helpful to include the actual offsets in this log, maybe just as an array?

crates/dekaf/src/session.rs
```diff
@@ -1103,13 +1104,21 @@ impl Session {
...
     pub async fn offset_commit(
         &mut self,
-        req: messages::OffsetCommitRequest,
+        mut req: messages::OffsetCommitRequest,
         header: RequestHeader,
     ) -> anyhow::Result<messages::OffsetCommitResponse> {
-        let mut mutated_req = req.clone();
-        for topic in &mut mutated_req.topics {
+        let collection_partitions = self
+            .fetch_collection_partitions(req.topics.iter().map(|topic| &topic.name))
```
psFried, line 1111: I noticed that `fetch_collection_partitions` calls `Collection::new`, which is called again on line 1149. And also that `Collection::new` is making a number of API requests. Seems worth trying to avoid the extra requests, if possible.

Also just a nit that you can feel free to ignore, but I personally did not expect a function called `new` to be making network requests. Might be worth renaming and/or refactoring `Collection::new` at some point.

crates/dekaf/src/session.rs
```diff
@@ -1318,6 +1350,30 @@ impl Session {
         }
     }
 
+    async fn fetch_collection_partitions(
```
psFried, line 1353: Now that I'm reading this again, it seems like this function could instead just return the `Collection`s themselves, and let the caller get the number of partitions from them if needed.

williamhbaker (approved):
LGTM

psFried (changes requested):
LGTM % avoiding the extra API calls, if possible.

crates/dekaf/src/session.rs
jshearer: I started out that way and the amount of duplication felt too smelly so I extracted it... but yeah, I think I went a tiny bit too far by throwing away the `Collection`s in favor of their `partitions.len()`. Will update this to return the corresponding Collections

crates/dekaf/src/session.rs
jshearer: I moved the log so it logs per partition -- I don't think this will make logs any noisier, and it's arguably easier to read than parsing an array out of a single log line. Most of the journals have a single partition anyway

crates/dekaf/src/session.rs
jshearer: Refactored this to be `fetch_collections` and then map those into their partition counts etc where it's called

jshearer:
@psFried thanks for the feedback! I believe I addressed it, thoughts? I'll squash the fixup commit before merging, but I left it for easier review

jshearer:
Moving the log stuff to a separate PR

psFried (approved):
LGTM

PR #2051, opened 2025-04-07
reactor: change state key for max load keys

psFried:
Updates the key that's used within materialization state to track the max key that's been observed. We had temporarily disabled the load key optimizations as part of the 0325 indcident recovery, and this effectively makes that disablement permanent for existing tasks, while allowing it to be used for any newly created tasks.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2051)


jgraettinger (approved):
LGTM

PR #2055, opened 2025-04-08
dekaf: Fix log forwarding dropping some messages

jshearer:
**Description:**

I realized we were dropping some log messages on the floor, especially messages that were logged soon before a task exited. I ended up realizing that because the append futures were being directly provided to `tokio::select!`, they will get cancelled if another future in the select resolves first. Per convo w/ @jgraettinger, I refactored the append loop to:

* First check if a new log or stats append can be started. If it can, start it and store it in a `MaybeFuture::Future`
* Poll either/both `MaybeFuture`s if they're not `Done` or `Gone`
* Poll for new messages
* Poll for a tick to send any gathered/rolled up stats since the last tick

Also per convo w/ @psFried, remove log forwarding from the registry. We never ended up setting a task name so no logs got sent, and it's best if we don't allow external data like from HTTP requests to get into the ops logs.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2055)

### DIFF
--- a/crates/dekaf/src/lib.rs
+++ b/crates/dekaf/src/lib.rs
@@ -243,7 +243,8 @@ impl App {
             // We only set this after successfully validating that this task exists and is a Dekaf
             // materialization. Otherwise we will either log auth errors attempting to append to
             // a journal that doesn't exist, or possibly log confusing errors to a different task's logs entirely.
-            logging::get_log_forwarder().set_task_name(username.clone(), labels.build.clone());
+            logging::get_log_forwarder()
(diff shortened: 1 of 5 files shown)


jshearer:
Confirmed logging still works locally and in `dekaf-def`, and that terminal messages are no longer dropped.

psFried (approved):
I'm good with merging this as is, and I don't see any problems. My general impression is that the `start` function could still be made simpler and clearer, and that handling logs that were generated before the shard is known doesn't seem worth the complexity. But I'm comfortable merging as is, and either deferring or abandoning further refactoring.

PR #2063, opened 2025-04-10
data-plane-controller: separate private_links column, restart on change

mdibaiee:
**Description:**

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2063)


crates/data-plane-controller/src/controller.rs
```diff
@@ -64,6 +64,32 @@ impl automations::Executor for Controller {
...
+                    line: "detected change in data plane configuration, restarting deployment"
+                        .to_string(),
+                })
+                .await
+                .context("failed to send to logs sink")?;
+
+            state.status = Status::Idle;
+            state.pending_converge = true;
+
+            return Ok(Outcome {
```
jgraettinger, line 83: I think it's incorrect to transition back to Idle on a config change (as in, could possibly break client interactions with a data-plane). For example, if we're polled while in state AwaitDNS2, then it could short-circuit the DNS await.

This is hand-wavy, but I think we want a mechanism where we more-unconditionally update the internal task state's private links based on the current state of the DB row, without actually changing the polling state of the data plane.

mdibaiee: @jgraettinger do you have any other examples where this would be wrong?

In case of AwaitDNS2, I don't see this as a problem either, since the config has changed, and we are now `pending_converge = true`, which means we will run a Pulumi Up 1 and so on after this.

PR #2068, opened 2025-04-15
Phil/controller next run fix

psFried:
Fixes a few bugs with controllers. See individual commits for more.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2068)

### DIFF
--- a/crates/agent/src/controllers/capture.rs
+++ b/crates/agent/src/controllers/capture.rs
@@ -61,6 +61,9 @@ async fn maybe_publish<C: ControlPlane>(
     control_plane: &C,
     model: &models::CaptureDef,
 ) -> anyhow::Result<bool> {
+    if model.shards.disable {
+        return Ok(false);
+    }
     let mut dependencies = Dependencies::resolve(state, control_plane).await?;
     let published = dependencies
         .update(state, control_plane, &mut status.publications, |deleted| {
--- a/crates/agent/src/controllers/capture.rs
+++ b/crates/agent/src/controllers/capture.rs
@@ -1,7 +1,7 @@
 mod auto_discover;
 use super::{
-    backoff_data_plane_activate, dependencies::Dependencies, ControlPlane, ControllerErrorExt,
-    ControllerState, Inbox, NextRun,
+    backoff_data_plane_activate, coalesce_results, dependencies::Dependencies, ControlPlane,
+    ControllerErrorExt, ControllerState, Inbox, NextRun,
 };
 use crate::controllers::{activation, periodic, publication_status};
 use anyhow::Context;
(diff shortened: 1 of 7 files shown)


crates/agent/src/controllers/capture.rs
```diff
@@ -24,28 +24,35 @@ pub async fn update<C: ControlPlane>(
...
         .and_then(auto_discover::next_run);
     let periodic_next = periodic::next_periodic_publish(state);
-    Ok(NextRun::earliest([
-        ad_next,
-        periodic_next,
-        activate_next_run,
-    ]))
+
+    coalesce_results([
+        published.map(|_| None),
```
mdibaiee, line 50: This makes `Ok(bool)` into `Ok(None)`, right? that's all it does

crates/agent/src/controllers/collection.rs
```diff
@@ -30,13 +30,17 @@ pub async fn update<C: ControlPlane>(
     let activate_next_run =
```
mdibaiee, line 30: It is renamed elsewhere to `activate_result`, should we rename here too?

crates/agent/src/controllers/mod.rs
```diff
@@ -326,6 +327,44 @@ impl NextRun {
...
+    let res = if errs.is_empty() {
+        Ok(min)
+    } else if errs.len() > 1 {
+        Err(anyhow::anyhow!(
+            "{} errors:\n- {}",
+            errs.len(),
+            errs.iter().join("\n- ")
+        ))
+    } else {
+        Err(errs.into_iter().next().unwrap())
```
mdibaiee, line 361:
```suggestion
        Err(errs.first().unwrap())
```

crates/agent/src/controllers/materialization.rs
```diff
@@ -36,16 +36,15 @@ pub async fn update<C: ControlPlane>(
     let activation_next =
```
mdibaiee, line 36: Similarly with the naming

mdibaiee (commented):
LGTM in general just some nits

crates/agent/src/controllers/capture.rs
psFried: Yep, the publish result contains a boolean that says whether it did publish or not, and the next runs are handled by separate functions for each sub-status. Admittedly, this is not the clearest factoring, but it doesn't seem like a good idea to expand the scope of the refactor any more than necessary at this time.

crates/agent/src/controllers/mod.rs
psFried: `first` is only for slices and vecs, and this is just an `IntoIterator`.

mdibaiee: We have `let mut errs = Vec::new()` above, so it is a vector, no?

mdibaiee (approved):

PR #2071, opened 2025-04-15
dekaf: include task name in fetch request metrics

jshearer:
Now that topic names are no longer collection names, we need a way to break down stats by task in order to correctly roll up by tenant, among other things.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2071)

### DIFF
--- a/crates/dekaf/src/read.rs
+++ b/crates/dekaf/src/read.rs
@@ -34,6 +34,9 @@ pub struct Read {
     // Stats are aggregated per collection
     collection_name: String,
 
+    // Include task name in emitted metrics
+    task_name: String,
+
     // Offset before which no documents should be emitted
     offset_start: i64,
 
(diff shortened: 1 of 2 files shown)


crates/dekaf/src/read.rs
```diff
@@ -106,6 +109,12 @@ impl Read {
 
             journal_name: partition.spec.name.clone(),
             collection_name: collection.name.to_owned(),
+            task_name: match auth {
+                SessionAuthentication::Task(task_auth) => task_auth.task_name.clone(),
+                SessionAuthentication::User(user_auth) => {
+                    format!("user-auth: {:?}", user_auth.claims.email)
+                }
```
mdibaiee, line 116: I'm not sure what this means, how would this case be triggered?

jshearer: Yeah... I went back and forth on the right way to deal with this.

`SessionAuthentication::User` is the old way of connecting to Dekaf, using a refresh token generated in the UI. As such, there is no task name. Because of the way `metrics` works, you can't have an optional field, and the logic in `fetch` and `Read` is already at the limit of the complexity I'm comfortable with, and I also didn't want to expand the scope here with a bigger refactor to clean that up, which I plan to do later.

Once everyone has migrated off of the old user-refresh-token based auth it's going to get turned off and the code for it removed, so I went with this path as it was easier and will get deleted fairly soon, as well as giving a nice way to find out exactly which users are still using the old auth strategy so we can nag them to move off of it.

mdibaiee (approved):

PR #2072, opened 2025-04-15
agent: fix skipping auto-discovers for disabled tasks

psFried:
Fixes an issue that causes the agent to rapidly re-run controllers in certain scenarios when a spec is disabled after previously encountering repeated auto-discover failures. This removes the `next_at` whenever the spec is disabled to ensure that we won't keep running the controller in a tight loop.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2072)


jgraettinger (approved):
LGTM

PR #2074, opened 2025-04-16
avro: Fix schematization of `type: null` fields

jshearer:
Previously when inferring an AVRO schema from a shape that solely represented the null type, we incorrectly bailed out with `RawJSON`, where the correct response should have been `avro::Schema::Null`. This corrects `shape_to_avro` to correctly map `{"type": ["null"]}` to `avro::Schema::Null`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2074)


jshearer:
Closing this for now. Talking w/ @jgraettinger, its behavior is intentional -- downstream systems won't know what to do with a purely null type, so it's better to emit it as a type they _do_ know what to do with, if we emit it at all.

PR #2079, opened 2025-04-17
basic control-plane and runtime support for task migration

jgraettinger:
**Description:**

* [required] Handle (in the runtime) a task which is migrated and awakes in a new data-plane, and must discard its stats ACK intent.
* [nice to have] support a new `estuary.dev/cordon` data-plane label, which is attached to ShardSpecs and JournalSpecs to disable them / mark them read-only, and preserved in future activations
* [nice to have] refactor `activate` and add new checks against races and expected ordering
* [nice to have] shard Apply RPC supports setting `primary_hints`, and related support in `gazette` & `activate`
* [nice to have] `migrate` crate implementing a prototype CLI for task and collection migration

Future PRs will include:
* a prototype migration CLI for task migration
* authorization API support for cordoning of tasks to support hot migration without task failure

Issue #2069 

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2079)

### DIFF
--- a/crates/activate/src/lib.rs
+++ b/crates/activate/src/lib.rs
@@ -4,6 +4,7 @@ use proto_flow::flow;
 use proto_gazette::{
     broker::{self, JournalSpec, Label, LabelSelector, LabelSet},
     consumer::{self, ShardSpec},
+    recoverylog,
 };
 use serde_json::json;
 use std::collections::BTreeMap;
@@ -427,6 +428,7 @@ fn unpack_shard_listing(resp: consumer::ListResponse) -> anyhow::Result<Vec<Shar
             id: spec.id,
             labels: set,
             mod_revision: resp.mod_revision,
+            primary_hints: None,
         });
     }
     Ok(v)
@@ -476,6 +478,7 @@ pub fn task_changes<'a>(
         id,
         labels: split,
         mod_revision: shard_revision,
+        primary_hints,
     } in shards
     {
         let template = match template {
@@ -487,6 +490,7 @@ pub fn task_changes<'a>(
                     expect_mod_revision: shard_revision,
                     upsert: None,
                     delete: id,
+                    primary_hints: None,
                 }));
                 continue;
             }
@@ -561,6 +565,7 @@ pub fn task_changes<'a>(
             expect_mod_revision: shard_revision,
             upsert: Some(shard_spec),
             delete: String::new(),
+            primary_hints,
         }));
         changes.push(Change::Journal(broker::apply_request::Change {
             expect_mod_revision: recovery_split.mod_revision,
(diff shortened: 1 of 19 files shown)


crates/activate/src/lib.rs
```diff
@@ -587,6 +568,8 @@ fn task_changes<'a>(
             upsert: Some(recovery_spec),
             delete: String::new(),
         }));
+
+        active = true;
```
psFried, line 577: nit: would it make sense to move this up to line 495? IIUC, it's always true unless we're deleting the shard. Maybe rename to `is_upsert`?

crates/activate/src/snapshots/activate__test__update.snap
```diff
@@ -228,7 +228,7 @@ expression: "(partition_changes, task_changes)"
               },
               {
                 "name": "estuary.dev/logs-journal",
-                "value": "ops/logs/name"
+                "value": "ops/tasks/BASE_NAME/logs/2020202020202020/kind=derivation/name=example%2Fderivation/pivot=00"
```
psFried, line 231: Just want to double check, but do we still have a test covering the case where the existing journal/shard ids don't have the generation id suffix? I thought perhaps this was the snapshot for that

crates/migrate/src/lib.rs
```diff
@@ -0,0 +1,687 @@
...
+    sqlx::query!(
+        r#"
+        SELECT
+            id AS "id: models::Id",
+            catalog_name,
+            spec_type AS "spec_type!: models::CatalogType",
+            built_spec AS "built_spec!: sqlx::types::Json<models::RawValue>"
+        FROM live_specs
+        WHERE starts_with(catalog_name, $1)
+        AND   spec_type IS NOT NULL
```
psFried, line 612: We don't actually set `spec_type` to null when deleting specs anymore, only the `spec` and `built_spec`. So I'm thinking this should change to `spec is not null`, assuming the desire is to filter out deleted specs.

psFried (approved):
LGTM

crates/activate/src/lib.rs
jgraettinger: removed in favor of using an iter().any(matches!(..))

crates/activate/src/snapshots/activate__test__update.snap
jgraettinger: Yes, it's create_and_delete.snap. This is unrelated -- it's replacing an ad-hoc fixture with the correct implied logs partition.

crates/migrate/src/lib.rs
jgraettinger: :+1: updated

Issue #2084, opened 2025-04-18
`billing` edge function not always returning for users


travjenkins:
`support` users are having issues being able to load the billing page. This does not always happen and sometimes if they reload several times they can get through

travjenkins:
This seems that it was not totally related to the edge function - but we took this time to clean some stuff up.

Issue #1243, opened 2023-10-16
Figure out issue with deploying billing edge functions


travjenkins:
It seems that when I am deploying the edge functions from my machine the start up fails but when Joseph deploys it works fine. I am not sure where this should live but since the edge functions are in this repo I am putting this issue here.

The error is this:
```
Log Event Message
module not found file:///Users/js/Documents/estuary/flow/supabase/functions/billing/index.ts
```

Note: 
This is not the end of the world but something we should really figure out before it comes a huge issue.

travjenkins:
This is now working - however there is an issue when trying to run. We are unable to get all the dependencies downloaded.

```
Serving functions on [http://127.0.0.1:5431/functions/v1/<function-name>](http://127.0.0.1:5431/functions/v1/%3Cfunction-name%3E)
Using supabase-edge-runtime-1.67.4 (compatible with Deno v1.45.2)
serving the request with supabase/functions/billing
serving the request with supabase/functions/billing
Download https://esm.sh/v117/math-intrinsics@1.1.0/deno/abs.js
Download https://esm.sh/v117/math-intrinsics@1.1.0/deno/floor.js
Download https://esm.sh/v117/math-intrinsics@1.1.0/deno/max.js
Download https://esm.sh/v117/math-intrinsics@1.1.0/deno/min.js
Download https://esm.sh/v117/math-intrinsics@1.1.0/deno/pow.js
Download https://esm.sh/v117/math-intrinsics@1.1.0/deno/round.js
Download https://esm.sh/v117/math-intrinsics@1.1.0/deno/sign.js
Download https://esm.sh/v117/dunder-proto@1.0.1/deno/get.js
worker boot error: failed to create the graph: Import 'https://esm.sh/v117/math-intrinsics@1.1.0/deno/abs.js' failed: 500 Internal Server Error
    at https://esm.sh/v117/get-intrinsic@1.2.7/deno/get-intrinsic.mjs:10:23
worker boot error: failed to create the graph: Import 'https://esm.sh/v117/math-intrinsics@1.1.0/deno/abs.js' failed: 500 Internal Server Error
    at https://esm.sh/v117/get-intrinsic@1.2.7/deno/get-intrinsic.mjs:10:23
InvalidWorkerCreation: worker boot error: failed to create the graph: Import 'https://esm.sh/v117/math-intrinsics@1.1.0/deno/abs.js' failed: 500 Internal Server Error
[...]

PR #2085, opened 2025-04-18
Some general clean up and version updating for edge functions

travjenkins:
**Description:**

https://github.com/estuary/flow/issues/2084
- Have the DB handle finding the specific tenant the user wants. This will not really change anything but still probably a better approach.

https://github.com/estuary/flow/issues/1243

- Switching off the hardcoded version of 117 for `esm`

Misc Tech Deb

- Using `semver` for some stuff so we are not using a single version. This is recently supports by `esm.sh` https://github.com/esm-dev/esm.sh/pull/875. Only potential issue is it might pull deprecated versions https://github.com/esm-dev/esm.sh/issues/1122

**Workflow steps:**

N/A

N/A

**Notes for reviewers:**

When testing local we need to make sure the `edge function` part of `tilt` can start up but also make sure all the functions can be called properly.

Tilt start up
image

Billing call made
image

Encrypt Config call
image

OAuth Call
image

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2085)


supabase/functions/alerts/alert_types/data_movement_stalled.ts
```diff
@@ -1,5 +1,5 @@
 import { AlertRecord, EmailConfig } from "../index.ts";
-import { isFinite } from "https://cdn.skypack.dev/lodash";
+import { isFinite } from "https://esm.sh/lodash?target=deno";
```
travjenkins, line 2: `skypack` is made specially for browsers. So I think `esm.sh` is probably a bit "safer" of a choice.

.vscode/settings.json
travjenkins, line 1: I don't think we need the `import-map` anymore.

deno.lock
travjenkins, line 1: We really needed this with the import-map stuff. However, we do not have that anymore. In the near future we should probably get this added back in to be safe. However, this is usually generated on an initial running of a `deno` function and since we run these _through_ Supabase I am not entirely sure how we want to generate/manage the lock files.

supabase/config.toml
```diff
@@ -61,11 +61,6 @@ enabled = false
 client_id = ""
 secret = ""
 
-[functions.alert-data-processing]
```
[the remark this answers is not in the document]
travjenkins: Removed: https://github.com/estuary/flow/commit/d148336758663c083e479561cd2b3756de722f36

supabase/functions/_shared/supabaseClient.ts
travjenkins, line 1: We can now use semantic versioning. Probably a bit safer to fetch the latest of the version we want.

mdibaiee (approved):

PR #2089, opened 2025-04-21
support "hot" migration through authorization cordons

jgraettinger:
**Description:**

This PR updates authorization snapshot mechanisms to introduce a concept of scheduled "cordons", where granted authorizations will bound their expiration time to a future scheduled migration timestamp. Upon reaching this timestamp, authorization APIs ask clients to retry periodically until the cordon condition is resolved.

This allows, for example, a running materialization -- which is _not_ migrated -- to gracefully handle a collection which _is_ migrated. Its reads of the collection proceed normally until the cordon timestamp, at which point it spins attempting to fetch a new authorization. Meanwhile, a migration of the collection is performed, and the cordon condition removed. The materialization authorization request can then proceed, and receives an authorization with an updated data-plane.

Similarly, for captures, cordoning ensures that we cannot have races where a non-migrated capture creates a new physical partition in the source data-plane, while the collection is in the process of migration to a different target data-plane.

Other updates to snapshots include:
* Perform ahead-of-time refreshes on reaching the max refresh interval, so we don't stall requests unnecessarily.
* Add extensive new unit test coverage of various authorization APIs.

Finally, add a "watch" mode to the migration tool which monitors for ready migrations and performs them.

Testing:
* In addition to unit test coverage, I performed extensive local testing using two data-planes where I migrated a variety of collections and tasks using snapshot cordons plus the migration tool running in "watch" mode.

**Workflow steps:**

* Enqueue new rows into `data_plane_migrations` to schedule a future migration.
* Run the `migrate watch` tool to monitor for & run ready migrations.
* Observe that collections and/or tasks are now homed in the target data plane.

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2089)

### DIFF
--- a/crates/migrate/src/lib.rs
+++ b/crates/migrate/src/lib.rs
@@ -663,11 +663,25 @@ async fn fetch_data_plane(pg_pool: &sqlx::PgPool, name: &str) -> anyhow::Result<
         )
         .context("failed to sign claims for data-plane")?;
 
+    // Data-plane gateway doesn't support journal suspension,
+    // nor do we want to update it to do so, so require that
+    // cronut is port-forwarded locally.
+    let broker_address = if row.broker_address.contains("svc.cluster.local:") {
+        "http://localhost:8080".to_string()
+    } else {
+        row.broker_address.clone()
+    };
+    let reactor_address = if row.reactor_address.contains("svc.cluster.local:") {
+        "http://localhost:9000".to_string()
+    } else {
+        row.reactor_address.clone()
+    };
+
     // Create the journal and shard clients that are used for interacting with the data plane
     let router = gazette::Router::new("local");
     let journal_client =
-        gazette::journal::Client::new(row.broker_address.clone(), metadata.clone(), router.clone());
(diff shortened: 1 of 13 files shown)


jgraettinger:
Issue #2069

crates/migrate/src/lib.rs
```diff
@@ -75,16 +111,76 @@ pub async fn run(args: Args) -> anyhow::Result<()> {
...
+                &migration.tgt_plane,
+                &migration.catalog_name_or_prefix,
+            )
+            .await
+            .with_context(|| format!("failed to perform migration {}", migration.id))?;
+
+            sqlx::query!(
+                r#"
+                UPDATE data_plane_migrations
+                SET active = false
```
mdibaiee, line 156: So once the migration is done, by setting the migration to inactive, the cordon is gone?

mdibaiee (approved):
I read this at least twice and some parts more, it looks good to me, with the caution that I'm not as knowledgable about this area

jgraettinger:
@jshearer please take a look at dekaf bits but going ahead to merge.

PR #2093, opened 2025-04-22
Dekaf: namespace group IDs and invalidate topic state when binding is backfilled or collection is reset

### LINKED ISSUE #2060: dekaf: Handle reset scenarios (backfill, collection delete/recreate, etc)
* https://github.com/estuary/flow/issues/2083
  Currently, when you change your `token` but keep the same group IDs, the encrypted topic names will fail to decode. We need to swap that logic to prefix the group name with the name of the materialization instead.
* We need to include the binding state key and full collection name including generation counter/build ID in the topic names we map to upstream Kafka in order to support backfills

### LINKED ISSUE #2083: dekaf: Transform group IDs to prevent cross-task conflicts

jshearer:
### Add `advanced.namespaced_ids` to Dekaf endpoint config

This flag will cause topics and groups to be namespaced when reported to/from upstream Kafka. This fixes a few problems described in https://github.com/estuary/flow/issues/2060

Specifically:
* Stop passing `token` to `to_upstream_topic_name`, as the net effect was breaking your consumer group if you changed your token.
* Namespacing group IDs to the materialization so that multiple consumers can share the same group ID without worrying about conflicts
  * **Note:** This doesn't save you from having multiple consumers of the _same_ task step on eachother's toes, like we saw happen last week. This is still a thing that can happen, and I haven't been able to think of a good way to mitigate it. Just don't run multiple consumers of the same task with the same group ID. Tinybird [tells you not to](https://www.tinybird.co/docs/forward/get-data-in/connectors/kafka#create-a-kafka-data-source), and we should probably do the same in our docs.
* Include the partition template name and binding state key in the mapped topic names. This causes backfills and collection resets to propagate to consumers as a fresh topic with no committed offsets.

##### Post-deploy steps:
Setting `advanced.namespaced_ids` to true will cause all of your committed offset state to be effectively reset, so it's set to false by default if not specified. After this change is deployed, we'll want to manually go and publish all Dekaf materializations with this field explicitly set to false. Then we can change its default value to true, and the net effect will be that all _new_ Dekaf materializations will have this turned on by default, leaving existing ones alone.

#### Other changes

We currently have a couple of consumers of materializations where the token changed after the group/topics had offsets committed. This is causing sessions to panic, and while we're actually handling that correctly because they're each an isolated Tokio task, I realized that we are "leaking" `dekaf_total_connections` -- we increment the counter when the connection is established, but panic before we can decrement it. So I refactored the way we report that metric to handle panic'd sessions.

I also fixed the panic by removing a couple of `unwrap()`s which I thought wouldn't fail, but do in this circumstance.

Fixes #2083, fixes #2060

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2093)

### DIFF
--- a/crates/dekaf/src/connector.rs
+++ b/crates/dekaf/src/connector.rs
@@ -47,6 +47,9 @@ pub struct DekafConfig {
     #[serde(default)]
     #[schemars(title = "Strict Topic Names")]
     pub strict_topic_names: bool,
+    #[serde(default)]
+    #[schemars(title = "Advanced")]
+    pub advanced: Advanced,
 }
 
 /// Configures a particular binding in a Dekaf-type materialization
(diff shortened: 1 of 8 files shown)


jshearer:
Obsolete

PR #2100, opened 2025-04-25
data-plane-controller: validate dpc state using jsonschema from ops repo

mdibaiee:
**Description:**

**Workflow steps:**

**Notes for reviewers:**


crates/data-plane-controller/src/controller.rs
```diff
@@ -681,7 +693,34 @@ impl Controller {
...
+        Ok(state)
+    }
+
+    async fn validate_state(&self, state: &State) -> anyhow::Result<()> {
+        let ops_checkout = self
+            .ops_repo
+            .checkout(
+                &self.logs_tx,
+                state.logs_token,
+                "mahdi/data-plane-jsonschema",
```
mdibaiee, line 709: to be replaced after https://github.com/estuary/ops/pull/758 is merged

crates/data-plane-controller/src/controller.rs
```diff
@@ -671,20 +683,51 @@ impl Controller {
             )
             .await?;
 
+        let branch = if remote == self.infra_remote {
+            &state.deploy_branch
+        } else {
+            "mahdi/data-plane-jsonschema"
```
mdibaiee, line 689: To be updated to `master` after https://github.com/estuary/ops/pull/758 is merged

crates/data-plane-controller/src/controller.rs
```diff
@@ -140,6 +143,12 @@ impl Controller {
         };
         let state = state.as_mut().unwrap();
 
+        let mut none = None;
```
jgraettinger, line 146: This is going to do a full clone of the `ops` repo on every single `on_poll`, which happens _frequently_.

I think this needs a refactor to leverage re-use such as `maybe_checkout`. Can you figure out a nice refactor that has _both_ repos living under sub-directories of `maybe_checkout` ?

crates/data-plane-controller/src/controller.rs
```diff
@@ -155,6 +164,8 @@ impl Controller {
             Status::AwaitDNS2 => self.on_await_dns_2(state).await?,
         };
 
+        validate_state(ops_checkout, state).await?;
```
jgraettinger, line 167: Not sure I understand the purpose of doing validation here. Haven't we already done the thing, this step? Shouldn't we just validate the next poll?

I suppose this is trying to catch invalid state transitions being attempted by `on_idle` pulling from `row_state`, but why not do that validation there, closer to where the transition is attempted?

jgraettinger:
I'm also expecting to see significant updates to DPC snapshot tests in this PR.

crates/data-plane-controller/src/controller.rs
mdibaiee: @jgraettinger we validate once before we operate on a state to make sure the data-plane-controller does not operate on a malformed state.

We then validate that after having done whatever step was necessary (pulumi up, ansible, etc.), the updated state as a result of these operations has not become malformed now, in effect preventing a malformed state from being written to `internal.tasks.inner_state`, to avoid having to manually fix malformed states

crates/data-plane-controller/tests/util.rs
```diff
@@ -124,7 +126,7 @@ pub fn initial_state() -> stack::State {
 
     stack::State {
         data_plane_id: models::Id::new([32; 8]),
-        deploy_branch: "main".to_string(),
+        deploy_branch: "test-branch".to_string(),
```
mdibaiee, line 129: Changed this so we can distinguish between git operations on infra repo from ops repo in snapshots

mdibaiee:
@jgraettinger updates:

1. Now there is a `Vec<HashMap<String, tempfile::TempDir>>` in place of the previous `Vec<Option<tempfile::TempDir>>` which allows each run to hold both the checkout for ops and infra repositories in the hashmap (keyed on the remote).
2. Now reports all JSONSchema validation errors in one go instead of only reporting the first error
3. Added a test case specifically for an invalid state

crates/data-plane-controller/tests/snapshots/test_initial_convergence_and_release__test.snap
```diff
@@ -139,9 +273,63 @@ expression: trace.lock().unwrap().as_slice()
...
+      "git clean --force"
+    ]
+  },
+  {
+    "Cmd": [
+      "git-fetch",
+      "git fetch"
+    ]
+  },
+  {
```
jgraettinger, line 309: I ... think this should be fine? But it do still have a latent concern over the sheer number of git operations / fetches in particular that we're doing.

Do you see an opportunity to reduce that at all?

jgraettinger (approved):
LGTM % couple comments

crates/data-plane-controller/tests/snapshots/test_initial_convergence_and_release__test.snap
mdibaiee: One possible way to reduce load of these operations is using `git sparse-checkout` so for ops repository we only checkout the one file we need, and subsequent checkouts will only get updates to that file and not the whole ops repository.

I would prefer keeping the validation before and after all operations to avoid bugs where we do not validate state in specific cases... but perhaps we can make the checkout less intense

mdibaiee: Using `git sparse-checkout` means _more_ commands to run and more complexity in the code, and I don't know how much it's going to help past the first `git clone` that will be slightly cheaper, otherwise subsequent checkouts already are not that expensive

mdibaiee: One other possibility is downloading the file using HTTP GET with a GitHub auth header instead of checking out the repository, that is probably somewhat cheaper compared to a git clone, but subsequent git checkouts are probably not much more expensive either

PR #2102, opened 2025-04-25
dekaf: Add timeout for metadata requests

jshearer:
Metadata requests appear to be sometimes hanging forever. We have an idle session timeout, but if a Session is actively handling a request it won't get killed, and if that request never returns then we'll effectively leak that connection.

We still absolutely need to figure out why exactly metadata requests are hanging, but for the moment this _should_ alleviate the unbounded connection growth issue Dekaf is seeing.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2102)


jshearer:
Rolled up relevant changes in https://github.com/estuary/flow/pull/2109

PR #2108, opened 2025-04-28
assemble: Switch `journal_selector` over to using `name:prefix` instead of `estuary.dev/collection`

jshearer:
**Description:**

This came out of a convo I had with @jgraettinger. We should now be in a state where this change can be deployed, and it should result in some amount of performance improvement for some Gazette RPCs, so let's get it out.

**Edit:** Need to update more snapshots here

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2108)


jgraettinger:
what's up with catalog tests here?

jgraettinger:
@jshearer I looked into this, and we'll need to first land and release https://github.com/estuary/flow/pull/2134 to reactors before we can make this change.

crates/validation/tests/snapshots/scenario_tests__connector_validation_is_skipped_when_shards_are_disabled.snap
```diff
@@ -428,8 +428,265 @@ Outcome {
             spec: MaterializationSpec {
                 name: "acmeCo/testMaterialization",
                 connector_type: Image,
-                config_json: b"{\"image\":\"someMaterializationImage\",\"config\":{\"any\":\"thing\"}}",
-                bindings: [],
+                config_json: "{\"image\":\"someMaterializationImage\",\"config\":{\"any\":\"thing\"}}",
+                bindings: [
```
jgraettinger, line 432: Whoa what's happening here? this snapshot change is not expected.

PR #2109, opened 2025-04-28
Dekaf: Collection of small updates

jshearer:
**Description:**

* Adds a timeout to `Metadata` API calls, as we found out that these were hanging indefinitely, and consumers were just opening more and more connections as a result. I'd like to have a more global timeout, but that's complicated by the fact that some requests are potentially long-running, such as `Fetch` with its `max_wait_ms` parameter. So I went with a more targeted timeout to start with.
* Skip calls to `dekaf::topology::Collection::fetch_spec()` if they're not needed -- this reduces the query load on supabase
* Switch from partition selectors using `estuary.dev/collection` when there is no binding partition selector to `name:prefix`. Combined with https://github.com/estuary/flow/pull/2108 to include `name:prefix` in partition selectors themselves, this will cause all calls to `fetch_partitions` to use `name:prefix`.
  **Note:** AFAIK, the only case where we won't have a binding partition selector is when a connection is using the old user-token auth which we're getting rid of ASAP. So the [change in this PR](https://github.com/estuary/flow/pull/2109/commits/cb484de24028ebaad5ddf44ec44ba95cb9df90e2) is only applicable to sessions using user-token auth, of which there are almost none at this point.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2109)

### DIFF
--- a/crates/dekaf/src/lib.rs
+++ b/crates/dekaf/src/lib.rs
@@ -38,7 +38,11 @@ use log_appender::SESSION_CLIENT_ID_FIELD_MARKER;
 use percent_encoding::{percent_decode_str, utf8_percent_encode};
 use proto_flow::flow;
 use serde::{Deserialize, Serialize};
-use std::{collections::BTreeMap, sync::Arc, time::SystemTime};
+use std::{
+    collections::BTreeMap,
+    sync::Arc,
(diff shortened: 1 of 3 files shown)


psFried (approved):
LGTM

PR #2115, opened 2025-04-29
Phil/validation tweaks tweaks

psFried:
Adds some additional "tweaks" on top of the already massive estuary/flow#2059

A lot of the work here ultimately stemmed from the need to change how connectors work in integration tests, since `NoOpConnectors` no longer works for that.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2115)


psFried:
There's at least one still pending refactor, to separate out model updates from the rest of the validation in `validation/src/materialize.rs`. @jgraettinger are you still up for doing that piece?

jgraettinger (approved):
LGTM - good to merge into validation-tweaks-2 when makes sense to you.

I'll then add a commit to refactor validation materialization as we discussed.

PR #2118, opened 2025-05-01
data-plane-controller: refactor for better test-ability

jgraettinger:
Use type-erased functions to run subcommands and log messages, and hoist dry-run mode into an implementation of a command runner which knows some of the details of controller subcommands (essentially: a lightweight mock).

Refactor Controller::on_* callbacks to be pure functions, with all database interaction driven by controller::Executor prior to entering a Controller::on_poll call.

This facilitates writing unit test loops driven by in-memory state, without DB interaction, using bespoke and test-integrated functions for tracking subcommands or collecting logs.

Add an example test which covers stack initialization and a subsequent rolling release.

Also add data-plane-controller as manually-enabled Tiltfile resource, to ease local integration testing. It works well with est_dry_dock and flowctl pointed at the local-stack database.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2118)


crates/data-plane-controller/tests/util.rs
```diff
@@ -0,0 +1,145 @@
...
+    Box::new(
+        move |_logs_token, stream, message| -> BoxFuture<'static, anyhow::Result<()>> {
+            trace.lock().unwrap().push(TraceEntry::Log(stream, message));
+
+            futures::future::ready(Ok(())).boxed()
+        },
+    )
+}
+
+pub fn mock_run_cmd_fn(trace: Arc<Mutex<Vec<TraceEntry>>>) -> RunCmdFn {
```
mdibaiee, line 31: So to write tests that check behaviour of specific states, we would want to extend this function to accept a custom fixture to use instead of `dry_run_fixture` so each test can customise the state it conveys. It would be cool actually to do a merge patch of `dry_run_fixture` with the input so that each test only applies whatever it specifically is testing instead of writing whole fixture files.

I can look into that in my pull-request, but want to make sure we are aligned on that

mdibaiee (approved):

crates/data-plane-controller/tests/util.rs
jgraettinger: Yea, I have no specific opinion on the nature of `RunCmdFn`'s used in tests and how they're built.

I put this together because I wanted to demonstrate a working and useful garden path leveraging snapshots, and I extracted into `util.rs` on the presumption that we'd want to re-use it, but IMO tests can do whatever they want here.

PR #2122, opened 2025-05-01
avro: Fix encoding of fields where ints are coerced to strings

jshearer:
**Description:**

This is similar to https://github.com/estuary/flow/pull/2039 from a few weeks ago, except this time the schema triggering it is a bit more complex:

```json
{
  "format": "number",
  "type": [
    "integer",
    "string"
  ]
}
```

This schema turns into the AVRO schema
```json
      {
        "type": [
          "string",
          "null"
        ]
      }
```

But similarly to before, a JSON document containing e.g `1.0` will fail to encode with this schema as `1.0` deserialized as a float, which wouldn't match the original schema. So we have to add additional handling for floats with 0-valued fractional components to treat them as ints when encoding string schemas.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2122)


crates/dekaf/tests/snapshots/field_extraction_tests__number_or_string_format_number.snap
```diff
@@ -0,0 +1,176 @@
...
+                                    String(
+                                        "null",
+                                    ),
+                                ),
+                            ],
+                        ),
+                    ),
+                ),
+                (
+                    "my_weird_field",
```
mdibaiee, line 100: this is indeed a weird field but I think this could use a better name to make it easier to follow the snapshots 😁 
perhaps something signifying what the field is actually... `string_int_fmt_number`?

mdibaiee (approved):
LGTM % note on naming of weird field

jgraettinger (approved):

PR #2127, opened 2025-05-02
control plane: only add new tenants to estuary_support/ role

psFried:
Updates the new tenant provisioning to only add newly created tenants to the `estuary_support/` role.  Previously it was adding all tenants to the role whenever any tenant was provisioned, which would re-add tenants that we had removed. Now they stay removed.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2127)


jgraettinger (approved):
LGTM

PR #2131, opened 2025-05-05
dekaf: Move all remaining agant-api calls into `TaskManager`

jshearer:
## Description

This PR build upon https://github.com/estuary/flow/pull/2116 from last week by moving all remaining agent-api calls into the `TaskManager`. The bulk of the remaining calls were being made by the stats and logs component since those require different `APPEND`-authorized data-plane tokens which I didn't get to in the PR last week. Also included are changes addressing the remaining review feedback and some improvements to keep a `TaskManager` around for longer in the face of short-lived sessions.  

### Changes:
* Swap `TaskManagerError(Arc<String>` for a `SharedError(Arc<anyhow::Error>)`
  * Implementing `std::error::Error` for `SharedError` lets Rust's auto-error mapping `?` keyword work correctly
* Rename `TaskManagerResult` to `task_manger::Result`
* Refactor `TaskManager` to scatter-gather state instead of passing around `Arc<Mutex>`es.
* Time out `TaskManager`s after a defined amount of time instead of after a single loop with no listeners
* Expose the cached and authenticated `journal::Client`, and use that in `topology`
  * This should deal with the remaining excess calls to `/authorize/dekaf`
* Update `TaskManager` to fetch and cache `APPEND`-authorized journal clients for ops logs and stats journals
* Update `GazetteAppender` and the surrounding logging logic to use the `TaskManager` provided journal clients instead of fetching fresh ones for each Session.
  * This should deal with the remaining excess calls to `/authorize/task`

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2131)


crates/dekaf/src/task_manager.rs
```diff
@@ -99,23 +104,30 @@ impl TaskStateListener {
 pub struct TaskManager {
     // Key: materialization/task name
     tasks: std::sync::Mutex<
-        HashMap<String, std::sync::Weak<watch::Receiver<Option<TaskManagerResult<TaskState>>>>>,
+        HashMap<
+            String,
+            (
+                // Activity signal to keep the task manager alive
```
jgraettinger, line 110: @nit: I have a sense these mechanics could be simplified, but I'm not taking the time to come up with an actionable proposal, so :shrug: for now.

jgraettinger (approved):
LGTM - a nice improvement

crates/dekaf/src/task_manager.rs
jshearer: Yeah... I agree it feels a bit circular. I'll keep thinking about simpler factorings

PR #2135, opened 2025-05-08
activate: create ops journals when the task is enabled, even without splits

jgraettinger:
This was missed in the previous refactor of `activate`. Dekaf tasks are activated with empty shard splits and an enabled task. We still need to create ops journals for it.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2135)

### DIFF
--- a/crates/activate/src/lib.rs
+++ b/crates/activate/src/lib.rs
@@ -583,7 +583,8 @@ pub fn task_changes<'a>(
         }));
     }
 
-    // Ensure existence of ops partitions iff there's a task shard upsert.
+    // Ensure existence of ops partitions iff there's a task shard upsert,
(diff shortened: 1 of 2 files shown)


jshearer (approved):
Dekaf tasks don't have shards, but _will_ have a `Some` template, so this should work to fix the problem of not creating ops logs/stats journals for Dekaf tasks. 

Beyond that I'm not entirely clear on how this fits into the broader picture WRT cross-dataplane migration etc, but I don't think it can hurt, e.g the worst that happens is this expands the set of circumstances when we create ops logs/stats journals too far and we end up with some superfluous journals.

PR #2137, opened 2025-05-09
Phil/materailization naming

### LINKED ISSUE #2010: models: update `sourceCapture` => `sources` and add `targetNaming`
Discussion in slack: https://estuaryworkspace.slack.com/archives/C03QBN83GQ4/p1740069603378199

Here's the summary:

- We rename sourceCapture => source (via alias)
- We rename targetSchema => targetNaming (via alias)
- We rename SourceCaptureSchemaMode to SourcedTargetNaming
    - variant LeaveEmpty => NoSchema
    - variant FromSourceName => WithSchema
    - NEW variant PrefixSchema

Then:
- /sourceCapture/capture becomes /source/capture (which avoids stuttering)
- /sources can still have a mode where it’s value is just the name of a capture.
- /sourceCapture/targetSchema: leaveEmpty => /source/targetNaming: noSchema
- /sourceCapture/targetSchema: fromSourceName => /source/targetNaming: withSchema
- /sources/targetNaming: prefixSchema (new)
- We would allow /source/capture to be unset. It would be valid to define only sources: {targetNaming: withSchema}. The control plane and UI would both honor targetNaming in all contexts where bindings are being added. (edited) 

edit: we ended up going with `source` instead of `sources`, so I updated the description here to reflect that.

psFried:
**Description:**

Resolves #2010, which has a complete description of the changes to models.

**Workflow steps:**

There's no change to workflows, and existing specs will continue to work with the old names.

This changes the default naming used for adding materialization bindings to the new `targetNaming: prefixNonDefaultSchema`. Given a collection `acmeCo/mySchema/foo`, this would result in a table named `mySchema_foo`. But if the collection were `acmeCo/public/foo`, then the table name would just be `foo`. This matches the previous behavior, which had changed some time ago.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2137)


crates/models/src/materializations.rs
```diff
@@ -16,7 +16,8 @@ use std::collections::BTreeMap;
 pub struct MaterializationDef {
     /// # Automatically materialize new bindings from a named capture
     #[serde(skip_serializing_if = "Option::is_none")]
-    pub source_capture: Option<SourceCapture>,
+    #[serde(alias = "sourceCapture")]
+    pub sources: Option<Sources>,
```
jgraettinger, line 20: this is an annoying change request, but I think it should be `source` instead of `sources`. This seems more consistent with how we pluralize things elsewhere in models, including with the `source` field within a binding.

Then we get:

```
source:
  capture: acmeCo/capture
  targetNaming: noSchema
```

```
source:
  targetNaming: withSchema
```

I think `sources: { capture: acmeCo/capture }` is confusingly inconsistent with `source: acmeCo/collection` under a binding -- I can't explain to a user why they're different -- and `source: {targetNaming: withSchema}` also seems fine.

crates/models/src/source_capture.rs
```diff
@@ -2,64 +2,98 @@ use super::Capture;
...
 
-impl Default for SourceCaptureSchemaMode {
+impl Default for TargetNaming {
     fn default() -> Self {
-        SourceCaptureSchemaMode::LeaveEmpty
+        TargetNaming::PrefixNonDefaultSchema
     }
 }
 
 /// SourceCaptureDef specifies configuration for source captures
```
jgraettinger, line 45: nit; update comment

jgraettinger (approved):
LGTM % comment on `sources` vs `source`

jgraettinger:
Noting for the record that the `sources` naming (instead of `source`) was originally proposed by me in a Slack thread. My bad.

crates/models/src/materializations.rs
jgraettinger: Trying to frame the counter argument / explanation:

`sources` is configuration with respect to _all_ sources of the materialization. `sources: { capture: acmeCo/capture }` is appropriate because that knob affects all materialization bindings and their (singular) `source`.

jgraettinger: > would you still feel the same if it were

I think `source` is more appropriate than `sources` in that context, though `captures` makes sense pluralized.

> AFAIK we'd still like to eventually allow specifying multiple source captures.

I think this is still a matter for debate. I currently lean the other way.

Issue #2095, opened 2025-04-23
`live_spec_flows` has issues being queries by support staff


travjenkins:
When running a query to fetch the `source_id` by matching on `target_id` and narrowing `flow_type` down to just `capture` it will timeout.

psFried:
Looking at the RLS policy on `live_spec_flows`, and I can see how it would be pretty slow:

```
CREATE POLICY "Users must be authorized to one referenced specification" ON public.live_spec_flows FOR SELECT USING ((((source_id)::macaddr8 IN ( SELECT live_specs.id
   FROM public.live_specs)) OR ((target_id)::macaddr8 IN ( SELECT live_specs.id
   FROM public.live_specs))));
```

TBH I'm not seeing an easy way to improve this. I'm thinking we'll need to reach for an API in order to fetch connected specs.

PR #2143, opened 2025-05-12
Phil/status summary again

### LINKED ISSUE #2139: agent-api: `/catalog/status/` response should include a summary
The status endpoint currently returns a pretty hefty and thorough status for each item. There's a fair bit of logic to take all that status information and coalesce it into something that could be used to show status in a list in the UI, where we'd like to display a simple colored indicator with maybe some hover text. So we'd like that logic to live as part of the API, and have the summary just get returned alongside the response.

The summary should be present in every response, and should have a machine-readable enum of statuses:

- `ok`: things appear to be working fine
- `taskDisabled`: the task shards are disabled
- `warning`: the task might not be currently working 100%, but we expect the condition to clear automatically within a reasonable amount of time
- `error`: something is not working right

Additionally, a `message` property should accompany the status, which just has a very brief human-readable description.

psFried:
**Description:**

Resolves #2139 and #2095 

Updates the `/catalog/stats` endpoint to add:

- A `summary` field to every status, which reduces all the available status information into a simple machine-readable enumeration and a short human-readable message
- An optional `short=true` URL parameter, to omit the full `controller_status` and `connector_status` fields, when all you need is the `summary`
- An optional `connected=true` URL parameter to include specs that are directly referenced by any of the specs given by the `name` parameters.

Also simplifies the output of `flowctl catalog status` using the new `summary`, and adds a new `--connected` argument to show the status of connected specs.

**Notes for reviewers:**

I'm feeling pretty confident about the general shape and approach of `summary`, but I expect we'll want to change the specifics of how we summarize the status and, and how we generate the `message`. For example, I expect we'll identify ways to improve the message in specific error scenarios once we see some more real world examples.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2143)

### DIFF
--- a/crates/agent/src/api/public/status.rs
+++ b/crates/agent/src/api/public/status.rs
@@ -1,3 +1,4 @@
+use std::collections::BTreeSet;
 use std::sync::Arc;
 
 use crate::api::error::ApiErrorExt;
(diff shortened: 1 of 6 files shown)


crates/agent/src/api/public/status.rs
```diff
@@ -9,37 +10,81 @@ use axum::{Extension, Json};
...
-use models::Id;
+use models::status::{self, connector::ConnectorStatus, StatusResponse, Summary};
+use models::{CatalogType, Id};
 
 /// Query parameters for the status endpoint
 #[derive(Debug, serde::Deserialize, schemars::JsonSchema)]
 pub struct StatusQuery {
     /// The catalog name of the live spec to get the status of
     pub name: Vec<String>,
+    /// Whether to return a smaller response that excludes the `connector_status` and `controller_status` fields.
```
jgraettinger, line 21: meta: these are great examples of api plumbing that we wouldn't have to do w/ async-graphql

jgraettinger (approved):
LGTM

PR #2152, opened 2025-05-15
dekaf: Forbid selecting null-only locations

### LINKED ISSUE #2082: dekaf: Forbid null-only locations in validation
Downstream systems don't know what to do with null-only locations -- there's no such thing as a "null-typed column" most of the time.

jshearer:
**Description:**

Fixes https://github.com/estuary/flow/issues/2082

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2152)


crates/dekaf/src/connector.rs
```diff
@@ -216,6 +216,16 @@ fn constraint_for_projection(
             r#type: constraint::Type::FieldForbidden.into(),
             reason: "Cannot materialize input data to '_is_deleted' when using CDC deletions mode as it will be generated by Dekaf".to_string(),
         }
+    } else if projection
+        .inference
```
jshearer, line 220: I was a little confused by this being called `inference`, but I'm 95% sure that this field is actually the shape of this location within the collection's effective read schema, at least based on this logic:

https://github.com/estuary/flow/blob/81f5877a64ff2cd5caa34af866fe888f816acb5d/crates/validation/src/collection.rs#L393-L602

mdibaiee: yes this was named inference before we had schema inference as such

mdibaiee (approved):

PR #2153, opened 2025-05-15
dekaf: Update handling for transient errors during session auth handshake

jshearer:
**Description:**

We're still getting complaints about error messages coming from the `DekafError::Unknown` codepath. So instead of bubbling those up, let's try two things:

* First, add a retry with backoff. All of the errors that we've seen here so far are transient, so this should help as long as the client is willing to wait.
* Second, if the error is still present after exhausting the retries, we must disconnect the client. But instead of sending the full error message, we'll just send a generic "Unknown error" code to avoid confusion.

Ideally there'd be some kind of message to send or action we could take that would signal to the consumer that something went wrong but it should retry, but there doesn't appear to be. I tried a bunch of different error codes, as well as just closing the connection immediately.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2153)

### DIFF
--- a/crates/dekaf/src/session.rs
+++ b/crates/dekaf/src/session.rs
@@ -174,29 +174,43 @@ impl Session {
         let authcid = it.next().context("expected SASL authcid")??;
         let password = it.next().context("expected SASL passwd")??;
 
-        let response = match self.app.authenticate(authcid, password).await {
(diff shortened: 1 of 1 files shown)


williamhbaker (approved):
LGTM

PR #2177, opened 2025-05-26
HMAC encrypt

### LINKED ISSUE #2165: dpc: encrypt HMAC keys

mdibaiee:
**Description:**

- HMAC keys are to be stored as a sops-encrypted JSON document in `hmac_keys` column instead of the current array of string format
- We then decrypt the keys in `fetch_loop` of snapshots and keep a cache that is shared between runs of `try_fetch` so long as the data_planes list is the same
- There are two other places where we need to decrypt HMAC keys, in `proxy_connectors` (used for discovers AFAIK) and in `controlplane.build_data_plane_context`
- I have tested snapshot decryption by publishing a task connected to a test data plane in a local flow stack, and seen that task activated (this tests `build_data_plane_context`). I also tested the `proxy_connectors` part by running a discover. I also tested running `flowctl list-shards` against the task on the remote data plane which tests snapshots

Migration plan:
I will manually encrypt all the existing HMAC keys and prepare a single query in which we alter the data type of `hmac_keys` to `String` and then set the new encrypted values. A backup of the previous state of the table will be retained and a query to revert this will also be available in case something goes wrong.

Sister pull-request: https://github.com/estuary/est-dry-dock/pull/116/files

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2177)


jgraettinger:
Looks like you'll also need to update `migrate` to account for HMAC ?

mdibaiee:
Tested the new version again using the same procedure: was able to publish a task to a test data plane through a local UI

crates/agent/src/api/snapshot.rs
```diff
@@ -13,6 +14,8 @@ pub struct Snapshot {
     pub collections_idx_name: Vec<usize>,
     // Platform data-planes.
     pub data_planes: tables::DataPlanes,
+    // Signing HMAC key of data-planes, keyed on data_plane_name
```
jgraettinger: Can we not continue to store these in `tables::DataPlane.hmac_keys` after decryption?

`tables::DataPlanes` implements Clone as well, so you can clone and pass into successive try_fetch just as you do with the HashMap

A nice-to-have would be to wrap tables::DataPlane.hmac_keys in zeroize as well. Not a deal breaker if that's non-trivial. I don't recall why i didn't do that initially, there may have been a challenge

crates/agent/src/api/snapshot.rs
```diff
@@ -451,7 +470,8 @@ async fn try_fetch(pg_pool: &sqlx::PgPool) -> anyhow::Result<Snapshot> {
             d.data_plane_name,
             d.data_plane_fqdn,
             false AS "is_default!: bool",
-            d.hmac_keys,
+            '{}'::text[] as "hmac_keys!",
```
jgraettinger: You should fetch both, and then use `encrypted_hmac_keys` only if `hmac_keys` is empty.

That lets us release this, and then update pulumi to introduce encrypted keys everywhere, and then remove the non-encrypted ones to cut over and verify nothing breaks, with very easy rollback points along the way.

crates/agent/src/controlplane.rs
```diff
@@ -242,13 +246,17 @@ impl<C: DiscoverConnectors + MakeConnectors> PGControlPlane<C> {
         let (ops_logs_template, ops_stats_template) =
             futures::try_join!(ops_logs_template, ops_stats_template)?;
 
+        let decrypted_hmac_keys = crate::decrypt_hmac_keys(&data_plane.encrypted_hmac_keys)
```
jgraettinger: We need these to be cached, we do a very high rate of these per second

Also, same comment applies here about preferring `hmac_keys` and using `encrypted_hmac_keys` only if they're not available. That gives us a very easy way to revert if we see problems during cutover.

Could we share the Snapshot approach, and have PGControlPlane keep a `Mutex<tables::DataPlanes>` ? If the requested `data_plane_id` is already in the table, then we're done. If it's not, then we would request _all_ data-planes, and decrypt hmac keys of (only) new ones.

Or even just ... put a Snapshot here? (perhaps with some flags to skip loads of collections & tasks, as they're chunky and not needed)

crates/agent/src/lib.rs
```diff
@@ -101,6 +102,48 @@ where
     }
 }
 
+#[derive(serde::Deserialize)]
+struct EncryptedHMACKeys {
```
jgraettinger: nit: HMACKeys? they're not encrypted.
Since this is single-use, I'll suggest inline-ing next to its usage.

But, stepping back, perhaps the `decrypt_hmac_keys` should take a &mut tables::DataPlane?

crates/migrate/Cargo.toml
```diff
@@ -10,6 +10,7 @@ license.workspace = true
 
 [dependencies]
 activate = { path = "../activate" }
+agent = { path = "../agent" }
```
jgraettinger: nit: not needed? at least not yet?

crates/tables/src/lib.rs
```diff
@@ -82,6 +82,9 @@ tables!(
         // HMAC-256 keys for this data-plane.
         // The first is used for signing, and any key may validate.
         val hmac_keys: Vec<String>,
+        // HMAC-256 keys for this data-plane in sops-encrypted yaml document format
+        // The first is used for signing, and any key may validate.
+        val encrypted_hmac_keys: String,
```
jgraettinger: nit: `models::RawValue` ? Basically a String, but enforces / clarifies that it's JSON, and is conventional here

supabase/migrations/20250526132206_data_plane_hmac_keys_string.sql
```diff
@@ -0,0 +1,7 @@
+-- Add encrypted_hmac_keys as a text column that holds an encrypted sops document
+
+begin;
+
+ALTER TABLE public.data_planes ADD encrypted_hmac_keys text not null default '';
```
jgraettinger: nit: TEXT => JSON ?

crates/agent/src/api/snapshot.rs
mdibaiee: @jgraettinger if we store them in `hmac_keys`, after we are done with migration and want to remove unencrypted `hmac_keys` from the database table, we will have a discrepency between the `tables::DataPlane` model and the actual table, which I would rather avoid. I think `sqlx` would complain as well.

Basically, it would work for this transition step, but afterwards we would be changing it again 🤷🏽

jgraettinger: > we will have a discrepency between the tables::DataPlane model and the actual table, which I would rather avoid. 

Yea, we have lots of other discrepancies today. Tables --despite its name -- is not tightly coupled to the control plane DB layout, and can roughly be thought of as the data-model "glue" of the control plane. As stuff gets handed off from one component to another, tables are how we pass it around, or record outcomes for debugging. The design is centered around the idea of progressive enrichment (start from raw models, de-structure, add context, build new representations, etc).

In that sense `hmac_keys` vs `encrypted_hmac_keys` actually fits quite well, but lemme know if we should VC to chat through it

crates/agent/src/api/snapshot.rs
```diff
@@ -335,13 +336,14 @@ impl Snapshot {
 
         let validation = jsonwebtoken::Validation::default();
 
-        for hmac_key in &data_plane.hmac_keys {
-            let key = jsonwebtoken::DecodingKey::from_base64_secret(hmac_key)?;
+        for hmac_key in data_plane.hmac_keys.clone() {
```
jgraettinger, line 339: nit: i don't think the clone is required

crates/agent/src/api/snapshot.rs
```diff
@@ -533,6 +540,22 @@ async fn try_fetch(pg_pool: &sqlx::PgPool) -> anyhow::Result<Snapshot> {
         "fetched authorization snapshot",
     );
 
+    let mut cached_keys = decrypted_hmac_keys.keys().collect::<Vec<&String>>();
```
jgraettinger: I think this is work-in-progress --`decrypted_hmac_keys` looks like it's intended to retain / pass-through the decrypted values, but it's being used here only as a set of the data-plane-names fetched in the last snapshot, and values aren't being used (nor is it being added to)?

One call-out just in case is that changes to the encrypted HMAC keys need to be accounted for too: for example, a new data-plane row will appear initially with encrypted HMAC keys as `{}`, and only have keys populated after Pulumi runs, but I don't think Snapshot would know it needs to apply decryption again.

Note that:
- `models::RawValue` implements Eq
- `tables::DataPlanes` implements Clone and has routines to look up by key, if you wanted to pass the prior snapshot value in

crates/agent/src/lib.rs
```diff
@@ -101,6 +102,54 @@ where
...
+    }
+
+    #[derive(serde::Deserialize)]
+    struct HMACKeys {
+        hmac_keys: Vec<String>,
+    }
+
+    // Note that input_output() pre-allocates an output buffer as large as its input buffer,
+    // and our decrypted result will never be larger than its input.
+    let async_process::Output {
```
jgraettinger, line 119: new data-planes have an initial value of `{}`, which I think will error out either during sops decryption (I'm unsure), or when parsing the output

crates/models/src/raw_value.rs
```diff
@@ -75,6 +78,33 @@ impl From<RawValue> for String {
     }
 }
 
+#[cfg(feature = "sqlx-support")]
```
jgraettinger: nit: Phil added handy macro:
```
crate::sqlx_json::sqlx_json!(RawValue);
```

crates/agent/src/main.rs
```diff
@@ -210,13 +210,31 @@ async fn async_main(args: Args) -> Result<(), anyhow::Error> {
...
+        .collect();
+
     let control_plane = agent::PGControlPlane::new(
         pg_pool.clone(),
         system_user_id,
         publisher.clone(),
         id_gen.clone(),
         discover_handler.clone(),
         logs_tx.clone(),
+        decrypted_hmac_keys,
```
mdibaiee, line 230: Turns out it is somewhat more tricky to get this working the same way as it works in Snapshot (i.e. refresh the decrypted hmac_keys when data planes are updated, etc.) since it requires a mutable reference to ControlPlane, which is fine until we hit [here](https://github.com/estuary/flow/blob/master/crates/agent/src/controllers/executor.rs#L30), and we could have `Arc<Mutex<_>>` here but I didn't want to complicate things further by introducing a lock here. I think that would mean LiveSpecControllerExecutors can no longer run in parallel due to the lock...

I'm trying to find a way to refresh the decrypted hmac keys that doesn't lead to a lock 🤔

crates/data-plane-controller/src/controller.rs
```diff
@@ -986,6 +986,50 @@ async fn fetch_releases(
...
+    // and our decrypted result will never be larger than its input.
+    let async_process::Output {
+        stderr,
+        stdout,
+        status,
+    } = async_process::input_output(
+        async_process::Command::new(sops).args([
+            "--encrypt",
+            "--gcp-kms",
+            "projects/estuary-control/locations/us-central1/keyRings/data-plane-controller/cryptoKeys/secrets",
```
jgraettinger: nit: should be a program argument / env

jgraettinger (approved):
LGTM

jgraettinger:
Whoops -- it opened on just the last commit adding encryption, which is what I LGTM'd without realizing the broader review context

crates/agent/src/main.rs
jgraettinger: Does it get easier at all to have it re-use `agent::api::App` ? Definitely a Frankenstein, but direction-ally we do want to move them closer together I think...

crates/agent/src/api/snapshot.rs
mdibaiee: > a new data-plane row will appear initially with encrypted HMAC keys as {}, and only have keys populated after Pulumi runs, but I don't think Snapshot would know it needs to apply decryption again.

I have updated the code now so that such a data-plane will not have a decrypted value in decrypted_hmac_values, so on a next run, we will try again to decrypt its hmac keys until it succeeds

crates/agent/src/api/snapshot.rs
mdibaiee: it is necessary because of the `return Ok(Some(data_plane))` below

crates/data-plane-controller/src/controller.rs
mdibaiee: I updated the code to use the existing `secrets_provider` argument, which points to the pulumi key. Is that okay or do we want a separate key?

jgraettinger: Works for me. That places it in the same authorization domain as Pulumi state secrets, which seems pretty reasonable?

mdibaiee: Works for me too 👍🏽

crates/agent/src/main.rs
mdibaiee: @jgraettinger I implemented a similar cache refreshing mechanism as the `fetch_loop` of Snapshots here

This uses an `RwLock` which distinguishes between reads and writes, so multiple controllers reading the same value won't lock it, but during the write, reads will be blocked (every minute). I think this is a reasonable approach for now

PR #2180, opened 2025-05-28
agent: skip authorization checks for controller-initiated discovers

psFried:
Updates discovers to skip authorization checks for discovers that are performed by controllers.  Previously, we'd be relying on the `estuary_support/` role to allow auto-discovers to properly resolve live specs, which caused auto-discovers to fail for tenants that weren't covered by that role grant.

Also updates the integration test setup to remove the `estuary_support/` role. This was sufficent to reproduce the issue wth auto-discovers, and seemed like a good idea generally.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2180)


jgraettinger (approved):
LGTM

PR #2182, opened 2025-05-29
Phil/activate moar

### LINKED ISSUE #2170: agent: controllers should check whether shards make it to PRIMARY after activation
There's failure modes in the data plane that cause it to transition task shards to FAILED status without ever emitting a `shard failed` log. There's also scenarios where shards can remain in BACKFILL status indefinitely.

We'd like to handle these cases by having controllers periodically check the shard statuses after each activation, and report that the task is not yet healthy until after all shards have a PRIMARY replica. If such a check shows the shards being FAILED, then they should be re-activated periodically. Once the controller sees all shards having a PRIMARY replica it can cease checking, relying on the data plane to emit a `ShardFailed` event if they fail after that point.

### LINKED ISSUE #2171: agent: controllers should periodically health check ops catalog shards
Proper functioning of the ops catalog is critical to ensure that the control plane can respond to task shard failures. This means that currently it's not possible for the control plane to respond to failures of task shards of the ops catalog itself. We currently shard supervisor cron jobs that handle restarting failed ops catalog shards, but we'd like to move that behavior into controllers.

The proposed solution is to have some additional controller behavior for derivations and materializations of the ops catalog, which will:

- periodically check their shard statuses
- restart any failed shards
- fire an alert if there's 3 or more consecutive failures

To identify tasks that are part of the ops catalog, we can check whether the catalog name starts with either the prefix `ops/` or `ops.us-central1.v1/`.

psFried:
**Description:**

Resolves #2170 
Resolves #2171

Introduces some major changes to how controllers keep tabs on task shards. Previously, controllers would always assume that any successful activation would mean that we'd be guaranteed to see a ShardFailed event if it failed again. But this turned out to be something we can't easily guarantee. Further, we can't ever get ShardFailed events for shards of the ops catalog itself, and so we need to rely on another mechanism for monitoring those shards.

The solution in both cases is for controller to periodically do shard listings in order to keep track of the status of task shards.

### Controller changes

Controllers will periodically check the status of task shards. There's a new `shard_status` field in the controller status, which tracks the results of those checks.

Controllers will use an increasingly long duration between shard status checks as long as the status hasn't changed. We want to avoid having controllers running too frequently or sending too many requests to reactors. A change in the observed shard status resets this duration, so that we'll check more frequently right after a change. Also, we use much longer durations when shards are healthy since we should be receiving ShardFailed events if there's a failure after a shard reaches Primary status.

Controllers will attempt to re-activate task shards if:
- any shards are Failed for 3 consecutive checks
- it's been at least 15 minutes (configurable via `FLOW_REACTIVATE_INTERVAL` env variable) since the last time shards were activated
- No `ShardFailed` event was observed

In scenarios where we _do_ get a ShardFailed event, the above logic will not apply, and the current behavior remains in place. This allows us to continue restarting failed shards immediately after the first couple of failures, with an incremental backoff after multiple failures.

In scenarios where we _don't_ get a ShardFailed event, it means that failed shards will be re-activated after 3 consecutive failed status checks (which takes ~2 minutes in total after the first failed check), and at most once every 15 minutes.

### Ops catalog shards

There's a special carve out for ops catalog tasks so that we check their shard statuses much more frequently, and re-start them more quickly if they're failed. Apart from that, ops catalog shards are covered by the same controller behavior that's used for all other tasks. Once this is deployed, we should be able to remove all the shard-supervisor cron jobs, as controllers will be able to handle all shard re-starts.

### Status summaries

I also updated how we summarize status information, so that the summary now accounts for shard failures that were observed via periodic shard listings. The new integration tests use the status summaries pretty extensively, so test coverage of status summaries has increased quite a bit, and even surfaced a few bugs that are fixed here.

**Workflow steps:**

No user-facing changes, except for the output of `flowctl catalog status`, which hopefully just got a bit more helpful and correct.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2182)


crates/agent/src/controllers/activation.rs
```diff
@@ -115,23 +193,263 @@ pub async fn update_activation<C: ControlPlane>(
...
+    status.shard_status = Some(ShardStatusCheck {
+        count,
+        first_ts,
+        last_ts: now,
+        status: new_status,
+    });
+
+    // If there's been at least 3 failed checks in a row, and it's been longer than the
+    // re-activation interval, then re-activate failed shards. We require at least two
+    // failed checks in a row because sometimes shards can temporarily be in failed state
```
jgraettinger: If it's failed, and it's not in a backoff, why wouldn't we restart it immediately?

FAILED shards will never restart themselves, nor should they ever be FAILED due to an ongoing hand-off / re-assignment.

Kept reading -- is this because we'll get confused about a FAILED prior which precedes our current / reset build?

crates/agent/src/controllers/activation.rs
```diff
@@ -115,23 +193,263 @@ pub async fn update_activation<C: ControlPlane>(
...
+    else {
+        anyhow::bail!("internal controller error: attempted to health check a catalog test");
+    };
+
+    let list_response = control_plane
+        .list_task_shards(state.data_plane_id, task_type, state.catalog_name.clone())
+        .await
+        .context("listing task shards")?;
+
+    if list_response.status != proto_gazette::consumer::Status::Ok as i32 {
```
jgraettinger: the gazette client already checks status and maps into an error:
https://github.com/estuary/flow/blob/master/crates/gazette/src/shard/mod.rs#L60

crates/agent/src/controllers/activation.rs
```diff
@@ -115,23 +193,263 @@ pub async fn update_activation<C: ControlPlane>(
...
+    if new_status == ShardsStatus::Failed
+        && time_since_activation >= get_reactivate_interval(state)
+        && count >= 3
+    {
+        // If we've reached this section, then we have _not_ received any
+        // ShardFailed events, which would have caused us to set a `next_retry`.
+        // This can happen if the ops catalog is broken, or in certain edge
+        // cases in the data plane.
+        tracing::warn!(%time_since_activation, failed_checks = %count, "re-activating task shards because they still show as Failed after prior activation");
+        do_activate(now, state, status, control_plane).await?;
```
jgraettinger: discussion: a little hand-wavy, but you could have it send itself a ShardFailed event which would keep a single control flow for how the controller "observes" a failure event. Not asserting that's worth it / makes sense but it's an intuitive recursive framing to me at least. 

The flow in here is complex (but so is the domain), so i'm just casting about for any patterns which might help simplify.

jgraettinger (approved):
LGTM

crates/agent/src/controllers/activation.rs
psFried: Yeah, this is because we'd rather have the normal shard restart mechanism handle these, but there's a race condition where we could observe the failure via a shard listing before the `ShardFailed` event gets materialized. So the idea was to give it a couple of minutes before acting on the this.

jgraettinger: :thinking:  is this improved under the pattern discussed below, where an unexpected FAILED shard in a listing delegates handling by sending a `ShardFailed` event to self? We might even want to literally add it to the `shard_failures` table, given we use that to get a sense of frequency of recent errors?

crates/agent/src/controllers/activation.rs
psFried: I did consider that, but only until it started getting a bit murky. I was uncertain about how to map an observation of Failed shards via listing into `ShardFailed` events, and saw a couple of challenges:

What if the actual `ShardFailed` event just gets materialized after we've observed the failed shards via listing? A slowdown in the ops catalog could result in duplicate `ShardFailed` events, which would result in longer restart backoffs. If there was a good way to deduplicate those ShardFailed events, we'd be all set, but I can't think of one. Definitely open to suggestions here, because I'm with you on wanting to reign in the complexity if we can.

jgraettinger: > What if the actual ShardFailed event just gets materialized after we've observed the failed shards via listing? A slowdown in the ops catalog could result in duplicate ShardFailed events, which would result in longer restart backoffs.

I don't think this is unique to this question -- for example, a reactor deployment will produce extra ShardFailed events just due to the elevated hand-offs.

IMO, creating extra / duplicate `shard_failures` entries due to a (rare, as i understand it) race isn't worth worrying about.

As a future option, we could segment shard_failures over time, and look at how many time segments had zero vs more-than-zero failures

psFried: Then I'll take a crack at it, and see how it goes. Will update here once I try it out

psFried: This worked out pretty well, I think. I added the changes as a new commit instead of re-writing the previous ones, so it's easier to see what changed. LMK what you think.

crates/agent/src/controllers/activation.rs
```diff
@@ -334,6 +294,48 @@ async fn update_shard_health<C: ControlPlane>(
...
+            if Some(state.last_build_id) == shard_build {
+                true
+            } else {
+                tracing::warn!(?shard_build, last_build_id = %state.last_build_id, "shard has mismatched build id");
+                false
+            }
+        })
+        .filter_map(|s| s.spec.as_ref().and_then(|spec| spec.labels.as_ref()))
+        .map(|set| {
+            //let set = s.spec.as_ref().and_then(|spec| spec.labels.as_ref()).unwrap_or(&proto_gazette::LabelSet::default());
```
jgraettinger: nit: commented code

crates/agent/src/controllers/activation.rs
```diff
@@ -334,6 +294,48 @@ async fn update_shard_health<C: ControlPlane>(
...
 
+fn synthesize_shard_failed_events(
+    now: DateTime<Utc>,
+    state: &ControllerState,
+    listing: &[list_response::Shard],
+) -> Vec<ShardFailure> {
+    listing.iter()
+        .filter(|s| s.status.iter().all(|rs| rs.code != (replica_status::Code::Primary as i32)))
+        .filter(|s| {
+            let shard_build = shard_build_id(s);
```
jgraettinger: nit: use `labels::shard::decode_labeling` and then pass through the `proto_flow::ops::ShardLabeling` for building the `models::ShardRef` below?

Using an extra level of scoping for external-crate types would be appreciated -- I thought this was `ops::ShardRef` in `proto_flow` and didn't realize there's a separate `models::ShardRef`.

jgraettinger (approved):
still LGTM % nit

PR #2161, opened 2025-05-19
Revert "Phil/materailization naming"

psFried:
Reverts estuary/flow#2137
Temporarily reverts the materialization naming changes so that we can deploy agent without these changes.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2161)

PR #2183, opened 2025-05-30
Phil/materialization naming again

psFried:
This was originally merged as #2137 , and then reverted by #2161 so that we could deploy other stuff while working on the UI for this. The UI is all ready now, so this is ready to re-merge.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2183)


jgraettinger (approved):
LGTM

Issue #2114, opened 2025-04-29
`edge functions` getting an error (not interrupting service)


travjenkins:
Some of our edge functions keep throwing the error below. Nothing is breaking but we should probably get this fixed.

![Image](https://github.com/user-attachments/assets/e20d8c13-6455-43ae-ad5c-8be9b79e8bbb)

travjenkins:
Supabase support helped when I reached out to them and mentioned this file as a possible cause:
https://github.com/estuary/flow/blob/484df179df4ec5bea35d9a7b861306250e4d7e6a/supabase/functions/billing/index.ts#L7

And to switch over to JSR or NPM imports. This should also help cold start times.

travjenkins:
Probably good to stop using `esm.sh` - https://deno.com/blog/not-using-npm-specifiers-doing-it-wrong

Issue #1412, opened 2024-03-11
Edge function `billing` is using a deprecated API


travjenkins:
We are using a deprecated API in our `billing` edge function. We should probably look into upgrading sometime soon so we are not "stuck" on Deno 1.

```
warning: Use of deprecated "Deno.stdin.rid" API. This API will be removed in Deno 2. Run again with DENO_VERBOSE_WARNINGS=1 to get more details.
```

PR #2198, opened 2025-06-06
Cleaning up some `edge function` tech debt

travjenkins:
**Description:**

completes https://github.com/estuary/flow/issues/2114
completes https://github.com/estuary/flow/issues/1412

Changes:

- Replace all uses of `esm.sh` as this was importing `deno std 1.177`
- Moved some functions to a shared folder and added tests
- Replaced an older SHA256 package with an approach using deno features

**Workflow steps:**

N/A

N/A

**Notes for reviewers:**

oAuth url fetching still working
image

billing calls working
image

edge functions running with no errors
image

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2198)


supabase/functions/_shared/helpers.ts
travjenkins, line 1: These functions did not 100% have to move here - but felt like functions that _could_ be reusable should just live here.

supabase/functions/_shared/helpers.ts
```diff
@@ -41,3 +43,27 @@ export const compileTemplate = (template: string, data: any) => {
 
     return mustacheOutput;
 };
+
+// https://github.com/chiefbiiko/sha256/issues/5#issuecomment-1766746363
+export async function hashStrBase64(str:string):Promise<string> {
+  const msgUint8 = new TextEncoder().encode(str)
+  const hashBuffer = await crypto.subtle.digest('SHA-256', msgUint8)
+  return fromUint8Array(new Uint8Array(hashBuffer)).padEnd(44, '=')
+}
```
travjenkins, line 52: While looking to see if packages were on `jsr` or `npm` I noticed an older package. While looking at the repo I noticed some people calling out performance issues (https://github.com/chiefbiiko/sha256/issues/5) so went ahead and replaced that with an approach the user recommended.

supabase/functions/_shared/supabaseClient.ts
travjenkins, line 1: Wanted to keep the client creations in a single spot to make sure they're on the same version and also always remember we have multiple clients.

supabase/functions/billing/index.ts
```diff
@@ -19,18 +20,15 @@ serve(async (req) => {
             const request = await req.json();
 
             const requested_tenant = request.tenant;
+
+            const authHeader = req.headers.get("Authorization");
+            if (!authHeader) {
+                throw new Error("Missing headers");
+            }
```
travjenkins, line 27: This probably should never happen... but since I was in here I decided to make things a tiny bit safer.

supabase/functions/oauth/auth-url.ts
```diff
@@ -29,7 +13,7 @@ interface OauthSettings {
 export async function authURL(req: { connector_id?: string; connector_config?: OauthSettings; config: object; redirect_uri?: string; state?: object }) {
     const { connector_id, config, redirect_uri, state, connector_config } = req;
 
-    let data: OauthSettings;
+    let data: OauthSettings | null;
```
travjenkins, line 16: The data can technically be null. So marking it as such.

supabase/functions/oauth/auth-url.ts
```diff
@@ -61,8 +45,7 @@ export async function authURL(req: { connector_id?: string; connector_config?: O
 
     // See https://www.oauth.com/oauth2-servers/pkce/authorization-request/
     const codeVerifier = generateUniqueRandomKey(50);
-    const codeChallenge = base64URLSafe(sha256(codeVerifier, "utf8", "base64") as string);
-    const codeChallengeMethod = "S256";
+    const codeChallenge = base64URLSafe(await hashStrBase64(codeVerifier));
```
travjenkins, line 48: Use the new function here. The output should pretty much be the same.

supabase/functions/oauth/auth-url.ts
```diff
@@ -72,7 +55,7 @@ export async function authURL(req: { connector_id?: string; connector_config?: O
             client_id: oauth2_client_id,
             config,
             code_challenge: codeChallenge,
-            code_challenge_method: codeChallengeMethod,
+            code_challenge_method: "S256",
```
travjenkins, line 58: This was hard coded already so not sure what the value of the `const' was.

supabase/functions/tests/shared_test.ts
travjenkins, line 1: This isn't the most helpful test - but it is a good spot to show the difference between the old hashing approach and the new approach. Showing it is _really_ close.

supabase/functions/tests/shared_test.ts
```diff
@@ -0,0 +1,54 @@
...
+      expect(randomNumberHashed).toBe(testData[2].newHash);
+    });
+
+    it("still hashes the same as before but with different separators", () => {
+      expect(randomStringHashed.split("-")).toEqual(testData[0].oldHash.split('+'))
+      expect(randomKeyHashed.split("-")).toEqual(testData[1].oldHash.split('+'))
+      expect(randomNumberHashed.split("_")).toEqual(testData[2].oldHash.split('/'))
+    })
+  });
+});
```
mdibaiee, line 60: am I right that the `base64URLSafe` version of both hashes are equal completely? it seems their difference is just that

travjenkins: Yeah. Once they are ran through the URL replaced they will be exactly the same.

mdibaiee (approved):

supabase/functions/tests/shared_test.ts
travjenkins: Added in a test to make sure this is fully documented.

jshearer (approved):
LGTM

PR #2208, opened 2025-06-09
Update allow-ip-addresses.md

dyaffe:
**Description:**

Updating the IP address labeling to be Europe vs US since that is more important than AWS vs GCP

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2208)


aeluce (approved):
👍

PR #2212, opened 2025-06-10
agent: fix follow up scheduling after new build activation

psFried:
Fixes the `next_run` time after controllers activate a new build. This had a 5m placeholder value mistakenly left in the code, and the fix is to just compute the actual value, which will result in 30s.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2212)


jgraettinger (approved):
LGTM

PR #2213, opened 2025-06-10
agent-api: filter out stale connector status in status response

psFried:
Updates the `/catalog/status` endpoint to filter our stale connector statuses when generating responses.  A stale status is one that has a non-current build id, or that has a timestamp indicating it was written prior to the most recent activation. In other words, we assume that if a connector emits a status at all, that it will do so upon every successful startup of the connector.

Currently, we are showing stale connector statuses in the UI, which is pretty confusing in cases where shards are failed and the connector status still says it's running. I can't think of a reason why we'd ever want to expose stale connector statuses via the `/catalog/status` endpoint, since users can simply look at the task logs for historical statuses. So I think it's better to filter them on the backend.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2213)

### DIFF
--- a/crates/agent/src/api/public/status.rs
+++ b/crates/agent/src/api/public/status.rs
@@ -153,6 +153,15 @@ impl StatusRow {
             controller_status.as_ref(),
             connector_status.as_ref(),
         );
+
+        // Omit the connector status if it's known to be stale.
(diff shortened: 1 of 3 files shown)


jgraettinger (approved):
LGTM

PR #2231, opened 2025-06-18
field selection tweaks & fix local stacks

jgraettinger:
**Description:**

* When `recommended: true`, don't select object leaves to maintain legacy selection behavior in this case
* Add temporary migration handling to populate `groupBy` clause when the current key is non-canonical.
* Streamline encrypted HMAC key handling for "manual" (local) data-planes.

Tested on a local stack by creating two collections, one with a projection of the collection key, the other without.
I confirmed that we added a `groupBy` to only the intended collection, and that thereafter the field selection logging output no differences.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2231)

### DIFF
--- a/crates/agent/src/api/create_data_plane.rs
+++ b/crates/agent/src/api/create_data_plane.rs
@@ -26,8 +26,6 @@ pub struct Manual {
     reactor_address: String,
     /// HMAC keys of the data-plane.
     hmac_keys: Vec<String>,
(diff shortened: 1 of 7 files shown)


psFried (approved):
LGTM

PR #2235, opened 2025-06-19
agent: fix create-data-plane

mdibaiee:
**Description:**

- `sqlx` treats `serde_json::Value` values as `JSONB` by default, and TIL in certain circumstances Postgres cannot automatically convert from `jsonb` to `json` when inserting data (and `sqlx` did not complain about this during compile time checks), so we got:
```
PgDatabaseError { severity: Error, code: "42846", message: "CASE/WHEN could not convert type jsonb to json", detail: None, hint: None, position: Some(Original(1012)), where: None, schema: None, table: None, column: None, data_type: None, constraint: None, file: Some("parse_coerce.c"), line: Some(1548), routine: Some("coerce_to_common_type") }
```

I simplified the code by getting rid of RawValue, and added an explicit cast to `::json` in the SQL query. Tested this locally by calling `create-data-plane`, first able to reproduce the issue and with this patch, able to actually have the row created.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2235)

### DIFF
--- a/crates/agent/src/api/create_data_plane.rs
+++ b/crates/agent/src/api/create_data_plane.rs
@@ -27,7 +27,7 @@ pub struct Manual {
     /// HMAC keys of the data-plane.
     hmac_keys: Vec<String>,
     /// HMAC keys of the data-plane in encrypted sops format
-    encrypted_hmac_keys: models::RawValue,
+    encrypted_hmac_keys: serde_json::Value,
 }
 
 #[derive(Debug, serde::Deserialize, schemars::JsonSchema, Validate)]
@@ -119,7 +119,7 @@ pub async fn create_data_plane(
             format!("https://gazette.{data_plane_fqdn}"),
             format!("https://reactor.{data_plane_fqdn}"),
             Vec::new(),
-            models::RawValue::from_string("{}".to_string()).unwrap(),
+            serde_json::json!({}),
         ),
         Category::Manual(Manual {
             broker_address,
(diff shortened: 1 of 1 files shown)


crates/agent/src/api/create_data_plane.rs
```diff
@@ -197,7 +197,7 @@ pub async fn create_data_plane(
         broker_address,
         reactor_address,
         &hmac_keys,
-        serde_json::to_value(encrypted_hmac_keys).unwrap(),
```
psFried, line 200: We actually implement the `sqlx::{Decode,Encode}` traits for `models::RawValue` so that it can be used directly for `JSON` columns. So technically this could have also been fixed by removing this conversion to `Value`. I see zero reason to change what you have here, but just wanted to point out [the `sqlx_json` call](https://github.com/estuary/flow/blob/a53a7825b824e980389d401e2b7b57f6f5405a94/crates/models/src/raw_value.rs#L81) that provides the trait impls, since we use that on a number of different types in `models`.

psFried (approved):
LGTM

crates/agent/src/api/create_data_plane.rs
mdibaiee: @psFried I tried that, `sqlx` complains `expected Value, instead got RawValue`

Issue #1686, opened 2024-10-07
Expand alerting capabilities


travjenkins:
#### Backend
Do not trigger controllers based on shard failures 
Add triggering for when a backfill happens

#### Resend
Need to get new email templates together
Need to make sure we have the proper account type with Resend

### Notes
Write up -> https://docs.google.com/document/d/1XOxMlug7OtOAogerSnb-j4-dR9xfk6mImNtPHoujFVo/edit

Related UI Issue -> https://github.com/estuary/ui/issues/1244

psFried:
1. We certainly _could_, but we'd need some place for them to store those thresholds, and I'm not sure where. Putting them on the models themselves is the obvious solution, but it might get a bit tedious if you needed to change the thresholds for many things at once. Maybe add a new table to store thresholds by prefix? Also, TBH I really can't see any reason for someone to set a threshold _lower_ than 3. We retry things pretty quickly after the first 2 failures, so a threshold of 3 is unlikely to delay alerts by a significant amount of time, and to go lower would likely result in a lot of false positives. My gut reaction is that we can do without allowing users to customize these thresholds right now, but we can definitely figure it out if you think it's important.
2. It sounds like you're thinking of a configuration knob where the user can opt in to notifications about schema changes, even if they don't break anything. So maybe we have two separate options, one to notify on "incompatible" schema changes, and another to notify on _any_ schema change? Notifying on every schema change could get pretty noisy. Would we want to limit the frequency of those notifications somehow?
3. This seems doable, but also seems pretty separate from the rest of the alerts here, which are focused more on individual tasks.

TBH all three of these feel like somewhat larger tasks, and I'd probably want to treat all of them as separate enhancements that come after the initial alerts.

We could also investigate an easier Slack integration, but I'd also think that should be something that comes afterward.

dyaffe:
Sounds good -- I'm all for a staged roll out for these and even other things you have in this ticket.

dyaffe:
Another idea is an alert for when storage maps are misconfigured

psFried:
@dyaffe @jgraettinger A question came up while I was thinking about alert histories and subscriptions: Should `alert_history` include alerts that nobody is subscribed to?

We'd like to start showing alerts on the details pages in the UI. But should that page still show alerts if nobody is subscribed to them? It seems like they should, yeah? But then what about alert _history_? The initial idea had been to query alert history in order to show the alerts on that page, but currently the history only includes alerts that we've actually sent.

[...]

PR #2238, opened 2025-06-20
control-plane: db migration in support of controller-based alerts

### LINKED ISSUE #2188: Alerts: allow alert subscriptions to exclude alert types
Part of #1686 
Enhance the `alert_subscriptions` table to allow subscriptions to exclude a subset of alert types. Modeling this as an exclusion means that any new alert types will automatically be sent to all subscribers unless/until they opt out.

psFried:
Part of: #1686
Resolves #2188
This is the first part of introducing controller-triggered alerting. There's no user-visible changes here. It's all just setting the stage for alerts that can be triggered by controllers.

- adds a new `controller_alerts` view and integrates it into `alert_all`
- improve performance and simplify the implementations of all existing alert views
- allow `alert_subscriptions` to exclude certain alert types
- fixes some bugs in existing alert views
- Moves the alert notifications API from supabase functions into cloudrun

The `controller_alerts` view exposes any and all alerts triggered by controllers, and should handle all future alert types that are triggered by live spec controllers.  The view extracts alert statuses from the controller status, which is enhanced in a subsequent commit to include alert details.

The performance improvements to the existing views are to allow the `alert_all` view to be evaluated much more frequently. A lot of the performance improvements were realized by moving the joins to `alert_subscriptions` and `auth.users` out of the individual alert views and into the `evaluate_alert_events` function. Thus we only need to join to them twice (once for new alerts, and again for closing alerts) instead of once per alert type. Additionally, we're now only joining to those tables for alerts that are actively transitioning, rather than for all possible alerts.

Also adds the ability for an `alert_subscriptions` row to exclude certain alert types, so that users can opt out of some alert types.

Fixes some bugs in the current alert views, which weren't accounting for the existence of multiple rows in `stripe.customers` for a given tenant. The new views use a group by to handle that.

Move the alert notifications API to cloudrun, and updates the `alert_history` trigger to point to the cloudrun deployment. This also required a few minor code changes in order to get  it working with the latest `deno` version, and to support the health checks performed by cloud run. I also updated this function to check the authorization header before validating the request body.

**Notes for reviewers:**

I'm planning to follow on with a separate PR to implement new alerts in controllers after this is deployed.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2238)


supabase/migrations/20250516163736_controller-alerts.sql
```diff
@@ -0,0 +1,345 @@
...
+        select 1
+        from json_each(json_extract_path(status, 'alerts'))
+    );
+$$;
+comment on function internal.has_controller_alert(json) is
+'Returns true if the controller status JSON contains any alerts,
+ used for the has_alert computed column';
+
+alter table public.controller_jobs
+add column has_alert boolean generated always as (internal.has_controller_alert(status)) stored;
```
jgraettinger: nit: able to inline this single-use function to keep the catalog smaller?

jgraettinger (approved):
LGTM

Perhaps change the `services` directory to something more-specifically about alerting? We have lots of things that could be considered services, and don't intend to move them here.

Also, geez that SQL is dense. After seeing it framed out like this I'm really coming around to your "should we do this in Rust?" question.

psFried:
> Also, geez that SQL is dense. After seeing it framed out like this I'm really coming around to your "should we do this in Rust?" question.

Yeah... I didn't anticipate quite how much work it was going to take in order to make the alert views perform well. I think it was still time well spent, though, and I kinda feel like we might as well just keep going with the approach where we evaluate alert events every minute.

We'll always be able to pretty easily change that up in the future if we want. We could have controllers diff current vs closed alerts and keep `alert_history` up-to-date, and then go back to a much slower evaluation interval for the other sql alerts. But I don't think we have to, especially not in the short term.

psFried:
> Perhaps change the services directory to something more-specifically about alerting? We have lots of things that could be considered services, and don't intend to move them here.

I was actually thinking that we'd eventually move everything from `supabase/functions/` into `services/`. That seem OK?

jgraettinger:
> I was actually thinking that we'd eventually move everything from supabase/functions/ into services/

`edge-functions/` ? `services` is just so general and all-encompassing in scope.

supabase/migrations/20250516163736_controller-alerts.sql
psFried: Yeah. This was a _little_ tricky because there's limitations on what you can do for generated columns, but it seemed to work to use `jsonb_path_exists`, and just cast the status to `jsonb`.

TBH this made me realize that controller statuses should probably have been stored as `jsonb` in the first place, but IDK whether that's worth trying to change at this time.

PR #2241, opened 2025-06-25
agent: new tenants use only the new public dataplane

psFried:
Updates the beta onboarding query to exclude Cronut and the new Cronut replacement from the `data_planes` array for new tenants.  We'd like to reserve the cronut replacement data plane for pre-existing tenants only, so that it doesn't become overloaded.


jgraettinger (approved):
LGTM

PR #2247, opened 2025-06-27
runtime: Allow connectors to tune their inferred schemas

### LINKED ISSUE #2229: schema inference: implement `x-initial-read-schema` and adjustability of complexity limit
When we first implemented capping schema complexity, we were reacting to pain caused by attempting to infer schemas in degenerate cases. Examples of degenerate schema inference scenarios are documents with dynamic keys such as timestamps or usernames, and systems that store truly arbitrary documents at certain locations. The problem is, when we try to infer meaningful schemas over these cases, the resulting schemas both explode in complexity while also not even being useful schemas: a naiive inferred schema over documents with timestamps as keys won't correctly match future documents, for example.

So we introduced logic to prune inferred schemas to keep their "complexity" under a certain limit. Complexity, in this case, just refers to the total count of unique locations within that schema. Originally, we arbitrarily picked a hardcoded complexity limit of 1,000. This struck a good balance of allowing sufficiently complex schemas, while still preventing unbounded degenerate growth.

The problem that we're currently running into is that *some* systems genuinely have >1,000 meaningful locations in their data-model. In addition, a particular failure scenario we recently encountered involves the `_meta` field. In short:

-   When deciding *which* locations to prune, we chose a heuristic that prefers more deeply nested locations over less deeply nested ones. The idea being that fields closer to the root will be more important vs fields in deeply nested structures.
-   Some connectors offer a way to include the full previous document when emitting CDC data. This effectively doubles the number of locations emitted, and also importantly is emitted under the `/_meta/before` key.
-   When certain SQL connectors encounter a keyless table, they fall back to using certain properties under `/_meta/source`
-   If there are more than ~500 columns in a table that both has `/_meta/before` turned on, and depends on `/_meta/source` for its key, the schema complexity limiter will use its heuristic to prune more deeply nested locations and remove key fields from the inferred schema. This then breaks downstream materializations.

### Requirements
So, there are a few things we can/should do:

-   Figure out how to *never* simplify away critical locations like the key fields and probably `/_meta/uuid` etc
-   Retain complexity capping for performance and system integrity protection
-   Understand that different connectors have different requirements: Mongodb will probably want a fairly small/restrictive complexity limit due to its "schema free" nature, while Postgres will probably want a fairly large/permissive complexity limit due to each column assumedly having real meaning that shouldn't be simplified away.
-   Understand that not every application of a particular connector will fit the mold, so having an escape hatch allowing particular tasks to change their own complexity limits (within reason) is probably a good idea

jshearer:
## Description

Modifies `x-infer-schema` to accept a full JSON schema that is merged into the collection's read schema, and adds support for tunable complexity limits via `x-inferred-schema-limit` annotation. 

**Note** that `x-infer-schema: true` behavior is retained for backwards compatibility.

- Changed `x-infer-schema` to optionally accept a whole schema, allowing connectors additional control over their read schemas while still opting into schema inference.
- Added `x-inferred-schema-limit` annotation within `x-infer-schema` to override the default complexity limit of 1000 (capped at 10,000 for safety)
- Updated discovery logic to add the value of `x-infer-schema` to the read schema's `allOf`, if present.
- Added read_schema_json and read_shape to capture Binding to properly extract and apply custom complexity limits
- Updated enforce_shape_complexity_limit() to cap limits at 10,000

The plumbing isn't entirely straightforward, it looks like this:
1. The connector emits a schema containing `x-infer-schema: {type: object, x-inferred-schema-limit: 2000, properties: ...}`
2. The logic in `merge_collections()` sees that and adds that schema into the `read_schema`'s `allOf` using `models::Schema::extended_inferred_read_schema()`
3. Then, at runtime that read schema is loaded in `Binding::new()` where we extract the `read_schema_json` and `read_shape`, at which point `x-inferred-schema-limit` will be present in the `Shape.annotations` field on `read_shape`, if it was present on the original schema from the connector.

## Testing

* Some applicable unit test coverage
* Manual testing:
  * Started a new stack
  * Created a postgres capture from a table with 10 columns:
  ``` sql
  CREATE TABLE test_table (
      col1 text,
      col2 text,
      col3 text,
      col4 text,
      col5 text,
      col6 text,
      col7 text,
      col8 text,
      col9 text,
      col10 jsonb
  );
  ```
  * Edited the discovered specs and manually added the following 3rd item to its read schema `allOf`, simulating what would happen if the connector emitted an object for `x-infer-schema` itself:
  ```yaml
  collections:
    acmeCo/test-cap/public/test_table-3:
      writeSchema: test_table.write.schema.yaml
      readSchema:
        allOf:
          - $ref: "flow://relaxed-write-schema"
          - $ref: "flow://inferred-schema"
          - x-infer-schema: 
            x-inferred-schema-limit: 5
            type: object
  ```
  * Confirmed that inferred schemas generated for that binding only have 5 total fields

**TODO:**
* Test the e2e behavior of a connector emitting an object for `x-infer-schema` containing `x-inferred-schema-limit`. Without this test, I'm not able to completely verify that the original problem of simplifying away important fields is solved.
* I would _love_ to have a full e2e test that covers this entirely flow from a connector emitting `x-infer-schema` with a low complexity cap to the runtime emitting a simplified inferred schema. Maybe a scenario/yaml-based test would work for this 🤔  
* ✅ Validate that `x-infer-schema` is either a bool or a valid JSON schema
* ✅ Validate that `x-inferred-schema-limit` is a number 0<n<10000, if present
* ✅ Document the process for manually overriding `x-infer-schema`. Essentially, you have to get it into the `readSchema`'s annotations one way or another.

---

#### Manual Override

To override the default complexity limit of 1,000 location, include an `x-inferred-schema-limit` annotation within your `x-infer-schema` object:

```yaml
collections:
  acmeCo/my-collection:
    writeSchema: my-schema.yaml
    readSchema:
      allOf:
        - $ref: "flow://relaxed-write-schema"
        - $ref: "flow://inferred-schema"
        - x-infer-schema:
             type: object
             x-inferred-schema-limit: 5000
             properties:
               critical_field:
                type: string
                description: "This field will be preserved even with high complexity"
```

Closes https://github.com/estuary/flow/issues/2229

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2247)


jshearer:
Closing in favor of https://github.com/estuary/flow/pull/2263

PR #2250, opened 2025-06-30
Phil/controller alerts

psFried:
**Description:**

Finishes the implementation of the first two controller-initiated alert types, for shard failures and auto-discover failures.

**Workflow steps:**

The deployment plan is to first deploy agent and the alerts cloudrun function (order does not matter there). Once that's done, the alerts still won't actually fire (be added to `alert_history`) until the migrations are run to add the respective `alert_type` values. We'll be able to check agent logs to verify that the alerts are working correctly before we actually start sending alert emails.


crates/agent/src/controllers/alerts.rs
```diff
@@ -0,0 +1,53 @@
...
+                resolved_at: None,
+                extra: HashMap::new(),
+            },
+        );
+    }
+}
+
+/// Clears the given alert entirely, removing it from `statuses`.
+/// This does not allow for the presence of separate `resolved_arguments`.
+/// If separate `resovlved_arguments` are needed, then we'll have to
```
psFried: I'll fix this comment that trails off. The design intentionally leaves open the possibility that we can have separate `resolved_arguments` by leaving the alert present with a `resolved` state. But the current implementation just removes the alert entirely, since we don't actually need to use separate `resolved_arguments` for now.

crates/agent/src/controllers/activation.rs
```diff
@@ -12,52 +12,73 @@ use models::{
...
-        } else {
-            chrono::Duration::hours(2)
-        }
-    });
+use super::{alerts, backoff_data_plane_activate, executor::Event, ControllerState, NextRun};
+
+/// Helper for getting a `chrono::Duration` from an environment variable, using humantime so it
+/// supports parsing durations like `3h`.
+macro_rules! env_config_interval {
+    ($var_ident:ident, $default_val:expr) => {
```
jgraettinger, line 25: nit (later): this is well done but still a bit of a smell -- it'd be better to thread these through as parameters coming from `clap`, which would handle the environment var mapping.

crates/agent/src/controllers/alerts.rs
jgraettinger: I'm not intuiting what it means to have separate `resolved_arguments` :shrug: I'm going fast here though.

crates/models/src/status/alerts.rs
```diff
@@ -0,0 +1,95 @@
...
+        for alert_type in AlertType::all() {
+            if alert_type.name() == name {
+                return Some(*alert_type);
+            }
+        }
+        None
+    }
+}
+
+impl serde::Serialize for AlertType {
```
jgraettinger, line 43: nit: can we not derive?

crates/models/src/status/alerts.rs
```diff
@@ -0,0 +1,95 @@
...
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        self.name().serialize(serializer)
+    }
+}
+
+impl<'de> serde::Deserialize<'de> for AlertType {
```
jgraettinger, line 52: nit: here too, it's not obvious why we need a customer deserializer

crates/models/src/status/alerts.rs
```diff
@@ -0,0 +1,95 @@
...
+pub enum AlertState {
+    /// The alert is currently firing.
+    Firing,
+    /// The alert has resolved. Resolved alerts may be retained in the status for a short while.
+    Resolved,
+}
+
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize, schemars::JsonSchema)]
+#[serde(rename_all = "snake_case")]
+pub struct ControllerAlert {
```
jgraettinger, line 74: nit: "Controller" here feels like a bit of an implementation detail. `CatalogAlert` or just `Alert` ?

edge-functions/alerts/alert_types/shard_failed.ts
```diff
@@ -0,0 +1,71 @@
...
+import { Recipient } from "../template.ts";
+import { ControllerAlertArguments } from "./controller_alerts.ts";
+
+type ShardFailedRecord = AlertRecord<"shard_failed", ControllerAlertArguments>;
+
+const formatAlertEmail = ({
+  arguments: { recipients, spec_type },
+  catalog_name,
+}: ShardFailedRecord): EmailConfig[] => {
+  const subject = `Estuary Flow: Shard failure detected for ${spec_type} ${catalog_name}`;
```
jgraettinger, line 12: nit: Remove "Shard" and just call it "Failure detected" ? Seems like we can avoid introducing the "shard" terminology here

jgraettinger (approved):
LGTM! Nice 🙌 

Good to merge as is, my feedback is all superficial and easily addressed later.

crates/agent/src/controllers/activation.rs
psFried: agreed :+1:

crates/models/src/status/alerts.rs
psFried: We need to access the actual string values of each variant like `some_alert_type.name()`, and we need the `name` to be consistent with the strings that serde uses to represent the enums. It's easier to have serde delegate to `self.name()` as opposed to having the `name` function somehow get the string value from serde. I'll add a comment.

crates/models/src/status/alerts.rs
psFried: This was intended to add clarity, because there are alerts that can't be represented here (like all the tenant/billing-related alerts). This is only used to represent the subset of alerts that are triggered by controllers. That make sense to you?

jgraettinger: Could that could be addressed (later) by making `spec_type` optional ? Or even an extra arg? That's the only field that's coupled to controllers.

psFried:
> I'm not intuiting what it means to have separate resolved_arguments 🤷 I'm going fast here though.

Just to clarify what's going on here: The design of alerts allows for having different `arguments` when an alert is resolved vs when it starts firing, so that different information can be passed into the alert resolution email. So far, I haven't actually seen a use case for doing that with any of the alerts we're planning to add. I tried to leave open the possibility that we _could_ use that portion of the API if we needed to, but didn't want to add the complexity to the controller code unless it actually became necessary.

crates/models/src/status/alerts.rs
jgraettinger: :+1: gotcha that makes sense. This is avoiding a "whoops" of `name()` somehow mis-matching the serde-derived value.

crates/models/src/status/alerts.rs
psFried: ah, yeah, that might make sense if we want to re-use this type to represent other alert types, like as part of an API response. I'm not certain whether that'll make sense or not, but will keep it in mind as something we could try to re-use.

PR #2262, opened 2025-07-08
network: add HSTS header to frontend HTTP responses

skord:
## Summary
• Add Strict-Transport-Security header to all HTTPS responses from the frontend
• Improves security by preventing protocol downgrade attacks
• Header includes includeSubDomains directive for comprehensive protection

## Test plan
- [x] Verify HSTS header is present in HTTP responses
- [x] Confirm no impact on existing functionality

Closes https://github.com/estuary/security/issues/382

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2262)

### DIFF
--- a/go/network/frontend.go
+++ b/go/network/frontend.go
@@ -393,6 +393,9 @@ func (p *Frontend) serveConnHTTP(user *frontendConn) {
 	var handle = func(w http.ResponseWriter, req *http.Request) {
 		httpStartedCounter.WithLabelValues(task, port, req.Method).Inc()
 
+		// Add HSTS header for all HTTPS responses
+		w.Header().Set("Strict-Transport-Security", "max-age=31536000; includeSubDomains")
+
 		if user.resolved.portIsPublic {
 			reverse.ServeHTTP(w, req)
 		} else if req.URL.Path == "/auth-redirect" {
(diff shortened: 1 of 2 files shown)


skord:
@jgraettinger, reviving this one, and answering your question directly: yes, local stacks continue to function correctly with this change.

I brought up a local data plane on a rebased version of this branch and confirmed both halves:
- the no-/unmatched-SNI reject path (`serveConnErr`) now carries `Strict-Transport-Security`, and
- normal routing is unaffected, so pass-through requests still return `200 OK` and the data-plane services come up healthy.

Two notes:
- The branch was very stale (~1000 commits behind `master`), so I rebased it; it applies cleanly and `go/network` builds + tests green.
- The rebase also surfaced a gap in the original change: HSTS was set only in the `serveConnHTTP` handler, but connections that fail SNI matching are answered by `serveConnErr` and bypassed it, so a probe against the bare endpoint still saw no HSTS. I added the header there too, which is the path that actually matters for an external scan.

Priority is low, since these are TLS API/proxy endpoints rather than browser-facing pages, so a missing HSTS header isn't really protecting anyone the way it would on a web app. The reason to land it is just that it keeps turning up in the monthly scan and the fix is trivial, so it's cheaper to close the finding out than to keep re-triaging it.

jgraettinger (approved):
LGTM

PR #2263, opened 2025-07-08
runtime: add tunable schema complexity limits and `x-initial-read-schema`

### LINKED ISSUE #2229: schema inference: implement `x-initial-read-schema` and adjustability of complexity limit

jshearer:
**Description:**

* Add `x-initial-read-schema` annotation for connector-specified initial read schemas
  * `x-inferred-schema: true` will continue to work
* Implement dynamic complexity limits: for the moment just bump the cap to 10k for SourcedSchema bindings, leaving it at 1k for all others
* Thread complexity limits through `jsonSchemaMerge` reduction annotation via `x-complexity-limit`

### Testing:
* ✅ Capture from connector using SourcedSchema (`source-postgres`) emits `x-complexity-limit: 10000`
Screenshot 2025-07-08 at 15 22 04
* ✅ Capture from connector **not** using SourcedSchema (`source-hello-world`) emits `x-complexity-limit: 1000`
Screenshot 2025-07-08 at 15 25 57
* ✅ `x-complexity-limit` is respected by the reducer https://github.com/estuary/flow/blob/c5cd44b5da79a085bff0693d09b6cc75b3cdeccd/crates/doc/src/reduce/strategy.rs#L1304-L1335
* ✅ `x-initial-read-schema` emitted by a connector during discover gets set as the read schema of the collection
https://github.com/estuary/flow/blob/c5cd44b5da79a085bff0693d09b6cc75b3cdeccd/crates/agent/src/discovers/specs.rs#L736-L748

Closes https://github.com/estuary/flow/issues/2229

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2263)

### DIFF
--- a/crates/agent/src/discovers/specs.rs
+++ b/crates/agent/src/discovers/specs.rs
@@ -1,6 +1,7 @@
 use crate::discovers::Changed;
 
 use anyhow::Context;
+use doc::shape::{X_INFER_SCHEMA, X_INITIAL_READ_SCHEMA};
 use itertools::Itertools;
 use models::{discovers::Changes, ResourcePath};
 use proto_flow::capture::{self, response::discovered};
@@ -318,6 +319,14 @@ pub fn merge_collections(
                 modified = true;
                 draft_model.write_schema = Some(document_schema);
             }
+        } else if let Some(initial_read_schema) = initializes_read_schema(&document_schema) {
+            modified = true;
+            draft_model.write_schema = Some(document_schema);
+            draft_model.schema = None;
+
+            draft_model.read_schema = Some(models::Schema::new(models::RawValue::from_value(
+                &initial_read_schema,
+            )));
         } else if uses_inferred_schema(&document_schema) {
             tracing::debug!(
                 %collection,
@@ -356,9 +365,17 @@ pub fn merge_collections(
 fn uses_inferred_schema(schema: &models::Schema) -> bool {
     matches!(
         // Does the connector use schema inference?
-        schema.to_value().get("x-infer-schema"),
+        schema.to_value().get(X_INFER_SCHEMA),
         Some(serde_json::Value::Bool(true))
     )
+}
+
+fn initializes_read_schema(schema: &models::Schema) -> Option<serde_json::Value> {
+    match schema.to_value().get(X_INITIAL_READ_SCHEMA) {
+        // Does the connector specify an initial read schema
+        Some(extension @ serde_json::Value::Object(_)) => Some(extension.clone()),
+        _ => None,
+    }
 }
 
 /// Returns whether the discovered schema is different from the current schema.
@@ -385,6 +402,7 @@ fn normalize_recommended_name(name: &str) -> String {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use doc::AsNode;
     use proto_flow::capture::{self, response::discovered};
     use serde_json::json;
     use tables::DraftRow;
@@ -525,6 +543,18 @@ mod tests {
                 collection_key: string_vec(&["/fallback", "/key"]),
                 is_fallback_key: true,
                 resource_path: string_vec(&["7"]),
+                disable: false,
+            },
+            // case/8: If there is no fetched collection and x-initial-read-schema is provided, it should be used for readSchema.
+            Binding {
+                target: models::Collection::new("case/8"),
+                document_schema: models::Schema::new(
+                    models::RawValue::from_str(r#"{ "const": "write!", "x-initial-read-schema": {"type": "object", "properties": {"id": {"type": "string"}}} }"#)
+                        .unwrap(),
+                ),
+                collection_key: string_vec(&["/id"]),
+                is_fallback_key: false,
+                resource_path: string_vec(&["8"]),
                 disable: false,
             },
         ];
@@ -703,6 +733,19 @@ mod tests {
                 },
                 is_touch: 0,
             },
+            DraftCollection {
+                collection: case/8,
+                scope: flow://collection/case/8,
+                expect_pub_id: "0000000000000000",
+                model: {
+                  "writeSchema": { "const": "write!", "x-initial-read-schema": {"type": "object", "properties": {"id": {"type": "string"}}} },
+                  "readSchema": {"properties":{"id":{"type":"string"}},"type":"object"},
+                  "key": [
+                    "/id"
+                  ]
+                },
+                is_touch: 0,
+            },
         ]
         "###);
 
@@ -762,6 +805,14 @@ mod tests {
                 ]: Changed {
                     target: Collection(
                         "case/7",
+                    ),
+                    disable: false,
+                },
+                [
+                    "8",
+                ]: Changed {
+                    target: Collection(
+                        "case/8",
                     ),
                     disable: false,
                 },
--- a/crates/doc/src/reduce/strategy.rs
+++ b/crates/doc/src/reduce/strategy.rs
@@ -481,6 +481,7 @@ fn true_value() -> bool {
 mod test {
     use super::super::test::*;
     use super::*;
+    use crate::shape::X_COMPLEXITY_LIMIT;
 
     #[test]
     fn test_append_array() {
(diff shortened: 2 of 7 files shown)


crates/agent/src/discovers/specs.rs
```diff
@@ -356,11 +365,19 @@ pub fn merge_collections(
...
     matches!(
         // Does the connector use schema inference?
-        schema.to_value().get("x-infer-schema"),
+        schema.to_value().get(X_INFER_SCHEMA),
         Some(serde_json::Value::Bool(true))
     )
 }
 
+fn initializes_read_schema(schema: &models::Schema) -> Option<serde_json::Value> {
+    match schema.to_value().get(X_INITIAL_READ_SCHEMA) {
```
jgraettinger, line 374: nit: you could use `remove` here I think, to avoid the clone.

crates/agent/src/discovers/specs.rs
```diff
@@ -318,6 +319,14 @@ pub fn merge_collections(
...
             }
+        } else if let Some(initial_read_schema) = initializes_read_schema(&document_schema) {
+            modified = true;
+            draft_model.write_schema = Some(document_schema);
+            draft_model.schema = None;
+
+            draft_model.read_schema = Some(models::Schema::new(models::RawValue::from_value(
+                &initial_read_schema,
+            )));
         } else if uses_inferred_schema(&document_schema) {
```
jgraettinger, line 330: nit: it'd be nicer to fold `initializes_read_schema` and `uses_inferred_schema` into a single thing.

Presence of `x-infer-schema` is now identically handled to an `x-initial-read-schema` with a particular fixed schema value.

crates/doc/src/reduce/strategy.rs
```diff
@@ -1192,6 +1193,148 @@ mod test {
         )
     }
 
+    #[test]
+    fn test_json_schema_merge_basic() {
```
jgraettinger: test SLOC are debt like any other SLOC. This is covering cases that are already covered (claude?)

crates/doc/src/shape/mod.rs
```diff
@@ -236,6 +236,20 @@ fn impute_property_shape(
     }
 }
 
+/// X_INFER_SCHEMA is a JSON-Schema annotation optionally added to binding schemas
+/// emitted by connectors during discovery. If present, it will be union'd with the
```
jgraettinger, line 240: this is wrong

crates/doc/src/shape/mod.rs
```diff
@@ -236,6 +236,20 @@ fn impute_property_shape(
...
+/// collection's inferred schema ref to form the discovered collection's read schema.
+pub const X_INFER_SCHEMA: &str = "x-infer-schema";
+
+/// X_INITIAL_READ_SCHEMA is a JSON-Schema annotation optionally added to binding schemas
+/// emitted by connectors during discovery. If present, it will be used as the
+/// collection's initial read schema.
+pub const X_INITIAL_READ_SCHEMA: &str = "x-initial-read-schema";
+
+/// X_COMPLEXITY_LIMIT is a JSON-Schema annotation added to emitted inferred schemas that
+/// allows for the modification of the default complexity limit applied to inferred schemas.
```
jgraettinger, line 250: nit: "allows for modification" is a confusing way to put this

crates/runtime/src/capture/protocol.rs
```diff
@@ -332,7 +353,14 @@ pub async fn recv_client_start_commit(
     // produce structured logs of all inferred schemas that have changed
     // in this transaction.
     for binding in txn.updated_inferences.iter() {
-        let serialized = doc::shape::schema::to_schema(shapes[*binding].clone());
+        let mut serialized = doc::shape::schema::to_schema(shapes[*binding].clone());
+
+        let complexity_limit = complexity_limit_for_binding(*binding, bindings_with_sourced_schema);
+
+        serialized.schema.extensions.insert(
```
jgraettinger, line 360: Take a look at how x-collection-generation-id gets initialized in the inferred Shape that we track.

Can you eliminate `bindings_with_sourced_schema` by similarly initializing a value, and then ratcheting it up permanently upon encountering a sourced schema?

crates/validation/src/collection.rs
```diff
@@ -394,6 +395,9 @@ fn walk_collection_schema(
         Error::from(err).push(scope, errors);
     }
 
+    // Validate x-initial-read-schema annotations
```
jgraettinger: Remove this entirely -- we don't need to validate the `x-initial-read-schema` of the collection's write schema. At this point it's entirely inert.

I'm a bit on the fence about whether we should strip it during the discovery merge. I currently think we should keep it as it's advisory and lets our support or the user understand if the recommended initial read schema has changed -- but that's all.

crates/validation/src/errors.rs
```diff
@@ -282,6 +282,14 @@ pub enum Error {
         this_entity: String,
         source_collection: String,
     },
+    #[error("x-infer-schema must be 'true' if present")]
```
jgraettinger: remove this too

jgraettinger (approved):
LGTM!

I have feedback here which I'd like to be addressed, but I have no correctness concerns and think we can merge as-is and fix using a subsequent PR.

crates/doc/src/reduce/strategy.rs
jshearer: 🤦 I searched for `jsonSchemaMerge` and thought there somehow weren't tests for this so I added one (written by Claude, yup)... but apparently I just didn't search wide enough because there totally are covering tests in `doc/src/reduce/schema.rs`. Will delete

crates/doc/src/shape/mod.rs
jshearer: Oops yup. What about:

> If present, it causes the collection's read schema to be set to
> ```json
> {
>   "allOf": [
>     {"$ref": "flow://relaxed-write-schema"},
>     {"$ref": "flow://inferred-schema"}
>   ]
> }

?

crates/validation/src/collection.rs
jshearer: Yeah, I was a bit unsure about this... my goal was to try and make the error messages for connector developers as clear as possible. If we don't explicitly validate `x-initial-read-schema`, it will still get validated by the regular read schema validator, but the error message won't make it clear that the validation error is caused by `x-initial-read-schema`, just that the collection's read schema is wrong.

I'll get rid of it here 👍

jshearer:
Merging now to get this out, will address remaining feedback in a follow-up PR

PR #2276, opened 2025-07-11
Dekaf redirects/migration support

### LINKED ISSUE #2261: dekaf(migrations): Support for automatically redirecting consumers to the correct instance
By design, an instance of Dekaf can only serve materializations hosted in the same dataplane that it's running in. This poses the question: what happens when we migrate a Dekaf materialization to a new dataplane? Currently, it breaks. The solution I've come up with is to (hopefully) seamlessly "redirect" those consumers to the correct broker address.

When a Kafka consumer connects to a broker, one of the first things it does is asks for the list of known broker addresses. Currently Dekaf just advertises itself as the only known broker and moves on with its day. What it _could_ do is instead look up the host dataplane of the task in question (which it fortunately will know at this point in the connection process), and if it's different than the current instance's dataplane, respond with _that_ instance's hostname as the only available broker. 

Ideally then, the consumer would know to start up a connection to that broker instead. It remains to be seen how this plays out in practice -- should the original broker immediately close the connection? Leave it open but error on other messages? Something else?

jshearer:
Use the Kafka metadata API to redirect consumers to the correct Dekaf instance by advertising it as the leader and sole replica node for all topics when authenticating to a Dekaf instance that lives in a different dataplane than the task in question. 

That results in `librdkafka` debug log output like this:

```
%7|1753295815.118|RECV|joseph_testing#consumer-1| [thrd:sasl_ssl://dekaf-dev.estuary-data.com:9092/bootstrap]: sasl_ssl://dekaf-dev.estuary-data.com:9092/bootstrap: Received MetadataResponse (v12, 139 bytes, CorrId 4, rtt 42.87ms)
%7|1753295815.121|METADATA|joseph_testing#consumer-1| [thrd:main]: sasl_ssl://dekaf-dev.estuary-data.com:9092/bootstrap: ===== Received metadata (for 1 requested topics): connected =====
%7|1753295815.121|METADATA|joseph_testing#consumer-1| [thrd:main]: sasl_ssl://dekaf-dev.estuary-data.com:9092/bootstrap: ClusterId: estuary-dekaf, ControllerId: 1
%7|1753295815.121|METADATA|joseph_testing#consumer-1| [thrd:main]: sasl_ssl://dekaf-dev.estuary-data.com:9092/bootstrap: 1 brokers, 1 topics
%7|1753295815.121|METADATA|joseph_testing#consumer-1| [thrd:main]: sasl_ssl://dekaf-dev.estuary-data.com:9092/bootstrap:   Broker #0/1: dekaf.gcp-us-central1-c2.dp.estuary-data.com:9092 NodeId 1
%7|1753295815.122|WAKEUPFD|joseph_testing#consumer-1| [thrd:main]: sasl_ssl://dekaf.gcp-us-central1-c2.dp.estuary-data.com:9092/1: Enabled low-latency ops queue wake-ups
%7|1753295815.122|BROKER|joseph_testing#consumer-1| [thrd:main]: sasl_ssl://dekaf.gcp-us-central1-c2.dp.estuary-data.com:9092/1: Added new broker with NodeId 1
%7|1753295815.122|METADATA|joseph_testing#consumer-1| [thrd:main]: sasl_ssl://dekaf-dev.estuary-data.com:9092/bootstrap:   Topic events with 1 partitions
%7|1753295815.122|STATE|joseph_testing#consumer-1| [thrd:main]: Topic events changed state unknown -> exists
%7|1753295815.122|BRKDELGT|joseph_testing#consumer-1| [thrd:main]: events [0]: delegate to broker sasl_ssl://dekaf.gcp-us-central1-c2.dp.estuary-data.com:9092/1 (rktp 0x13d809800, term 0, ref 3)
%7|1753295815.122|BRKDELGT|joseph_testing#consumer-1| [thrd:main]: events [0]: delegating to broker sasl_ssl://dekaf.gcp-us-central1-c2.dp.estuary-data.com:9092/1 for partition with 0 messages (0 bytes) queued
%7|1753295815.122|BRKMIGR|joseph_testing#consumer-1| [thrd:main]: Migrating topic events [0] 0x13d809800 from (none) to sasl_ssl://dekaf.gcp-us-central1-c2.dp.estuary-data.com:9092/1 (sending PARTITION_JOIN to sasl_ssl://dekaf.gcp-us-central1-c2.dp.estuary-data.com:9092/1)
```

### `/authorize/dekaf` changes:
* https://github.com/estuary/flow/pull/2285
* ~https://github.com/estuary/flow/pull/2292~ 

I decided to deal with the schema registry issue by proxying schema registry requests to their corresponding Dekaf instance, as it was the far simpler solution.

### Issue:
Closes #2261

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2276)

### DIFF
--- a/crates/dekaf/src/lib.rs
+++ b/crates/dekaf/src/lib.rs
@@ -36,8 +36,10 @@ use aes_siv::{aead::Aead, Aes256SivAead, KeyInit, KeySizeUser};
 use flow_client::client::{refresh_authorizations, RefreshToken};
 use log_appender::SESSION_CLIENT_ID_FIELD_MARKER;
 use percent_encoding::{percent_decode_str, utf8_percent_encode};
+use proto_flow::flow::MaterializationSpec;
 use serde::{Deserialize, Serialize};
 use std::{
+    any,
     sync::Arc,
     time::{Duration, SystemTime},
 };
@@ -69,6 +71,7 @@ pub struct DeprecatedConfigOptions {
     pub deletions: connector::DeletionMode,
 }
 
+#[derive(Clone)]
 pub struct UserAuth {
     client: flow_client::Client,
     refresh_token: RefreshToken,
@@ -77,6 +80,7 @@ pub struct UserAuth {
     config: DeprecatedConfigOptions,
 }
 
+#[derive(Clone)]
 pub struct TaskAuth {
     client: flow_client::Client,
     task_name: String,
@@ -87,9 +91,16 @@ pub struct TaskAuth {
     exp: time::OffsetDateTime,
 }
 
+#[derive(Clone)]
 pub enum SessionAuthentication {
     User(UserAuth),
     Task(TaskAuth),
+    Redirect {
+        target_dataplane_fqdn: String,
+        spec: MaterializationSpec,
+        config: connector::DekafConfig,
+        task_state_listener: task_manager::TaskStateListener,
+    },
 }
 
 impl SessionAuthentication {
@@ -99,6 +110,10 @@ impl SessionAuthentication {
                 std::time::UNIX_EPOCH + std::time::Duration::new(user.claims.exp, 0)
             }
             SessionAuthentication::Task(task) => task.exp.into(),
+            SessionAuthentication::Redirect { .. } => {
+                // Redirects are valid for a short duration
+                SystemTime::now() + std::time::Duration::from_secs(60 * 5)
+            }
         }
     }
 
@@ -106,6 +121,9 @@ impl SessionAuthentication {
         match self {
             SessionAuthentication::User(auth) => auth.authenticated_client().await,
             SessionAuthentication::Task(auth) => auth.authenticated_client().await,
+            SessionAuthentication::Redirect { .. } => {
+                anyhow::bail!("Cannot get flow client for redirected task")
+            }
         }
     }
 
@@ -117,6 +135,9 @@ impl SessionAuthentication {
             SessionAuthentication::Task(auth) => {
                 auth.client = auth.client.clone().with_fresh_gazette_client();
             }
+            SessionAuthentication::Redirect { .. } => {
+                unreachable!("This session is a redirect and cannot communicate with Gazette");
+            }
         }
     }
 
@@ -124,6 +145,7 @@ impl SessionAuthentication {
         match self {
             SessionAuthentication::User(user_auth) => user_auth.config.deletions,
             SessionAuthentication::Task(task_auth) => task_auth.config.deletions,
+            SessionAuthentication::Redirect { config, .. } => config.deletions,
         }
     }
 }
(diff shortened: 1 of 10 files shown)


crates/dekaf/src/lib.rs
```diff
@@ -117,13 +134,17 @@ impl SessionAuthentication {
             SessionAuthentication::Task(auth) => {
                 auth.client = auth.client.clone().with_fresh_gazette_client();
             }
+            SessionAuthentication::Redirect { .. } => {
+                unreachable!("This session is a redirect and cannot communicate with Gazette");
```
jshearer, line 139: It should be impossible for these `unreachable!` lines to get executed, and if they do then I consider it an implementation bug to be fixed.

I could instead have changed these functions to return `Result<>` and `bail!` here instead, but I didn't because I think it's better to be noisy. AFAIK no flaky network condition can cause this

crates/dekaf/src/log_appender.rs
```diff
@@ -208,20 +223,32 @@ impl GazetteAppender {
...
+                crate::task_manager::TaskState::Authorized {
+                    ops_stats_client, ..
+                } => ops_stats_client
+                    .map(|(client, _claims)| client)
+                    .map_err(|err| err.into()),
+                crate::task_manager::TaskState::Redirect {
+                    target_dataplane_fqdn,
+                    ..
+                } => {
+                    anyhow::bail!("Task has been redirected to {}", target_dataplane_fqdn);
```
jshearer: We don't have a dataplane access token here, so we're unable to write any logs

psFried: Do we ever call `get_client` if this is a redirect? I'm not seeing how we'd end up in this code path.

crates/dekaf/src/session.rs
```diff
@@ -270,85 +297,112 @@ impl Session {
...
+                        let maybe_collection = Collection::new(
+                            auth,
+                            pg_client,
+                            from_downstream_topic_name(topic.name.to_owned().unwrap_or_default())
+                                .as_str(),
+                        )
+                        .await?;
+                        Ok((topic.name.unwrap_or_default(), maybe_collection))
+                    }))
+                    .await;
```
psFried: kind of a silly nit, by why have the `?` below on line 342 instead of here?

psFried (approved):
LGTM

crates/dekaf/src/session.rs
jshearer: The answer is that moving it seems to cause Rust to be unable to infer the type of the `Ok` on 336: 

```
error[E0283]: type annotations needed
   --> crates/dekaf/src/session.rs:336:25
    |
336 |                         Ok((topic.name.unwrap_or_default(), maybe_collection))
    |                         ^^ cannot infer type of the type parameter `E` declared on the enum `Result`
337 |                     }))
338 |                     .await?;
    |                           - type must be known at this point
    |
    = note: multiple `impl`s satisfying `anyhow::Error: From<_>` found in the following crates: `anyhow`, `core`:
            - impl<E> From<E> for anyhow::Error
              where E: StdError, E: std::marker::Send, E: Sync, E: 'static;
            - impl<T> From<T> for T;
help: consider specifying the generic arguments
    |
336 |                         Ok::<(TopicName, std::option::Option<topology::Collection>), E>((topic.name.unwrap_or_default(), maybe_collection))
```

But rewriting it with `.map` instead of `?` and then returning an `Ok` works just fine, so I'll do that 👍

crates/dekaf/src/log_appender.rs
jshearer: Nope, switched to unreachable

PR #2277, opened 2025-07-11
control-plane: update catalog_stats* table indexes

psFried:
Removes some unneeded indexes, and creates a new index on `ts` in order to facilitate more efficient deletions of old hourly stats. This is intended to resolve statement timeouts in the `delete_old_hourly_stats`, and cleanup some accumulated junk.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2277)


psFried:
Updated to use a BRIN index, and successfully applied the migration ✅

PR #2284, opened 2025-07-16
agent: log in json format by default

psFried:
Changes agent to log in JSON format by default, so that we can more easily use logs in Grafana.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2284)


jshearer (approved):
🪵

PR #2285, opened 2025-07-16
agent-api(`/authorize/dekaf`): Handle redirects when `iss`uing dataplane differs from `sub`ject's host dataplane

jshearer:
**Description:**

When a dekaf materialization is migrated, its host dataplane will change. Correspondingly, Dekaf needs to know about this in order to redirect consumers. As such, we now return enough information to serve that redirect:

* The FQDN of the new host dataplane
* The `MaterializationSpec` so that Dekaf can still advertise the correct topics. Without this, consumers will think there are no topics available.

This is part of https://github.com/estuary/flow/issues/2261, and needed in order to test out the [Dekaf changes](https://github.com/estuary/flow/pull/2276) on `dekaf-dev`.

I tested this locally with a 2-dataplane stack in conjunction with #2276 and it worked wonderfully

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2285)

### DIFF
--- a/crates/agent/src/api/authorize_dekaf.rs
+++ b/crates/agent/src/api/authorize_dekaf.rs
@@ -94,7 +94,7 @@ pub async fn do_authorize_dekaf<'a>(
             evaluate_authorization(snapshot, task_name, shard_data_plane_fqdn, &token)
         },
     ) {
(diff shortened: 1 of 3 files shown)


psFried (approved):
LGTM

PR #2286, opened 2025-07-16
Docs: Add target resource naming options

aeluce:
**Description:**

Includes target resource naming convention options in docs, along with minor tweaks on language around default schema for clarity.

The resource naming section and mirror schemas option can be updated when `source.targetNaming` receives further updates.

Updates the [Materialization concepts page](https://docs.estuary.dev/concepts/materialization/)

**Notes for reviewers:**

Trying to catch up on some docs requests; thanks for the review!

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2286)

### DIFF
--- a/site/docs/concepts/materialization.md
+++ b/site/docs/concepts/materialization.md
@@ -359,33 +359,26 @@ This functionality is only supported for Materialization connectors that have th
 field implemented. Consult the individual connector documentation for details.
 :::
 
-### How It Works
+### Schema Naming Hierarchy
 
-1. **Source Capture Level:**
+There are several chances to set or update schema names for your materialization bindings.
+From lowest priority in the hierarchy to highest priority (overwriting lower-priority settings),
+these are:
 
-   If the source capture provides a schema or namespace, it will be used as the default schema for all bindings in the materialization.
+1. **Materialization Default**
 
-2. **Manual Overrides:**
+   The default schema name can be set at the materialization level in the Endpoint Config, ensuring that all new bindings within that materialization automatically inherit the default schema name.
 
-   You can still manually configure schema names for each binding, overriding the default schema if needed.
+2. **Naming Convention**
 
-3. **Materialization-Level Configuration:**
+   The [target resource naming convention](#target-resource-naming-convention-options) can be set in a materialization's Source Collections section.
+   This option provides general rules for schema and table names for all newly-added bindings.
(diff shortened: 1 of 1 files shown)


site/docs/concepts/materialization.md
```diff
@@ -371,7 +371,7 @@ field implemented. Consult the individual connector documentation for details.
 
 3. **Materialization-Level Configuration:**
 
-   The default schema name can be set at the materialization level, ensuring that all new captures within that materialization automatically inherit the default schema name.
+   The default schema name can be set at the materialization level, ensuring that all new bindings within that materialization automatically inherit the default schema name.
```
jwhartley: In this "How It Works". section, "1. Source Capture Level" is now untrue - we replaced Source Capture naming with Target Resource Naming (we expanded the scope of the naming setting to apply to all bindings, not just bindings from a source Capture).

So I would combine this section with the next one on target naming, with the schema waterfall being materialization default -> target naming (if mirror schemas) -> binding level override.

site/docs/concepts/materialization.md
```diff
@@ -385,4 +385,44 @@ field implemented. Consult the individual connector documentation for details.
...
+|   | Collection: `acmeCo/anvils/orders` | Collection: `acmeCo/public/orders` |
+| - | --- | --- |
+| Prefix schema | Table: `anvils_orders` | Table: `public_orders` |
+|               | Schema: default        | Schema: default        |
+| Prefix non-default schema | Table: `anvils_orders` | Table: `orders` |
+|                           | Schema: default        | Schema: default |
+| Mirror schemas | Table: `orders`  | Table: `orders`  |
+|                | Schema: `anvils` | Schema: `public` |
+| Use table name only | Table: `orders` | Table: `orders` |
+|                     | Schema: default | Schema: default |
```
jwhartley: I find this table a bit difficult to read, I think because of the alternating row colour
Is it possible to change it so that the table+schema for the same option are grouped together, e.g. they're the same colour, or removing row lines between them?

site/docs/concepts/materialization.md
aeluce: Great, thanks for the clarification!

site/docs/concepts/materialization.md
aeluce: How does it look now? I updated it so naming conventions span both table and schema name rows. If it's not clear, I can fiddle with styling or rethink the table altogether. 🤔

PR #2288, opened 2025-07-17
agent: allow dekaf materializations to use source captures

psFried:
Eases a restriction on dekaf materializations, which prevented them from using source captures.  I'm not exactly sure why this restriction was still in place, but my recollection is that it was there because we didn't have a `connector_tags` row for dekaf materializations. We definitely have one now, though.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2288)


jshearer (approved):
LGTM

PR #2292, opened 2025-07-21
agent(`/authorize/dekaf`): Return control-plane `dekaf` token even when redirecting

jshearer:
**Description:**

It turns out that the schema registry doesn't support HTTP redirects, so in order for Dekaf redirects to work, _any_ instance must be able to serve _all_ schema registry requests. This lays the groundwork for that by modifying https://github.com/estuary/flow/pull/2285 so that `/authorize/dekaf` returns the necessary token even when a request is a redirect.

Tested this working end to end with a dual-dataplane local stack.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2292)

### DIFF
--- a/crates/agent/src/api/authorize_dekaf.rs
+++ b/crates/agent/src/api/authorize_dekaf.rs
@@ -112,13 +112,9 @@ pub async fn do_authorize_dekaf<'a>(
                 email: None,
             };
 
-            // Only return a token if we are not redirecting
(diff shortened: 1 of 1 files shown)


psFried (approved):
LGTM

jshearer:
I've decided to deal with the schema registry issue by proxying schema registry requests to their corresponding Dekaf instance, as it was the far simpler solution from the Dekaf side of things.

PR #2294, opened 2025-07-22
flowctl: automatically encrypt endpoint configs

psFried:
**Description:**

Updates flowctl to always and automatically encrypt any unencrypted endpoint configs.  There's no reason to ever use plain text credentials, so there's no option to disable encryption. The encryption happens whenever we upsert to `draft_specs` for any reason, which means that endpoint configs will be encrypted whenever users run any of the following commands:
- `draft author`
- `catalog test`
- `catalog publish`

Encryption will be skipped for any tasks that don't have their connector as part of `connector_tags`.  This is because we currently need to pull the `endpoint_spec_schema` from the `connector_tags` table.  A future improvement can allow flowctl to run the `spec` RPC to get the endpoint spec schemas of connectors that don't have a `connector_tags` row.

The local endpoint configs are never updated as part of this. If users desire to replace their local plain text configs with the encrypted ones, then it's recommended that they run `draft author` followed by `draft develop` with `--overwrite`.

**Workflow steps:**

- Create a `flow.yaml` with at least one capture and/or materialization using an un-encrypted endpoint config
- run `flowctl catalog test|publish` and expect that the `draft_specs` have encrypted endpoint configs
- run `flowctl draft author` and expect that the `draft_specs` should have encrypted the endpoint config
- run `flowctl draft develop --overwrite`, and expect that the local endpoint config has now been encrypted
- In all cases, endpoint configs that are already encrypted should not be modified at all

Some other tests that I ran:

- try the above with both inlined and externalized endpoint configs
- Include a task that uses a connector that _isn't_ in `connector_tags`, and expect that it logs a warning but continues to still author|test|publish

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2294)

### DIFF
--- a/crates/flow-client/src/client.rs
+++ b/crates/flow-client/src/client.rs
@@ -21,6 +21,8 @@ pub struct Client {
     // a single connection pool. The clones can have different headers while
     // still re-using the same connection pool, so this will work across refreshes.
     pg_parent: postgrest::Postgrest,
+    // URL of the config encryption service.
+    config_encryption_url: url::Url,
 }
 
 impl Client {
@@ -30,6 +32,7 @@ impl Client {
         pg_api_token: String,
         pg_url: Url,
         user_access_token: Option<String>,
+        config_encryption_url: Url,
     ) -> Self {
         // Build journal and shard clients with an empty default service address.
         // We'll use their with_endpoint_and_metadata() routines to cheaply clone
@@ -56,6 +59,7 @@ impl Client {
             journal_client,
             shard_client,
             user_access_token,
+            config_encryption_url,
         }
     }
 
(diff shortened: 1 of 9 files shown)


crates/flow-client/src/client.rs
```diff
@@ -167,6 +171,52 @@ impl Client {
...
+        // The encryption service does not currently require any sort of
+        // authentication, so there's no auth header added here. We'll of course
+        // need to update this if we ever add authentiation to that endpoint.
+        let mut req = self
+            .http_client
+            .post(&encrypt_endpoint)
+            .header("Content-Type", "application/json")
+            .json(&request);
+
+        if let Some(token) = &self.user_access_token {
```
jshearer: I thought that the encryption service does not currently require any sort of authentication?

jshearer (approved):
LGTM!

crates/flow-client/src/client.rs
psFried: Whoops, you're right, I removed it.

PR #2295, opened 2025-07-23
dekaf: Permit session reauthentication where the username doesn't change

jshearer:
**Description:**

This is happening because sessions now only live for as long as their access token, which is ~1h. The consumer will try to reauthenticate when a session is about to expire, and we should let it.

I'm thinking this will also resolve a lot of customer complaints about noisy auth logs.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2295)

### DIFF
--- a/crates/dekaf/src/log_appender.rs
+++ b/crates/dekaf/src/log_appender.rs
@@ -364,7 +364,10 @@ impl<W: TaskWriter + Clone + 'static> TaskForwarder<W> {
                 maybe_msg = event_stream.next() => {
                     match maybe_msg {
                         Some(TaskWriterMessage::SetTaskName{name, build}) => {
-                            anyhow::bail!("You can't change the task name after it has already been set ({shard_ref:?} -> ({name}, {build}))");
+                            let shard = dekaf_shard_ref(name.clone(), build.clone());
+                            if shard != shard_ref {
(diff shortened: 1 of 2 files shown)


crates/dekaf/src/log_appender.rs
```diff
@@ -364,7 +364,10 @@ impl<W: TaskWriter + Clone + 'static> TaskForwarder<W> {
                 maybe_msg = event_stream.next() => {
                     match maybe_msg {
                         Some(TaskWriterMessage::SetTaskName{name, build}) => {
-                            anyhow::bail!("You can't change the task name after it has already been set ({shard_ref:?} -> ({name}, {build}))");
+                            let shard = dekaf_shard_ref(name.clone(), build.clone());
+                            if shard != shard_ref {
```
psFried, line 368: Would this be sensitive to changes in the build id? Seems like we'd error if only the build id has changed, but I'm not sure whether that's intentional.

psFried (approved):
LGTM % the one question

crates/dekaf/src/log_appender.rs
jshearer: Yeah, I think that's fine. This errors every hour now anyway, and the session reconnects just fine. I haven't 100% thought through the consequences of the build ID changing in the middle of a session, so I'd rather have it force a new session to be created in that case.

PR #2297, opened 2025-07-23
flowctl: fix default config-encryption url

psFried:
I fat-fingered this URL when I made it a lazy static, so this just removes the extra `s`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2297)


Alex-Bair (approved):
LGTM

PR #2299, opened 2025-07-25
dekaf: Don't transform schemas into canonical form in `/subjects/:subject/versions/latest`

jshearer:
**Description:**

This got missed when we did https://github.com/estuary/flow/pull/1956, and an integration partner just pointed it out as a discrepancy when investigating an issue.

Currently, `/schemas/ids/:id` returns the full-fat schema whereas `/subjects/:subject/versions/latest` returns the canonical form which is missing a lot of data.

Tested this locally with kcat and on `dekaf-dev`

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2299)

### DIFF
--- a/crates/dekaf/src/registry.rs
+++ b/crates/dekaf/src/registry.rs
@@ -107,9 +107,12 @@ async fn get_subject_latest(
             (value_id, &collection.value_schema)
         };
 
+        let serialized_schema =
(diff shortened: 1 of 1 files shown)


psFried (approved):
LGTM

PR #2300, opened 2025-07-25
dekaf: Entirely remove user refresh token auth

jshearer:
Old MSK is off, nobody can use this flow anymore, let's delete some code!

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2300)


crates/dekaf/src/topology.rs
```diff
@@ -119,57 +92,43 @@ impl Collection {
...
-        pg_client: &postgrest::Postgrest,
         topic_name: &str,
     ) -> anyhow::Result<Option<Self>> {
         let binding = match auth {
             SessionAuthentication::Task(task_auth) => {
                 if let Some(binding) = task_auth.get_binding_for_topic(topic_name).await? {
                     Some(binding)
                 } else {
-                    bail!("{topic_name} is not a binding of {}", task_auth.task_name)
+                    tracing::warn!("{topic_name} is not a binding of {}", task_auth.task_name);
```
jshearer, line 102: This is a correctness change I noticed -- we used to close the session if a binding for the requested topic name wasn't found, but the intended behavior is to return `UnknownTopicOrPartition`, which is what happens when we return `None` here. 

We have a bunch of consumers asking for topics that are no longer bindings of materializations, and I'm hoping this will better signal to them that they're never going to succeed.

crates/dekaf/src/topology.rs
```diff
@@ -247,7 +206,8 @@ impl Collection {
                 auth.deletions(),
             )?
         } else {
-            utils::build_LEGACY_field_extractors(collection_schema_shape.clone(), auth.deletions())?
+            // For non-task authentication, use default field extraction
```
psFried: I don't understand this comment. Wouldn't returning None result in `UnknownTopicOrPartition`?

psFried (approved):
LGTM

crates/dekaf/src/topology.rs
jshearer: Yeah, and actually `binding` doesn't need to be `Option` in here at all, we are already bailing out early if it's None. I'll fix it

PR #2302, opened 2025-07-25
update Rust and Go dependencies

jgraettinger:
**Description:**

Update Rust and Go dependencies, and deal with various minor fallout from breaking changes in our own code base. No changes to function / behavior.

The intent is that we're better set up to do regular incremental `cargo update` runs from here.

**Notes for reviewers:**

See individual commits for more detail.


jgraettinger:
Did a round of testing on a local stack, looks good.

Cargo.toml
```diff
@@ -24,159 +25,142 @@ aide = { version = "0", features = [
...
-bytesize = "1.1.0"
-bumpalo = { version = "3.14", features = ["collections"] }
-bytelines = "2.4"
-byteorder = "1.4"
-caseless = "0.2"
-chardetng = "0.1"
-chrono = { version = "0.4", features = ["serde"] }
+bytes = "1"
+bumpalo = { version = "3", features = ["collections"] }
+bytelines = "2"
```
danielnelson, line 45: We don't use any features added in bytelines 2.1.x 2.2.x etc?

Cargo.toml
```diff
@@ -24,159 +25,142 @@ aide = { version = "0", features = [
...
+byteorder = "1"
+caseless = "0"
```
danielnelson, line 47: It would be allowed for a library to break their api in each 0-version release, so I don't think it can ever be safe to support all 0 versions.

danielnelson (commented):
Just skimming, but had a couple questions about the new style dependency versions in Cargo.toml.

Cargo.toml
jgraettinger: Nope. In fact, it looks like it's not used at all and is here vestigially.

I removed as many of these as I could find, but it's painstaking...

Cargo.toml
jgraettinger: Yes, that's right. We're not trying to, though, because we've pinned exact workspace versions in Cargo.lock.

This relaxation of Cargo.toml done here is targeted at enabling bulk updates through a single `cargo update` command, which can then be tested / reviewed / canaried / released.

I debated removing _all_ versions for deps we don't need to specifically hold-back by using `"*"`...

PR #2304, opened 2025-07-28
agent: Address some PR feedback

jshearer:
**Description:**

This addresses the remaining PR feedback from https://github.com/estuary/flow/pull/2263

* use remove() to avoid cloning in initializes_read_schema
* simplify initial read schema discovery codepaths
* simplify setting `x-complexity-limit` on inferred schemas
  * Initialize shapes with default complexity limit annotation
  * Raise complexity limit to 10,000 when sourced schemas are applied in apply_sourced_schemas

Tested this locally with a `hello-world` and `postgres` connector, confirming that their behavior is expected: hello-world has `x-complexity-limit: 1000` and postgres has `x-complexity-limit: 10000`
Screenshot 2025-09-05 at 12 19 51
Screenshot 2025-09-05 at 12 19 33

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2304)

### DIFF
--- a/crates/control-plane-api/src/discovers/specs.rs
+++ b/crates/control-plane-api/src/discovers/specs.rs
@@ -319,7 +319,7 @@ pub fn merge_collections(
             // Collection is using separate read & write schemas.
             // The read schema has been initialized and we do not modify it.
             &mut draft_model.write_schema
-        } else if let Some(initial_read_schema) = initializes_read_schema(&connector_schema) {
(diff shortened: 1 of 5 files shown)


jshearer:
@jgraettinger I put this together a while ago and _finally_ got around to actually testing it, and in the intervening time some other changes happened to `crates/control-plane-api/src/discovers/specs.rs`. I _believe_ I resolved the conflict appropriately, but I'm just pointing that out as a place I could have misunderstood things.

PR #2307, opened 2025-07-29
Update MySQL.md

dyaffe:
**Description:**

Make it obvious that a user can use lower binlog retention durations.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2307)


aeluce (approved):
LGTM 👍

PR #2314, opened 2025-08-04
dekaf: Allow redirected sessions to field group management requests

jshearer:
**Description:**

Consumers assume that all brokers can serve group management requests. I had initially forbidden this in order to allow for segregating certain Dekaf instances onto their own MSK cluster, but given that we're ultimately planning on rolling our own global group state, there's no point in trying to keep that torch alive, especially since rejecting these requests by consumers is just confusing them.


crates/dekaf/src/session.rs
```diff
@@ -300,8 +324,70 @@ impl Session {
...
+                };
+
+                let partitions = collection
+                    .partitions
+                    .iter()
+                    .enumerate()
+                    .map(|(index, _)| {
+                        messages::metadata_response::MetadataResponsePartition::default()
+                            .with_partition_index(index as i32)
+                            .with_leader_id(messages::BrokerId(1))
```
jgraettinger, line 371: are broker IDs something the client could potentially be caching? Do we need to use distinct values for distinct data-plane FQDN's ?

jgraettinger (approved):
LGTM

crates/dekaf/src/session.rs
jshearer: I did think about that. It appears that unlike partition IDs, sparse broker IDs would be perfectly fine, so a straightforward hash of the dataplane's name/FQDN, collapsed into an `i32` for the broker ID ought to have a sufficiently small collision chance that it seems feasible.

I did some more digging into `librdkafka` though, and it appears that brokers changing their address is normal and expected. Upon redirect, when the consumer gets an error message or connection reset, it will ask for the metadata again, get the new address for broker ID 1, notice the address has changed, close all other connections to the old address and proceed with the new one.

jshearer:
Going to revert this as it has a performance regression that I haven't been able to pin down yet, and we want to get https://github.com/estuary/flow/pull/2353 out soon

PR #2320, opened 2025-08-06
dekaf: Fix propagation of `last_source_published_at` in stats documents

jshearer:
**Description:**

This was just never getting set on emitted stats documents, even through we were recording it at read time correctly.

**Testing:**

Confirmed that this emits `lastSourcePublishedAt` whereas it didn't before: 
```json
{
  "_meta": {
    "uuid": "f0e9592d-72f3-11f0-8000-1dede6ce7f7c"
  },
  "materialize": {
    "joseph/c2-to-c1-test/hello/events": {
      "lastSourcePublishedAt": "2025-08-06T18:34:08.431948212+00:00",
      "out": {
        "bytesTotal": 9407679,
        "docsTotal": 77913
      },
      "right": {
        "bytesTotal": 9407679,
        "docsTotal": 77913
      }
    }
  },
  "shard": {
    "build": "11fa3c5c178000e4",
    "keyBegin": "00000000",
    "kind": "materialization",
    "name": "joseph/c2-to-c1-test/dekaf/dekaf-generic",
    "rClockBegin": "00000000"
  },
  "ts": "2025-08-06T18:34:09.292934193+00:00"
}
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2320)

### DIFF
--- a/crates/dekaf/src/log_appender.rs
+++ b/crates/dekaf/src/log_appender.rs
@@ -51,6 +51,9 @@ impl StatsAggregator {
         if let Some(out) = &stats.out {
             ops::merge_docs_and_bytes(out, &mut binding.out);
         }
+        if let Some(last_source_published_at) = stats.last_source_published_at {
+            binding.last_source_published_at = Some(last_source_published_at);
+        }
     }
 
     // If any stats have been written, return them and reset the counter. Otherwise None
--- a/crates/dekaf/src/snapshots/dekaf__log_appender__tests__test_stats.snap
+++ b/crates/dekaf/src/snapshots/dekaf__log_appender__tests__test_stats.snap
@@ -9,6 +9,7 @@ expression: captured_logs
     },
     "materialize": {
       "test_collection": {
+        "lastSourcePublishedAt": "1970-01-01T00:00:06.000000007+00:00",
         "left": {
           "bytesTotal": 2,
           "docsTotal": 1


Alex-Bair (approved):
LGTM

crates/dekaf/src/log_appender.rs
```diff
@@ -51,6 +51,21 @@ impl StatsAggregator {
         if let Some(out) = &stats.out {
             ops::merge_docs_and_bytes(out, &mut binding.out);
         }
+        if let Some(last_source_published_at) = stats.last_source_published_at {
+            binding.last_source_published_at = Some(match binding.last_source_published_at {
+                Some(prev) => {
+                    if prev.seconds < last_source_published_at.seconds
```
psFried: I don't understand why this check would be needed. Why would `last_source_published_at` ever be less than `prev`? Is it just when we backfill? My knee jerk is that this condition doesn't really seem right.

psFried (commented):
One question on this

crates/dekaf/src/log_appender.rs
jshearer: `StatsAggregator`'s job is to roll up a bunch of stats events into a single stats event so it can be emitted periodically. My thinking here is, just like how we merge in the `DocsAndBytes` above for `left`, `right`, and `out`, we should also merge in `last_source_published_at` by always selecting the sooner of the `last_source_published_at`s, so the emitted stats document always has the soonest value.

I suppose if we make the assumption that all of the emitted stats documents are in order, and remain in order through the whole codepath then this check isn't needed, but it seemed like cheap insurance in case that's not the case

psFried: > if we make the assumption that all of the emitted stats documents are in order, and remain in order through the whole codepath

The intent of `last_source_published_at` is that it should be the `flow_published_at` of the last document that was materialized. That's how we treat it elsewhere (we don't have any similar conditionals for regular materialzations), and I think it makes sense for dekaf to treat it the same. If we somehow manage to read or materialize documents out of order then it seems like that would be its own problem, and it probably wouldn't help to have this logic potentially paper over that issue.

jshearer: Alright, sounds good I'll update this to just overwrite with the new val every time whatever it is 👍

psFried (approved):
LGTM

PR #2324, opened 2025-08-10
dekaf: Revert build

jshearer:
I started from the build of Dekaf we'd like to revert back to, `ghcr.io/estuary/flow:v0.5.11-272-gbe1988e8d2`, created `dekaf/revert-base` at its commit of `be1988e8d2`, and then cherry-picked https://github.com/estuary/flow/pull/2320 ontop of it.

This essentially reverts all of the changes in https://github.com/estuary/ops/pull/838 while keeping the fix for `last_source_published_at`.

_So long as_ we don't have anyone using Dekaf redirects in production, which is true given that I don't see any `Task has been redirected` logs in `dekaf.estuary-data.com` other than my testing, all of the rest of these changes are QoL improvements that _should_ be safe to revert

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2324)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -8,7 +8,7 @@ on:
     paths-ignore:
       - "site/**"
   pull_request:
-    branches: [master, dekaf/revert-cherry-picked]
+    branches: [master, dekaf/revert-base]
     paths-ignore:
       - "site/**"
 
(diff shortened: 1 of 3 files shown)


jshearer:
Closing this now that https://github.com/estuary/flow/pull/2333 is out and should fix the problem. This revert wouldn't have helped anyway.

PR #2327, opened 2025-08-11
dpc: allow_cidrs for data planes

mdibaiee:
**Description:**

- allow setting a custom allow_cidrs config for data planes. See https://github.com/estuary/est-dry-dock/pull/165

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2327)

### DIFF
--- a/crates/data-plane-controller/src/data_plane_fixture.json
+++ b/crates/data-plane-controller/src/data_plane_fixture.json
@@ -4,6 +4,7 @@
     "deployments": [],
     "gcp_project": "testing-ground-123456",
     "ssh_subnets": [],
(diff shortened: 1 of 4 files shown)


jgraettinger (commented):
I think this can just be part of the config and does not need to be a separate DB column.

The salient difference between this and private links is that this incorrect values cannot cause pulumi up to fail (as is the case with private links), so we don't expect to ever need to update a config of a data-plane currently in the PulumiUp1 state.

jgraettinger (approved):
LGTM

PR #2332, opened 2025-08-13
agent: fix optimistic concurrency bug in publication commit

psFried:
Fixes a gnarly bug in the publisher, which allowed multiple concurrent publications affecting the same specs commit. This resulted in cases where an initial publication result was clobbered by the results of another concurrent publication. The new integration test was able to reproduce this issue by spawning separate tasks for committing publications concurrently.

The fix for the concurrency issue was just to use `repeatable read` transaction isolation, instead of the default `read committed` level. Using `repeatable read` requires that we handle database errors due to transaction serialization failures, though, and that required a few other changes.

First, it required changing how we determine the `live_specs.id` values. Previously, we were having the database determine those, and we'd then update the `UncommittedBuild` with the ids after executing the insert statement. That made retries more difficult, since the `UncommittedBuild` would be modified by an attempt to commit. So this changes how we determine those ids, to have the agent determine them up front, just before attempting the commit.

This also required changing the signature of `WithCommit` to accept an `UncommittedBuild` and a separate `JobStatus` argument instead of the single `PublicationResult` argument. This made it easier to retry the commit, since we no longer need to convert to a `PublicationResult` in order to call `WithCommit`.

### DIFF
--- a/crates/agent/src/integration_tests/locking_retries.rs
+++ b/crates/agent/src/integration_tests/locking_retries.rs
@@ -9,6 +9,96 @@ use crate::{
 
 #[tokio::test]
 #[serial_test::serial]
+async fn test_publication_concurrent_commits() {
+    let harness = TestHarness::init("test_publication_concurrent_commits").await;
+
+    let user_id = harness.setup_tenant("beavers").await;
+
+    let draft = draft_catalog(serde_json::json!({
+        "collections": {
+            "beavers/dens": minimal_collection(None),
+        },
+        "captures": {
+            "beavers/dams": minimal_capture(None, &["beavers/dens"]),
+        }
+    }));
+
+    // Try to reproduce a scenario where multiple different publications all try to commit
+    // concurrently. We'll expect exactly one of them to succeed, and the others to fail.
+    for test_iteration in 0..5 {
+        let publisher_1 = harness.publisher.clone();
+        let publisher_2 = harness.publisher.clone();
+        let publisher_3 = harness.publisher.clone();
+
+        let build_1 = publisher_1
+            .build(
+                user_id,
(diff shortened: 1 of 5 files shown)


crates/agent/src/integration_tests/locking_retries.rs
```diff
@@ -7,6 +7,96 @@ use crate::{
...
+            tokio::spawn(async move { publisher_3.commit(build_3, NoopWithCommit).await.unwrap() });
+
+        let mut success_count = 0;
+        for (build, handle) in [handle_1, handle_2, handle_3].into_iter().enumerate() {
+            let result = handle.await.unwrap();
+
+            if result.status.is_success() {
+                success_count += 1;
+            }
+            // if result.as_ref().is_ok_and(|s| s.is_success()) {
```
jgraettinger: nit: remove?

crates/agent/src/publications.rs
```diff
@@ -528,38 +534,140 @@ impl<MC: MakeConnectors> Publisher<MC> {
...
+    /// this here instead of in the database because we need to use these ids
+    /// for updating `live_spec_flows` and other things. The ids need to be
+    /// added to the built rows in `output.built`, and assigning them up front
+    /// helps simplify the process of retrying after a transaction serialization
+    /// failure.
+    fn assign_control_ids(&self, output: &mut build::Output) {
+        let mut id_gen = self.id_gen.lock().unwrap();
+        let mut new_captures = 0;
+        for r in output.built.built_captures.iter_mut() {
+            // TODO: use r.is_insert() instead?
```
jgraettinger: nit: still TODO?

jgraettinger (approved):
LGTM

PR #2333, opened 2025-08-13
dekaf: Move TLS handshake out of main accept loop to prevent blocking

jshearer:
**Description:**

The TLS handshake should not block the main TCP accept loop. When a client connects but doesn't complete the TLS handshake promptly, we should still allow other connections to be accepted. This refactors the connection handling to spawn the TLS handshake in the connection handler task rather than the accept loop:

* Unify TLS and non-TLS code paths into a single accept loop with an optional TLS acceptor
* Move TLS handshake into `serve()` where it happens after spawning, preventing accept loop blocking

The main accept loop can now continue accepting new connections immediately, even if TLS handshakes are slow or hang indefinitely.

Also included is a small change to start up an explicitly created runtime, and explicitly tear it down when we're done.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2333)

### DIFF
--- a/crates/dekaf/src/main.rs
+++ b/crates/dekaf/src/main.rs
@@ -96,6 +96,10 @@ pub struct Cli {
     #[arg(long, env = "TASK_REFRESH_INTERVAL", value_parser = humantime::parse_duration, default_value = "30s")]
     task_refresh_interval: std::time::Duration,
 
+    /// Timeout for TLS handshake completion
+    #[arg(long, env = "TLS_HANDSHAKE_TIMEOUT", value_parser = humantime::parse_duration, default_value = "10s")]
+    tls_handshake_timeout: std::time::Duration,
+
     /// The fully-qualified domain name of the data plane that Dekaf is running inside of
     #[arg(
         long,
@@ -233,7 +237,8 @@ fn main() {
         }
     };
 
-    let result = runtime.block_on(async_main(cli));
+    let handle = runtime.spawn(async_main(cli));
+    let result = runtime.block_on(async { handle.await.unwrap() });
 
     // Explicitly shut down the runtime without waiting for blocking background tasks.
     // This prevents hangs from tasks that may be waiting on I/O that will never complete.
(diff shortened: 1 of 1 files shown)


crates/dekaf/src/main.rs
```diff
@@ -334,119 +336,98 @@ async fn main() -> anyhow::Result<()> {
...
-                    };
-                    socket.set_nodelay(true)?;
+    let mut backoff = None;
 
-                    let task_cancellation = cancel_token.child_token();
+    // Accept and serve Kafka sessions until we're signaled to stop.
+    loop {
+        tokio::select! {
+            accept = kafka_listener.accept() => {
+                if let Some(sleep_until) = backoff {
```
psFried: I'm not seeing why we'd want to delay here, after `accept()` has already returned. I'd think we'd want any delay to occur before we call `accept()`. Delays at this point would be observable to a connected client, unless I'm missing something.

crates/dekaf/src/main.rs
```diff
@@ -334,119 +336,98 @@ async fn main() -> anyhow::Result<()> {
...
+                    tokio::time::sleep_until(sleep_until).await;
+                    backoff = None;
+                }
+                let (socket, addr) = match accept {
+                    Ok(v) => v,
+                    Err(e) => {
+                        tracing::error!("Failed to accept connection: {e:?}");
+                        backoff = Some(tokio::time::Instant::now() + std::time::Duration::from_millis(100));
```
psFried: nit: seems a little cleaner to pass a duration only, and use plain `sleep` instead of `sleep_until`

crates/dekaf/src/main.rs
```diff
@@ -334,119 +336,98 @@ async fn main() -> anyhow::Result<()> {
...
+                        continue;
+                    }
+                };
+
+                if let Err(e) = socket.set_nodelay(true) {
+                    tracing::error!("Failed to set nodelay: {e:?}");
+                    backoff = Some(tokio::time::Instant::now() + std::time::Duration::from_millis(100));
```
psFried: Take this with a grain of salt, but this doesn't seem like it should necessarily trigger a delay. IDK why this would ever fail in practice, though, so feel free to ignore.

crates/dekaf/src/main.rs
```diff
@@ -457,6 +438,18 @@ where
 
     tracing::info!("accepted client connection");
 
+    // Optionally drive TLS handshake if needed
+    let socket: Box<dyn Connection> = match tls_acceptor {
+        Some(acceptor) => {
+            let tls_stream = acceptor
```
psFried: Given that we think we may have observed a case where this `accept` function never returned, I think we should ensure that there's some sort of timeout on this. IDK whether that's part of the tls config, or if we need an explicit timeout here, but seems like one of those would be a good idea.

crates/dekaf/src/main.rs
```diff
@@ -222,6 +221,31 @@ async fn main() -> anyhow::Result<()> {
...
+
+    let runtime = match runtime {
+        Ok(runtime) => runtime,
+        Err(error) => {
+            tracing::error!(%error, "couldn't build Tokio runtime");
+            std::process::exit(1);
+        }
+    };
+
+    let result = runtime.block_on(async_main(cli));
```
psFried: My understanding is that for `$reasons`, you want to use:

```rust
let handle = runtime.spawn(async_main(cli));
let result = runtime.block_on(async { handle.await });
```

Here's a thread on why: https://github.com/tokio-rs/tokio/discussions/6831

psFried (commented):
Had a few comments here, but overall looks good.

crates/dekaf/src/main.rs
jshearer: Ah 🤦 yeah, the backoff should totally be before the socket is accepted from the system, not after.

crates/dekaf/src/main.rs
jshearer: Yup, planned on doing this before marking ready for review, just pushed it. Does a default of 10s seem reasonable here? I feel like that should be more than enough 🤔

crates/dekaf/src/main.rs
jshearer: Yeah fair enough, I made it not trigger the backoff

crates/dekaf/src/main.rs
psFried: yeah, or even shorter (5s) would probably be fine. If we can't do the TLS handshake in under that time, then mostly likely things aren't going to work well for that connection anyway.

PR #2342, opened 2025-08-15
agent: use timeout when listing task shards

psFried:
Updates the activation controller to use a timeout when listing task shards. The lack of a timeout was an oversight. The client will re-try indefinitely when it's unable to connect to a data plane, so the timeout is needed in order for that to bubble up to a controller error.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2342)


jshearer (approved):
LGTM

PR #2345, opened 2025-08-18
go/*: some minor tweaks to reduce Go allocations

willdonnelly:
**Description:**

This is the consolation prize for all that profiling work, a couple bits of very low-hanging fruit which should be fairly uncontroversial improvements.

In my benchmarking these improve throughput by about 2-3MBps on a wide range of document sizes by reducing some unnecessary allocations in the hot path of document processing.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2345)

### DIFF
--- a/go/bindings/task_service.go
+++ b/go/bindings/task_service.go
@@ -21,6 +21,7 @@ import (
 	"github.com/sirupsen/logrus"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials/insecure"
+	"google.golang.org/grpc/experimental"
 )
 
 type TaskService struct {
@@ -120,6 +121,7 @@ func NewTaskService(
 		grpc.WithUnaryInterceptor(grpc_prometheus.UnaryClientInterceptor),
 		grpc.WithStreamInterceptor(grpc_prometheus.StreamClientInterceptor),
 		grpc.WithDefaultCallOptions(grpc.MaxCallRecvMsgSize(maxMessageSize), grpc.MaxCallSendMsgSize(maxMessageSize)),
+		experimental.WithRecvBufferPool(grpc.NewSharedBufferPool()),
 	)
 
 	if err != nil {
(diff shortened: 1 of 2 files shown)


go/flow/mapping.go
```diff
@@ -175,7 +176,9 @@ func pickPartition(logicalPrefix []byte, hexKey []byte, journals []pb.ListRespon
 		if l, r := len(logicalPrefix), len(pre); l < r {
 			pre = pre[:l] // Compare over the prefix.
 		}
-		switch bytes.Compare([]byte(pre), logicalPrefix) {
+		// Equivalent to `var unsafePrefixBytes = []byte(pre)` but without extra string allocations
```
jgraettinger, line 179: Huh, you're sure this is allocating? I recall doing an analysis (admittedly a while ago) of escaping heap allocations in here, and that the compiler was smart enough to not actually clone...

go/bindings/task_service.go
```diff
@@ -120,6 +121,7 @@ func NewTaskService(
 		grpc.WithUnaryInterceptor(grpc_prometheus.UnaryClientInterceptor),
 		grpc.WithStreamInterceptor(grpc_prometheus.StreamClientInterceptor),
 		grpc.WithDefaultCallOptions(grpc.MaxCallRecvMsgSize(maxMessageSize), grpc.MaxCallSendMsgSize(maxMessageSize)),
+		experimental.WithRecvBufferPool(grpc.NewSharedBufferPool()),
```
jgraettinger, line 124: Do you have a sense of how this interacts with the gRPC upgrade (to v1.74.2) in https://github.com/estuary/flow/tree/johnny/update-world ?

On that branch I switch raw json protobuf fields to `bytes` from `string` because I observed use-after-free errors given the incorrect string => bytes cast semantics we were using (to json.RawMessage which is []byte), tickled by a changed gRPC decoder / buffer system

jgraettinger (approved):
LGTM % comments

PR #2348, opened 2025-08-19
dekaf: Improve timeout handling when refreshing tokens

jshearer:
**Description:**

There was an event early this morning that ended up causing a lot of connection drops. After investigation, I believe it has to do with waiting indefinitely on requests to `/authorize/task` and `/authorize/dekaf`. I was not able to find any logs in agent-api indicating long requests that eventually succeeded, but other evidence includes:

* Dekaf logs complaining about expired tokens: `verifying Authorization: token has invalid claims: token is expired`. This means that we're attempting to use a journal client with an expired token.
* Dekaf logs from `Read::next_batch()` saying `second time provided was later than self`
  This is coming from `SystemTime::duration_since`. The only way I could imagine this happening is if `Read::new_stream()`, which fetches the latest `TaskState` from its `TaskStateListener` was to get an expired token. The assumption is that `TaskStateListener::get()` will either return a valid non-expired token, or an error.

  If the TaskManager's loop were to get stuck waiting on one of the network requests it makes, this assumption would be broken and we would see the behavior that we saw.

So, to fix the problem:
* I updated `TaskManager` to:
  * Proactively refresh its tokens earlier, starting 10 minutes before their expiration timestamp.
  * Added timeouts to the network calls, and made it resilient in the face of those timeouts: while the token is still valid, keep retrying and returning the cached token if refresh requests time out.
* I updated `Read` to error if, even after attempting to refresh its token, the provided expiration is still in the past. 
* Added `TASK_REQUEST_TIMEOUT` / `--task-request-timeout` which allows configuring the TaskManager timeout. I defaulted it to 20s which seems long, but agent-api has a backoff mechanism that includes multi-second delays which are covered under this timeout, and I wanted to avoid false-positive timeouts.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2348)

### DIFF
--- a/crates/dekaf/src/main.rs
+++ b/crates/dekaf/src/main.rs
@@ -87,9 +87,13 @@ pub struct Cli {
     #[arg(long, env = "IDLE_SESSION_TIMEOUT", value_parser = humantime::parse_duration, default_value = "30s")]
     idle_session_timeout: std::time::Duration,
 
-    /// How long to cache materialization specs and other task metadata for before re-refreshing
+    /// How long to cache materialization specs and other task metadata for before refreshing
     #[arg(long, env = "TASK_REFRESH_INTERVAL", value_parser = humantime::parse_duration, default_value = "30s")]
     task_refresh_interval: std::time::Duration,
+
+    /// How long before a request for materialization specs and other task metadata times out
+    #[arg(long, env = "TASK_REQUEST_TIMEOUT", value_parser = humantime::parse_duration, default_value = "30s")]
+    task_request_timeout: std::time::Duration,
 
     /// Timeout for TLS handshake completion
     #[arg(long, env = "TLS_HANDSHAKE_TIMEOUT", value_parser = humantime::parse_duration, default_value = "10s")]
@@ -223,6 +227,7 @@ async fn async_main(cli: Cli) -> anyhow::Result<()> {
 
     let task_manager = Arc::new(TaskManager::new(
         cli.task_refresh_interval,
+        cli.task_request_timeout,
         client_base.clone(),
         cli.data_plane_fqdn.clone(),
         signing_token.clone(),
--- a/crates/dekaf/src/task_manager.rs
+++ b/crates/dekaf/src/task_manager.rs
@@ -45,7 +45,10 @@ impl fmt::Display for SharedError {
 
 pub type Result<T> = core::result::Result<T, SharedError>;
 
+/// How long to keep a TaskManager alive without any listeners.
 const TASK_TIMEOUT: Duration = Duration::from_secs(60 * 3);
+/// How long before the end of an access token should we start trying to refresh it
+const REFRESH_START_AT: Duration = Duration::from_secs(60 * 5);
 
 #[derive(Clone)]
 pub enum TaskState {
@@ -85,11 +88,24 @@ impl TaskStateListener {
             // Scope to force the borrow to end before awaiting
             {
                 let current_value = temp_rx.borrow_and_update();
-                if let Some(ref result) = *current_value {
-                    return result
-                        .as_ref()
-                        .map(|arc| Arc::clone(arc))
-                        .map_err(|e| anyhow::Error::from(e.clone()));
+                match &*current_value {
+                    Some(Ok(state)) => match state.as_ref() {
+                        TaskState::Authorized {
+                            access_token_claims,
+                            ..
+                        } if access_token_claims.exp
+                            <= time::OffsetDateTime::now_utc().unix_timestamp() as u64 =>
+                        {
+                            anyhow::bail!("Access token has expired and the task manager has been unable to refresh it.");
+                        }
+                        _ => return Ok(state.clone()),
+                    },
+                    Some(res) => {
+                        return res.clone().map_err(anyhow::Error::from);
+                    }
+                    None => {
+                        tracing::debug!("No task state available yet, waiting for the next update");
+                    }
                 }
             }
 
@@ -125,6 +141,7 @@ pub struct TaskManager {
         >,
     >,
     interval: Duration,
+    timeout: Duration,
     client: flow_client::Client,
     data_plane_fqdn: String,
     data_plane_signer: jsonwebtoken::EncodingKey,
@@ -132,6 +149,7 @@ pub struct TaskManager {
 impl TaskManager {
     pub fn new(
         interval: Duration,
+        timeout: Duration,
         client: flow_client::Client,
         data_plane_fqdn: String,
         data_plane_signer: jsonwebtoken::EncodingKey,
@@ -139,6 +157,7 @@ impl TaskManager {
         TaskManager {
             tasks: std::sync::Mutex::new(HashMap::new()),
             interval,
+            timeout,
             client,
             data_plane_fqdn,
             data_plane_signer: data_plane_signer,
@@ -228,6 +247,8 @@ impl TaskManager {
         let mut cached_ops_stats_client: Option<Result<(journal::Client, proto_gazette::Claims)>> =
             None;
 
+        let mut cached_dekaf_auth: Option<DekafTaskAuth> = None;
+
         let mut timeout_start = None;
 
         loop {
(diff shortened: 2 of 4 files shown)


crates/dekaf/src/task_manager.rs
```diff
@@ -85,8 +89,21 @@ impl TaskStateListener {
             // Scope to force the borrow to end before awaiting
             {
                 let current_value = temp_rx.borrow_and_update();
-                if let Some(ref result) = *current_value {
-                    return result.clone().map_err(anyhow::Error::from);
+                match &*current_value {
```
jshearer, line 91: We used to happily hand out expired tokens here, just assuming that the `TaskManager` loop could keep up. That's the real cause of the problem

crates/dekaf/src/task_manager.rs
```diff
@@ -85,8 +89,21 @@ impl TaskStateListener {
...
+                    Some(Ok(TaskState::Authorized {
+                        access_token_claims,
+                        ..
+                    })) if access_token_claims.exp
+                        <= time::OffsetDateTime::now_utc().unix_timestamp() as u64 =>
+                    {
+                        anyhow::bail!("Access token has expired and the task manager has been unable to refresh it.");
```
jshearer: An alternative to crashing here would be awaiting until the channel has seen a new value, but that felt like it could result in even worse problems if the `TaskManager` gets stuck for some other reason. Better to crash and try again than potentially wait forever.

crates/dekaf/src/task_manager.rs
```diff
@@ -576,6 +649,60 @@ pub enum DekafTaskAuth {
     },
 }
 
+impl DekafTaskAuth {
+    fn exp(&self) -> u64 {
+        match self {
+            DekafTaskAuth::Redirect { fetched_at, .. } => {
+                // Redirects are valid for 10 minutes
```
jshearer, line 643: I just made this up. Since redirects don't get a token, they don't get an `agent-api`-specified expiration. 10 minutes seems more than fine, as the only time where this would matter is if a task were migrated such that it is no longer a redirect.

crates/dekaf/src/main.rs
```diff
@@ -92,10 +92,14 @@ pub struct Cli {
     #[arg(long, env = "IDLE_SESSION_TIMEOUT", value_parser = humantime::parse_duration, default_value = "30s")]
     idle_session_timeout: std::time::Duration,
 
-    /// How long to cache materialization specs and other task metadata for before re-refreshing
+    /// How long to cache materialization specs and other task metadata for before refreshing
     #[arg(long, env = "TASK_REFRESH_INTERVAL", value_parser = humantime::parse_duration, default_value = "30s")]
     task_refresh_interval: std::time::Duration,
 
+    /// How long before a request for materialization specs and other task metadata times out
+    #[arg(long, env = "TASK_REQUEST_TIMEOUT", value_parser = humantime::parse_duration, default_value = "20s")]
```
psFried: I think this default timeout should probably be a little higher. The [minimum snapshot refresh interval](https://github.com/estuary/flow/blob/8fa5cd9e1f062e6af6548668e2320e5ee123b7af/crates/agent/src/api/snapshot.rs#L395) is 20s, which would be the maximum amount of time that we might need to delay in order to authorize a request. So maybe more like 25-30s?

crates/dekaf/src/task_manager.rs
psFried: Does returning an error here crash the whole dekaf process? Or is it just the handler for this one connection?

crates/dekaf/src/task_manager.rs
```diff
@@ -255,21 +277,38 @@ impl TaskManager {
...
-                // fresh MaterializationSpecs, even if the access token may live for a while.
-                let dekaf_auth = fetch_dekaf_task_auth(
-                    &self.client,
-                    &task_name,
-                    &self.data_plane_fqdn,
-                    &self.data_plane_signer,
-                )
-                .await
-                .context("error fetching dekaf task auth")?;
+                let dekaf_auth = if let Some(ref cached_auth) = cached_dekaf_auth {
```
psFried: nit: I found this conditional a little confusing, because either code path just calls `fetch_dekaf_task_auth`. It seems like you could make the `get_or_refresh` function take an `&Option<DekafTaskAuth>` instead of `&self` in order to consolidate these two branches.

crates/dekaf/src/task_manager.rs
```diff
@@ -493,29 +537,55 @@ async fn get_or_refresh_journal_client(
...
+            data_plane_signer,
+            capability,
+            selector,
+        ),
     )
     .await
+    {
+        Ok(resp) => resp,
+        Err(_) => {
+            if let Some((allowed_until, cached_client, cached_claims)) = timeouts_allowed_until {
```
psFried, line 565: So we'll hand out the cached client if the `/authorize/task` call fails with a timeout, but not if it fails for any other reason (e.g. HTTP 503)? This feels a little weird. I'm not super familiar with all the callsites of this, but it seems like we'd probably want to treat other errors (except for 4xx) the same as a timeout, and return the cached client.

psFried (commented):
Couple of questions on this, but nothing major.

crates/dekaf/src/task_manager.rs
jshearer: It does not, it bubbles up to the serve loop for the connection in question and ends it early.

crates/dekaf/src/main.rs
jshearer: bumped to 30 :+1:

crates/dekaf/src/task_manager.rs
jshearer: Yeah agree, tbh I went back and forth on this. Updated to do it the way you talk about :+1:

crates/dekaf/src/task_manager.rs
jshearer: I think that approach makes sense, but it requires a bit more work so I'm thinking I want to do it as a separate PR. Currently `agent_unary` just returns a simple `anyhow::bail!`: https://github.com/estuary/flow/blob/44565eac547b810805b33005615c0a361a7a3178/crates/flow-client/src/client.rs#L167-L172

I'm thinking in order to do this the right way, we'd want it to return something more like `gazette::Error` which has: https://github.com/estuary/flow/blob/44565eac547b810805b33005615c0a361a7a3178/crates/gazette/src/lib.rs#L65-L71

Which we could then use to determine whether the error is able to be retried or should be cached and returned as a more permanent error.

psFried: Yeah, it does seem like `flow-client` ought to return error types that allow for easier introspection. Makes sense to handle that in a separate pr 👍

crates/dekaf/src/task_manager.rs
psFried: 🤔 Does this mean that we'll now start to expire `Redirect`s, where previously we haven't? Just wanting to double check whether that poses any risk for existing tasks that may not have encountered that before.

crates/dekaf/src/topology.rs
```diff
@@ -1,5 +1,6 @@
 use crate::{connector, utils, SessionAuthentication, TaskState};
 use anyhow::{anyhow, bail, Context};
+use futures::StreamExt;
```
psFried, line 3: This seems unintended?

psFried (approved):
LGTM

crates/dekaf/src/topology.rs
jshearer: https://github.com/estuary/flow/pull/2373 yeah 🤦 it got missed in another PR

crates/dekaf/src/task_manager.rs
jshearer: We used to re-fetch every 30s no matter what. Now we only fetch when the (made up) expiration is coming up. I don't think this substantively changes the behavior, just the amount of time a redirect response can be cached for

PR #2353, opened 2025-08-21
dekaf: add basic field folding, and update selection constraints

jgraettinger:
Simplify constraint generation in light of new field selection semantics, and also support a new fold of `/` characters (replacing with underscore).

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2353)


crates/dekaf/src/connector.rs
```diff
@@ -186,41 +181,27 @@ where
...
-            folded_field: String::new(),
-        }
-    } else if projection.field == "_is_deleted"
+    } else if folded_field == "_is_deleted"
         && matches!(endpoint_config.deletions, DeletionMode::CDC)
     {
         materialize::response::validated::Constraint {
             r#type: constraint::Type::FieldForbidden.into(),
             reason: "Cannot materialize input data to '_is_deleted' when using CDC deletions mode as it will be generated by Dekaf".to_string(),
-            folded_field: String::new(),
```
jshearer, line 223: Oh interesting, so I guess field folding was planned for a while but just not implemented until now? I didn't even remember this field being here 🤔

crates/dekaf/src/connector.rs
```diff
@@ -186,41 +181,27 @@ where
...
             reason: format!(
                 "Field name {} contains characters forbidden in AVRO schemas. It must match the regex '{}'. If you still want to include this field, you can create a projection for this field on the source collection with a name that matches the regex. See https://go.estuary.dev/projections for more information.",
-                projection.field,
+                &folded_field,
                 avro::AVRO_FIELD_RE.as_str()
             ),
-            folded_field: String::new(),
+            folded_field,
         };
-    } else if projection.is_primary_key {
```
jshearer, line 205: Why are we getting rid of these constraints? Something about how the new field selection behavior is responsible for this now I'm guessing?

crates/dekaf/src/lib.rs
```diff
@@ -756,6 +756,15 @@ fn bool<const U: bool>() -> bool {
     U
 }
 
+// field_fold maps a projection field name to its folded AVRO-compatible name.
+// Currently this is the most basic possible transform that provides somewhat
+// reasonable UX, and will forbid any field which isn't AVRO_FIELD_RE relaxed
+// to also allow '/'.
+// TODO(johnny): Fold unicode (punycode?) and whitespace.
```
jshearer, line 763: Does anything depend on this TODO? Or just something to keep in mind? As it stands, unicode characters in field names will be rejected by the `AVRO_FIELD_RE` constraint, so I guess we need to add unicode folding if enough people have non-ascii field names.

crates/dekaf/src/connector.rs
jgraettinger: Yes.

crates/dekaf/src/lib.rs
jgraettinger: Nothing depends on it. Yes, since we reject anything else, we can safely open up the set of cases that we fold over time (but not shrink it), so long as we release dekaf runtime instances before agent instances.

PR #2354, opened 2025-08-21
proto string => bytes WIP

jgraettinger:
Update all string protobuf fields which use casttype
`encoding/json.RawMessage` to instead be bytes.

This is a correctness issue, and is required for correct processing
semantics in later Go gRPC library versions. The decoder must be able to
assume that a cast of []byte to a string field allows the []byte to be
re-used post-cast, and the use of json.RawMessage breaks this
assumption.

It's also a performance concern, because `prost` in Rust is able to
use zero-copy bytes::Bytes to back a `bytes` field, while String _must_
be deeply cloned.

No semantic or wire changes to the protocol -- this is just a low-level refactor of the in-memory type representation.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2354)


crates/runtime/src/materialize/protocol.rs
```diff
@@ -452,9 +454,9 @@ pub fn send_connector_store(
             delete: meta.deleted(),
             doc_json,
             exists: meta.front(),
-            key_json: String::new(), // TODO(johnny)
+            key_json: bytes::Bytes::new(), // TODO(johnny)
```
psFried, line 457: nit: do you remember what these TODOs were for? Remove em?

psFried (approved):
LGTM

crates/runtime/src/materialize/protocol.rs
jgraettinger: They're for plumbing in a json encoding (array) of the key, for connectors that prefer to work with JSON instead of packed foundation-DB tuples. Obviously hasn't been super important, but may be if we start to support user-implemented connectors.

PR #2359, opened 2025-08-25
Phil/graphql

psFried:
**Description:**

Introduces an initial take on a control plane GraphQL API, using the `async-graphql` crate. It's expected that we'll continue to iterate on the graphql schema as we continue building UI, but the idea is to merge what we have so far so we can more easily collaborate.

This also includes a refactoring of agent crates. The `agent-sql` crate is removed, in favor of a new crate called `control-plane-api`, which includes the REST and GraphQL APIs, and also exports Rust functions for use by `agent`.

**Workflow steps:**

- Visit `http://localhost:8675/graphiql` to see a UI for testing GraphQL queries and exploring the schema/docs.
- Copy the auth token from the admin page and paste it into the Headers section.
- Then click the "refresh" button in the lower left corner of the page to re-fetch the graphql schema (which requires the auth token)
- Start testing queries

None for now. Eventually, we may want some public docs for this, but I think it makes sense to let things stabilize for a while before we encourage outside use of this API.

**Notes for reviewers:**

The agent refactoring is a bit of a work in progress. There might be some functions that should move back into `agent` from `control-plane-api`, and there's almost certainly some better ways to organize the code within `control-plane-api` itself. I tried to keep this initial refactoring as minimal as I could in order to make it easier to review. Most `agent-sql` modules were simply renamed into `control_plane_api::<something>::db` modules just to make reviewing easier. It'll probably make sense to do another refactor to better organize the code within `control-plane-api`, but I'm leaving that for the future.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2359)

### DIFF
--- a/.github/workflows/dekaf-ci.yaml
+++ b/.github/workflows/dekaf-ci.yaml
@@ -17,6 +17,7 @@ on:
 env:
   CARGO_TERM_COLOR: always
   RUSTFLAGS: "-C link-arg=-fuse-ld=lld"
+  SQLX_OFFLINE: "1"
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
--- a/.github/workflows/flowctl-release.yaml
+++ b/.github/workflows/flowctl-release.yaml
@@ -13,6 +13,7 @@ jobs:
       ASSET_NAME: ${{ matrix.config.assetName }}
       # build.rs reads this env variable and uses to set the value that's printed by flowctl --version
       FLOW_VERSION: ${{ github.event.release.tag_name }}
+      SQLX_OFFLINE: "1"
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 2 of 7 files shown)


crates/flow-client/control-plane-api.graphql
```diff
@@ -0,0 +1,202 @@
...
+	free_trial
+	free_trial_ending
+	free_trial_stalled
+	missing_payment_method
+}
+
+"""
+A prefix to which the user is authorized.
+"""
+type AuthorizedPrefix {
```
jgraettinger, line 70: Just `Prefix`. I think these become 1:1 with storage mapping and data-plane configs attached to prefixes. Let's make it one concept.

jgraettinger (approved):
LGTM 🎉 

We had a good VC discussing where we want to take the graphQL server spec from here, and the disposition is to merge this and keep iterating in future PRs.

crates/flow-client/control-plane-api.graphql
jgraettinger: this can be deferred until later

PR #2362, opened 2025-08-26
dekaf: Fix performance issue caused by excessive `TaskState` cloning

jshearer:
**Description:**

`task_state_listener.get()` was cloning the whole `TaskState` upon every call, even if it hadn't changed. https://github.com/estuary/flow/pull/2314 introduced a bunch more frequent calls to this function, with the assumption that calling it would be cheap. Changing it to return an Arc: 

```rust
pub async fn get(&self) -> anyhow::Result<Arc<TaskState>>
```

fixed the issue entirely.  Also included is a small change to logging to backoff if we get an error.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2362)


psFried (approved):
LGTM

PR #2365, opened 2025-08-26
Dekaf: re-merge migration support with performance fix applied

jshearer:
**Description:**

https://github.com/estuary/flow/pull/2314 got reverted because of a performance regression. That was fixed in https://github.com/estuary/flow/pull/2362 and included in this PR.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2365)

### DIFF
--- a/crates/dekaf/src/log_appender.rs
+++ b/crates/dekaf/src/log_appender.rs
@@ -236,7 +236,7 @@ impl GazetteAppender {
                     target_dataplane_fqdn,
                     ..
                 } => {
-                    unreachable!("Task has been redirected to {}", target_dataplane_fqdn);
+                    anyhow::bail!("Task has been redirected to {}", target_dataplane_fqdn);
                 }
             },
             GazetteAppender::OpsLogs(state) => match state.task_listener.get().await? {
@@ -249,7 +249,7 @@ impl GazetteAppender {
                     target_dataplane_fqdn,
                     ..
                 } => {
-                    unreachable!("Task has been redirected to {}", target_dataplane_fqdn);
+                    anyhow::bail!("Task has been redirected to {}", target_dataplane_fqdn);
                 }
             },
         }
--- a/crates/dekaf/src/log_appender.rs
+++ b/crates/dekaf/src/log_appender.rs
@@ -151,7 +151,7 @@ impl GazetteWriter {
 
         let initial_state = task_listener.get().await?;
 
-        let (ops_logs_journal, ops_stats_journal) = match initial_state {
+        let (ops_logs_journal, ops_stats_journal) = match initial_state.as_ref() {
             crate::task_manager::TaskState::Authorized {
                 ops_logs_journal,
                 ops_stats_journal,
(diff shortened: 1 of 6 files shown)


crates/dekaf/src/session.rs
```diff
@@ -383,6 +383,11 @@ impl Session {
         &mut self,
         request: messages::FindCoordinatorRequest,
     ) -> anyhow::Result<messages::FindCoordinatorResponse> {
+        // if matches!(self.auth, Some(SessionAuthentication::Redirect { .. })) {
```
psFried: Was this left commented intentionally?

crates/dekaf/src/log_appender.rs
```diff
@@ -365,6 +365,9 @@ impl<W: TaskWriter + Clone + 'static> TaskForwarder<W> {
         let mut log_append_state: MaybeDone<PinnedAppendFuture> = MaybeDone::Gone;
         let mut stats_append_state: MaybeDone<PinnedAppendFuture> = MaybeDone::Gone;
 
+        let mut backoff_duration = std::time::Duration::from_millis(100);
+        let max_backoff = std::time::Duration::from_secs(10);
```
psFried: nit: make this `const`? IIRC clippy warns on this

crates/dekaf/src/log_appender.rs
```diff
@@ -383,14 +386,24 @@ impl<W: TaskWriter + Clone + 'static> TaskForwarder<W> {
                 // Poll log and stats appends, if running
                 _ = &mut log_append_state, if !log_append_state.is_terminated() => {
                     if let Some(Err(e)) = Pin::new(&mut log_append_state).take_output(){
-                        tracing::error!(error = ?e, "Error appending logs");
+                        tracing::error!(error = ?e, ?backoff_duration, "Error appending logs");
+
+                        tokio::time::sleep(backoff_duration).await;
+                        backoff_duration = (backoff_duration * 2).min(max_backoff);
```
psFried: It doesn't seem like anything ever resets the `backoff_duration`, and I'm wondering whether that could pose a problem. If this loop is potentially very long lived, then I'd probably want to reset the backoff duration after a successful append. Or maybe just use a fixed backoff for now.

psFried (approved):
I left a few comments, but otherwise LGTM

crates/dekaf/src/log_appender.rs
psFried: Ah, nvm, I didn't notice the reset below. Disregard.

crates/dekaf/src/session.rs
```diff
@@ -383,12 +411,7 @@ impl Session {
         &mut self,
         request: messages::FindCoordinatorRequest,
     ) -> anyhow::Result<messages::FindCoordinatorResponse> {
-        // if matches!(self.auth, Some(SessionAuthentication::Redirect { .. })) {
-        //     return Ok(messages::FindCoordinatorResponse::default()
-        //         .with_error_code(ResponseError::CoordinatorNotAvailable.code()));
-        // }
-
-        let (broker_host, broker_port) = match self.get_redirect_address() {
+        let (broker_host, broker_port) = match self.get_redirect_address().await? {
```
jshearer, line 414: @psFried I can't figure out how to link you to this directly, but wrt your comment about leftover comments in this function, they're removed here. I should probably have just squashed the two commits so it's easier to review commit-by-commit, my bad.

crates/dekaf/src/session.rs
jshearer: [It's removed in the subsequent commit](https://github.com/estuary/flow/pull/2365#discussion_r2302037509)

PR #2367, opened 2025-08-27
parser: print file content prefix when parsing fails completely

psFried:
This is intended to help diagnose issues with unexpected file content or parser behavior by logging out a small prefix of the file content when parsing fails.

This is only done when no records could be successfully parsed from the file. That's just because it's much easier to do that way, and still seems helpful.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2367)


Alex-Bair (approved):
LGTM

PR #2368, opened 2025-08-27
agent: Fix alert state serialization

psFried:
Fixes a bug with the alert state serialization. These were getting serialized with `PascalCase`, but the `controller_alerts` view expects `lowercase`. This prevented alerts from be fired by controllers.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2368)


jshearer (approved):
LGTM

PR #2370, opened 2025-08-28
dekaf: Add `name:prefix={collection_name}` to List RPCs

jshearer:
**Description:**

This changes the behavior from using `Binding.partition_selector`, which does _not_ currently include `name:prefix`

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2370)

### DIFF
--- a/crates/dekaf/src/task_manager.rs
+++ b/crates/dekaf/src/task_manager.rs
@@ -7,6 +7,7 @@ use anyhow::Context;
 use futures::StreamExt;
 use gazette::{broker, journal};
 use itertools::Itertools;
+use models::LabelSelector;
 use proto_flow::flow::MaterializationSpec;
 use rand::Rng;
 use std::{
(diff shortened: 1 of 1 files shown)


crates/dekaf/src/task_manager.rs
```diff
@@ -427,14 +428,25 @@ async fn update_partition_info(
...
             .context("expected partition template")?;
 
+        let template_name = partition_template.name.clone();
+        let task_name_clone = task_name.to_string();
+
         let partition_selector = binding
             .partition_selector
             .as_ref()
+            .map(|selector| {
+                let include = labels::add_value(
```
jgraettinger: You'll want to use set_value instead of add_value.

[The gazette optimization only kicks in if there's a single label value](https://github.com/gazette/core/blob/2580071332a6bf7f9302af1e513391f8c6539f5d/broker/list_apply_api.go#L147), so set_value makes this resilient to the partition selector including `name:prefix` as well.

Also, you're discarding `selector.exclude`, which potentially isn't empty.

jshearer: Thanks -- swapped to `set_value` and fixed droping `selector.exclude`

jgraettinger (approved):
LGTM

PR #2371, opened 2025-08-28
agent: allow deferring auto-discovers

psFried:
This is a quick and dirty way to allowing us to defer auto-discovers based on a probability given in an env variable.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2371)


jgraettinger (approved):
LGTM

PR #2379, opened 2025-09-04
update `rkyv` and implement tape-length optimization

jgraettinger:
**Description:**

Upgrade rkyv, which includes some breaking changes, add an inline tape-length optimization, and then use it within the reduction machinery.

See individual commit messages for many more details.

Performance isn't noticably different in the available `combiner_perf` benchmark:
* The updating to rkyv 0.8 lost us ~5% performance in that benchmark.
* The introduction of the tape-length optimization gained that performance right back.

The `parser_perf` benchmark also shows no detectable change (though we're now building ArchivedNodes with tape lengths able to accelerate reduction / redaction).

I suspect this PR can only (substantially 🤞) improve the common case of reduction, where simd-doc is used for parsing and large chunks of the tree are no longer walked during `lastWriteWins` reduction, but I don't have a ready benchmark to prove that.

**Workflow steps:**

No change in any semantics.

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2379)


jgraettinger:
new error check for conflicting strategies caught some legit ambiguities in catalog tests / examples.

PR #2381, opened 2025-09-04
[DRAFT] Docs: Improve docs structure

aeluce:
**Description:**

As Estuary has grown and added more information to docs, the current structure has not kept pace. It is not always clear where information lives and so the docs can be difficult to navigate.

This update is meant to be a restructure to break docs out into smaller categories. Categories should be:
* Defined by topic
* Organized roughly in order of relevance/when users will encounter the concepts
* Reasonably sized, with sub-categories collecting related topics so the eye can more easily land on items of interest without a lot of scrolling

To minimize breaking links, moved documents retain their old URL. As much of the PR is therefore fixing up references for doc locations, it's probably easier to review using the GitHub preview and comparing the resulting structure with the current docs.

Lots. Moved docs should retain their original URL. Apart from an initial bare-bones Security & Compliance section, no new information has been added, simply reorganized.

**Notes for reviewers:**

Thanks for any suggestions or comments!


site/docs/transformations/_category_.json
```diff
@@ -0,0 +1,4 @@
+{
+    "label": "Transforming Data",
```
danthelion, line 2: I love this as a separate section

site/docs/getting-started/comparisons.md
danthelion, line 1: I'm a little conflicted about the position of this comparison section. It compares Flow itself on a fairly technical level to other frameworks, so definitely valuable but I do wonder if this type of comparison is something freshly onboarding users need or they'd be more interested in comparisons with other hosted ETL services.

danthelion (commented):
As a generic comment I think we should consider pulling the connector doc folders to the top level - they are kinda hidden and I feel like a lot of users are looking for them from the main page. Maybe just putting the "Connectors" group to the top level would be enough. What do you think?

aeluce:
> As a generic comment I think we should consider pulling the connector doc folders to the top level - they are kinda hidden and I feel like a lot of users are looking for them from the main page. Maybe just putting the "Connectors" group to the top level would be enough. What do you think?

Agreed. They're some of the most-used docs and some of the most deeply nested. What do you think of bringing the connector reference up to the top navbar and treating it more like an API reference? I added a rough example of this to the branch now. Some kind of styling might help draw attention to the reference as well.

(This temporarily puts the Reduction Strategies category in limbo but I can find a new home for it if this otherwise seems like a good direction.)

site/docs/getting-started/comparisons.md
aeluce: That's reasonable. Orienting against comparable services can help users get up to speed quickly and showcase how Estuary's different, but it may be diving into the technical weeds as-is. I can consider restructuring. We may be able to have a streamlined upfront comparison guide and link to more technical details elsewhere in docs.

aeluce:
Thanks for catching these! I agree that the Connectors section link at the top may not be immediately obvious, especially for people who are already familiar with the platform and used to finding connector docs at the bottom of the sidebar. Would you have any suggestions on improving the experience?

travjenkins (approved):
lgtm

site/src/css/custom.css
```diff
@@ -69,4 +69,13 @@ body.running-in-iframe .theme-doc-breadcrumbs {
 .docs-section-link:hover {
     text-decoration: underline;
     background-color: rgb(233, 238, 237);
+}
+
+.navbar__link--active {
+    color: var(--ifm-color-primary-darker);
+    background-color: rgb(233, 238, 237);
+}
```
travjenkins, line 78: Can you please add a little bit of margin between the buttons?

```
.navbar__items .navbar__items {
margin-left: 2px;
margin-right: 2px;
}
```

travjenkins: Welp... that is the wrong css... sorry

`.navbar__items .navbar__link`

aeluce: No problem; that's what I ended up using anyway. 😄

PR #2382, opened 2025-09-04
dekaf: Fix excessive `/authorize/task` calls

jshearer:
**Description:**

The `refresh_at()` method was subtracting milliseconds from a Unix timestamp in seconds, causing it to return a time ~3.5 days in the past. This made every refresh check fail, triggering API calls on every task manager interval (30s) instead of only when tokens were actually close to expiration.

Fixed by using `as_secs()` instead of `as_millis()`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2382)


jgraettinger (approved):
LGTM

PR #2383, opened 2025-09-04
Add support for schema-driven redaction

jgraettinger:
**Description:**

This PR introduces schema-driven data redaction for Flow, allowing sensitive data to be automatically redacted from documents based on JSON Schema `redact` annotations. The feature supports two redaction strategies:       
  - block: Removes annotated properties/items entirely from documents
  - sha256: Replaces values with salted SHA-256 hash digests

Redaction is applied before documents leave the runtime's MemTable, preventing writes to disk or being surfaced in error messages. Salt is generated and managed for capture and derivation tasks to ensure deterministic hashing, with an ability to specify a manual salt value.

**Workflow steps:**

 1. Add top-level `redact` annotations to the collection schema / writeSchema:
 ```json
  {
    "$defs": { },
    "$ref": "flow://connector-schema",
    "properties": {
      "ssn": {        
        "redact": {"strategy": "sha256"}                                                                                                                          
      },                                                                                                                                                          
      "internal_id": {                                                                                                                                            
        "redact": {"strategy": "block"}                                                                                                                           
      }
    }                                                                                                                                                             
  } 
  ```                                                                                                                                                              
  2. Salt configuration (optional): For captures/derivations, salts can be specified as `redactSalt` (base64). An existing salt is passed-through from a live spec, and (on creation) a salt is generated deterministically via `xxhash(init_vector + task_name)`.

**Notes for reviewers:**

  - Validation prevents redaction of document locations used as keys (would break collection semantics)
  - New fuzz test on redaction logic
  - Redaction also uses the new tape-length optimization to avoid traversing un-redacted portions of the document
  - Existing live specs do NOT initialize salt values.
    - This was a implementation trade-off to slow the rate of introducing `redactSalt`'s in production, to reduce the blast radius of issues upon release.
  - Similarly Projection.Inference does _not_ yet have `redact` or `reduce` wired in, to avoid breaking the current UI.
  - UI must update to next `flow-web` version to prevent `redact` annotations from breaking the UI.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/flow/2383)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3024,7 +3024,7 @@ dependencies = [
 
 [[package]]
 name = "flow-web"
-version = "0.5.18"
+version = "0.5.19"
 dependencies = [
  "console_error_panic_hook",
  "doc",
--- a/crates/flow-web/Cargo.toml
+++ b/crates/flow-web/Cargo.toml
@@ -2,7 +2,7 @@
 name = "flow-web"
 
 # wasm-pack isn't yet fully compatible with workspace inheritance, so we can't use that for these fields
-version = "0.5.18"
+version = "0.5.19"
 authors = ["Estuary developers  <engineering@estuary.dev>"]
 edition = "2021"
 license = "BSL"


crates/doc/src/combine/memtable.rs
```diff
@@ -76,17 +76,29 @@ impl Entries {
...
 
             let rhs_valid = validator
-                .validate(schema.as_ref(), &rhs.root)
-                .map_err(Error::SchemaError)?
+                .validate(schema.as_ref(), &rhs.root)?
                 .ok()
-                .map_err(|err| {
-                    Error::FailedValidation(self.spec.names[rhs.meta.binding()].clone(), err)
+                .map_err(|invalid| {
+                    // Best-effort redaction using available outcomes, prior to generating an error.
```
psFried, line 86: This was a great idea to redact the documents in the validation errors. I'm wondering under what circumstances we _wouldn't_ be able to redact a field here. Given a document with `{"redactMe": ...}`, it's hard to think of a scenario where the validation outcome wouldn't have the necessary information for redacting `redactMe`, unless you were using conditionals in the schema. I suppose what makes this "best effort" is primarily just that the incoming document could put the sensitive data in a location other than `/redactMe`?

crates/validation/src/capture.rs
```diff
@@ -389,6 +393,19 @@ async fn walk_capture<C: Connectors>(
     }
     let inactive_bindings = live_bindings_spec.values().map(|v| (*v).clone()).collect();
 
+    // Use manual salt if provided, otherwise the live salt, otherwise generate a new one.
```
psFried, line 396: What do you think about requiring a backfill if the salt changes? Kinda seems like if you care enough to set the salt manually, then you'd want to backfill if it changes. Definitely doesn't seem necessary, especially in the short term, though.

crates/doc/src/combine/memtable.rs
jgraettinger: Yea, the kind of case I was thinking of is a redact annotation that applies to (for example) `pattern: .*@.*.com`, but it turns out you have a `.dev`. Or perhaps the `redactMe` subschema itself fails to validate (wrong type or format perhaps). It will apply as broadly as possible without failing-fast, but there may be errors of expression that violate the users's intention, etc.

crates/validation/src/capture.rs
jgraettinger: It's a fairly advanced feature so I don't have a handle on when or why people will use it. Rotation? Alignment with another capture? Matching a migrated table ? :shrug:

PR #2387, opened 2025-09-05
billing: Improve performance by using prefix rollups, and more

jshearer:
**Description:**

-   Only skip invoices with `NoDataMoved` or `NoFullPipeline` if the customer also does not have a payment method on file. If a customer had billable usage, but then deleted all of their captures or materializations, they wouldn't have gotten an invoice prior to this change. If they have a payment method on file, we can and should bill for this legitimate usage.
  - Fix bug where invoices would get skipped if they didn't have an associated Stripe customer in the database
  - Use pre-aggregated `catalog_stats_monthly` table with exact tenant matching for improved query performance
  - Pre-fetch tenant trial start date from db instead of issuing another query per tenant to fetch it. This dramatically speeds things up
  - Filter manual invoices by `date_start` to only include those beginning within the target month
  - Enable `auto_advance=true` during invoice finalization, allowing Stripe to handle automatic charging and retries
-   Manual invoices for the billing period are now correctly identified and included by fetching all manual invoices and performing date filtering clientside. Stripe's API doesn't support date filters in this case :(

### DIFF
--- a/crates/billing-integrations/src/publish.rs
+++ b/crates/billing-integrations/src/publish.rs
@@ -153,8 +153,7 @@ struct Invoice {
     invoice_type: InvoiceType,
     extra: Option<sqlx::types::Json<Option<Extra>>>,
     has_payment_method: Option<bool>,
-    capture_hours: Option<f64>,
-    materialization_hours: Option<f64>,
+    has_full_pipeline: bool,
     payment_provider: PaymentProvider,
     tenant_trial_start: Option<NaiveDate>,
 }
@@ -250,9 +249,7 @@ impl Invoice {
                         return Ok(InvoiceResult::NoDataMoved);
                     }
 
-                    if (self.capture_hours.unwrap_or_default() == 0.0
-                        || self.materialization_hours.unwrap_or_default() == 0.0)
-                        && !matches!(&self.invoice_type, InvoiceType::Manual)
+                    if !self.has_full_pipeline && !matches!(&self.invoice_type, InvoiceType::Manual)
                     {
                         return Ok(InvoiceResult::NoFullPipeline);
                     }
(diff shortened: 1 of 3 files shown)


crates/billing-integrations/src/publish.rs
```diff
@@ -204,36 +208,54 @@ impl Invoice {
...
+                        )
+                        .await?
+                        .and_then(|customer| customer.invoice_settings)
+                        .and_then(|i| i.default_payment_method);
+
+                        if has_payment_method != real_default_payment_method.is_some() {
+                            tracing::warn!(
+                                ?has_payment_method,
+                                stripe_payment_method = real_default_payment_method.is_some(),
+                                "Inconsistent payment method state"
```
jgraettinger, line 232: q: does this log in practice?

crates/billing-integrations/src/publish.rs
```diff
@@ -507,32 +526,34 @@ pub async fn do_publish_invoices(cmd: &PublishInvoice) -> anyhow::Result<()> {
...
                 	group by billed_prefix
                 ) as customer on true
-                inner join lateral(
+                left join lateral(
                 	select
-                		sum(catalog_stats.usage_seconds) filter (where live_specs.spec_type = 'capture') / (60.0 * 60) as capture_hours,
-                    	sum(catalog_stats.usage_seconds) filter (where live_specs.spec_type = 'materialization') / (60.0 * 60)  as materialization_hours
-                    from catalog_stats
-                    join live_specs on live_specs.catalog_name = catalog_stats.catalog_name
+                		sum(catalog_stats_monthly.usage_seconds) filter (where live_specs.spec_type = 'capture') / (60.0 * 60) as capture_hours,
```
jgraettinger: is this right? doesn't this project and sum the singular top-level `catalog_stats_monthly.usage_seconds` across _all_ contained live specs ?

crates/billing-integrations/src/publish.rs
jshearer: Yep! Our stripe capture sometimes doesn't keep up, or something. I haven't looked too deeply into it. It didn't log at all this most recent time running billing, but previously it logged quite a lot.

crates/billing-integrations/src/publish.rs
jshearer: We spoke about this a few 1:1s ago and I ended up refactoring this [here](https://github.com/estuary/flow/pull/2387/commits/87d38e7455542390ecb69c86c0cd04ec806a6d00) as per our convo

jgraettinger (approved):
LGTM

PR #2392, opened 2025-09-08
Phil/graphql updates

psFried:
**Description:**

This rolls up a few big changes to the graphql schema, and a bunch of smaller ones.

The biggest change is the introduction of `LiveSpecRef` to represent a catalog name along with the user's authorized capability to that name. The top-level `liveSpecs` query now returns a paginated list of `LiveSpecRef`s, and that stuct has accessors for `status`, `activeAlerts`, `liveSpec`, etc. The main advantage this affords is that we can serve this top-level `liveSpecs` query from the `Snapshot`, without hitting the database at all. We also avoid needing to fetch the `LiveSpec` (requiring a db query) in cases where all the client wants is something else like the `status`. The `LiveSpec` type is still there, but it is now indirected behind a `liveSpec` field on the `LiveSpecRef`.

I also added some new fields to the `LiveSpec` type for retrieving connected specs:
- `readsFrom`
- `writesTo`
- `readBy`
- `writtenBy`

These all return paginated lists of `LiveSpecRef`s, so you can chain on calls to `status` or `activeAlerts`. The return type of these functions is intentionally the same as the return type of the top-level query. This allows you to re-use query fragments, like in this example, which gets the status of each materialization, in addition to the status of everything that the materialization reads from:

```graphql
fragment Statuses on LiveSpecRefConnection {
  edges {
    node {
      catalogName
      status {
        summary
        type
      }
    }
  }
}

fragment RecurseStatuses on LiveSpecRefConnection {
  edges {
    node {
      liveSpec {
        readsFrom {
          ...Statuses
        }
      }
    }
  }
  ...Statuses
}

query MyQuery {
  liveSpecs(by: {prefix: "acmeCo/", catalogType: materialization}) {
    edges {
      node {
        liveSpec {
          readsFrom {
            ...RecurseStatuses
          }
        }
      }
    }
    ...Statuses 
  }
}
```

(Note graphql does not have _real_ recursion, so this is limited to just a few levels of depth)

**Other stuff**

- Fix the graphql uri in the GraphiQL html
- Rename AuthorizedPrefix to Prefix
- Rename separate `captures`, collections, etc to just `live_specs` that is paginated
- Removed the extra `Node` type from all the Connection types in the schema
- LiveSpecs rename `spec` -> `model`
- Pagination on all the top-level stuff that fetches by prefix
    - alerts
    - live specs
    - prefixes
- Add `isDisabled` to LiveSpecs
- Move graphql client queries out of `flow-client`, into `flowctl` crate
- Rename `firing` to `active` in alerts api
- Updated the error messages that we generate for shard failure and auto-discover alerts to
include more details

### DIFF
--- a/crates/control-plane-api/src/server/public/graphql/live_specs.rs
+++ b/crates/control-plane-api/src/server/public/graphql/live_specs.rs
@@ -200,15 +200,15 @@ impl dataloader::Loader<LiveSpecKey> for super::PgDataLoader {
                 ls.updated_at,
                 case when inputs.with_built then ls.built_spec::text else null end as "built_spec: crate::TextJson<async_graphql::Value>",
                 coalesce(ls.spec->'shards'->>'disable', ls.spec->'derive'->'shards'->>'disable', 'false')::boolean as "is_disabled!: bool",
-                ls.reads_from as "reads_from: Vec<String>",
-                ls.writes_to as "writes_to: Vec<String>",
+                ls.reads_from as "reads_from?: Vec<String>",
+                ls.writes_to as "writes_to?: Vec<String>",
                 case json_typeof(ls.spec->'source')
                 when 'object' then ls.spec->'source'->>'capture'
                 when 'string' then ls.spec->>'source'
                 else null
-                end as "source_capture: String",
-                array_agg(distinct in_flows_specs.catalog_name) filter (where ls.spec_type = 'collection' and in_flows.flow_type = 'capture') as "written_by: Vec<String>",
-                array_agg(distinct out_flows_specs.catalog_name) filter (where ls.spec_type = 'collection') as "read_by: Vec<String>"
+                end as "source_capture?: String",
+                array_agg(distinct in_flows_specs.catalog_name) filter (where ls.spec_type = 'collection' and in_flows.flow_type = 'capture') as "written_by?: Vec<String>",
+                array_agg(distinct out_flows_specs.catalog_name) filter (where ls.spec_type = 'collection' and out_flows.flow_type is not null) as "read_by?: Vec<String>"
             from unnest($1::catalog_name[], $2::boolean[], $3::boolean[]) inputs(name, with_model, with_built)
             join live_specs ls on inputs.name = ls.catalog_name
             left outer join live_spec_flows in_flows on in_flows.target_id = ls.id
--- a/sqlx-data.json
+++ b/sqlx-data.json
@@ -1129,6 +1129,178 @@
     },
     "query": "\n        SELECT\n            d.id AS \"control_id: Id\",\n            d.data_plane_name,\n            d.hmac_keys,\n            d.encrypted_hmac_keys AS \"encrypted_hmac_keys: models::RawValue\",\n            d.data_plane_fqdn,\n            d.broker_address,\n            d.reactor_address,\n            d.ops_logs_name AS \"ops_logs_name: models::Collection\",\n            d.ops_stats_name AS \"ops_stats_name: models::Collection\"\n        FROM data_planes d\n        WHERE id = $1\n        "
   },
+  "3b9e730c56083687f87833d7de187abbe2b7d1060fa8d2e5ea6f9f807763f98b": {
+    "describe": {
+      "columns": [
+        {
+          "name": "catalog_name",
+          "ordinal": 0,
+          "type_info": "Text"
+        },
+        {
+          "name": "with_model!: bool",
+          "ordinal": 1,
+          "type_info": "Bool"
+        },
+        {
+          "name": "with_built!: bool",
+          "ordinal": 2,
+          "type_info": "Bool"
+        },
+        {
+          "name": "live_spec_id: models::Id",
+          "ordinal": 3,
+          "type_info": "Macaddr8"
+        },
+        {
+          "name": "spec_type!: models::CatalogType",
+          "ordinal": 4,
+          "type_info": {
+            "Custom": {
+              "kind": {
+                "Enum": [
+                  "capture",
+                  "collection",
+                  "materialization",
+                  "test"
+                ]
+              },
+              "name": "catalog_spec_type"
+            }
+          }
+        },
+        {
+          "name": "model: crate::TextJson<async_graphql::Value>",
+          "ordinal": 5,
+          "type_info": "Text"
+        },
+        {
+          "name": "last_build_id: models::Id",
+          "ordinal": 6,
+          "type_info": "Macaddr8"
+        },
+        {
+          "name": "last_pub_id: models::Id",
+          "ordinal": 7,
+          "type_info": "Macaddr8"
+        },
+        {
+          "name": "created_at",
+          "ordinal": 8,
+          "type_info": "Timestamptz"
+        },
+        {
+          "name": "updated_at",
+          "ordinal": 9,
+          "type_info": "Timestamptz"
+        },
+        {
+          "name": "built_spec: crate::TextJson<async_graphql::Value>",
+          "ordinal": 10,
+          "type_info": "Text"
+        },
+        {
+          "name": "is_disabled!: bool",
+          "ordinal": 11,
+          "type_info": "Bool"
+        },
+        {
+          "name": "reads_from?: Vec<String>",
+          "ordinal": 12,
+          "type_info": "TextArray"
+        },
+        {
+          "name": "writes_to?: Vec<String>",
+          "ordinal": 13,
+          "type_info": "TextArray"
+        },
+        {
+          "name": "source_capture?: String",
+          "ordinal": 14,
+          "type_info": "Text"
+        },
+        {
+          "name": "written_by?: Vec<String>",
+          "ordinal": 15,
+          "type_info": {
+            "Custom": {
+              "kind": {
+                "Array": {
+                  "Custom": {
+                    "kind": {
+                      "Domain": "Text"
+                    },
+                    "name": "catalog_name"
+                  }
+                }
+              },
+              "name": "_catalog_name"
+            }
+          }
+        },
+        {
+          "name": "read_by?: Vec<String>",
+          "ordinal": 16,
+          "type_info": {
+            "Custom": {
+              "kind": {
+                "Array": {
+                  "Custom": {
+                    "kind": {
+                      "Domain": "Text"
+                    },
+                    "name": "catalog_name"
+                  }
+                }
+              },
+              "name": "_catalog_name"
+            }
+          }
+        }
+      ],
+      "nullable": [
+        false,
+        null,
+        null,
+        false,
+        true,
+        null,
+        false,
+        false,
+        false,
+        false,
+        null,
+        null,
+        true,
+        true,
+        null,
+        null,
+        null
+      ],
+      "parameters": {
+        "Left": [
+          {
+            "Custom": {
+              "kind": {
+                "Array": {
+                  "Custom": {
+                    "kind": {
+                      "Domain": "Text"
+                    },
+                    "name": "catalog_name"
+                  }
+                }
+              },
+              "name": "_catalog_name"
+            }
+          },
+          "BoolArray",
+          "BoolArray"
+        ]
+      }
+    },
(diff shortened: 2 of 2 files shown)


crates/agent/src/integration_tests/harness.rs
```diff
@@ -1316,6 +1320,102 @@ impl TestHarness {
...
+            self.pool.clone(),
+            self.publisher.clone(),
+        ));
+
+        // Start the snapshot fetch loop
+        let app_clone = app.clone();
+        tokio::spawn(control_plane_api::server::snapshot::fetch_loop(app_clone));
+
+        // Give it time to load initial snapshot
+        tokio::time::sleep(tokio::time::Duration::from_millis(500)).await;
```
jgraettinger: nit: the sleeps in here are a smell.

We may want a `tokio::sync::watch` for observing snapshot updates.

crates/control-plane-api/src/server/public/graphql/live_spec_refs.rs
```diff
@@ -0,0 +1,242 @@
...
+}
+
+/// Represents a reference from one live spec to another.
+#[derive(Debug, Clone, SimpleObject)]
+#[graphql(complex)]
+pub struct LiveSpecRef {
+    /// The catalog_name of the referent.
+    pub catalog_name: models::Name,
+    /// The current user's capability to the referent. None indicates no access.
+    /// Note that
```
jgraettinger: nit: cut off comment

crates/control-plane-api/src/server/public/graphql/live_spec_refs.rs
```diff
@@ -0,0 +1,242 @@
...
+
+#[ComplexObject]
+impl LiveSpecRef {
+    /// Returns the live spec that the reference points to, if the user has access to it.
+    async fn live_spec(
+        &self,
+        ctx: &Context<'_>,
+    ) -> async_graphql::Result<Option<live_specs::LiveSpec>> {
+        if self.user_capability.is_none() {
+            // return Err(async_graphql::Error::new(
```
jgraettinger: nit: cleanup?

crates/control-plane-api/src/server/public/graphql/live_spec_refs.rs
```diff
@@ -0,0 +1,242 @@
...
+        let live_spec = loader.load_one(key).await?.ok_or_else(|| {
+            async_graphql::Error::new(format!("no live spec found for {}", self.catalog_name))
+        })?;
+        Ok(Some(live_spec))
+    }
+
+    /// Returns all alerts that are currently firing for this live spec.
+    async fn active_alerts(&self, ctx: &Context<'_>) -> async_graphql::Result<Vec<alerts::Alert>> {
+        if self.user_capability.is_none() {
+            tracing::info!(catalog_name = %self.catalog_name, "not showing firing_alerts because user is not authorized");
```
jgraettinger: nit: do we need these? it could get noisy. Or is this left-over from testing?

crates/control-plane-api/src/server/public/graphql/live_spec_refs.rs
```diff
@@ -0,0 +1,242 @@
...
+    ///
+    /// Note that the `user_capability` that's returned as part of the reference
+    /// represents the user's capability to the whole prefix, and it is possible
+    /// that there are more specific grants for a broader capability. In other
+    /// words, this capability represents the _minimum_ capability that the user
+    /// has for the given spec.
+    pub async fn live_specs(
+        &self,
+        ctx: &Context<'_>,
+        by: ByPrefix,
```
jgraettinger: if we wanted to also support fetching by `id`, how would we extend this?

What about fetching a specific name?

crates/control-plane-api/src/server/snapshot.rs
```diff
@@ -246,6 +248,56 @@ impl Snapshot {
         Err(Ok(backoff.to_std().unwrap()))
     }
 
+    /// Returns an iterator of tasks matching the given prefix. If `after_task` is provided,
+    /// only tasks with names greater than `after_task` will be returned.
+    pub fn list_tasks<'a, 'p: 'a>(
```
jgraettinger: let's talk this out in person, but: I think this needs to go to the database because it otherwise leads to violations of read-your-writes consistency.

crates/control-plane-api/src/server/snapshot.rs
```diff
@@ -246,6 +248,56 @@ impl Snapshot {
...
+
+        slice
+            .iter()
+            .map(move |task_idx| &tasks[*task_idx])
+            .take_while(move |task| task.task_name.starts_with(prefix.as_str()))
+    }
+
+    /// Returns an iterator of collections matching the given prefix. If `after_collection` is provided,
+    /// only collections with names greater than `after_collection` will be returned.
+    pub fn list_collections<'a, 'p: 'a>(
```
jgraettinger: same here.

crates/flow-client/control-plane-api.graphql
```diff
@@ -56,26 +92,90 @@ type AlertEdge {
...
+	If auto-discovery has failed, this will include information about that failure.
+	This field is cleared as soon as a successful auto-discover is run.
+	"""
+	failure: AutoDiscoverFailure
+}
+
+"""
+Input type for returning live specs references by prefix and catalog type.
+"""
+input ByPrefix {
```
jgraettinger: The ByPrefix name seems to constrain us, from adding additional means of pushing down predicates. I had originally proposed:

```
type LiveSpecBy {
  id: ID
  prefixes: [String!]
  captures: bool
  collections: bool
  materializations: bool
  test: bool
}
```

Not married to this, but we need to account for being able to add look-ups by `id`, and there may be future predicates we want to support.

crates/flow-client/control-plane-api.graphql
```diff
@@ -97,42 +197,245 @@ enum CatalogType {
...
+	"""
+	A cursor for use in pagination
+	"""
+	cursor: String!
+}
+
+"""
+Represents an optimistic lock failure when trying to update live specs.
+"""
+type LockFailure {
```
jgraettinger, line 447: This seems like dirty laundry that shouldn't be in the public API. What's the rationale for having it?

crates/flow-client/control-plane-api.graphql
```diff
@@ -155,42 +458,355 @@ type PageInfo {
...
-		prefixes: [String!]!
-	): [Alert!]!
-	authorizedPrefixes(minCapability: Capability!): [AuthorizedPrefix!]!
+		prefix: String!,
+		"""
+		Optionally filter alerts by whether or not they are active
+		"""
+		active: Boolean,		before: String,		last: Int
+	): AlertConnection!
+	prefixes(minCapability: Capability!, after: String, first: Int): PrefixConnection!
```
jgraettinger: Similar: `by: ByPrefix!` predicates ?

crates/flow-client/control-plane-api.graphql
```diff
@@ -155,42 +458,355 @@ type PageInfo {
...
 		Show alerts for the given catalog prefixes
 		"""
-		prefixes: [String!]!
-	): [Alert!]!
-	authorizedPrefixes(minCapability: Capability!): [AuthorizedPrefix!]!
+		prefix: String!,
+		"""
+		Optionally filter alerts by whether or not they are active
+		"""
+		active: Boolean,		before: String,		last: Int
```
jgraettinger: Let's discuss having a standard of a single / universal `by` argument that's passed to queries alongside relay parameters, as `foo(by: FooBy, before: String, last: Int): FooConnection!`

So in this case we'd have:
```
struct AlertsBy {
  prefix: String!,
  active: Boolean,
}
```

I'd like to think through / understand pros / cons of this vs adding kitchen-sink predicates directly into the query. I have a knee-jerk expectation this will help future extensibility, but I'm not certain it makes a difference.

crates/flow-client/control-plane-api.graphql
```diff
@@ -155,42 +458,355 @@ type PageInfo {
...
+	): AlertConnection!
+	prefixes(minCapability: Capability!, after: String, first: Int): PrefixConnection!
+}
+
+"""
+The shape of a connector status, which matches that of an ops::Log.
+"""
+type ShardFailure {
```
jgraettinger, line 622: Is it possible to have ShardFailure / ShardRef etc all be TaskFailure / TaskRef / TaskStatusCheck / etc ?

I'm sure that's a larger refactor so perhaps done separately...

crates/models/src/status/capture.rs
```diff
@@ -66,7 +69,10 @@ pub struct AutoDiscoverOutcome {
     /// Errors that occurred during the discovery or evolution process.
     #[serde(default, skip_serializing_if = "Vec::is_empty")]
     pub errors: Vec<draft_error::Error>,
-    /// Collections that were re-created due to the collection key having changed.
+    /// Deprecated. Evolution is now handled as part of the build, and results
+    /// are reported as part of the `model_fixes`, which are added to the `detail`
+    /// in `publication_specs`.
+    #[cfg_attr(feature = "async-graphql", graphql(skip))]
     #[serde(default, skip_serializing_if = "Vec::is_empty")]
```
jgraettinger: nit: can we remove?

crates/flow-client/control-plane-api.graphql
jgraettinger: over vc, determined that this also carries expectPubId violations, so we just need to update comment to clarify

crates/agent/src/integration_tests/harness.rs
psFried: I refactored how we deal with snapshots in tests, so that we refresh the snapshot once, synchronously, before executing the first graphql query. So there's no longer any need to run the fetch loop, and tests get deterministic behavior around refreshes.

crates/control-plane-api/src/server/public/graphql/live_spec_refs.rs
psFried: fixed

crates/control-plane-api/src/server/public/graphql/live_spec_refs.rs
psFried: I went ahead and removed these. You're right, it could get pretty spammy.

crates/control-plane-api/src/server/public/graphql/live_spec_refs.rs
psFried: I added support for fetching by name, and I think ids would work essentially the same way. The `LiveSpecsBy` type is a `@oneOf`, so the query now looks like either:

- `liveSpecs(by: { prefixAndType: { prefix: "foo/", catalogType: capture } })`
- `liveSpecs(by: { names: [ "foo/capture" ] })`

crates/control-plane-api/src/server/snapshot.rs
psFried: We talked over VC and decided to just fetch live specs from the database instead of the snapshot.

crates/flow-client/control-plane-api.graphql
psFried: I updated the comment to reflect that these are due to expectPubId mismatches

crates/flow-client/control-plane-api.graphql
psFried: updated

crates/flow-client/control-plane-api.graphql
psFried: :+1: I've updated all the queries to use `by` for any filters.

crates/control-plane-api/src/server/public/graphql/live_spec_refs.rs
```diff
@@ -11,21 +11,35 @@ use crate::server::{
...
 /// Input type for returning live specs references by prefix and catalog type.
 #[derive(Debug, Clone, async_graphql::InputObject)]
-pub struct ByPrefix {
+pub struct ByPrefixAndType {
     pub prefix: models::Prefix,
     pub catalog_type: models::CatalogType,
 }
 
+#[derive(Debug, Clone, async_graphql::OneofObject)]
+pub enum LiveSpecsBy {
```
jgraettinger, line 24: :heart:

crates/flow-client/control-plane-api.graphql
```diff
@@ -415,8 +433,16 @@ type LiveSpecRefEdge {
...
+input LiveSpecsBy @oneOf {
+	prefixAndType: ByPrefixAndType
+	names: [Name!]
+}
+
 """
-Represents an optimistic lock failure when trying to update live specs.
+Represents an optimistic lock failure when trying to update live specs. This
+typically means that an `expectPubId` was not matched after a successful
+build, meaning that another publication may have changed the spec while this
```
jgraettinger: nit: perhaps:

Represents an optimistic lock failure when trying to update a specification.
This typically means that an `expectPubId` was not matched, implying that another
publication has applied a conflicting update to this specification since it was last fetched.


I'm trying to capture that this can also happen because you're working with an outdated view and need to `flowctl catalog pull-specs`.

jgraettinger (approved):
LGTM! This looks awesome

PR #2393, opened 2025-09-09
Docs: Snowflake.md: Make TIMESTAMP_TYPE_MAPPING configuration more prominent

jwhartley:
We get a lot of questions / issues around snowflake timestamps, making `TIMESTAMP_TYPE_MAPPING` configuration and docs more prominent.

### DIFF
--- a/site/docs/reference/Connectors/materialization-connectors/Snowflake.md
+++ b/site/docs/reference/Connectors/materialization-connectors/Snowflake.md
@@ -75,8 +75,8 @@ grant CREATE SCHEMA, MONITOR, USAGE on database identifier($database_name) to ro
 use role ACCOUNTADMIN;
 grant CREATE INTEGRATION on account to role identifier($estuary_role);
 use role sysadmin;
--- Use `TIMESTAMP_LTZ` or `TIMESTAMP_TZ`
-ALTER USER $estuary_user SET TIMESTAMP_TYPE_MAPPING = 'TIMESTAMP_LTZ';
+-- Optional: Set the timestamp type to `TIMESTAMP_TZ`; otherwise defaults to using `TIMESTAMP_LTZ`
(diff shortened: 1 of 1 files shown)


site/docs/reference/Connectors/materialization-connectors/Snowflake.md
```diff
@@ -72,6 +75,8 @@ grant CREATE SCHEMA, MONITOR, USAGE on database identifier($database_name) to ro
 use role ACCOUNTADMIN;
 grant CREATE INTEGRATION on account to role identifier($estuary_role);
 use role sysadmin;
+-- Use `TIMESTAMP_LTZ` or `TIMESTAMP_TZ`
+ALTER USER $estuary_user SET TIMESTAMP_TYPE_MAPPING = 'TIMESTAMP_LTZ';
```
aeluce: In general, if I understand correctly, there's no need to set this, since we'll default to LTZ anyway. Just that we need to make users aware of what we're doing. I'm not sure whether adding the setting to a script users may simply copy/paste will help. Perhaps if we framed it as an option?

```
-- Optional: Set the timestamp type to `TIMESTAMP_TZ`; otherwise defaults to using `TIMESTAMP_LTZ`
-- ALTER USER $estuary_user SET TIMESTAMP_TYPE_MAPPING = 'TIMESTAMP_TZ';
```

I'll leave it to your judgment which will cause fewer headaches! :grin:

aeluce (approved):
Just a comment for your consideration. Otherwise looks good, thanks!

site/docs/reference/Connectors/materialization-connectors/Snowflake.md
jwhartley: Great point, will make it an option

PR #2396, opened 2025-09-10
dekaf: Fix `OffsetCommitResponse` topic names

jshearer:
**Description:**

Topic names are encrypted when sent to upstream Kafka for group management requests. In other places, we map the topic names in proxied responses back to their plaintext value before sending the response along to the client. It appears we forgot to do that for `OffsetCommitResponse`. We got a report from a user of `franz-go` that it's complaining about this with the following message:

```
broker replied to our OffsetCommitRequest incorrectly! Topic at request index 0: <plaintext_topic_name>, reply at index: <encrypted_topic_name>; num partitions on request topic: 1, in reply: 1, we cannot handle this!
```


jgraettinger (approved):
LGTM

PR #2398, opened 2025-09-10
alerts: fix deployment action for alerts cloudrun service

psFried:
The sender address was originally setup as a secret, so it fails to deploy with the previous yaml, which specified that as a regular env variable.  This updates it to use the secret.


.github/workflows/deploy-alert-api.yaml
```diff
@@ -37,10 +37,10 @@ jobs:
           timeout: 1m
           env_vars: |-
             ALERT_REPLY_TO_ADDRESS=support@estuary.dev
-            RESEND_EMAIL_ADDRESS=alerts@email.estuary.dev
           secrets: |-
             RESEND_API_KEY=ALERTS_RESEND_API_KEY:latest
             ALERT_EMAIL_FUNCTION_SECRET=ALERT_EMAIL_FUNCTION_SECRET:latest
+            RESEND_EMAIL_ADDRESS=ALERTS_SENDER_EMAI:latest
```
jshearer: Typo?

`ALERTS_SENDER_EMAI:latest`

psFried: ooh good catch, thanks!

PR #2400, opened 2025-09-10
dekaf: Refresh `MaterializationSpec`s more frequently

### LINKED ISSUE #2395: dekaf: Refresh task spec more frequently
Since https://github.com/estuary/flow/pull/2348 changed `TaskManager`'s behavior to only call `/authorize/dekaf` ~5 minutes before token expiration, we now only fetch updated task specs that frequently. This is a bug, we need to be refreshing task specs much more frequently than that.

jshearer:
**Description:**

Since #2348 changed `TaskManager`'s behavior to only call `/authorize/dekaf` ~5 minutes before token expiration, we would only fetch updated task specs that frequently. This is a bug, we need to be refreshing task specs much more frequently than that.

So instead, let's set a fairly short upper bound on how stale materialization specs can be and start trying to refresh after that passes.

Fixes #2395

### DIFF
--- a/crates/dekaf/src/task_manager.rs
+++ b/crates/dekaf/src/task_manager.rs
@@ -50,6 +50,8 @@ pub type Result<T> = core::result::Result<T, SharedError>;
 const TASK_TIMEOUT: Duration = Duration::from_secs(60 * 3);
 /// How long before the end of an access token should we start trying to refresh it
 const REFRESH_START_AT: Duration = Duration::from_secs(60 * 5);
+/// How long to cache a MaterializationSpec before re-fetching it, even if the token is still valid.
+const SPEC_TTL: Duration = Duration::from_secs(60 * 2);
 
 #[derive(Clone)]
 pub enum TaskState {
@@ -321,6 +323,7 @@ impl TaskManager {
                         ops_logs_journal,
                         ops_stats_journal,
                         spec,
+                        ..
                     } => {
                         // Continue with normal processing
                         partitions_and_clients = update_partition_info(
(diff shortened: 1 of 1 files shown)


crates/dekaf/src/task_manager.rs
```diff
@@ -642,22 +645,32 @@ pub enum DekafTaskAuth {
...
-    fn exp(&self) -> u64 {
+    fn exp(&self) -> anyhow::Result<time::OffsetDateTime> {
         match self {
             DekafTaskAuth::Redirect { fetched_at, .. } => {
                 // Redirects are valid for 10 minutes
-                fetched_at.unix_timestamp() as u64 + 60 * 10
+                Ok(*fetched_at + Duration::from_secs(60 * 10))
+            }
+            DekafTaskAuth::Auth { claims, .. } => {
+                time::OffsetDateTime::from_unix_timestamp(claims.exp as i64).map_err(|e| e.into())
```
jgraettinger, line 660: note that realistically, SPEC_TTL is the only bound that matters here.

Dekaf is assured that `exp` will far exceed the values that we're looking at, because Snapshot::evaluate() lower-bounds to half of the maximum authorization time (so, currently 40 minutes).

crates/dekaf/src/task_manager.rs
```diff
@@ -691,7 +704,9 @@ async fn get_or_refresh_dekaf_auth(
         {
             Ok(resp) => resp,
             Err(_) => {
-                if time::OffsetDateTime::now_utc().unix_timestamp() < cached_auth.exp() as i64 {
+                // This isn't checking SPEC_TTL, so it will potentially hand out
```
jgraettinger, line 707: hopefully you're intending to refactor already, but noting that `fetch_dekaf_task_auth` being run in the tokio timeout appears to be a thundering herd waiting to happen under load. That's because `fetch_dekaf_task_auth` may just only be able to fire off a request for an authorization before getting timed out. The client moves on and doesn't care, but the upstream server still needs to deal with the flurry of /authorize/dekaf requests, which are expensive and certain to get slower under a timeout storm.

https://docs.rs/tokio/latest/tokio/sync/watch/index.html may be a good fit here? Dekaf would periodically fetch dekaf authorization in a tokio task _without_ a timeout, and then broadcast through the watch. Receivers are able to grab the current value without waiting, or wait for a next value if they need to (e.x. because a periodic refresh couldn't complete before expiration, perhaps due to migration, and they need to await a next refresh).

jgraettinger (approved):
LGTM

crates/dekaf/src/task_manager.rs
jshearer: Indeed, I guess I could have a comment here to that effect. Even though I know that as currently configured, `REFRESH_START_AT` won't come into play for this at all, that felt implicit enough when I was writing this that I wanted to cover as many bases as possible.

crates/dekaf/src/task_manager.rs
jshearer: Yeah... I think that's a good point WRT the thundering herd. Like you say, it seems like the right solution here is another concurrent task that's purely responsible for fetching and making available a relatively fresh response from `/authorize/dekaf`. "Relatively fresh" means to start fetching once the age exceeds `SPEC_TTL`, and broadcast the update thereafter as soon as a response is returned.

I _do_ still think that the individual requests to `/authorize/dekaf` need to be wrapped in a timeout though. That was the original cause of the problem: transient network conditions (maybe we're actually not setting keepalive somewhere, or something like that?) caused the request to hang indefinitely, so I feel like the behavior in that case should be a timeout with a backoff-retry, which also still takes into account a "please retry no sooner than X seconds" response like it does today.

PR #2401, opened 2025-09-10
dekaf: Fix refreshing MSK tokens

jshearer:
**Description:**

The intent was always to refresh tokens once they have less than 30s of life left, but the existing code was refreshing once they have less than 30 _milliseconds_ of life left. If the refresh took >30ms, which is entirely plausible, the Kafka client would attempt to use an expired token, get an authorization error, and pass that along to the client which would log noisily and make people think there's something wrong.


jgraettinger (approved):
LGTM

PR #2407, opened 2025-09-19
dekaf: Fix missed field folding

jshearer:
**Description:**

Field folding was applied to the `avro::RecordField.name`, but not applied to the avro schema's location itself.

Without the fix, we see a similar error to the one observed in production:
```
running 1 test
test test_field_selection_recommended_fields ... FAILED

successes:

successes:

failures:

---- test_field_selection_recommended_fields stdout ----

thread 'test_field_selection_recommended_fields' panicked at crates/avro/src/schema.rs:251:28:
called `Result::unwrap()` on an `Err` value: Invalid schema name root.invoice/additional_charges._items._flow_extra.RawJSON. It must match the regex '^((?P<namespace>([A-Za-z_][A-Za-z0-9_]*(\.[A-Za-z_][A-Za-z0-9_]*)*)?)\.)?(?P<name>[A-Za-z_][A-Za-z0-9_]*)$'
```

### DIFF
--- a/crates/dekaf/tests/snapshots/field_extraction_tests__field_selection_recommended_fields.snap
+++ b/crates/dekaf/tests/snapshots/field_extraction_tests__field_selection_recommended_fields.snap
@@ -1,5 +1,5 @@
 ---
-source: crates/dekaf/tests/dekaf_schema_test.rs
+source: crates/dekaf/tests/field_extraction_tests.rs
 expression: output?
 ---
 Record(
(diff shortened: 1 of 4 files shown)


crates/dekaf/tests/snapshots/field_extraction_tests__field_selection_recommended_fields.snap
```diff
@@ -1,5 +1,5 @@
 ---
-source: crates/dekaf/tests/dekaf_schema_test.rs
+source: crates/dekaf/tests/field_extraction_tests.rs
```
jshearer, line 2: I guess this snapshot was created and then the test file was since renamed. This must not be asserted when validating the snapshot 🤔

jgraettinger (approved):
LGTM

PR #2413, opened 2025-09-23
agent: fix controller retries of inferred schema updates

psFried:
We've observed a bug in prod where a publication to update the inferred schema fails, but the controller's next attempt is scheduled for ~4 hours out, even after the first failure.  The issue here was a `.do_not_retry()?` in the collection controller, which meant that it fell back to the 4 hour periodic check rather than attempting to re-try.  But the larger issue was that this behavior was not covered by tests, so I added an integration test for it.

### DIFF
--- a/crates/agent/src/integration_tests/inferred_schemas.rs
+++ b/crates/agent/src/integration_tests/inferred_schemas.rs
@@ -1,0 +1,260 @@
+use super::harness::{
+    draft_catalog, get_collection_generation_id, mock_inferred_schema, TestHarness,
+};
+use crate::{controllers::ControllerState, integration_tests::harness::InjectBuildError};
+use chrono::{DateTime, Utc};
+use serde_json::json;
+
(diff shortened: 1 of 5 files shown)


crates/agent/src/integration_tests/inferred_schemas.rs
```diff
@@ -0,0 +1,260 @@
+use super::harness::{
+    draft_catalog, get_collection_generation_id, mock_inferred_schema, TestHarness,
+};
+use crate::{controllers::ControllerState, integration_tests::harness::InjectBuildError};
+use chrono::{DateTime, Utc};
+use serde_json::json;
+
+#[tokio::test]
+#[serial_test::serial]
+async fn test_inferred_schema_updates() {
```
jshearer, line 10: These are really nice integration tests!

jshearer (approved):
LGTM

PR #2419, opened 2025-09-26
dekaf: Make `KafkaApiClient` more resilient to transient errors

jshearer:
**Description:**

We were seeing errors like this in the logs, causing sessions to crash unexpectedly:

```
Failed to increase partition count for topic ...: Some(TopicAuthorizationFailed)
```
and
```
Failed to increase partition count topic="..." error=Some(InvalidPartitions) Topic already has 1 partition(s).
```

These are caused by race conditions either originating from auto-topic creation, or multiple consumer sessions both attempting the same action. They aren't functionally meaningful, as a subsequent retry will behave correctly, but they do cause session drops which result in consumers logging noisy warnings about unexpected disconnects.

### DIFF
--- a/crates/dekaf/src/api_client.rs
+++ b/crates/dekaf/src/api_client.rs
@@ -525,18 +525,16 @@ impl KafkaApiClient {
     }
 
     #[instrument(skip_all)]
-    async fn increase_partition_counts(
-        &mut self,
-        topics: Vec<(messages::TopicName, usize)>,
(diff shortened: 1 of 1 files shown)


jshearer:
Tested locally and in `c5` dataplane, doesn't break anything and the errors in question are fixed.

jgraettinger (approved):
LGTM % refactoring out retry loop as discussed over VC

PR #2421, opened 2025-09-29
perf: 3x faster JSON Schema validation and combiner key prefix optimization

jgraettinger:
**Description**

This PR refactors JSON Schema validation to work directly with the AsNode trait instead of walking serde deserializers, resulting in significantly improved performance (~3x or more, depending on the schema). The AsNode trait and Pointer implementation have been migrated from the doc crate to the json crate as their more appropriate home.

Key improvements:
  - Complete reimplementation of JSON Schema validation using a cache-friendly, single-pass algorithm
  - Updated to JSON Schema Draft 2020-12 standard with backward compatibility
  - doc::Validator now handles document redaction on the error path internally, simplifying MemTable and providing a stronger guarantee we cannot forget to do this.
  - A new Combiner optimization embeds a prefix of the document key, which greatly reduces random memory reads that are otherwise required during sorting / merging (as observed by perf)
    - This may be **just** as important as faster validation, in very large transactions
  - Support for push-down predicates to track specific annotations during validation

**Workflow steps:**

  This is an internal refactoring with no changes to external APIs or workflows. All existing functionality is preserved while providing significant
  performance improvements under the hood.

**Notes for reviewers:**

  - The largest changes are in the json crate where validation has been completely reimplemented. The new implementation eliminates O(N²) behavior when validating objects with many schematized properties, and focuses on cache efficiency and avoiding all unnecessary work.
  - The doc crate changes primarily involve removing the now-migrated AsNode/Pointer code and updating to use the json crate versions. The Validator interface has been refreshed to handle redaction internally and support annotation collection predicate push-down.
  - The sources crate's load_schema_node() was significantly impacted since Keyword::Application was removed in favor of flattening all JSON Schema keywords. The function now explicitly handles all keyword types.
  - The combiner optimization (last commit) embeds the first 13 bytes of extracted key tuples directly in metadata, reducing memory fetches during sorting/merging operations.
  - All other changes are mechanical updates to use the new interfaces.
  - Benchmarks show ~3x improvement on citi_bike and github event validation. I suspect this undersells the improvement, as these benches easily fit in L3 cache.
  - Using `perf` with `flowctl raw combine` shows that schema validations and combiner reductions are about 5% of total time in test data-sets I've tried. It's dominated by simdjson parsing and LZ4 compress / decompress.


crates/doc/src/heap.rs
```diff
@@ -122,16 +127,120 @@ impl<'alloc> HeapNode<'alloc> {
...
+                    break (false, 0);
+                }
+                HeapNode::Null => unreachable!("null already handled"),
+            };
+        };
+
+        // Walk back up the stack, adjusting tape lengths as we go.
+        for (tape_length, local_delta) in stack.into_iter().rev() {
+            built_delta += local_delta;
+            *tape_length += built_delta;
```
psFried, line 228: 🤯

crates/json/src/validator/mod.rs
```diff
@@ -0,0 +1,196 @@
...
+pub struct Validator<'s, A>
+where
+    A: Annotation,
+{
+    index: &'s schema::Index<'s, A>,
+    stack: Vec<Frame<'s, A>>,
+    active: Vec<u32>,
+}
+
+struct Frame<'s, A>
```
psFried, line 60: Just to double check my understanding: Frame represents the context for visitation of the JSON nodes at a particular JSON location?

crates/doc/src/heap.rs
jgraettinger: This was actually just moved from ptr.rs in this same crate, which was _mostly_ moved to `json` except for this routine.

But yea, that was part of the tape-length optimization work.

crates/json/src/validator/mod.rs
jgraettinger: Mostly. A Frame is conceptually like a stack frame, and is driving at the idea that we're concurrently "executing" multiple different functions through a single pass walk of the document.

Concretely a Frame is the combination of a JSON schema being evaluated, and a document node at which it's being evaluated. Many Frames can be active because each sub-schema of (for example) an `anyOf` keyword gets its own Frame which is individually tracked with respect to validation status.

PR #2424, opened 2025-09-30
agent: guard against null bytes in controller errors

psFried:
We've seen a case where controllers fail to update their status due to null bytes being included in one of the columns.  I've not yet been able to pin down exactly how that's happening, but wanted to add a guard against null bytes in the error message, as that seems like the most likely suspect.  Also add some xtra logging to help pin down the issue.

### DIFF
--- a/crates/agent/src/controllers/executor.rs
+++ b/crates/agent/src/controllers/executor.rs
@@ -98,7 +98,7 @@ impl automations::Outcome for Outcome {
             status,
             next_run,
             failures,
-            error,
+            mut error,
             live_spec_deleted,
         } = self;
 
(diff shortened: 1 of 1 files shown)


crates/agent/src/controllers/executor.rs
```diff
@@ -114,7 +114,13 @@ impl automations::Outcome for Outcome {
             return Ok(Action::Done);
         }
 
-        control_plane_api::controllers::update_status(
+        // Guard against any null bytes in an error message, which would be disallowed by postgres.
+        if error.as_ref().is_some_and(|e| e.contains('\0')) {
```
jgraettinger, line 118: related: https://github.com/estuary/flow/blob/master/crates/agent/src/logs.rs#L50-L61

Re-use? Also, should this be in `update_status` rather than here?

Any other places we're storing errors or arbitrary Rust Strings in non-JSON text columns?

psFried: I'd taken a while to respond to this, partly because I'd hoped that we might not need to merge this PR at all. The main goal is to identify the specific issue that had caused a controller update failure, because I'm not actually sure that null bytes are even the issue. I went ahead and merged this as-is so we can release the `ops/` validation fix without losing this logging.

I didn't want to re-use the logs function because I wanted to add some logging and I worried that it'd be too noisy to add it there. And wanted to keep the logging all here in one place, rather than move some into `update_status`. We _shouldn't_ have any null bytes in the controller status, so if that's actually the problem I'd hope that we can eliminate them at the source rather than doing a last-ditch sanitization.

PR #2436, opened 2025-10-06
validation: adjust Reject ordering

jgraettinger:
Reject::ConnectorIncompatible must be lower-rank than Reject::DuplicateFold.

If two fields are ambiguous, we want to reject the second field due to the DuplicateFold instead of ConnectorIncompatible. The former will cause the field to simply not be included, while the latter will trigger the onIncompatibleSchemaChange action.

**Workflow steps:**

**Notes for reviewers:**


psFried (approved):
LGTM
That test case is 🤌

PR #2437, opened 2025-10-06
Phil/flowctl status

psFried:
**Description:**

Converts a selection of `flowctl catalog` subcommands to use our new graphql API. This also rolls in a few changes to our graphql schema. The goal was to some usage of our new graphql APIs so we can see what breaking changes we'd want to make before deploying the schema to production. I'd say that's been achieved, given that most of the schema changes here are relatively minor.

Best to review this commit by commit.

**Workflow steps:**
**Breaking changes to CLI arguments**

The commit to change `flowctl catalog list|delete|pull-specs` makes a few minor breaking changes to the CLI arguments. These changes probably won't affect most users too much, but we should try to clearly communicate it regardless. Details from the commit message:

The arguments for selecting specs by catalog type now only permit a single catalog type to be selected.  They also no longer accept explicit boolean values, so for example `--captures=true` would now be passed as just `--captures`.

The `--data-plane-name` argument now only accepts a single value, where previously it could be passed multiple times.

Both of these changes were made because they were infrequently used corners of the API, and it didn't seem worth the extra code complexity to support them.

Lastly, it is now required that you pass a `--prefix` or `--name` argument if the current user is authorized to more than 5 different catalog prefixes. This is because our graphql API requires a prefix selector when listing live specs.  When running `flowctl catalog list|pull-specs` without any `--name` or `--prefix` argument, flowctl will first fetch a list of the user's authorized roles. If there are fewer than 5 of them (after removing redundant nested roles), then it behaves as if the user had passed each of them as a `--prefix` argument. For the cosmopolitan users with access to more than 5 distinct prefixes, you'll be required to pass an explicit `--prefix` argument.


crates/control-plane-api/src/server/public/graphql/live_spec_refs.rs
```diff
@@ -16,7 +16,10 @@ const DEFAULT_PAGE_SIZE: usize = 50;
 #[derive(Debug, Clone, async_graphql::InputObject)]
 pub struct ByPrefixAndType {
     pub prefix: models::Prefix,
-    pub catalog_type: models::CatalogType,
+    /// Optionally filter by catalogType
```
jgraettinger, line 22: As a general pattern, should we flatten `LiveSpecsBy` into a flat set of predicates like:

```
pub struct LiveSpecsBy {
  pub prefix: Option<models::Prefix>,
  pub names: Vec<models::Name>,
  pub catalog_type: Option<models::CatalogType>,
  pub data_plane_name: Option<models::Name>,
}
```

Then, the rule is that the conjunction of **all** presented predicates must be true. You can provide an explicit list of names, and a data_plane_name, and only those rows which are satisfied by _both_ are returned.

Or you can provide names with a catalog type, and just get the sub-set of named specs that match the desired type.

Or you can provide a prefix and names, and while it might be silly, we'll only return rows where both are true.

While this allows for some query combinations that don't make a lot of sense, it also avoids our needing to guess ahead of time what sets of combos are potentially useful and might show up one day.

This feels more scalable and open for extension from an API ergonomics perspective to me, but open for debate?

crates/control-plane-api/src/server/public/graphql/mod.rs
```diff
@@ -170,3 +171,16 @@ pub async fn graphql_graphiql() -> impl axum::response::IntoResponse {
...
     )
 }
+
+// It's critical that we preserve the order of fields within the endpoint config
+// of task specs, or we'll invalidate sops encrypted configs. This property is
+// guaranteed by the use of `IndexMap` to represent objects inside an
+// `async_graphql::Value`. This test ensures that we'll catch it if that ever
+// changes.
+#[test]
+fn graphql_json_values_preserves_field_order() {
```
jgraettinger, line 181: nice :+1:

crates/flowctl/src/catalog/status.rs
```diff
@@ -1,45 +1,147 @@
...
-#[serde(transparent)]
-pub struct StatusOutput(StatusResponse);
+#[derive(serde::Serialize, Debug)]
+#[serde(rename_all = "camelCase")]
+struct StatusRow {
+    catalog_name: models::Name,
+    catalog_type: Option<models::CatalogType>,
+    live_spec_updated_at: Option<DateTime>,
+    status: Option<connected_status_query::SelectStatus>,
+    // User capability is here in case it helps differentiate cases where specs are deleted vs
```
jgraettinger: nit: incomplete comment

jgraettinger (commented):
This looks really well done. One important question to discuss w.r.t. the query predicate for LiveSpecsBy and other future predicates, but everything else is :+1:

crates/control-plane-api/src/server/public/graphql/live_spec_refs.rs
psFried: Personally, I feel pretty good about keeping the `@oneOf`. I don't see any reason why it _needs_ to be that way, but the advantages I'm seeing are:

- Intended use is clearer for users/readers of the schema
- Saves us from needing to do more structural validation of the query input, which can't otherwise be expressed in the schema
- Saves us from having to actually implement filtering by `catalog_type` or `data_plane_name` when selecting specs by name

I don't feel super strongly about the benefits of `@oneOf`, but I also don't really see a big benefit to _not_ using it. The way I see it, `@oneOf` ought to be a little easier to keep clean as we add new filters and such, because the schema can encode which ones are valid to pass along with each other. The downside is that, say we add a filter that could be applied to _both_ `ByName` and `ByPrefixAndType`, then we'd need to add it to both structs. But I also see that as an upshot, because I imagine it'll be clearer to users which predicates can go together. It's hard to say, though. I'd be OK with switching to a flat struct if you feel like we'll regret `@oneOf`.

jgraettinger: Let's sync quickly tomorrow? I don't have a handle on how painful it would be to change `@oneOf` if we start to regret it.

Assorted thoughts:
* My question is motivated by a lack of confidence that I understand what intended usage will be a year or two from now.
* I'm assuming we wouldn't bother to do structural validation of co-presented predicates. Garbage in, garbage out.
* An essential assumption is that we could throw this into one "kitchen sink" query that the SQL optimizer would do its thing on, because these are basically a bunch of toggle-able AND conditions on the WHERE clause, and the optimizer has insight into which predicates are actually doing something:

```
WHERE
   ...
   AND ($data_plane_name IS NULL OR $data_plane_name == dp.data_plane_name)
   AND ($catalog_type IS NULL OR $catalog_type == ls.catalog_type)
   AND ($names IS NULL OR ls.catalog_name IN UNNEST($names))
```
 
 If we **can't** do this, I agree that it blows up the approach and we need 1:1 correspondence between `@oneOf` variants and specific query.

jgraettinger (approved):
LGTM :tada:

crates/control-plane-api/src/server/public/graphql/live_spec_refs.rs
psFried: Just to close the loop here, this ended up working out just fine, and is implemented in the last commit.

crates/control-plane-api/src/server/public/graphql/alerts.rs
```diff
@@ -255,6 +248,118 @@ async fn fetch_alert_history_by_prefix(
...
+    resolved_at: Option<DateTime<Utc>>,
+    arguments: crate::TextJson<async_graphql::Value>,
+}
+
+async fn fetch_alerts_by_prefix_before(
+    AlertsBy { prefix, active }: AlertsBy,
+    before: PaginatedAlertsCursor,
+    last: usize,
+    db: &sqlx::PgPool,
+) -> sqlx::Result<(Vec<AlertRow>, bool)> {
```
GregorShear, line 264: I'm pretty sure this after/before pagination convention is implemented backwards here, where the descending sort order of timestamped alerts is conflated with the normal ascending order of paged results. We should fix this.

PR #2439, opened 2025-10-06
Docs: Add History Mode, Source Tag

aeluce:
**Description:**

Updates docs to include some missing connector config fields as well as some information about usage.

* Adds History Mode to CDC SQL captures
* Adds Source Tag to SQL captures
* Creates a landing page for the "Customizing Data Flows" section to collect connector options, especially ones that don't necessarily require their own separate page

Updates affected SQL capture connector docs and creates a new page for Customizing Data Flows.

**Notes for reviewers:**

Thanks for reviewing!


jwhartley (approved):
LGTM, although I would reword this for simplicity/clarity:
> Capture change events without reducing them to a final state.
to
> Capture each change event, without merging.

I like the idea to break out the new page, thanks

PR #2440, opened 2025-10-07
validation: use explicit data plane for ops/ prefix specs

jgraettinger:
The ops/ prefix is special as these specs exist in all data-planes. When validating existing ops/ specs, we now pass the current data_plane_id as explicit_plane to avoid errors about missing mapped data-planes.

For regular specs, we continue to pass None to allow walk_prefix to use the first storage mapping plane as a no-op placeholder, avoiding errors when data-planes are removed from mappings or intended for other specs.

**Workflow steps:**

**Notes for reviewers:**

### DIFF
--- a/crates/validation/src/lib.rs
+++ b/crates/validation/src/lib.rs
@@ -520,6 +520,7 @@ where
                 }
                 .push(Scope::new(draft.scope()), errors);
             }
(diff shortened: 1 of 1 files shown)


psFried (approved):
LGTM

PR #2445, opened 2025-10-07
runtime: compute flow-connector-init image from $FLOW_VERSION

jgraettinger:
Rather than hard-coding a specific pin, use the same version as the current build.

This works because an in-progres build prefers a resolved pre-built flow-connector-init binary, and doesn't attempt to pull this image.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -74,8 +74,8 @@ jobs:
 
       - run: make install-tools
       - run: go mod download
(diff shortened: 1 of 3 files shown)


psFried (approved):
LGTM

jgraettinger:
This isn't gonna work without a bigger refactor of our build system, because flow-connector-init is built in the parallel musl job, but is required for running tests here.

PR #2457, opened 2025-10-15
control-plane-api: default to aws-us-east-1 data plane for new tenants

psFried:
Changes the ordering of data planes when provisioning a new tenant so that the aws-us-east-1 data plane will be first in the list. This is a short-ish term solution until we can build out UI for users to pick which one they'd like to be the default.


skord (approved):
LGTM

jshearer (approved):
LGTM 👍

PR #2464, opened 2025-10-20
derive-python: implement Python derivations

jgraettinger:
**Description:**

Implements a new Python derivation connector for Flow, enabling users to write derivation transforms in Python using async functions with Pydantic-typed inputs and outputs.
                                                                                                                                                                                                                                                             
  This follows the same architectural pattern as `derive-typescript`:                                                                                                                                                                                         
  - **Rust connector binary** handles protocol messages and generates code                                                                                                                                                                                    
  - **Type generation** from JSON Schema → doc::Shape → Pydantic models                                                                                                                                                                                       
  - **Runtime wrapper** manages Python execution via `uv` and asyncio
  - **Docker image** using `ghcr.io/astral-sh/uv:python3.14-trixie-slim` base

  ## Key Changes

  ### Protocol & Model Integration
  - Add `PYTHON = 5` to `ConnectorType` enum in `flow.proto`
  - Add `DeriveUsingPython` to catalog models with `module` field                                                              
  - Runtime routes Python derivations to `ghcr.io/estuary/derive-python:dev`
  - This supports module source management and code / stub generation similar to derive-typescript, so users have an ergonomic IDE development experience out of the box.

  ### Code Generation (`crates/derive-python/src/codegen`)                                                                                                                                                                                                    
  - **Schema → AST mapper**: Converts JSON Schema via `doc::Shape` to intermediate AST                                                                                                                                                                        
  - **Type support**: Union types, nested objects, optional fields, literal enums                                                                                                                                                                             
  - **Naming conventions**: PascalCase classes from `$anchor`, snake_case methods                                              
  - **Pydantic models**: Protocol types (Open, StartCommit, etc.) with `extra='allow'`                                         
  - **Transform classes**: Generated `Read{Transform}` and `Source{Transform}` types                                           
  - **IDerivation ABC**: Abstract base class with `@abstractmethod`-decorated transforms
  - **Python identifier sanitization**: Handles keywords and unicode correctly

  ### Runtime Implementation
  - Template-based `main.py` with asyncio event loop
  - Designed to support intra-transaction task concurrency via asyncio

  ### Build & CI
  - Integrate `derive-python` into Cargo workspace
  - Add GitHub workflow for building and publishing Docker image
  - Dockerfile with `uv` for fast dependency management

  ### Testing
  - Lots of snapshot tests over code generation
  - Manual testing using `flowctl preview`
  - Manual testing on a local stack

**Workflow steps:**

* Users scaffold a derivation (typically using UI)
* `flowctl generate` or `flowctl catalog pull-specs` will produce `flow_generated` / `py_project.toml` / `my_module.py` stub scaffolding:

```python
  """Derivation implementation for patterns/sums."""
  from patterns.sums import IDerivation, Document, Request
  from typing import AsyncIterator

  class Derivation(IDerivation):
      async def from_ints(self, read: Request.ReadFromInts) -> AsyncIterator[Document]:
          # Transform and generate output documents as an incremental, streaming generator.
          yield Document(
              Key=read.doc.Key,
              Sum=read.doc.Int,
              Count=1
          )
```

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -3,6 +3,7 @@ target/
 .build/
 .build-ci/
 
+!target/x86_64-unknown-linux-musl/release/derive-python
 !target/x86_64-unknown-linux-musl/release/derive-typescript
 
 # Typical IDE stuff.
--- a/.github/workflows/derive-python.yaml
+++ b/.github/workflows/derive-python.yaml
@@ -1,0 +1,49 @@
+name: derive-python
+
+on:
+  push:
+    branches: [master]
+    paths: [crates/derive-python/**]
+  pull_request:
+    branches: [master]
+    paths: [crates/derive-python/**]
+
+jobs:
+  build-and-test:
+    runs-on: ubuntu-24.04
+    permissions:
+      contents: read
+      packages: write
+
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+          submodules: false
+
+      - uses: dtolnay/rust-toolchain@stable
+        with:
+          targets: x86_64-unknown-linux-musl
+
+      - run: sudo apt install -y musl-tools
+
+      - name: Login to GitHub container registry
+        if: ${{ github.ref == 'refs/heads/master' }}
+        run: |
+          echo "${{ secrets.GITHUB_TOKEN }}" | \
+            docker login --username ${{ github.actor }} --password-stdin ghcr.io
+
+      - uses: Swatinem/rust-cache@v2
+
+      - name: build `derive-python`
+        run: cargo build -p derive-python --target x86_64-unknown-linux-musl --release
+
+      - name: test `derive-python`
+        run: cargo test -p derive-python --target x86_64-unknown-linux-musl --release
+
+      - name: package docker image
+        run: docker build -t ghcr.io/estuary/derive-python:dev -f crates/derive-python/Dockerfile .
+
+      - name: push docker image
+        if: ${{ github.ref == 'refs/heads/master' }}
+        run: docker push ghcr.io/estuary/derive-python:dev
--- a/crates/derive-python/Dockerfile
+++ b/crates/derive-python/Dockerfile
@@ -1,0 +1,16 @@
+FROM ghcr.io/astral-sh/uv:python3.14-trixie-slim
+
+RUN apt update -y \
+     && apt install --no-install-recommends -y \
+     ca-certificates \
+     nodejs \
+     && rm -rf /var/lib/apt/lists/*
+
+COPY target/x86_64-unknown-linux-musl/release/derive-python /
+
+USER nobody
+ENV UV_CACHE_DIR=/tmp/uv-cache
+
+ENTRYPOINT ["/derive-python"]
+LABEL FLOW_RUNTIME_CODEC=json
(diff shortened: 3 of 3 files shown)


crates/derive-python/src/codegen/ast.rs
```diff
@@ -0,0 +1,251 @@
...
+        ctx.into.push('\n');
+    }
+}
+
+impl AST {
+    pub fn render(&self, w: &mut String) {
+        match self {
+            AST::Any => w.push_str("typing.Any"),
+            AST::Bool => w.push_str("bool"),
+            AST::Float => w.push_str("float"),
```
Alex-Bair, line 157: Any concerns with precision loss by using `float`? `decimal.Decimal` is an alternative that better keeps the precision of the original input, although at the cost of performance. A quick & dirty benchmark I ran showed parsing strings with `float` is ~2x faster than parsing with `decimal.Decimal`.

```python
from decimal import Decimal

small_num = "0.123456789012345678901234567890"

print(float(small_num))   # 0.12345678901234568
print(Decimal(small_num)) # 0.123456789012345678901234567890

big_num = "9007199254740993.1"

print(float(big_num))     # 9007199254740994.0
print(Decimal(big_num))   # 9007199254740993.1
```

crates/derive-python/src/lib.rs
```diff
@@ -115,17 +190,240 @@ fn validate(validate: derive::request::Validate) -> anyhow::Result<derive::respo
...
+        dependencies.insert("pyright".to_string(), ">=1.1".to_string());
+    }
+
+    result.push_str("[project]\n");
+    result.push_str(&format!(
+        "name = \"{}\"\n",
+        collection_name.replace('/', "-")
+    ));
+    result.push_str("version = \"0.1.0\"\n");
+    result.push_str("requires-python = \">=3.12\"\n");
```
Alex-Bair, line 407: The Dockerfile is using `uv:python3.14-trixie-slim`. Should the pyproject.toml require Python 3.14 or greater?

Alex-Bair (approved):
LGTM, just had a couple small questions as I was going through the commits.

crates/derive-python/src/codegen/ast.rs
jgraettinger: Good question. No, no concerns, and the reason is that a Python float is backed by a C double / Rust `f64`, which is also how the Flow runtime represents them (in an enum, along with i64 for negative numbers, and u64 for non-negative). So, they're using the same representation.

In the common case of a `type: number`, that maps into an AST::Union of AST::Float and AST::Int. For `type: integer`, Pydantic matches the JSON Schema behavior of allowing `5.0` as `int` (while rejecting `5.1`).

I should do a follow-up PR with updates to examples/ though to prove this...

crates/derive-python/src/lib.rs
jgraettinger: The choice of 3.12 here was for developer ergonomics, so that older versions of python on a `flowctl` user's machine would work out of the box. I went with this somewhat arbitrarily as it's /usr/bin/python3 on Ubuntu 24.04

crates/derive-python/src/codegen/ast.rs
Alex-Bair: That all makes sense, thanks!

PR #2466, opened 2025-10-21
control-plane-api: discovers add inferred schema placeholder

### LINKED ISSUE #2459: Inline the `inferred schema` to the discover response
The UI will need the inferred schema to be inline during discover.

psFried:
Resolves #2459
Discovers add the inferred schema placeholder when initializing new collection schemas. This ensures that the collection schemas don't have a dangling reference to it, so the UI can get projections from them.


jgraettinger (approved):
LGTM

PR #2475, opened 2025-10-24
Docs: Add new SingleStore materialization variant

aeluce:
**Description:**

Adds reference docs for the new SingleStore materialization, which is a variant of the MySQL materialization.

**Notes for reviewers:**

Thanks for reviewing!

### DIFF
--- a/site/docs/reference/Connectors/materialization-connectors/MySQL/singlestore-mysql.md
+++ b/site/docs/reference/Connectors/materialization-connectors/MySQL/singlestore-mysql.md
@@ -117,5 +117,5 @@ The default is to use standard updates.
 
 ## Reserved words
 
-SingleStore uses MySQL, which has a list of reserved words that must be quoted in order to be used as an identifier.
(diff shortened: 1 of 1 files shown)


site/docs/reference/Connectors/materialization-connectors/MySQL/singlestore-mysql.md
```diff
@@ -0,0 +1,121 @@
...
+```
+
+## Delta updates
+
+This connector supports both standard (merge) and [delta updates](/concepts/materialization/#delta-updates).
+The default is to use standard updates.
+
+## Reserved words
+
+SingleStore uses MySQL, which has a list of reserved words that must be quoted in order to be used as an identifier.
```
mdibaiee: Actually let's link this one:

https://docs.singlestore.com/db/v9.0/reference/sql-reference/restricted-keywords/list-of-restricted-keywords/

I will change the code as well to add the ones here to the connector

PR #2476, opened 2025-10-24
Docs: Add new alert types

aeluce:
**Description:**

Updates the Notifications docs with basic info about new alert types and where to find alerts.

This PR will not be merged until alert history has been added in the UI.

Updates [Notifications](https://docs.estuary.dev/reference/notifications/)

**Notes for reviewers:**

Thanks for reviewing!

### DIFF
--- a/site/docs/guides/dashboard/notifications.md
+++ b/site/docs/guides/dashboard/notifications.md
@@ -15,7 +15,7 @@ Active and historical notifications include the type of alert, when it was fired
 
 ## Data Movement Alerts
 
-A user can select an interval for tracking zero data movement for a specific capture or materialization. From the connector details page, select the `Alerts` tab. Under the `Notification Settings` card, select a time interval from the dropdown labeled `Interval`. There is no need to save, but you must also have already configured notifications in order for the alert to take effect. If you are not yet subscribed to notifications, an info box will appear prompting you to set up a subscription by clicking on `CLICK HERE`.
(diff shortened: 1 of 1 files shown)


site/docs/guides/dashboard/notifications.md
```diff
@@ -5,16 +5,38 @@ slug: /reference/notifications/
...
+Active and historical notifications include the type of alert, when it was fired, and alert details.
+
 ## Data Movement Alerts
 
-When navigating to the main view of a capture or a materialization, a user can select an interval for tracking zero data movement. Under the `Notification Settings` card, select a time interval from the dropdown labeled `Interval`. There is no need to save, but you must also have already configured notifications in order for the alert to take effect. If you are not yet subscribed to notifications, an info box will appear prompting you to set up a subscription by clicking on `CLICK HERE`.
+A user can select an interval for tracking zero data movement for a specific capture or materialization. From the connector details page, select the `Alerts` tab. Under the `Notification Settings` card, select a time interval from the dropdown labeled `Interval`. There is no need to save, but you must also have already configured notifications in order for the alert to take effect. If you are not yet subscribed to notifications, an info box will appear prompting you to set up a subscription by clicking on `CLICK HERE`.
```
travjenkins: This might just be back to the `task vs connector` issue again - but this wording is a bit confusing to me:

`From the connector details page, select the `Alerts` tab`

I assume `connector` = `task` here?

aeluce: I was using "connector" as shorthand for "capture or materialization." I can change it.

Would "task" be too broad a term here? I didn't see the data movement `Notification Settings` card on derivations.

psFried: We recently discussed trying to avoid the term "shard" in favor of "task", and this feels similar to me. My 2 cents is that "task" is probably also preferable to "connector", which I see as somewhat more ambiguous than "shard". We may not currently show alert settings for derivations, but from a backend perspective we support data movement and shard failure alerts on them. And I'd guess that most users aren't even aware that derivations exist, and so won't be confused by the ambiguity.

PR #2481, opened 2025-10-29
graphql: support `startedAt` parameter for non-blocking authZ checks

### LINKED ISSUE #2480: graphql: support `startedAt` parameter to control authZ blocking
This pertains to the behavior of our APIs when a request tries to access a resource that the user is not authorized to. We perform authorization checks using a `Snapshot`, which gets periodically updated, and can thus be out of date. When a request tries to access an unauthorizede catalog namespace, we cannot know whether the authorization check would pass after the next snapshot refresh (in other words, if role grants were just recently added, but not yet observed). Our REST endpoints generally accept a `started` parameter for the timestamp at which the client began the request, and we can use that timestamp to determine whether the snapshot is up-to-date enough that we can definitively reject it.

We'd like to support this in our GraphQL API as well. All top-level resolvers should accept an optional `startedAt` parameter. If it's not provided, then any failed authZ check will wait until the next snapshot refresh before checking again, and will return a terminal authZ error if it still fails. If it's set to a timestamp that's _after_ the time of the last snapshot refresh, and the authZ check fails, then we should return a response with a `retryAt` timestamp, to have the client retry the request after the snapshot should have been refreshed. If it's set to a timestamp _before_ the snapshot was last refreshed, then we can treat any authZ failure as terminal.

psFried:
Resolves #2480

Adds an optional `startedAt` parameter to top-level graphql queries, which controls the blocking behavior when a request tries to access resources that the user lacks capability to. Because the `Snapshot` is only updated periodically, we need to wait until the next snapshot refresh in order to definitively say that a user lacks a given capability. If `startedAt` is not provided, then we'll block until the next snapshot refresh in order to be certain about the authorization decision.  But if `startedAt` is provided, and the snapshot was last refreshed _before_ `startedAt`, then we'll return an error response that includes a `retryAfter` extension, which indicates to the client that it may retry the request in order to get a more definitive answer. If `startedAt` is provided and the value is _after_ the time of the last snapshot refresh, then the authorization decision is definitive and no `retryAfter` will be returned.

### DIFF
--- a/crates/control-plane-api/src/server/public/graphql/mod.rs
+++ b/crates/control-plane-api/src/server/public/graphql/mod.rs
@@ -133,7 +133,7 @@ pub async fn graphql_graphiql() -> impl axum::response::IntoResponse {
               import 'graphiql/setup-workers/esm.sh';
 
               const fetcher = createGraphiQLFetcher({
-                url: 'http://localhost:8675/api/graphql',
+                url: '/api/graphql',
               });
               const plugins = [HISTORY_PLUGIN, explorerPlugin()];
 


crates/agent/src/integration_tests/harness.rs
```diff
@@ -1371,12 +1373,41 @@ impl TestHarness {
...
 
-        // Execute the query
-        let response = schema.execute(request).await;
+        // Execute the query using a pretty short timeout. This is necessary because the
+        // server will try to wait for a Snapshot refresh when an authZ check fails, but
+        // snapshots are never automatically refreshed during integration tests.
+        let response =
+            tokio::time::timeout(std::time::Duration::from_secs(1), schema.execute(request))
+                .await
+                .context("graphql query timed out (note: authorization failures could be a common cause of this if the query omitted a startedAt parameter)")?;
```
jgraettinger, line 1382: :+1:

crates/control-plane-api/src/server/error.rs
```diff
@@ -129,20 +142,38 @@ impl ApiError {
     }
 }
 
+impl async_graphql::ErrorExtensions for ApiError {
+    fn extend(&self) -> async_graphql::Error {
+        let err = async_graphql::Error::new(self.error.to_string());
```
jgraettinger: use format!("{:#}") so we include full error chains.

to_string() drops all but the highest-level error context.

Any other places we're doing this?

crates/control-plane-api/src/server/mod.rs
```diff
@@ -88,38 +88,74 @@ impl App {
...
     pub async fn verify_user_authorization(
         &self,
         claims: &ControlClaims,
-        prefixes: Vec<String>,
+        req_started_at: Option<chrono::DateTime<chrono::Utc>>,
+        prefixes_or_names: Vec<String>,
         capability: models::Capability,
     ) -> Result<Vec<String>, ApiError> {
-        let started = chrono::Utc::now();
+        let now = chrono::Utc::now();
```
jgraettinger: nit: `let started_at = req_started_at.unwrap_or(chrono::Utc::now())`

`now` is a little confusing because it's no longer "now" in later iterations.

crates/control-plane-api/src/server/mod.rs
```diff
@@ -88,38 +88,74 @@ impl App {
...
 
+    /// Verifies tha the user has at least `capability` to each of the
+    /// `prefixes_or_names`. This ensures that the error extensions will include
+    /// a `retryAt` timestamp when appropriate, so that clients can retry the
+    /// request after the snapshot is refreshed.
+    pub async fn verify_user_authorization_graphql(
+        &self,
+        claims: &ControlClaims,
+        req_started_at: Option<chrono::DateTime<chrono::Utc>>,
+        prefixes_or_names: Vec<String>,
```
jgraettinger, line 150: optional / for later: this could take an IntoIterator + Copy over AsRef, to avoid having to materialize to a vector.

crates/control-plane-api/src/server/error.rs
psFried: > Any other places we're doing this?

No. I only used it here because the one and only usage is with authZ errors, which don't wrap any other errors. I went ahead and just changed it to use `#` though, just in case any other usages come up in the future.

crates/control-plane-api/src/server/mod.rs
psFried: I just inlined both these usages 👍

crates/control-plane-api/src/server/mod.rs
psFried: Yeah, I'd like to come up with a better factoring for these, but am holding off until I have a better idea for the return value (related to your todo on `verify_user_authorization`)

PR #2489, opened 2025-11-04
Phil/inferred cooloff

### LINKED ISSUE #2488: agent: allow limiting the frequency of inferred schema updates

psFried:
**Description:**

Resolves #2488 

Introduces the `INFERRED_SCHEMA_UPDATE_COOLDOWN` env variable (or cli argument), which limits the frequency of inferred schema updates on a per-collection basis. This can be set to `5m`, for example, to ensure that controllers will only try to publish inferred schema updates every 5 minutes. If an inferred schema update is observed prior to that, the controller will wait until the cooldown has elapsed before publishing it, and will set the `nextMd5` and `nextUpdateAfter` fields in the inferred schema status.

Note that any publication of the collection will always update the inferred schema. So there's still cases where inferred schema updates could happen more frequently, for example on derivations that get published in response to changes in their source collections. We might also consider limiting the frequency of publications in response to dependency changes, but I'm leaving that for a separate PR, if ever.

Also fixes a bug in `connector_tags.rs`, which prevented those jobs from succeeding.

### DIFF
--- a/crates/agent/src/controllers/collection.rs
+++ b/crates/agent/src/controllers/collection.rs
@@ -6,11 +6,9 @@ use super::{
 };
 use crate::controllers::{activation, publication_status};
 use anyhow::Context;
+use control_plane_api::publications::PublicationResult;
 use itertools::Itertools;
-use models::status::{
-    collection::{CollectionStatus, InferredSchemaStatus},
-    publications::PublicationStatus,
-};
+use models::status::collection::CollectionStatus;
 
 pub async fn update<C: ControlPlane>(
     status: &mut CollectionStatus,
@@ -26,7 +24,6 @@ pub async fn update<C: ControlPlane>(
         return Ok(Some(NextRun::immediately()));
     }
 
-    // TODO(phil): remove `inferred_schema` status
     let CollectionStatus {
         inferred_schema: _,
         publications,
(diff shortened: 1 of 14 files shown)


crates/agent/src/controllers/collection.rs
```diff
@@ -144,32 +172,47 @@ fn handle_deleted_dependencies(
...
+            {
+                let next_update_after = last_updated + cooldown;
+                tracing::info!(%next_update_after, ?cooldown, next_md5 = %md5, "awaiting cooldown before inferred schema update");
+                inferred_schema_status.next_md5 = Some(md5);
+                inferred_schema_status.next_update_after = Some(next_update_after);
+                return Ok(false);
+            }
 
-        if status.schema_md5.as_ref() != Some(&md5) {
+            let mut pending_pub = PendingPublication::new();
```
jgraettinger, line 209: I think I was expecting an implementation where:

* The `if status.schema_md5.as_ref() != Some(&md5)` test additionally considers whether `last_update_at` is sufficiently far in the past
* If it's _not_ far enough in the past, then it does nothing but also lower-bounds the next controller run time to be at `last_update_at + cooldown`

Would you unpack the motivation you're seeing for `next_md5` ? Is this ☝️ not possible, or is there another benefit? Squinting, it seems like an extra two-stage control flow for inference updates that's only rarely taken, which would make me more wary of lurking bugs.

psFried: That's how it works here already. Line 194 above checks `inferred_schema_status.schema_md5.as_ref() != Some(&md5)` and then we immediately check:

```
if let Some(last_updated) = inferred_schema_status.schema_last_updated
                && (last_updated + cooldown) > control_plane.current_time()
```

and if that is true, then we need to wait for the cooldown. We set `next_md5` and `next_update_after` to record the fact that we're waiting on the cooldown, and the controller `next_run` time will be set based on `next_update_after`.

If that check is false, then we just proceed with the inferred schema publication as normal, and don't bother setting `next_md5` or `next_update_after`.

jgraettinger: Ooohhkay. So to confirm, `next_md5` and `next_update_after` are purely advisory for API consumers of the controller status? That makes sense.

I'm under the weather here and operating at 50% brain power.

jgraettinger (approved):
LGTM

crates/agent/src/controllers/collection.rs
psFried: No worries. Technically, we do use `next_update_after` in the controller, just as part of determining the `next_run`. But their primary purpose is really just to indicate to consumers that we're holding off on a pending inferred schema update.

Issue #2376, opened 2025-09-02
dekaf: Support collection reset and binding backfill via "leader epochs"


jshearer:
1.  Consumer State: A modern Kafka consumer doesn't just track its current offset; it also receives and tracks the `LeaderEpoch` of the partition it's fetching from.
2.  Fetch with Epoch: When the consumer sends a `FetchRequest` to a broker, it includes the `CurrentLeaderEpoch` it knows for that partition.
3.  Detecting Truncation: Let's say your journal resets.
    -   Your emulator increments its internal "generation," which you map to the `LeaderEpoch`. Let's say it goes from epoch 5 to 6.
    -   The consumer connects, still thinking it's on epoch 5. It sends a `FetchRequest` for `(partition=0, offset=123, current_epoch=5)`.
    -   Your emulator sees the incoming epoch (5) is less than its current epoch (6). It knows the consumer's state is stale.
    -   Your emulator responds with the error `FENCED_LEADER_EPOCH`.
4.  Consumer Recovery:
    -   When `librdkafka` receives this error, it knows the consumer's position is likely invalid due to truncation.
    -   It then triggers its offset validation logic, which sends an `OffsetsForLeaderEpochRequest`. This request basically asks the broker: "I was last at epoch 5. What is the end offset for that epoch?"
    -   Your emulator would respond with the last valid offset from the previous journal generation.
    -   More commonly, if the consumer was far behind, your emulator might just say, "I have no record of that epoch," or return offset 0 for the new epoch.
    -   The consumer now knows its position is invalid. It will either reset to the offset provided by the broker or fall back to its `auto.offset.reset` policy if the broker couldn't provide a valid offset for its old epoch.

PR #2491, opened 2025-11-05
dekaf: Report protocol message versions in metrics

jshearer:
**Description:**

One of the options for supporting [collection reset](https://github.com/estuary/flow/issues/2376) will require bumping the minimum supported message version for a few of the messages Dekaf supports. This will expose visibility into the versions that are currently being used.


psFried (approved):
LGTM

PR #2492, opened 2025-11-05
agent: clear inbox when executing an already completed directive

psFried:
Due to the way directives are handled, it's possible for an applied directive to already be completed when the executor runs. The agent expects this and returns `Action::Done` for such tasks, but we neglected to clear the task inbox first. This caused the automations server to re-enqueue the task instead of deleting it.  This clears the inbox, allowing the task to be removed.


jshearer (approved):
LGTM

PR #2494, opened 2025-11-06
automations: dequeue jobs in ascending order

psFried:
Attempts to mitigate a issues with controller jobs, which cause us to re-publish captures and materializations too frequently, in response to aa bunch of inferred schema updates.


jgraettinger (approved):
LGTM

PR #2499, opened 2025-11-13
docs/source-postgres: Clarify version requirements for standby captures

willdonnelly:
**Description:**

Adds a note to the docs clarifying that only PostgreSQL v16+ supports captures from standby replicas, along with the error message that earlier versions would give.

Also I cleaned up the wording of the following paragraph because it jumped out at me as especially awkward.


Alex-Bair (approved):
LGTM

PR #2502, opened 2025-11-18
docs/source-mysql-batch: Actual documentation!

willdonnelly:
**Description:**

Following on from https://github.com/estuary/flow/pull/2499, this PR adds the bare minimum of fleshed-out user documentation for source-mysql-batch.

### DIFF
--- a/site/docs/reference/Connectors/capture-connectors/MySQL/mysql-batch.md
+++ b/site/docs/reference/Connectors/capture-connectors/MySQL/mysql-batch.md
@@ -1,28 +1,169 @@
 # MySQL Batch Query Connector
 
-This connector captures data from MySQL into Flow collections by periodically
-executing queries and translating the results into JSON documents.
+This connector captures data from MySQL databases into Flow collections by
+periodically executing queries and translating the results into JSON documents.
+
+## When to use this connector
 
 We recommend using our [MySQL CDC Connector](http://go.estuary.dev/source-mysql)
-instead where possible. Using CDC provides lower latency data capture, delete and
-update events, and usually has a smaller impact on the source database.
+instead when possible. CDC provides lower latency data capture, delete and
+update events, and typically has a smaller impact on the source database.
 
-However there are some circumstances where this might not be feasible. Perhaps you need
-to capture from a managed MySQL instance which doesn't support logical replication. Or
-perhaps you need to capture the contents of a view or the result of an ad-hoc query.
-That's the sort of situation this connector is intended for.
+However, the batch connector is the right choice when:
 
-The number one caveat you need to be aware of when using this connector is that **it will
-periodically execute its update query over and over**. The default polling interval is set
-to 24 hours to minimize the impact of this behavior, but even then it could mean a lot of
-duplicated data being processed depending on the size of your tables.
+- Your MySQL instance doesn't support binary log replication (e.g., some managed services)
+- You need to capture from database views
+- You want to execute ad-hoc or custom queries
+- You need to capture from a read replica that doesn't have binary logging enabled
 
-If the dataset has a natural cursor which could be used to identify only new or updated
-rows, this should be specified by editing the `Cursor` property of the binding. Common
-examples of suitable cursors include:
+## Supported versions and platforms
 
-  - Update timestamps, which are usually the best choice if available since they
-    can be used to identify all changed rows, not just updates.
-  - Creation timestamps, which can be used to identify newly added rows in append-only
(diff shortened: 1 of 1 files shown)


site/docs/reference/Connectors/capture-connectors/MySQL/mysql-batch.md
```diff
@@ -1,28 +1,169 @@
...
+
+## Supported versions and platforms
+
+This connector works with all supported MySQL versions on major cloud
+platforms (including Amazon RDS and Aurora, Google Cloud SQL, Azure Database
+for MySQL, and other managed services), as well as self-hosted instances.
+
+:::tip Configuration Tip
+To capture data from databases hosted on your internal network, you must
+use [SSH tunneling](/guides/connect-network/).
```
mdibaiee, line 27: I would suggest also mentioning that they can use Private Link (or perhaps more generically "private cloud networking features") if they have a private deployment

mdibaiee (approved):
LGTM % one small point. I think also a good idea to get reviews from @aeluce

site/docs/reference/Connectors/capture-connectors/MySQL/mysql-batch.md
```diff
@@ -1,28 +1,169 @@
...
+| `JSON` | Native JSON | Passed through without modification |
+
+## Query templates
+
+Query templates use Go's template syntax to generate SQL queries. The connector
+uses a default template which implements appropriate behavior for full-refresh
+and cursor-incremental bindings, but you can override this for custom behavior.
+
+Overriding the query template is best left to power-users or done at the direction
+of Estuary support. Consult the [connector source](https://github.com/estuary/connectors/blob/main/source-mysql-batch/main.go)
```
aeluce, line 168: I'd like to include query template examples for the batch connectors at some point (especially since that field in the dashboard says to consult docs). But that can be a later addition.

site/docs/reference/Connectors/capture-connectors/MySQL/mysql-batch.md
willdonnelly: Yeah, that seems like a generally useful addition to that info box. TBH, that bit was just copy-pasted from the SQL CDC connector docs, so I should probably add it there as well.

site/docs/reference/Connectors/capture-connectors/MySQL/mysql-batch.md
willdonnelly: Yeah, I started including some of that but it ended up being fairly verbose and amounted to "here is the default template and the description of the available template variables, have fun" unless I also included a bunch of worked examples, and I wasn't sure whether custom queries were really common enough in practice to be worth padding out the documentation with that many examples.

site/docs/reference/Connectors/capture-connectors/MySQL/mysql-batch.md
willdonnelly: @aeluce how would this sound for an updated "hey you'll need to do something if your DB isn't publicly reachable" info box?

> To capture data from databases hosted on your internal network, you may need to use [SSH tunneling](https://docs.estuary.dev/guides/connect-network/). Users with [private deployments](https://docs.estuary.dev/getting-started/deployment-options/#private-deployment) can also use private cloud networking features.

willdonnelly: Hmm no, on rereading that has a weird tone shift from direct second-person address to a more passive "users can ...". Perhaps:

> To capture data from databases hosted on your internal network, you may need to use [SSH tunneling](https://docs.estuary.dev/guides/connect-network/). If you have a [private deployment](https://docs.estuary.dev/getting-started/deployment-options/#private-deployment), you can also use private cloud networking features to reach your database.

site/docs/reference/Connectors/capture-connectors/MySQL/mysql-batch.md
aeluce: I've definitely seen folks asking how to use this. We could spin off query examples into their own page and link to that from the batch query connectors to avoid bogging down the reference pages.

site/docs/reference/Connectors/capture-connectors/MySQL/mysql-batch.md
willdonnelly: I'm going to go ahead and merge the current docs PRs as-is and then follow up with a single PR tweaking this infobox for various connectors.

PR #2503, opened 2025-11-18
docs/source-sqlserver-batch: Documentation!

willdonnelly:
**Description:**

Following on from https://github.com/estuary/flow/pull/2499, this PR adds the bare minimum of fleshed-out user documentation for source-sqlserver-batch.

### DIFF
--- a/site/docs/reference/Connectors/capture-connectors/SQLServer/sqlserver-batch.md
+++ b/site/docs/reference/Connectors/capture-connectors/SQLServer/sqlserver-batch.md
@@ -1,81 +1,180 @@
 # SQL Server Batch Query Connector
 
-This connector captures data from SQL Server into Flow collections by periodically
-executing queries and translating the results into JSON documents.
+This connector captures data from SQL Server databases into Flow collections by
+periodically executing queries and translating the results into JSON documents.
 
-For local development or open-source workflows, [`ghcr.io/estuary/source-sqlserver-batch:dev`](https://ghcr.io/estuary/source-sqlserver-batch:dev) provides the latest version of the connector as a Docker image. You can also follow the link in your browser to see past image versions.
+## When to use this connector
 
-We recommend using our [SQL Server CDC Connector](./sqlserver.md) instead
-if possible. Using CDC provides lower latency data capture, delete and update events, and usually
-has a smaller impact on the source database.
+We recommend using our [SQL Server CDC Connector](http://go.estuary.dev/source-sqlserver)
+instead when possible. CDC provides lower latency data capture, delete and
+update events, and typically has a smaller impact on the source database.
 
-However there are some circumstances where this might not be feasible. Perhaps you need
-to capture from a managed SQL Server instance which doesn't support logical replication.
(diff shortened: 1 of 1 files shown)


site/docs/reference/Connectors/capture-connectors/SQLServer/sqlserver-batch.md
```diff
@@ -1,81 +1,180 @@
...
+## Supported versions and platforms
 
-## Configuration
+This connector works with all supported versions of SQL Server on major cloud platforms
+(including Amazon RDS, Azure SQL Database, and other managed services), as well as
+self-hosted instances.
+
+:::tip Configuration Tip
+To capture data from databases hosted on your internal network, you must
+use [SSH tunneling](/guides/connect-network/).
```
mdibaiee, line 27: Same suggestion as mysql batch, to mention private cloud networking for private deployments

site/docs/reference/Connectors/capture-connectors/SQLServer/sqlserver-batch.md
```diff
@@ -1,81 +1,180 @@
 # SQL Server Batch Query Connector
 
-This connector captures data from SQL Server into Flow collections by periodically
-executing queries and translating the results into JSON documents.
+This connector captures data from SQL Server databases into Flow collections by
+periodically executing queries and translating the results into JSON documents.
 
-For local development or open-source workflows, [`ghcr.io/estuary/source-sqlserver-batch:dev`](https://ghcr.io/estuary/source-sqlserver-batch:dev) provides the latest version of the connector as a Docker image. You can also follow the link in your browser to see past image versions.
```
aeluce, line 6: Is there a reason to remove the GitHub container link? We include that on a lot of our connector reference pages.

willdonnelly: That paragraph just doesn't feel actually relevant to any real concern the reader might have. Possibly I'm unfamiliar with how some of our users are using this, but concretely:
1. The `:dev` version tag isn't really relevant to most users, they'll just use the tagged major version in the UI.
2. Users accessing this in the UI don't need to care about the Docker image URL in the least.
3. Users who want to set a capture up via flowctl will have enough other trouble trying to write that YAML from scratch that they'll generally want to start from a working example anyway, and they'll get the image URL there.
4. We don't provide any sort of useful changelog at that URL, so the only thing a user could learn is "the image was pushed recently" unless they're also watching our repository for a specific commit to be merged, and in that case they usually don't need to care about the Docker image releases.

I think that (3) is probably the closest we come to a real use-case for including the image URL, and I would argue that if we think we need to make that easier the best way to do it would be a complete example of a YAML task spec (which is something that other connector docs have, IIRC) which happens to include our current recommended image URL with appropriate major version.

aeluce: Fair enough. If we add docs later to specifically support local/open-source use cases, we can add something about finding the image there. As-is, we already mention elsewhere that self-hosting support is spotty.

PR #2504, opened 2025-11-18
docs/source-oracle-batch: Documentation

willdonnelly:
**Description:**

Following on from https://github.com/estuary/flow/pull/2499, this PR adds the bare minimum of fleshed-out user documentation for source-oracle-batch. I think it's correct? It's hard to be sure with Oracle.

### DIFF
--- a/site/docs/reference/Connectors/capture-connectors/OracleDB/oracle-batch.md
+++ b/site/docs/reference/Connectors/capture-connectors/OracleDB/oracle-batch.md
@@ -1,29 +1,174 @@
 # Oracle Batch Query Connector
 
-This connector captures data from Oracle into Flow collections by periodically
-executing queries and translating the results into JSON documents.
+This connector captures data from Oracle databases into Flow collections by
+periodically executing queries and translating the results into JSON documents.
 
-For local development or open-source workflows, [`ghcr.io/estuary/source-oracle-batch:dev`](https://ghcr.io/estuary/source-oracle-batch:dev) provides the latest version of the connector as a Docker image. You can also follow the link in your browser to see past image versions.
+## When to use this connector
 
-We recommend using our [Oracle CDC Connector](http://go.estuary.dev/source-oracle) instead
-if possible. Using CDC provides lower latency data capture, delete and update events, and usually
-has a smaller impact on the source database.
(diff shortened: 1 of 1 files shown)


site/docs/reference/Connectors/capture-connectors/OracleDB/oracle-batch.md
```diff
@@ -1,29 +1,174 @@
...
-ad-hoc queries, you need to either have some way of restricting the query to "just the new
-rows since last time" or else have your polling interval set high enough that the data rate
-`<DatasetSize> / <PollingInterval>` is an amount of data you're willing to deal with.
+This connector works with Oracle Database 11g and later on major cloud platforms
+(including Amazon RDS, Oracle Cloud Infrastructure, and other managed services),
+as well as self-hosted instances.
+
+:::tip Configuration Tip
+To capture data from databases hosted on your internal network, you must
+use [SSH tunneling](/guides/connect-network/).
```
mdibaiee, line 27: Same here as mysql and sqlserver

site/docs/reference/Connectors/capture-connectors/OracleDB/oracle-batch.md
```diff
@@ -1,29 +1,174 @@
 # Oracle Batch Query Connector
 
-This connector captures data from Oracle into Flow collections by periodically
-executing queries and translating the results into JSON documents.
+This connector captures data from Oracle databases into Flow collections by
+periodically executing queries and translating the results into JSON documents.
 
-For local development or open-source workflows, [`ghcr.io/estuary/source-oracle-batch:dev`](https://ghcr.io/estuary/source-oracle-batch:dev) provides the latest version of the connector as a Docker image. You can also follow the link in your browser to see past image versions.
```
aeluce, line 6: Same here. If the phrasing seems off, we can revise it. Or if it's not useful to provide to users, I could remove it from other connector pages as well.

willdonnelly: Yeah, I addressed my perspective on this paragraph in the other PR comment, but the TL;DR is that I don't think it's useful and would suggest probably removing it across the board. This paragraph feels generally not-actionable, and worse it feels like it could potentially distract a user away from the happy path of just point-and-click configuring a capture in our UI.

PR #2511, opened 2025-11-19
Add reset functionality for captures and materializations

jgraettinger:
Extends the existing collection reset feature to also support resetting captures and materializations. When `reset: true` is published for a task, it regenerates the shard ID prefix, effectively resetting all task state while applying the change as a single publication (rather than requiring separate delete and create operations).

For derived collections, the reset flag now also applies to the derivation task state, effectively re-building the derivation in its entirety rather than just resetting the collection journals.

Implementation details:
- Added `reset` field to CaptureDef and MaterializationDef models
- Updated derivation validation to reset task state when collection is reset
- Updated validation logic to regenerate shard IDs when reset is true
- Added check preventing reset of Dekaf materializations

### DIFF
--- a/crates/agent/src/discovers.rs
+++ b/crates/agent/src/discovers.rs
@@ -300,6 +300,7 @@ async fn prepare_discover(
                 expect_pub_id: None,
                 bindings: Vec::new(),
                 delete: false,
+                reset: false,
             };
             draft.captures.insert(tables::DraftCapture {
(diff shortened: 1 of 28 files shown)


jgraettinger:
No, it's simply because reset doesn't do anything for Dekaf tasks (their generation ID is always zero), and better to explicitly error than let the user take an action they incorrectly believe is doing something it isn't.

PR #2513, opened 2025-11-19
docs/source-redshift-batch: Add documentation

willdonnelly:
**Description:**

Apparently we didn't actually have any docs at all for our current non-deprecated Redshift connector, and the docs URL for it was pointing at the docs for the old deprecated one. This adds some basic documentation based on source-postgres-batch, so that I can get the go-links pointing to the right places.


willdonnelly:
> I hope it's okay that I added a commit to your branch

Entirely fine with me.

PR #2520, opened 2025-11-22
docs: add mapping docs for materialize-elasticsearch

danielnelson:
**Description:**

Documentation for https://github.com/estuary/connectors/pull/3525, In this change I added a new method for manually specifying the field mapping, similar to the ddl option for the sql materializations.


Alex-Bair (approved):
LGTM

PR #2522, opened 2025-11-24
control-plane: change `catalog_stats.usage_seconds` column type

psFried:
Updates the column type to `bigint` instead of `integer`. This is needed because tenant-level rollups of monthly stats can actually roll over the maximum integer size. Fixes a production issue, hopefully.

### DIFF
--- a/supabase/migrations/20251124130455_stats-usage.sql
+++ b/supabase/migrations/20251124130455_stats-usage.sql
@@ -1,0 +1,136 @@
+-- Change the column type for usage_seconds in each partition of the catalog_stats table.
+-- We need to drop and re-create views in order to allow that.
+begin;
+
+drop view internal.alert_all;
+drop view internal.alert_data_movement_stalled;
+drop VIEW internal.new_free_trial_tenants;
+
+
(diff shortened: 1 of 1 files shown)


mdibaiee (approved):

supabase/migrations/20251124130455_stats-usage.sql
```diff
@@ -0,0 +1,135 @@
...
+ 'Unified view of all alert types. Alerts in this view can be either firing or
+ not. This view is not intended to be queried by end users, as it can be rather
+ slow.';
+
+ CREATE VIEW internal.new_free_trial_tenants AS WITH hours_by_day AS (
+   SELECT
+     t.tenant,
+     cs.ts,
+     (
+       cs.usage_seconds :: numeric / 3600.0
```
mdibaiee, line 69: I'm not an expert but is this cast okay? I see `::bigint` cast elsewhere for `bytes_processed`

psFried: Yeah, it should be. The range is:
> up to 131072 digits before the decimal point; up to 16383 digits after the decimal point

PR #2523, opened 2025-11-24
agent: generalize publication cooldown

### LINKED ISSUE #2505: agent: generalize cooldown to apply to all controller-initiated publications
In #2488 we introduced a cooldown timer that applied to inferred schema updates. We've now realized that this cooldown should apply more generally to any controller-initiated publications. We still see cases where connected captures and materializations get published very frequently when there's inferred schema updates for a bunch of bound collections. So this cooldown is intended to prevent that by applying to publications in response to changed dependencies.

psFried:
Resolves #2505

Replaces the `INFERRED_SCHEMA_UPDATE_COOLDOWN` with a more general `CONTROLLER_PUBLICATION_COOLDOWN`, which applies to all controller-initiated publications. In scenarios where a bunch of collections all have inferred schema updates in rapid succession, this prevents the connected captures and materializations from being repeatedly published in response to changes in the collections. In prod, we've observed that connected tasks are getting published very frequently during those times, and our goals here are to:

- try to reclaim some system capacity that could be used for publication for other tenants, tasks, etc
- prevent repeated publications from causing too many task restarts, and thus impacting task performance

This ended up requiring a bit of refactoring, particularly in the materialization controller.

The cooldown is implemented as a special error. The `RetryableError` type now has an `is_backoff` field, which is used to coalesce multiple different backoff errors into one. This ensures that we have consistent reporting of cooldown errors, even if a controller makes multiple attempts to publish. The status summary logic was also updated to ignore these cooldown errors. Generally, a cooldown error is not a condition that users will care about or be able to act on.

### DIFF
--- a/crates/agent/src/controllers/activation.rs
+++ b/crates/agent/src/controllers/activation.rs
@@ -60,6 +60,7 @@ const ALERT_AFTER_SHARD_FAILURES: std::sync::LazyLock<u32> = std::sync::LazyLock
 pub async fn update_activation<C: ControlPlane>(
     status: &mut ActivationStatus,
     alerts_status: &mut Alerts,
+    next_pending_pub: Option<chrono::DateTime<chrono::Utc>>,
     state: &ControllerState,
     events: &Inbox,
     control_plane: &C,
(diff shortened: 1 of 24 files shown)


crates/agent/src/controllers/activation.rs
```diff
@@ -282,6 +154,178 @@ pub async fn update_activation<C: ControlPlane>(
...
+    };
+    status.recent_failure_count = failures.len() as u32;
+
+    // If we've just observed a shard failure and we haven't already determined
+    // the time of the next retry, then determine that now.
+    if observed_shard_failures > 0 && status.next_retry.is_none() {
+        // If we're waiting on a cooldown before publishing the spec, then don't restart the task shards until
+        // after the spec is published. Not always, but often if the controller needs to publish
+        // the spec, then that publication is actually necessary in order to fix the failure condition.
+        // We add some extra time to this in order to
```
jgraettinger: nit: runs off

jgraettinger (approved):
LGTM

PR #2527, opened 2025-11-26
flowctl: preview streaming fixture

mdibaiee:
**Description:**

- Add a new streaming fixture format to `flowctl preview` which instead of loading all of the fixture into memory, it streams documents to the connector. This is useful for running huge fixtures for performance testing (I have tested this with a 17GB fixture file on a machine with 16GB of memory, and the memory usage is less than 1GB and constant)
- This is used in a new benchmarking tool I have prepared in connectors repository which I will create a pull-request for soon

**Workflow steps:**

**Notes for reviewers:**


mdibaiee:
> I haven't reviewed in depth, but could we reasonably support _only_ streaming for fixtures ? Why keep in-memory around?

@jgraettinger yep I would also would like to only keep streaming fixture, but I was not sure if there is anything other than connector tests that depend on it, so I kept both

crates/runtime/src/harness/streaming_fixture.rs
```diff
@@ -0,0 +1,276 @@
...
+        resume: consumer::Checkpoint,
+    ) -> BoxStream<'static, anyhow::Result<Read>> {
+        let skip = resume
+            .sources
+            .get("fixture")
+            .as_ref()
+            .map(|source| source.read_through as usize)
+            .unwrap_or_default();
+
+        let producer = crate::uuid::Producer([7, 19, 83, 3, 3, 17]);
```
psFried, line 103: nit: I'd appreciate a comment on this hard-coded producer. Why is that value important?

jgraettinger: I created that no doubt, as an arbitrary fixture (I recognize the numbers).

mdibaiee: yeah these already existed 👍🏽

jgraettinger:
> @jgraettinger yep I would also would like to only keep streaming fixture, but I was not sure if there is anything other than connector tests that depend on it, so I kept both

I can confirm that the only thing using it is connector tests.

crates/runtime/src/harness/mod.rs
```diff
@@ -5,6 +5,7 @@ mod capture;
 mod derive;
 pub mod fixture;
```
psFried: Do we still need the `fixture` module now that the streaming fixture is the only one that can be used?

crates/runtime/src/harness/streaming_fixture.rs
```diff
@@ -0,0 +1,276 @@
...
+        .boxed()
+    }
+}
+
+// Helper function to check if a line is an ack marker
+fn is_ack_line(line: &str) -> anyhow::Result<bool> {
+    // Try to parse as JSON object with "ack" field
+    if let Ok(val) = serde_json::from_str::<serde_json::Value>(line) {
+        if let Some(obj) = val.as_object() {
+            if let Some(ack) = obj.get("ack") {
```
psFried: This may or may not be worth changing, but I found this representation a bit confusing, due to the resemblance to the acknowledgement documents that we use in flow collections.  I think it could be clearer and more obviously different from our ack documents, like `{"commit": true}`. Noting also that the `ack` here has different semantics from the acknowledgements in collections. Here it means to commit the materialization transaction immediately, vs in collections where it just means that a certain set of documents from a collection may get added to the materialization transaction along with potentially many others.

mdibaiee: Good point 👍🏽

PR #2532, opened 2025-12-01
control-plane-api: cache CORS headers

### LINKED ISSUE #2528: Cache CORS headers for agent and reactors
The UI had to break up calls a bit when loading large tables. This means we end up duplicating a lot of CORS header calls that is pretty wasteful and impacts performance _a bit_.

I think we should set `Access-Control-Max-Age` to do _some_ caching. If we're worried about it hanging around too long I think setting to like 5 seconds will still help things a lot.

The endpoints this would be great on:
`agent-api` - `authorize/user/task`
`reactor` - `shards/list`

psFried:
Resolves #2528

Sets the Access-Control-Max-Age header on CORS pre-flight responses, to allow browsers to cache the results. This just cuts down on the number of pre-flight requests that the browser needs to make to our API.

### DIFF
--- a/crates/control-plane-api/src/server/mod.rs
+++ b/crates/control-plane-api/src/server/mod.rs
@@ -263,7 +263,13 @@ pub fn build_router(
     .map(|h| h.parse().unwrap())
     .collect::<Vec<_>>();
 
+    // Sets the Access-Control-Max-Age header to 1 hour, to allow browsers to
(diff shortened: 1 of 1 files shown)


jshearer (approved):
I don't really have a sense for the timeframes involved, but AFAIU we aren't expecting to change CORS headers all that frequently (or... at all?) so 1h seems reasonable

PR #2552, opened 2025-12-11
doc: check allocations after every document drained from spill

williamhbaker:
**Description:**

Previously `alloc` was size checked only if the entire document was successfully reduced. Schemas with a nested non-associative "last write wins" reduction strategy (such as history mode captures) do not get reduced because of this, but allocations may occur prior to the reduction traversing to that nested location from reducing a top-level merge, for example. This can lead to unbounded memory growth, and in extreme cases OOMs.

The fix here is to check the allocator size for every document drained.

A benchmark simulating history mode was added to verify that memory usage stays under the threshold as documents are drained. Prior to this change, memory usage would continue to grow indefinitely, potentially reaching multiple gigabytes.

**Workflow steps:**

**Notes for reviewers:**

### DIFF
--- a/crates/doc/src/combine/spill.rs
+++ b/crates/doc/src/combine/spill.rs
@@ -429,14 +429,14 @@ impl<F: io::Read + io::Seek> SpillDrainer<F> {
                 // Safety: we allocated `reduced` out of `self.alloc`.
                 let reduced = unsafe { OwnedHeapNode::new(reduced, self.alloc.clone()) };
 
-                // Safety: we must hold `alloc` constant for all reductions of a yielded entry.
-                if bump_mem_used(&self.alloc) > BUMP_THRESHOLD {
-                    self.alloc = Arc::new(Bump::new());
-                }
-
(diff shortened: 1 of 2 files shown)


crates/doc/src/combine/spill.rs
```diff
@@ -429,15 +429,15 @@ impl<F: io::Read + io::Seek> SpillDrainer<F> {
...
-                if bump_mem_used(&self.alloc) > BUMP_THRESHOLD {
-                    self.alloc = Arc::new(Bump::new());
-                }
-
                 OwnedNode::Heap(reduced)
             }
         };
 
+        // Safety: we must hold `alloc` constant for all reductions of a yielded entry.
+        if bump_mem_used(&self.alloc) > BUMP_THRESHOLD {
```
jgraettinger, line 437: I think it very likely that this has negligible performance impact due to super scaling / branch prediction, but I worry a bit about regressions from a change like this.

PR #2557, opened 2025-12-12
Dekaf E2E Testing WIP

jshearer:


crates/dekaf/tests/e2e/mod.rs
```diff
@@ -0,0 +1,612 @@
...
+            .iter()
+            .find(|l| l.name == "estuary.dev/expose-port")
+            .map(|l| &l.value)
+            .context("no port label")?;
+
+        let base = endpoint.replace("https://", "");
+        let url = format!("https://{hostname}-{port}.{base}/{path}");
+
+        let client = reqwest::Client::builder()
+            .danger_accept_invalid_certs(true)
```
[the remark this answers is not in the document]
jshearer: It's just for testing

.cargo/config.toml
```diff
@@ -1,3 +1,6 @@
+[build]
+rustflags = ["-D", "warnings"]
```
jshearer: I had a bunch of issues with tests failing after fairly long CI runs, so I added this to make `cargo check`/`cargo build` fail locally in the same way as CI will.

.github/workflows/platform-test.yaml
```diff
@@ -65,3 +65,44 @@ jobs:
       - run: mise run ci:gotest
       - run: mise run ci:catalog-test
       - run: mise run build:flow-schema
+  dekaf-test:
```
jshearer: Had to separate this out because `mise run ci:nextest-run` and/or `mise run ci:catalog-test` are leaving the control-plane in an unclean state, so I figured it would be simpler to fully isolate things since Dekaf's e2e tests also create/delete specs, publish to the data-plane, etc.

PR #2558, opened 2025-12-15
Fix Cloud Run deployments after Dockerfile relocation

skord:
## Summary

Fixes three broken Cloud Run deployment workflows that were never updated after the December 1st Dockerfile relocation (commit fec63eca6f).

## Problem

The mise infrastructure refactoring moved Dockerfiles from crate directories to `docker/`:
- `crates/agent/Dockerfile` → `docker/control-plane-agent.Dockerfile`
- `crates/data-plane-controller/Dockerfile` → `docker/data-plane-controller.Dockerfile`  
- `crates/oidc-discovery-server/Dockerfile` → `docker/oidc-discovery-server.Dockerfile`

The new Dockerfiles use `COPY ${TARGETARCH}/binary` to copy binaries from architecture-specific subdirectories, but the workflows were still copying binaries directly to crate directories without creating the `amd64/` subdirectory structure.

This caused all three deployments to fail when triggered.

## Solution

Updated all three workflows to:
1. Create `amd64/` subdirectory in the deployment source directory
2. Copy built binaries into `amd64/`
3. Copy required files (entrypoint scripts, sops) into `amd64/`
4. Copy the Dockerfile from `docker/` to the source directory
5. Proceed with Cloud Run deployment

## Testing

Locally verified that Docker builds succeed with this directory structure by simulating the workflow steps and running `docker build`.

## Fixed Workflows

- ✅ `deploy-agent-api.yaml`
- ✅ `deploy-data-plane-controller.yaml` (immediately broken by #2549)
- ✅ `deploy-oidc-discovery-server.yaml`

## Related

- Follows up on #2549 which exposed this issue
- Fixes deployments broken since commit fec63eca6f (Dec 1, 2025)

### DIFF
--- a/.github/workflows/deploy-data-plane-controller.yaml
+++ b/.github/workflows/deploy-data-plane-controller.yaml
@@ -19,6 +19,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
+        with:
+          fetch-depth: 0  # Required for git describe to find tags
 
       - name: Get image tag
         id: image-tag
--- a/.github/workflows/deploy-oidc-discovery-server.yaml
+++ b/.github/workflows/deploy-oidc-discovery-server.yaml
@@ -19,6 +19,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
+        with:
+          fetch-depth: 0  # Required for git describe to find tags
 
       - name: Get image tag
         id: image-tag


jgraettinger (approved):
LGTM, though this ought to be cleaned up further.

A better structure would be to let the Platform Build CI task complete and then deploy the pre-built/pushed images directly, rather than compiling and building an image from scratch.

.github/workflows/deploy-data-plane-controller.yaml
```diff
@@ -1,39 +1,35 @@
...
-        run: cargo build --release -p data-plane-controller
 
-      - run: mv target/release/data-plane-controller crates/data-plane-controller/
+      - name: Get image tag
+        id: image-tag
+        run: |
+          if [[ "${{ github.event_name }}" == "workflow_dispatch" ]]; then
+            TAG="dev-next"
+          else
+            TAG=$(git describe --tags --always)
```
jgraettinger, line 31: this may require `depth: 0` on the actions/checkout to pull tags? Otherwise i thought it was shallow

.github/workflows/deploy-agent-api.yaml
```diff
@@ -41,7 +33,7 @@ jobs:
           service: agent-api
           project_id: estuary-control
           region: us-central1
-          source: crates/agent/
+          image: ${{ steps.image-tag.outputs.image }}
```
jgraettinger, line 36: sanity check: does it pin the sha256 of the _current_ `dev-next`, or will it truly pull the "latest" dev-next every time Cloud Run decides to pull the image?

jgraettinger (approved):
LGTM % questions

.github/workflows/deploy-agent-api.yaml
skord: According to https://docs.cloud.google.com/run/docs/deploying#service

> Deploying to a service for the first time creates its first revision. Note that revisions are immutable. If you deploy from a container image tag, it will be resolved to a digest and the revision will always serve this particular digest.

.github/workflows/deploy-data-plane-controller.yaml
skord: Good catch!

PR #2559, opened 2025-12-15
Docs: Split SuiteQL connector into its own reference page

aeluce:
**Description:**

Improvements to NetSuite docs include:

* Update the SuiteAnalytics connector to remove the SuiteQL mode and bring other properties up to date
* Add a separate page for SuiteQL, documenting the new connector's specific configuration details
* Fully remove the old SuiteTalk connector documentation

Incorporates some info from a couple old PRs that never got merged.

Updates existing NetSuite docs and adds a new page for SuiteQL

**Notes for reviewers:**

Thanks for reviewing!

### DIFF
--- a/site/docs/reference/Connectors/capture-connectors/netsuite-suiteanalytics.md
+++ b/site/docs/reference/Connectors/capture-connectors/netsuite-suiteanalytics.md
@@ -4,7 +4,7 @@ import ReactPlayer from "react-player";
 
 This connector captures data from Oracle NetSuite into Estuary collections. It uses the SuiteAnalytics Connect feature in order to both load large amounts of data quickly, as well as introspect the available tables, their schemas, keys, and cursor fields.
 
-[`ghcr.io/estuary/source-netsuite:dev`](https://ghcr.io/estuary/source-netsuite-v2:dev) provides the
+[`ghcr.io/estuary/source-netsuite-v2:dev`](https://ghcr.io/estuary/source-netsuite-v2:dev) provides the
 latest connector image. You can also follow the link in your browser to see past image versions.
 
 In general, SuiteAnalytics Connect is the preferred method to retrieve data from NetSuite.
@@ -87,7 +87,7 @@ Create a NetSuite _integration_ to obtain a Consumer Key and Consumer Secret.
 
    6. (IMPORTANT) Click **Lists** and add all the dropdown entities with either **full** or **view** access level.
 
-   7. (IMPORTANT) Click **Setup** an add all the dropdown entities with either **full** or **view** access level.
+   7. (IMPORTANT) Click **Setup** and add all the dropdown entities with either **full** or **view** access level.
 
    8. (IMPORTANT) If you have multiple subsidiaries, make sure to select all of the subsidiaries you want the connector to have access to under the **Role** > **Subsidiary Restrictions** configuration.
 
@@ -145,7 +145,7 @@ See [connectors](../../../concepts/connectors.md#using-connectors) to learn more
 | Property                      | Title                  | Description                                                                                      | Type   | Required/Default |
 | ----------------------------- | ---------------------- | ------------------------------------------------------------------------------------------------ | ------ | ---------------- |
 | `/account`                     | Netsuite Account ID    | Netsuite realm/Account ID e.g. 2344535, as for `production` or 2344535_SB1, as for `sandbox` | string | Required         |
-| `/suiteanalytics_data_source` | Data Source            | Which NetSuite data source to use. Options are `NetSuite.com`, or `NetSuite2.com`                | string | Required         |
+| `/suiteanalytics_data_source` | Data Source            | Which NetSuite data source to use. This should generally be `NetSuite2.com`                | string | Required         |
 | `/authentication`             | Authentication Details | Credentials to access your NetSuite account                                                      | object | Required         |
 | `/authentication/auth_type`   | Authentication Type    | Type of authentication used, either `token` or `user_pass`.                                      | string | `token`          |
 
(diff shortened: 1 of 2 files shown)


site/docs/reference/Connectors/capture-connectors/netsuite-suiteanalytics.md
```diff
@@ -126,57 +145,77 @@ See [connectors](../../../concepts/connectors.md#using-connectors) to learn more
...
 | `/advanced/connection_limit` | Connection Limit | The maximum number of concurrent data streams to attempt at once.                                                                                                                                                   | int  | 10 Connections   |
 | `/advanced/task_limit`       | Task Limit       | The maximum number of concurrent tasks to run at once. A task is either a backfill or incremental load. Backfills can load multiple chunks in parallel, so this must be strictly &lt;= `/advanced/connection_limit` | int  | 5 Tasks          |
 | `/advanced/start_date`       | Start Date       | The date that we should attempt to start backfilling from. If not provided, backfill from the beginning.                                                                                                            | date | Not Required     |
+| `/advanced/cursor_fields`              | Cursor Fields     | Columns to use as cursor for incremental replication, in order of preference. Case insensitive. | string array | Not Required. `last_modified_date`, `lastmodifieddate`, and more are used as the default list. |
```
jshearer: `cursor_fields` is no more:

https://github.com/estuary/source-netsuite/pull/186/changes#diff-95e2e93e11615c31eea253f5d1e6c4272af209059945ee13b6ad76ae19d6b57dL100-L114

aeluce: Good catch, thanks! Noting, though, that `Cursor Fields` still shows up for me under the connector's advanced options when I start to create a new SuiteAnalytics capture. Is that intended?

jshearer: Screenshot 2026-01-08 at 12 34 18
Looks like I had forgotten to refresh the schema in `connector_tags`. Fixed now, thanks for pointing it out!

site/docs/reference/Connectors/capture-connectors/netsuite-suiteanalytics.md
```diff
@@ -190,16 +228,13 @@ captures:
                      consumer_secret_sops: xxx
                      token_id: xxx
                      token_secret_sops: xxx
-                  connection_type: suiteanalytics
-                  role_id: 3
                   suiteanalytics_data_source: NetSuite2.com
                   advanced:
                      connection_limit: 20
                      cursor_fields: []
```
jshearer: `cursor_fields` is no longer an option

site/docs/reference/Connectors/capture-connectors/netsuite-suiteanalytics.md
```diff
@@ -2,18 +2,19 @@ import ReactPlayer from "react-player";
 
 # NetSuite SuiteAnalytics Connect
 
-This connector captures data from Oracle NetSuite into Flow collections. It uses the SuiteAnalytics Connect feature in order to both load large amounts of data quickly, as well as introspect the available tables, their schemas, keys, and cursor fields.
+This connector captures data from Oracle NetSuite into Estuary collections. It uses the SuiteAnalytics Connect feature in order to both load large amounts of data quickly, as well as introspect the available tables, their schemas, keys, and cursor fields.
 
-[`ghcr.io/estuary/source-netsuite:dev`](https://ghcr.io/estuary/source-netsuite:dev) provides the
+[`ghcr.io/estuary/source-netsuite:dev`](https://ghcr.io/estuary/source-netsuite-v2:dev) provides the
```
jshearer: The link text should also have `-v2` here. FWIW... should we even be showing this? I kind of feel like nobody else but is could reasonably be expected to understand how to drive our connector images. What's the point of showing them like this at all?

site/docs/reference/Connectors/capture-connectors/netsuite-suiteql.md
```diff
@@ -0,0 +1,208 @@
...
+
+This connector captures data from Oracle NetSuite into Estuary collections using SuiteQL.
+
+To use SuiteAnalytics Connect to sync your data instead, see the [Netsuite SuiteAnalytics connector](./netsuite-suiteanalytics.md).
+
+## Supported data resources
+
+Currently, this connector supports a subset of NetSuite tables, such as:
+
+* Customer
```
jshearer, line 13: Also: 
* `Account`
* `DeletedRecord`

site/docs/reference/Connectors/capture-connectors/netsuite-suiteanalytics.md
```diff
@@ -126,57 +145,76 @@ See [connectors](../../../concepts/connectors.md#using-connectors) to learn more
...
 | `/query_limit`                              | Query Limit             | Maximum number of rows to fetch in a query. Will be divided between all connections if `/concurrency` > 1                             | int                                                                                                                              | 100,000 Rows                        |
-| `/query_timeout`                            | Query Timeout           | Timeout for queries. Typically left as the default as some tables just take a very long time to respond.                              | [`ISO8601` Duration](https://www.digi.com/resources/documentation/digidocs/90001488-13/reference/r_iso_8601_duration_format.htm) | `PT10M` (10 Minutes)                |
+| `/select_columns`         | Manually Selected Columns     | Override the columns to load from the table. If empty, all columns will be loaded. Ideally this should only be set when loading specific columns is neccesary, as it won't automatically update when new columns are added or removed. | string array | `[]`  |
```
jshearer: nit: `neccesary` => `necessary`

site/docs/reference/Connectors/capture-connectors/netsuite-suiteql.md
```diff
@@ -0,0 +1,208 @@
...
+
+   3. Scroll to the **Permissions** section.
+
+   4. (IMPORTANT) Click **Transactions** and add all the dropdown entities with either **full** or **view** access level.
+
+   5. (IMPORTANT) Click **Reports** and add all the dropdown entities with either **full** or **view** access level.
+
+   6. (IMPORTANT) Click **Lists** and add all the dropdown entities with either **full** or **view** access level.
+
+   7. (IMPORTANT) Click **Setup** an add all the dropdown entities with either **full** or **view** access level.
```
jshearer: nit: `an add` => `and add`

site/docs/reference/Connectors/capture-connectors/netsuite-suiteanalytics.md
```diff
@@ -126,57 +145,76 @@ See [connectors](../../../concepts/connectors.md#using-connectors) to learn more
...
 | `/name`                                     | Name                    | The name of the table this binding refers to                                                                                          | string                                                                                                                           | Required                            |
 | `/interval`                                 | Interval                | How frequently to check for incremental changes                                                                                       | [`ISO8601` Duration](https://www.digi.com/resources/documentation/digidocs/90001488-13/reference/r_iso_8601_duration_format.htm) | `PT1H` (1 Hour)                     |
+| `/schedule`         | Schedule     | Schedule to automatically rebackfill this binding. Accepts a cron expression.      | string |   |
```
jshearer: We should probably call out a bit more of the complexity involved here, as it's not trivial. Here is the relevant portion from the connector:

```python
    # We need a schedule if and only if we have a key and page cursor, but no log cursor.
    # Snapshots, which happen when we don't have a key, already run on a schedule
    # defined by `interval`, and should not be periodically re-backfilled as they depend
    # on keeping track of the previous row count in order to properly issue deletions.
    # We also do _not_ want to schedule periodic re-backfills if we are actually able to
    # load incrementally via an association from another table.
    if (
        key != ["/_meta/row_id"]
        and initial_page_cursor != ""
        and initial_log_cursor == ""
        and len(hinted_associations_for_table(table.table_name, config)) == 0
    ):
        needs_schedule = True
```

So we should indicate that `schedule` is only meaningful in cases where you want to do periodic full-refreshes using the paginated backfill mode. Otherwise, if you're using snapshot mode, `schedule` should explicitly _not_ be set as snapshots manage their own periodic refreshes (using `interval`) and state internally. Backfilling a snapshot binding (like with `schedule`, or manually) will just cause it to not emit deletions events properly.

site/docs/reference/Connectors/capture-connectors/netsuite-suiteanalytics.md
```diff
@@ -126,57 +145,76 @@ See [connectors](../../../concepts/connectors.md#using-connectors) to learn more
 | Property                      | Title                  | Description                                                                                      | Type   | Required/Default |
 | ----------------------------- | ---------------------- | ------------------------------------------------------------------------------------------------ | ------ | ---------------- |
 | `/account`                     | Netsuite Account ID    | Netsuite realm/Account ID e.g. 2344535, as for `production` or 2344535_SB1, as for `sandbox` | string | Required         |
-| `/role_id`                    | Role ID                | The ID of the role you created. Defaults to 3, which is the ID of the administrator role.        | int    | 3                |
-| `/connection_type`            | Connection Type        | Type of connection to use. `suiteanalytics` is preferred over `suiteql`.                      | string | Required |
 | `/suiteanalytics_data_source` | Data Source            | Which NetSuite data source to use. Options are `NetSuite.com`, or `NetSuite2.com`                | string | Required         |
```
jshearer: Should probably indicate that this is deprecated. Netsuite is killing `Netsuite.com` datasource very soon: 
> The NetSuite.com data source is no longer supported as of 2025.1 and will be removed in 2026.1.

[[1]](https://docs.oracle.com/en/cloud/saas/netsuite/ns-online-help/article_0405125116.html)

AFAIK, 2026.1 already went out to most customers, so tbh I would say just pre-emptively get rid of this property in the docs as it's likely not meaningful anymore.

site/docs/reference/Connectors/capture-connectors/netsuite-suiteanalytics.md
aeluce: 🤦

site/docs/reference/Connectors/capture-connectors/netsuite-suiteanalytics.md
aeluce: This is probably a sign for me to finally go through and remove this paragraph from all the connectors. We don't go into further depth on what to do with this info in the connector page, so yes, as-is it detracts from the intended flow of working with the connector.

When working directly with YAML files, being able to see past versions and choose a version could be helpful if we had detailed change logs and extended support for past versions. But again, not how we do things right now. 😅

site/docs/reference/Connectors/capture-connectors/netsuite-suiteanalytics.md
aeluce: Thanks! Added a section at the end of the doc. Let me know if that captures it all.

PR #2562, opened 2025-12-17
Docs: Standardize Estuary product name

aeluce:
**Description:**

We variously call our service "Estuary," "Estuary Flow," or "Flow" throughout our docs and other resources. This PR is an attempt to standardize our language usage.

* In most possible instances, the product should now simply be "Estuary" across our docs and in the main repo README and examples.
* We do still have the concept of "Data Flows," and I opted to change instances of "Flow specification file" to "Data Flow specification file" and similar to help explain `flow.yaml` and `flowctl`.
* Some connectors still have properties, property descriptions, or default property values that include `flow`. I generally left the docs for these alone to match up with the user experience and avoid breakages. The "Data Flow" concept may also be able to explain many of these, though we may wish to update some connectors' spec schemas eventually.

There are some other minor updates to language on the main README and docs landing page to match how we currently describe the platform. Other than that, I attempted to keep changes focused, though a few typo fixes may have made their way into the PR as well.

cc @SourabhEstuary for awareness

Almost every docs page

**Notes for reviewers:**

The platform docs may warrant more attention than the connector docs for review. Since the connector references are fairly formulaic, as long as the most-used connectors (like Postgres, Snowflake, etc) are reviewed, other connector docs are likely to be similar.

### DIFF
--- a/site/docs/reference/Connectors/capture-connectors/zoho-crm.md
+++ b/site/docs/reference/Connectors/capture-connectors/zoho-crm.md
@@ -1,6 +1,6 @@
 # Zoho CRM
 
-This connector captures data from Zoho CRM modules into Flow collections.
+This connector captures data from Zoho CRM modules into Estuary collections.
 It uses Zoho's [Bulk API 2.0](https://www.zoho.com/crm/developer/docs/api/v7/bulk-read/overview.html) and [COQL API](https://www.zoho.com/crm/developer/docs/api/v7/COQL-Overview.html).
 
 This connector offers several unique advantages:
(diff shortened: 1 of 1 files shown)


aeluce:
With a change like this, I think there's going to be cleanup efforts for a while. I don't think we need to wait for connector field descriptions, etc, to change before getting the main updates out.

That said, what changes were you thinking would be needed for yaml specs? Ex. we can keep existing terminology like `flow.yaml` files. The description in docs about data flow specification files would simply explain the connection, that "flow" files relate to your data flow.

site/docs/concepts/connectors.md
```diff
@@ -7,41 +7,41 @@ import TabItem from '@theme/TabItem';
...
 
 Most — if not all — of your Data Flows will use at least one connector.
 You configure connectors within capture or materialization specifications.
 When you publish one of these entities, you're also deploying all the connectors it uses.
 
-You can interact with connectors using either the Flow web application or the flowctl CLI.
+You can interact with connectors using either Estuary's web application or the flowctl CLI.
 
-### Flow web application
+### Web application
```
travjenkins, line 35: Maybe not for this PR as `web application` is used in a lot of places. But we might want to look into just calling this the `dashboard` since that is how we usually talk about it internally.

CONTRIBUTING.md
travjenkins, line 1: I assume this is not a _deeply_ binding document. So I assume this is safe. I'll also change the link to this document in UI as it says 'Estuary Flow'.

site/docs/concepts/connectors.md
aeluce: Definitely. That may end up being its own standardization task. We variously say "the web app," "the UI," "the dashboard," etc throughout our docs.

PR #2563, opened 2025-12-17
mise: fix vm:claude on macOS

mdibaiee:
**Description:**

- on macOS the keychain record is an API key, not JSON, and it should be added to `~/.claude.json`

**Workflow steps:**

**Notes for reviewers:**


mise/tasks/vm/create-post
```diff
@@ -25,6 +25,8 @@ if ! grep -q "mise" "${HOME}/.bashrc"; then
 source <(cd $(pwd) && ~/.local/bin/mise activate bash)
 source <(cd $(pwd) && ~/.local/bin/mise completion bash --include-bash-completion-lib)
 
+echo 'export PATH="$HOME/cargo-target/debug:$HOME/cargo-target/release:$PATH"' >> "${HOME}/.bashrc"
```
jgraettinger: I don't think we can assume this intent for everybody in all cases. It leads to easy confusion over which binary you're actually running, but this also misses musl targets.

mise/tasks/vm/claude
```diff
@@ -55,12 +57,9 @@ elif [[ "${HOST_OS}" == "linux" ]]; then
...
         exit 1
     fi
-fi
 
-# Copy .claude.json configuration file
-scp ~/.claude.json "${HOSTNAME}:~/.claude.json"
-
-# Set appropriate permissions
-ssh "${HOSTNAME}" 'chmod 700 ~/.claude && chmod 600 ~/.claude/.credentials.json'
+    # Set appropriate permissions
```
jgraettinger, line 70: This also applies to the MacOS flow on my box.

mise/tasks/vm/claude
```diff
@@ -36,12 +36,14 @@ echo "Syncing credentials from host to VM..."
...
+# Copy .claude.json configuration file
+scp ~/.claude.json "${HOSTNAME}:~/.claude.json"
+
 # Extract and copy credentials based on host OS
 if [[ "${HOST_OS}" == "mac" ]]; then
     # Extract from macOS Keychain and pipe directly to VM
-    if ! security find-generic-password -s "Claude Code-credentials" -w 2>/dev/null | \
-         jq -c 'del(.claudeAiOauth.refreshToken)' | \
-         ssh "${HOSTNAME}" 'cat > ~/.claude/.credentials.json'; then
+if ! security find-generic-password -s "Claude Code" -w 2>/dev/null | \
```
jgraettinger: Let's add this as a new case? I think it depends on how the key is set up. I have `Claude Code-credentials`, and the current flow works, but I don't have `Claude Code`.

mise/tasks/vm/claude
```diff
@@ -36,16 +36,26 @@ echo "Syncing credentials from host to VM..."
...
+
 # Extract and copy credentials based on host OS
 if [[ "${HOST_OS}" == "mac" ]]; then
-    # Extract from macOS Keychain and pipe directly to VM
-    if ! security find-generic-password -s "Claude Code-credentials" -w 2>/dev/null | \
-         jq -c 'del(.claudeAiOauth.refreshToken)' | \
-         ssh "${HOSTNAME}" 'cat > ~/.claude/.credentials.json'; then
+    # Try extracting from macOS Keychain with both possible service names
+    if security find-generic-password -s "Claude Code-credentials" -w 2>/dev/null | \
+        ssh "${HOSTNAME}" 'cat > ~/.claude/.credentials.json'; then
```
jgraettinger, line 47: compare to the diff -- this dropped logic to strip the refresh token, which shouldn't go into the VM (otherwise it breaks the host's single-use refresh token)

jgraettinger (approved):
LGTM

PR #2564, opened 2025-12-17
reactor: add logging to help troubleshoot logging

psFried:
Add a log statement so that we can determine which file descriptors are used for which tasks.


jgraettinger (approved):
LGTM

PR #2566, opened 2025-12-18
dekaf: Implement e2e testing framework

jshearer:
### Summary

* E2E test framework: 
  * publishes test specs into namespaced prefixes
  * uses `source-http-ingest` to inject documents
  * uses existing Dekaf Kafka API client to interact with low-level Kafka APIs
  * uses `rdkafka` Rust crate to interact with Dekaf via high-level librdkafka consumer
* During testing I found some poorly specified behavior around collections with missing journals, so I added `CollectionStatus` in order to return a retryable error to consumers and reduce connection churn.
* I added `--spec-ttl` to allow tests to shorten the `TaskManager` spec refresh cycle and speed up tests that depend on changes to specs.
* A few changes to `KafkaApiClient`:
  - Plaintext connections are supported via `tcp://` URL scheme
  - SASL PLAIN authentication support
  - `--upstream-auth=none` flag to skip authenticating to upstream Kafka broker if it doesn't require auth (such as when running tests)
* Added Dekaf to CI:
  *  `dekaf-test` job in `platform-test.yaml`
  *  Exclude Dekaf tests in default nextest profile as they require things like a local Kafka broker to be running as a prerequisite. Add `dekaf-e2e` profile to run e2e tests explicitly 
  *  Mise tasks (`local:dekaf`, `local:dekaf-kafka`, `local:test-tenant`) for local Dekaf + Kafka services, provision `test/` tenant etc

### Test scenarios

| Test file | Scenarios |
|-----------|-----------|
| `collection_reset.rs` | Collection-reset behaviors such as dealing with leader epochs in Metadata, ListOffsets, Fetch, `FENCED_LEADER_EPOCH`, `UNKNOWN_LEADER_EPOCH`, `OffsetForLeaderEpoch`, etc |
| `not_ready.rs` | `LeaderNotAvailable` when journals don't exist |
| `list_offsets.rs` | Earliest/latest offset queries, unknown partition handling |
| `empty_fetch.rs` | Empty fetch responses don't break subsequent fetches |
| `basic.rs` | Document roundtrip |

---

I intentionally left out more complex tests from here such as higher volume load testing, testing with other high-level consumers such as sarama, franz-go, etc. This PR is already getting fairly large, and I wanted to get it reviewed and out the door rather than add more heft to it.

---

### Additional PRs adding more E2E tests based ontop of this branch:
* https://github.com/estuary/flow/pull/2588
* https://github.com/estuary/flow/pull/2597
* https://github.com/estuary/flow/pull/2625

Note: Some of those PRs are targeting `master` instead of `dekaf/collection_reset_with_e2e_tests` in order to get CI to run the tests. Once this PR is merged I'll mark them ready for review


crates/dekaf/tests/e2e/harness.rs
```diff
@@ -0,0 +1,799 @@
...
+    ///           strict_topic_names: false
+    ///     bindings:
+    ///       - source: test_collection
+    ///         resource: { topic_name: test_topic }
+    /// ```
+    pub async fn setup(test_name: &str, fixture_yaml: &str) -> anyhow::Result<Self> {
+        let suffix = format!("{:04x}", rand::random::<u16>());
+        // Use test/ prefix which is provisioned by `mise run local:test-tenant`
+        // with storage mappings and user grants for the local system user
+        let namespace = format!("test/dekaf/{test_name}/{suffix}");
```
jshearer: `test/` is a constant, should be defined as such

crates/dekaf/tests/e2e/harness.rs
```diff
@@ -0,0 +1,799 @@
...
+            .iter()
+            .find(|l| l.name == "estuary.dev/expose-port")
+            .map(|l| &l.value)
+            .context("no port label")?;
+
+        let base = endpoint.replace("https://", "");
+        let url = format!("https://{hostname}-{port}.{base}/{path}");
+
+        let client = reqwest::Client::builder()
+            .danger_accept_invalid_certs(true)
```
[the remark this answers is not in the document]
jshearer: test only

.cargo/config.toml
```diff
@@ -1,3 +1,6 @@
+[build]
+rustflags = ["-D", "warnings"]
```
jshearer: Made this change because I kept on running into annoying CI failures that I could have caught locally. Now `cargo check`/`cargo build` will fail on warnings like CI will.

psFried: There's also [this line](https://github.com/estuary/flow/blob/7a7dd8d258be4c8eec938cdcb4727203fad37eb9/mise.toml#L29) in mise.toml. TBH I think I like your approach better, assuming it would only apply to `cargo build` commands, and not `cargo check`. The current method of setting this applies to check as well, which causes the IDE to present them as errors instead of warnings. I personally find that annoying, but ofc others may disagree. Seems like we probably shouldn't have both of these set, though

crates/dekaf/src/api_client.rs
```diff
@@ -19,50 +19,72 @@ use std::{io::BufWriter, pin::Pin, sync::Arc};
...
+    let host = parsed
+        .host()
+        .context("missing host in broker URL")?
+        .to_string();
+    let port = parsed.port().unwrap_or(9092);
+    Ok((scheme, host, port))
+}
 
-    let tls_connector = tokio_rustls::TlsConnector::from(Arc::new(tls_config));
+static ROOT_CERT_STORE: OnceCell<Arc<RootCertStore>> = OnceCell::const_new();
```
jgraettinger: nit: **especially** for cases like tokio::sync::OnceCell vs std::cell::OnceCell, as a reader I deeply appreciate seeing a more explicit module path so I know which one we're talking about.

crates/dekaf/src/session.rs
```diff
@@ -262,18 +262,26 @@ impl Session {
...
-                self.build_topic_metadata(encoded_name, &coll)
+            .map(|(name, status)| {
+                let encoded_name = self.encode_topic_name(name.clone())?;
+                match status {
+                    CollectionStatus::Ready(coll) => self.build_topic_metadata(encoded_name, &coll),
+                    CollectionStatus::NotFound => {
+                        anyhow::bail!("Collection '{}' not found or not accessible", name)
+                    }
+                    CollectionStatus::NotReady => {
+                        // Collection exists but journals aren't available - return LeaderNotAvailable so clients will retry
```
jgraettinger, line 280: I presume you thought about this carefully, but: the "obvious" approach would represent such a collection as a topic with zero partitions. That breaks at a protocol level?

Or, it looks like we previously returned a single partition. I'm guessing this was a placeholder which was semantically empty, and then popped into actual existence with the first created journal? It looks like you junked this tactic to make the E2E tests work better (or were there other reasons, too?).

FYI a sentinel partition approach would fit well with the `tokens` work and [the task-manager sketch](https://github.com/estuary/flow/pull/2601#issuecomment-3745514743), which re-frames journal listing as a long-lived watch. Such a sentinel could block awaiting a journal listing update, be notified immediately when it's ready, and then transition to doing actual reads.

crates/dekaf/src/session.rs
```diff
@@ -1950,19 +2150,37 @@ impl Session {
...
-                .map(|p| {
-                    EpochEndOffset::default()
-                        .with_partition(p.partition)
-                        .with_error_code(ResponseError::UnknownTopicOrPartition.code())
-                })
-                .collect();
-            return Ok(OffsetForLeaderTopicResult::default()
-                .with_topic(topic.topic)
-                .with_partitions(partitions));
+        let collection = match Collection::new(auth, &collection_name).await? {
```
jgraettinger: The quantity of new code, here, matching on CollectionStatus is a little alarming / a smell. Do these all truly need to be such distinct cases? Is there no factoring that could be applied, to collapse some of the commonality of these paths?

Nothing here looks incorrect, per se, but the change _is_ violating my mental yardstick re: how much churn such a change should result in.

.cargo/config.toml
jgraettinger: I'm guessing this is when building on your host, but you're doing it outside of using mise? Would you try using mise first instead of this change? It's designed to work well on hosts outside of VMs, as well as within VMs.

.github/workflows/platform-test.yaml
```diff
@@ -65,3 +65,44 @@ jobs:
...
+
+      - name: Cache/Restore Go workspace.
+        uses: actions/cache@v4
+        with:
+          path: ~/go/pkg/mod
+          key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
+
+      - uses: mozilla-actions/sccache-action@v0.0.9
+      - run: echo 'SCCACHE_GHA_ENABLED=true' >> $GITHUB_ENV
+      - run: mise run build:rocksdb
```
jgraettinger: Break this out into a separate workflow, dekaf-test.yaml ?

Also please add a mise task `ci:dekaf-test` paralleling `ci:platform-test` so that there's a one-shot way to run these in a dev VM.

jgraettinger (commented):
This looks great! Nice job and great to see this level of test coverage. A few comments below for discussion but nothing big.

.cargo/config.toml
jshearer: I can get rid of it and just maintain this change locally instead, it doesn't need to block this PR. FWIW though, I feel like I remember that even running on a GCP dev VM created by mise, a plain old `cargo check` or `cargo build` will pass with warnings whereas in CI it'll fail.

**Edit:** Removed it from this PR

crates/dekaf/src/api_client.rs
jshearer: Yeeeep, good point. Fixed.

.github/workflows/platform-test.yaml
jshearer: I broke the gh action into a separate file. WRT `ci:dekaf-test`, do you mean something other than [`ci:dekaf-e2e`](https://github.com/estuary/flow/blob/7452c05d034bb6440b1fd7903eaba9e04ef4166b/mise/tasks/ci/dekaf-e2e)?

crates/dekaf/src/session.rs
jshearer: I changed it from returning a single partition because consumers would then try to `ListOffsets`/`Fetch` from that partition and get `UnknownTopicOrPartition`, which is a permanent error, so they wouldn't retry. In practice they (usually) eventually do retry when the application logic times out and restarts the consumer, but it felt smelly to have that in the tests.

I tried returning no partitions instead, but the consumer also treated that as a "permanent" situation for the lifetime of the session. `LeaderNotAvailable` is supposed to be a transient condition and it causes the consumer to retry according to its retry policy.

crates/dekaf/src/session.rs
jshearer: Yeah, agreed that the size of the diff is big. Some of it is real behavior improvements that I noticed during testing, for example `metadata` now returns topic-specific errors instead of failing the whole request, the group management APIs now validate that collections exist and are available whereas before they would let you interact with groups (join/sync/commit offset etc) with topics that didn't map to any extant collection at all, etc.

Still, I took another pass over it and ended up moving the Kafka error code matching logic to a couple of helpers on `CollectionStatus`. So for example instead of having the error codes spread throughout `session`, like this:
https://github.com/estuary/flow/blob/109edbf6fc39fe58d5dd8a4bc07c0ccf692cb2c2/crates/dekaf/src/session.rs#L1659-L1672

You can instead just do this:
https://github.com/estuary/flow/blob/76e90af5a2bddcf3c6fa860d1df9151ea314571a/crates/dekaf/src/session.rs#L1636-L1638

crates/dekaf/src/api_client.rs
```diff
@@ -255,26 +256,45 @@ async fn sasl_auth(
 
     let mut state_buf = BufWriter::new(Vec::new());
     let mut state = session.step(None, &mut state_buf)?;
+    // Flush the BufWriter to ensure all data is written to the underlying Vec
```
psFried: nit: I'm supposing maybe `into_inner()` does a `flush` internally? Either way, this comment seems a bit confusing, as it says "flush" but then does not call flush

crates/dekaf/src/api_client.rs
```diff
@@ -255,26 +256,45 @@ async fn sasl_auth(
...
                 .map(|code| format!("{code:?}"))
-                .unwrap_or(format!("Unknown error {}", handshake_resp.error_code));
+                .unwrap_or(format!("Unknown error {}", auth_resp.error_code));
             bail!(
                 "Error performing SASL authentication: {err} {:?}",
                 auth_resp.error_message
             )
         }
+
+        if !state.is_running() {
```
psFried: nit: Why do we need to repeat this check here? Seems maybe worthy of a comment

psFried (approved):
LGTM 🎉 Nice work on the tests. This looks like a huge improvement

crates/dekaf/src/api_client.rs
jshearer: yeah fair enough, the flushing aspect of `into_inner()` is a bit confusing to point out here vs the semantic meaning of "get me the full value of the inner `Vec` (which inherently requires flushing the writer first)". I just removed the comment

crates/dekaf/src/api_client.rs
jshearer: I ended up refactoring this loop a bit more so it's clearer what's going on:

```rust
    loop {
        let mut buf = BufWriter::new(Vec::new());
        let state = session.step(server_response.as_deref(), &mut buf)?;
        let client_message = buf.into_inner()?;

        match (state.is_running(), client_message.is_empty()) {
            // Still running, send and continue
            (true, _) => {
                let resp = send_sasl_message(conn, client_message).await?;
                server_response = Some(resp.auth_bytes.to_vec());
            }
            // Finished, but have final message to send first
            (false, false) => {
                send_sasl_message(conn, client_message).await?;
                break;
            }
            // Finished, nothing more to send
            (false, true) => break,
        }
    }
```

PR #2568, opened 2025-12-18
agent-api: fix sub-second precision mismatch in agent-api causing incorrect terminal authorization errors

jshearer:
**Description:**

The authorization evaluator compared `snapshot.taken` which contains sub-second precision against the JWT's `iat` claim which only contains second-level precision. This caused the authorization API to return terminal failures when fetching auth within the same second as a snapshot refresh.

Example timeline:
```
> Snapshot fetched at 15:32:44.76
2025-12-18T15:32:44.767374Z INFO control_plane_api::server::snapshot: fetched authorization snapshot collections=8 data_planes=1 migrations=0 role_grants=7 user_grants=4 tasks=7

>  Publication committed at 15:32:44.86
2025-12-18T15:32:44.861475Z INFO try_publish:commit{build_id=13555487ce800352}: control_plane_api::publications: finished publication commit attempt=0 attempt_elapsed_ms=5 commit_elapsed_ms=5 result="success" temp_dashboard_compat="successfully committed publication"

> Task started and attempted to fetch authorization at 15:32:44.98
2025-12-18T15:32:44.981443Z WARN authorize_task: control_plane_api::server::authorize_task: error=status: 412 Precondition Failed, error: task shard capture/test/my-task/source_ingest/13555487ce800352/00000000-00000000 within data-plane local-cluster.dp.estuary-data.com is not known
```

The comparison `snapshot.taken > started` evaluated as: `15:32:44.763 > 15:32:44.000 = TRUE` (terminal failure)

But the snapshot (taken at `.763`) was actually OLDER than the request (`.981`).

The fix is to truncate `snapshot.taken` to seconds when comparing.

### DIFF
--- a/crates/control-plane-api/src/server/snapshot.rs
+++ b/crates/control-plane-api/src/server/snapshot.rs
@@ -1,5 +1,6 @@
 use super::App;
 use anyhow::Context;
+use chrono::SubsecRound;
 use std::collections::HashMap;
 use std::sync::Arc;
 
@@ -219,7 +220,11 @@ impl Snapshot {
             }
             // Authorization is invalid and the Snapshot was taken after the
             // start of the authorization request. Terminal failure.
-            Err(err) if snapshot.taken > started => return Err(Err(err)),
+            //
+            // Because JWT `iat` is second-precision, we should only terminally fail
+            // if the snapshot was taken in a later second, otherwise we can't say
+            // definitively that the request is invalid.
+            Err(err) if snapshot.taken.trunc_subsecs(0) > started => return Err(Err(err)),
 
             // Authorization is valid but is currently cordoned, and we must
             // hold it in limbo until the cordoned condition is resolved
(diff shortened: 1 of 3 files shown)


crates/control-plane-api/src/server/snapshot.rs
```diff
@@ -219,7 +220,11 @@ impl Snapshot {
             }
             // Authorization is invalid and the Snapshot was taken after the
             // start of the authorization request. Terminal failure.
-            Err(err) if snapshot.taken > started => return Err(Err(err)),
+            //
+            // Because JWT `iat` is second-precision, we should only terminally fail
+            // if the snapshot was taken in a later second, otherwise we can't say
+            // deffinitively that the request is invalid.
```
jgraettinger: nit: spelling

jgraettinger (approved):
LGTM

crates/control-plane-api/src/server/snapshot.rs
jshearer: 🤦 thx

PR #2571, opened 2025-12-19
DOCS : `npm audit` and major `redirect` update

travjenkins:
**Description:**

**Workflow steps:**

**Notes for reviewers:**

### DIFF
--- a/.github/workflows/docs-ci-previews.yaml
+++ b/.github/workflows/docs-ci-previews.yaml
@@ -24,7 +24,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-node@v3
         with:
-          node-version: 18
+          node-version: 20
           cache: npm
           cache-dependency-path: ./site/package-lock.json
       - name: Install dependencies
--- a/site/docusaurus.config.js
+++ b/site/docusaurus.config.js
@@ -114,10 +114,6 @@ const config = {
             to: '/reference/Connectors/capture-connectors/HubSpot-real-time/'
           },
           {
-            from: '/overview/comparisons/',
-            to: '/getting-started/comparisons/'
-          },
-          {
             from: '/guides/configure-cloud-storage/',
             to: '/getting-started/installation/'
           },
--- a/site/docusaurus.config.js
+++ b/site/docusaurus.config.js
@@ -48,11 +48,16 @@ const config = {
   baseUrl: BASE_URL,
   onBrokenAnchors: 'warn', // TODO(johnny): Fix broken links and make this 'throw'.
   onBrokenLinks: 'throw',
-  onBrokenMarkdownLinks: 'throw',
   favicon: 'img/favicon-2.ico',
   organizationName: 'estuary',
   projectName: 'flow',
   trailingSlash: true,
+
+  markdown: {
+    hooks: {
+      onBrokenMarkdownLinks:'throw',
+    },
+  },
 
   scripts: [
     {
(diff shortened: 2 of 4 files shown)


.github/workflows/docs-ci-previews.yaml
travjenkins, line 1: Ran a `local` build and the `preview` was able to run a build using `20` so we should be good.

site/docusaurus.config.js
```diff
@@ -48,12 +48,17 @@ const config = {
...
   favicon: 'img/favicon-2.ico',
   organizationName: 'estuary',
   projectName: 'flow',
   trailingSlash: true,
 
+  markdown: {
+    hooks: {
+      onBrokenMarkdownLinks:'throw',
+    },
+  },
```
travjenkins, line 60: Having this in root has been deprecated.

https://docusaurus.io/docs/api/docusaurus-config#onBrokenMarkdownLinks

site/docusaurus.config.js
travjenkins, line 1: These redirects were any that Google indexing was complaining about and still were 404ing in production.

site/docusaurus.config.js
```diff
@@ -89,13 +94,74 @@ const config = {
...
+            from : '/reference/Connectors/materialization-connectors/Redshift/',
+            to: '/reference/Connectors/materialization-connectors/amazon-redshift/',
+          },
+          {
+            from : '/reference/Connectors/capture-connectors/netsuite-odbc/',
+            to: '/reference/Connectors/capture-connectors/netsuite-suiteanalytics/',
+          },
+          {
+            from: '/reference/Connectors/capture-connectors/hubspot-real-time/',
+            to: '/reference/Connectors/capture-connectors/HubSpot-real-time/'
```
aeluce, line 114: So, this branch starts up locally okay for me, but when I run `npm run build`, I get a `Redirect file creation error` for the HubSpot redirect:

```
[ERROR] Error: Unable to build website for locale en.
    {…}
  [cause]: Error: The redirect plugin is not supposed to override existing files.
```

I believe I ran into this before; it doesn't seem to keep the docs site from building, but it does essentially block other warnings from displaying, making it harder to track down broken anchors.

Any thoughts on how to get around this?

travjenkins: Can you try clearing the `cache` maybe? This is basically me just guessing.

`npm run clear`


In addition - you'll need to make sure you're running with Node `v20`. I am using `v20.17.0` specifically.

aeluce: Thanks! Somewhere along the line, I was using v22... Updated that, cleared the cache, and I'm still seeing the error when building.

site/docusaurus.config.js
```diff
@@ -89,13 +94,74 @@ const config = {
...
+          // {
+          //   from : '/reference/Connectors/capture-connectors/netsuite-suitetalk/',
+          //   to: '/reference/Connectors/capture-connectors/netsuite-suiteql/',
+          // },
+
+          // There is no longer a landing page. Will need to make one or select a specific
+          //  parquet connector now. 
+          // {
+          //   from: '/reference/Connectors/materialization-connectors/Parquet/',
+          //   to: '???'
```
aeluce: An admittedly brief search isn't turning up this particular path, but `/reference/Connectors/materialization-connectors/apache-parquet/`, when it was around, was specifically for S3. So I think we can redirect to `/reference/Connectors/materialization-connectors/amazon-s3-parquet/`.

site/docusaurus.config.js
```diff
@@ -89,13 +94,74 @@ const config = {
...
+            from: '/reference/Connectors/dekaf/dekaf-tinybird/',
+            to: '/reference/Connectors/materialization-connectors/Dekaf/tinybird/'
+          },
+          {
+            to: '/getting-started/comparisons/',
+            from: '/overview/comparisons/',
+          },
+          {
+            to: '/getting-started/who-should-use-flow/',
+            from: '/overview/who-should-use-flow/',
```
aeluce, line 150: Do you want these fields to be in a particular order? I see the ones originally ordered to/from were reordered to from/to, and these last two redirects break that pattern.

travjenkins: I always think of redirects as `to` -> `from` and it makes sense in my brain :shrug: so I just went with it since there weren't a lot of them yet.

aeluce (approved):
LGTM!

PR #2576, opened 2025-12-19
DOCS: Fixing security scan issue 

travjenkins:
https://github.com/estuary/flow/security/code-scanning/8

**Description:**

Escaping the `tenant` name
Adding in `state` to reduce chances of security issues (mentioned here https://learn.microsoft.com/en-us/entra/identity-platform/v2-oauth2-auth-code-flow#request-an-authorization-code)

**Workflow steps:**

This will impact how Azure gets setup. This still needs tested

**Notes for reviewers:**

Used fake code and incorrect session storage
image

Used fake code and correct session storage 
image

### DIFF
--- a/site/docs/getting-started/azureAuthorize.jsx
+++ b/site/docs/getting-started/azureAuthorize.jsx
@@ -28,7 +28,6 @@ export const AzureAuthorizeComponent = () => {
     const [tenant, setTenant] = React.useState("");
 
     // Try to get the auth code but first ensure the states match
-    //  if there is an issue this will fail silently (Q4 2025)
     const authCode = React.useMemo(() => {
         const params = new URLSearchParams(window.location.search);
         const code = params.get(SETTINGS.responseType);
--- a/site/docs/getting-started/azureAuthorize.jsx
+++ b/site/docs/getting-started/azureAuthorize.jsx
@@ -1,19 +1,51 @@
 import React from "react";
 
+const SETTINGS = {
+    appId: "42cb0c6c-dab0-411f-9c21-16d5a2b1b025",
+    redirectUri: "https://eyrcnmuzzyriypdajwdk.supabase.co/functions/v1/azure-dpc-oauth",
+    resourceId: 'https://storage.azure.com',
+    responseType: 'code',
+    stateKey: 'state',
+    storageKey: 'docs_est:azure_oauth_state'
+}
+
+const generateAuthorizeUrl = (theirTenant) => {
+    const state = crypto.randomUUID()
+    sessionStorage.setItem(SETTINGS.storageKey, state);
+
+    const params = new URLSearchParams({
+      client_id: SETTINGS.appId,
+      redirect_uri: SETTINGS.redirectUri,
+      resource_id: SETTINGS.resourceId,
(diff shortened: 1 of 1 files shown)


site/docs/getting-started/azureAuthorize.jsx
```diff
@@ -1,19 +1,51 @@
...
+    redirectUri: "https://eyrcnmuzzyriypdajwdk.supabase.co/functions/v1/azure-dpc-oauth",
+    resourceId: 'https://storage.azure.com',
+    responseType: 'code',
+    stateKey: 'state',
+    storageKey: 'docs_est:azure_oauth_state'
+}
+
+const generateAuthorizeUrl = (theirTenant) => {
+    const state = crypto.randomUUID()
+    sessionStorage.setItem(SETTINGS.storageKey, state);
```
travjenkins, line 14: This is the MAJOR update here. We need to store this off and validate it down below.

site/docs/getting-started/azureAuthorize.jsx
```diff
@@ -1,19 +1,51 @@
...
-        `https://login.microsoftonline.com/${theirTenant}/oauth2/authorize?client_id=${ourAppId}&response_type=code&redirect_uri=${encodeURIComponent(
-            redirectUri
-        )}&resource_id=${encodeURIComponent(resourceId)}`;
+    const params = new URLSearchParams({
+      client_id: SETTINGS.appId,
+      redirect_uri: SETTINGS.redirectUri,
+      resource_id: SETTINGS.resourceId,
+      response_type: SETTINGS.responseType,
+      [SETTINGS.stateKey]: state,
+    });
```
travjenkins, line 22: Setting up the params in this way should handle escaping for us.

site/docs/getting-started/azureAuthorize.jsx
```diff
@@ -1,62 +1,102 @@
...
+        return null;
     }, []);
 
+    if (authCode === "error") {
+        return (
+            <span style={{ color: "red" }}>
+                We were unable to verify this worked. Please contact support.
+            </span>
+        ); 
+    }
```
travjenkins, line 56: This is where things get weird. Normally in an oAuth situation we might consume this `code` and do something with it. However, here we are just seeing that things are setup. So I think this is the best we could do here 🤷 .

Like if this happens either their browser messed up or possibly someone messed with their stuff and they just setup something wrong.

site/docs/getting-started/azureAuthorize.jsx
```diff
@@ -1,62 +1,101 @@
...
+            return code;
+        }
+
+        return null;
     }, []);
 
+    if (authCode === "error") {
+        return (
+            <span style={{ color: "red" }}>
+                We were unable to verify this worked. Please contact support.
```
mdibaiee: I would say "try again and if persists, contact support"

mdibaiee (approved):
LGTM, let me know if you need help testing this

site/docs/getting-started/azureAuthorize.jsx
travjenkins: Updated. Took the same content we use for the dashboard.

PR #2583, opened 2025-12-30
Phil/alert evals

### LINKED ISSUE #2506: Replace alerts cloudrun function with a agent automation tasks
Alert evaluation is occasionally hitting the postgres statement timeout, causing all alert evaluation to fail until the next attempt. This makes our alerting less reliable. We've also seen cases where sending alert notification emails fails, and users miss important notifications. So we'd like to make alerting more robust, and easier to monitor. The plan is:

- Add a new `AlertNotifications` automation task, and a corresponding executor in `agent` for sending alert emails for a specific alert instance. A task of this type will get created whenever an alert starts firing. The task should stick around while the alert is active. After the alert resolves, this task will send the final resolution email and then complete, deleting itself by returning `Action::Done` from the `Outcome`.
- Alter the `alert_history` table to add a nullable `task_id flowid` column. This column should be populated with a newly generated `flowid` when a new row is inserted.
- Alter the insert trigger on `alert_history` to create an `AlertNotifications` task for each new alert, with a `task_id` that's the same as the `alert_history.task_id`. The newly created task should be immediately queued to run.
- Alter the update trigger on `alert_history` to send a final `Resolved` message to the `AlertNotifications` task
- Add a new alert evaluation automation task for tenant-related alerts (free_trial, free_trial_ending, etc), which should periodically query the `tenant_alerts` view and insert/update `alert_history`
- Add a new alert evaluation automation task for `data_movement_stalled` alerts, which should periodically query the corresponding alert view and insert/update `alert_history`
- Update the live spec controllers automation executor to insert/update `alert_history` in response to changes in the `alerts` sub-status, after each controller run.
- Remove the alerts cloudrun function

psFried:
**Description:**

Refactors alerts, and makes them more reliable.
Fixes #2506 
Also related: #1686 

Our previous alerting system suffered from a few different reliability issues:

- Alert evaluations would sometimes run into postgres statement timeouts, meaning no alerts would be fired or resolved until the next successful evaluation 10 minutes later
- Alert notification requests sometimes fail, meaning no notification would be sent for some alerts and resolutions. This was especially common because Resend has a 9 req/s rate limit, so we'd generally fail at least one notification whenever there's more than 9 alerts firing or resolving at the same time (which was common since we evaluated all alerts at the same time).

The new design fixes those issues, and adds a few other improvements. The basic idea is to use the `automations` crate to manage a few new background tasks for evaluating alerts and sending any notifications.

- A new `AlertNotifications` task gets created for every alert, as part of adding it to `alert_history`. This task is responsible for sending all of the notifications pertaining to a single alert. After the alert has resolved and all resolution notifications have been sent, the task will delete itself.
- Two new `AlertEvaluator` tasks get created as part of the migration. One evaluates `data_movement_stalled` alerts, and the other evaluates all of the tenant-related alerts (`free_trial`, `missing_payment_method`, etc). They each just add/update `alert_history`.
- Controller runs now return a set of `AlertAction`s, which will fire and resolve alerts while committing the controller outcome.

The `alert_history` table structure is unchanged except for the addition of an `id` column. The `id` is essentially a surrogate key, though the original `(catalog_name, alert_type, fired_at)` primary key was left in place. The shape of the `arguments` and `resolved_arguments` JSON remains as it was, with the resolved `recipients` being embedded in it. I have regrets about the `recipients` being embedded like that, but I'm not sure that it's worth changing at this point, and I thought it would help to keep the transition as simple as possible. 

**Deployment plan:**

- Run the migration, which will disable the old alerting system
- Deploy new agent version, which starts the new one

Alerts that were fired by the legacy system can be resolved by the new system. The migration creates a notification task for each open alert, and sets the state to indicate that the "fired" notifications have already been sent. When the alert resolves, the new notification system will send the emails.

If we need to rollback, we can do so without much risk by deploying the previous agent version and re-enabling the evaluation cron job and `alert_history` triggers.

### DIFF
--- a/crates/notifications/src/missing_payment_method.rs
+++ b/crates/notifications/src/missing_payment_method.rs
@@ -15,7 +15,7 @@ pub fn register_templates<'a>(registry: &mut handlebars::Handlebars<'a>) -> anyh
     registry
         .register_template_string(
             &resolved_body,
-            r#"<p class="body-text">We hope you are enjoying Estuary Flow. We have received your payment method for your account <span class="identifier">{{arguments.tenant}}</span>. {{#if (eq arguments.plan_state "free_trial")}}After your free trial ends on <strong>{{arguments.trial_end}}</strong>, you will automatically be switched the paid tier.{{else}}you are now on the paid tier.{{/if}}</p>
+            r#"<p class="body-text">We hope you are enjoying Estuary Flow. We have received your payment method for your account <span class="identifier">{{arguments.tenant}}</span>. {{#if (eq arguments.plan_state "free_trial")}}After your free trial ends on <strong>{{arguments.trial_end}}</strong>, you will automatically be switched the paid tier.{{else}}You are now on the paid tier.{{/if}}</p>
 <a href="https://dashboard.estuary.dev/admin/billing" class="button">📈 See your bill</a>
 
 <hr style="border: 0; border-top: 1px dashed lightgrey; margin: 40px 0 10px 0;">
--- a/crates/notifications/src/snapshots/notifications__tests__missing_payment_method_resolved_paid_body.snap
+++ b/crates/notifications/src/snapshots/notifications__tests__missing_payment_method_resolved_paid_body.snap
@@ -40,7 +40,7 @@ expression: email.body
         </div>
         <div class="divider"></div>
         <div class="email-content">
-            <p class="body-text">We hope you are enjoying Estuary Flow. We have received your payment method for your account <span class="identifier">acmeCo/</span>. you are now on the paid tier.</p>
+            <p class="body-text">We hope you are enjoying Estuary Flow. We have received your payment method for your account <span class="identifier">acmeCo/</span>. You are now on the paid tier.</p>
             <a href="https://dashboard.estuary.dev/admin/billing" class="button">📈 See your bill</a>
             
             <hr style="border: 0; border-top: 1px dashed lightgrey; margin: 40px 0 10px 0;">


crates/notifications/src/missing_payment_method.rs
```diff
@@ -0,0 +1,49 @@
...
+        .register_template_string(
+            &resolved_subject,
+            r#"Estuary: Thanks for Adding a Payment Method🎉"#,
+        )
+        .context("registering missing_payment_method-resolved-subject template")?;
+
+    registry
+        .register_template_string(
+            &resolved_body,
+            r#"<p class="body-text">We hope you are enjoying Estuary Flow. We have received your payment method for your account <span class="identifier">{{arguments.tenant}}</span>. {{#if (eq arguments.plan_state "free_trial")}}After your free trial ends on <strong>{{arguments.trial_end}}</strong>, you will automatically be switched the paid tier.{{else}}you are now on the paid tier.{{/if}}</p>
```
jgraettinger: nit: **Y**ou are now on the paid tier. (capitalization)

crates/notifications/Cargo.toml
```diff
@@ -0,0 +1,58 @@
...
+doc = { path = "../doc" }
+migrate = { path = "../migrate" }
+gazette = { path = "../gazette" }
+json = { path = "../json" }
+labels = { path = "../labels" }
+locate-bin = { path = "../locate-bin" }
+models = { path = "../models", features = ["async-graphql", "sqlx-support"] }
+ops = { path = "../ops" }
+proto-flow = { path = "../proto-flow" }
+proto-gazette = { path = "../proto-gazette" }
```
jgraettinger: nit: perhaps I missed it, but I don't recall seeing many of these crates. Overly-broad copy/paste?

crates/control-plane-api/src/alerts.rs
```diff
@@ -0,0 +1,346 @@
...
+
+    tracing::info!(%id, %catalog_name, %alert_type, %resolved_at, "resolved alert");
+
+    Ok(())
+}
+
+pub async fn fetch_open_alerts_by_type(
+    alert_types: &[AlertType],
+    pool: &sqlx::PgPool,
+) -> sqlx::Result<Vec<Alert>> {
```
jgraettinger, line 276: This could potentially return a very large result set, right? Should this use `fetch()` instead of `fetch_all()`, in order to return a stream that the caller can step through?

crates/agent/src/lib.rs
```diff
@@ -24,6 +25,25 @@ lazy_static! {
...
+/// shards are reseved for what.
+pub mod id_generator {
+    pub fn tenant_alerts() -> models::IdGenerator {
+        models::IdGenerator::new(1)
+    }
+    pub fn data_movement_stalled_alerts() -> models::IdGenerator {
+        models::IdGenerator::new(2)
+    }
+
+    pub fn with_random_shard() -> models::IdGenerator {
```
jgraettinger, line 39: My intuition is that we should consider a stable process-global IdGenerator, which is initialized / logged at startup using `hostname` so that distinct pods (001, 002, etc) are highly likely to have distinct id generation shards with values > 0 (as the DB uses shard zero IIRC).

Randomization will work _mostly_, but the shard ID space is somewhat narrow, and if we see weird ID race effects we'll never be certain if it was due to a happenstance collision of shard & clock.

(If we change this, it can / should be a separate PR to not block release of this work)

crates/agent/src/alerts/evaluator.rs
```diff
@@ -0,0 +1,452 @@
...
+/// alerts that have resolved. But it's also possible to include the alert
+/// with `firing: false`, in which case any `base_arguments` will become the
+/// `resolved_arguments` in alert history.
+pub fn evaluate_alert_actions(
+    eval_timestamp: DateTime<Utc>,
+    id_gen: &mut models::IdGenerator,
+    open_alerts: Vec<Alert>,
+    current_alert_state: Vec<AlertViewRow>,
+) -> Vec<AlertAction> {
+    let mut indexed_open: HashMap<(String, AlertType), (models::Id, DateTime<Utc>, ArgsObject)> =
```
jgraettinger, line 54: Why do this join in-process, instead of in the database? To state the obvious downside, it means moving (potentially a lot of) rows from DB => process memory to evaluate down to a small number of transitions. We also have no lever to bound the number of rows we're talking about (though acknowledged that it's almost certainly fine, today).

crates/agent/src/alerts/notifier.rs
```diff
@@ -0,0 +1,312 @@
...
+        // Note on retries: We don't technically need to handle retries here, as
+        // we could instead return and just schedule ourselves to run again.
+        // It's common for many alerts to fire more or less simultaneously, and
+        // the resend rate limit is only 9 req/s. So my current thinking is that
+        // it's better to handle retries here, so that we don't end up having a
+        // bunch of other notifier tasks run and then have this one hit another
+        // rate limit error when we retry it. Better to retry each notification
+        // until it succeeds. If we exhaust the number of retries, we'll return
+        // an error and back off somewhat longer.
+        let response = resend_rs::rate_limit::send_with_retry_opts(
```
jgraettinger, line 137: optional: AlertNotifications already has mechanics for jittered, backed-off retries on send errors. You could have the first send to an AlertNotification task react by _not_ actually attempting a send, but instead putting itself to sleep for an initial backoff of zero plus a jittered amount, after which it would make the first attempt. This would smooth out a burst of sends while collapsing all retries / jittering into a single mechanism.

crates/agent/src/lib.rs
psFried: Agreed. Something like `struct SharedIdGenerator(Arc<Mutex<IdGenerator>>);` seems better, but just requires some extra care to make sure we don't hold the mutex lock any longer than we need to. So separate PR makes sense to me

crates/agent/src/alerts/evaluator.rs
psFried: This was motivated mostly by trying to keep it DRY. If we have this logic in the database, it's going to pretty much look like our existing `evaluate_alert_events` function. Meaning that it's hard to factor out reusable logic for checking against open alerts in `alert_history`, without essentially lumping all the alerts together in the `alert_all` view.

So my thinking here is to try to break down alert evaluation into small enough parts that they can comfortably fit into memory. If we end up with more open alerts than could easily fit into memory (TBH I kinda doubt we ever will), then I'm thinking our next move could be to break up the `TenantAlerts` into separate evaluators for each alert type. Or for `data_movement_stalled` alerts, we could move evaluation into controllers.

One other thought is that if we wanted to reduce memory usage, it might actually be better to paginate (or stream) the alert _view_ results, rather than the set of open alerts. For tenant alerts, at least, the view is far larger since it contains alerts where `firing = false`.

crates/control-plane-api/src/alerts.rs
psFried: Using `fetch` instead of `fetch_all` is an easy way to halve the memory usage of this function, so I might as well. That said, I don't think any of these will be all that large, no. Based on currentish prod data, the most we'd ever load is about ~15k rows for the tenant alerts evaluation. The vast majority of that is the `missing_payment_method` alerts, which we'd expect to be firing for ~99% of the tenants in the system. But I think we could safely ~100X the number of tenants before this becomes a memory problem. And I think we'd also be good if we ~100X the number of tasks and they all fired data movement stalled alerts. Well, not _good_, but, you know...

crates/agent/src/alerts/notifier.rs
psFried: I think that seems like a reasonable idea, though I'm inclined to keep the current method for now. One issue I was seeing is that we have to identify rate-limiting errors and handle those differently from other types of errors by re-trying perhaps a bit more quickly and logging them at a lower severity. So using `send_with_retry_opts` seemed like an easy way to make that determination.

But more importantly, I think this will result in better behavior in situations where we're sending a lot of alerts and hitting the rate limits. If we suddenly fired a ton of alerts, then we could end up exacerbating the issue if we immediately give up after one failed attempt and then try the next notifier task. I'd worry that we'd end up generating a lot more requests that way, and handling the retries this way seemed like an easy way to avoid that.

crates/notifications/Cargo.toml
psFried: oops, good catch!

PR #2588, opened 2026-01-05
dekaf: Fix task names containing periods

jshearer:
Task names containing periods (e.g. `foo.com/tenant/task`) were being corrupted during authentication. The issue: `decode_safe_name()` replaces `.` with `%` then attempts percent-decoding, so `foo.com` becomes `foo%com`, and `%co` is invalid hex, leaving the name corrupted.

The `decode_safe_name` function was added in #1516 to support dot-encoded JSON config objects as usernames (e.g. `{}` encoded for platforms that don't allow `{` in usernames). When materialization task-based auth was introduced in #1665, the dot-encoding was never removed despite no longer having a valid use case: we're no longer trying to do things like stuff JSON objects into usernames.

This PR removes the `decode_safe_name()` call from `authenticate()`. It's still used by `from_downstream_topic_name` for decoding Kafka topic names in order to avoid breaking backwards compatibility.  It also fixes a small bug in Dekaf's Kafka API client (which the e2e test infrastructure uses) that was treating an error code of `-1` as non-terminal.

**NOTE:** This will break any user using `%`-encoded or `.`-encoded SASL usernames (of which there are none at present AFAICT)

### DIFF
--- a/crates/dekaf/src/api_client.rs
+++ b/crates/dekaf/src/api_client.rs
@@ -225,7 +225,7 @@ async fn send_sasl_message(
     )
     .await?;
 
-    if resp.error_code > 0 {
+    if resp.error_code != 0 {
         let err = kafka_protocol::ResponseError::try_from_code(resp.error_code)
             .map(|code| format!("{code:?}"))
             .unwrap_or(format!("Unknown error {}", resp.error_code));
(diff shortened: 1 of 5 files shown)


williamhbaker (approved):
LGTM

PR #2590, opened 2026-01-05
Create storage mapping mutation

GregorShear:
**Description:** 
This PR adds a `createStorageMapping` GraphQL mutation to the control-plane API that allows users to create storage mappings with health checks. Key changes:

- New GraphQL mutation (`createStorageMapping`) that:
  - Validates user has admin access to the catalog prefix
  - Resolves data plane names and checks user authorization for each
  - Runs health checks against each data plane + store combination (with 30s timeout)
  - Only saves the mapping if all health checks pass (unless `dryRun: true`)
- New `build_journal_client` helper in data_plane.rs that creates an authenticated Gazette journal client using HMAC-signed JWT claims
- New `fragment_store_health` method on `gazette::journal::Client` to call the broker's health check API
- HMAC keys plumbing: The decrypted HMAC keys cache is now passed through to the App struct so storage health checks can authenticate with data planes
- Integration test for validation errors (tests requiring actual broker connections are deferred until broker mocking is available)

**Workflow steps:** Call the GraphQL mutation:

```gql
mutation CreateStorageMapping {
  createStorageMapping(input: {
    catalogPrefix: "gregCo/"
    storage: {
      stores: [
        { provider: "GCS", bucket: "your-bucket", prefix: "collection-data/" }
      ]
      data_planes: ["ops/dp/public/local-cluster", "no/such/dataplane"]
    }
    dryRun: true
  }) {
    created
    catalogPrefix
  }
}
```

and expect a response like:
```json
{
  "data": {
    "createStorageMapping": {
      "created": false,
      "catalogPrefix": "gregCo/",
    }
  }
}
```
or failure:
```json
{
  "data": null,
  "errors": [
    {
      "message": "Storage health checks failed",
      "locations": [
        {
          "line": 3,
          "column": 3
        }
      ],
      "path": [
        "createStorageMapping"
      ],
      "extensions": {
        "healthCheckErrors": [
          {
            "dataPlane": "gregCo/",
            "error": "Data plane not found",
            "fragmentStore": "gs://your-bucket/collection-data/"
          },
          {
            "dataPlane": "gregCo/foo/",
            "error": "Data plane not found",
            "fragmentStore": "gs://your-bucket/collection-data/"
          }
        ]
      }
    }
  ]
}
```


crates/control-plane-api/src/server/public/graphql/storage_mappings.rs
```diff
@@ -0,0 +1,276 @@
...
+            }
+        } else {
+            missing.push(name.clone());
+        }
+    }
+
+    Ok(ResolvedDataPlanes { found, missing })
+}
+
+const HEALTH_CHECK_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(30);
```
GregorShear, line 44: is 30 seconds a reasonable timeout here?
@psFried

crates/control-plane-api/src/server/public/graphql/storage_mappings.rs
```diff
@@ -0,0 +1,276 @@
...
+    let fut = client.fragment_store_health(broker::FragmentStoreHealthRequest {
+        fragment_store: fragment_store.clone(),
+    });
+
+    let error = match tokio::time::timeout(HEALTH_CHECK_TIMEOUT, fut).await {
+        Ok(Ok(resp)) if resp.store_health_error.is_empty() => None,
+        Ok(Ok(resp)) => Some(resp.store_health_error),
+        Ok(Err(err)) => Some(err.to_string()),
+        Err(_) => Some("Health check timed out".to_string()),
+    };
```
GregorShear, line 61: is this too dense? nested results?

crates/control-plane-api/src/server/public/graphql/storage_mappings.rs
psFried: I think so, yeah. Seems fine to just bump it later if we need to.

crates/control-plane-api/src/server/public/graphql/storage_mappings.rs
```diff
@@ -0,0 +1,300 @@
...
+            found: data_planes,
+            missing,
+        } = resolve_data_planes(app, claims, &storage.data_planes).await?;
+
+        // Run health checks
+        let mut health_checks =
+            run_storage_health_checks(app, &data_planes, &fragment_stores).await;
+
+        // Add error results for missing data planes
+        for name in &missing {
```
psFried, line 227: I'm thinking that we should not try to support a partial success like this in the case of authorization failures. IMO it's pretty fair to just fail the entire mutation if the user lacks authorization to any of the data planes. And it would allow you to call `app.verify_user_authorization_graphql(..., &storage.data_planes)` up front, and `resolve_data_planes` wouldn't need to return `missing`. I don't think we're likely to run into that type of auth failure, in practice, since our default role grants confer `write` capability to the entire tenant prefix and `read` to all public data planes.

crates/control-plane-api/src/server/public/graphql/storage_mappings.rs
```diff
@@ -0,0 +1,300 @@
...
+}
+
+/// Resolved data planes and any that were not found or unauthorized.
+struct ResolvedDataPlanes {
+    found: Vec<tables::DataPlane>,
+    missing: Vec<String>,
+}
+
+/// Resolve data plane names to data plane records, checking authorization.
+async fn resolve_data_planes(
```
psFried: The `Snapshot` already includes all data planes, and it seems safe for us use that instead of querying the database. (The downside of the snapshot is just that the data there can be a touch stale. But if you create a data plane, then it's certain that the snapshots will have all refreshed before such time as the data plane is running and capable of handling a health check request.)

Also, see my other comment below about the authorization checks

crates/control-plane-api/src/server/public/graphql/storage_mappings.rs
psFried: No, I think this looks pretty good. It helps a lot that the source of the nested results is so close to the match arms.

crates/control-plane-api/src/server/public/graphql/storage_mappings.rs
```diff
@@ -0,0 +1,324 @@
...
+            vec![input.catalog_prefix.to_string()],
+            models::Capability::Admin,
+        )
+        .await?;
+
+        // Check if any existing tasks or collections would be affected by this new storage mapping.
+        // We disallow creating storage mappings that would change the storage for existing specs.
+        let prefix_str = input.catalog_prefix.as_str();
+        let affected_specs: Vec<String> =
+            Snapshot::evaluate(app.snapshot(), chrono::Utc::now(), |snapshot: &Snapshot| {
```
psFried: I'm thinking it's worth it to query the database instead of using the snapshot here. If we use the snapshot, then it might be possible (though admittedly pretty unlikely) for a user to create a task and then immediately create a storage mapping, before the snapshot is refreshed. And given the relative infrequency of storage mapping creation, I'm not concerned about the additional database query. My apologies. I think I'd mentioned using Snapshot, but hadn't considered the possible race condition.

We have [these functions](https://github.com/estuary/flow/blob/a9271106431d6e93c6f3d1b97b64e449b2d4c6f8/crates/control-plane-api/src/server/public/graphql/live_spec_refs.rs#L386) for fetching live spec names, and could move those into `control-plane-api/src/live_specs/db.rs` to make them accessible.

crates/control-plane-api/src/server/public/graphql/storage_mappings.rs
```diff
@@ -0,0 +1,324 @@
...
+        }
+
+        if input.dry_run {
+            return Ok(CreateStorageMappingResult {
+                created: false,
+                catalog_prefix: input.catalog_prefix.to_string(),
+            });
+        }
+
+        let mut txn = app.pg_pool.begin().await?;
```
psFried: optional nit: In this case, the transaction seems unnecessary since we're only performing a single operation. We've generally just passed a `&sqlx::PgPool` parameter to the insert function in cases like this where we don't need a transaction, as it simplifies the code a bit and also serves as a clear indication that we're not doing a bunch of other stuff as part of the same transaction.

crates/control-plane-api/src/server/public/graphql/storage_mappings.rs
```diff
@@ -0,0 +1,324 @@
...
+        let insert_result = crate::directives::storage_mappings::insert_storage_mapping(
+            detail,
+            &input.catalog_prefix,
+            storage,
+            &mut txn,
+        )
+        .await;
+
+        if let Err(sqlx::Error::Database(db_err)) = &insert_result {
+            // 23505 is the PostgreSQL error code for unique_violation
```
psFried: I'll suggest a different approach for this, where it's kept internal to `insert_storage_mapping`. We can use a query like:

```
let new_id = sqlx::query!(
  r#"insert into storage_mappings ...
     on conflict (catalog_prefix) do nothing
     returning id"#,
  ...
)
.fetch_optional(db)
.await
.ok_or_else(|| anyhow::anyhow!("storage mapping already exists");
```

I don't actually have a strong opinion on the error code vs using `returning`, but I feel a bit more strongly that this check belongs inside `insert_storage_mapping`.

crates/flow-client/control-plane-api.graphql
```diff
@@ -262,6 +262,46 @@ type Controller {
...
+	"""
+	success: Boolean!
+	"""
+	The catalog prefix for which the storage mapping was created.
+	"""
+	catalogPrefix: String!
+	"""
+	Results of health checks for each data plane and store combination.
+	"""
+	healthChecks: [StorageHealthResult!]!
```
psFried: Looks like this file needs to be re-generated?

crates/control-plane-api/src/server/public/graphql/storage_mappings.rs
```diff
@@ -0,0 +1,324 @@
...
+                .map(|r| {
+                    serde_json::json!({
+                        "dataPlane": r.data_plane_name,
+                        "fragmentStore": r.fragment_store,
+                        "error": r.error.as_deref().unwrap_or("unknown error"),
+                    })
+                })
+                .collect();
+
+            let errors_value: async_graphql::Value =
```
psFried, line 255: I'd been thinking that we'd return multiple errors here using `Context::add_error`, but now I'm realizing that might not be allowed by the graphql spec, since all the errors would have the same `path`. I'm not seeing a clear line about whether it's permissible to have multiple errors with the same `path`, but it kinda seems like it's not really supported. Does that agree with your understanding? Is that why you used a single error extension here?

crates/control-plane-api/src/server/public/graphql/storage_mappings.rs
GregorShear: oh yeah, i had thought to ask you about that. The existing insert function just expects a transaction - seems easy to change, but it would also require regenerating some query cache files... i wasn't sure if the impact on the diff was worth this small detail. happy to do it tho

GregorShear: Wait, am I dumb? Yes. That insert function is entirely new. I'll change the function signature.

crates/control-plane-api/src/server/public/graphql/storage_mappings.rs
GregorShear: I'm not seeing anything that requires unique error paths...

That said, I'll defend a single error for health check failures as a category, with the individual failures detailed in the extension. From the client's perspective: if I'm handling health check failures, I'd rather inspect one error's extensions than filter/map the entire errors array to extract the relevant details.

crates/control-plane-api/src/server/public/graphql/storage_mappings.rs
GregorShear: I think you're talking about `fetch_live_specs_names_after()` and `fetch_live_specs_names_before()` - but these return paginated results, and i'd have to truncate later results (query will return `philCo/` when i'm only interested in the `gregCo/` tree). 

I added a new `fetch_live_spec_names_by_prefix()` which uses something like `where starts_with(catalog_name, 'gregCo/')`, and doesn't paginate. Let me know if I'm way off base here

crates/control-plane-api/src/server/public/graphql/storage_mappings.rs
GregorShear: i think you're pointing out my leaky abstraction here, and that's been fixed

PR #2597, opened 2026-01-07
Dekaf: E2E tests for group management APIs

jshearer:
This PR adds to the Dekaf e2e test suite to cover consumer group management behavior and its interactions with leader epochs and collection resets.

- Add consumer-group behavior coverage:
  - Commit and resume with a stable `group.id` (rdkafka), including committed-offset readback
  - `Metadata` and `OffsetCommit` error mapping for unknown topics (`UnknownTopicOrPartition`)
  - `OffsetFetch` behavior for never-committed groups (committed_offset = -1, committed_leader_epoch = -1)
- Collection reset behavior:
  - After a collection reset, OffsetFetch does not return offsets committed under the prior epoch

### DIFF
--- a/crates/dekaf/tests/e2e/kafka.rs
+++ b/crates/dekaf/tests/e2e/kafka.rs
@@ -2,7 +2,8 @@ use anyhow::Context;
 use futures::StreamExt;
 use rdkafka::Message;
 use rdkafka::config::RDKafkaLogLevel;
-use rdkafka::consumer::{Consumer, StreamConsumer};
+use rdkafka::consumer::{CommitMode, Consumer, StreamConsumer};
+use rdkafka::topic_partition_list::{Offset, TopicPartitionList};
 use schema_registry_converter::async_impl::avro::AvroDecoder;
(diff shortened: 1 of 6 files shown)


crates/dekaf/tests/e2e/kafka.rs
```diff
@@ -22,13 +23,29 @@ pub struct DecodedRecord {
 
 impl KafkaConsumer {
     pub fn new(broker: &str, registry: &str, username: &str, password: &str) -> Self {
+        Self::with_group_id(
```
jshearer, line 27: This change is duplicated over [here](https://github.com/estuary/flow/blob/4afb88d099fad0268160837e855c0cbf52dc2d90/crates/dekaf/tests/e2e/kafka.rs#L26-L40) in #2625. Shouldn't be a problem, but will probably need to rebase this if that merges first

jshearer: #2625 merged, rebased this 👍

PR #2601, opened 2026-01-08
introduce `tokens` and wire through `gazette` and `flow-client-next`

jgraettinger:
## Summary

This PR introduces auto-refreshing authentication token infrastructure and integrates it throughout the control-plane and data-plane clients.

This is a _partial_ rollout of the new approach. Work remains to update downstream crates and transition them to the patterns of `flow-client-next`.

### New Crates

**`tokens`** - Framework for managing tokens that need periodic refreshing (JWTs, API tokens, OAuth tokens). Handles the lifecycle of obtaining, caching, and refreshing tokens with proper error handling and backoff.

- `Source` trait for producing tokens on demand
- `Watch`/`PendingWatch` for shared access to periodically-refreshed tokens
- Built-in sources: `jwt::SignedSource` (self-signed JWTs), `RestSource` (REST API fetch), `StreamSource` (futures stream wrapper)
- JWT utilities for signing, verification, and parsing with capability-based access control

**`flow-client-next`** - Replacement for `flow-client` that integrates with the tokens crate for automatic token lifecycle management. Provides authorization workflows for:

| Workflow | Use Case |
|----------|----------|
| `UserCollectionAuth` | Users reading/writing collections |
| `UserTaskAuth` | Users accessing task shards and journals |
| `UserPrefixAuth` | Users accessing a catalog prefix within a data-plane |
| `TaskCollectionAuth` | Tasks accessing their bound collections |
| `TaskDekafAuth` | Dekaf tasks fetching specs and schemas |

### Gazette Integration

Refactored gazette journal and shard clients to use `tokens::PendingWatch` for automatic credential refresh:

- Moved `Metadata` from gazette to `proto_grpc` with builder-style API
- Added `new_with_tokens()` constructors for token-based auth
- Changed synchronous `into_sub()` to async `subclient()` that awaits tokens
- Updated callers in flow-client, control-plane-api, flowctl, migrate

This enables clients to automatically refresh expiring credentials without manual intervention.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3223,6 +3223,30 @@ dependencies = [
 ]
 
 [[package]]
+name = "flow-client-next"
+version = "0.0.0"
+dependencies = [
+ "anyhow",
+ "futures",
+ "gazette",
+ "jsonwebtoken",
+ "lazy_static",
+ "models",
+ "postgrest",
+ "proto-flow",
+ "proto-gazette",
+ "proto-grpc",
+ "reqwest 0.12.23",
+ "serde",
+ "serde_json",
+ "tokens",
+ "tokio",
+ "tonic",
+ "tracing",
+ "url",
+]
+
+[[package]]
 name = "flow-web"
 version = "0.6.0"
 dependencies = [
--- a/crates/flow-client-next/Cargo.toml
+++ b/crates/flow-client-next/Cargo.toml
@@ -1,0 +1,30 @@
+[package]
+name = "flow-client-next"
+version.workspace = true
+rust-version.workspace = true
+edition.workspace = true
+authors.workspace = true
+homepage.workspace = true
+repository.workspace = true
+license.workspace = true
+
+[dependencies]
+gazette = { path = "../gazette" }
+models = { path = "../models" }
+proto-flow = { path = "../proto-flow" }
+proto-gazette = { path = "../proto-gazette" }
+proto-grpc = { path = "../proto-grpc" }
+tokens = { path = "../tokens" }
+
+anyhow = { workspace = true }
+futures = { workspace = true }
+jsonwebtoken = { workspace = true }
+lazy_static = { workspace = true }
+postgrest = { workspace = true }
+reqwest = { workspace = true }
+serde = { workspace = true }
+serde_json = { workspace = true }
+tokio = { workspace = true }
+tonic = { workspace = true }
+tracing = { workspace = true }
+url = { workspace = true }
--- a/crates/flow-client-next/README.md
+++ b/crates/flow-client-next/README.md
@@ -1,0 +1,61 @@
+# flow-client-next
+
+Client library for Estuary Flow control-plane APIs with auto-refreshing authentication.
+
+This crate is a replacement for `flow-client`. Dependent crates should use
+`use flow_client_next as flow_client;` to ease the transition.
+
+## Overview
+
+Provides HTTP clients and authentication workflows for interacting with:
+- Control-plane REST APIs (agent endpoints)
+- PostgREST/Supabase APIs
+- Data-plane services (brokers, reactors) via authorized tokens
+
+All authentication sources integrate with the `tokens` crate for automatic
+token refresh, eliminating manual token lifecycle management.
+
+## Modules
+
+- `rest` - Generic REST client wrapping reqwest
+- `postgrest` - PostgREST query execution with pagination support
+- `user_auth` - User token management (access/refresh token pairs)
+- `workflows` - Authorization source implementations
+
+## Authorization Workflows
+
+The `workflows` module provides `tokens::Source` implementations for different
+authorization scenarios:
+
+| Workflow | Use Case |
+|----------|----------|
+| `UserCollectionAuth` | Users reading/writing collections |
+| `UserTaskAuth` | Users accessing task shards and journals |
+| `UserPrefixAuth` | Users accessing a catalog prefix within a data-plane |
+| `TaskCollectionAuth` | Tasks accessing their bound collections |
+| `TaskDekafAuth` | Dekaf tasks fetching specs and schemas |
+
+Each workflow handles token refresh automatically. For task-based workflows,
+use the `new_signed_source()` helper to construct the JWT claims with the
+appropriate data-plane signing key.
+
+## Building Gazette Clients
+
+Each user workflow module exports `new_journal_client()` and (where applicable)
+`new_shard_client()` helpers that wire a `tokens::PendingWatch` into a
+Gazette client, providing automatic token injection on each request.
+
+```rust
+let tokens = tokens::spawn(UserCollectionAuth { ... });
+let client = user_collection_auth::new_journal_client(
+    fragment_client,
+    router,
+    tokens.pending_watch(),
+);
+```
+
+## Stream Adapters
+
+`adapt_gazette_retry_stream()` converts `gazette::RetryResult` streams into
+`tonic::Result` streams, handling transient error suppression with caller-provided
+logic.
--- a/crates/flow-client-next/src/lib.rs
+++ b/crates/flow-client-next/src/lib.rs
@@ -1,0 +1,63 @@
+use futures::StreamExt;
+
+pub mod postgrest;
+pub mod rest;
+pub mod user_auth;
+pub mod workflows;
+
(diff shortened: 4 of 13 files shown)


crates/tokens/src/jwt.rs
```diff
@@ -0,0 +1,402 @@
+use crate::{DateTime, Source, TimeDelta};
+use base64::Engine;
+
+/// EncodingKey is a type alias for jsonwebtoken::EncodingKey.
+/// Prefer it over jsonwebtoken::EncodingKey.
+pub type EncodingKey = jsonwebtoken::EncodingKey;
```
jshearer, line 6: Maybe it's obvious, but why this instead of just `use jsonwebtoken::EncodingKey`?

crates/tokens/src/jwt.rs
```diff
@@ -0,0 +1,402 @@
...
+
+/// DecodingKey is a type alias for jsonwebtoken::DecodingKey.
+/// Prefer it over jsonwebtoken::DecodingKey.
+pub type DecodingKey = jsonwebtoken::DecodingKey;
+
+/// Sign claims to produce a JWT token.
```
jshearer, line 12: nit: json web token token 🤓

crates/tokens/src/jwt.rs
```diff
@@ -0,0 +1,402 @@
...
+pub fn sign<Claims>(claims: Claims, key: &EncodingKey) -> Result<String, tonic::Status>
+where
+    Claims: serde::Serialize,
+{
+    // Select an appropriate algorithm and header for the key.
+    // NOTE: AlgorithmFamily is not exported, so we cannot match over it
```
jshearer, line 18: I'm not clear what this comment is saying. What does this mean?
> `algorithms().first()` has hard-coded behavior

crates/tokens/src/jwt.rs
```diff
@@ -0,0 +1,402 @@
...
+            #[serde(default)]
+            cap: Option<u32>,
+        }
+        let skim: ClaimsSkim = serde_json::from_str(claims.get()).map_err(|err| {
+            tonic::Status::unauthenticated(format!(
+                "failed to skim `cap` claim from verified JWT: {err}"
+            ))
+        })?;
+
+        if skim.cap.unwrap_or_default() & require_capability != require_capability {
```
jshearer, line 111: `require_capability != require_capability`? 

**Edit:**
Ohh.. this a bitwise thing isn't it, read as `(cap & require_capability) != require_capability`?

crates/tokens/src/lib.rs
```diff
@@ -0,0 +1,181 @@
...
+    #[cfg(debug_assertions)]
+    {
+        // TIME_POINT is a (Instant, DateTime) pair captured at the same time.
+        // It allows for mapping between Instant and DateTime.
+        //
+        // Beware! There's potential for clock drift over time due to NTP steps
+        // or because of subtle drifts between the two clocks. This is acceptable
+        // in tests (only).
+        static TIME_POINT: std::sync::LazyLock<(std::time::Instant, DateTime)> =
+            std::sync::LazyLock::new(|| (std::time::Instant::now(), chrono::Utc::now()));
```
jshearer, line 167: These two operations will run at different times, and so will have slightly different values internally... but that's not a problem because the difference will be on the order of micro/nanoseconds, whereas the tests that matter for e.g expiration are operating on the order of minutes/hours, right?

crates/tokens/src/lib.rs
```diff
@@ -0,0 +1,181 @@
...
+///
+/// The timestamps returned by this routine are aware of tokio test time,
+/// and will return coherent values in paused testing contexts where
+/// tokio::time::advance() or auto-advance is used.
+///
+/// Use this routine over SystemTime or jsonwebtoken::get_current_timestamp().
+pub fn now() -> DateTime {
+    // In debug builds, use a fixed point to map between Instant and DateTime,
+    // allowing tokio test time to influence the result.
+    #[cfg(debug_assertions)]
```
jshearer, line 158: I frequently run debug builds locally during development. Could this possibly introduce some subtle behavior differences vs prod that would be difficult to track down? Is there any way to gate this behavior to only run during tests, instead of all debug builds?

crates/tokens/src/rest.rs
```diff
@@ -0,0 +1,118 @@
...
+/// RestSource is a trait fulfilling Source via REST API requests and responses.
+pub trait RestSource: Send + Sync {
+    /// Model is the deserialized response type from the REST API.
+    type Model: for<'de> serde::Deserialize<'de> + Send + Sync + 'static;
+    /// Token type extracted from Model. Often the same type as Model.
+    type Token: Send + Sync + 'static;
+
+    /// Build an API request whose 200 OK response is a JSON serialization of Model.
+    /// A server-side (5XX) error is logged and retried but is not client-facing.
+    /// All other error statuses are mapped and surfaced as tonic::Status.
```
jshearer, line 12: Cool, so this codifies the behavior of "500 to retry, anything else is terminal". 

I have [this branch](https://github.com/estuary/flow/commit/4be3e802a750cc1c6a467da2f740e6ef2d918071) (which I haven't made a PR for yet) to update `/authorize/dekaf` to return a proper 404 when the task isn't found, rather than returning 500. I... wonder if that's actually the correct behavior? On the one hand, it'd be nice to know promptly that a task doesn't exist vs retrying forever, but OTOH it could have been recently created and not included in the current snapshot. WDYT?

crates/tokens/src/rest.rs
```diff
@@ -0,0 +1,118 @@
...
+    type Revoke = std::future::Pending<()>;
+
+    async fn refresh(
+        &mut self,
+        started: DateTime,
+    ) -> tonic::Result<Result<(Self::Token, TimeDelta, Self::Revoke), TimeDelta>> {
+        let request = self.build_request(started).await?;
+
+        // If we need to retry, it will be in range 5-15 seconds (mean of 10s).
+        let retry = TimeDelta::milliseconds(rand::random_range(5000..15000));
```
jshearer: Probably not actionable, but noting that this sort of thing is spread throughout the codebase in various forms (jitter, exponential backoff, etc). Might be nice at some point to centralize it?

crates/tokens/src/jwt.rs
jgraettinger: To allow EncodingKey to one-day _not_ be `jsonwebtoken`. This crate is ... fine, but I think we'll want to directly implement more of its functionality over time.

crates/tokens/src/jwt.rs
jgraettinger: algorithms() has a hard-coded iteration order. This is a work-around for jsonwebtoken's oversight of not exporting AlgorithmFamily.

crates/tokens/src/lib.rs
jgraettinger: there's no direct comparison between a SystemTime and an Instant so "internally different values" doesn't have meaning -- they're physically different hardware clocks. What this is establishing is an arbitrary, close-enough timepoint where we consider the values to be equal. Both SystemTime and Instant support relative deltas (as Duration), and having this fixed timepoint lets us use those deltas to map between Instant and SystemTime.

The concern discussed in the comment above isn't that SystemTime and Instant are "off" from each other at the time of establishing this mapping -- it's that the _relative progression of time_ of SystemTime and Instant isn't guaranteed to align over long periods of time. The hardware clocks can run at different speeds.

That doesn't matter for a test because they don't run for that long, so the accumulated skew over time is small. What we're achieving here is that use of `tokio::time::advance()` will show up in tokens::now() in a logically consistent way.

crates/tokens/src/lib.rs
jgraettinger: No, there isn't a way to distinguish, because this crate is built as a library (!cfg(test)) when being included in some _other_ crate's test build. Ergo using debug.

It only matters if you're leaving a debug build running for very long periods of time (weeks or months), as skew can accumulate between the different physical hardware clocks of Instant vs SystemTime.

crates/tokens/src/rest.rs
jgraettinger: Yep, 500 is automatically retried, anything else is terminal.

I have a pretty substantial overhaul of control-plane-api I'm working on that happens to introduce more appropriate errors (404 in this case).

PR #2605, opened 2026-01-09
Update storage mapping mutation

GregorShear:
**Description**: Adds `updateStorageMapping` GraphQL mutation for modifying existing storage mappings. Refactors shared validation and health check logic into reusable helper functions. The update response includes a `republish` field indicating whether the primary storage bucket changed, which signals that affected specs will need republishing. 

**Workflow steps**: Call `updateStorageMapping` with the same input shape as `createStorageMapping`:

```gql
mutation {
  updateStorageMapping(
    catalogPrefix: "gregCo/"
    dryRun: false
    storage: {
      stores: [
        {provider: "GCS", bucket: "your-bucket", prefix: "collection-data/"}
      ]
      data_planes: ["ops/dp/public/local-cluster"]
    }
  ) {
    updated
    catalogPrefix
    republish
  }
}
```

Use `dryRun: true` to validate input and check if a republish would be required without persisting changes. 

**Notes for reviewers:**

- The `republish` field compares the incoming list of stores with the existing
- The TODO comment about actually triggering republish remains - this PR only reports whether it's needed

### DIFF
--- a/crates/control-plane-api/src/server/public/graphql/storage_mappings.rs
+++ b/crates/control-plane-api/src/server/public/graphql/storage_mappings.rs
@@ -172,7 +172,7 @@ impl StorageMappingsMutation {
         catalog_prefix: models::Prefix,
         detail: Option<String>,
         storage: async_graphql::Json<models::StorageDef>,
-        dry_run: bool,
+        #[graphql(default = false)] dry_run: bool,
     ) -> async_graphql::Result<CreateStorageMappingResult> {
         let env = ctx.data::<crate::Envelope>()?;
         let claims = env.claims()?;
@@ -286,7 +286,7 @@ impl StorageMappingsMutation {
         catalog_prefix: models::Prefix,
         detail: Option<String>,
         storage: async_graphql::Json<models::StorageDef>,
-        dry_run: bool,
+        #[graphql(default = false)] dry_run: bool,
     ) -> async_graphql::Result<UpdateStorageMappingResult> {
         let env = ctx.data::<crate::Envelope>()?;
         let claims = env.claims()?;
--- a/crates/flow-client/control-plane-api.graphql
+++ b/crates/flow-client/control-plane-api.graphql
@@ -560,7 +560,7 @@ type MutationRoot {
 	
 	All health checks must pass before the storage mapping is created.
 	"""
-	createStorageMapping(catalogPrefix: Prefix!, detail: String, storage: JSON!, dryRun: Boolean!): CreateStorageMappingResult!
+	createStorageMapping(catalogPrefix: Prefix!, detail: String, storage: JSON!, dryRun: Boolean! = false): CreateStorageMappingResult!
 	"""
 	Update an existing storage mapping for the given catalog prefix.
 
@@ -570,7 +570,7 @@ type MutationRoot {
 
 	All health checks must pass before the storage mapping is updated.
 	"""
-	updateStorageMapping(catalogPrefix: Prefix!, detail: String, storage: JSON!, dryRun: Boolean!): UpdateStorageMappingResult!
+	updateStorageMapping(catalogPrefix: Prefix!, detail: String, storage: JSON!, dryRun: Boolean! = false): UpdateStorageMappingResult!
 	"""
 	Creates a new alert subscription. Returns an error if there is already
 	an existing subscription for the same prefix and email address.


crates/control-plane-api/src/server/public/graphql/storage_mappings.rs
```diff
@@ -76,6 +76,17 @@ pub struct CreateStorageMappingResult {
...
 }
 
+/// Result of updating a storage mapping.
+#[derive(Debug, Clone, async_graphql::SimpleObject)]
+pub struct UpdateStorageMappingResult {
+    /// Whether the storage mapping was updated (false if dry_run was true).
+    pub updated: bool,
+    /// The catalog prefix of the storage mapping that was updated.
+    pub catalog_prefix: String,
+    /// Whether a republish is required because the primary storage bucket changed.
```
psFried: We'd need to republish if _any_ of the storage buckets have changed, not just the primary one. The first is the only one that's used for writing data, but the others are all still used for reading historical data, so we'd want to activate any changes to those, too.

crates/control-plane-api/src/server/public/graphql/storage_mappings.rs
```diff
@@ -139,33 +275,20 @@ impl StorageMappingsMutation {
     pub async fn create_storage_mapping(
         &self,
         ctx: &Context<'_>,
-        input: CreateStorageMappingInput,
+        input: StorageMappingInput,
```
psFried: I don't love the name `input`, because it feels too generic. Just sort of casting about for other options, we could call these parameters `new` and `update` respectively for the create and update functions. Or maybe it's better to use separate parameters instead, since the grouping doesn't seem to be giving us a whole lot? I'll defer to your experience here, but just wanted to raise the question, so that we can hopefully agree on a pattern that we can apply consistenly across the graphql api. WDYT?

crates/control-plane-api/src/server/public/graphql/storage_mappings.rs
```diff
@@ -299,4 +343,101 @@ impl StorageMappingsMutation {
...
+
+        if input.dry_run {
+            return Ok(UpdateStorageMappingResult {
+                updated: false,
+                catalog_prefix: input.catalog_prefix.to_string(),
+                republish,
+            });
+        }
+
+        let detail = input.detail.as_deref().unwrap_or("updated via GraphQL API");
```
psFried: I think if we're going to generate a detail automatically, we should try to make it as clear and helpful as we can, since these messages are shown to users. I'll suggest something like `Updated by {user_email_or_id}`. But I think leaving it null would be OK, too, and would effectively convey the same information (that someone updated it via api and didn't specify a detail)

crates/control-plane-api/src/server/public/graphql/storage_mappings.rs
```diff
@@ -299,4 +343,101 @@ impl StorageMappingsMutation {
...
+
+        tracing::info!(
+            catalog_prefix = %input.catalog_prefix,
+            data_planes = ?validated.storage.data_planes,
+            stores_count = validated.storage.stores.len(),
+            republish,
+            "updated storage mapping"
+        );
+
+        // TODO: if the primary (first) fragmentStore changed, republish the entire prefix.
```
psFried: nit: seems like the plan is for the client to handle the republication, so this can be removed?

crates/control-plane-api/src/server/public/graphql/storage_mappings.rs
```diff
@@ -299,4 +343,101 @@ impl StorageMappingsMutation {
...
+
+        let existing_stores: std::collections::HashSet<String> = existing_storage
+            .stores
+            .iter()
+            .map(|s| s.to_url(&input.catalog_prefix).to_string())
+            .collect();
+        let new_stores: std::collections::HashSet<String> =
+            validated.fragment_stores.iter().cloned().collect();
+
+        let primary_changed = validated.fragment_stores.first()
```
psFried: Unless I'm missing something, the `existing_stores != new_stores` check should be entirely sufficient to determine the need to republish. I don't see why we'd need to also check anything related to the primary store.

GregorShear: This check ignores the ordering of secondary stores - am i overthinking this?

GregorShear: Simplified this to just compare the existing list with the input

crates/control-plane-api/src/server/public/graphql/storage_mappings.rs
GregorShear: Split these out into named params

crates/control-plane-api/src/server/public/graphql/storage_mappings.rs
```diff
@@ -77,68 +118,211 @@ async fn run_storage_health_checks(
...
+            SELECT catalog_name
+            FROM live_specs
+            WHERE starts_with(catalog_name, $1)
+            AND spec IS NOT NULL
+            LIMIT 5
+            "#,
+            &catalog_prefix,
+        )
+        .fetch_all(&mut *txn)
+        .await?;
```
GregorShear, line 215: @psFried 
okay to leave these inline or would you rather keep all queries in `crates/control-plane-api/src/directives/storage_mappings.rs`?

crates/control-plane-api/src/server/public/graphql/storage_mappings.rs
```diff
@@ -246,50 +401,54 @@ impl StorageMappingsMutation {
...
+                republish,
+            });
         }
 
-        Ok(UpsertStorageMappingResult {
-            created: current.is_none(),
+        sqlx::query!(
+            r#"
+            UPDATE storage_mappings
+            SET spec = $2, detail = $3, updated_at = now()
```
GregorShear: should we be manually updating the `updated_at` value? i would expect this to be handled with an UPDATE trigger...

crates/control-plane-api/src/server/public/graphql/storage_mappings.rs
```diff
@@ -246,50 +401,54 @@ impl StorageMappingsMutation {
...
+            WHERE catalog_prefix = 'recovery/' || $1
+            "#,
+            &catalog_prefix as &str,
+            crate::TextJson(&recovery_storage) as crate::TextJson<&models::StorageDef>,
+            detail,
+        )
+        .execute(&mut *txn)
+        .await?;
```
GregorShear: I think this `recovery/` logic is right but would love some extra scrutiny here

crates/control-plane-api/src/server/public/graphql/storage_mappings.rs
GregorShear: does this gql operation need to be aware of the republishing process, or is that just handled elsewhere when changes are detected?

crates/control-plane-api/src/server/public/graphql/storage_mappings.rs
psFried: Yeah, we just handle those manually instead of with triggers. It's not something I feel especially strongly about, and we do still use triggers here and there for other things. Though I think my general position is to try to avoid triggers when we can, to make things more explicit.

That said, I think the [existing upsert function](https://github.com/estuary/flow/blob/71da382b2c713e42bca3b141ceae3ed9af0c920b/crates/control-plane-api/src/directives/storage_mappings.rs#L20) is probably what we should use here instead of inline sql, and that already takes care of setting `updated_at`.

crates/control-plane-api/src/server/public/graphql/storage_mappings.rs
psFried: I think inline is generally fine, as long as it's not something we're trying to re-use elsewhere. For this particular query, inline seems fine. But we have some existing functions in `crates/control-plane-api/src/directives/storage_mappings.rs` that seem like they'd work for inserts and updates to storage mappings. I think it's worth using those, unless there's a reason not to.

crates/control-plane-api/src/server/public/graphql/storage_mappings.rs
```diff
@@ -77,68 +118,211 @@ async fn run_storage_health_checks(
...
+        } = storage;
+
+        let collection_storage = models::StorageDef {
+            data_planes,
+            stores: stores
+                .iter()
+                .cloned()
+                .map(|mut store| {
+                    let prefix = store.prefix_mut();
+                    *prefix = models::Prefix::new(format!("{prefix}collection-data/"));
```
psFried: I foresee potential issues with this logic in update scenarios. The UI will fetch the existing storage mappings, which already have the `collection-data/` prefix, and then pass those storage mappings to update, which would add the prefix again. So I'm thinking at a minimum we'd want to change the corresponding logic that's used during the update to check whether the prefix is there already. But also, this feels like it ought to be in a function that's shared between insert and update operations. Ideally, we'd have a pure function that accepts a single `models::StorageDef`, and returns a tuple of both the collection and recovery storage defs. Then we could continue to use that same function, even after we no longer persist the recovery mappings in postgres.

crates/control-plane-api/src/server/public/graphql/storage_mappings.rs
psFried: See my other comments about factoring out a function that returns both the storage defs from the original input, and about reusing the existing functions for updating the database. I'm thinking we should use [upsert_storage_mapping](https://github.com/estuary/flow/blob/71da382b2c713e42bca3b141ceae3ed9af0c920b/crates/control-plane-api/src/directives/storage_mappings.rs#L20), and then this inline query would not be necessary.

crates/control-plane-api/src/directives/storage_mappings.rs
```diff
@@ -64,6 +64,32 @@ where
     Ok(result.rows_affected() > 0)
 }
 
+pub async fn update_storage_mapping<'e, T, E>(
+    detail: Option<&str>,
+    catalog_prefix: &str,
+    spec: T,
+    executor: E,
+) -> sqlx::Result<bool>
+where
```
GregorShear, line 73: Same rationale as the GraphQL mutations - separate create / update functions will generally reduce complexity. Keeping upsert for now to minimize the diff and it's convenient for the `recovery/` mappings

crates/control-plane-api/src/directives/storage_mappings.rs
```diff
@@ -89,3 +115,153 @@ pub async fn fetch_storage_mappings(
...
+const COLLECTION_DATA_SUFFIX: &str = "collection-data/";
+
+/// Split a user-provided `StorageDef` into separate collection and recovery storage definitions.
+///
+/// The collection storage gets `collection-data/` appended to each store's prefix (if not already
+/// present) and retains the data plane assignments. The recovery storage uses the base prefixes
+/// (with `collection-data/` stripped if present) and has no data plane assignments.
+pub fn split_collection_and_recovery_storage(
+    storage: models::StorageDef,
+) -> (models::StorageDef, models::StorageDef) {
```
GregorShear, line 128: Is this a good place for this util function to live? and related tests below?

crates/control-plane-api/src/server/public/graphql/storage_mappings.rs
```diff
@@ -152,37 +309,13 @@ impl StorageMappingsMutation {
...
-                    ext.set("healthCheckErrors", errors_value.clone());
-                }),
-            );
-        }
-
-        // Begin a transaction to check for conflicts and upsert the storage mapping.
+        // Begin a transaction to fetch existing mapping and update.
         let mut txn = env.pg_pool.begin().await?;
 
+        // Fetch existing storage mapping to compare stores and verify it exists.
```
GregorShear, line 316: the query just below here looks similar to the existing `fetch` function in the shared file, except here we're locking the row with `FOR UPDATE` - i'm inclined to keep this query defined here until we find another place we need the row lock

crates/control-plane-api/src/server/public/graphql/storage_mappings.rs
```diff
@@ -77,68 +121,181 @@ async fn run_storage_health_checks(
...
+
+        // using upsert here to simplify recovery mapping update/insert
+        // which we'll eventually remove when we stop storing recovery mappings separately
+        upsert_storage_mapping(
+            detail.as_deref(),
+            &format!("recovery/{catalog_prefix}"),
+            &recovery_storage,
+            &mut txn,
+        )
+        .await?;
```
GregorShear, line 259: i mention this elsewhere but i'm keeping the upsert function here because it does feel convenient for the recovery case. I'd make a case for removing upsert entirely when the recovery/ mappings are gone

crates/control-plane-api/src/directives/storage_mappings.rs
```diff
@@ -89,3 +115,153 @@ pub async fn fetch_storage_mappings(
...
+    let collection_storage = models::StorageDef {
+        data_planes,
+        stores: stores
+            .iter()
+            .cloned()
+            .map(|mut store| {
+                let prefix = store.prefix_mut();
+                if !prefix.as_str().ends_with(COLLECTION_DATA_SUFFIX) {
+                    *prefix = models::Prefix::new(format!("{prefix}{COLLECTION_DATA_SUFFIX}"));
+                }
```
GregorShear, line 143: Just to talk this through - the `collection-data/` part may have already been added in a previous operation, so we want to check if it's there first

crates/control-plane-api/src/directives/storage_mappings.rs
```diff
@@ -89,3 +115,153 @@ pub async fn fetch_storage_mappings(
...
+            .collect(),
+    };
+
+    let recovery_storage = models::StorageDef {
+        data_planes: Vec::new(),
+        stores: stores
+            .into_iter()
+            .map(|mut store| {
+                let prefix = store.prefix_mut();
+                if let Some(base) = prefix.as_str().strip_suffix(COLLECTION_DATA_SUFFIX) {
```
GregorShear, line 155: similar to above - the `collection-data/` part may exist from a previous operation, so we remove it for the `recovery/` mapping

GregorShear:
@psFried took another pass here - i believe i've incorporated all of your feedback. I'm making a case to (eventually) deprecate the existing upsert function and move toward separate create/update functions in directives/storage_mappings.rs for the same reason we gave for the gql mutations. Happy to chat through this if you disagree

crates/control-plane-api/src/server/public/graphql/storage_mappings.rs
psFried: Sounds good to me 👍

crates/control-plane-api/src/server/public/graphql/storage_mappings.rs
psFried: Yeah, this seems good to me

crates/control-plane-api/src/directives/storage_mappings.rs
psFried: 👍 yep, sounds good to me

crates/control-plane-api/src/directives/storage_mappings.rs
psFried: Yeah, I think so. We'll probably eventually move this whole file out from `directives`, but I think it's the best place for this logic.

crates/control-plane-api/src/directives/storage_mappings.rs
psFried: I'm realizing that maybe it's not actually great for us to return the storage mappings with the `/collection-data/` added... we'd presumably display `/user-prefix/collection-data/` on the storage mappings UI, which seems like it'd be confusing to users if we then went and _also_ wrote data to `/user-prefix/recovery/`.

We started out with the idea that we'd want to only expose a single representation of the storage mapping to users, and hide the split behind our API. I still think that's a good idea, but I think I was wrong about my previous suggestion to only conditionally add the `collection-data/`. Technically, it works, but it doesn't feel right.

I think it might be better if we instead strip the `collection-data/` when we return the storage mappings in the API response? Then the UI would only ever see `/user-prefix/`, and we'd automatically add the `collection-data/` when we persist changes. Does that make sense to you?

jshearer:
@GregorShear test comment please ignore

PR #2612, opened 2026-01-14
compact DB migrations and begin to support `sqlx::test`

jgraettinger:
**Description:**

Compact all migrations into 01_compacted.sql (produced via pg_dump) and
update 00_polyfill.sql to support both Supabase and sqlx::test environments.
    
For sqlx::test, the polyfill creates stubs for Supabase-provided schemas:
    - auth.users table with required columns
    - auth.uid() function
    
A new, temporary hello-world test is added to verify that sqlx::test is able
to apply all migrations, alongside supabase.
    
The compacted migration comments out pg_dump's search_path reset since all
objects use explicit schema qualification, and an empty search_path breaks
sqlx's migration tracking.

Testing: pgTAP and sanity checked with a local stack.


jgraettinger:
Let's VC -- I've been experimenting with sqlx::test for control-plane-api with positive results, and wanted to land this separately just to reduce the scope.

PR #2618, opened 2026-01-15
control-plane-api: refactor towards tokens::Watch Snapshot and AuthZ retries via 307

jgraettinger:
**Description:**

This PR has two large architectural changes, and a lot of other fallout:

* We invert the way in which the Snapshot is examined: rather than obtaining a lock, we now use a tokens::watch to broadcast immutable Snapshot instances which are fully owned by their holders (via Arc).
* We introduce a standard, semi-transparent AuthZ retry mechanism via 307 Temporary Redirect responses, which allows us to hold the Snapshot steady across request handling

Further fallout:

* Introduces `Envelope` axum extractor that bundles common request context (verified claims, snapshot refresh, timing parameters, pg pool) for cleaner authorization handling
* Refactors agent to use snapshot watch for data-plane lookups instead of direct database queries
* Adds bijective gRPC <=> HTTP status code mapping for consistent API error responses
  * Use tonic::Status as a fundamental error type in both gRPC and HTTP contexts
* Consolidates SOPS age key configuration for tests and local development

Additionally, this PR introduces use of `sqlx::test` for DB-in-the-loop tests. They're fast to run, and they can be run in parallel.

**Notes for reviewers:**

Please review commit by commit. Apologies for the size -- I tried several times to limit scope, but this was a case of pulling on one thread and the whole ball detangling. I'm still in the process of testing, but wanted to get this in front of you both given the impact on your own work.

### DIFF
--- a/crates/control-plane-api/src/envelope.rs
+++ b/crates/control-plane-api/src/envelope.rs
@@ -225,11 +225,19 @@ impl axum::extract::FromRequestParts<Arc<crate::App>> for Envelope {
                         token = exchanged_token.as_ref().unwrap();
                     }
 
-                    let verified = tokens::jwt::verify(
+                    let verified = tokens::jwt::verify::<crate::ControlClaims>(
                         token.as_bytes(),
                         0,
                         &state.control_plane_jwt_decode_keys,
                     )?;
+
+                    if verified.claims().aud != "authenticated" {
+                        return Err(tonic::Status::unauthenticated(
+                            "authorization bearer claims missing required `aud` of 'authenticated'",
+                        )
+                        .into());
+                    }
+
                     MaybeControlClaims::with_verified(verified)
                 }
                 None => MaybeControlClaims::with_unauthenticated(),


jgraettinger:
> I'm still in the process of testing

I've now done a good bit of testing on a local stack, integrated with the UI and `flowctl`. Everything checks out: both the UI and current flowctl are happy to follow 307 redirects.

Further, if I edit flowctl to not send `started_unix` (which triggers 307 retries instead of custom responses), it will follow redirects and perform retries as directed, implying that we could remove the `started_unix` / custom retry response handling without impacting ecosystem installs of `flowctl`.

The exception is the authorize_task handler, as Go's http.Client needs a bit more help to follow redirects.

crates/control-plane-api/src/envelope.rs
```diff
@@ -0,0 +1,252 @@
...
+                "This is an authenticated API but the request is missing a required Authorization: Bearer token",
+            )),
+        }
+    }
+}
+
+/// Envelope packages common fields and request-derived parameters which are
+/// universal across the Estuary API.
+#[derive(Debug)]
+pub struct Envelope {
```
psFried, line 30: I'm not sure whether I'm just missing something, but this structure doesn't seem to be providing much benefit over having separate axum extractors. IMO it would be more clear to have, for example, `handle_post(claims: AuthClaims, snapshot: SnapshotRef, db: sqlx::PgPool)`, where `AuthClaims` could bundle a non-null `ControlClaims`, `retry_after`, and `started`. I think it's less clear to have handlers all get these fields if they don't need them.

A bit of a separate topic, but I'm also not really convinced that it benefits us to move the authentication requirements into the handler function bodies. Having handler functions accept something like a `ControlClaims` seems to like a better way to express "this requires an authenticated user", versus just doing a check inside the function body.

crates/control-plane-api/src/interval.rs
```diff
@@ -0,0 +1,52 @@
+/// Wraps a `chrono::TimeDelta` to allow it to be used as a Postgres `interval`
+/// type. This is necessary because `chrono::TimeDelta` does not implement
+/// `Decode`. Note that converting a `chrono::TimeDelta` to an `interval` may
```
psFried, line 3: Where's this conversion being done? Where would we bubble up those errors?

crates/control-plane-api/src/lib.rs
```diff
@@ -1,35 +1,60 @@
...
-
-// Re-export from the old agent-sql crate
 mod text_json;
+
+/// TextJson encodes JSON for Postgres while preserving property ordering.
 pub use text_json::TextJson;
 
-pub use models::{Capability, CatalogType, Id};
-pub use tables::RoleGrant;
+/// TODO(johnny): Could we use sqlx's native PgInterval type?
```
psFried, line 29: can we? I'm still reading, so not sure where this type is being used, but it seems like you probably wouldn't have created the separate type if PgInterval worked

crates/flow-client/control-plane-api.graphql
```diff
@@ -539,15 +508,15 @@ type LockFailure {
...
 	
 	This validates that the user has admin access to the catalog prefix,
 	runs health checks to verify that data planes can access the storage buckets,
-	and then saves the storage mapping to the database.
+	and then saves the storage mapping to the database if `dry_run` is false.
 	
 	All health checks must pass before the storage mapping is created.
 	"""
-	createStorageMapping(input: CreateStorageMappingInput!): CreateStorageMappingResult!
+	upsertStorageMapping(catalogPrefix: Prefix!, detail: String, storage: JSON!, dryRun: Boolean!): UpsertStorageMappingResult!
```
psFried, line 519: Let's follow up on VC about upserts vs separate create/update operations?

crates/models/src/authorizations.rs
```diff
@@ -7,6 +7,8 @@ pub struct ControlClaims {
     // Note that many more fields, such as additional user metadata,
     // are available if we choose to parse them.
 
+    // Audience for which the token is intended.
+    pub aud: String,
```
psFried, line 11: Looks like this is just refactoring the previous approach of calling `set_audience`, but I'm curious about what part of the system is actually using this. Dekaf?

crates/control-plane-api/src/envelope.rs
jgraettinger: Discussed over VC -- recapping:

* Envelope defines some behaviors that need a lot of this stuff. It's a choice of either having a single containing struct implementing FromRequestParts, or making the handler responsible for collecting all of it and plumbing it into routines like `authorization_outcome`.

* MaybeControlClaims could be refactored to impl FromRequestParts itself. Then, the graphQL handler could add it to the GQL context as a separate piece of data. I'm on the fence myself on whether this is worth it, as 95% of handlers want _both_ the Envelope and ControlClaims, but :shrug:

* The deferred assertion of AuthN offered by MaybeControlClaims is nice for GraphQL APIs, as it means the single axum handler can handle both AuthN and un-AuthN routes, or that assertion of claims can be dynamic based on arguments, or an un-AuthN API can layer in additional data if the user happens to be AuthN (thinking about connectors here). It also greatly simplifies construction of the axum router, which is a little dense (previously, toggling a route from being un-AuthN => AuthN would be an undertaking).

* sqlx::PgPool also doesn't _need_ to be on Envelope. This too is a balance of "if 90% of handlers want it, is it worth having it threaded separately and pushing that onto every handler?"

crates/control-plane-api/src/interval.rs
jgraettinger: [Original source of this comment ](https://github.com/estuary/flow/commit/b19f1e1b97b23593d9fb745c253c8bf54018653c#diff-1b198bfe67f43d43b9426f75a5eff9d4332f33f3bc84a7371de4dc50cb218a82) :grin: 

I just moved it from lib.rs, and I switched an impl Into => impl From to fix a clippy

crates/control-plane-api/src/lib.rs
jgraettinger: Yep as per above, I just moved `Interval` and decided to not pick further at that thread in this PR.
It's used (only) in an interaction with `connector_tags` IIRC.

crates/models/src/authorizations.rs
jgraettinger: Ah, good note. I'd intended to verify this in envelope.rs, but forgot to do so (adding it now).

The broader back-story is that we use to have a jsonwebtoken::Validator that was specifically configured for validating control-plane claims, and would verify `aud` is "authenticated". I wanted to be able to use `tokens::jwt::verify()` for consistency, but that required making `aud` a higher order check (as data-plane tokens don't set `aud`), so the compromise was to have an extra check in `envelope.rs` when verifying control-plane claims.

crates/control-plane-api/src/interval.rs
psFried: lol solid gold 🤣

PR #2625, opened 2026-01-20
dekaf: Migration E2E testing

jshearer:
This PR adds test coverage for Dekaf's handling of cross-dataplane migrations. 

Previously, the convention was that Dekaf addresses live at `dekaf.{plane_fqdn}:9092`. In order to test migration locally, I had to manually and temporarily change this logic. In order to support automated testing of migrations, we needed a first-class way to represent the address of a data-plane's Dekaf instance. We already have `reactor_address` and `broker_address` in the `data_planes` table, so I added `dekaf_address` and `dekaf_registry_address`. 

One wrinkle is that unlike reactors and brokers, data planes do not always run Dekaf. In order to handle this properly, I updated data-plane-controller to actively manage the `dekaf_address` and `dekaf_registry_address` columns, such that they're only set if Dekaf is actually deployed in that plane, otherwise they're null. Dekaf will now emit a helpful error message when attempting to serve a redirect to a data-plane that does not have an instance deployed.

**Note:** As it turns out, our mapping of redirects onto the Kafka protocol isn't perfect. Specifically, since we reuse the same broker ID for the redirect target:
https://github.com/estuary/flow/blob/71da382b2c713e42bca3b141ceae3ed9af0c920b/crates/dekaf/src/session.rs#L249-L262
librdkafka gets confused and shuts down its session rather than smoothly connecting to the newly advertised broker. The subsequent connection only sees the new broker and connects to it properly, but I believe this is the source of the periodic error logging we've been seeing. Fixing it is out of scope here, but it involves a mechanism for assigning each data-plane's Dekaf instance(s) a stable numeric identifier that we can use as the broker ID.

**Note:** I discovered a race condition in Gazette while working on these tests that causes them to have an unfortunately high upper bound on runtime: about 8 minutes. 

Briefly, when a journal is migrated, one of the steps is suspending it in the source data-plane. When a journal is suspended, two things happen: its un-persisted fragments are scheduled for immediate upload, and its replication factor is reduced from 3 to 1. The race condition is: if the broker that _was_ the primary for the journal being suspended ends up being picked as the sole remaining broker in the topology, the fragment persistence happens promptly. Otherwise, it falls back to the backup fragment upload pathway which takes up to 3 minutes. 

If all of the necessary fragments aren't present when the journal is then published to the destination data-plane, we have to wait out the 5 minute default fragment refresh interval before they get picked up. Hence the 8 minute upper bound on runtime.

The following output is with a temporarily modified faster fragment refresh interval:
```
 Nextest run ID db351aec-39d7-4872-a37e-e4ef30c2f1fb with nextest profile: dekaf-e2e
    Starting 2 tests across 1 binary (14 tests and 3 binaries skipped)
        SLOW [> 60.000s] dekaf::e2e migration::test_rdkafka_handles_redirect
        SLOW [> 60.000s] dekaf::e2e migration::test_migration_protocol_responses
        SLOW [>120.000s] dekaf::e2e migration::test_migration_protocol_responses
        SLOW [>120.000s] dekaf::e2e migration::test_rdkafka_handles_redirect
        PASS [ 170.702s] dekaf::e2e migration::test_migration_protocol_responses
        SLOW [>180.000s] dekaf::e2e migration::test_rdkafka_handles_redirect
        PASS [ 187.367s] dekaf::e2e migration::test_rdkafka_handles_redirect
```


crates/data-plane-controller/src/controller.rs
```diff
@@ -591,6 +591,27 @@ impl Controller {
             .deployments
             .retain_mut(stack::Deployment::mark_current);
 
+        // Compute dekaf addresses based on whether Dekaf is deployed with current > 0.
```
jshearer, line 594: I waffled a bit on where exactly in the state machine to update these values, and settled on here because it felt like the best spot:

* It's the first place where the deployment is considered current
* The instances will be up (Pulumi) and configured/running Dekaf (Ansible)

Anybody disagree?

crates/dekaf/src/session.rs
```diff
@@ -262,9 +287,24 @@ impl Session {
             .await?;
         tracing::debug!(collections=?ops::DebugJson(&collection_names), "fetched all collections");
 
-        let collection_statuses = self
-            .fetch_collections_for_metadata(collection_names)
-            .await?;
+        let names_for_redirect = collection_names.clone();
```
jshearer, line 290: Caught this in testing: the `Metadata` request can either include a list of topics to explicitly request metadata for, or if that list is empty, the request is for all available topics. 

Back in [3b84916590](https://github.com/estuary/flow/commit/3b84916590) (PR #2497) I broke this codepath specifically if the task is a redirect. That hasn't been deployed yet, specifically to `dekaf.estuary-data.com`, so it didn't get caught until I exercised it in these tests.

local/systemd/flow-plane-link@.service
```diff
@@ -15,9 +15,15 @@ ExecStartPre=sleep 2
 
 # Call control-plane API to register the data plane
 # Use printf to build JSON and avoid shell quoting complexities
+# DEKAF_ADDRESS and DEKAF_REGISTRY_ADDRESS are optional; if not set, omit them from the payload
```
jshearer, line 18: I'm not sure whether this should be optional or not. Basically, do we ever care about running a local data-plane using `mise run local:data-plane` _without_ it including Dekaf? My initial reaction was that Dekaf should be optional and opt-in, but it does introduce some additional branching here. I don't imagine most people want to run Dekaf locally by default though.

mise/tasks/local/gazette
```diff
@@ -32,14 +32,14 @@ BROKER_ALLOW_ORIGIN=http://localhost:3000
 BROKER_AUTH_KEYS=${HMAC_KEY}
 BROKER_AUTO_SUSPEND=true
 BROKER_FILE_ONLY=true
-BROKER_FILE_ROOT=${FLOW_LOCAL}/fragments/${DATA_PLANE}/
+BROKER_FILE_ROOT=${FLOW_LOCAL}/fragments/
```
jshearer, line 35: This is a notable change, it causes all local data-planes to share the same backing storage, so fragments persisted by one will be visible to another.

crates/data-plane-controller/src/controller.rs
psFried: I don't disagree, but also am not the best qualified to say, so thinking we should solicit another opinion on this. @qaoj are you able to weigh in here?

local/systemd/flow-plane-link@.service
psFried: My 2 cents is that it'd be fine to just always run dekaf locally, since it probably wouldn't consume much memory or cpu at idle. I'm not opposed to it being conditional, assuming it's easy to enable.

mise/tasks/ci/dekaf-e2e
```diff
@@ -32,39 +33,60 @@ until nc -z localhost 8675 2>/dev/null; do
...
-    sleep 1
-    ELAPSED=$((ELAPSED + 1))
-    if [ "${ELAPSED}" -ge "${TIMEOUT}" ]; then
-        echo "Error: Dekaf failed to start within ${TIMEOUT} seconds"
-        exit 1
-    fi
+# local-cluster with Dekaf is expected to already be running from local:stack
+# Only set up local-cluster-2 for migration tests
+echo "Setting up local-cluster-2 dataplane with Dekaf..."
+mise run local:data-plane local-cluster-2 8200 --num-brokers 4 --num-reactors 1 --link --dekaf
```
psFried, line 45: I guess mise doesn't have a way to provide arguments to the task if you declare it as a normal dependency? Feels a little weird to run it this way, when there's also declarative dependencies, and just makes me wonder whether we're missing some other way to do this. But I took a peek at the mise docs, and didn't see anything so I'm supposing that's why we're doing this? 🤷 No action necessary here, assuming that's the case

supabase/migrations/20260109190246_dekaf_address.sql
```diff
@@ -0,0 +1,19 @@
+-- Add dekaf_address and dekaf_registry_address to data_planes table.
+-- dekaf_address: URL with scheme, e.g., "tls://dekaf.gcp-us-central1-c1.dp.estuary-data.com:9092"
+-- dekaf_registry_address: URL with scheme, e.g., "https://dekaf.gcp-us-central1-c1.dp.estuary-data.com"
+-- Nullable because not every data-plane has a Dekaf instance.
+-- For managed data-planes, these are set by the data-plane-controller when Dekaf is deployed.
+-- For manual data-planes, these can be set via the create_data_plane API.
+ALTER TABLE public.data_planes
```
psFried, line 9: Wrap all this in a transaction for safety

psFried (approved):
LGTM % a few comments. The tests look great.

supabase/migrations/20260109190246_dekaf_address.sql
jshearer: Done

crates/data-plane-controller/src/controller.rs
qaoj: Spoke about this on standup, it makes sense for this to be here and I don't really see anything that could disrupt the workings of dekaf or the rest of the control plane. Also makes sense in the context of the supabase migration file

local/systemd/flow-plane-link@.service
jshearer: Alright... I'm leaning slightly towards keeping it optional/opt-in just to avoid pushing Dekaf on everyone unnecessarily.

mise/tasks/ci/dekaf-e2e
jshearer: Yeah, this was the best way I could find to do it

PR #2628, opened 2026-01-21
control-plane: add aud in generate_access_token

psFried:
A previous commit had updated our JWT claims parsing to require the `aud` field. This updates the `generate_access_token` database function to set the `aud` field, so that refreshed tokens can be used by flowctl.


skord (approved):
LGTM

PR #2633, opened 2026-01-26
rocksdb: batch state merges to handle many operands

jgraettinger:
Connectors may emit many small merge-patch updates that accumulate into hundreds of thousands of merge operands. Previously this could exhaust memory during compaction or recovery.

Refactor do_merge to process documents in memory-bounded batches (starting at 32MB), iterating until convergence to a single output. If batch count doesn't decrease, the threshold doubles up to 1024MB before failing with an explicit error.

### DIFF
--- a/crates/runtime/src/rocksdb.rs
+++ b/crates/runtime/src/rocksdb.rs
@@ -160,8 +160,6 @@ impl RocksDB {
         {
             let state = String::from_utf8(state).context("decoding connector state as UTF-8")?;
             let state = models::RawValue::from_string(state).context("decoding state as JSON")?;
-
-            tracing::debug!(state=?ops::DebugJson(&state), "loaded a persisted connector state");
             return Ok(state);
         };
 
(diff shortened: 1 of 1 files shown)


crates/runtime/src/rocksdb.rs
```diff
@@ -323,60 +321,147 @@ fn do_merge(
...
     schema: &str,
 ) -> anyhow::Result<Vec<u8>> {
-    let bundle = doc::validation::build_bundle(schema.as_bytes()).unwrap();
-    let validator = doc::Validator::new(bundle).unwrap();
+    // Holds outputs of prior iterations, which are now inputs of the next one.
+    let mut input_storage: Vec<Vec<u8>>;
+
+    // Collect all input documents (initial + operands, each newline-delimited).
+    let mut inputs: Vec<&[u8]> = Vec::new();
+    if let Some(initial) = initial {
```
williamhbaker, line 329: Mostly curious: Can `initial` ever have more than one document?

jgraettinger: In practice it's always either None or Some("{}"). There was some migration handling at one point which applied a connector state stored in a different way as `initial`, but that was done long ago now.

PR #2634, opened 2026-01-26
dekaf: preserve upstream connection mode for group coordinator / controller connections

jshearer:
Previous to https://github.com/estuary/flow/pull/2566, Dekaf would always use TLS when talking to upstream brokers. When we introduced support for non-TLS connections (for e2e testing) by interpreting the URI scheme, I forgot about `connect_to_group_coordinator` and `connect_to_controller` which need to be made aware of the configured URI in order to work properly.

Now,
  - `tcp://` means plaintext
  - `tls://` means TLS

Since `connect_to_group_coordinator` and `connect_to_controller` had hardcoded `tcp://`, which now means plaintext, they worked fine in e2e tests which use plaintext brokers, but fail when connecting to MSK clusters which require TLS.

This change now extracts the scheme out of the URL from `try_connect` and uses it later when opening other connections using the same `KafkaApiClient`.

### DIFF
--- a/crates/dekaf/src/api_client.rs
+++ b/crates/dekaf/src/api_client.rs
@@ -35,6 +35,26 @@ enum ConnectionScheme {
     Tls,
 }
 
+impl ConnectionScheme {
+    fn as_str(&self) -> &'static str {
+        match self {
+            ConnectionScheme::Plaintext => "tcp",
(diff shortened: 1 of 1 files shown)


jgraettinger (approved):
LGTM

PR #2643, opened 2026-01-28
Revert "runtime/capture: Return ~immediately on connector EOF"

jgraettinger:
Reverts estuary/flow#2591

This commit causes a hot-loop of polls between the Go and Rust portions of the runtime, for classes of connectors which perform a periodic sweep and then exit.


willdonnelly (approved):
LGTM

PR #2646, opened 2026-01-30
Greg/mutation/connection health

GregorShear:
**Description:**

- expose a `testConnectionHealth` mutation that returns per data-plane/store results
- return structured health check items (including errors) instead of GraphQL extensions
- remove `dryRun` param from create/update mutations
- only fail update mutations when newly added stores or data planes fail health checks
- create mutation logic unchanged

**Workflow steps:**
request:
```gql
mutation {
  testConnectionHealth(
    catalogPrefix: "gregCo/"
    storage: {
      stores: [
        {provider: "GCS", bucket: "my-data-a"}
        {provider: "GCS", bucket: "my-data-b"}
      ]
      data_planes: ["ops/dp/public/local-cluster"]
    }
  ) {
    results {
      fragmentStore
      dataPlaneName
      error
    }
  }
}
```

response:
```json
{
  "data": {
    "testConnectionHealth": {
      "results": [
        {
          "fragmentStore": "gs://my-data-a/",
          "dataPlaneName": "ops/dp/public/local-cluster",
          "error": null
        },
        {
          "fragmentStore": "gs://my-data-b/",
          "dataPlaneName": "ops/dp/public/local-cluster",
          "error": "Simulated random failure"
        }
      ]
    }
  }
}
```

### DIFF
--- a/crates/control-plane-api/src/server/public/graphql/storage_mappings.rs
+++ b/crates/control-plane-api/src/server/public/graphql/storage_mappings.rs
@@ -18,13 +18,20 @@ pub struct StorageHealthItem {
 
 /// Result of checking storage health for a catalog prefix.
 #[derive(Debug, Clone, async_graphql::SimpleObject)]
+#[graphql(complex)]
 pub struct ConnectionHealthTestResult {
     /// The catalog prefix for which storage health was checked.
     pub catalog_prefix: models::Prefix,
-    /// Whether all health checks passed.
-    pub all_passed: bool,
     /// Individual health check results for each data plane and store combination.
     pub results: Vec<StorageHealthItem>,
+}
+
+#[async_graphql::ComplexObject]
+impl ConnectionHealthTestResult {
+    /// Whether all health checks passed.
+    pub async fn all_passed(&self) -> bool {
+        self.results.iter().all(|c| c.error.is_none())
+    }
 }
 
 /// Result of creating a storage mapping.
@@ -320,6 +327,10 @@ impl StorageMappingsMutation {
         });
 
         if has_new_failures {
+            // We only fail on health check errors for newly added stores or data planes.
+            // Tasks under this storage mapping will still be broken if there are any failing
+            // health checks, but we allow the update so long as the user isn't adding more
+            // problems than there already were.
             return Err(async_graphql::Error::new(
                 "Storage health checks failed for newly added stores or data planes",
             ));
@@ -398,11 +409,9 @@ impl StorageMappingsMutation {
 
         // Run health checks and collect results.
         let results = run_all_health_checks(&catalog_prefix, &data_planes, &storage.stores).await;
-        let all_passed = results.iter().all(|c| c.error.is_none());
 
         Ok(ConnectionHealthTestResult {
             catalog_prefix,
-            all_passed,
             results,
         })
     }


crates/control-plane-api/src/server/public/graphql/storage_mappings.rs
```diff
@@ -336,28 +305,29 @@ impl StorageMappingsMutation {
...
+            let is_new_store = !current.0.stores.contains(&c.fragment_store.0);
+            let is_new_dp = !current.0.data_planes.contains(&c.data_plane_name);
+            is_new_store || is_new_dp
+        });
+
+        if has_new_failures {
+            return Err(async_graphql::Error::new(
+                "Storage health checks failed for newly added stores or data planes",
+            ));
+        }
```
GregorShear, line 337: @psFried i'm thinking on the update mutation we don't block the change if previously-saved dataplanes/stores are failing (but the front end will still call attention to those connections in the UI)

crates/control-plane-api/src/server/public/graphql/storage_mappings.rs
```diff
@@ -2,33 +2,42 @@ use crate::directives::storage_mappings::{
...
 /// Result of testing storage health for a single data plane and store.
-#[derive(Debug, Clone)]
-struct StorageHealthResult {
+#[derive(Debug, Clone, async_graphql::SimpleObject)]
+pub struct StorageHealthItem {
+    /// Name of the data plane that was checked.
     data_plane_name: String,
-    fragment_store: url::Url,
+    /// The fragment store that was checked.
+    fragment_store: async_graphql::Json<models::Store>,
```
GregorShear, line 14: i'm changing this from the resolved store URL to the original json format that it was sent from the client so that the client doesn't need to parse that URL in order to match the health check outcome with the store in the UI

crates/control-plane-api/src/server/public/graphql/storage_mappings.rs
```diff
@@ -2,33 +2,42 @@ use crate::directives::storage_mappings::{
...
     error: Option<String>,
 }
 
+/// Result of checking storage health for a catalog prefix.
+#[derive(Debug, Clone, async_graphql::SimpleObject)]
+pub struct ConnectionHealthTestResult {
+    /// The catalog prefix for which storage health was checked.
+    pub catalog_prefix: models::Prefix,
+    /// Whether all health checks passed.
+    pub all_passed: bool,
```
psFried: nit: There's a way of handling computed fields like this dynamically, which I think would be a little more DRY:

```
#[async_graphql::ComplexObject]
impl ConnectionHealthTestResult {
  pub async fn all_passed(&self) -> bool {
    self.results.iter().all(|r| r.error.is_none())
  }
}
```

Here's the docs on that: https://async-graphql.github.io/async-graphql/en/define_simple_object.html#user-defined-resolvers

crates/control-plane-api/src/server/public/graphql/storage_mappings.rs
psFried: My initial reaction is that this seems unexpected, and might not be what we want. If any of the stores (regardless of whether it's been modified or is the primary store) have health check errors, then none of the tasks will be able to make progress. I guess I could see an argument for letting people remove old stores, even if some of their current stores are still failing. Is that what you were thinking about here?

GregorShear: My thought is that the failures are unrelated to the changes the user is trying to make - do we want to block the addition of a healthy store because of a pre-existing condition? The failing health checks will remain when we turn the user away.

What if the user just wants to update the `detail` field on the storage mapping? Should we block that until they resolve the health checks? (maybe? genuine question)

Is there a scenario where a user's primary bucket is failing and they want to go add a new one so they can continue capturing data while they troubleshoot the old one? I'm kinda making things up here, but my reasoning is that the health checks should prevent the user from causing problems, and not prevent the user from making good changes when there are problems nearby.

Happy to defer to your gut feeling on this - easy to set this special handling aside and add it back in later if needed.

psFried: > Is there a scenario where a user's primary bucket is failing and they want to go add a new one so they can continue capturing data while they troubleshoot the old one?

My initial hesitance here was because this is _not_ something that would work today. If any of the stores fail the health check, then the task is not going to work. One of those stores might contain data that's necessary in order for the task to run, and it'd be very difficult for the data plane to ever rule that out.

That said, I think it's pretty fair to say that you can update broken storage mappings as long as you're not breaking them any more than they already were, as long as the UX makes it clear that nothing is going to work until all the stores are healthy. I think we're good to move forward the way it is here.

psFried (approved):
LGTM

PR #2650, opened 2026-02-02
dekaf: fix `OffsetFetch` during leader epoch migration

jshearer:
## Problem

While canarying #2497, I noticed some unexpected consumer behavior that looked like a backfill. No backfills were triggered, so I started looking for a bug, and found this.

The context is that #2497 includes a migration that changes the way we map topic names into upstream Kafka topic names for the purposes of serving group management APIs. Specifically, we must now include the leader epoch (materialization binding backfill counter) as part of the topic name, as any committed offsets with one epoch are meaningless in the context of another. 

In order to avoid this exact problem of consumers needing to start over from the beginning, I implemented logic to fall back to the previous topic name mapping in the event that an `OffsetFetch` request found no committed offsets for the requested topic/partition. The bug is that these requests can contain multiple topic/partitions to fetch offsets for, and it would only fall back to looking up "legacy" offsets if _none_ of the requested topic/partitions come back using the new mapping. If even one does, then all the other topic/partitions that haven't been committed yet will show as `committed_offset=-1`, which causes the consumer to apply its `auto.offset.reset` logic and start over.

## Fix
In `crates/dekaf/src/session.rs`, refactored `Session::offset_fetch` to perform fallback fetching per requested topic. I feel like it's actually much easier to follow now as well.
- Query epoch-qualified offsets first.
- If any are missing (`error_code == 0` and `committed_offset == -1`), also query for offsets using the old topic name mapping.
- Merge responses together rather than being all-or-nothing

This prevents offsets from being reported as uncommitted if they haven't been committed yet under the new mapping and others have, avoiding the consumer reset behavior observed after the rollout.
## Tests
`crates/dekaf/tests/e2e/consumer_group.rs` updates `test_offset_fetch_legacy_fallback` with `crates/dekaf/tests/e2e/fixtures/two_topics.flow.yaml` to simulate partial migration, including directly interacting with the test Kafka broker to set up the proper scenario:
- `topic_a` has an epoch-qualified committed offset.
- `topic_b` has only a legacy committed offset.
- A single explicit-topics OffsetFetch must return both offsets.

Before the fix is applied, the test fails showing the unexpected committed offset of `-1`:
```
[ci:dekaf-e2e]  Nextest run ID aafd2f6d-abe1-4521-b8dd-38fe1fcbedd8 with nextest profile: dekaf-e2e
[ci:dekaf-e2e]     Starting 1 test across 1 binary (16 tests and 3 binaries skipped, including 3 binaries via profile.dekaf-e2e.default-filter)
[ci:dekaf-e2e]        START (1/1) dekaf::e2e consumer_group::test_offset_fetch_legacy_fallback
[ci:dekaf-e2e]
[ci:dekaf-e2e] running 1 test
[ci:dekaf-e2e] 2026-02-02T20:19:50.388500Z  INFO e2e::harness: Setting up test environment namespace=test/dekaf/legacy_fallback/d4e4
[ci:dekaf-e2e] 2026-02-02T20:19:54.450455Z  INFO e2e::harness: Publish succeeded
[ci:dekaf-e2e] 2026-02-02T20:19:54.450475Z  INFO e2e::harness: Waiting for shard to become primary task_name=test/dekaf/legacy_fallback/d4e4/source_ingest
[ci:dekaf-e2e]
[ci:dekaf-e2e] thread 'consumer_group::test_offset_fetch_legacy_fallback' (2684572) panicked at crates/dekaf/tests/e2e/consumer_group.rs:314:5:
[ci:dekaf-e2e] assertion `left == right` failed: expected topic_b to fall back to legacy committed offset 7
[ci:dekaf-e2e]   left: -1
[ci:dekaf-e2e]  right: 7
[ci:dekaf-e2e] note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
[ci:dekaf-e2e] test consumer_group::test_offset_fetch_legacy_fallback ... FAILED
[ci:dekaf-e2e]
[ci:dekaf-e2e] failures:
[ci:dekaf-e2e]
[ci:dekaf-e2e] failures:
[ci:dekaf-e2e]     consumer_group::test_offset_fetch_legacy_fallback
[ci:dekaf-e2e]
[ci:dekaf-e2e] test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 16 filtered out; finished in 18.54s
[ci:dekaf-e2e]
[ci:dekaf-e2e]         FAIL [  18.541s] (1/1) dekaf::e2e consumer_group::test_offset_fetch_legacy_fallback
[ci:dekaf-e2e]   Cancelling due to test failure:
[ci:dekaf-e2e] ────────────
[ci:dekaf-e2e]      Summary [  18.541s] 1 test run: 0 passed, 1 failed, 16 skipped
[ci:dekaf-e2e]         FAIL [  18.541s] (1/1) dekaf::e2e consumer_group::test_offset_fetch_legacy_fallback
[ci:dekaf-e2e] error: test run failed
```
---
**Note:** This branch also includes a couple more e2e tests for consumer group API behavior. I decided to roll #2597 into this PR to make reviewing easier:

- Add consumer-group behavior coverage:
  - Commit and resume with a stable `group.id` (rdkafka), including committed-offset readback
  - `Metadata` and `OffsetCommit` error mapping for unknown topics (`UnknownTopicOrPartition`)
  - `OffsetFetch` behavior for never-committed groups (committed_offset = -1, committed_leader_epoch = -1)
- Collection reset behavior:
  - After a collection reset, OffsetFetch does not return offsets committed under the prior epoch

### DIFF
--- a/crates/dekaf/src/session.rs
+++ b/crates/dekaf/src/session.rs
@@ -6,6 +6,7 @@ use crate::{
 };
 use anyhow::{Context, bail};
 use bytes::{Buf, BufMut, Bytes, BytesMut};
+use itertools::Itertools;
 use kafka_protocol::{
     error::{ParseResponseErrorCode, ResponseError},
     messages::{
@@ -1929,7 +1930,7 @@ impl Session {
     #[instrument(skip_all, fields(group=?req.group_id))]
     pub async fn offset_fetch(
         &mut self,
-        mut req: messages::OffsetFetchRequest,
+        req: messages::OffsetFetchRequest,
         header: RequestHeader,
     ) -> anyhow::Result<messages::OffsetFetchResponse> {
         let redirect_response = messages::OffsetFetchResponse::default().with_topics(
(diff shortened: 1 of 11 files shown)


crates/dekaf/src/session.rs
```diff
@@ -1929,7 +1930,7 @@ impl Session {
     #[instrument(skip_all, fields(group=?req.group_id))]
     pub async fn offset_fetch(
```
jshearer, line 1931: Sorry about the large diff in this function, I've been wanting to clean it up and haven't been confident that doing so won't break its behavior until now.

crates/dekaf/src/session.rs
```diff
@@ -1955,152 +1956,179 @@ impl Session {
...
-            {
-                Ok(statuses) => Some(statuses),
-                Err(e) if Self::is_redirect_error(&e) => return Ok(redirect_response),
-                Err(e) => return Err(e),
-            }
+        let mut requested_topics = if let Some(topics) = req.topics.clone() {
+            topics
         } else {
-            None
+            // We don't currently support fetching all topics' offsets.
```
jshearer, line 1962: Just making this condition explicit as previously it was implicit

crates/dekaf/src/session.rs
```diff
@@ -1955,152 +1956,179 @@ impl Session {
...
+        let collection_statuses = match self
+            .fetch_collections(requested_topics.iter().map(|topic| &topic.name))
+            .await
+        {
+            Ok(statuses) => statuses,
+            Err(e) if Self::is_redirect_error(&e) => return Ok(redirect_response),
+            Err(e) => return Err(e),
+        };
+
+        // Check for NotFound/NotReady collections - return per-topic errors and remove from fetch list
```
jshearer, line 1976: The previous logic would error all-or-nothing if even one collection was in the NotReady state, now we return specific errors and keep processing the rest of the requested topics

crates/dekaf/src/session.rs
```diff
@@ -1955,152 +1956,179 @@ impl Session {
...
-        if let Some(ref mut topics) = fallback_req.topics {
-            for topic in topics.iter_mut() {
-                let decrypted = self.decrypt_topic_name(topic.name.clone())?;
-                topic.name = self.encrypt_topic_name(decrypted, None)?;
-            }
-        }
+        // The upgrade path can result in a partially-migrated state where some partitions
+        // have an epoch-qualified committed offset while others still only have a committed
+        // offset from before the migration. Fetch legacy offsets for topics that need them.
+        let (topics_needing_fallback, topics_with_epoch_qualified_offset): (Vec<_>, Vec<_>) =
```
psFried, line 2052: We're able to remove this logic as soon as all existing tasks are migrated, yeah?

psFried (approved):
I love seeing all that new test coverage ❤️ LGTM

crates/dekaf/src/session.rs
jshearer: Yep 👍

PR #2651, opened 2026-02-03
Add GraphQL query for storage mappings

GregorShear:
**Description:**

Adds a new `storageMappings` GraphQL query endpoint that allows users to fetch storage mapping configurations for their catalog prefixes.

Strips the internal `collection-data/` suffix from store prefixes before returning to users, matching the format they originally provided when creating the mapping

```gql
query {
  storageMappings(by:  {
     exactPrefixes: ["gregCo/"]
  }) {
    edges {
      node {
        catalogPrefix
        detail
        storage
        userCapability
      }
    }
  }
}
```

```json
{
  "data": {
    "storageMappings": {
      "edges": [
        {
          "node": {
            "catalogPrefix": "gregCo/",
            "detail": "applied via directive",
            "storage": {
              "stores": [
                {
                  "provider": "GCS",
                  "bucket": "estuary-trial",
                  "prefix": ""
                }
              ],
              "data_planes": [
                "ops/dp/public/local-cluster"
              ]
            },
            "userCapability": "admin"
          }
        }
      ]
    }
  }
}
```

```json
{
  "data": null,
  "errors": [
    {
      "message": "PermissionDenied: test@example.com is not authorized to read catalog prefix 'acmeCo/'",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "storageMappings"
      ]
    }
  ]
}
```

### DIFF
--- a/crates/control-plane-api/src/server/public/graphql/storage_mappings.rs
+++ b/crates/control-plane-api/src/server/public/graphql/storage_mappings.rs
@@ -201,7 +201,7 @@ impl StorageMappingsMutation {
             r#"
             SELECT catalog_name
             FROM live_specs
-            WHERE starts_with(catalog_name, $1)
+            WHERE catalog_name ^@ $1
             AND spec IS NOT NULL
             LIMIT 5
             "#,
@@ -571,7 +571,6 @@ impl StorageMappingsQuery {
 
         // Verify user has read capability to the queried prefixes.
         let claims = env.claims()?;
-        let user_email = claims.email.as_deref().unwrap_or("user");
 
         let prefixes_to_check: Vec<&models::Prefix> = if let Some(ref prefix) = under_prefix {
             vec![prefix]
@@ -579,22 +578,13 @@ impl StorageMappingsQuery {
             exact_prefixes.iter().collect()
         };
 
-        for prefix in prefixes_to_check {
-            let policy_result = if tables::UserGrant::is_authorized(
-                &env.snapshot().role_grants,
-                &env.snapshot().user_grants,
-                claims.sub,
-                prefix,
-                models::Capability::Read,
-            ) {
-                Ok((None, ()))
-            } else {
-                Err(tonic::Status::permission_denied(format!(
-                    "{user_email} is not authorized to read catalog prefix '{prefix}'",
-                )))
-            };
-            env.authorization_outcome(policy_result).await?;
-        }
+        let policy_result = crate::server::evaluate_names_authorization(
+            &env.snapshot(),
+            claims,
+            models::Capability::Read,
+            prefixes_to_check,
+        );
+        env.authorization_outcome(policy_result).await?;
 
         let (rows, has_prev, has_next) =
             connection::query_with::<String, _, _, _, async_graphql::Error>(
@@ -703,10 +693,10 @@ async fn fetch_storage_mappings_after(
             detail,
             spec as "spec!: crate::TextJson<models::StorageDef>"
         FROM storage_mappings
-        WHERE NOT starts_with(catalog_prefix, 'recovery/')
+        WHERE NOT catalog_prefix ^@ 'recovery/'
         AND (
             catalog_prefix::text = any($1::text[])
-            OR ($2::text IS NOT NULL AND starts_with(catalog_prefix, $2::text))
+            OR ($2::text IS NOT NULL AND catalog_prefix ^@ $2::text)
         )
         AND ($3::text IS NULL OR catalog_prefix > $3::text)
         ORDER BY catalog_prefix ASC
@@ -737,8 +727,6 @@ async fn fetch_storage_mappings_before(
     before: Option<&str>,
     limit: i64,
 ) -> anyhow::Result<Vec<StorageMappingRow>> {
-    // TODO: Greg is skeptical about this _catalog_name workaround - investigate further.
-
     // Convert to plain strings to avoid sqlx encoding as `_catalog_name` domain array,
     // which would trigger domain constraint validation on bind.
     let exact_prefixes: Vec<String> = exact_prefixes.iter().map(|p| p.to_string()).collect();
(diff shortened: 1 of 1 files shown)


crates/control-plane-api/src/server/public/graphql/storage_mappings.rs
```diff
@@ -485,3 +489,294 @@ fn resolve_data_planes<'s>(
...
+                prefix,
+                models::Capability::Read,
+            ) {
+                Ok((None, ()))
+            } else {
+                Err(tonic::Status::permission_denied(format!(
+                    "{user_email} is not authorized to read catalog prefix '{prefix}'",
+                )))
+            };
+            env.authorization_outcome(policy_result).await?;
```
GregorShear: Just want to verify that it makes sense to call `authorization_outcome` after throwing the formatted `permission_denied` error...

Doesn't `authorization_outcome` handle the case where permission is denied at first so we wait for a fresh snapshot?

crates/control-plane-api/src/server/public/graphql/storage_mappings.rs
```diff
@@ -485,3 +489,294 @@ fn resolve_data_planes<'s>(
...
+        r#"
+        SELECT
+            catalog_prefix as "catalog_prefix!: String",
+            detail,
+            spec as "spec!: crate::TextJson<models::StorageDef>"
+        FROM storage_mappings
+        WHERE NOT starts_with(catalog_prefix, 'recovery/')
+        AND (
+            catalog_prefix::text = any($1::text[])
+            OR ($2::text IS NOT NULL AND starts_with(catalog_prefix, $2::text))
```
psFried: We've been bitten by `starts_with` in the past, where we've seen that it won't use an index with `starts_with`. But it will use the index if you use the diabolically obfuscated equivalent `catalog_prefix ^@ $2`. This is contrary to the postgres docs, but it caused an outage, so we always just use `^@` now.

crates/control-plane-api/src/server/public/graphql/storage_mappings.rs
psFried: Yeah, but I think this ought to use [`evaluate_names_authorization`](https://github.com/estuary/flow/blob/285a556795cc147e54d90e4ffed8bcd88c02f4dd/crates/control-plane-api/src/server/mod.rs#L66) once for all the prefixes, followed by a single `authorization_outcome` call.

And yeah, the `authorization_outcome` handles everything around snapshot refreshes and retries and such.

crates/control-plane-api/src/server/public/graphql/storage_mappings.rs
```diff
@@ -485,3 +489,294 @@ fn resolve_data_planes<'s>(
...
+}
+
+async fn fetch_storage_mappings_before(
+    db: &sqlx::PgPool,
+    exact_prefixes: &[models::Prefix],
+    under_prefix: Option<&models::Prefix>,
+    before: Option<&str>,
+    limit: i64,
+) -> anyhow::Result<Vec<StorageMappingRow>> {
+    // TODO: Greg is skeptical about this _catalog_name workaround - investigate further.
```
psFried: still todo?

crates/control-plane-api/src/server/public/graphql/storage_mappings.rs
```diff
@@ -485,3 +489,294 @@ fn resolve_data_planes<'s>(
...
+/// Input type for querying storage mappings.
+#[derive(Debug, Clone, async_graphql::InputObject)]
+pub struct StorageMappingsBy {
+    /// Fetch storage mappings by exact catalog prefixes.
+    /// At least one of `exactPrefixes` or `underPrefix` must be provided.
+    pub exact_prefixes: Option<Vec<models::Prefix>>,
+    /// Fetch all storage mappings under this prefix pattern.
+    /// For example, "acmeCo/" returns mappings for "acmeCo/", "acmeCo/team-a/", etc.
+    /// At least one of `exactPrefixes` or `underPrefix` must be provided.
+    pub under_prefix: Option<models::Prefix>,
```
psFried, line 506: I like the clarity of the names here. Elsewhere in the schema we call this one `prefix`, and I'm wondering whether it's worth doing so here, or if we should consider using a more explicit `underPrefix` or `prefixedBy` more universally.

crates/control-plane-api/src/server/public/graphql/storage_mappings.rs
GregorShear: oh ... yeah, this todo eventually led me to the question "should we even be using `models::Prefix` on the input params?"

I notice the alerts endpoints are just using `prefix: String` [here](https://github.com/estuary/flow/blob/1d75158d1b9c15a37adda32647ec06a5b3144983/crates/control-plane-api/src/server/public/graphql/alerts.rs#L12)

crates/control-plane-api/src/server/public/graphql/storage_mappings.rs
GregorShear: Oh yeah, I debated whether or not to stick to the convention, but honestly I think `prefix` is so overloaded that I'd welcome any disambiguation

crates/control-plane-api/src/server/public/graphql/storage_mappings.rs
GregorShear: Ahhh right, i'll make this change.

(and my confusion was that the error is assigned to `policy_result`, not actually thrown before calling `authorization_outcome` ... cool )

crates/control-plane-api/src/server/public/graphql/storage_mappings.rs
psFried: I think probably I should just change the alerts endpoint to use a `models::Prefix`, and we should use that consistently. My reasoning being that `Prefix` then shows up in the graphql schema and has a description saying what it is, and hopefully adds clarity. TBH I'd probably trust your opinion more than my own regarding how much clarity is actually added, so please push back if you feel differently.

There's a `models::Name`, too, which we could be using instead of strings in some types.

I have mixed feelings on whether those types are actually worth using. IMHO, `Prefix` and `Name` feel like a little more trouble than they're worth, outside the context of a public-facing schema. There aren't borrowed variants of them, so we end up just using `&str` a lot, anyway. And in `models::Catalog` and `tables`, we actually don't even use `Name`, but instead use separate string wrapper types for each `Collection`, `Capture`, `Materialization`, and `Test` 🤢 

So I'll propose the following:
- We always use `Prefix` or `Name` in our graphql schema
- In code, we continue to use `&str` whenever we can
- If we need owned variants, we use `Prefix` or `Name` when it's guaranteed to be one or the other
- Use `String` when we need an owned variant that can represent _either_ a `Name` or a `Prefix`

WDYT?

crates/control-plane-api/src/server/public/graphql/storage_mappings.rs
psFried: Sounds good. Maybe we try to use `underPrefix` consistently then? There's a few places that would need updated, but I feel like now's our best chance to try to make it consistent.

psFried (approved):
LGTM

PR #2652, opened 2026-02-03
Add GraphQL query for data-planes

GregorShear:
**Description:**

- Add a new dataPlanes GraphQL query 
- Parse data plane names to extract structured cloudProvider, region, tag (cluster), and isPublic fields for the GraphQL response so the client doesn't have to
- Lazy fetch cloud credential fields needed for storage mappings: 

```rust
cidr_blocks: Vec<String>,
gcp_service_account_email: Option<String>,
aws_iam_user_arn: Option<String>,
azure_application_name: Option<String>,
azure_application_client_id: Option<String>,
```

example gql query:

```gql
query {
  dataPlanes {
    edges {
      cursor
      node {
        dataPlaneName
        cloudProvider
        region
        isPublic
        tag
        userCapability
      }
    }
  }
}
```

```json
{
  "data": {
    "dataPlanes": {
      "edges": [
        {
          "cursor": "ops/dp/public/aws-us-east-1-c1",
          "node": {
            "dataPlaneName": "ops/dp/public/aws-us-east-1-c1",
            "cloudProvider": "AWS",
            "region": "us-east-1",
            "isPublic": true,
            "tag": "1",
            "userCapability": "read"
          }
        },
        {
          "cursor": "ops/dp/public/local-cluster",
          "node": {
            "dataPlaneName": "ops/dp/public/local-cluster",
            "cloudProvider": "LOCAL",
            "region": "local",
            "isPublic": true,
            "tag": "1",
            "userCapability": "read"
          }
        }
      ]
    }
  }
}
```

### DIFF
--- a/crates/control-plane-api/src/server/public/graphql/data_planes.rs
+++ b/crates/control-plane-api/src/server/public/graphql/data_planes.rs
@@ -1,0 +1,425 @@
+use async_graphql::{
+    Context, SimpleObject,
+    types::connection::{self, Connection},
+};
+use std::collections::HashMap;
+
+const DEFAULT_PAGE_SIZE: usize = 50;
+
+/// Cloud provider where the data plane is hosted.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, async_graphql::Enum)]
+pub enum DataPlaneCloudProvider {
+    Aws,
+    Azure,
+    Gcp,
+    Local,
+}
+
+/// A data plane where tasks execute and collections are stored.
+#[derive(Debug, Clone, SimpleObject)]
+pub struct DataPlane {
+    /// Name of this data-plane under the catalog namespace.
+    pub data_plane_name: String,
+    /// Fully-qualified domain name of this data-plane.
+    pub data_plane_fqdn: String,
+    /// Address of reactors within the data-plane.
+    pub reactor_address: String,
+    /// The current user's capability to this data plane's name prefix.
+    pub user_capability: models::Capability,
+    /// Cloud provider where this data-plane is hosted.
+    pub cloud_provider: DataPlaneCloudProvider,
+    /// Cloud region where this data-plane is hosted.
+    /// For example: "us-east-1" (AWS), "us-central1" (GCP), "eastus" (Azure).
+    pub region: String,
+    /// Tag (cluster) identifier within the region.
+    pub tag: String,
+    /// Whether this is a public data-plane.
+    pub is_public: bool,
+    /// CIDR blocks for this data-plane.
+    pub cidr_blocks: Vec<String>,
+    /// GCP service account email for this data-plane.
+    pub gcp_service_account_email: Option<String>,
+    /// AWS IAM user ARN for this data-plane.
+    pub aws_iam_user_arn: Option<String>,
+    /// Azure application name for this data-plane.
+    pub azure_application_name: Option<String>,
+    /// Azure application client ID for this data-plane.
+    pub azure_application_client_id: Option<String>,
+}
+
+/// Fetches detail fields for the given data plane names from the database.
+/// Returns a map from data_plane_name to its detail fields.
+async fn fetch_data_plane_details(
+    pg_pool: &sqlx::PgPool,
+    names: &[String],
+) -> async_graphql::Result<HashMap<String, DataPlaneDetails>> {
+    tracing::debug!(count = names.len(), "loading data_plane details");
+
+    let names_ref: Vec<&str> = names.iter().map(String::as_str).collect();
+
+    let rows = sqlx::query!(
+        r#"select
+            dp.data_plane_name,
+            dp.cidr_blocks::text[] as "cidr_blocks!: Vec<String>",
+            dp.gcp_service_account_email,
+            dp.aws_iam_user_arn,
+            dp.azure_application_name,
+            dp.azure_application_client_id
+        from unnest($1::text[]) as input(name)
+        join data_planes dp on dp.data_plane_name = input.name
+        "#,
+        &names_ref as &[&str],
+    )
+    .fetch_all(pg_pool)
+    .await?;
+
+    Ok(rows
+        .into_iter()
+        .map(|row| {
+            (
+                row.data_plane_name,
+                DataPlaneDetails {
+                    cidr_blocks: row.cidr_blocks,
+                    gcp_service_account_email: row.gcp_service_account_email,
+                    aws_iam_user_arn: row.aws_iam_user_arn,
+                    azure_application_name: row.azure_application_name,
+                    azure_application_client_id: row.azure_application_client_id,
+                },
+            )
+        })
+        .collect())
+}
+
+struct DataPlaneDetails {
+    cidr_blocks: Vec<String>,
+    gcp_service_account_email: Option<String>,
+    aws_iam_user_arn: Option<String>,
+    azure_application_name: Option<String>,
+    azure_application_client_id: Option<String>,
+}
+
+/// Parses a data plane name into its component parts.
+/// Returns None if the name format is invalid.
+///
+/// Expected formats:
+/// - Cloud: "ops/dp/public/aws-us-east-1-c1" or "ops/dp/private/gcp-us-central1-c2"
+/// - Local: "ops/dp/local/local-foo" (any suffix after "local-")
+pub(crate) fn parse_data_plane_name(
+    name: &str,
+) -> Option<(DataPlaneCloudProvider, String, String, bool)> {
+    let last_segment = name.rsplit('/').next()?;
+    let (provider_str, after_provider) = last_segment.split_once('-')?;
+
+    match provider_str {
+        "local" => Some((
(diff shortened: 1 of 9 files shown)


crates/control-plane-api/src/server/public/graphql/data_planes.rs
```diff
@@ -0,0 +1,304 @@
...
+        let env = ctx.data::<crate::Envelope>()?;
+        let claims = env.claims()?;
+        let snapshot = env.snapshot();
+
+        // Filter to only data planes the user can read and that have valid names.
+        let accessible_data_planes: Vec<&tables::DataPlane> = snapshot
+            .data_planes
+            .iter()
+            .filter(|dp| {
+                parse_data_plane_name(&dp.data_plane_name).is_some()
```
psFried: This seems less than ideal, as it will silently skip rows having an invalid name.  My sense is that we should try to return an error, but maybe it's worth talking through.

GregorShear: failing the entire query feels a little heavy handed - i'm at least adding a log warning so it's not completely silent. happy to chat about it, too if you like

crates/control-plane-api/src/server/public/graphql/data_planes.rs
```diff
@@ -0,0 +1,489 @@
...
+        let env = ctx.data::<crate::Envelope>()?;
+        let claims = env.claims()?;
+        let snapshot = env.snapshot();
+
+        // Filter to only data planes the user can read and that have valid names.
+        let accessible_data_planes: Vec<&tables::DataPlane> = snapshot
+            .data_planes
+            .iter()
+            .filter(|dp| {
+                if parse_data_plane_name(&dp.data_plane_name).is_none() {
```
psFried, line 193: I think we should also validate the names as part of [`create_data_plane`](https://github.com/estuary/flow/blob/78f970992c03ccbcb098db80407ac80426553610/crates/control-plane-api/src/server/create_data_plane.rs#L57), so that we don't accidentally add planes with invalid names. Otherwise, just skipping them here is liable to make things difficult to debug, since most people creating data planes won't know to go look for this in the agent-api logs.

crates/control-plane-api/src/server/public/graphql/data_planes.rs
```diff
@@ -0,0 +1,489 @@
...
+    /// Whether this is a public data-plane.
+    pub is_public: bool,
+}
+
+/// Typed key for loading cloud-provider details of data planes from the database.
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct DataPlaneDetailsKey(pub String);
+
+#[derive(Clone)]
+pub struct DataPlaneDetails {
```
psFried: I'm recalling feeling ambivalent during our last conversation about whether it's worth pulling any fields from the Snapshot apart from the name. But now that I'm looking at the code, it seems like it might be simpler to have the `Loader` impl just return a `DataPlane` struct with all the fields filled in. Could still do all the pagination just based on the names from the snapshot, and then call `loader.load_all(names_page)` and return the results as edges. Then we wouldn't need to have separate resolver functions or `DataPlaneDetails`. What do you think?

crates/flow-client/control-plane-api.graphql
```diff
@@ -836,6 +930,13 @@ type QueryRoot {
 	Results are paginated and sorted by catalog_prefix.
 	"""
 	storageMappings(by: StorageMappingsBy!, after: String, before: String, first: Int, last: Int): StorageMappingConnection!
+	"""
+	Returns data planes accessible to the current user.
+	
+	Results are paginated and sorted by data_plane_name.
+	Only data planes the user has at least read capability to are returned.
+	"""
+	dataPlanes(after: String, before: String, first: Int, last: Int): DataPlaneConnection!
```
psFried, line 939: Does the UI want _all_ the data planes that a user has permission to see? I ask because I'm wondering whether we'd want to filter them based on the effective storage mapping. If this query is being run as part of a publish workflow, then we could potentially filter the list to only the data planes that are permitted by the storage mapping. Not sure if that's helpful, though.

crates/control-plane-api/src/server/public/graphql/data_planes.rs
```diff
@@ -0,0 +1,489 @@
...
+    match provider_str {
+        "local" => Some((
+            CloudProvider::Local,
+            "local".to_string(),
+            "1".to_string(),
+            true,
+        )),
+        "aws" | "az" | "azure" | "gcp" => {
+            // Must have privacy indicator in path.
+            if !name.contains("ops/dp/private/") && !name.contains("ops/dp/public/") {
```
psFried: nit: can we use `starts_with` for these? Just seems like it'd help readability since these are prefixes.

crates/control-plane-api/src/server/public/graphql/data_planes.rs
```diff
@@ -0,0 +1,489 @@
...
+                return None;
+            }
+
+            // Parse cluster suffix (e.g., "-c1", "-c5").
```
psFried: Do we need to parse these? Seems maybe preferable to leave them opaque, just to leave ourselves more options in the future. Can `cluster` be something like `c2`?

crates/flow-client/control-plane-api.graphql
GregorShear: it makes sense to me to constrain dataplane access to users, maybe depending on a (some) storage mapping, but i don't know enough to make a strong case for it yet.

crates/control-plane-api/src/server/public/graphql/data_planes.rs
GregorShear: Yep, I'm fine with this. In fact, I don't think we need the data loader at all if we take this path (do we?) - i've made this change if you want to see how it looks.

crates/control-plane-api/src/server/public/graphql/data_planes.rs
GregorShear: Made this change - currently reaching al the way into graphql/data_planes.rs from create ... is there a better place for this parse function to live?

psFried: Thanks, I think what you did in `create_data_plane` is perfect for right now.

> is there a better place for this parse function to live?

I feel like there _should_ be, but I don't think there is one yet. I've been thinking about some possible agent/control-plane-api refactors, and maybe we just leave this as is for now and try to address it later, after we can align on what a better factoring would look like?

crates/flow-client/control-plane-api.graphql
```diff
@@ -274,6 +274,16 @@ enum CatalogType {
 	test
 }
 
+"""
+Cloud provider where the data plane is hosted.
+"""
+enum CloudProvider {
```
psFried: this is pretty nit picky, so feel free to disagree: it might be clearer to name this `DataPlaneCloudProvider`, so it doesn't get confused with cloud providers in other contexts, like storage mappings.

crates/flow-client/control-plane-api.graphql
```diff
@@ -168,7 +168,7 @@ enum AlertType {
 	are peformed on all specs for a variety of reasons. For example,
 	updating inferred schemas, or updating materialization bindings to match
 	the source capture. When these publications fail, tasks are likely to
-	stop functioning correctly until the issue can be address.
+	stop functioning correctly until the issue can be addressed.
```
psFried, line 171: Thank you!

crates/control-plane-api/src/server/public/graphql/data_planes.rs
```diff
@@ -0,0 +1,448 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::test_server;
+
+    #[sqlx::test(
+        migrations = "../../supabase/migrations",
+        fixtures(path = "../../../fixtures", scripts("data_planes", "alice"))
```
psFried, line 322: I was initially confused by why the `data_planes` fixture wasn't sufficient for this test, but now I see that the names there are likely not parseable, so that's why they don't show up in the output?

Long term at least, it seems like we ought to update the `data_planes` fixture to use realistic names. I'm not sure about the blast radius of that change, though. Did you try it? I think it's OK to leave this for now, but maybe worth adding a comment explaining what the fixture is actually doing in this scenario. Or possibly add a `valid(or something)_data_planes` fixture, which we could re-use in other test scenarios. We don't necessarily need to do anything here at the moment, though.

crates/flow-client/control-plane-api.graphql
GregorShear: pick the nits! i'll change it

crates/control-plane-api/src/server/public/graphql/data_planes.rs
GregorShear: Oh ... i didn't try it, was wary of blast radius so i just punted. 

Taking a closer look now and yes, i'll update the existing features and use them.

PR #2660, opened 2026-02-05
Check for stale query cache during CI

GregorShear:
**Description:**

- Add SQLx check to github CI (and local `ci:platform-test`) so Greg doesn't forget to update the query cache and break builds anymore.
- Run format check and SQLx check in separate lightweight jobs, running in parallel with the main platform test, on standard ubuntu-24.04 runners to avoid queuing behind heavy jobs on the large runner pool
- Cache the sqlx-cli binary (pinned to 0.8.6) so it only compiles on the first run


jshearer (approved):
Seems good to me. We've been bitten a couple times now by not remembering to run `sqlx prepare` 👍

PR #2662, opened 2026-02-05
runtime: fix batched RocksDB state merges to preserve null deletion markers

jgraettinger:
## Summary

- Reapplies the batched merge operator for RocksDB connector state (previously reverted), which processes large operand lists in bounded-memory batches to avoid OOM on merge-heavy workloads
- Fixes a correctness bug where non-first batches used `full` reduction, causing LWW delete to strip null deletion markers within the batch instead of preserving them for re-merge with the base value. Now only the first batch (which contains the base/initial value) uses `full`; subsequent batches use associative reduction.
- Adds targeted regression tests and a QuickCheck fuzz suite for the batched merge path

## Background

RocksDB merge operators for connector state use the RFC 7396 merge-patch schema,
where `null` values are deletion markers that remove the corresponding property
from the base object during full reduction. When many operands accumulate
(e.g. 50k+ cursor updates), the merge operator processes them in batches to
bound memory usage.

The bug: all batches were passed the caller's `full` flag. Batches after the
first contain only operands (no base value), so `full=true` caused null markers
to be consumed within the batch — deleting properties that should only be deleted
when the batch output is later merged with the actual base. The fix restricts
`full` to only the first batch of each iteration, which always contains the
base value.

## What changed

**`crates/runtime/src/rocksdb.rs`**

- Extracted testable `do_merge(full, key, inputs, schema, batch_byte_target, batch_op_target)` from the RocksDB-specific `do_merge_rocks` wrapper, enabling direct unit testing without a RocksDB instance
- Added `is_first_batch` tracking: only the first batch per iteration uses `full` reduction; subsequent batches use associative (partial) reduction to preserve null deletion markers
- Added `batch_op_target` parameter (document count limit per batch) alongside the existing byte threshold, enabling deterministic batch boundaries in tests

## Testing

The `doc` crate already has extensive merge-patch test coverage:

- **`doc/tests/merge_patch_fuzz.rs`** — QuickCheck tests for `reduce_stack` and `reduce_combiner` paths, verifying associative vs. full reduction semantics against `json_patch::merge`
- **`doc/tests/spill_merge_fuzz.rs`** — QuickCheck tests for merge behavior across spill/memtable boundaries
- **`doc/src/reduce/strategy.rs`** — Unit tests for RFC 7396 merge-patch examples, null deletion in both associative and full modes, nested object merging, and type-switching edge cases
- **`doc/src/combine/memtable.rs`** — Snapshot tests for merge-patch reduction sequences through the combiner, including hold-back semantics and non-associative stacking

This PR adds coverage specifically for the RocksDB batched merge path, which sits above the `doc` reduction layer:

- [x] `test_null_deletion_survives_batching` — targeted regression: null markers must survive batch boundaries and correctly delete properties when re-merged with the base
- [x] `test_merge_many_operands_batched` — integration test through a real RocksDB instance with 50k operands
- [x] `fuzz_batched_merge_matches_reference` — QuickCheck (1000 cases): random merge-patch sequences at batch sizes 2, 3, 5, and unbounded, verifying both full-merge and partial-then-full-merge paths match `json_patch::merge`

### DIFF
--- a/crates/runtime/src/rocksdb.rs
+++ b/crates/runtime/src/rocksdb.rs
@@ -283,7 +283,7 @@ fn set_json_schema_merge_operator(opts: &mut rocksdb::Options, schema: &str) -> 
                               initial: Option<&[u8]>,
                               operands: &rocksdb::merge_operator::MergeOperands|
           -> Option<Vec<u8>> {
-        match do_merge(true, initial, key, operands, &schema_1) {
+        match do_merge_rocks(true, initial, key, operands, &schema_1) {
             Ok(ok) => Some(ok),
             Err(err) => {
                 tracing::error!(%err, "error within RocksDB full-merge operator");
@@ -298,7 +298,7 @@ fn set_json_schema_merge_operator(opts: &mut rocksdb::Options, schema: &str) -> 
                                  initial: Option<&[u8]>,
                                  operands: &rocksdb::merge_operator::MergeOperands|
           -> Option<Vec<u8>> {
-        match do_merge(false, initial, key, operands, &schema_2) {
+        match do_merge_rocks(false, initial, key, operands, &schema_2) {
             Ok(ok) => Some(ok),
             Err(err) => {
                 tracing::error!(%err, "error within RocksDB partial-merge operator");
@@ -314,16 +314,13 @@ fn set_json_schema_merge_operator(opts: &mut rocksdb::Options, schema: &str) -> 
     Ok(())
 }
 
-fn do_merge(
+fn do_merge_rocks(
     full: bool,
     initial: Option<&[u8]>,
     key: &[u8],
     operands: &rocksdb::merge_operator::MergeOperands,
     schema: &str,
 ) -> anyhow::Result<Vec<u8>> {
-    // Holds outputs of prior iterations, which are now inputs of the next one.
-    let mut input_storage: Vec<Vec<u8>>;
-
     // Collect all input documents (initial + operands, each newline-delimited).
     let mut inputs: Vec<&[u8]> = Vec::new();
     if let Some(initial) = initial {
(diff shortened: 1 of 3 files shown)


williamhbaker (approved):
LGTM

crates/runtime/src/rocksdb.rs
```diff
@@ -316,73 +314,237 @@ fn set_json_schema_merge_operator(opts: &mut rocksdb::Options, schema: &str) ->
...
+        // Stop when we have a single batch output.
+        if outputs.len() <= 1 {
+            return Ok(outputs.into_iter().next().unwrap_or_default());
+        }
+
+        // Batch count didn't decrease: double thresholds to make progress.
+        // Bail if both thresholds are already at their caps — we've exhausted
+        // our escalation budget and still can't converge.
+        if outputs.len() >= prev_batches {
+            if batch_byte_target >= MAX_BYTE_THRESHOLD && batch_op_target >= 1_024 * 1_024 {
```
williamhbaker, line 407: Might be worth noting that this `batch_op_target` is only applicable in testing.

williamhbaker: Also there's a very minor change here that I wanted to confirm is intentional: Previously we were using `mb_threshold > 1024` instead of `>=`, so we'd be a little more aggressive on memory usage. I assume this is just being a little more conservative here?

crates/runtime/src/rocksdb.rs
```diff
@@ -400,4 +562,232 @@ mod test {
...
+                    do_merge(true, key, final_inputs, &schema, usize::MAX, usize::MAX).unwrap();
+                if serde_json::from_slice::<serde_json::Value>(&result).unwrap() != expected {
+                    return false;
+                }
+            }
+            true
+        }
+
+        quickcheck::QuickCheck::new()
+            .tests(1000)
```
williamhbaker, line 790: Any concerns with test execution time with this many repetitions? It's probably really fast even with this many, but I know there is some overhead from constructing a validator and schema bundle as well.

crates/runtime/src/rocksdb.rs
```diff
@@ -316,73 +314,237 @@ fn set_json_schema_merge_operator(opts: &mut rocksdb::Options, schema: &str) ->
...
+        }
+
+        // Iterate again with batch outputs as new inputs.
+        // The first output descends from the original base value, so
+        // `is_first_batch` will correctly be `true` for it on the next iteration.
+        prev_batches = outputs.len();
+        input_storage = outputs;
+        inputs = input_storage
+            .iter()
+            .flat_map(|batch| batch.split(|c| *c == b'\n'))
```
williamhbaker, line 428: It took me a moment to realize that each batch output can contain multiple newline-separated documents (one per distinct key in the combiner), so the `flat_map(split('\n'))` here is re-exploding them into individual docs for the next iteration. A brief comment explaining that might help future readers — at first glance it looks like each batch output should just be one document.

williamhbaker (commented):
A few non-blocking comments that might be worth consideration below:

crates/runtime/src/rocksdb.rs
jgraettinger: The semantics are unchanged -- there's also a re-ordering of the test that also happened (to _before_ doubling), so while we'd previously:
* do a round at 1GB
* double the threshold
* fail this check

we now:
* double the threshold 512MB => 1GB
* do a round
* fail this check

crates/runtime/src/rocksdb.rs
jgraettinger: Yep, it's reasonable -- nextest says 1-2s on a debug build.

crates/runtime/src/rocksdb.rs
jgraettinger: >  can contain multiple newline-separated documents (one per distinct key in the combiner)

Just to clarify, there's only one actual key in use (`connector-state` IIRC). The multiple values here are reduction operands that were held back from combine because they're not associative.

PR #2665, opened 2026-02-06
Phil/storage mapping republish

psFried:
**Description:**

Rolls up a few different related things:

First, we introduce a new `background_publication_failed` alert type, which triggers when controller-initiated publications fail. Then we add the capability to trigger background publications of all the specs under a given prefix, and alter the `updateStorageMapping` mutation to do so after a successful modification to a storage mapping.

Recommend reviewing commit by commit.

**Workflow steps:**

This doesn't have direct impact on existing workflows, except that users will start to get notifications for issues that were previously unreported.

I took a stab at adding a section on the new alert type. I'd appreciate any feedback or updates from @aeluce 
https://docs.estuary.dev/reference/notifications/

### DIFF
--- a/site/docs/guides/dashboard/notifications.md
+++ b/site/docs/guides/dashboard/notifications.md
@@ -37,6 +37,19 @@ You can check the current status of the connector in your dashboard.
 
 Additional details about the failure will be available in the connector's `Alerts` tab.
 
+## Background Publication Failed Alerts
+
+Triggers when an automated background process needs to publish a spec, but is unable to because of publication errors. Background publications are performed on all specs for a variety of reasons. For example, updating inferred schemas, or updating materialization bindings to match the source capture. When these publications fail, tasks are likely to stop functioning correctly until the issue can be addressed.
+
+There are many different reasons why publications might fail, but some common scenarios are:
+
+- A network error or misconfiguration between the connector and your source or destination system
+- The credentials for connecting to a source or destination system have been changed or revoked
+- A materialization requires a specific column, which no longer appears in the source collection schema
+- There was an incompatible change to the data type of a particular field (like `string -> boolean`) and the materialization has `onIncompatibleSchemaChange: abort`
+
+A recommended troubleshooting step is to try to publish the spec yourself, resolving any validation errors you encounter along the way. Once you're able to publish the spec, our background automation should be able to as well.
+
 ## Billing Alerts
 
 Billing alerts are automatically subscribed to when a user inputs their email into the `Organization Notifications` table. Alerts will be sent out for the following events:


crates/agent/src/controllers/publication_status.rs
```diff
@@ -218,16 +239,16 @@ impl PendingPublication {
...
             Ok(r) => {
                 record_result(status, pub_info(r));
-                if r.status.is_success() {
-                    control_plane
-                        .notify_dependents(state.live_spec_id)
-                        .await
-                        .context("notifying dependents after successful publication")?;
-                    status.max_observed_pub_id = r.pub_id;
-                }
+                control_plane
```
jshearer: This seems notable: I haven't yet wrapped my head around exactly when this codepath gets hit, but why the change to notifying dependents unconditionally now vs only upon success before? Is the "successful publication" condition enforced elsewhere now? If not, it seems like the `.context()` string below should be updated at least.

site/docs/guides/dashboard/notifications.md
```diff
@@ -37,6 +37,12 @@ You can check the current status of the connector in your dashboard.
 
 Additional details about the failure will be available in the connector's `Alerts` tab.
 
+## Background Publication Failed Alerts
+
+Triggers when an automated background process needs to publish a spec, but is unable to because of publication errors. Background publications are peformed on all specs for a variety of reasons. For example, updating inferred schemas, or updating materialization bindings to match the source capture. When these publications fail, tasks are likely to stop functioning correctly until the issue can be addressed.
```
aeluce:
```suggestion
Triggers when an automated background process needs to publish a spec, but is unable to because of publication errors. Background publications are performed on all specs for a variety of reasons. For example, updating inferred schemas, or updating materialization bindings to match the source capture. When these publications fail, tasks are likely to stop functioning correctly until the issue can be addressed.
```

site/docs/guides/dashboard/notifications.md
```diff
@@ -37,6 +37,12 @@ You can check the current status of the connector in your dashboard.
...
+
+A recommended troubleshooting step is to try to publish the spec yourself, resolving any validation errors you encounter along the way. Once you're able to publish the spec, our background automation should be able to as well.
```
aeluce, line 51:
```suggestion
A recommended troubleshooting step is to try to [publish the spec yourself](/guides/flowctl/edit-specification-locally/), resolving any validation errors you encounter along the way. Once you're able to publish the spec, our background automation should be able to as well.
```

aeluce: It could be helpful to link to instructions on working with/publishing the spec here.

I don't know that we have anything about handling validation errors in the spec elsewhere right now, so it could also be helpful to provide a brief example on what users should be looking for or what they might do to address the issue.

crates/agent/src/controllers/publication_status.rs
```diff
@@ -244,31 +265,88 @@ impl PendingPublication {
...
+        // Do we need to fire or resolve an alert?
+        let Some(alert_state) = alerts else {
+            return result;
+        };
+
+        if publish_successful {
+            resolve_alert(alert_state, AlertType::BackgroundPublicationFailed);
+        } else {
+            let last_entry = &status.history[0];
+            if last_entry.count >= ALERT_AFTER_PUBLICATION_FAILURES {
```
jshearer, line 283: I went digging for where `.count` comes from/is incremented, and I found it here: https://github.com/estuary/flow/blob/39e1f9e44627f2c17705c9da0f0c5cc61dfac738/crates/models/src/status/publications.rs#L95-L108

Notably, because that checks for equality between the `result`s, it will fail to reduce (and thus increment the count, and thus ultimately trigger an alert) if the `result`s are not the same. Following the chain a bit deeper, `result` is an enum: https://github.com/estuary/flow/blob/39e1f9e44627f2c17705c9da0f0c5cc61dfac738/crates/agent/src/connector_tags.rs#L17-L29

So... is my interpretation correct that a publication can fail >3 times in a row without triggering an alert, so long as it fails differently each time? I guess the assumption is just that that won't happen? 

I wonder if, for example, the `ValidationError` inside `JobStatus::ValidationFailed` contains something like a timestamp or something else that would make any two `ValidationError`s trivially non-equal. Just thinking of any possibility here.

OTOH, I could have gotten this chain all wrong and this isn't relevant at all

crates/control-plane-api/src/controllers.rs
```diff
@@ -5,6 +5,34 @@ use serde_json::value::RawValue;
...
+pub enum Message {
+    /// A dependency of the controlled spec has been updated.
+    DependencyUpdated,
+    /// The controlled spec has just been published.
+    SpecPublished {
+        /// The ID of the publication that touched or modified the spec.
+        pub_id: models::Id,
+    },
+    /// Signals that a publication of the spec is necessary, and should be performed
+    /// as soon as practical after the controller recieves this message.
```
jshearer: nit: `recieves` => `receives`

crates/flow-client/control-plane-api.graphql
```diff
@@ -126,13 +126,51 @@ input AlertSubscriptionsBy {
...
+	"""
+	Triggers after repeated task failures have been observed. The task may or may not
+	continue to make progress in between failures, but at a minimum, performance will
+	be degraded. And in many scenarios, the task will be unable to process data at all.
+	"""
 	shard_failed
+	"""
+	Triggers when an automated background process needs to publish a spec,
+	but is unable to because of publication errors. Background publications
+	are peformed on all specs for a variety of reasons. For example,
```
jshearer, line 168: nit: `peformed`=>`performed`

crates/control-plane-api/src/controllers.rs
```diff
@@ -95,6 +123,38 @@ pub async fn update_status(
...
+    message: T,
+    db: &mut sqlx::PgConnection,
+) -> anyhow::Result<i64> {
+    let message_json = serde_json::to_value(message)?;
+    let result = sqlx::query_scalar!(r#"
+        with ids as (
+          select controller_task_id
+          from live_specs
+          where catalog_name::text ^@ $1
+          and coalesce(spec->'shards'->>'disable', 'false') = 'false'
```
jshearer, line 144: Took me a sec to figure out how this was describing the "ignore disabled captures or materializations but not derivations" filter, and eventually realized that derivations are disabled with `spec.derive.shards.disable`, not `spec.shards.disable`. No action here, just confirming that's how it works?

crates/agent/src/controllers/publication_status.rs
psFried: Good catch! Technically, the call won't hurt anything if the publication had failed, but it was definitely intended that we avoid unnecessary wake ups of dependents in that case. Fixed

crates/control-plane-api/src/controllers.rs
psFried: yeah, that's right, I'll update that comment

crates/agent/src/controllers/publication_status.rs
psFried: Yeah, that's right. The `JobStatus::ValidationFailed` you referenced is from `connector_tags::JobStatus`, which is different from `publications::JobStatus`. The `publications` one does not embed the actual validation errors at all, so should still reduce fine. One case where it would still fail to alert is if we were to bounce back and forth between `PublicationFailed` and `BuildFailed` errors. The `PublicationFailed` usually means an error talking with postgres, for example, which indicates a problem with control plane (and we should probably be alerting on). So thing this is probably OK for now that we wouldn't alert on that case.

But I think that this is still something that could be modeled better and more explicitly. There's a similar situation in [this function](https://github.com/estuary/flow/blob/master/crates/agent/src/controllers/mod.rs#L456), which is trying to determine a failure count for a certain type of publication based on matching the detail message.

We got here because I'd initially thought that we'd always only be looking at the most recent entry in the publication history in order to determine the failure count, but that turned out to not be workable. For example, if a "dependency update" publication fails to publish a capture, then it's possible that an auto-discover publication is necessary to get it working again. Or if an auto-discover publication fails, then we'd still want to attempt a pending "config update" publication because that might fix the auto-discover problem.

So I think a more desirable data model would be to have an explicit machine-readable identifier that corresponds to the reason for the publication (i.e. `auto_discover`, `dependency_update`, etc), and to use that both when reducing publication history entries, and when determining previous failure counts. This seems like a refactor that ought to go in a separate PR though.

site/docs/guides/dashboard/notifications.md
psFried: Good point about the errors... I'm not quite sure what to do about that. There's a pretty good variety of different publication/validation errors that users might run into, and I don't think I could enumerate them all. But having at least a few common examples seems useful. I added a few bullet point examples for now. LMK what you think of that

aeluce: This is great, thanks!

PR #2667, opened 2026-02-09
docs: document NetSuite sync modes, delete handling, and table types

jwhartley:
## Summary
- Add "Sync modes and data loading" sections to both NetSuite connector docs (SuiteAnalytics and SuiteQL)
- Explain how incremental vs full refresh is determined (presence of `log_cursor` / `lastmodifieddate`)
- Document delete handling via NetSuite's `DeletedRecord` system table, including which table types support it (base tables) and which don't (linking/junction tables)
- Clarify the distinction between paginated backfill, snapshot backfill, and scheduled full refresh
- Add tip for switching a table from full refresh to incremental

## Context
Customer conversations revealed that none of this behavior was documented. The docs previously listed configuration properties but didn't explain the resulting sync behavior, which tables get deletes, or how to check/change sync modes.

## Test plan
- [ ] Verify Docusaurus renders the new sections correctly (headings, tip box, internal links)
- [ ] Joseph to review for technical accuracy against connector implementation\


site/docs/reference/Connectors/capture-connectors/netsuite-suiteanalytics.md
```diff
@@ -16,6 +16,45 @@ However, if you don't have SuiteAnalytics, see the [SuiteQL connector](./netsuit
 
 Estuary discovers all of the tables to which you grant access during [setup](#setup), including `Transactions`, `Reports`, `Lists`, and `Setup`.
 
+## Sync modes and data loading
+
+Each table binding uses one of two sync modes, determined by the cursor fields available on the table.
```
jshearer: will come back after reading the whole thing, but... I would say the connector has at least 3 sync modes: snapshots, periodic backfills, and incremental replication.

site/docs/reference/Connectors/capture-connectors/netsuite-suiteanalytics.md
```diff
@@ -16,6 +16,45 @@ However, if you don't have SuiteAnalytics, see the [SuiteQL connector](./netsuit
...
+
+Tables with a date-time column suitable for change tracking (configured via [`log_cursor`](#bindings)) are synced **incrementally**. Most base record tables — such as `Transaction`, `Account`, `Customer`, `Employee`, `Item`, `Subsidiary`, and `Vendor` — have a `lastmodifieddate` column that the connector discovers automatically.
+
+During incremental sync, the connector queries only for rows modified since the last checkpoint. The polling frequency is controlled by the [`interval`](#bindings) setting (default: 1 hour).
+
+**How to tell if a table is incremental:** Check the binding's `log_cursor` field. If it's set (e.g., `lastmodifieddate`), the table is incremental.
+
+### Full refresh
+
+Tables without a `log_cursor` are synced via **full refresh** — the connector re-reads the entire table on each sync. This is common for linking and junction tables (e.g., `TransactionLine`, `NextTransactionLineLink`) that lack a `lastmodifieddate` column.
```
jshearer: https://github.com/estuary/source-netsuite/pull/193 will make these referenced tables out of date

site/docs/reference/Connectors/capture-connectors/netsuite-suiteanalytics.md
```diff
@@ -16,6 +16,45 @@ However, if you don't have SuiteAnalytics, see the [SuiteQL connector](./netsuit
...
+
+There are two full refresh modes:
+
+- **Paginated backfill** — Uses the [`page_cursor`](#bindings) to read the table in ordered pages. Pair this with a [`schedule`](#setting-a-schedule) cron expression to control how often the full refresh runs (e.g., daily).
+- **Snapshot backfill** — Set [`snapshot_backfill: true`](#bindings) when no good page cursor exists and the table is small enough for a single query. Snapshot mode manages its own schedule via the `interval` field. Do **not** combine `snapshot_backfill` with a cron `schedule` — this will cause issues with delete emission.
```
jshearer: It's probably worth noting that periodic snapshots using `interval` are capable of capturing deletions on tables that are not covered by `DeletedRecord`

site/docs/reference/Connectors/capture-connectors/netsuite-suiteanalytics.md
```diff
@@ -16,6 +16,45 @@ However, if you don't have SuiteAnalytics, see the [SuiteQL connector](./netsuit
...
+
+**Base record tables** (such as `Transaction`, `Account`, `Customer`, `Subsidiary`, `Currency`, `CurrencyRate`, `Department`, `Employee`, `Item`, `Vendor`, etc.) support delete tracking. When a record is deleted in NetSuite, it appears in the `DeletedRecord` table, and Estuary emits a deletion event to the corresponding collection.
+
+**Linking and junction tables** (such as `TransactionLine`, `TransactionAccountingLine`, `NextTransactionLineLink`, `NextTransactionAccountingLineLink`) do **not** support delete tracking via `DeletedRecord`. NetSuite does not track deletions for these table types. If a linking table is configured for full refresh (snapshot mode), rows that no longer appear in the source are detected as removals during the next full read.
+
+### Table associations
+
+Linking tables can optionally be loaded as **associations** of a parent table instead of — or in addition to — standalone bindings. For example, `TransactionAccountingLine` can be associated with `Transaction` so that when a transaction is modified, its related accounting lines are also loaded.
```
jshearer: I... have been somewhat hesitant to actually talk about this feature because of the major footgun that unless every change in the child table is reflected in an update to the parent table's last-modified cursor (which is almost never the case), using this feature will only appear to work, but actually miss a bunch of change events.

site/docs/reference/Connectors/capture-connectors/netsuite-suiteql.md
```diff
@@ -23,6 +23,26 @@ Currently, this connector supports a subset of NetSuite tables, such as:
 If you need to capture a table that is not yet supported, [contact support](mailto:support@estuary.dev) with the table name(s).
 Estuary support will be able to confirm availability and, if needed, add the table(s) to the connector.
 
+## Sync modes and data loading
```
jshearer, line 26: SuiteQL does not support incremental sync, deletion tracking (except via snapshots, those still capture deletes via `_meta/row_id`), associations, etc. This was an intentional decision we made to simplify the implementation given the constraints. Here's a comparison table that I had Claude put together. Are you in the netsuite connector repo? We should probably get you in there so you (or your agents ;)) can poke around.

---

## Fundamental SuiteQL API Constraints That Drove the Design

These four API-level constraints shaped every design decision:

1. **No table metadata introspection** - SuiteQL has no equivalent of `OA_TABLES`/`OA_COLUMNS`/`OA_FKEYS`. You cannot programmatically discover schemas, column types, primary keys, or foreign keys. The connector requires manual table/key specification via `EndpointConfig.tables`, with defaults from a small `hints.py` (19 lines, 10 tables).

2. **100-column silent failure** - SuiteQL silently returns zero results if a table exceeds 100 columns. Users can specify explicit `columns` lists, or use `SELECT *` with the understanding it fails silently on wide tables.

3. **100k row result limit** - SuiteQL caps results at 100,000 rows per query. The connector uses rownum-based subquery pagination (`wrap_in_paged_subquery`), but each subsequent page re-queries already-seen data, making it progressively slower. Practical ceiling is low single-digit millions of rows.

4. **Datetime information loss** - SuiteQL returns datetime columns as date-only strings. Previous attempts to use `TO_CHAR()` conversions caused timezone mismatch bugs, and including many date-to-datetime conversions in a single query also causes failures. The new connector accepts this and returns data as-is, without hour/minute/second information.

## Features SuiteQL Does NOT Support vs SuiteAnalytics

| Capability | SuiteAnalytics | SuiteQL |
|---|---|---|
| **Incremental replication** | Yes (via `lastmodifieddate` cursor fields) | No. Full refreshes only. |
| **Automatic schema discovery** | Yes (queries `OA_COLUMNS` with types, PKs, FKs) | No. Manual table/key config only. |
| **Deletion tracking** | Yes (`DeletedRecord` table + tombstones) | No. Full refresh overwrites. |
| **Associated document loading** | Yes (child rows loaded with parent changes) | No. |
| **Maximum columns supported** | 1,000 | 100 |
| **Concurrent backfill chunks** | Yes (parallel cursor-range splitting) | No. Single-threaded only. |
| **Typed schema models** | Yes (full Pydantic models with correct types) | No. All fields are passed through from the API response, relies on schema inference. |
| **Boolean normalization** | Yes (`T`/`F` to `true`/`false`) | No. Data as-is. |
| **Datetime with timezone** | Yes (UTC assumption, proper datetime types) | No. Dates only. |

jshearer (changes requested):
Mainly the SuiteQL information here is factually incorrect. If you don't already have access to the netsuite repo just make an access request and Mike or I can add you, there's a lot of valuable data in there about the history of the connectors, decisions we made in PR/commit messages etc.

jwhartley:
All review feedback addressed in commits eaede1f, 66ce4fe, and 745263d. Inline replies on each thread below; high-level summary:

SuiteAnalytics
- "Sync modes" rewritten as three modes (incremental / paginated backfill / snapshot)
- Linking-table examples updated for source-netsuite#193 — `TransactionLine`, `NextTransactionLineLink`, `PreviousTransactionLineLink`, `NextTransactionAccountingLineLink`, and `PreviousTransactionAccountingLineLink` are listed as incremental; `TransactionAccountingLine` is the paginated-backfill example
- Snapshot row-id diffing documented as the delete mechanism for tables not in `DeletedRecord`
- Delete-handling section now calls out that the `DeletedRecord` binding must itself be enabled, and that paginated-backfill bindings can't auto-emit deletes
- Dropped the dedicated "Table associations" prose section given the parent-cursor footgun; the `/associations` properties stay in the bindings reference
- Bumped sample image tag `v3` → `v4` and trimmed the duplicated "Setting a Schedule" section
- Fixed the `exclude_*` defaults in the Advanced Options table to reflect what discovery actually configures (calculated/custom/hidden = `true` for new bindings per source-netsuite#181; `non_display` = `false` because too many useful columns are flagged non-display in NetSuite metadata)

SuiteQL
- Replaced the incorrect incremental / `DeletedRecord` / associations claims; the page now states up front that the connector is full-refresh-only by design and points users to SuiteAnalytics if they need incremental
- New "API constraints" section covers the four hard SuiteQL limits
- Snapshot-mode delete inference via `/_meta/row_id` documented; paginated-backfill deletes flagged as customer-managed
- Fixed `/interval` and `/schedule` defaults in the bindings table to match what discovery actually configures (snapshots: `P1D` interval, empty schedule; paginated: ignored interval, `0 0 * * *` schedule)

Each claim verified against `source_netsuite/{resources,hints,models}.py`, `source_netsuite_suiteql/{resource,hints,models,query}.py`, and CLAUDE.md.

site/docs/reference/Connectors/capture-connectors/netsuite-suiteanalytics.md
jwhartley: Done — rewrote the section to describe three modes (incremental, paginated backfill, snapshot) and clarified that the connector picks one automatically during discovery from the cursor and key columns it finds.

site/docs/reference/Connectors/capture-connectors/netsuite-suiteanalytics.md
jwhartley: Updated. `TransactionLine` and the four `*LineLink` / `*AccountingLineLink` tables are now listed as incremental examples (with `lastmodifieddate` or `linelastmodifieddate` cursors per hints.py). `TransactionAccountingLine` is now the example for paginated backfill — it's still log-cursor-less in hints.py, so it should be a stable example.

site/docs/reference/Connectors/capture-connectors/netsuite-suiteanalytics.md
jwhartley: Added. The Snapshot section now spells out that snapshots diff each run's `/_meta/row_id` set against the previous run's and emit deletions for any rows that disappear. The Delete-handling section calls this out explicitly as the only built-in mechanism for tables not covered by `DeletedRecord`, and notes that paginated-backfill bindings (e.g., `transactionLine`) can't infer deletions on their own — those need a downstream cleanup query.

site/docs/reference/Connectors/capture-connectors/netsuite-suiteanalytics.md
jwhartley: Fair concern. I dropped the dedicated prose walkthrough that framed associations as a recommended pattern. The `/associations` properties remain in the bindings reference for users who already configure them intentionally, but there's no longer anything in the page steering new customers toward the feature. Happy to remove the property rows entirely if you'd rather hide it from the docs altogether.

site/docs/reference/Connectors/capture-connectors/netsuite-suiteql.md
jwhartley: Rewrote the SuiteQL section completely. It now opens by stating that the connector is full-refresh-only by design and routes users to SuiteAnalytics for incremental needs. Added an "API constraints" section covering the four hard SuiteQL limits from your comparison table (no metadata introspection, 100-column silent failure, 100k-row cap, date-only timestamps) so the design tradeoffs are explicit. Snapshot-mode delete inference via `/_meta/row_id` is documented; paginated-backfill deletes are explicitly flagged as customer-managed.

Also fixed the `/interval` and `/schedule` defaults in the bindings table to match discovery reality — snapshots use `P1D` interval with no schedule, paginated bindings ignore `interval` and run on `0 0 * * *`. The previous `PT1H` interval default was the Pydantic-level default that no real binding actually uses.

site/docs/reference/Connectors/capture-connectors/netsuite-suiteanalytics.md
```diff
@@ -16,6 +16,39 @@ However, if you don't have SuiteAnalytics, see the [SuiteQL connector](./netsuit
...
+
+Estuary captures deletions in two ways:
+
+- **Via the `DeletedRecord` system table.** For base record tables (such as `Transaction`, `Account`, `Customer`, `Subsidiary`, `Currency`, `Department`, `Employee`, `Item`, and `Vendor`), Estuary reads NetSuite's `DeletedRecord` system table and emits a deletion event to the corresponding collection for each row that appears there. **The `DeletedRecord` binding must itself be enabled in your capture for these events to be emitted.** NetSuite does not record deletions in `DeletedRecord` for many linking and junction tables (notably `transactionLine` and `TransactionAccountingLine`).
+- **Via snapshot diffing.** Snapshot bindings detect deletions by comparing each run's `/_meta/row_id` set to the previous run's. Any row that disappears is emitted as a deletion. This is the only built-in mechanism for capturing deletions on tables that aren't covered by `DeletedRecord`. For paginated-backfill bindings on those tables (for example, `transactionLine`), the connector cannot infer deletions on its own — schedule periodic backfills and run a downstream cleanup query that removes rows older than the most recent backfill start time.
```
jshearer: > For paginated-backfill bindings on those tables (for example, `transactionLine`), the connector cannot infer deletions on its own — schedule periodic backfills and run a downstream cleanup query that removes rows older than the most recent backfill start time.

We have talked about this solution internally (and may actually have one or two customers actually using it), and it _does_ work, but IMHO it's kind of a rube goldberg machine and probably only really useful if it's the last thing needed to get a contract over the finish line.

The real feature here is first-party support for truncations, which is on the roadmap but not built yet. Personally, I wouldn't even mention the "downstream cleanup query" here.... but that's somewhat of a taste decision, so up to you, not blocking.

site/docs/reference/Connectors/capture-connectors/netsuite-suiteql.md
```diff
@@ -23,6 +23,35 @@ Currently, this connector supports a subset of NetSuite tables, such as:
...
+
+## API constraints
+
+SuiteQL has several hard API limits that shape how the connector operates. Plan your bindings with these in mind:
+
+- **No metadata introspection.** Unlike the SuiteAnalytics ODBC driver, SuiteQL has no way to programmatically discover table schemas, primary keys, or foreign keys. The connector relies on a static list of supported tables and keys; if you need a table that isn't supported, [contact support](mailto:support@estuary.dev).
+- **100-column limit.** SuiteQL silently returns zero rows for any query that selects more than 100 columns. To capture wide tables, set the binding's [`columns`](#bindings) field to an explicit list of 100 or fewer columns.
+- **100,000-row result limit.** SuiteQL caps query results at 100,000 rows. The connector works around this with paginated subqueries, but each successive page re-scans previously read rows, so very large tables (tens of millions of rows or more) become impractical.
```
jshearer:
```suggestion
- **100,000-row result limit.** SuiteQL caps query results at 100,000 rows. For tables with keys, this is not a practical concern: the connector pages through results using the key's sort order, and each page fetches only unread rows. For tables without keys on the other hand, the connector uses a limit-offset strategy. Each subquery must re-examine the full table scan to locate its window of rows, so performance degrades as the table grows. Snapshot bindings are best suited to small reference/lookup tables.
```

Or something like that. The 100k result limit isn't relevant for backfilling with a key, just snapshots, and just because it forces us to use rownum pagination.

site/docs/reference/Connectors/capture-connectors/netsuite-suiteql.md
```diff
@@ -23,6 +23,35 @@ Currently, this connector supports a subset of NetSuite tables, such as:
...
+The SuiteQL connector is **full-refresh only** — it does not support incremental change capture. The SuiteQL REST API does not expose the metadata required for incremental replication (no equivalent of SuiteAnalytics' `OA_TABLES`/`OA_COLUMNS`/`OA_FKEYS`), so each binding is re-read from scratch on every run. Use the [`schedule`](#bindings) cron expression to control how often a binding re-runs and picks up new and changed rows.
+
+If incremental sync is required, use the [SuiteAnalytics connector](./netsuite-suiteanalytics.md) instead.
+
+### Paginated backfill vs. snapshot
+
+The connector picks a mode per binding based on whether the table has a configured key:
+
+- **Paginated backfill** — Tables with a key (specified in the endpoint's `tables` config) are read in ordered pages using a `page_cursor` (defaults to the first key field). Each scheduled run starts a new full backfill.
+- **Snapshot** — Tables with no key (for example, `Account`, `DeletedRecord`, `transactionHistory`, `TransactionStatus`) are read as a single query that returns the entire table. Snapshots run on the binding's `interval` (defaulting to once a day) and use `/_meta/row_id` as the collection key.
```
jshearer: tbh `Account` does have a key, it just isn't hinted properly. we should fix this in the connector's hints

jshearer (approved):
This looks good to me @jwhartley, thanks for the update! I had a couple of small suggestions/comments, but all within the realm of your discretion.

site/docs/reference/Connectors/capture-connectors/netsuite-suiteanalytics.md
jwhartley: Kept the cleanup-query workaround and added a sentence noting that first-party truncation support is on the roadmap and would automate this end-to-end. Pushed in ea68515.

site/docs/reference/Connectors/capture-connectors/netsuite-suiteql.md
jwhartley: Applied your suggestion verbatim — pushed in ea68515.

site/docs/reference/Connectors/capture-connectors/netsuite-suiteql.md
jwhartley: Agreed — dropped `Account` from the snapshot example list in the doc (ea68515) so we're not modeling it as keyless going forward.

Looking at the connector, your `jshearer/suiteql_table_analysis` branch (commit 93e4857, "expand table coverage from 10 to 50 with verified keys") already flips `Account` from snapshot to cursor mode with a verified `id` key. Are you planning to land that branch soon? If so I'll leave the hint as-is and let it carry the fix; if it's blocked or going to sit a while, happy to put up a smaller targeted PR for just the `Account` hint to get the immediate bug fixed.

PR #2668, opened 2026-02-09
assemble: add name:prefix to collection journal selector

jgraettinger:
**Description:**

Finally enable the long-commented-out `name:prefix` attached to build partition selectors.
We have handling in Dekaf and the Go runtime which unconditionally set this label value after-the-fact (such as in Dekaf's `update_partition_info`). We can roll those back once this is deployed and all catalog specs have been rebuilt.


jgraettinger:
Ping

PR #2669, opened 2026-02-10
Block redact strategies on /_meta and /_meta/uuid in collection schemas

jgraettinger:
**Description:**

Redacting /_meta or /_meta/uuid would destroy the UUID placeholder that the runtime depends on. This adds a validation check that rejects schemas with redact strategies on those locations, and hardens the Go runtime so that a missing UUID placeholder is recovered as an error (with a logged stack trace) rather than crashing the process.

**Workflow steps:**

Attempting to redact `/_meta/` or `/_meta/uuid` now generates a publication error.
At runtime, a former "missing UUID" panic is now an error with a log.

### DIFF
--- a/crates/validation/src/collection.rs
+++ b/crates/validation/src/collection.rs
@@ -180,6 +180,17 @@ fn walk_collection(
         errors,
     )?;
 
+    // Check that /_meta and /_meta/uuid are not redacted, as redaction would
+    // destroy the UUID placeholder that the runtime depends on.
+    if read_model.is_some() {
+        check_uuid_ptr_not_redacted(scope.push_prop("writeSchema"), &write_spec.spec, errors);
+        if let Some(read) = &read_spec {
+            check_uuid_ptr_not_redacted(scope.push_prop("readSchema"), &read.spec, errors);
+        }
+    } else {
+        check_uuid_ptr_not_redacted(scope.push_prop("schema"), &write_spec.spec, errors);
+    }
+
     let effective_read_spec = read_spec
         .as_ref()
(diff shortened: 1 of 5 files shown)


go/flow/mapping.go
```diff
@@ -219,14 +220,37 @@ func (m Mappable) SetUUID(uuid message.UUID) {
...
+			m.Spec.Name,
+		))
 	}
 
 	// Replace it with the string-form UUID.
 	var str = uuid.String()
 	copy(m.Doc[ind:ind+36], str[0:36])
 }
 
+// PublishMappable wraps a publish call (PublishUncommitted or PublishCommitted)
```
williamhbaker: I think it's undeniably a "good thing" to keep and issue from a single task from repeatedly crashing the entire reactor, but I'm feeling like this panic recovery is a bit of a half-measure.

Since the root cause (the redaction of `_meta/uuid`) has been made impossible, we should be able to continue with the invariant that the UUID placeholder is always valid. On the other hand, the extra bit of a defensiveness with the `recover()` here doesn't seem unreasonable. BUT there are other places where we are using `PublishUncommitted` and `PublishCommitted` directly without this wrapper, so it seems inconsistent.

I'd advocate for either A) being confident that we've address this panic condition by effectively making it impossible or B) more strictly enforcing error recovery from `SetUUID` so we use it consistently.

Unless I'm missing something about the specific call sites in `capture.go` and `derive.go` that make us want to be extra careful there?

jgraettinger:
As discussed over VC, I've backed out the panic / recover changes, leaving only the new validation assertions. PTAL.

PR #2670, opened 2026-02-10
docs: Add connection pooling requirement to Neon Postgres capture

jwhartley:
## Summary

- Adds a "Connection Pooling" section to the Neon PostgreSQL capture docs explaining that the CDC connector requires a direct connection (not pooled)
- Adds prerequisite noting direct connection requirement
- Adds inline warning in the connector setup step
- Documents how to identify pooled vs direct hostnames (`-pooler` in hostname)
- Includes the error message users see when using a pooled connection (`syntax error at or near "IDENTIFY_SYSTEM"`)

This is a common support issue — Neon's connection pooler doesn't support the PostgreSQL replication protocol, so captures fail silently or with confusing errors.

## Test plan

- [ ] Verify docs preview renders the caution/warning admonitions correctly
- [ ] Verify anchor link from prerequisites to Connection Pooling section works

### DIFF
--- a/site/docs/reference/Connectors/capture-connectors/PostgreSQL/neon-postgres.md
+++ b/site/docs/reference/Connectors/capture-connectors/PostgreSQL/neon-postgres.md
@@ -107,13 +107,7 @@ To get your direct connection string:
 3. Ensure **Connection pooling** is toggled **off**
 4. Copy the connection string
 
-If you use a pooled connection, the capture will fail because the pooler does not pass through replication commands. You will see an error like:
-
-```
-pq: syntax error at or near "IDENTIFY_SYSTEM"
-```
-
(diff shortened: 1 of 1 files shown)


aeluce (approved):
LGTM. As a minor style note for future, be wary of competing formats in a small amount of space. Info/warning boxes, bolded text, inline code, or other special formatting all together can make it unclear where the reader should focus attention.

Image

Also, if users run into this issue pretty frequently, the connector could potentially provide a more specific error. For example, we check for Supabase connection poolers as well as noting the requirement in the Supabase docs:
https://github.com/estuary/connectors/blob/555b3309d417d412f49dff2c4df16da42fb9ff79/source-postgres/main.go#L262

PR #2671, opened 2026-02-11
docs: Note delta updates incompatibility with formula field refreshes

jwhartley:
## Summary
- Adds a note to the Salesforce Native docs that formula field refreshes emit partial documents which produce `null` values for non-formula fields when used with delta updates materializations
- Adds the equivalent note to the HubSpot Real-Time docs for calculated property refreshes
- This is a frequently asked support question (most recently [Slack thread](https://estuary-dev.slack.com/archives/C04MSMG4QLG/p1770735595773129))

## Test plan
- [ ] Verify doc pages render correctly on preview


site/docs/reference/Connectors/capture-connectors/Salesforce/salesforce-native.md
```diff
@@ -125,6 +125,8 @@ Salesforce objects can contain [formula fields](https://help.salesforce.com/s/ar
 
 To address this challenge, the Salesforce connector is able to refresh the values of formula fields on a schedule after the initial backfill completes. This is controlled at a binding level by the cron expression in the [`schedule` property](#bindings). When a scheduled formula field refresh occurs, the connector fetches every record's current formula field values and merges them into the associated collection with a top-level [`merge` reduction strategy](/reference/reduction-strategies/merge).
 
+Note that formula field refreshes emit partial documents containing only the record's key and formula field values. By default, these are combined with previously captured complete documents, and this works well for standard [updates materializations](/concepts/materialization/#updates-materializations). However, [delta updates materializations](/concepts/materialization/#delta-updates) do not reduce against existing destination state, so partial documents from formula field refreshes are materialized as-is with all non-formula fields as `null`.
```
Alex-Bair: It looks like there isn't an `#updates-materializations` anchor on the `/concepts/materializations` page. The `#how-continuous-materializations-work` anchor describes how documents are reduced, so it makes sense to me to link to that anchor instead.

Also, I'd suggest including `standard` in the link for consistency with the delta updates materializations link.

Alex-Bair: So maybe something like:

```suggestion
Note that formula field refreshes emit partial documents containing only the record's key and formula field values. By default, these are combined with previously captured complete documents, and this works well for [standard updates materializations](/concepts/materialization/#how-continuous-materializations-work). However, [delta updates materializations](/concepts/materialization/#delta-updates) do not fully reduce documents, so partial documents from formula field refreshes are materialized as-is with all non-formula fields as `null`.
```

site/docs/reference/Connectors/capture-connectors/HubSpot-real-time.md
```diff
@@ -93,4 +93,6 @@ captures:
...
\ No newline at end of file
+To address this challenge, the HubSpot connector is able to refresh the values of calculated properties on a schedule after the initial backfill completes. This is controlled at a binding level by the cron expression in the [`schedule` property](#bindings). When a scheduled calculated property refresh occurs, the connector fetches every record's current calculated property values and merges them into the associated collection using [`merge` reduction strategies](/reference/reduction-strategies/merge).
+
+Note that calculated property refreshes emit partial documents containing only the record's key and calculated property values. By default, these are combined with previously captured complete documents, and this works well for standard [updates materializations](/concepts/materialization/#updates-materializations). However, [delta updates materializations](/concepts/materialization/#delta-updates) do not reduce against existing destination state, so partial documents from calculated property refreshes are materialized as-is with all non-calculated properties as `null`.
```
Alex-Bair: See the suggestion on `salesforce-native.md`. It also applies here.

PR #2676, opened 2026-02-13
ops: expanded stats for tracking tasks progress

williamhbaker:
**Description:**

- Accumulate a new "bytes behind" statistic for for materialization bindings and derivation transforms, and roll this up through the reporting pipeline into the `catalog_stats` table.
- Also add a new "last published at" statistic for captures and derived outputs, similar to the existing `lastSourcePublishedAt` stats field, that is written in stats documents based on the timestamp of the latest document published to the collection. Roll this up into the `catalog_stats` table as well, populated for collections.
- Roll up `lastSourcePublishedAt` into `catalog_stats` too, alongside "bytes behind".
- Expose these new stats in the control plane OpenMetrics API.

A nuance of this implementation is that updating of the task progress is driven by the task reading documents for its source journals. They will be stale until more documents are read. The assumption is that for practical purposes this should be fine, since task are actively reading data from their sources. However it may seem a little weird if for example a task has a `notAfter` set and its `bytes_behind` metric never moves because of that.

Manually tested on a local stack. Made sure that everything works as expected: Stats are written with the `lastPublishedAt` for collection writers, and `bytesBehind` for collection readers. They are aggregated / rolled up by the ops catalog a materialized into `catalog_stats` as expected. Also verified that the new gauges are available in the OpenMetrics API.

**Workflow steps:**

TODO: Will need to update the OpenMetrics API docs.

**Notes for reviewers:**

Had both Claude and Codex do a pre-review and addressed a couple of their findings (from Codex, specifically).

### DIFF
--- a/supabase/seed.sql
+++ b/supabase/seed.sql
@@ -125,6 +125,13 @@ begin
 end;
 $$ language plpgsql;
 
+-- Seed the JWT secret into the vault so that refresh token exchange works locally.
+-- This is required for the OpenMetrics API and any other endpoint that uses refresh tokens.
+select vault.create_secret(
+  'super-secret-jwt-token-with-at-least-32-characters-long',
+  'app.jwt_secret'
+);
+
 -- TODO(johnny): Support deprecated gateway_auth_token() RPC to be removed:
 insert into internal.gateway_auth_keys (secret_key, detail) values (
   'supersecret', 'Used for development only. This value will be changed manually when deployed to production.'


williamhbaker (commented):
N/A - Accidental double comment, nothing to see here

go/protocols/ops/ops.proto
```diff
@@ -124,21 +124,30 @@ message Stats {
     uint64 bytes_total = 2;
   }
 
-  // Binding represents counts of JSON documents and their
-  // cumulative total size in bytes, passing through the binding
-  // of a capture or materialization.
-  message Binding {
```
williamhbaker, line 130: Splitting these up felt somewhat dramatic, but it seemed necessary to prevent a singular `Binding` from getting very cumbersome trying to have a different name for a capture's "last published at" vs. a materialization's "last source published at".

go/runtime/derive.go
```diff
@@ -145,6 +153,7 @@ func (d *deriveApp) ConsumeMessage(_ consumer.Shard, env message.Envelope, _ *me
 	var uuid = isr.UuidParts[isr.Index]
 	var keyPacked = isr.Arena.Bytes(isr.PackedKey[isr.Index])
 	var docJson = isr.Arena.Bytes(isr.Docs[isr.Index])
+	d.tracker.observe(env.Journal.Name, isr)
```
williamhbaker: I don't think this is as efficient as it theoretically could be, since we are usually indexing the same `IndexShuffleResponse` for multiple documents, which has the same `readThrough` and `writeHead`. And thinking about further, it's probably not a precise as it could be, since if the Flow transaction boundary doesn't line up exactly with the last document in in the shuffle response, we'll overestimate how far we are in the journal when writing stats.

I've implemented it this way since regarding efficiency, doing an extra map write for every document seemed acceptable. I'm only just thinking about the precision issue now while writing this comment. Perhaps it's alright as an overly-optimistic estimate, but I'm also wondering if we could instead keep track of how many bytes each `ConsumeMessage` document represents from the journal and be more accurate :thinking:.

go/runtime/derive.go
```diff
@@ -175,18 +185,41 @@ func (d *deriveApp) FinalizeTxn(shard consumer.Shard, pub *message.Publisher) er
...
 			}
-			if _, err = pub.PublishUncommitted(mapper.Map, flow.Mappable{
+			var mappable = flow.Mappable{
 				Spec:       d.term.taskSpec,
 				Doc:        response.Published.DocJson,
 				PackedKey:  responseExt.Published.KeyPacked,
 				Partitions: partitions,
 				List:       d.watch,
-			}); err != nil {
+				UUIDOffset: bytes.Index(response.Published.DocJson, pf.DocumentUUIDPlaceholder),
```
williamhbaker: I certainly don't love this mechanism for managing the UUIDs, but it was the best I could come up with to prevent us from having to parse the UUID here for every document written.

ops-catalog/stats.schema.yaml
```diff
@@ -51,6 +55,11 @@ properties:
           format: date-time
           description: |
             The publication timestamp of the most recently processed source document.
+        bytesBehind:
```
williamhbaker, line 58: An alternative thought I had here was to keep the raw `readThrough` and `writeHead` here, still aggregated by collection. Then the rollup derivation would do the `writeHead - readThrough` math to come up with the bytes behind. I couldn't really think of a great motivator for doing this though.

go/runtime/derive.go
williamhbaker: Seems like we could tighten this up by referencing the [offsets](https://github.com/estuary/flow/blob/c7bbaa4fa47037785d39065f0d0e7028b2ffca04/go/protocols/runtime/runtime.pb.go#L260-L263) from the shuffle response, along with the document index.

go/protocols/ops/ops.proto
jgraettinger: :shrug: agreed it's time and there's already precedent (`Transform` as a derivation's `Binding` equivalent).

super nit: Move them to be directly adjacent to where they're used in a `map`, as is the case with `message Derive` and `message Interval`?

go/protocols/ops/stats.go
```diff
@@ -5,22 +5,114 @@ import (
...
+// Shadow struct for Stats_DocsAndBytes with camelCase json tags.
+type jsonDocsAndBytes struct {
 	DocsTotal  uint64 `json:"docsTotal,omitempty"`
 	BytesTotal uint64 `json:"bytesTotal,omitempty"`
 }
 
-// MarshalJSONPB allows us to bypass the protobuf-specific marshalling which would quote uint64
-// fields and instead use regular json encoding which does not apply quotes.
+// Shadow struct for Stats_MaterializeBinding with camelCase json tags.
+type jsonBinding struct {
```
jgraettinger: nit: jsonMaterializeBinding ?

ops-catalog/stats.schema.yaml
jgraettinger: Agreed, I think a `bytesBehind` gauge is cleaner.

ops-catalog/catalog-stats.schema.yaml
```diff
@@ -52,6 +52,35 @@ properties:
...
+        format: date-time
+        reduce:
+          strategy: maximize
+      progress:
+        description: >-
+          Per-source-collection progress for materializations and derivations,
+          tracking bytesBehind and lastSourcePublishedAt.
+        type: object
+        additionalProperties:
+          type: object
```
jgraettinger: What's the motivation for having this live in `progress` as well as under `taskStats` ? Couldn't I query it there?

I'm noting that stats.schema.yaml doesn't have reduction annotations on these locations, but shouldn't it? Other parts do already, such as `$defs/docsAndBytes`

jgraettinger (commented):
I've been going back and forth on this - I think it's done about as well as can be done given the current architecture headwinds, but I'm having trouble choking it down.

I'd like to discuss a proposal: landing the ops protobuf changes, reporting updates, tests, and schema changes, and open_metrics updates, and _maybe_ Go runtime changes that can be done cleanly (if there are any), but holding back the rest. We'd instead focus on getting it implemented properly under the shuffle-v2 implementation -- I think there's a bit of specific work there to support this, but probably not much -- and transitioning sequencing and appends to be done from Rust. Thoughts?

ops-catalog/tests.flow.yaml
```diff
@@ -90,7 +90,10 @@ tests:
                 tenant/test/collection:
```
williamhbaker, line 90: This file is getting pretty difficult to edit/review manually. I had Claude handle adding the `verify` fields and spot-checked. The test pass, and the derivation logic isn't particularly complex, so it seems to be working as intended.

go/protocols/ops/stats.go
williamhbaker: Done

ops-catalog/catalog-stats.schema.yaml
williamhbaker: Agreed, after thinking about it, there's no reason to have a separate & redundant `progress` in `statsSummary`. Removed the `progress` field entirely.

williamhbaker: For the reduction annotations, what I've got now is:
- `lastPublishedAt` in `catalog-stats.schema.yaml`: Uses `maximize`. This is populated for collections, and needs to keep the maximum in cases where there are multiple writers to a collection.
- `lastPublishedAt` and `bytesBehind` in `stats.schema.yaml`: Both don't have a reduction listed, meaning the default to last write wins. I think this is correct for `bytesBehind` because the stats are for a single task as it progresses, so no summing or maximizing/etc. would make sense. `lastPublishedAt` is a little more mind-bending since the collection-level field needs to `maximize`, but this one is for a single task, and its `lastPublishedAt` should always be increasing, so I think the last write wins is correct here too.

go/protocols/ops/ops.proto
williamhbaker: Done

crates/control-plane-api/src/server/public/open_metrics.rs
```diff
@@ -210,8 +228,19 @@ fn encode_metrics(buf: &mut BufferParts, _scrape_at: f64, stats: CatalogStats) {
...
+                    &l_task_collection_transform,
+                    ts,
+                    false,
+                );
             }
+            DERIVED_BYTES_BEHIND.gauge(
+                buf,
+                &l_task_collection_transform,
+                bytes_behind as f64,
+                true,
```
jgraettinger: I think we want to set this and others to `false`, at least for now, for the pragmatic reason of: if we start emitting these as zero, it'll show up in user's DataDog / Prometheus integrations and we'll get support requests about "why is this wrong?"

jgraettinger (approved):
LGTM % one comment about not emitting zero-value lag gauges

crates/control-plane-api/src/server/public/open_metrics.rs
williamhbaker: Good idea, made that update along witha TODO to emit explicit zeros sometime in the future, when stats are widely populated with `bytesBehind`.

PR #2678, opened 2026-02-16
Improve expectPubId mismatch error message

jwhartley:
## Summary
- Rewrites the `ExpectPubIdNotMatched` error to be more human-readable: leads with what happened, gives a clear action, and demotes publication IDs to a parenthetical
- Motivated by customer confusion — the current message is truncated in the UI and the hex IDs are not actionable

**Before:**
```
expected publication ID 0000000000000000 was not matched (it's actually 117efc2f6100029f):
your changes have already been published or another publication has modified this spec;
please try again with a fresh copy of the spec.
```

**After:**
```
This task was updated while you were editing — please refresh the spec and re-apply your changes.
This may have been an automated system update. (expected publication ID 0000000000000000, actual 117efc2f6100029f)
```

## Context
This error commonly occurs when automated system updates (e.g. inferred schema changes, auto-discovers, controller convergence) modify a spec while a user is editing it. The old message was written for developers and doesn't help customers understand what to do.

**Question:** Do inferred schema changes still update `last_pub_id`? The Aug 2024 plan (#1520) outlined separating `last_pub_id` from `last_build_id` so that dependency-only changes wouldn't bump `last_pub_id`. If that's been implemented, this error should be less frequent — but when it does occur, it should still be clear.

## Test plan
- [x] `cargo test -p validation -- test_insert_but_already_exists test_update_but_does_not_exist` passes locally
- [ ] Integration test in `locking_retries.rs` updated to match
- [ ] CI passes

### DIFF
--- a/crates/validation/src/errors.rs
+++ b/crates/validation/src/errors.rs
@@ -267,7 +267,7 @@ pub enum Error {
         larger_id: models::Id,
     },
     #[error(
-        "This task was updated while you were editing — please refresh the spec and re-apply your changes.\nThis may have been an automated system update. (expected publication ID {expect_id}, actual {actual_id})"
+        "This spec was updated while you were editing — please refresh and re-apply your changes.\nThis may have been an automated system update. (expected publication ID {expect_id}, actual {actual_id})"
     )]
     ExpectPubIdNotMatched {
         expect_id: models::Id,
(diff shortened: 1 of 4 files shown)


crates/validation/src/errors.rs
```diff
@@ -267,7 +267,7 @@ pub enum Error {
         larger_id: models::Id,
     },
     #[error(
-        "expected publication ID {expect_id} was not matched (it's actually {actual_id}): your changes have already been published or another publication has modified this spec; please try again with a fresh copy of the spec."
+        "This task was updated while you were editing — please refresh the spec and re-apply your changes.\nThis may have been an automated system update. (expected publication ID {expect_id}, actual {actual_id})"
```
jgraettinger: Use "spec" or "specification", as this error applies to collection as well (which are usually not tasks).

jwhartley: done

PR #2679, opened 2026-02-16
docs: document Additional Table Create SQL for creating indexes

jwhartley:
## Summary

- Adds a usage section to the PostgreSQL materialization docs explaining how to use the `additional_table_create_sql` binding property for creating indexes
- Adds the missing `additional_table_create_sql` property to the Databricks materialization docs bindings table, with a similar usage section
- Includes a `CREATE INDEX` example (Postgres) and notes about when the SQL executes

## Context

This field exists in config reference tables for Postgres-family connectors but the description ("Additional SQL statement(s) to be run in the same transaction that creates the table") doesn't make practical use cases obvious. The only existing usage example is on the TimescaleDB page (for hypertables).

A recent support thread showed this is a common source of confusion — customers aren't sure what the field is for or how to use it for things like creating indexes. Adding concrete examples should reduce repeat questions.

The Databricks connector also implements the field but it was missing from the docs entirely.

## Test plan

- [ ] Verify the Postgres and Databricks pages render correctly on the docs site


aeluce (approved):
LGTM. Did you want to update the Postgres materialization variants as well? If you didn't want to copy the new section to all of them, you could add a link to the binding property directed to the "Additional Table Create SQL" section in the main Postgres doc.

PR #2680, opened 2026-02-16
docs: Move GA4 BigQuery Native to sort with Google connectors

jwhartley:
**Description:**

Renames `bigquery-ga4-native.md` to `google-analytics-4-bigquery-native.md` so the Google Analytics 4 BigQuery Native connector appears in the sidebar next to the other Google Analytics connectors (under "G") instead of being separated under "B".

A prospect couldn't find it because it wasn't listed with the other Google connectors. See [Slack thread from Zulf](https://estuary-dev.slack.com/archives/C04MSMG4QLG).

**Workflow steps:**

No user-facing workflow change. The page content is identical — only the sidebar position changes.

- Old URL `/reference/Connectors/capture-connectors/bigquery-ga4-native/` now redirects to the new path via `plugin-client-redirects`

**Notes for reviewers:**

- The page title ("Google Analytics 4 BigQuery Native Connector") was already correct — only the filename was causing the mis-sort
- A redirect is included so any existing bookmarks or external links continue to work


aeluce (approved):
This works. It's also possible to keep the existing slug instead of using a redirect by adding a tag at the top of the page, such as:

```
---
slug: /reference/Connectors/capture-connectors/bigquery-ga4-native/
---
```

PR #2689, opened 2026-02-17
control-plane: exclude defunct data planes from snapshots

psFried:
We had data planes rows in production that don't have hmac keys or encrypted hmac keys. They represent data planes that are currently unused, since hmac keys are required in order to interact with data planes. So this updates the snapshot refresh logic to filter them out instead of erroring on them.

I had to deal with a handful of flakey tests in order to make this work, by running all the control_plane_api::server tests sequentially. Hopefully that won't slow down our test runs too much.

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -19,3 +19,11 @@ test-group = 'serial-db-tests'
 [[profile.default.overrides]]
 filter = 'package(automations) and test(test_fibonacci_bench)'
 test-group = 'serial-db-tests'
+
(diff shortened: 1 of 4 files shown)


psFried:
Ah, I forgot to mention that. If a data plane is missing those keys, then it can't be used for anything. Any attempt to publish to the plane would have failed anyway. Any authorization requests coming from that data plane would fail after this merges, but there shouldn't be anything running there, and failing the authZ requests seems appropriate if by chance some tasks are still running.

PR #2695, opened 2026-02-19
validation: allow key fields with write-schema defaults despite inferred schema additionalProperties

JustinASmith:
**Description:**

## Summary

- Treats `default` as a pure annotation during schema intersection, preserving it through `allOf` composition even when the intersected type is `INVALID`
- Adds `&& default.is_none()` guards to `PtrCannotExist` checks in `walk_ptr()` so locations with a default bypass the error
- Fixes the source-shopify-native scenario where `/_meta/store` has a write-schema default but the inferred schema hasn't observed it yet

## Problem

When a collection key includes a field (e.g. `/_meta/store`) defined in the write schema with `default: ""`, but the inferred schema has `additionalProperties: false` on the parent object and hasn't observed the field yet, publication fails with:

> location /_meta/store is prohibited from ever existing by the schema

The root cause is in `intersect_default()`: the inferred schema's `additionalProperties: false` imputes a false schema for unobserved properties, and intersecting with the write schema produces `type_: INVALID`. `intersect_default()` then drops the default because its type (`STRING`) doesn't overlap with `INVALID`, even though the default was valid in its originating schema.

## Fix

`default` is an annotation, not a validation keyword. Local validation already rejects mismatched defaults (e.g. `type: integer, default: "hello"`) at the schema that directly contains them. The intersection doesn't need to re-validate.

- **`intersect_default()`**: Replaced with `lhs.default.or(rhs.default)`, matching how `title`, `description`, and `reduce` are intersected
- **`walk_ptr()`**: Added `&& read_shape.default.is_none()` / `&& write_shape.default.is_none()` to `Cannot`-exist checks. When `read_exists` is `Cannot` with a default, skips read-side key checks (type, reduce, redact) since the `INVALID` shape is not meaningful
- **`shape_to_avro()`**: Updated the nullable check to treat a null default as equivalent to no default. When `default` is preserved as `Some(null)` through intersection, the field must still be nullable in the Avro schema since absence maps to null

## Test plan

- [x] Added two scenario tests demonstrating the bug (simple write schema + $ref chain matching source-shopify-native)
- [x] Updated `intersect_default` unit test to reflect annotation semantics
- [x] All 124 doc lib tests pass
- [x] All 75 validation scenario tests pass
- [x] Full validation test suite (124 tests across all test files) passes
- [x] All 10 dekaf field extraction tests pass (including `test_allof_with_null_default`)
- [x] All 2 avro crate tests pass

**Workflow steps:**

**Notes for reviewers:**

### DIFF
--- a/crates/validation/src/schema.rs
+++ b/crates/validation/src/schema.rs
@@ -73,36 +73,40 @@ impl Schema {
             return Ok(());
         }
 
-        // When read_exists is Cannot but has a default, we skipped the
-        // PtrCannotExist error above. The read shape's type is INVALID in
-        // this case, so key-specific checks against it are meaningless.
-        if read_exists != Exists::Cannot {
-            if !read_shape.type_.is_keyable_type() {
-                return Err(Error::KeyWrongType {
-                    ptr: ptr.to_string(),
-                    type_: read_shape.type_,
-                    schema: read.curi.clone(),
-                });
-            }
+        // When the location Cannot exist but has a default, the shape type
+        // is INVALID. Union with the default value's type so the keyable
+        // check passes when the default itself is a keyable type.
+        let default_type = read_shape
+            .default
+            .as_ref()
+            .map(|d| types::Set::for_node(&d.0))
+            .unwrap_or(types::INVALID);
+        let effective_type = read_shape.type_ | default_type;
 
-            if !matches!(
-                read_shape.reduce,
-                shape::Reduce::Unset
-                    | shape::Reduce::Strategy(doc::reduce::Strategy::LastWriteWins(_)),
-            ) {
-                return Err(Error::KeyHasReduce {
-                    ptr: ptr.to_string(),
-                    schema: read.curi.clone(),
-                    strategy: read_shape.reduce.clone(),
-                });
-            }
-            if !matches!(read_shape.redact, shape::Redact::Unset) {
-                return Err(Error::KeyHasRedact {
-                    ptr: ptr.to_string(),
-                    schema: read.curi.clone(),
-                    strategy: read_shape.redact.clone(),
-                });
-            }
+        if !effective_type.is_keyable_type() {
+            return Err(Error::KeyWrongType {
+                ptr: ptr.to_string(),
+                type_: read_shape.type_,
+                schema: read.curi.clone(),
+            });
+        }
+
+        if !matches!(
+            read_shape.reduce,
+            shape::Reduce::Unset | shape::Reduce::Strategy(doc::reduce::Strategy::LastWriteWins(_)),
+        ) {
+            return Err(Error::KeyHasReduce {
+                ptr: ptr.to_string(),
+                schema: read.curi.clone(),
+                strategy: read_shape.reduce.clone(),
+            });
+        }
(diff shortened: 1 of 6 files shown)


crates/validation/src/schema.rs
```diff
@@ -62,46 +64,69 @@ impl Schema {
                 schema: read.curi.clone(),
             });
         } else if read_exists == Exists::Cannot {
-            return Err(Error::PtrCannotExist {
-                ptr: ptr.to_string(),
-                schema: read.curi.clone(),
-            });
+            // A location that Cannot exist in the read schema may still be
```
jgraettinger: This is mixing / mis-understanding a couple of nuanced concepts:

* The runtime does not "fill in" a default value within a captured document.
   * It will, however, use a default annotation when mapping and extracting through a projection (from a JSON pointer to a columnar field) in order to produce a columnar materialization value.

crates/validation/src/schema.rs
```diff
@@ -62,46 +64,69 @@ impl Schema {
...
+            // valid if the write schema provides a default value. The runtime
+            // fills in write-schema defaults, so the field will always be
+            // populated despite the inferred read schema not knowing about it
+            // yet (e.g., a newly-added key field before any data is written).
+            let write_provides_default = write_located
```
jgraettinger: This is mixing checks over the schema vs write schema in an odd way. I'd like to see:

1) Start by rolling back the implementation change of this routine to a clean slate, then

1) extend the `else if read_exists == Exists::Cannot` condition to instead be
  `else if read_exists == Exists::Cannot && read_shape.default.is_none()` 

That allows for a `default` annotation to signal the user's "out" that they are aware it cannot exist and would like to proceed anyway, using the `default` when projecting from the location.

2) Do the same  with the `else if write_exists == Exists::Cannot {` below, but in terms of `write_shape.default`.

jgraettinger: I suspect this ☝️ plan will achieve the behavior that you're after, but if it doesn't, please ping me and let's discuss further

crates/validation/src/schema.rs
```diff
@@ -73,60 +73,68 @@ impl Schema {
...
         }
 
-        if !read_shape.type_.is_keyable_type() {
-            return Err(Error::KeyWrongType {
-                ptr: ptr.to_string(),
-                type_: read_shape.type_,
-                schema: read.curi.clone(),
-            });
-        }
+        // When read_exists is Cannot but has a default, we skipped the
```
jgraettinger: This specific change block should be rolled back -- we don't want to disable checks over `reduce` or `redact`.

Instead, we should adapt the `if !read_shape.type_.is_keyable_type()` guard to "and" with a new check that `default` is none or is not a key-able type.

Put differently: we're relaxing the check to not fire if `default` is set AND is itself a key-able type.

I'd recommend structuring by taking the union of `read_shape.type_` and the projected type of the `default` (where None is INVALID -- IIRC there's an existing routine for mapping Value to its type?), and then testing if that type `.is_keyable_type()`.

jgraettinger (commented):
One more action item below, but the rest looks good :+1:

jgraettinger (approved):
LGTM!

PR #2700, opened 2026-02-20
dekaf: set RecordBatch `partition_leader_epoch` from `binding_backfill_counter`

jshearer:
## Problem

A consumer reading from Dekaf after a collection reset and subsequent session disconnects/reconnects would silently lose offset commits. After the first consumer session committed successfully, subsequent generations' commits were dropped, causing consumers to re-read messages they had already processed. The root cause was a mismatch between two places where Dekaf reports the leader epoch: the `partition_leader_epoch` in RecordBatch headers (hardcoded to `1`) and the `committed_leader_epoch` in OffsetFetch responses (derived from `binding_backfill_counter`, which increments on reset). librdkafka compares these two values internally and refuses to commit when the RecordBatch epoch is behind the OffsetFetch epoch.

This wasn't caught in testing because it requires the consumer to experience a rebalance or some other circumstance causing it to reconnect, which doesn't happen very often in practice.

## Fix

`crates/dekaf/src/read.rs` sets `partition_leader_epoch: self.partition_leader_epoch` instead of hardcoding `1`. The `partition_leader_epoch` field is initialized from `collection.binding_backfill_counter`, keeping RecordBatch epochs consistent with OffsetFetch's `committed_leader_epoch`.

## Test

`test_commits_persist_across_rebalances_after_reset` in `crates/dekaf/tests/e2e/collection_reset.rs` reproduces the scenario that the customer reported: a consumer initially works fine, is able to commit offsets, and then after reconnecting/collection reset, offset commits appear to be silently skipped, causing _subsequent_ reconnections to roll back to the last committed offset from the initial connection (resulting in duplicate records being seen by the consumer).

The test uses `commit_consumer_state` (`rd_kafka_commit(rk, NULL, ...)`) rather than `commit_message`, because `commit_message` constructs a raw offset without leader epoch and bypasses `rd_kafka_fetch_pos_cmp` entirely.

### Before fix (Dekaf serving `partition_leader_epoch: 1`)

```
2026-02-20T18:12:45.135858Z  INFO e2e::harness: Shard is primary task_name=test/dekaf/commits_after_reset/329c/source_ingest
2026-02-20T18:12:45.135870Z  INFO e2e::collection_reset: Injecting document to create new journal
2026-02-20T18:12:45.175779Z  INFO e2e::harness: Injecting documents count=1 path=data
2026-02-20T18:12:45.284784Z  INFO e2e::collection_reset: Waiting for epoch change and partitions topic=test_topic partition=0 previous_epoch=1 timeout_secs=30
2026-02-20T18:12:55.288608Z  INFO e2e::collection_reset: Epoch changed and partitions available topic=test_topic partition=0 previous_epoch=1 new_epoch=2
2026-02-20T18:12:55.288649Z  INFO e2e::collection_reset: Collection reset completed initial_epoch=1 new_epoch=2
2026-02-20T18:12:55.330229Z  INFO e2e::harness: Injecting documents count=3 path=data
test collection_reset::test_commits_persist_across_rebalances_after_reset has been running for over 60 seconds
2026-02-20T18:12:59.122673Z  INFO e2e::harness: Injecting documents count=1 path=data
2026-02-20T18:13:02.657697Z  WARN e2e::kafka: commit_consumer_state returned NoOffset (commit skipped due to epoch mismatch)
2026-02-20T18:13:02.800396Z  INFO e2e::harness: Injecting documents count=1 path=data

thread 'collection_reset::test_commits_persist_across_rebalances_after_reset' (265698) panicked at crates/dekaf/tests/e2e/collection_reset.rs:748:9:
assertion `left == right` failed: gen 3 should see only doc e (gen 2's commit must have persisted), but got 2 records: [String("d"), String("e")]
  left: 2
 right: 1
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
test collection_reset::test_commits_persist_across_rebalances_after_reset ... FAILED

```

After the fix is applied, the test passes.

### DIFF
--- a/crates/dekaf/src/read.rs
+++ b/crates/dekaf/src/read.rs
@@ -51,6 +51,12 @@ pub struct Read {
     deletes: DeletionMode,
 
     pub(crate) rewrite_offsets_from: Option<i64>,
+
+    // Leader epoch for this partition, used in RecordBatch headers.
+    // Must match the epoch reported in OffsetFetch's committed_leader_epoch
+    // so that librdkafka's epoch-aware commit filtering doesn't silently
+    // skip offset commits.
+    partition_leader_epoch: i32,
 }
 
 pub enum BatchResult {
@@ -132,6 +138,7 @@ impl Read {
             rewrite_offsets_from,
             deletes: auth.deletions(),
             offset_start: offset,
+            partition_leader_epoch: collection.binding_backfill_counter as i32,
         })
     }
 
@@ -442,7 +449,7 @@ impl Read {
                 headers: Default::default(),
                 key,
                 offset: kafka_offset,
-                partition_leader_epoch: 1,
+                partition_leader_epoch: self.partition_leader_epoch,
                 producer_epoch: 1,
                 producer_id: producer.as_i64(),
                 sequence: kafka_offset as i32,
--- a/crates/dekaf/tests/e2e/collection_reset.rs
+++ b/crates/dekaf/tests/e2e/collection_reset.rs
@@ -662,6 +662,99 @@ async fn test_fetch_response_includes_leader_epoch() -> anyhow::Result<()> {
     Ok(())
 }
 
+/// This catches a bug where the RecordBatch `partition_leader_epoch` didn't match
+/// the `committed_leader_epoch` returned by OffsetFetch. librdkafka compares these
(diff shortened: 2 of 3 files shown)


jgraettinger (approved):
LGTM

PR #2702, opened 2026-02-23
Shuffle V2: Coordinated disk-backed shuffle protocol (part one)

jgraettinger:
## Summary

- Introduces `crates/shuffle/`, a new Rust crate implementing the V2 shuffle protocol as a three-level RPC hierarchy (Session → Slice → Queue) that replaces the in-memory, per-shard `go/shuffle/` system
- Adds `go/protocols/shuffle/shuffle.proto` defining the wire protocol and generated Rust/gRPC bindings
- Includes `flowctl raw shuffle`, a development client that runs the full pipeline locally against a live collection

## Motivation

The legacy `go/shuffle/` system has scaling and correctness limitations that block further progress:

1. **In-memory document staging** limits how far reads can progress ahead of downstream processing, creating memory pressure vs throughput tradeoffs when processing larger documents 
2. **Per-shard independent checkpoints** prevent coordinated multi-shard transactions — there's no single frontier representing "data ready to process across all shards"
3. **Per-shard, per-journal RPCs** don't scale: M shards × N journals = up to M×N streams (e.g. 1M streams for 10 shards × 100k journals)

The V2 design addresses all three: documents are routed to on-disk queue files (eliminating memory pressure), a single Session coordinates a unified checkpoint frontier across all members, and M + M² streams replace M×N (110 streams for 10 members, regardless of journal count).

## Architecture

```
Coordinator
  └─ Session (one, on member 0)
       ├─ Slice 0 ──┬── Queue 0  (in-process)
       │             ├── Queue 1  (remote)
       │             └── Queue 2  (remote)
       ├─ Slice 1 ──┬── Queue 0  (remote)
       │             ├── Queue 1  (in-process)
       │             └── Queue 2  (remote)
       └─ Slice 2 ──┬── Queue 0  (remote)
                     ├── Queue 1  (remote)
                     └── Queue 2  (in-process)
```

**Session** — Coordinates the shuffle. Receives journal discoveries from Slices, assigns journal reads via range-overlap routing, pulls progress deltas, aggregates them into a four-stage checkpoint pipeline (progressed → unresolved → ready → consumed), and serves `NextCheckpoint` to the Coordinator.

**Slice** — Reads assigned journals, sequences documents with per-producer clock tracking (filtering duplicates from conservative reads and at-least-once delivery), orders them by priority then adjusted clock via a max-heap, routes each document to the owning Queue(s) by key hash and r-clock, and autonomously flushes Queues after observed commits.

**Queue** — Receives routed documents from all Slices, merges them into a single ordered stream via a priority heap (best-effort read leveling), and writes to disk. Responds to Flush once all preceding documents are durable.

### Protocol (`shuffle.proto`)
- Three bidirectional streaming RPCs: `Session`, `Slice`, `Queue`
- Wire types for frontiers (`FrontierChunk`, `JournalFrontier`, `ProducerFrontier`) with delta-encoded journal names
- `Enqueue` carries routed documents with priority, adjusted clock, packed key, and archived document bytes

### `crates/shuffle/`
- **`binding`**: Extracts and validates per-binding shuffle configuration from task specs (derivation transforms, materialization bindings, ad-hoc reads). Assigns cohorts by unique `(priority, read_delay)` tuples.
- **`frontier`**: Sorted frontier data structure with `reduce` (sorted merge), `resolve_hints` (causal hint resolution for cross-journal transactions), `project_unresolved_hints` (recovery projection), and chunked `Drain` for streaming over gRPC.
- **`session/`**: Handler opens Slice RPCs and reads the resume checkpoint. Actor runs the `select!` loop dispatching Session and Slice messages. State implements `Topology` (routing journal reads to members), `CheckpointPipeline` (four-stage state machine that gates on causal hint resolution).
- **`slice/`**: Handler opens Queue RPCs. Actor runs the `select!` loop over listing watches, journal probes, journal reads, and the ready-read heap. State implements `FlushState` (autonomous flush pipelining), `ProgressState` (pull-based progress reporting), and `sequence_document` (per-producer duplicate filtering with rollback detection). Sub-modules: `listing` (Gazette journal watch), `producer` (ProducerState with dual settled/pending maps), `routing` (r-clock rotation, key hash + range routing), `read` (ReadState with journal probing), `heap` (max-heap by priority DESC, adjusted clock ASC).
- **`queue/`**: Handler coordinates `QueueJoin` synchronization (defers actor spawn until all Slices connect). Actor runs the `select!` loop merging Enqueues from all Slices. Heap matches Slice ordering for best-effort read leveling.
- **`service`**: gRPC server entry point with peer channel caching, `spawn_{session,slice,queue}`, and HTTP/2 keep-alive tuning.

### Changes to existing crates
- **`proto-gazette/uuid`**: `Ord`/`Hash` on `Producer` (for use as map keys with passthrough hashers), `Clock` arithmetic and accessors, custom `Debug` impls for readability
- **`proto-flow`**: Generated Rust bindings and serde impls for `shuffle.proto`
- **`proto-grpc`**: Generated gRPC client/server stubs for the three shuffle RPCs
- **`labels/partition`**: `decode_field_range` now takes explicit field names and validates coverage (needed for shuffle key partition field extraction)
- **`doc/validation`**: Custom `Debug` on `Validator` showing schema URI instead of full schema

### Development tooling
- **`flowctl raw shuffle`**: Runs the full Session → Slice → Queue pipeline locally against a live production collection, useful for end-to-end validation

## Not yet implemented

- **Queue disk writes**: Currently emulated with a 50ms sleep stub; actual disk-backed queue file format is deferred to a follow-up
- **Queue dequeue**: Consumers will read queue files directly via file access; the dequeue protocol is out of scope for this PR
- **Lambda shuffle keys**: Binding tracks `uses_lambda` but actual lambda-based key extraction is not yet wired (and may be removed eventually? TBD).
- **Integration with reactor runtime**: This PR builds the shuffle crate and development client; wiring into the Gazette consumer framework reactor is a follow-up

## Testing

Extensive new unit tests of business logic, which is generally extracted into pure functions and state machines wherever possible.

E2E test via `RUST_LOG=shuffle=info cargo run -p flowctl -- raw shuffle --name <collection> --members 3`, which runs the full pipeline against a live collection.

A future PR will introduce integration tests which exercise shuffled reads using real Etcd & Gazette, as we have with the legacy `go/shuffle/` package.


go/protocols/shuffle/shuffle.proto
```diff
@@ -0,0 +1,285 @@
...
+  // session. The Session will read through these and emit the read-through
+  // frontier as the first NextCheckpoint response.
+  FrontierChunk resume_checkpoint_chunk = 2;
+
+  // NextCheckpoint requests the next available checkpoint delta.
+  // This is a blocking request: the Session only responds when progress is
+  // available. The client requests a next checkpoint at times of its choosing
+  // (e.g., after completing processing of the previous checkpoint).
+  message NextCheckpoint {}
+  NextCheckpoint next_checkpoint = 4;
```
williamhbaker, line 105: Is skipping field 3 intentional here?

crates/shuffle/src/lib.rs
```diff
@@ -0,0 +1,183 @@
...
+
+struct Verify<'p> {
+    source: &'static str,
+    expect: &'static str,
+    peer_endpoint: &'p str,
+    peer_index: usize,
+}
+
+impl<'p> Verify<'p> {
+    #[must_use]
```
williamhbaker, line 128: Is this `#[must_use]` redundant for an `anyhow::Result` return type?

crates/shuffle/src/binding.rs
```diff
@@ -0,0 +1,442 @@
...
+    /// JSON pointer for extracting document UUIDs.
+    pub source_uuid_ptr: json::Pointer,
+    /// True if shuffle key is dynamically computed via a lambda function.
+    pub uses_lambda: bool,
+    /// True if shuffle key equals the source collection key.
+    pub uses_source_key: bool,
+    /// Pre-parsed schema for validation and inference.
+    pub schema: doc::Validator,
+    /// True if documents should be validated on read (derivation transforms only).
+    pub validate_on_read: bool,
```
williamhbaker, line 51: Just for my own understanding, I wanted to ask about what this means. At a very superficial level my thought is that materializations do (schema) validation on reads as well, but clearly this must be referring to something a little different than that, and I wasn't able to deduce it.

crates/shuffle/src/binding.rs
```diff
@@ -0,0 +1,442 @@
...
+                seen.len() - 1
+            }
+        };
+        binding.cohort = cohort as u32;
+    }
+}
+
+/// Parse a collection schema bundle into a Validator and inferred Shape.
+/// Prefers the read schema; falls back to the write schema.
+fn build_schema(
```
williamhbaker, line 371: nit: I'm noting that we do this pattern of "build schema, validator, and shape" in at least [one other place](https://github.com/estuary/flow/blob/f31bcb27e83d33f83af1ffb93afc09789f9b6eb3/crates/validation/src/schema.rs#L17-L27) in the code base, and elsewhere too like [materialize-kafka](https://github.com/estuary/connectors/blob/63cc47b47dba07a8f320f17e5938a569c6d04afc/materialize-kafka/src/binding_info.rs#L186-L196). It would be nice to have a helper maybe like `doc::validation::build_validator_and_shape(bundle) -> Result<(Validator, Shape)>` for this repeated pattern.

crates/shuffle/src/session/state.rs
```diff
@@ -0,0 +1,1138 @@
...
+        let shuffle::slice_response::ListingAdded {
+            binding,
+            spec,
+            create_revision,
+            mod_revision,
+            route,
+        } = added;
+
+        let binding = &self.bindings[binding as usize];
+        let journal_name = spec.as_ref().map(|s| s.name.as_str()).unwrap_or("");
```
williamhbaker, line 133: Would it make sense to error here, instead of using an empty journal name? My understanding is that it shouldn't be possible to not have a name here.

crates/shuffle/src/slice/read.rs
```diff
@@ -0,0 +1,142 @@
...
+) -> anyhow::Result<(i64, Option<broker::Header>)> {
+    use futures::StreamExt;
+
+    let stream = client.read(broker::ReadRequest {
+        journal: journal.to_string(),
+        offset: -1,
+        block: false,
+        do_not_proxy: true,
+        header,
+        ..Default::default()
```
williamhbaker, line 102: Does setting the `metadata_only` flag here do any good? Looking at the broker code [here](https://github.com/gazette/core/blob/1ebeffaff75a949bcd8d425eeb88f1850ab2ed2c/broker/read_api.go#L136-L141) I suppose the `req.EndOffset != 0 && resp.Offset >= req.EndOffset` would always end up being true with a request offset of `-1` so perhaps it is redundant.

crates/shuffle/src/slice/read.rs
```diff
@@ -0,0 +1,142 @@
...
+                producer,
+            },
+            tail,
+            inner: read,
+        })
+    }
+}
+
+/// Probe the current write head of a journal via a non-blocking read at offset -1.
+/// Returns `(write_head, header)`. JournalNotFound yields `(0, None)`.
```
williamhbaker, line 87: The "JournalNotFound yields `(0, None)`" part from the comment doesn't seem to track, unless I'm missing something: Seems like it would hit the `map_read_error` and return an `Err(<something>)`.

crates/shuffle/src/slice/state.rs
```diff
@@ -0,0 +1,742 @@
...
+        self.in_flight.resize(member_count, true);
+        self.flushing = frontier;
+        self.seq
+    }
+
+    /// Record a Flushed response from a queue member.
+    /// Returns the completed frontier when all members have flushed.
+    pub fn on_flushed(&mut self, member_index: usize) -> Option<crate::Frontier> {
+        let Some(in_flight) = self.in_flight.get_mut(member_index) else {
+            return None;
```
williamhbaker, line 79: Is this a normally possible condition to happen? I'm thinking it would take some kind of application error, for a `flushed` to be sent twice for the same member, and we might want to error instead of silently dropping it.

crates/shuffle/src/binding.rs
```diff
@@ -0,0 +1,442 @@
...
+            uuid_ptr,
+            write_schema_json,
+        } = spec;
+
+        let (schema, shape) = build_schema(read_schema_json, write_schema_json)?;
+
+        Ok(Binding {
+            index: 0,
+            filter_r_clocks: false,
+            journal_read_suffix: String::new(), // No suffix for ad-hoc reads.
```
williamhbaker, line 325: Discussed in VC and will be taken care of in a later PR, but just noting here:

For an ad-hoc collection read with `flowctl raw shuffle --name=<collection_name>`, I found this not to work with an error of `Journal.metadata path segment: invalid length (0; expected 1 <= length <= 512)`. If I put a dummy suffix here though, it worked.

crates/shuffle/src/slice/actor.rs
```diff
@@ -0,0 +1,752 @@
...
+            }
+            return Ok(());
+        };
+
+        let mut lines_batch = match result {
+            Err(gazette::RetryError {
+                attempt,
+                inner: err,
+            }) => match err {
+                gazette::Error::BrokerStatus(broker::Status::JournalNotFound) => {
```
williamhbaker, line 339: I tested `flowctl raw shuffle` on a materialization task reading from collections with suspended journals, and got an error `unexpected broker status: Suspended`. I think there will need to be some additional handling for suspended journals.

williamhbaker (approved):
LGTM! Lots to digest here.

Since there is more to come on this and it isn't actually hooked up to anything production-related yet, none of these comments are blocking. The most substantive thing I found was using the `flowctl raw shuffle` on bindings with suspended journals does not work, so that will need to be taken care of either here or in a follow-up if it makes more sense.

go/protocols/shuffle/shuffle.proto
jgraettinger: Nope, will fix.

crates/shuffle/src/lib.rs
jgraettinger: Yep, it is. Will remove.

crates/shuffle/src/binding.rs
jgraettinger: I've removed this in future commits. It was a hold-over behavior from the legacy implementation, which validates derivations (because our next step is to feed the json bytes into the derivation connector) but not materializations (because we'll plumb the document into a combiner, which will _also_ validate, so it's redundant).

In upcoming work, the combiner becomes aware of the validation status from the shuffled reader, so there's no redundant validation work, and it's always beneficial to validate early, at read.

crates/shuffle/src/session/state.rs
jgraettinger: Yep this can unwrap(), will do.

crates/shuffle/src/slice/read.rs
jgraettinger: I'll add the following comment:

    // A non-blocking read at offset -1 returns OffsetNotYetAvailable immediately.
    // Note that the `client.read()` controls ReadRequest::metadata_only internally.

Unpacking, the `gazette` crate client controls metadata_only and uses a two-request strategy: the first is metadata only to fetch a potential fragment URL, and the second is a non-metadata streaming read.

crates/shuffle/src/slice/read.rs
jgraettinger: Agreed; removing.

crates/shuffle/src/slice/state.rs
jgraettinger: Sure, mapped into an error.

crates/shuffle/src/binding.rs
jgraettinger: Yep, fixed.

crates/shuffle/src/slice/actor.rs
jgraettinger: Yep, fixed in the next set of changes.

PR #2703, opened 2026-02-23
Gazette Rust client: ReadLines stream, list Subscriber, and simd-doc extraction

jgraettinger:
**Description:**

Foundation work for the shuffle-v2 effort: adds new streaming primitives to the `gazette` Rust client and extracts low-level parsing functions from `simd-doc` so they can be used with custom buffering strategies.

- **`ReadLines` stream** (`gazette::journal::read::lines`): Accumulates newline-aligned content from journal reads into large contiguous buffers suited for batch parsing (e.g. simdjson). Supports configurable target batch size and padding via const generics, deferred error handling, content reinsertion via `put_back()`, and `FusedStream` semantics.

- **`Fold` trait and `Subscriber`** (`gazette::journal::list`): Refactors journal listing into a `Fold` trait for custom processing of streamed `ListResponse` snapshots. Adds `Subscriber` / `SubscriberFold` that performs streaming sorted-merge diffing to detect journal adds/removes, with delta-encoded name storage and partial-progress retry recovery.

- **`delta` module** (`gazette::delta`): UTF-8-safe delta encoding for sorted string sequences (used by `SubscriberFold` to compactly represent journal names).

- **`simd-doc` extraction**: Introduces `SimdParser` wrapper and extracts `parse_one`, `parse_many`, and `transcode_many` as standalone public functions. The existing `Parser` type delegates to these, preserving all current behavior while enabling advanced users to manage their own input buffering.

- **Router header handling**: `Router::route()` now takes `&mut Option<Header>` instead of `Option<&mut Header>` and clears the entire header after routing, since request headers are a server-to-server proxy mechanism not intended for client-to-server requests.

- **Assemble**: Enables the `name:prefix` label in collection journal selectors (previously commented out pending connector propagation).

- **Error classification**: `FetchFragment` errors with HTTP 4xx status codes are now classified as non-transient, while network-level failures (no status) remain retryable.

**Notes for reviewers:**

The base `assemble` commit is also an open PR here: https://github.com/estuary/flow/pull/2668


crates/simd-doc/src/lib.rs
```diff
@@ -221,38 +178,166 @@ impl Parser {
...
+    alloc: &'a doc::Allocator,
+    output: &mut Vec<(doc::HeapNode<'static>, i64)>,
+) -> Result<doc::HeapNode<'a>, std::io::Error> {
+    // Safety: we transmute alloc to 'static for internal use, and transmute
+    // the returned HeapNode back to 'a. This is sound because the node is
+    // allocated from `alloc` which lives for 'a.
+    let alloc: &'static doc::Allocator = unsafe { std::mem::transmute(alloc) };
+
+    assert!(output.is_empty());
+    input.push(b'\n');
```
williamhbaker, line 210: Would it make sense to clear the input at the end of this function, similar to how the `parse_many` and `transcode_many` drain consumed bytes from the input buffer? Especially since the input is being modified here to add a newline at the end, it seems like it would make sense for the input to be consumed.

crates/gazette/src/journal/mod.rs
```diff
@@ -3,11 +3,11 @@ use proto_gazette::broker;
...
 
 mod append;
-mod list;
-mod read;
+pub mod list;
+pub mod read;
 
-mod read_json_lines;
-pub use read_json_lines::{ReadJsonLine, ReadJsonLines};
+// TODO(johnny): Update usages to gazette::journal::read::ReadJsonLines;
```
williamhbaker, line 9: Not sure I'm understanding this comment; would we want to eventually update this to ReadLines?

crates/gazette/src/journal/read/lines.rs
```diff
@@ -0,0 +1,792 @@
...
+    fn take_aligned(
+        &mut self,
+        yield_count: u16,
+        buffered_bytes: usize,
+        padding: u16,
+    ) -> LinesBatch {
+        // `self.offset` is the end of all buffered content.
+        // The start of buffers[0] is `self.offset - buffered_bytes`.
+        let batch_offset = self.offset - buffered_bytes as i64;
+        let tailing = self.tailing;
```
williamhbaker, line 147: I'm assuming this "remember the tailing status of the data from when it was added to the batch" assignment will be important for later shuffle read work, and am curious to see how it is used.

crates/simd-doc/src/lib.rs
jgraettinger: In practice the sole current caller clears immediately after calling this routine.

From a design standpoint, I'd say that we don't own `input`, the caller does, so we wouldn't want to clear a buffer that the caller may have another use for. Perhaps it should pop the appended newline though.

crates/gazette/src/journal/mod.rs
jgraettinger: Yes, there's now a fair amount of conceptual overlap between `ReadJsonLines`, vs `ReadLines` paired with a simd-doc `SimdParser`.

`ReadJsonLines` is used by `dekaf`, and I wouldn't want to monkey with it's implementation until `ReadLines` has been battle tested, but eventually we should kill off `ReadJsonLines`.

crates/gazette/src/journal/read/lines.rs
jgraettinger: Yep! It's used to account for "reads which are known to be behind the write-head which we should await before draining reads from a priority heap", vs "reads which could block indefinitely before they next yield data".

PR #2705, opened 2026-02-23
validation: keep as document field when incompatible

danielnelson:
**Description:**

If the document field is incompatible, the materialization will need to be backfilled.  Previously we would lose track of the document fields special status and select it as a value.

**Workflow steps:**

No changes.

None

**Notes for reviewers:**

### DIFF
--- a/crates/validation/src/field_selection.rs
+++ b/crates/validation/src/field_selection.rs
@@ -490,7 +490,7 @@ pub fn group_outcomes(
             (Some(select), Some(reject)) => EOB::Both(select, reject),
         };
 
-        if let EOB::Left(_select) = &outcome {
+        if outcome.has_left() {
             // Track selected fold and pointer for subsequent evaluations.
             if field_ptr != "" {
                 selected_ptrs.push(field_ptr);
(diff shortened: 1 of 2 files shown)


crates/validation/src/field_selection.rs
```diff
@@ -490,7 +490,7 @@ pub fn group_outcomes(
             (Some(select), Some(reject)) => EOB::Both(select, reject),
         };
 
-        if let EOB::Left(_select) = &outcome {
+        if outcome.has_left() {
```
jgraettinger, line 493: I'm confused -- this change appears to do nothing, because we previously were destructing by reference, which does not consume the Option.

danielnelson: We now enter the condition when outcome is `EOB::Both`, instead of only when it is `EOB::Left`.  This allows us to set `document_field`.  Later in `build_selection` the field can be set as the document instead of as a value.

jgraettinger: 🤦 thanks, I need to be walked through that.

PR #2707, opened 2026-02-23
dekaf: introduce retries with exponential backoff for SOPS

jshearer:
We encountered a couple of SOPS errors under heavy load recently, so this commit adds retries with backoff.

Sample error:
```
WARN authenticate{username="..."}: dekaf: error=Unknown(decrypting sops document failed: Failed to get the data key required to decrypt the SOPS file.
Group 0: FAILED
  projects/...: FAILED
    - | failed to decrypt sops data key with GCP KMS key: context
      | deadline exceeded

Recovery failed because no master key was able to decrypt the file. In
order for SOPS to recover the file, at least one key has to be successful,
but none were.

Stack backtrace:

   0: std::backtrace::Backtrace::create
   1: anyhow::error::<impl anyhow::Error>::msg
   2: dekaf::topology::extract_dekaf_config::{{closure}}
```

### DIFF
--- a/crates/dekaf/src/topology.rs
+++ b/crates/dekaf/src/topology.rs
@@ -603,10 +603,30 @@ pub async fn extract_dekaf_config(
     }
     let config = serde_json::from_slice::<models::DekafConfig>(&spec.config_json)?;
 
-    let decrypted_endpoint_config =
-        unseal::decrypt_sops(&RawValue::from_str(&config.config.to_string())?).await?;
+    let raw_config = RawValue::from_str(&config.config.to_string())?;
 
-    let dekaf_config =
-        serde_json::from_str::<connector::DekafConfig>(decrypted_endpoint_config.get())?;
-    Ok(dekaf_config)
+    // SOPS decryption calls out to GCP KMS, which can transiently fail.
+    // Retry with backoff to handle these transient failures.
+    let mut last_err = None;
+    for attempt in 0..5u32 {
+        if attempt > 0 {
+            let base_ms = 1000 * 2u64.pow(attempt - 1);
+            let jitter_ms = rand::random_range(0..=base_ms / 2);
+            tokio::time::sleep(std::time::Duration::from_millis(base_ms + jitter_ms)).await;
+        }
+        match unseal::decrypt_sops(&raw_config).await {
+            Ok(decrypted) => {
+                let dekaf_config = serde_json::from_str::<connector::DekafConfig>(decrypted.get())?;
+                return Ok(dekaf_config);
+            }
+            Err(err) => {
+                tracing::warn!(attempt, error = ?err, "sops decryption failed, retrying");
+                last_err = Some(err);
+            }
+        }
+    }
+
+    // Unwrap is safe: the loop always executes at least once, and we only
+    // reach here via the Err arm which sets last_err = Some.
+    Err(last_err.unwrap())
 }


crates/dekaf/src/topology.rs
```diff
@@ -603,10 +603,30 @@ pub async fn extract_dekaf_config(
     }
     let config = serde_json::from_slice::<models::DekafConfig>(&spec.config_json)?;
 
-    let decrypted_endpoint_config =
-        unseal::decrypt_sops(&RawValue::from_str(&config.config.to_string())?).await?;
+    let raw_config = RawValue::from_str(&config.config.to_string())?;
+
+    // SOPS decryption calls out to GCP KMS, which can transiently fail.
+    // Retry with backoff to handle these transient failures.
+    let mut last_err = None;
```
jgraettinger, line 610: nit: this doesn't need to be Option. You can simply declare it but not initialize it. Rust is smart enough to recognize that by the time you're at loop end, it's been initialized.

jgraettinger (approved):
LGTM % nit

crates/dekaf/src/topology.rs
jshearer: Hm, am I doing it wrong?

```rust
pub async fn extract_dekaf_config(
    spec: &proto_flow::flow::MaterializationSpec,
) -> anyhow::Result<connector::DekafConfig> {
    if spec.connector_type != proto_flow::flow::materialization_spec::ConnectorType::Dekaf as i32 {
        anyhow::bail!("Not a Dekaf materialization")
    }
    let config = serde_json::from_slice::<models::DekafConfig>(&spec.config_json)?;

    let raw_config = RawValue::from_str(&config.config.to_string())?;

    // SOPS decryption calls out to GCP KMS, which can transiently fail.
    // Retry with backoff to handle these transient failures.
    let mut last_err;
    for attempt in 0..5u32 {
        if attempt > 0 {
            let base_ms = 1000 * 2u64.pow(attempt - 1);
            let jitter_ms = rand::random_range(0..=base_ms / 2);
            tokio::time::sleep(std::time::Duration::from_millis(base_ms + jitter_ms)).await;
        }
        match unseal::decrypt_sops(&raw_config).await {
            Ok(decrypted) => {
                let dekaf_config = serde_json::from_str::<connector::DekafConfig>(decrypted.get())?;
                return Ok(dekaf_config);
            }
            Err(err) => {
                tracing::warn!(attempt, error = ?err, "sops decryption failed, retrying");
                last_err = err;
            }
        }
    }

    // Unwrap is safe: the loop always executes at least once, and we only
    // reach here via the Err arm which sets last_err = Some.
    Err(last_err)
}
```

```
error[E0381]: used binding `last_err` is possibly-uninitialized
   --> crates/dekaf/src/topology.rs:631:9
    |
610 |     let mut last_err;
    |         ------------ binding declared here but left uninitialized
...
624 |                 last_err = err;
    |                 -------- binding initialized here in some conditions
...
631 |     Err(last_err)
    |         ^^^^^^^^ `last_err` used here but it is possibly-uninitialized

For more information about this error, try `rustc --explain E0381`.
error: could not compile `dekaf` (lib) due to 1 previous error
```

jgraettinger: :shrug: never mind then, sorry for the noise

jshearer: Too bad, this would have been really cool!

PR #2712, opened 2026-02-25
Docs: Update connector image tags

aeluce:
**Description:**

Ours docs have been using a `:dev` image tag as a default for connector images. However, `flowctl` can have some unexpected behavior when working with a spec where the image tag is not explicitly defined in the `connector_tags` table.

This update therefore changes `:dev` tags in docs to specific versions so that users who base their spec off the examples start with a solid base. Some exceptions: the NetSuite SuiteQL and Exchange Rates API connectors only have a `:dev` tag in `connector_tags` and so still use this in their examples.

Going forward, when connectors upgrade to a new major version, associated docs will also need to be updated. This may be more complex when adding variants to the mix. We could also potentially rethink how we use the main README for capture and materialization connectors; currently, these pages add a secondary location where the version tag must be kept updated. While overhauling them is likely outside the scope of this PR, I wanted to make a note for the future.

A couple people have also commented that the paragraph in connector reference docs directing users to the associated GitHub package is unnecessary or oddly placed, causing more confusion than actual benefit. As it also uses the `:dev` tag, this seemed as good a time as any to go through and remove it.

(Almost) all connector reference docs. Also a few guides that include spec examples and a note in the `flowctl` concepts page.

**Notes for reviewers:**

Chose a couple reviewers from those involved in the engineering review conversation about this, but feel free to bring someone else into the conversation as needed. Thanks!

### DIFF
--- a/site/docs/concepts/connectors.md
+++ b/site/docs/concepts/connectors.md
@@ -53,7 +53,7 @@ or any other public image registry provider.
 
 To interface with a connector, the Estuary runtime needs to know:
 
-1. The specific image to use, through an image name such as `ghcr.io/estuary/source-postgres:v5`.
+1. The specific image to use, through an image name such as `ghcr.io/estuary/source-postgres:v3`.
    Notice that the image name also conveys the specific image registry and version tag to use.
 
 2. Endpoint configuration such as a database address and account, with meaning that is specific to the connector.
@@ -185,7 +185,7 @@ captures:
   acmeCo/postgres-capture-ssh:
     endpoint:
       connector:
-        image: ghcr.io/estuary/source-postgres:v5
+        image: ghcr.io/estuary/source-postgres:v3
         config:
           address: 127.0.0.1:5432
           database: flow


site/docs/concepts/connectors.md
```diff
@@ -53,7 +53,7 @@ or any other public image registry provider.
 
 To interface with a connector, the Estuary runtime needs to know:
 
-1. The specific image to use, through an image name such as `ghcr.io/estuary/source-postgres:dev`.
+1. The specific image to use, through an image name such as `ghcr.io/estuary/source-postgres:v5`.
```
willdonnelly: Is it intentional that this example version doesn't exist? I see that the version in the actual Postgres capture docs is correct (`v3`) so I'm not sure if using `v5` here was intentional or not.

aeluce: Eagle-eyed catch, thank you! There are also mentions of `materialize-postgres` on the same page, which is on `:v5`, so I glossed over the `source` part here. 🤦

PR #2729, opened 2026-03-03
Docs: Add missing connector sample specs

aeluce:
**Description:**

A few of our connector docs don't include a sample YAML configuration, or include a validation spec or something else rather than an example on how users would configure their connector.

This update fixes these docs to provide at least a minimal config to get users started. This also ensures that the latest image tag is available in the connector reference documentation as part of the example spec (as a follow up to #2712 which changed already existing `:dev` tags).

In some cases, connector properties also needed to be updated to correctly reflect the configuration.

The PR currently excludes HTTP ingest/webhook capture connectors, which are also missing example specs, as some of their properties are in flux. I can wrap these up after documentation changes for ECDSA are merged.
(**Edit:** these are added now.)

**Notes for reviewers:**

Chose this week's triage lead for a default reviewer, though certainly feel free to tag someone else for the review instead/as needed. Thanks!

### DIFF
--- a/site/docs/reference/Connectors/capture-connectors/MySQL/mysql-batch.md
+++ b/site/docs/reference/Connectors/capture-connectors/MySQL/mysql-batch.md
@@ -151,7 +151,7 @@ captures:
       connector:
         image: ghcr.io/estuary/source-mysql-batch:v1
         config:
-          address: host:3306
+          address: host:port
           user: flow_capture
           password: secret
     bindings:
--- a/site/docs/reference/Connectors/capture-connectors/woocommerce.md
+++ b/site/docs/reference/Connectors/capture-connectors/woocommerce.md
@@ -35,9 +35,9 @@ You configure connectors either in the Estuary web app, or by directly editing t
 #### Endpoint
 | Property           | Title           | Description                                                        | Type   | Required/Default |
 | ------------------ | --------------- | ------------------------------------------------------------------ | ------ | ---------------- |
-| `/customer_key`    | Customer Key    | Customer Key for API in WooCommerce shop                           | string | Required         |
-| `/customer_secret` | Customer Secret | Customer Secret for API in WooCommerce shop                        | string | Required         |
-| `/shop_name`       | Shop Name       | The name of the store.                                             | string | Required         |
+| `/api_key`    | Customer Key    | Customer Key for API in WooCommerce shop                           | string | Required         |
+| `/api_secret` | Customer Secret | Customer Secret for API in WooCommerce shop                        | string | Required         |
+| `/shop`       | Shop Name       | The name of the store.                                             | string | Required         |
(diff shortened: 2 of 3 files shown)


site/docs/reference/Connectors/capture-connectors/MySQL/mysql-batch.md
```diff
@@ -142,6 +142,26 @@ to learn more about using connectors.
...
+### Sample
+
+```yaml
+captures:
+  ${PREFIX}/${CAPTURE_NAME}:
+    endpoint:
+      connector:
+        image: ghcr.io/estuary/source-mysql-batch:v1
+        config:
+          address: host:3306
```
Alex-Bair: nit: In these documentation updates, would it be clearer / more consistent to use `host:port` as the sample's `address` instead of specifying a port? I noticed the `oracle-batch.md` docs have `host:port` but others use `host:3306` & I figured we'd want the samples across the docs to be as consistent as possible with each other .

site/docs/reference/Connectors/capture-connectors/woocommerce.md
```diff
@@ -51,12 +51,22 @@ You configure connectors either in the Estuary web app, or by directly editing t
...
+```yaml
+captures:
+  ${PREFIX}/${CAPTURE_NAME}:
+    endpoint:
+      connector:
+        image: ghcr.io/estuary/source-woocommerce:v1
+        config:
+          customer_key: api-key
+          customer_secret: secret
+          shop_name: acmeCo
```
Alex-Bair: note: These endpoint config properties' names line up with the ones listed above in the "Endpoint" section, but I noticed that these actually aren't the names used by captures. The actual property names I'm seeing used by the connector are:
`customer_key` -> `api_key`
`customer_secret` -> `api_secret`
`shop_name` -> `shop`

We don't necessarily need to fix that in this PR (it's out of the original scope of adding missing samples), but I wanted to mention it so we can come back later and align the property names here in the docs with what's used by the connector.

aeluce: Thanks for catching this! I'm not so familiar with figuring out the proper spec info for old Airbyte connectors. But yes, always good to update property names where they're incorrect. 😄

Alex-Bair: Absolutely! I only noticed it by looking at the spec of a capture task, which I don't think is something non-developers can easily check.

A circuitous way of checking what the specs look like for ATF connectors is by seeing what version of the Airbyte connector is being used then checking Airbyte's docs / code for the spec, keeping in mind it may have changed from the version we're running. But that's a lot of effort to spend investigating when a developer can pretty easily check what the spec looks like for an ATF connector with `flowctl raw spec` or checking an active capture task. If you want us to confirm what a spec should look like for these connectors, feel free to reach out and we can do that.

PR #2731, opened 2026-03-04
Dekaf: Prevent accumulation of Read RPCs due to abandoned journal reads

jshearer:
## Summary

Periodically reap abandoned Read RPCs, dropping any `PendingRead` not consumed by a `Fetch` request within `STALE_READ_TIMEOUT` (default 1m). This prevents once-Fetched but now abandoned Reads from accumulating stalled gRPC streams that exhaust the HTTP/2 connection-level flow control window and block all broker responses, including partition List RPCs

Additionally,
- Add a 30-second timeout to `fetch_partitions()` List RPCs in the task manager loop. It isn't expected that this codepath actually gets hit, and is indicative of a problem if it is. Leaving this timeout in here mainly to prove to ourselves via logging that it's not happening.
- We realized that if the TaskManager returns `Err` or panics, we would never know. Added logic to capture the task handle and log if it errors or panics.

### DIFF
--- a/crates/dekaf/src/session.rs
+++ b/crates/dekaf/src/session.rs
@@ -29,6 +29,9 @@ struct PendingRead {
     offset: i64,          // Journal offset to be completed by this PendingRead.
     last_write_head: i64, // Most-recent observed journal write head.
     leader_epoch: i32,    // Leader epoch (binding backfill counter) for this read.
+    // Last time this read was included in a Fetch request. Used to reap reads for
+    // partitions the client has stopped fetching.
+    last_accessed: std::time::Instant,
     handle: tokio_util::task::AbortOnDropHandle<anyhow::Result<(Read, BatchResult)>>,
 }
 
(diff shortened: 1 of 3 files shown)


crates/dekaf/src/session.rs
```diff
@@ -685,22 +701,10 @@ impl Session {
                 let pending_info = self
                     .reads
                     .get(&key)
-                    .map(|(p, s)| (p.offset, p.leader_epoch, s.elapsed()));
+                    .map(|p| (p.offset, p.leader_epoch));
 
                 match pending_info {
-                    Some((_, _, elapsed)) if elapsed > std::time::Duration::from_secs(60 * 5) => {
```
jshearer, line 691: There was some old code in here to time out Reads every 5 minutes. This was to work around an issue with data-plane gateway, which is no longer a factor, so I got rid of it.

jgraettinger (approved):
LGTM

PR #2734, opened 2026-03-04
ops-catalog: one-off script to update L1 reporting for all data planes

williamhbaker:
**Description:**

After updating the `data-plane-template.bundle.json`, we need to re-publish the catalog-stats and stats collections for all data planes.

I'm not intending to merge this or keep it around permanently. If it's something we need to do more of, maybe a new control plane API would make sense.

This should be run after https://github.com/estuary/flow/pull/2676 is merged and the agent-api deployed from it. The L2 derivation will be updated using the `update-l2-reporting` endpoint then too.

**Workflow steps:**

**Notes for reviewers:**

### DIFF
--- a/ops-catalog/update-l1-reporting.sh
+++ b/ops-catalog/update-l1-reporting.sh
@@ -1,0 +1,72 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# One-off script to re-publish the catalog-stats and stats collections
+# for all data planes, after updating data-plane-template.bundle.json.
+# Based on the mise/tasks/local/stack script.
+
+DB_URL="${DB_URL:-postgresql://postgres:postgres@localhost:5432/postgres}"
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+TEMPLATE=$(cat "$SCRIPT_DIR/data-plane-template.bundle.json" | sed "s/'/''/g")
+
+# Only data planes in this list will be updated. All others are skipped.
+APPLICABLE_DATA_PLANES=(
+  "ops/dp/public/not-a-real-dataplane"
+  "ops/dp/public/local-cluster"
+)
+
+ALL_DATA_PLANES=$(psql "$DB_URL" -t -A -c \
+  "select data_plane_name from data_planes order by data_plane_name")
+
+# Warn about any entries in APPLICABLE_DATA_PLANES that don't exist in the database.
+for WANTED in "${APPLICABLE_DATA_PLANES[@]}"; do
+  if ! echo "$ALL_DATA_PLANES" | grep -qx "$WANTED"; then
+    echo "Warning: '$WANTED' is not a known data plane" >&2
+  fi
+done
+
+for DATA_PLANE in $ALL_DATA_PLANES; do
+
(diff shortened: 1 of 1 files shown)


ops-catalog/update-l1-reporting.sh
```diff
@@ -0,0 +1,50 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# One-off script to re-publish the catalog-stats and stats collections
+# for all data planes, after updating data-plane-template.bundle.json.
+# Based on the mise/tasks/local/stack script.
+
+DB_URL="${DB_URL:-postgresql://postgres:postgres@localhost:5432/postgres}"
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+TEMPLATE=$(cat "$SCRIPT_DIR/data-plane-template.bundle.json" | sed "s/'/''/g")
```
psFried, line 10: What's the sed replacement for?

ops-catalog/update-l1-reporting.sh
```diff
@@ -0,0 +1,50 @@
...
+
+# Query all data planes, deriving the base name by stripping the 'ops/dp/' prefix.
+for BASE_NAME in $(psql "$DB_URL" -t -A -c \
+  "select replace(data_plane_name, 'ops/dp/', '') from data_planes order by data_plane_name"); do
+
+  echo "Updating L1 reporting for data plane: $BASE_NAME"
+
+  CATALOG=$(echo "$TEMPLATE" | sed "s|BASE_NAME|$BASE_NAME|g")
```
psFried, line 39: TIL you can use almost any character you want for delimiters in a sed expression 👍

ops-catalog/update-l1-reporting.sh
psFried: ah, nvm, looks like there's some single quote characters in the ops catalog bundle, which this is doubling to satisfy postgres

ops-catalog/update-l1-reporting.sh
```diff
@@ -0,0 +1,50 @@
...
+    insert into draft_specs (draft_id, catalog_name, spec_type, spec)
+    select new_draft_id, "key", 'collection'::catalog_spec_type, "value"
+    from json_each(json_extract_path(bundled_catalog_arg, 'collections'))
+    where "key" like '%/catalog-stats' or "key" like '%/stats';
+end \$\$
+language plpgsql;
+commit;
+EOF
+
+  echo "Done: $BASE_NAME"
```
psFried: I'd recommend adding a `sleep 30` here. As it is, this will likely run into a current weakness in our publication handling, because all of these publications will implicitly pull in the L2 derivation. What we've seen in similar scenarios is lots of optimistic locking failures on `last_build_id`, and retries which tend to exacerbate the problem, which ultimately presents as long delays in handling user-initiated publications. Basically, we gum up all our publication handlers with retries of publications that all contend over the `last_build_id` of the common specs. If we can get the publications to happen one at a time, then it'll avoid that issue.

williamhbaker: Added the 30s sleep, thanks!

Also since I am planning on running this on only a few at a time initially at least, I modified the script so it takes a list of applicable data planes and only runs on those. I'll actually pull the full list of data planes prior to starting this, and then run it in smaller batches at first to make sure things are working as they should be.

PR #2742, opened 2026-03-05
Phil/disco schemas

psFried:
**Description:**

Rolls up a revert of #2720 along with a change to ensure that auto-discovers will update collection specs that already exist but correspond to currently disabled bindings.

**Notes for reviewers:**

The change here is kind of subtle, but I tried to show it in the integration test. Previously that new test scenario would fail for auto-discovers, but would pass if the discover were performed by a user. Now it works the same for both.

The plan will be to deploy this change, let auto-discovers re-publish any disabled collections that already exist, and then re-apply #2720 once all the existing collections have the `flow://connector-schema` definition added to them.

### DIFF
--- a/crates/agent/src/main.rs
+++ b/crates/agent/src/main.rs
@@ -1,3 +1,4 @@
+#![recursion_limit = "256"] // Rust 1.94 fails to compile agent without this
 // Links in the allocator crate, which sets the global allocator to jemalloc
 extern crate allocator;
 


jgraettinger (approved):
LGTM

A sanity check -- the commit says:

> This changes `prune_unchanged_specs` so that it no longer prunes specs
for bindings that are disabled. Instead, controller-initiated
publications now use the `PruneUnboundCollections` publication finalizer
in order to avoid publishing collections that meet both the following
criteria:
> - The drafted collection represents an insert (no existing live spec)
> - The capture binding is disabled (the collection is 'unbound')

From reviewing, I understand this is _not_ meant to say that **user**-initiated discovers _no longer_ prune new, disabled bindings. That still happens, because both the auto-discover and also the user-initiated publication paths both use the finalizer. Right?

PR #2745, opened 2026-03-06
Phase 1a/b: Invite links gql

GregorShear:
- Add `internal.invite_links` table to replace the directives-based invite flow with a simplified model
- Implement GraphQL query and mutations (`inviteLinks`, `createInviteLink`, `redeemInviteLink`, `deleteInviteLink`)
- Single-use links are deleted upon redemption; multi-use links persist
- Include transitional dual-write triggers that sync between `public.directives` and `internal.invite_links`, plus a backfill of existing unredeemed grant directives — to be removed after the UI adopts the new GraphQL API

```gql
mutation Create {
  createInviteLink(
    catalogPrefix: "acmeCo/"
    capability: admin
    singleUse: true
  ) {
    token
    catalogPrefix
    capability
    singleUse
    detail
    createdAt
  }
}

mutation Redeem {
  redeemInviteLink(token: "a1b2c3d4-e5f6-7890-abcd-ef1234567890") {
    catalogPrefix
    capability
  }
}

mutation Delete {
  deleteInviteLink(token: "a1b2c3d4-e5f6-7890-abcd-ef1234567890")
}


query {
  inviteLinks(filter:  {
     prefix:  {
        startsWith: "gregCo/"
     }
     singleUse:  {
        eq: true
     }
  }) {
    edges {
      node {
        catalogPrefix
        token
        capability
      }
    }
  }
}

```

## Test plan

- [x] Snapshot tests for create, redeem, unauthorized create, bad token, and exhausted single-use redemption
- [x] Manual test of transitional triggers: create a directive via PostgREST and confirm it appears in `internal.invite_links`
- [x] Verify backfill picks up existing unredeemed grant directives

### DIFF
--- a/crates/control-plane-api/src/server/public/graphql/authorized_prefixes.rs
+++ b/crates/control-plane-api/src/server/public/graphql/authorized_prefixes.rs
@@ -1,0 +1,181 @@
+/// Returns catalog prefixes where the authenticated user has at least
+/// `min_capability`, optionally narrowed to those overlapping `prefix_filter`.
+///
+/// Intended for use by GraphQL queries that list resources scoped to the
+/// caller's authorized prefixes, with an optional prefix filter.
+///
+/// When `prefix_filter` is provided, a prefix is included if the filter is a
+/// sub-prefix of the grant OR the grant is a sub-prefix of the filter. This
+/// bidirectional check lets callers query with a filter that is either broader
+/// or narrower than their grants.
+pub(super) fn authorized_prefixes(
+    role_grants: &tables::RoleGrants,
+    user_grants: &tables::UserGrants,
+    user_id: uuid::Uuid,
+    min_capability: models::Capability,
+    prefix_filter: Option<&str>,
+) -> Vec<String> {
+    let mut prefixes: Vec<String> =
+        tables::UserGrant::transitive_roles(role_grants, user_grants, user_id)
+            .filter(|grant| grant.capability >= min_capability)
+            .filter(|grant| {
+                prefix_filter.is_none_or(|pf| {
+                    grant.object_role.starts_with(pf) || pf.starts_with(&*grant.object_role)
+                })
+            })
+            .map(|grant| grant.object_role.to_string())
+            .collect();
+
+    // Sort and remove child prefixes that are already covered by a parent prefix.
+    prefixes.sort();
+    prefixes.dedup();
+
+    let mut pruned: Vec<String> = Vec::new();
+    for p in prefixes {
+        if pruned
+            .last()
+            .is_none_or(|parent| !p.starts_with(parent.as_str()))
+        {
+            pruned.push(p);
+        }
+    }
+
+    pruned
+}
+
+#[cfg(test)]
+mod tests {
+    use super::authorized_prefixes;
+    use models::Capability::{Admin, Read, Write};
+
+    fn make_grants(
+        user_grants: &[(uuid::Uuid, &str, models::Capability)],
+        role_grants: &[(&str, &str, models::Capability)],
+    ) -> (tables::UserGrants, tables::RoleGrants) {
+        let ug = tables::UserGrants::from_iter(user_grants.iter().map(|(id, obj, cap)| {
+            tables::UserGrant {
+                user_id: *id,
+                object_role: models::Prefix::new(*obj),
+                capability: *cap,
+            }
+        }));
+        let rg = tables::RoleGrants::from_iter(role_grants.iter().map(|(sub, obj, cap)| {
+            tables::RoleGrant {
+                subject_role: models::Prefix::new(*sub),
+                object_role: models::Prefix::new(*obj),
+                capability: *cap,
+            }
+        }));
+        (ug, rg)
+    }
+
+    const ALICE: uuid::Uuid = uuid::Uuid::from_bytes([0x11; 16]);
+
+    #[test]
+    fn no_filter_returns_all_at_or_above_capability() {
+        let (ug, rg) = make_grants(
+            &[
+                (ALICE, "acmeCo/", Admin),
+                (ALICE, "widgets/", Write),
+                (ALICE, "readonly/", Read),
+            ],
+            &[],
+        );
+
+        let result = authorized_prefixes(&rg, &ug, ALICE, Admin, None);
+        assert_eq!(result, vec!["acmeCo/"]);
+
+        let result = authorized_prefixes(&rg, &ug, ALICE, Write, None);
+        assert_eq!(result, vec!["acmeCo/", "widgets/"]);
+
+        let result = authorized_prefixes(&rg, &ug, ALICE, Read, None);
+        assert_eq!(result, vec!["acmeCo/", "readonly/", "widgets/"]);
+    }
+
+    #[test]
+    fn filter_narrower_than_grant() {
+        // Grant is on "acmeCo/", filter is "acmeCo/data/" — the grant covers
+        // the filter, so "acmeCo/" is included.
+        let (ug, rg) = make_grants(&[(ALICE, "acmeCo/", Admin)], &[]);
+
+        let result = authorized_prefixes(&rg, &ug, ALICE, Admin, Some("acmeCo/data/"));
+        assert_eq!(result, vec!["acmeCo/"]);
+    }
+
+    #[test]
+    fn filter_broader_than_grant() {
+        // Grant is on "acmeCo/data/", filter is "acmeCo/" — the grant starts
+        // with the filter, so "acmeCo/data/" is included.
+        let (ug, rg) = make_grants(&[(ALICE, "acmeCo/data/", Admin)], &[]);
+
+        let result = authorized_prefixes(&rg, &ug, ALICE, Admin, Some("acmeCo/"));
+        assert_eq!(result, vec!["acmeCo/data/"]);
+    }
+
+    #[test]
+    fn filter_excludes_non_overlapping() {
+        let (ug, rg) = make_grants(&[(ALICE, "acmeCo/", Admin), (ALICE, "other/", Admin)], &[]);
+
+        let result = authorized_prefixes(&rg, &ug, ALICE, Admin, Some("acmeCo/"));
+        assert_eq!(result, vec!["acmeCo/"]);
+    }
+
+    #[test]
+    fn no_grants_returns_empty() {
+        let (ug, rg) = make_grants(&[], &[]);
+
+        let result = authorized_prefixes(&rg, &ug, ALICE, Admin, None);
+        assert!(result.is_empty());
+    }
+
+    #[test]
+    fn includes_transitive_roles() {
+        // Alice has Admin on "acmeCo/", which transitively grants Write on "shared/".
+        let (ug, rg) = make_grants(
+            &[(ALICE, "acmeCo/", Admin)],
+            &[("acmeCo/", "shared/", Write)],
+        );
+
+        let result = authorized_prefixes(&rg, &ug, ALICE, Write, None);
+        assert_eq!(result, vec!["acmeCo/", "shared/"]);
+
+        // Admin threshold excludes the transitive Write grant.
+        let result = authorized_prefixes(&rg, &ug, ALICE, Admin, None);
+        assert_eq!(result, vec!["acmeCo/"]);
+    }
+
+    #[test]
+    fn omits_child_prefixes_covered_by_parent() {
+        // Alice has Admin on "acmeCo/" and a direct grant on "acmeCo/data/".
+        // The child prefix should be pruned.
+        let (ug, rg) = make_grants(
+            &[(ALICE, "acmeCo/", Admin), (ALICE, "acmeCo/data/", Admin)],
+            &[],
+        );
+
+        let result = authorized_prefixes(&rg, &ug, ALICE, Admin, None);
+        assert_eq!(result, vec!["acmeCo/"]);
+    }
+
+    #[test]
+    fn omits_transitive_child_prefixes() {
+        // Alice has Admin on "acmeCo/", which grants Write on "acmeCo/team/".
+        // "acmeCo/team/" is already covered by "acmeCo/" so it should be pruned.
+        let (ug, rg) = make_grants(
+            &[(ALICE, "acmeCo/", Admin)],
+            &[("acmeCo/", "acmeCo/team/", Write)],
+        );
+
+        let result = authorized_prefixes(&rg, &ug, ALICE, Write, None);
+        assert_eq!(result, vec!["acmeCo/"]);
+    }
+
+    #[test]
+    fn different_user_sees_nothing() {
+        let bob = uuid::Uuid::from_bytes([0x22; 16]);
+        let (ug, rg) = make_grants(&[(ALICE, "acmeCo/", Admin)], &[]);
+
+        let result = authorized_prefixes(&rg, &ug, bob, Read, None);
+        assert!(result.is_empty());
+    }
+}
--- a/crates/control-plane-api/src/server/public/graphql/filters.rs
+++ b/crates/control-plane-api/src/server/public/graphql/filters.rs
@@ -1,0 +1,9 @@
+#[derive(Debug, Clone, Default, async_graphql::InputObject)]
+pub struct BoolFilter {
+    pub eq: Option<bool>,
+}
+
+#[derive(Debug, Clone, Default, async_graphql::InputObject)]
+pub struct PrefixFilter {
+    pub starts_with: Option<String>,
+}
--- a/crates/control-plane-api/src/server/public/graphql/invite_links.rs
+++ b/crates/control-plane-api/src/server/public/graphql/invite_links.rs
@@ -1,0 +1,1121 @@
+use super::filters;
+use async_graphql::{Context, types::connection};
+
+/// An invite link that grants access to a catalog prefix.
+#[derive(Debug, Clone, async_graphql::SimpleObject)]
+pub struct InviteLink {
+    /// The secret token for this invite link.
+    pub token: uuid::Uuid,
+    /// The catalog prefix this invite link grants access to.
+    pub catalog_prefix: models::Prefix,
+    /// The capability level granted by this invite link.
+    pub capability: models::Capability,
+    /// Whether this invite link can only be used once.
+    pub single_use: bool,
+    /// Optional description of this invite link.
+    pub detail: Option<String>,
+    /// When this invite link was created.
+    pub created_at: chrono::DateTime<chrono::Utc>,
+}
+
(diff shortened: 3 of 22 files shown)


crates/control-plane-api/src/server/public/graphql/filters.rs
```diff
@@ -0,0 +1,6 @@
+/// Reusable filter input types for GraphQL queries.
+
+#[derive(Debug, Clone, Default, async_graphql::InputObject)]
+pub struct BoolFilter {
```
GregorShear, line 2: Wanting to start a parallel discussion here about standardizing optional filters that could eventually be exposed in the UI. Bool is the simplest example, here are a couple others that would follow the pattern:

```rust
#[derive(Debug, Clone, Default, async_graphql::InputObject)]
pub struct StringFilter {
    pub eq: Option<String>,
    pub ne: Option<String>,
    pub contains: Option<String>,
}

#[derive(Debug, Clone, Default, async_graphql::InputObject)]
pub struct IntFilter {
    pub eq: Option<i64>,
    pub ne: Option<i64>,
    pub gt: Option<i64>,
    pub gte: Option<i64>,
    pub lt: Option<i64>,
    pub lte: Option<i64>,
}
```

crates/control-plane-api/src/server/public/graphql/invite_links.rs
```diff
@@ -0,0 +1,675 @@
...
+#[derive(Debug, Clone, async_graphql::SimpleObject)]
+pub struct InviteLink {
+    /// The secret token for this invite link.
+    pub token: uuid::Uuid,
+    /// The catalog prefix this invite link grants access to.
+    pub catalog_prefix: models::Prefix,
+    /// The capability level granted by this invite link.
+    pub capability: models::Capability,
+    /// Whether this invite link can only be used once.
+    pub single_use: bool,
```
GregorShear, line 14: this replaces the `directives.uses_remaining` field
(and we'll delete consumed invites, instead of setting `uses_remaining = 0`)

crates/control-plane-api/src/server/public/graphql/invite_links.rs
```diff
@@ -0,0 +1,675 @@
+use super::filters;
+use async_graphql::{Context, types::connection};
+
+/// An invite link that grants access to a catalog prefix.
+#[derive(Debug, Clone, async_graphql::SimpleObject)]
+pub struct InviteLink {
+    /// The secret token for this invite link.
+    pub token: uuid::Uuid,
+    /// The catalog prefix this invite link grants access to.
+    pub catalog_prefix: models::Prefix,
```
GregorShear, line 10: related to the optional filters, i'd like these resolvers to do their own check of which prefixes the user can see, and then this `catalog_prefix` can be an optional filter to narrow the results further.

would love to talk this through w/ phil & joseph

GregorShear: The other thing i like about this is that we can define the operators and standardize the prefix filters that phil and i were chatting about a few weeks ago (`prefix` vs `prefixedBy` vs exactPrefixes` etc...)

```rust
#[derive(Debug, Clone, Default, async_graphql::InputObject)]
pub struct PrefixFilter {
    pub eq: Option<models::Prefix>,
    pub starts_with: Option<models::Prefix>,
}
```

so an example gql query might look like 
```gql
query links {
  inviteLinks(filter:  {
     singleUse:  {
        eq: true
     }
     catalogPrefix: {
        startsWith: "gregCo"
     }
  }) {
    ...
  }
}
```

crates/control-plane-api/src/server/public/graphql/invite_links.rs
```diff
@@ -0,0 +1,553 @@
...
+        let invite = match invite {
+            Some(row) => row,
+            None => {
+                return Err(async_graphql::Error::new("Invalid invite link"));
+            }
+        };
+
+        // Delete single-use invite links upon redemption.
+        if invite.single_use {
+            sqlx::query!("DELETE FROM internal.invite_links WHERE token = $1", token,)
```
GregorShear, line 260: just calling out new behavior - delete single-use invites when they're redeemed

crates/control-plane-api/src/server/public/graphql/invite_links.rs
```diff
@@ -0,0 +1,553 @@
...
+            single_use,
+            detail.as_deref(),
+        )
+        .fetch_one(&env.pg_pool)
+        .await?;
+
+        tracing::info!(
+            %catalog_prefix,
+            ?capability,
+            %claims.sub,
```
GregorShear, line 209: @jshearer i suppose this could be the record of who invited the user...

jshearer: We don't keep logs indefinitely. If it's not required for compliance then I'm 100% whatevs about tracking who invited whom, I was just spitballing features that the existing directive's implementation _could_ offer that we might care about

crates/control-plane-api/src/server/public/graphql/invite_links.rs
jshearer: Yeah... I both do and don't like this. Back when I wrote a very similar generalized graphql resolver filter for Narwhal's GraphQL API, the reason it was tractable at all was that the entities we were querying over were homogenous enough that there was one centralized place (the "query builder") where we could turn these filters (`startsWith`, `lt`, `eq`, `and`, `or`, etc) into a big `WHERE` clause which could then be applied to any query we wanted.

If we _can't_ do that, then it becomes a combinatorial nightmare of having to support every type of filter on every graphql resolver's individual query/queries.

jshearer: I will also note, I spent a _lot_ of time in there optimizing the monstrous queries it would generate. So we should probably not follow that same exact pattern here :P

crates/control-plane-api/src/server/public/graphql/filters.rs
jshearer: Like I was talking about below, unless we're applying these filters over a homogenous-enough surface area (i.e, it's not actually clear to me that the UI should be the same datagrid/table sort of thing for captures, collections, materializations, storage mappings, data planes, tenant members, etc), then I'd actually vote for keeping it specific to each resolver. Thoughts?

Or maybe keep these `StringFilter`/`IntFilter` reusable chunks, but only implement them as needed in the resolvers where they make sense?

supabase/migrations/20260305120000_invite_links.sql
```diff
@@ -0,0 +1,95 @@
...
+    execute function internal.mirror_grant_directive_to_invite_links();
+
+-- Reverse sync: when a directive is consumed (uses_remaining drops to 0) or deleted
+-- via the old path, remove the corresponding invite_links row.
+create or replace function internal.sync_directive_removal_to_invite_links()
+returns trigger as $$
+begin
+    if TG_OP = 'DELETE' then
+        delete from internal.invite_links where token = OLD.token;
+        return OLD;
```
psFried: From the postgres docs:
> The return value of a row-level trigger fired AFTER or a statement-level trigger fired BEFORE or AFTER is always ignored; it might as well be null.

I think it's preferable to return `null`, since otherwise it makes it seem like the return value is significant. Also applies to `mirror_grant_directive`

supabase/migrations/20260305120000_invite_links.sql
```diff
@@ -0,0 +1,95 @@
...
+end;
+$$ language plpgsql security definer;
+
+create trigger sync_directive_removal
+    after update of uses_remaining or delete on public.directives
+    for each row
+    execute function internal.sync_directive_removal_to_invite_links();
+
+-- Backfill existing unredeemed grant directives into invite_links.
+insert into internal.invite_links (token, catalog_prefix, capability, single_use, detail, created_at)
```
psFried, line 82: This migration needs to be wrapped in a transaction. Otherwise it's possible for a trigger to create an `invite_links` row prior to this statement running. Also just a good guideline to always use a `begin`/`commit`, since it alleviates needing to think about recovering from partial successes.

crates/control-plane-api/src/server/public/graphql/filters.rs
GregorShear: i think this is the right compromise - having a standardized pattern can only help, and how far we take the implementation depends on what's needed vs how painful it gets as it grows

crates/flow-client/control-plane-api.graphql
```diff
@@ -525,6 +529,65 @@ type InferredSchemaStatus {
...
+	The item at the end of the edge
+	"""
+	node: InviteLink!
+	"""
+	A cursor for use in pagination
+	"""
+	cursor: String!
+}
+
+input InviteLinksFilter {
```
psFried, line 587: I actually like this `_Filter` naming better, though the inconsistency with `_By` seems unfortunate. IMO if we're going to switch the convention, then it's probably worth trying to switch everything else over, too. I'm kinda on the fence as to whether it's worth while, though. What do you think?

crates/control-plane-api/src/server/public/graphql/filters.rs
psFried: I can definitely see the benefit of having consistent terminology for different predicates. Though a problem I foresee with the approach is that not all resolvers will be able to support all predicates. For example, for some we might only support `eq`, and others maybe only `startsWith`. I mostly just wouldn't want to have our graphql schema suggest that you could use a `contains` filter on liveSpecs names, for example, when we don't want to support such queries in the database.

I'm a little on the fence about `liveSpecs(by: { prefix: "acmeCo/" })` vs `liveSpecs(by: { catalogName: { startsWith: "acmeCo/" } })`. I do kinda like the latter, though IDK whether I'd like it better _enough_ to justify migrating existing resolvers or living with two different conventions. Maybe it'd help to understand what the migration process would look like? If we wanted to have some limited backward compatibility during a transition period, would that be pretty easy?

GregorShear: migration would be easy - we can deprecate the current param and add it to the filter section ... decide later when we think it's safe to remove the deprecated one.

crates/control-plane-api/src/server/public/graphql/filters.rs
```diff
@@ -0,0 +1,9 @@
...
+    pub eq: Option<bool>,
```
jshearer, line 3: After our convo yesterday, I'm good with `filters` as a bucket for all set-filtering operations, and reusable `*Filter` structs for the operators/operands. Since there will only ever be one operator for booleans, the `eq` feels redundant and this feels like it should just be: 

```rust
pub struct BoolFilter(Option<bool>)
```

crates/control-plane-api/src/server/public/graphql/invite_links.rs
```diff
@@ -0,0 +1,911 @@
...
+            .as_ref()
+            .and_then(|f| f.single_use.as_ref())
+            .and_then(|f| f.eq);
+        let prefix_starts_with = filter
+            .and_then(|f| f.catalog_prefix)
+            .and_then(|f| f.starts_with);
+
+        // Collect the prefixes this user has admin access to, optionally
+        // narrowed by the prefix filter.
+        let admin_prefixes: Vec<String> = tables::UserGrant::transitive_roles(
```
GregorShear: @psFried @jshearer how do you guys feel about this?

i'm pulling grants out of the snapshot, and then filtering those by the input prefix filter param, if provided. 

then i'm bailing out if that list of prefixes is greater than `MAX_PREFIXES` (set to 20 right now).

Is this the right approach? is that the right complexity threshold?

crates/control-plane-api/src/server/public/graphql/invite_links.rs
```diff
@@ -0,0 +1,911 @@
...
+                    r#"
+                SELECT
+                    token,
+                    catalog_prefix AS "catalog_prefix!: String",
+                    capability AS "capability!: models::Capability",
+                    single_use AS "single_use!: bool",
+                    detail,
+                    created_at AS "created_at!: chrono::DateTime<chrono::Utc>"
+                FROM internal.invite_links
+                WHERE catalog_prefix::text ^@ ANY($1)
```
GregorShear, line 118: here's the clause for invite prefix starts with any of the available prefixes from above

crates/control-plane-api/src/server/public/graphql/invite_links.rs
```diff
@@ -0,0 +1,911 @@
...
+
+        // Collect the prefixes this user has admin access to, optionally
+        // narrowed by the prefix filter.
+        let admin_prefixes: Vec<String> = tables::UserGrant::transitive_roles(
+            &env.snapshot().role_grants,
+            &env.snapshot().user_grants,
+            env.claims()?.sub,
+        )
+        .filter(|grant| grant.capability >= models::Capability::Admin)
+        .filter(|grant| {
```
jshearer: thinking about this bit: 
```rust
        .filter(|grant| {
            prefix_starts_with
                .as_deref()
                .is_none_or(|sw| grant.object_role.starts_with(sw))
        })
```
does that work? if `prefix_starts_with` is unset(`None`), then `is_none_or` will return… `true` I guess, and pass every field through the filter. Cool makes sense!

crates/control-plane-api/src/server/public/graphql/invite_links.rs
jshearer: 20 seems fine to me I think. The question is, does pulling the prefixes from the grant work here? Prefixes should change infrequently, and the only case where this would be a problem is if you gained an admin grant and tried to list all prefixes before the agent-api had refreshed its snapshot. But I... don't see that as being a big problem. I can't think of any meaningful read-after-write consistency issues here. @psFried can you think of any eventual-consistency issues here?

psFried: I think this logic seems fine, though I feel like it'd be worthwhile to factor out a separate function, similar to [`get_default_prefix_arguments`](https://github.com/estuary/flow/blob/6584c806799a8807be1bc7e82a2f9acb243a0824/crates/flowctl/src/lib.rs#L256) in flowctl.

psFried: To be explicit, the only eventual consistency issue I can foresee here would be if a user just got granted access to the tenant, then they might not see invlite links for it immediately. Seems fine

crates/control-plane-api/src/server/public/graphql/invite_links.rs
jshearer: Just confirming that we've seen issues with `^@` and indexes before. This PR adds the following index: 

```sql
create index idx_invite_links_catalog_prefix on internal.invite_links
    (catalog_prefix text_pattern_ops);
```

I am by no means a postgres expert, but just pattern matching on this from previous issues, @psFried thoughts on whether we should be using spgist here?

jshearer: Yeah I think we want this index instead: 

```sql
CREATE INDEX idx_invite_links_catalog_prefix ON internal.invite_links
    USING spgist ((catalog_prefix::text));
```

crates/control-plane-api/src/server/public/graphql/filters.rs
jshearer: We talked about this, and I'm fine with holding this opinion weakly: the argument is that keeping the pattern as `{filter: { <fieldName>: { <operator> : <value> }}}` it makes it simpler to implement generic field filtering components in the UI. I'm still not sure if we have enough commonality between any domain entity to make generic field filtering components in the UI make sense, but... /shrug

crates/control-plane-api/src/server/public/graphql/invite_links.rs
GregorShear: great catch, fixed.

crates/control-plane-api/src/server/public/graphql/invite_links.rs
```diff
@@ -0,0 +1,969 @@
...
+
+        let bob_edges = bob_list["data"]["inviteLinks"]["edges"]
+            .as_array()
+            .expect("should have edges");
+        assert_eq!(bob_edges.len(), 0, "bob should see no invite links");
+
+        // Filtering with a sub-prefix more specific than the admin grant returns
+        // no results, because the filter narrows which admin prefixes are considered
+        // and "aliceCo/data/" does not start-with-match the grant "aliceCo/".
+        let sub_prefix_list: serde_json::Value = server
```
jshearer, line 1067: Hmmmm... this feels surprising/wrong from an API standpoint. The `catalogPrefix` filter here should be filtering InviteLinks, not grants, right?

If I have an admin grant to `aliceCo/` (meaning I transitively have an admin grant to anything prefixed by `aliceCo/`), and there's an invite link with `catalogPrefix: aliceCo/data/foo`, I would expect a query for inviteLinks filtered on `{ catalogPrefix: { startsWith: "aliceCo/data/" } }` to return that link, no?

I think this is still doable with the current pattern/without exploding the query complexity too much. The grant filtering needs to work in both directions: keep a grant if the grant is a prefix of the filter value **or** if the filter value is a prefix of the grant. This preserves the `aliceCo/` admin prefix in `admin_prefixes`

GregorShear: great catch. this is fixed and added a couple tests to exercise it. thanks!

crates/control-plane-api/src/server/public/graphql/invite_links.rs
GregorShear: pulled this logic into `.../graphql/authorized_prefixes.rs`, and added a bunch of tests

crates/control-plane-api/src/server/public/graphql/authorized_prefixes.rs
```diff
@@ -0,0 +1,138 @@
...
+/// bidirectional check lets callers query with a filter that is either broader
+/// or narrower than their grants.
+pub(super) fn authorized_prefixes(
+    role_grants: &tables::RoleGrants,
+    user_grants: &tables::UserGrants,
+    user_id: uuid::Uuid,
+    min_capability: models::Capability,
+    prefix_filter: Option<&str>,
+) -> Vec<String> {
+    tables::UserGrant::transitive_roles(role_grants, user_grants, user_id)
```
psFried: Just to ask the question, should we seek to avoid returning prefixes that duplicate each other? Like, if the user has a grant to `acmeCo/` and also `acmeCo/nested/`, is it worth having this function only return `acmeCo/`? We do that in flowctl to avoid listing the same things multiple times, if they are included in multiple prefix filters

GregorShear: yep, good idea. made this change

PR #2746, opened 2026-03-06
dekaf: cap per-partition fetch reads with a configurable combined limit

jshearer:
Dekaf placed no upper bound on how much data it buffered per Fetch request. A consumer could send large `partition_max_bytes` values across many partitions and cause unbounded memory growth.

This adds a `--combined-partition-fetch-limit` / `COMBINED_PARTITION_FETCH_LIMIT` config (default 50MB) that caps the total bytes read across all partitions in a Fetch. The limit is divided evenly across all fetched partitions, and each partition's read target is capped at `min(partition_max_bytes, limit / num_partitions)`.

* `fetch.max_bytes` from the client remains ignored because it is meant as an after-the-fact cap on response size, not a pre-flight allocation. Dividing it across partitions would starve active partitions when most partitions in the request are idle, which is a common pattern when consumers subscribe to many partitions but only a few have data.
  * In the future, I could see logic that 
* `max_wait_ms` is not capped since bounded bytes prevent resource overcommit regardless of wait duration
* Adds `dekaf_fetch_partition_bytes` histogram to track the actual byte size of each partition's fetch response

### DIFF
--- a/crates/dekaf/src/main.rs
+++ b/crates/dekaf/src/main.rs
@@ -142,6 +142,14 @@ pub struct Cli {
     /// can be up to 130K in size.
     #[arg(long, env = "READ_BUFFER_CHUNK_LIMIT", default_value = "20")]
     read_buffer_chunk_limit: usize,
+    /// Total byte budget for all partitions in a single Fetch request.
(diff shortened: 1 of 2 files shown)


jgraettinger (approved):
LGTM!

PR #2750, opened 2026-03-08
docs: guide for multiple bindings on file source connectors

jwhartley:
## Summary

- **New guide**: `guides/flowctl/multiple-file-source-bindings.md` — documents how to capture from multiple paths/folders within a single file source connector using flowctl
- **Google Drive connector docs fix**: Replaced incorrect `file_id`/`path` binding properties with `stream`, added missing `folderUrl` endpoint property, documented the `/u/N/` URL format caveat
- **Cross-references**: Added tip boxes linking to the new guide on S3, GCS, SFTP, and Google Drive connector pages

## Context

File source connectors all support multiple bindings at runtime, but the UI/discovery only creates a single binding. This leads to confusion — users think multiple paths require multiple captures. This guide covers the flowctl workflow for adding extra bindings, with two options (same collection vs. separate collections).

## Tested

Verified end-to-end with real captures:
- **SFTP**: Created capture with one binding → published → added second binding for a different directory → re-published → confirmed both directories captured into separate collections
- **Google Drive**: Created capture via UI with one folder → pulled specs → added second folder binding → re-published → confirmed both folders captured into separate collections

### DIFF
--- a/site/docs/guides/flowctl/multiple-file-source-bindings.md
+++ b/site/docs/guides/flowctl/multiple-file-source-bindings.md
@@ -4,7 +4,7 @@ sidebar_position: 4
 
 # Capture Multiple Paths with File Source Connectors
 
-File source connectors like Amazon S3, Google Cloud Storage, SFTP, Google Drive, Azure Blob Storage, Dropbox, and HTTP File all support capturing from multiple paths within a single capture task. However, the Estuary web app only creates a single binding during initial setup. To add additional paths, you can use flowctl to manually configure extra bindings.
+File source connectors like Amazon S3, Google Cloud Storage, SFTP, Google Drive, Azure Blob Storage, Dropbox, and HTTP File all support capturing from multiple paths within a single capture task. However, the Estuary web app only creates a single binding during initial setup. To add additional paths, you can use flowctl or the Advanced Specification Editor to manually configure extra bindings.
 
 This is useful when you need to:
 
@@ -14,11 +14,13 @@ This is useful when you need to:
 ## Prerequisites
 
 - An existing file source capture created through the Estuary web app.
-- [flowctl installed and authenticated](/guides/get-started-with-flowctl).
+- [flowctl installed and authenticated](/guides/get-started-with-flowctl) (for Options A and B via CLI).
 
 ## Option A: Multiple paths into the same collection
 
 Add extra bindings that capture from different paths but write to the same target collection.
+
+You can also make this change directly in the web app using the **Advanced Specification Editor** instead of flowctl. Edit your capture, open the Advanced Specification Editor, and add new entries to the `bindings` array with a different `stream` value but the same `target` collection. This only works for Option A — Option B requires creating new collections via flowctl.
 
 ### 1. Pull the capture specification
 
--- a/site/docs/reference/Connectors/capture-connectors/azure-blob-storage.md
+++ b/site/docs/reference/Connectors/capture-connectors/azure-blob-storage.md
@@ -1,4 +1,3 @@
-
 # Azure Blob Storage
 
 This connector captures data from an Azure Blob Storage Account.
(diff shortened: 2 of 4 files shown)


site/docs/guides/flowctl/multiple-file-source-bindings.md
```diff
@@ -0,0 +1,249 @@
+---
+sidebar_position: 4
+---
+
+# Capture Multiple Paths with File Source Connectors
+
+File source connectors like Amazon S3, Google Cloud Storage, SFTP, Google Drive, Azure Blob Storage, Dropbox, and HTTP File all support capturing from multiple paths within a single capture task. However, the Estuary web app only creates a single binding during initial setup. To add additional paths, you can use flowctl to manually configure extra bindings.
```
aeluce: Since you mention Dropbox here, it looks like the Dropbox docs currently say: `This connector is designed for files located in a specific Dropbox folder.`

Would it be helpful to update this language to make it clear that multiple folders can be configured?

site/docs/guides/flowctl/multiple-file-source-bindings.md
```diff
@@ -0,0 +1,249 @@
...
+  evolveIncompatibleCollections: true
+```
+
+:::tip
+In the web app, this corresponds to unchecking **Automatically add new collections** while keeping **Automatically keep schemas up to date** checked.
+:::
+
+### 3. Add a new binding
+
+In the `bindings` array, add a new entry with a different `stream` value but the same `target` collection:
```
aeluce, line 49: Would the Advanced Specification Editor work for this use case as well? We could briefly mention it at the beginning of Option A.

site/docs/reference/Connectors/capture-connectors/google-drive.md
```diff
@@ -21,16 +27,19 @@ The values and specification sample below provide configuration details specific
...
+|--------------------|----------------|-----------------------------------------------------------------------------------------------------------------------------|--------|------------------|
+| **`/folderUrl`**   | Folder URL     | URL of the Google Drive folder to capture from. Must be `https://drive.google.com/drive/folders/FOLDER_ID`. If your URL contains `/u/0/` or `/u/1/` (from Google's account switcher), remove that segment. | string | Required         |
+| **`/credentials`** | Credentials    | Google OAuth2 credentials or service account JSON for authentication.                                                        | object | Required         |
+| `/matchKeys`       | Match Keys     | Filter applied to file paths under the folder. If provided, only files whose paths match this regex will be read.            | string |                  |
+| `/parser`          | Parser Configuration | Configures how files are parsed (optional).                                                                             | object |                  |
```
aeluce, line 35: Technically, we could include properties for the `credentials` and `parser` objects here as well (`/credentials/auth_type`, `/parser/compression`, etc).

But the parser config especially is complex enough and shared by enough different connectors that we may want to give it its own page instead sometime (no action needed for this PR).

site/docs/guides/flowctl/multiple-file-source-bindings.md
jwhartley: Good catch — updated the Dropbox docs. Changed "This connector is designed for files located in a specific Dropbox folder" and "captures the data within the specified Dropbox folder into a single Estuary collection" to reflect that multiple folders are supported. Also added a tip box cross-referencing the new guide, matching the other connector pages.

While I was at it, added the same cross-reference tip boxes to Azure Blob Storage and HTTP File connector docs too — those were the remaining file source connectors that didn't have them.

site/docs/guides/flowctl/multiple-file-source-bindings.md
jwhartley: Good call — yes it does! I tested it with Google Drive: edited a capture via the Advanced Spec Editor, added a second binding with a different folder ID and the same target collection, and it saved and published successfully.

Added a paragraph at the top of Option A mentioning the Advanced Spec Editor as a lighter-weight alternative to flowctl. It also notes that this only works for Option A (same collection) — Option B requires creating new collections, which must be done via flowctl.

site/docs/reference/Connectors/capture-connectors/google-drive.md
jwhartley: Agreed — a dedicated parser config page would be good as a follow-up. The parser config is basically identical across all file source connectors so it makes sense to document it once rather than repeating it on every connector page. No action for this PR.

PR #2752, opened 2026-03-09
agent: detect and disable abandoned tasks

### LINKED ISSUE #2710: agent: detect and disable abandoned tasks
We have a bunch of abandoned tasks [^1]. We want to figure out a way to deal with them. In order to do that, we need to:
* Identify when a task is abandoned
* Notify the owner of the task that it's abandoned, and give them some time to remedy the situation
* Disable the abandoned task

**Out of scope** for this piece of work are tasks that are running but haven't moved any data, dealing with the collections of abandoned tasks, and subsequently deleting the disabled abandoned tasks.

## Proposed UX

1. A user publishes a database capture on day 1 of their POC
2. On day 3, they change the firewall rules so that database is no longer reachable
3. The task starts failing, and the control-plane diligently restarts it according to the task failure backoff strategy.
4. On day 3+N (14 maybe?), they get an email saying something like "Your task has been failing for N days. We've tried restarting it Q times without success. It will be disabled in M (maybe 7?) days from now. <links to troubleshooting or whatever>"
5. On day 3+N+M, the control-plane will create a publication that disables that task. 
## Definition

A task counts as "abandoned" when:
- It's enabled
- It is a type of entity that has shards (captures, materializations, derivations)
  - **Note:** This excludes all Dekaf materializations, which is fine because they shouldn't be incurring the same kind of control-plane overhead as actual abandoned tasks do anyway.
- No shard has been PRIMARY since a configurable threshold of days ago. When PRIMARY has _never_ been observed, `created_at` is used as the fallback, so new tasks aren't flagged until they've existed for the full threshold.

## Approach

We'll use the controller-evaluated alert path (same as `ShardFailed`, `AutoDiscoverFailed`, `BackgroundPublicationFailed`). The per-spec controller already does periodic shard listings, so we already have most of the data we need.

One more piece of data is needed:

- `last_primary_ts: Option<DateTime<Utc>>` on `ActivationStatus`. Set to `now()`[^2] whenever `aggregate_shard_status()` returns `Ok` during a health check. No new data-plane calls, this piggy-backs on the existing `list_task_shards()`. Starts as `None` for existing tasks and updates with the normal health-check cadence.

Then, we'll add a new `evaluate_abandoned()` function and wire it into the capture, materialization, and collection codepaths after the activation stage. The controller runs the full pipeline on every wake, and broken tasks still wake regularly as e.g shard health checks back off to ~60min [^3]. Derivations are collections, so they're covered by the collection pipeline. We'll gate with `has_task_shards()` which returns `false` for plain collections, Dekaf tasks, disabled tasks etc. I'll call `set_alert_firing(TaskAbandoned, ...)` or `resolve_alert(TaskAbandoned)` based on whether the task meets the criteria above.

**Concept:** Would it also be worth tracking the number of times the controller has restarted a failing task, since it was last `PRIMARY`? That way we could include in the email some text like "We have attempted to restart your task 826 times since it started failing 14 days ago.". This would also be a useful metric to have in general. 

## Rollout

We want to roll this out in stages.

**Detection only.** Fire alerts into `alert_history` but don't notify anyone. Don't add `task_abandoned` to default `include_alert_types` in `alert_subscriptions`. Deploy with a generous threshold (30 days) and let `last_primary_ts` accumulate. Monitor `alert_history` for `task_abandoned` entries and manually review what gets flagged.

**Notification.** Once we're confident the detection is flagging the right tasks, add `task_abandoned` to default `include_alert_types`, and probably run a migration to opt existing tenants in as well. By this point we'll need a real email template describing what "abandoned" means and what the user should do about it.

**Auto-disable.** When `TaskAbandoned` has been firing for longer than `ABANDONED_TASK_DISABLE_AFTER` (default 7 days), the controller creates a background publication setting `shards.disable = true`.

- On success, the task is disabled, `is_enabled_task()` returns false on the next run, the alert resolves, and the controller stops scheduling health checks. 
  - **Note:** Make sure this doesn't send a "your task is no longer abandoned" alert resolution email when we disable lol.
- On failure, the alert stays firing and the controller retries on its next wake.

## (Un)answered questions

**Data-plane outage**: if the data-plane that a task is running on goes down, what happens? Will the controller handle not being able to make shard listings? Will the controller handle not being able to publish the disablement? I imagine as designed, we'll sit there trying over and over to disable these tasks.
  * Answer from talking with @psFried: yep, we don't have a good story around data-plane deletions. currently this will get stuck in a loop trying to activate the disablement or deletion unless support manually removes all tasks from a data-plane before removing it.

**Existing abandoned tasks:** As designed, this will apply to both existing and new abandoned tasks, after the threshold window has passed. That's probably what we want, just calling it out

**How do we want to communicate:** Do we want an email after X number of days without `PRIMARY`, then disable after Y? Multiple emails? Do we want a resolution email? etc

**Windows:** What are we thinking WRT the thresholds here? 14 days to warning email, then another week until disable? 

[^1]: How many? 
[^2]: Is `now()` correct? Do we have some other concept of time to worry about here?
———
**Answer:** The controller uses `control_plane.current_time()` rather than raw `now()`. In production this is just `Utc::now()`, but the indirection exists so that integration tests can use a controllable clock.
[^3]: Is this correct? I want to make sure I/we understand how often this automation will run. I feel like it doesn't _need_ to run more than roughly once per day, but more often also shouldn't hurt. 
———
**Answer:** the controller picks the soonest of the reported wake-at times from the sub-tasks, so `evaluate_abandoned()` will just report whatever interval it cares about, like 24h, and that will guarantee that it gets woken no later than that.

jshearer:
## Summary

Implements abandoned task detection from #2710. Two alert sequences identify tasks that are persistently failing or not processing data, warn the user, and then disable the task after a grace period. Auto-disable publication is gated behind `DISABLE_ABANDONED_TASKS` env var (off by default).

## Sequence 1: Chronically Failing tasks

`TaskChronicallyFailing` fires when all of:
- `has_task_shards` is true (enabled, non-Dekaf task with shards)
- `ShardFailed` has been continuously firing >= 30 days (`CHRONICALLY_FAILING_THRESHOLD`)
- No user publication within 14 days (`ABANDON_USER_PUB_RECENCY`)

Since `ShardFailed` only resolves after 2 hours of continuous healthy operation, the 30-day condition means the task has not stayed alive for 2 consecutive hours at any point during that window.

After 7 days (`CHRONICALLY_FAILING_DISABLE_AFTER`), `TaskAutoDisabledFailing` fires and the task is disabled.

Resolves when `ShardFailed` resolves, a user publishes, or the task is disabled. A user publication resets a 14-day suppression window before the alert can re-fire. `ShardFailed.first_ts` (the 30-day clock) is unaffected by publications and only resets when the task recovers for 2+ hours. However, when `TaskChronicallyFailing` resolves and later re-fires, its own `first_ts` (which anchors the 7-day disable grace period) resets to the new fire time.

## Sequence 2: Idle tasks (no data movement)

`TaskIdle` fires when all of:
- `has_task_shards` is true
- No `ShardFailed` or `TaskChronicallyFailing` alert is active
- No data movement in `catalog_stats_daily` within 30 days (`ABANDON_IDLE_THRESHOLD`)
- No user publication within 14 days (`ABANDON_USER_PUB_RECENCY`)

After 7 days (`ABANDON_IDLE_DISABLE_AFTER`), `TaskAutoDisabledIdle` fires and the task is disabled.

Idle detection is suppressed while failure alerts are active. A task that is both failing and idle follows the first sequence only, so users only receive one set of notifications.

## Scenarios

**Task with intermittent 3-hour runs:** Shard reaches PRIMARY, runs for 3 hours, then fails. Each cycle, `ShardFailed` resolves (2+ hours healthy) then re-fires, resetting `ShardFailed.first_ts`. `TaskChronicallyFailing` never fires. This is intentional: the task is doing some work each cycle. If it doesn't move any data during the period where it's `PRIMARY`, the `TaskIdle` alert will take over.

**Healthy task that never moves data:** A task stays in `PRIMARY` with no failures, but no data flows through it (e.g. a Google Sheets capture where the sheet is empty). No `ShardFailed` fires, so the chronically-failing path is never involved. After 30 days with no data movement and no user publication in the preceding 14 days, `TaskIdle` fires. After 37 days total, the task is auto-disabled.

**Failing task that also has no data movement:** A task starts erroring and never moves data. `ShardFailed` fires within hours (after 3 failures), which immediately suppresses `TaskIdle` for as long as it remains active. Even though the task meets all other idle conditions, `TaskIdle` cannot fire while `ShardFailed` or `TaskChronicallyFailing` exists in the alert map. The task follows the chronically-failing path: `TaskChronicallyFailing` at 30 days, auto-disable at 37. If the task later recovers and `ShardFailed` resolves but still has no data movement, idle evaluation resumes and `TaskIdle` can fire on the next evaluation.

**Failing task where user publishes a fix attempt:** `TaskChronicallyFailing` fires on day 30. User publishes on day 31, which resolves the alert silently. If the task continues failing, the alert re-fires on day 45 (when the publication ages past 14 days) with a fresh `TaskChronicallyFailing.first_ts` and a new 7-day grace period. `ShardFailed.first_ts` remains day 0 throughout. The user publication bought 21 extra days total (14 + 7).

**Task re-enabled after auto-disable:** All abandon alerts were silently resolved on disable. On re-enablement, `has_task_shards` returns true and evaluation starts from scratch. For failures, `ShardFailed.first_ts` restarts from zero, so the user has at least 30 + 7 = 37 days before auto-disable. For idle, the re-enablement publication suppresses alerts for 14 days (it updates `last_user_pub_at`), then `TaskIdle` can fire if no data has moved. That gives 14 + 7 = 21 days before auto-disable.

## Rollout

On merge, controllers immediately begin evaluating the four new alert types on every wake cycle. `alert_history` rows will be created for tasks that meet the firing conditions. However, no emails will be sent and no tasks will be disabled:

**No emails:** The `fetch_recipients` query matches `alert_type = any(include_alert_types)` against `alert_subscriptions`. The new alert types are not in `DEFAULT_ALERT_TYPES` (the list applied when creating subscriptions during onboarding), and no existing subscription rows include them. The `alert_history` rows will have empty `recipients` arrays, so the notification task will produce zero emails.

**No disablement:** `DISABLE_ABANDONED_TASKS` defaults to `false`. The controller evaluates alerts and computes `DisableReason`, but skips the `maybe_disable_task` publication.

### Monitoring

```sql
select catalog_name, alert_type, fired_at, arguments
from alert_history
where alert_type in (
  'task_chronically_failing', 'task_auto_disabled_failing',
  'task_idle', 'task_auto_disabled_idle'
)
order by fired_at desc;
```

### Enabling notifications

Two steps, which can be done independently:

1. **New tenants:** Add the four alert types to `DEFAULT_ALERT_TYPES`. New subscriptions created during onboarding will then include them.

2. **Existing tenants:** Run a migration to append the new types to `include_alert_types` for existing `alert_subscriptions` rows:

```sql
update alert_subscriptions
set include_alert_types = include_alert_types
  || '{task_chronically_failing,task_auto_disabled_failing,task_idle,task_auto_disabled_idle}'::alert_type[]
where not include_alert_types @> '{task_chronically_failing}'::alert_type[];
```

### Enabling disablement

Set `DISABLE_ABANDONED_TASKS=true` in the agent env. This should be done >= 7 days after subscribing everyone to the above notifications, otherwise we might disable tasks that didn't get a warning email first.

### Environment variable knobs

| Variable | Default | Description |
|---|---|---|
| `CHRONICALLY_FAILING_THRESHOLD` | 30 days | How long `ShardFailed` must be continuously firing before `TaskChronicallyFailing` fires |
| `CHRONICALLY_FAILING_DISABLE_AFTER` | 7 days | Grace period between `TaskChronicallyFailing` firing and `TaskAutoDisabledFailing` firing (auto-disable) |
| `ABANDON_IDLE_THRESHOLD` | 30 days | How long without data movement (in `catalog_stats_daily`) before `TaskIdle` can fire |
| `ABANDON_USER_PUB_RECENCY` | 14 days | How long without a user publication before either sequence can fire. A user publication within this window suppresses both `TaskChronicallyFailing` and `TaskIdle` |
| `ABANDON_IDLE_DISABLE_AFTER` | 7 days | Grace period between `TaskIdle` firing and `TaskAutoDisabledIdle` firing (auto-disable) |
| `DISABLE_ABANDONED_TASKS` | `false` | When `true`, the controller publishes `shards.disable = true` when an auto-disable alert fires. When `false`, the alerts still fire and appear in `alert_history`, but no publication occurs |
| `ABANDON_CHECK_INTERVAL` | 24 hours | Minimum interval between abandonment evaluations for a given task. Reduces `catalog_stats_daily` query load by skipping re-evaluation when the last check was recent |

Related vars in the activation controller (not introduced here, but relevant to understanding the chronically-failing sequence):
| Variable | Default | Description |
|---|---|---|
| `ALERT_AFTER_SHARD_FAILURES` | 3 | Number of recent shard failures within the retention window needed to fire `ShardFailed` |
| `RESOLVE_SHARD_FAILED_ALERT_AFTER` | 2 hours | Duration of continuous `Ok` shard status needed to resolve `ShardFailed` |

Closes #2710

### DIFF
--- a/crates/agent/src/controllers/abandon.rs
+++ b/crates/agent/src/controllers/abandon.rs
@@ -1,0 +1,1157 @@
+use super::{ControllerState, NextRun, alerts};
+use crate::controllers::activation::has_task_shards;
+use crate::controllers::publication_status::PendingPublication;
+use crate::controlplane::ControlPlane;
+use chrono::{DateTime, Utc};
+use models::status::publications::PublicationStatus;
+use models::status::{AlertType, Alerts};
+
+env_config_interval! {
+    /// ShardFailed must be continuously firing for this long before we consider the task chronically failing.
+    CHRONICALLY_FAILING_THRESHOLD, chrono::Duration::days(30)
+}
+env_config_interval! {
+    /// Grace period after firing the chronically-failing warning before auto-disabling.
+    CHRONICALLY_FAILING_DISABLE_AFTER, chrono::Duration::days(7)
+}
+env_config_interval! {
+    /// Task must have no data movement for this long before we consider it idle.
+    IDLE_THRESHOLD, chrono::Duration::days(30)
+}
+env_config_interval! {
+    /// A user publication within this timeframe will prevent task abandonment alerts from firing and tasks from being disabled.
+    USER_PUB_THRESHOLD, chrono::Duration::days(14)
+}
+env_config_interval! {
+    /// Grace period after firing the idle warning before auto-disabling.
+    IDLE_DISABLE_AFTER, chrono::Duration::days(7)
+}
+env_config_bool! {
+    /// Allows for running the abandonment controller without actually disabling tasks.
+    pub(crate) DISABLE_ABANDONED_TASKS, false
+}
+
+/// Extra field keys stored in alert `extra` maps for notification templates.
+const EXTRA_DISABLE_AT: &str = "disable_at";
+const EXTRA_FAILING_SINCE: &str = "failing_since";
+
+/// Why a task is being auto-disabled, used as the publication detail message.
+#[derive(Debug, PartialEq, Eq)]
+enum DisableReason {
+    ChronicallyFailing,
+    Idle,
+}
+
+#[derive(Clone, Copy, Default)]
+struct AbandonmentTimestamps {
+    last_data_movement_ts: Option<DateTime<Utc>>,
+    last_user_pub_at: Option<DateTime<Utc>>,
+}
+
+impl std::fmt::Display for DisableReason {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::ChronicallyFailing => f.write_str("auto-disabling chronically failing task"),
+            Self::Idle => f.write_str("auto-disabling idle task"),
+        }
+    }
+}
+
+pub async fn evaluate_abandoned<C: ControlPlane>(
+    alerts_status: &mut Alerts,
+    publications: &mut PublicationStatus,
+    state: &ControllerState,
+    control_plane: &C,
+) -> anyhow::Result<Option<NextRun>> {
+    // Tasks without shards (disabled, Dekaf, etc.) don't need abandon monitoring.
+    if !has_task_shards(state) {
+        resolve_all_abandon_alerts(alerts_status);
+        return Ok(None);
+    }
+
+    let now = control_plane.current_time();
+    let timestamps = fetch_abandonment_timestamps(alerts_status, state, control_plane, now).await?;
+
+    let disable_reason = evaluate_alerts(alerts_status, state, now, timestamps);
+
+    if let Some(reason) = disable_reason {
+        if DISABLE_ABANDONED_TASKS.get() {
+            if maybe_disable_task(reason, state, publications, control_plane).await? {
+                return Ok(Some(NextRun::immediately()));
+            }
+        }
+    }
+
+    // Safety-net wake: controllers already wake every ~2 hours for shard health
+    // checks, so this 24-hour timer just ensures abandon evaluation isn't skipped
+    // indefinitely if those checks stop scheduling wakes.
+    Ok(Some(NextRun::after_minutes(24 * 60)))
+}
+
+async fn fetch_abandonment_timestamps<C: ControlPlane>(
+    alerts_status: &Alerts,
+    state: &ControllerState,
+    control_plane: &C,
+    now: DateTime<Utc>,
+) -> anyhow::Result<AbandonmentTimestamps> {
+    let last_user_pub_at = control_plane
+        .fetch_last_user_publication_at(state.live_spec_id)
+        .await?;
+
+    let has_failure_alerts = alerts_status.contains_key(&AlertType::ShardFailed)
+        || alerts_status.contains_key(&AlertType::TaskChronicallyFailing);
+    let user_pub_stale = last_user_pub_at.map_or(true, |ts| (now - ts) >= USER_PUB_THRESHOLD.get());
+
+    let last_data_movement_ts = if has_failure_alerts || !user_pub_stale {
+        None
+    } else {
+        control_plane
+            .fetch_last_data_movement_ts(state.catalog_name.clone())
+            .await?
+    };
+
+    Ok(AbandonmentTimestamps {
+        last_data_movement_ts,
+        last_user_pub_at,
+    })
+}
+
+/// Pure evaluation of abandon alert state. Returns `Some(reason)` when the
+/// grace period for an auto-disable sequence has expired and the task should
+/// be disabled, or `None` if no disable is needed.
+fn evaluate_alerts(
+    alerts_status: &mut Alerts,
+    state: &ControllerState,
+    now: DateTime<Utc>,
+    timestamps: AbandonmentTimestamps,
+) -> Option<DisableReason> {
+    // Tasks that are disabled, Dekaf, or lack shards don't get abandonment checks.
+    // Silently resolve any alerts that may have been firing so re-enablement starts clean.
+    let Some(spec) = state.live_spec.as_ref().filter(|_| has_task_shards(state)) else {
+        resolve_all_abandon_alerts(alerts_status);
+        return None;
+    };
+
+    let catalog_type = spec.catalog_type();
+
+    let user_pub_stale = timestamps
+        .last_user_pub_at
+        .map_or(true, |ts| (now - ts) >= USER_PUB_THRESHOLD.get());
+
+    // Sequence 1: Chronically Failing
+    // Fires when ShardFailed has been continuously active for > CHRONICALLY_FAILING_THRESHOLD
+    // and no user has published changes recently (a recent publication signals active debugging).
+    let shard_failed_since = alerts_status
+        .get(&AlertType::ShardFailed)
+        .map(|a| a.first_ts);
+
+    let is_chronically_failing = shard_failed_since
+        .is_some_and(|first_ts| (now - first_ts) >= CHRONICALLY_FAILING_THRESHOLD.get())
+        && user_pub_stale;
+
+    let mut disable_reason = None;
+
+    if is_chronically_failing {
+        let shard_failed_first_ts = shard_failed_since.unwrap();
+
+        // Only fire the alert once. On subsequent evaluations the stored
+        // disable_at and failing_since are the source of truth, so we don't
+        // shift the disable date if the env config changes between runs.
+        if !alerts_status.contains_key(&AlertType::TaskChronicallyFailing) {
+            let failing_since = shard_failed_first_ts.format("%Y-%m-%d").to_string();
+            let disable_at = (now + CHRONICALLY_FAILING_DISABLE_AFTER.get())
+                .format("%Y-%m-%d")
+                .to_string();
+
+            alerts::set_alert_firing(
+                alerts_status,
+                AlertType::TaskChronicallyFailing,
+                now,
+                format!("task shards have been failing since {failing_since}"),
+                1, // Each abandonment alert is only ever fired once.
+                catalog_type,
+            );
+            let alert = alerts_status
+                .get_mut(&AlertType::TaskChronicallyFailing)
+                .expect("just inserted");
+            alert
+                .extra
+                .insert(EXTRA_DISABLE_AT.to_string(), disable_at.into());
+            alert
+                .extra
+                .insert(EXTRA_FAILING_SINCE.to_string(), failing_since.into());
+        }
+
+        let alert = alerts_status
+            .get(&AlertType::TaskChronicallyFailing)
+            .expect("TaskChronicallyFailing alert was just set");
+        let disable_at = alert
+            .extra
+            .get(EXTRA_DISABLE_AT)
+            .expect("disable_at was just inserted")
+            .as_str()
+            .expect("disable_at must be a string")
+            .to_string();
+        let failing_since = alert
+            .extra
+            .get(EXTRA_FAILING_SINCE)
+            .expect("failing_since was just inserted")
+            .as_str()
+            .expect("failing_since must be a string")
+            .to_string();
+        let disable_at = chrono::NaiveDate::parse_from_str(&disable_at, "%Y-%m-%d")
+            .expect("disable_at was set from a valid date");
+
+        if now.date_naive() >= disable_at {
+            if !alerts_status.contains_key(&AlertType::TaskAutoDisabledFailing) {
+                alerts::set_alert_firing(
+                    alerts_status,
+                    AlertType::TaskAutoDisabledFailing,
+                    now,
+                    format!("task auto-disabled after shards failing since {failing_since}"),
+                    1,
+                    catalog_type,
+                );
+                alerts_status
+                    .get_mut(&AlertType::TaskAutoDisabledFailing)
+                    .expect("just inserted")
+                    .extra
+                    .insert(
+                        EXTRA_FAILING_SINCE.to_string(),
+                        failing_since.clone().into(),
+                    );
+            }
+            disable_reason = Some(DisableReason::ChronicallyFailing);
+        }
+    } else {
+        alerts::resolve_alert(alerts_status, AlertType::TaskChronicallyFailing);
+        alerts::resolve_alert(alerts_status, AlertType::TaskAutoDisabledFailing);
+    }
+
+    // Sequence 2: Idle Task
+    // Fires when no data has moved AND no user publication for extended periods.
+    // Suppressed when ShardFailed or TaskChronicallyFailing is active to avoid
+    // sending both "your task is failing" and "your task is idle" simultaneously.
+    let has_failure_alerts = alerts_status.contains_key(&AlertType::ShardFailed)
+        || alerts_status.contains_key(&AlertType::TaskChronicallyFailing);
+
+    let data_stale = timestamps
+        .last_data_movement_ts
+        .map_or(true, |ts| (now - ts) >= IDLE_THRESHOLD.get());
+
+    let is_idle = !has_failure_alerts && data_stale && user_pub_stale;
+
+    if is_idle {
+        if !alerts_status.contains_key(&AlertType::TaskIdle) {
+            let disable_at = (now + IDLE_DISABLE_AFTER.get())
+                .format("%Y-%m-%d")
+                .to_string();
+
+            alerts::set_alert_firing(
+                alerts_status,
+                AlertType::TaskIdle,
+                now,
+                format!(
+                    "task has not moved data{}",
+                    timestamps
+                        .last_data_movement_ts
+                        .map(|ts| format!(" since {}", ts.format("%Y-%m-%d")))
+                        .unwrap_or_else(|| " since it was created".to_string())
+                ),
+                1,
+                catalog_type,
+            );
+            alerts_status
+                .get_mut(&AlertType::TaskIdle)
+                .expect("just inserted")
+                .extra
+                .insert(EXTRA_DISABLE_AT.to_string(), disable_at.into());
+        }
+
+        let disable_at = alerts_status
+            .get(&AlertType::TaskIdle)
+            .expect("TaskIdle alert was just set")
+            .extra
+            .get(EXTRA_DISABLE_AT)
+            .expect("disable_at was just inserted")
+            .as_str()
+            .expect("disable_at must be a string");
+        let disable_at = chrono::NaiveDate::parse_from_str(disable_at, "%Y-%m-%d")
+            .expect("disable_at was set from a valid date");
+
+        if now.date_naive() >= disable_at {
+            if !alerts_status.contains_key(&AlertType::TaskAutoDisabledIdle) {
+                alerts::set_alert_firing(
+                    alerts_status,
+                    AlertType::TaskAutoDisabledIdle,
+                    now,
+                    "task auto-disabled due to inactivity".to_string(),
+                    1,
+                    catalog_type,
+                );
+            }
+            disable_reason = Some(DisableReason::Idle);
+        }
+    } else {
+        alerts::resolve_alert(alerts_status, AlertType::TaskIdle);
+        alerts::resolve_alert(alerts_status, AlertType::TaskAutoDisabledIdle);
+    }
+
+    disable_reason
+}
+
+/// Publishes the spec with `shards.disable = true`. Returns `true` if the
+/// publication succeeded, signaling the caller to return `NextRun::immediately()`
+/// so the controller re-enters and processes the now-disabled spec.
+async fn maybe_disable_task<C: ControlPlane>(
+    reason: DisableReason,
+    state: &ControllerState,
+    publications: &mut PublicationStatus,
+    control_plane: &C,
+) -> anyhow::Result<bool> {
+    let mut spec = state
+        .live_spec
+        .clone()
+        .expect("must have live spec to disable");
+
+    set_spec_disabled(&mut spec);
+    let mut pending = PendingPublication::update_model(
+        &state.catalog_name,
+        state.last_pub_id,
+        spec,
+        &reason.to_string(),
+    );
+
+    // `finish()` normally fires BackgroundPublicationFailed after repeated
+    // publication failures. Pass None to suppress that: if we're disabling
+    // an abandoned task, an additional alert about the disable publication
+    // failing is just noise. The failure is still recorded in publication
+    // history and we'll retry on the next controller run.
+    let result = pending
+        .finish(state, publications, None, control_plane)
+        .await?;
+    Ok(result.status.is_success())
+}
+
+fn set_spec_disabled(spec: &mut models::AnySpec) {
+    match spec {
+        models::AnySpec::Capture(c) => c.shards.disable = true,
+        models::AnySpec::Collection(c) => {
+            c.derive
+                .as_mut()
+                .expect("collection without derivation should not have task shards")
+                .shards
+                .disable = true;
+        }
+        models::AnySpec::Materialization(m) => m.shards.disable = true,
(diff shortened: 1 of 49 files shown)


crates/agent/src/controllers/activation.rs
```diff
@@ -19,42 +17,85 @@ use models::{
 };
 
 /// Helper for getting a `chrono::Duration` from an environment variable, using humantime so it
-/// supports parsing durations like `3h`.
+/// supports parsing durations like `3h`. The value is read once on first access. In test builds,
```
jshearer: This kind of spiraled a bit. Happy to split it out into a separate PR if anyone cares, but the thought process is:

* I needed a way to parse durations and booleans in `abandon.rs`, so I added `env_config_bool` as a copy of `env_config_interval`
* I needed a way to change these values at runtime for testing, so I added `EnvConfig` with a `#[cfg(test)] pub fn set` method
* I didn't like how `ALERT_AFTER_SHARD_FAILURES` and `SUSTAINED_PRIMARY_MIN_CHECKS` were all lonely by themselves so I made `env_config_num` and updated them to use it so that everything would be consistent

It's possible I could come up with some sort of meta-macro that reduces the duplication of these three, but macros are dense enough that I figured this is actually better.

crates/control-plane-api/src/controllers.rs
```diff
@@ -79,13 +82,32 @@ pub async fn fetch_controller_job(
...
         join controller_jobs cj on ls.id = cj.live_spec_id
         left outer join data_planes dp on ls.data_plane_id = dp.id
+        left join lateral (
+            select max(ts) as last_data_movement_ts
+            from catalog_stats_daily
+            where catalog_name = ls.catalog_name
+                and ts > now() - interval '60 days'
+                and (bytes_written_by_me + bytes_read_by_me
+                   + bytes_written_to_me + bytes_read_from_me) > 0
+        ) cdm on true
```
jshearer: ## Claude's optimization feedback

**`catalog_stats_daily` lateral join (last data movement)**

```sql
left join lateral (
    select max(ts) as last_data_movement_ts
    from catalog_stats_daily
    where catalog_name = ls.catalog_name
        and ts > now() - interval '60 days'
        and (bytes_written_by_me + bytes_read_by_me
           + bytes_written_to_me + bytes_read_from_me) > 0
) cdm on true
```

**Index coverage:** The PK `(catalog_name, grain, ts)` on the parent `catalog_stats` table works here. Since `catalog_stats_daily` is a LIST partition where `grain = 'daily'` is implicit, the effective index key is `(catalog_name, ts)`. The planner can do an index scan for `catalog_name = X` and range-scan `ts > now() - '60 days'`.

**Cardinality:** Max ~60 rows per catalog_name (one row per day, bounded by the 60-day window). This is well-bounded.

**Recommendation: Switch `max(ts)` to `ORDER BY ts DESC LIMIT 1`.** This allows a backward index scan with early termination. For active tasks (the common case), Postgres stops after examining 1-2 rows instead of aggregating all ~60. Minor but free improvement:

```sql
left join lateral (
    select ts as last_data_movement_ts
    from catalog_stats_daily
    where catalog_name = ls.catalog_name
        and ts > now() - interval '60 days'
        and (bytes_written_by_me + bytes_read_by_me
           + bytes_written_to_me + bytes_read_from_me) > 0
    order by ts desc
    limit 1
) cdm on true
```

**No new index needed.** The per-name cardinality is too low to justify one. The 60-day window is well-chosen (2x the 30-day idle threshold).

crates/control-plane-api/src/controllers.rs
```diff
@@ -79,13 +82,32 @@ pub async fn fetch_controller_job(
...
+        left join lateral (
+            select ps.published_at as last_user_pub_at
+            from publication_specs ps
+            where ps.live_spec_id = ls.id
+                and ps.user_id != $2
+            order by ps.pub_id desc
+            limit 1
+        ) lup on true
```
jshearer: ## Claude's optimization feedback

**`publication_specs` lateral join (last user publication)**

```sql
left join lateral (
    select ps.published_at as last_user_pub_at
    from publication_specs ps
    where ps.live_spec_id = ls.id
        and ps.user_id != $2
    order by ps.pub_id desc
    limit 1
) lup on true
```

**Index coverage:** The PK `(live_spec_id, pub_id)` supports the backward scan pattern well. Postgres seeks to the highest `pub_id` for the given `live_spec_id` and scans backward.

**Concern: `user_id != $2` requires heap fetches.** The `user_id` column isn't in the PK index, so each candidate row requires a table page access. For specs where the most recent publications are human-authored (common case), this stops after 1 heap fetch. Worst case: specs with many consecutive system-only publications require scanning through all of them.

**No new index recommended right now.** A partial index on `WHERE user_id != '<uuid>'` would be ideal but couples the index to a specific UUID. If this becomes a bottleneck, denormalizing `last_user_pub_at` into `live_specs` (updated during the publication path, which already touches that table) is the cleanest fix.

crates/control-plane-api/src/controllers.rs
jshearer: changed to `ORDER BY ts DESC LIMIT 1` 👍

crates/control-plane-api/src/controllers.rs
jshearer: left alone 👍

crates/agent/src/controllers/abandon.rs
```diff
@@ -0,0 +1,852 @@
...
+
+/// Why a task is being auto-disabled, used as the publication detail message.
+#[derive(Debug, PartialEq, Eq)]
+enum DisableReason {
+    ChronicallyFailing,
+    Idle,
+}
+
+#[derive(Clone, Copy, Default)]
+struct AbandonmentTimestamps {
```
jshearer, line 22: Still waffling on if i I like this struct or not. I think it's justified because the alternative is using an unnamed tuple, which is confusing.

crates/agent/src/controllers/abandon.rs
```diff
@@ -0,0 +1,852 @@
...
+) -> anyhow::Result<AbandonmentTimestamps> {
+    let last_user_pub_at = control_plane
+        .fetch_last_user_publication_at(state.live_spec_id)
+        .await?;
+
+    let has_failure_alerts = alerts_status.contains_key(&AlertType::ShardFailed)
+        || alerts_status.contains_key(&AlertType::TaskChronicallyFailing);
+    let user_pub_stale = last_user_pub_at.map_or(true, |ts| (now - ts) >= USER_PUB_THRESHOLD.get());
+
+    let last_data_movement_ts = if has_failure_alerts || !user_pub_stale {
```
jshearer, line 96: this complexity is all here just so we can short-circuit the query against `catalog_stats` if it's not needed. may be premature optimization...

crates/agent/src/controllers/abandon.rs
```diff
@@ -0,0 +1,852 @@
...
+    if is_chronically_failing {
+        let shard_failed_first_ts = shard_failed_since.unwrap();
+        let failing_since = shard_failed_first_ts.format("%Y-%m-%d").to_string();
+
+        alerts::set_alert_firing(
+            alerts_status,
+            AlertType::TaskChronicallyFailing,
+            now,
+            format!("task shards have been failing since {failing_since}"),
+            activation.recent_failure_count,
```
jshearer: is this the right thing to pass for `count` here?

crates/agent/src/controllers/abandon.rs
```diff
@@ -0,0 +1,960 @@
...
+
+    let mut disable_reason = None;
+
+    if is_chronically_failing {
+        let shard_failed_first_ts = shard_failed_since.unwrap();
+
+        // Only fire the alert once. On subsequent evaluations the stored
+        // disable_at and failing_since are the source of truth, so we don't
+        // shift the disable date if the env config changes between runs.
+        if !alerts_status.contains_key(&AlertType::TaskChronicallyFailing) {
```
jshearer, line 152: This is worth pointing out. All of the other alerts call `set_alert_firing` on every invocation of the controller. I didn't want a scenario where we send someone an email saying that we'll disable their task in a week, then decide to shorten that timespan and they get their task disabled sooner, for example. So I decided that the alert itself should control, rather than re-computing the required timeframe every time. Thoughts? Is there any other reason we need to be calling `alerts::set_alert_firing` every run?

crates/agent/src/controllers/abandon.rs
```diff
@@ -0,0 +1,1155 @@
...
+                .get_mut(&AlertType::TaskIdle)
+                .expect("just inserted")
+                .extra
+                .insert(EXTRA_DISABLE_AT.to_string(), disable_at.into());
+        }
+
+        let disable_at = alerts_status
+            .get(&AlertType::TaskIdle)
+            .expect("TaskIdle alert was just set")
+            .extra
```
jshearer, line 266: Would be really nice if these `extra` maps could be strongly typed, but that's very much out of scope of this PR

crates/agent/src/integration_tests/abandoned_tasks.rs
```diff
@@ -0,0 +1,386 @@
...
+    override_shard_status_last_ts(catalog_name, chrono::Duration::minutes(5), harness).await;
+    let _state = harness.run_pending_controller(catalog_name).await;
+
+    let alert = harness
+        .assert_alert_firing(catalog_name, AlertType::TaskChronicallyFailing)
+        .await;
+
+    let extra = &alert.alert.arguments.0;
+    assert!(
+        extra.get("disable_at").and_then(|v| v.as_str()).is_some(),
```
jshearer, line 160: I wanted to do an inline snapshot here but the timestamp would need to be omitted as it's not constant, so this is as good as I think we're going to get

crates/agent/src/controllers/activation.rs
psFried: You're gonna groan, but I think perhaps we should just abandon `env_config_interval` altogether, and just commit to only using our clap arguments. My original motivation for using `env_config_interval` was because I had low confidence that the vars I was introducing would be the "right" knobs, and so I didn't want to make them CLI arguments. But now, I think I regret that, because "in for a penny, in for a pound". So I think my recommendation would be to rip the bandaid now and only use clap args for everything. That way we get consistent `--help` docs for all these vars, and they're all documented in one place. And I'm sorry that you tripped over my dirty laundry here.

crates/agent/src/integration_tests/abandoned_tasks.rs
psFried: The only other thing I could think of would be to assert that it's within some % of a desired interval in the future. We do that in the `inferred_schemas` integration tests to test backoffs and such.

crates/models/src/status/alerts.rs
```diff
@@ -45,6 +45,37 @@ pub enum AlertType {
     /// continue to make progress in between failures, but at a minimum, performance will
     /// be degraded. And in many scenarios, the task will be unable to process data at all.
     ShardFailed,
+    /// Warning that a task has been unable to run for an extended period and will
+    /// be auto-disabled unless someone intervenes.
+    ///
+    /// Fires when all of:
+    /// - `has_task_shards` (enabled, non-Dekaf task with shards)
```
psFried: fyi, these doc comments can actually become user-facing. They're part of the graphql schema, and IIRC that's what the UI is using to render alert type descriptions.

crates/agent/src/controllers/abandon.rs
psFried: Why would we shorten it? I'm not sure I follow

Just to maybe clarify some things, `set_alert_firing` only updates the `alerts_status`, it won't result in duplicate alert emails being sent. The `alert_history` table get's updated as part of the controller update transaction [here](https://github.com/estuary/flow/blob/master/crates/agent/src/controllers/executor.rs#L211). Once we fire an alert for a given `(catalog_name, alert_type)`, we'll never fire another one or update it until it resolves. Meaning that the `alert_history.arguments` will always stay as they were when the first email was sent.

The meaning of incrementing the `count` is just that we've evaluated the alert condition again, and still think that it's firing. AFAIK we only ever use it in the shard failures alert email, and in others it's never rendered. That's why other controllers call `set_alert_firing` every time, just to record that we've looked at at and still think it should be firing.

psFried (changes requested):
I left some feedback here and asked for some changes. There's parts of the code that I'm still trying to fully understand, but wanted to get you whatever feedback I have at this point, in case there's anything we should talk about on VC today

crates/agent/src/controllers/activation.rs
jshearer: Groan groan, I think you're right. I will work on plumbing these through. Any reason you can think of for me to not also apply it to "your" knobs as well? In for a pound, in for two pounds lol

crates/models/src/status/alerts.rs
jshearer: Great point, I've updated them to be appropriately user-facing instead of internal.

crates/agent/src/integration_tests/abandoned_tasks.rs
jshearer: Yep, done 👍

crates/agent/src/controllers/activation.rs
jshearer: This is refactored to be fully clap-based now: https://github.com/estuary/flow/pull/2752/changes#diff-ff12110284dfd3b2eb0ceec777999e80805a22f48e94426d485c7bb60843f668R32-R87

crates/agent/src/controllers/abandon.rs
jshearer: > Why would we shorten it? I'm not sure I follow

If we decide that we want it shorter, I guess. It just makes sense to me that we set an alert firing with a particular scheduled date for disabling, and then the alert's metadata itself should control when we do that thing we said we're going to do. 

> The meaning of incrementing the count is just that we've evaluated the alert condition again, and still think that it's firing. AFAIK we only ever use it in the shard failures alert email, and in others it's never rendered. That's why other controllers call set_alert_firing every time, just to record that we've looked at at and still think it should be firing.

Is the intended pattern that whenever a controller runs, and an alert is firing, it should call `set_alert_firing` again? Is there any reason, other than updating `count` (which isn't used here)?

crates/agent/src/controllers/mod.rs
```diff
@@ -25,44 +24,68 @@ use std::fmt::Debug;
...
-            .unwrap()
-            .get(&std::thread::current().id())
-        {
-            return v;
-        }
-        self.value
-    }
+/// Configuration values for controller automations, parsed from CLI
+/// arguments and threaded through `ControlPlane`.
+#[derive(Copy, Clone, Debug, clap::Parser)]
```
psFried: This struct seems like it's a little large for `Copy`. Can we instead make `ControlPlane::controller_config` return either a `&ControllerConfig` or `Arc<ControllerConfig>`?

crates/agent/src/controllers/abandon.rs
psFried: > the alert's metadata itself should control when we do that thing we said we're going to do.

This is the part that I was feeling skeptical about, and now that I've looked through the code again, I'd like to withdraw the concern. We're already needing to look at alert state as part of this, so why not just use that alert state as the source of truth? Fair enough. And yeah, the `count` isn't something we need to update.

crates/agent/src/controllers/abandon.rs
psFried: IMO it's good to err on the side of structs with named fields 👍

crates/agent/src/controllers/abandon.rs
```diff
@@ -0,0 +1,1199 @@
...
+    control_plane: &C,
+) -> anyhow::Result<Option<NextRun>> {
+    // Tasks without shards (disabled, Dekaf, etc.) don't need abandon monitoring.
+    if !has_task_shards(state) {
+        resolve_all_abandon_alerts(alerts_status);
+        return Ok(None);
+    }
+
+    let now = control_plane.current_time();
+    let config = control_plane.controller_config();
```
psFried, line 50: I love `control_plane.controller_config()`

crates/agent/src/controllers/abandon.rs
```diff
@@ -0,0 +1,1199 @@
...
+    let timestamps =
```
psFried, line 64: A possible concern with this is that we could potentially end up with significantly more queries on `catalog_stats_daily`, which is already an area of concern (meaning with stats performance, more broadly). I think it might be worthwhile to limit how frequently we'll evaluate this controller, somehow. One thought would be to add either a `abandonmentLastChecked` or `nextAbandonmentCheck` timestamp to the status, and ensure we go at least some minimum time between checks. Seems like daily ought to be plenty frequent enough.

For auto-discovers, we added a `auto_discover_probability`, which could potentially also be useful here, though I tend to think that just a minimum interval between checks might be the best way to model it. WDYT? Did you do any testing of the `catalog_stats_daily` query in prod? Is this maybe not worth worrying about?

crates/agent/src/controllers/abandon.rs
psFried: I think this is a good idea to avoid querying catalog stats if we don't need to. See my other comment about limiting evaluation frequency 😬

psFried (commented):
I love the ControllerConfig refactor. I left one other comment about possibly limiting the evaluation frequency, which seems worth talking through before merging. And removing `Copy` from `ControllerConfig` seems worthwhile too.

crates/agent/src/controllers/mod.rs
jshearer: Good point, got rid of `Copy` here and swapped to `Arc`.

crates/agent/src/controllers/abandon.rs
jshearer: I added `ABANDON_CHECK_INTERVAL` defaulting to 24h to throttle down abandonment checking. It doesn't need to run more frequently than that, given the timescales involved are on the order of weeks.

psFried (approved):
LGTM

PR #2754, opened 2026-03-10
Reapply "control-plane-api: enforce strict $defs check in update_conn…

psFried:
…ector_schema"

This reverts commit 856a6c98e129f8946f0389d023d83b558d8b5098.

This brings back the changes from #2720 , which was reverted in order to give auto-discovers time to update the collections for disabled bindings to add `flow://connector-schema`. At this point, all the the collections that are going to be updated by auto-discovers will have been. So now we can re-merge #2720 and re-enable the requirement for `flow://connector-schema` to exist prior to discover merge.


jgraettinger (approved):
LGTM

PR #2757, opened 2026-03-11
graphql: add basic connector queries

psFried:
Backend for - https://github.com/estuary/ui/issues/1929

Rolls up a few commits, to add graphql schema for querying connectors and connector tags, and to use the new graphql schema in flowctl when querying connector schemas in order to encrypt endpoint configs. This also updates flowctl to _require_ a connector schema for every task that has an unencrypted endpoint config. In other words, endpoint config encryption is no longer optional in flowctl.


jshearer:
I've reworked the connector GraphQL types a bit to better match what we want the API to look like going forward: connectors as a catalog/menu, with spec resolution as a separate concern. The connector-tags internals are no longer exposed in preparation for refactoring them to be dynamically served in the futur .

### `Connector` type

```graphql
type Connector {
  id: Id!
  imageName: String!
  title: String
  shortDescription: String
  longDescription: String
  logoUrl: String
  externalUrl: String!
  recommended: Boolean!
  createdAt: DateTime!

  # These are new:
  protocol: ConnectorProto          # capture or materialization, directly on the connector
  defaultImageTag: String           # the blessed tag, e.g. ":v1"
  defaultSpec: ConnectorSpec        # full spec for the blessed tag
  spec(imageTag: String!): ConnectorSpec   # look up a specific tag's spec
}
```

The old `connectorTag(imageTag, orDefault)` resolver is replaced by two fields: `defaultSpec` for the common case, and `spec(imageTag)` for looking up a specific version. `tags` and `ConnectorTagRef` are removed.

### `ConnectorSpec` type (was `ConnectorTag`)

Renamed from `ConnectorTag` and trimmed to only the fields relevant to configuring a connector:

```graphql
type ConnectorSpec {
  imageTag: String!
  protocol: ConnectorProto
  documentationUrl: String
  endpointSpecSchema: JSON
  resourceSpecSchema: JSON
  disableBackfill: Boolean!
  defaultCaptureInterval: String
}
```

Internal fields from the `connector_tags` table (`id`, `connectorId`, `createdAt`, `updatedAt`) are removed.

### Top-level query

`connectorTag(fullImageName, id)` is replaced by `connectorSpec(fullImageName: String!)`. If the requested tag is not available, it falls back to the default tag. Check the returned `imageTag` to see which tag was actually resolved.

### Other fixes

- Fixed `has_previous_page` / `has_next_page` being swapped in paginated connector results.
- Protocol filter on the `connectors` query no longer filters the tag list within each connector. Previously, querying `connectors(filter: {protocol: {eq: "capture"}})` would only include capture tags in each connector's internal state. Now the filter controls which connectors appear, but each connector retains all of its tags internally.

### Impact on UI work

I took a look at estuary/ui#1949 and I'm hoping that these changes should be straightforward to adapt. The fields the UI queries (`endpointSpecSchema`, `resourceSpecSchema`, `documentationUrl`, `disableBackfill`, `defaultCaptureInterval`) are all still present on `ConnectorSpec`. `protocol` is now available directly on `Connector` without needing to resolve a spec. `ConnectorSpec.connectorId` was redundant with the parent `connector.id` which the queries already select. The main change is updating `connectorTag(orDefault: true)` to use `defaultSpec` (for the create flow) or `spec(imageTag)` + `defaultSpec` (for the edit flow where a specific tag is known).

jshearer (approved):
LGTM, but I'd like another review maybe from @travjenkins who has been running this locally for UI integration work to avoid approving and merging some of my own changes without review from someone else

travjenkins (approved):
lgtm

crates/control-plane-api/src/server/public/graphql/connectors.rs
```diff
@@ -0,0 +1,696 @@
...
+    tags: Vec<ConnectorTagRef>,
+}
+
+impl Connector {
+    fn default_image_tag_ref(&self) -> Option<&ConnectorTagRef> {
+        self.tags
+            .iter()
+            .filter(|t| t.spec_succeeded_sync())
+            .max_by_key(|t| &t.image_tag)
+    }
```
GregorShear, line 85: This is fragile - `max_by_key(":v9", ":v10")` returns `:v9`.

or worse yet `max_by_key(":v9", ":wip")` returns `:wip`.

Could we instead have a `is_default` boolean column, with a constraint like `CREATE UNIQUE INDEX ON connector_tags (connector_id) WHERE is_default` ?

GregorShear: Chatted w/ Joseph - plan is to remove the `connector_tags` table entirely, and so long as we do so before we reach `v10` on any of the connectors (the highest one is `v5` right now) then this shouldn't be an issue.

GregorShear (approved):

crates/control-plane-api/src/server/public/graphql/connectors.rs
travjenkins: So we have the history here. Today the UI does this logic using `sort` powered by `localeCompare` and would do the same thing so this is kind of good we are not changing this.

image

jshearer: Yep, that was the intent with not changing it. Eventually the idea is to get rid of this implicit sorting and have the connector tag upgrade path be explicit by introducing a mapping of some sort onto `connectors` that says e.g "if you're on `:v3`, then upgrade to `:v4`" sort of thing

PR #2763, opened 2026-03-12
billing: expand invoice generator to detect back-dated manual bills

jshearer:
The manual invoice date filter added in #2387 only matched bills whose `date_start` fell within the `--month` parameter. Multi-month manual bills (e.g. annual contracts) were only visible to the single monthly run matching their start date, and invisible to all subsequent runs. This means that back-dated manual bills were missed unless the invoices generator was re-run for the first month of the back-dated bill.

This changes the invoice generator to process all manual invoices that cover the target billing month. This means that quarterly or yearly invoices will be processed multiple times, which is fine as processing an existing invoice is essentially a no-op: open invoices will be left untouched, and draft invoices will have their line items re-written.

Tested it against the motivating case: a manual bill was created in November 2025 _after_ invoices were run, and never got picked up by subsequent runs. After applying this change, a run of the invoice generator for Feb 2026 picked it up and it got created as expected.


crates/billing-integrations/src/send.rs
```diff
@@ -83,15 +83,19 @@ pub async fn do_send_invoices(cmd: &SendInvoices) -> anyhow::Result<()> {
...
-                                && period_start_date <= month_end_date;
+                    match (inv.period_start(), inv.period_end()) {
+                        (Some(start_str), Some(end_str)) => {
+                            let start = NaiveDate::parse_from_str(&start_str, "%Y-%m-%d");
+                            let end = NaiveDate::parse_from_str(&end_str, "%Y-%m-%d");
+                            match (start, end) {
+                                (Ok(start), Ok(end)) => {
+                                    start <= month_end_date && end >= month_start_date
+                                }
+                                _ => false,
```
jshearer: we should probably fail noisily if these ever fail to parse, rather than silently like we do here

psFried (approved):
LGTM

PR #2775, opened 2026-03-16
Updated storage mapping admin docs to reflect new UI

GregorShear:


site/docs/getting-started/installation.mdx
```diff
@@ -74,176 +63,67 @@ Storage connected with public data planes should not limit access by IP.
 Storage for private data planes can use [PrivateLink](/private-byoc/privatelink) to restrict access.
 :::
```
GregorShear: I assume this is still relevant, but just making sure

site/docs/getting-started/installation.md
```diff
@@ -74,176 +63,67 @@ Storage connected with public data planes should not limit access by IP.
...
-
-3. Gather the following information. You'll need this when you contact us to complete setup.
+You can set a [bucket lifecycle policy](../concepts/storage-mappings.md#bucket-lifecycle-policies) to limit
+how long collections keep data.
+Lifecycle policies should be specific to the `collection-data/` sub-directory that Estuary creates,
+**not** cover the entire bucket.
 
-    - Your **Azure AD tenant ID**. You can find this in the **Azure Active Directory** page.
-      ![Azure AD Tenant ID](https://storage.googleapis.com/estuary-marketing-strapi-uploads/uploads//Azure_AD_Tenant_ID_1b60184837/Azure_AD_Tenant_ID_1b60184837.png)
+## Provider Requirements
```
GregorShear, line 79: I'm unsure if we should keep this section - here's my claude-assisted arguement for and against it

> Case for removing:
> 
> - The dashboard now provides the exact same information with real values filled in. This section is a less useful version of what the user already gets in-app.
> - Having it here risks confusion — a user might land on this page from a search, follow the docs instead of the dashboard, and have a worse experience without the pre-filled values.
>
> Case for keeping:
> 
> - Some users want to understand what they're granting before they start. An IT/security team reviewing access requirements won't be in the dashboard — they'll want a doc they can read and approve.
> - Search discoverability. Someone googling "estuary S3 permissions" or "estuary azure role" should find this page and get an answer.
> - If the dashboard is down or the user is evaluating Estuary before signing up, the docs are the only reference.

aeluce: I'd lean towards keeping this section. While the dashboard is the "correct" route, as you say, not all the information there will be publicly accessible.

site/docs/getting-started/installation.mdx
aeluce: This info came from [this conversation](https://estuaryworkspace.slack.com/archives/C03Q9A3GFEE/p1764763533950489?thread_ts=1764601887.222069&cid=C03Q9A3GFEE) though there was another discussion recently about using IPv6 addresses instead. I updated the note and moved it to a more relevant location.

site/docs/getting-started/installation.md
GregorShear: cool, works for me

PR #2776, opened 2026-03-17
materialize-clickhouse: new connector

jacobmarble:
**Description:**

Document `materialize-clickhouse`.

See https://github.com/estuary/connectors/pull/4022

### DIFF
--- a/site/docs/reference/Connectors/materialization-connectors/ClickHouse.md
+++ b/site/docs/reference/Connectors/materialization-connectors/ClickHouse.md
@@ -9,7 +9,16 @@ This connector writes batches directly to ClickHouse using the
 [Native protocol](https://clickhouse.com/docs/interfaces/tcp) and
 [Native format](https://clickhouse.com/docs/interfaces/formats/Native).
 
-Estuary also provides a [Dekaf-based integration](./Dekaf/clickhouse.md) for users who prefer to ingest via ClickPipes.
+## When to use this connector
+
+We recommend this direct materialization for most ClickHouse use cases.
+It writes to ClickHouse natively, supports hard and soft deletes, and works with both self-hosted and ClickHouse Cloud deployments.
+
+Estuary also offers a [Dekaf-based integration](./Dekaf/clickhouse.md) that ingests via ClickPipes.
+Choose Dekaf when:
+
+- You are required to use ClickPipes for data ingestion (e.g., organizational policy)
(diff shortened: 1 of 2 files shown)


mdibaiee (approved):

site/docs/reference/Connectors/materialization-connectors/ClickHouse.md
```diff
@@ -0,0 +1,123 @@
...
+# ClickHouse
+
+This connector materializes Estuary collections into tables in a ClickHouse database.
+
+[ClickHouse](https://clickhouse.com/) is a column-oriented OLAP database designed for real-time analytics.
+This connector writes batches directly to ClickHouse using the
+[Native protocol](https://clickhouse.com/docs/interfaces/tcp) and
+[Native format](https://clickhouse.com/docs/interfaces/formats/Native).
+
+Estuary also provides a [Dekaf-based integration](./Dekaf/clickhouse.md) for users who prefer to ingest via ClickPipes.
```
aeluce: We could be a little stronger in our preference for the direct materialization.

For situations where we have multiple similar connectors, we'll sometimes call out when it makes sense to use the alternative (ex. [batch vs. CDC](https://docs.estuary.dev/reference/Connectors/capture-connectors/OracleDB/oracle-batch/#when-to-use-this-connector) or the various [SQL Server](https://docs.estuary.dev/reference/Connectors/capture-connectors/SQLServer/#when-to-use-this-connector) capture options). So it could be something like: "We recommend the direct materialization with ClickHouse. Use Dekaf when…"

site/docs/reference/Connectors/materialization-connectors/ClickHouse.md
```diff
@@ -0,0 +1,123 @@
...
+SELECT * FROM my_table FINAL;
+```
+
+## Hard deletes
+
+When `hardDelete: true` is set in the endpoint configuration, the connector adds an `_is_deleted` (UInt8) column to each table.
+When a source document is deleted, the connector inserts a **tombstone row** with `_is_deleted = 1` and the same key columns as the original row.
+The `ReplacingMergeTree` engine uses `_is_deleted` to exclude these rows from `FINAL` queries, and automatic cleanup merges eventually remove the tombstoned records from disk.
+
+## Soft deletes
```
aeluce: My preference would be to nest the soft and hard delete sections under a larger "Handling deletions" section or similar to keep them grouped together, especially if the documentation expands later. Not a blocker, though.

PR #2778, opened 2026-03-17
Shuffle v2: disk-backed log pipeline with zero-copy combiner pass-through

jgraettinger:
## Summary

Complete the working end-to-end implementation of the v2 shuffle pipeline, replacing the placeholder Queue with a disk-backed segmented log for document storage and retrieval. Also adds an efficient `doc::Combiner` path for pre-archived nodes that can pass through without deserialization when reduction/redaction isn't required.

### Segmented log subsystem (`shuffle::log`)

- **Block encoding format**: columnar rkyv-serialized blocks storing journals, producers, per-document metadata (binding, flags, clock), and packed document content. Enables zero-copy document access within loaded / decompressed blocks
- **Writer**: appends encoded blocks to segment files with rolling at byte thresholds or the 65k block-per-segment limit. LZ4 compression support
- **Reader**: frontier-driven `FrontierScan` that yields only entries committed relative to the current `Frontier`, with `BlockScan` using binary search on clock visibility indexes; lazy segment open/close with monotonic LSN watermarks
- **LogActor**: event loop tying together block encoder, writer, and reader into a persistent document merging pipeline with disk backlog back-pressure
- **LogState**: tracks block accumulation lifecycle and multi-stage flush

### Zero-copy combiner pass-through (`doc::HeapRoot::Embedded`)

- New `HeapRoot` enum extending `HeapNode` with an `Embedded` variant carrying pre-serialized `ArchivedNode` bytes (16 bytes, matching `HeapNode` size)
- Documents from the shuffle reader flow through the Combiner without deserialization; only promoted to `HeapNode` when reduction is actually required
- On spill, embedded entries are written as raw byte copies instead of re-serialized through rkyv
- `MemTable::add_embedded()` and `Meta::from_packed_prefix()` skip key re-extraction for pre-validated shuffle documents

### Protocol and infrastructure changes

- Queue → Log rename throughout protobuf definitions and Rust code
- `flushed_lsn` tracking in `Frontier` for segmented log read-through horizons
- `bytes_read_delta` / `bytes_behind_delta` observability accumulators
- Causal hint extraction and projection for cross-journal transaction visibility
- Tightened UUID "deep rollback" detection to handle journal re-read scenarios

### New crates

- **`publisher`**: pipelined journal appends with partition mapping, transaction ACK intents with delta-encoded causal hints
- **`e2e-support`**: hermetic data-plane integration testing with managed etcd and Gazette broker processes over UDS

### Testing

- Scenario fixture tests covering single/multi-producer transactions, multi-member routing, checkpoint resume/recovery, rollback handling, clock window filtering, and partition-filtered causal hints
- Quickcheck fuzz tests with a transactional correctness oracle verifying document ordering, completeness, and causal consistency
- Cross-validation of `MemDrainer` vs `SpillDrainer` output in doc combiner fuzz tests
- Block encoding round-trip and size calculation fuzz tests

### Supporting changes

- `gazette`: switch to `zlib-rs` for ~10-20% faster gzip decompression; error handling refinements; `SubscriberFold` filtering for suspended journals; remove legacy UDS support
- `labels::partition`: decompose field and key-range label operations into independent composable primitives
- `tokens`: fix `Source::refresh()` to select on drop signal, preventing lingering gRPC connections during shutdown
- `doc::Extractor`: moved packed hash calculation for broader reuse
- `doc::json::Fields::get`: linear scan optimization (~2x faster on small field sets in production profiling)
- `validation`: `NoOpConnectors` return expected bindings for local test builds
- `build::for_local_test`: consolidate repeated `managed_build()` patterns

## Test plan

- [x] Scenario fixture tests pass (`cargo test -p shuffle scenario_fixtures`)
- [x] Fuzz tests pass (`cargo test -p shuffle scenario_fuzz`)
- [x] Block encoding fuzz tests pass (`cargo test -p shuffle block`)
- [x] Doc combiner fuzz tests pass (`cargo test -p doc spill_merge_fuzz`)
- [x] `flowctl raw shuffle` exercises the full Session→Log pipeline end-to-end
- [x] e2e-support hello_world test passes (`cargo test -p e2e-support`)

### DIFF
--- a/crates/e2e-support/src/lib.rs
+++ b/crates/e2e-support/src/lib.rs
@@ -38,7 +38,7 @@ impl DataPlane {
         let subscriber = tracing_subscriber::FmtSubscriber::builder()
             .with_env_filter(
                 tracing_subscriber::EnvFilter::builder()
-                    .with_default_directive(tracing::level_filters::LevelFilter::INFO.into())
+                    .with_default_directive(tracing::level_filters::LevelFilter::DEBUG.into())
                     .from_env()
                     .context("parsing RUST_LOG filter")?,
             )


crates/publisher/src/intents.rs
```diff
@@ -0,0 +1,511 @@
...
+                            // Decoder interprets as `this_producer` & `this_commit`
+                            // without consuming space in the journal.
+                            serde_json::json!({})
+                        } else {
+                            let p = base64::engine::general_purpose::STANDARD
+                                .encode(hinted_producer.as_bytes());
+
+                            serde_json::json!({
+                                "p": p,
+                                "c": format!("{:016x}", hinted_commit.as_u64()),
```
williamhbaker, line 67: Are these u64's being hex-encoded to avoid any possible problems with them being larger than an i64?

crates/e2e-support/src/gazette.rs
```diff
@@ -0,0 +1,133 @@
...
+impl GazetteCluster {
+    /// Start a cluster of `broker_count` gazette brokers connected to the given etcd.
+    pub async fn start(etcd: &EtcdInstance, broker_count: usize) -> anyhow::Result<Self> {
+        // Fixed HMAC key for test auth. Pre-computed base64 of "test-data-plane-hmac-secret!".
+        let base64_key = "dGVzdC1kYXRhLXBsYW5lLWhtYWMtc2VjcmV0IQ==".to_string();
+
+        let (encode_key, _decode_keys) = tokens::jwt::parse_base64_hmac_keys([&base64_key])
+            .map_err(|status| anyhow::anyhow!("failed to parse HMAC key: {status}"))?;
+
+        let gazette_bin = format!("{}/go/bin/gazette", std::env::var("HOME")?);
```
williamhbaker, line 28: Does this need to hard-code the path at `~/go/bin/gazette`? Not that I'd expect it to really be a problem, but the etcd path just uses `"etcd"` as the PATH lookup.

crates/shuffle/src/frontier.rs
```diff
@@ -181,27 +201,46 @@ impl JournalFrontier {
...
 /// Frontier tracks journal progress including causal hints.
+///
+/// A Frontier may represent either a cumulative checkpoint (full state of all
+/// journals and producers) or a checkpoint delta (only journals and producers
+/// that progressed since the last checkpoint). The `reduce` method merges a
+/// delta into a cumulative base: new journals from the delta are added, base
+/// journals absent from the delta are preserved, and matching entries are
+/// reduced by maximizing clocks.
+///
+/// See the CheckpointPipeline (README §11) for how deltas are produced.
```
williamhbaker: nit: Referencing a specific section of the README seems like it has a high chance of going stale, so might be better to make this more of a general reference to the README docs.

crates/shuffle/src/session/handler.rs
```diff
@@ -18,11 +18,13 @@ where
...
 
-    let shuffle::session_request::Open {
-        session_id,
-        task,
-        members,
-    } = open.open.context("first message must be Open")?;
+    let shuffle::session_request::Open { task, members } =
+        open.open.context("first message must be Open")?;
+
+    let session_id: u32 = std::time::SystemTime::now()
```
williamhbaker, line 24: Is it going to be an issue for this to wrap around every ~4.3s? Since this is used as part of the key in `log_joins`, it seems like collisions could be problematic, although rare.

crates/flowctl/src/raw/shuffle.rs
```diff
@@ -97,171 +111,228 @@ impl Shuffle {
...
+        {
+            let memtable = accumulator.memtable()?;
+
+            for entry in scan.block_iter() {
+                memtable.add_embedded(
+                    entry.meta.binding.to_native(),
+                    &entry.doc.packed_key_prefix,
+                    entry.doc.doc.to_heap(memtable.alloc()),
+                    false,
+                    true,
```
williamhbaker: This is bypassing schema validation in `flowctl` right? Can we get the value for `known_valid` from the block entry flags?

crates/shuffle/src/log/handler.rs
```diff
@@ -19,47 +19,55 @@ where
...
-    let join_key = (session_id, queue_member_index);
+    let join_key = (session_id, directory.to_string(), log_member_index);
 
     // Scope `guard` to prove it's not held across await points.
     let connections = {
-        let mut guard = service.queue_joins.lock().unwrap();
+        let mut guard = service.log_joins.lock().unwrap();
 
-        let join = guard.entry(join_key).or_insert_with(|| QueueJoin {
+        let join = guard.entry(join_key.clone()).or_insert_with(|| LogJoin {
```
williamhbaker, line 50: Can slices every be removed from `log_joins`? For example, for a session with 2 slices, if Slice 0 joins but Slice 1 fails and the Session is torn down, will the `log_join` entry for Slice 0 stayed here forever?

crates/shuffle/src/log/writer/sealed.rs
```diff
@@ -0,0 +1,407 @@
...
+
+    let dir = path
+        .parent()
+        .context("segment path must have a parent directory")?;
+    let filename = path
+        .file_name()
+        .context("segment path must have a filename")?;
+    let tmp_path = dir.join(format!(".compress-{}", filename.to_string_lossy()));
+
+    let mut tmp_file = std::fs::OpenOptions::new()
```
williamhbaker: Noting that this temp file would be left around if there are any errors before the `rename_exchange` on line 207. Worth adding some kind of catch-all that ensures the temp file is cleaned up on an early return? I'm not sure what that would look like or how viable it would be, but I know we've had to deal with reactor disks filling up due to repeated failures in the past, so it would be nice to mitigate these if possible.

williamhbaker (approved):
LGTM!

A few comments for consideration. I think the potential for collisions with `session_id` is probably the closest one I had to being a blocker, although it would be quite rare.

crates/e2e-support/src/gazette.rs
jgraettinger: Given the current `mise` project structure, yes. `etcd` is installed by mise as a "tool" and is contained in an opaque mise-managed tool directory on the PATH. `gazette` is built by us (e.x. via a mise task) and is _not_ on the PATH.

crates/flowctl/src/raw/shuffle.rs
jgraettinger: Yes, this was an oversight, will fix.

crates/publisher/src/intents.rs
jgraettinger: Nope, we handle u64 vs i64 just fine. I hex encoded because:

* They're very large integers, so hex is moderately more space efficient
* They're monotonic, and IMO it's easier to visually compare large monotonic integers as fixed-width hex vs arbitrary-length base-10

crates/shuffle/src/log/writer/sealed.rs
jgraettinger: Yep, switched to use `tempfile::NamedTempFile` which deletes on drop.

crates/shuffle/src/log/handler.rs
jgraettinger: It'll stick around until process termination. There could be some kind of timed sweep, but 🤷.

crates/shuffle/src/session/handler.rs
jgraettinger: I'm removing `session_id` as a key component of log joins -- that's left over cruft.

The directory and member index already need to be unique (or sessions will clobber each other on file names and fail), so the intention is that the self-reported member directory of each member is already unique (corresponds to a shard assignment lifetime).

crates/shuffle/src/frontier.rs
jgraettinger: Fixed

crates/shuffle/src/log/handler.rs
williamhbaker: Probably not a big deal, then. I suppose it would be extremely unlikely for this ever to cause a real problem.

PR #2787, opened 2026-03-18
docs: fix inaccurate delete event field descriptions

jwhartley:
## Summary

- Fixes docs that incorrectly stated delete events contain *only* the primary key and meta fields
- The runtime doesn't strip fields from delete events — what's present depends on the source connector
- PostgreSQL CDC with `REPLICA IDENTITY FULL`, for example, includes all column values in deletes

## Changed files

- `site/docs/guides/deletions.md` — main deletions reference page
- `site/docs/guides/advanced-usage/metadata-fields.md` — metadata fields reference

### DIFF
--- a/site/docs/guides/deletions.md
+++ b/site/docs/guides/deletions.md
@@ -69,7 +69,7 @@ There are a couple best practices to consider when writing derivations in conjun
 
 **Filtering**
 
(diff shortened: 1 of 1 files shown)


site/docs/guides/deletions.md
aeluce:
```suggestion
Since delete events can lack standard document fields, it can be easy to accidentally filter them out from downstream systems.
```

PR #2796, opened 2026-03-19
Phase 4a: SSO provider on tenants + grant migration

GregorShear:
- Add `sso_provider_id` column to tenants table. 
- Add grant migration trigger on `auth.identities` insert: when a new SSO user's email matches an existing social user, transfer grants from old account to new. Grants on tenants with a _different_ `sso_provider_id` (i.e. `sso_provider_id IS NOT NULL AND sso_provider_id != new_provider`) are not transferred. 
- Old account is banned, sessions/refresh tokens revoked.
- Adding claude skill for setting up mocksaml.com as sso provider for local testing (if you try this let me know if it works)

No user-visible behavior change for existing users without them intentionally going through the SSO login flow. The trigger only fires when a brand new SSO identity is created.

> Need to document and make sure customers realize that flowctl / refresh tokens will stop working if they're connected to a social user who migrates to SSO. Remind me to raise this in appropriate channels when we merge this work

**Verify:**

- [ ] SSO user signs in with email matching social user → grants transferred, old account banned
- [ ] Grants on tenants with a different `sso_provider_id` are NOT transferred
- [ ] Grants on tenants with no `sso_provider_id` ARE transferred
- [ ] No existing user access is affected


supabase/migrations/20260319120000_sso_grant_migration.sql
```diff
@@ -0,0 +1,121 @@
...
+comment on column public.tenants.sso_provider_id is
+  'SSO provider that governs this tenant, if any. NULL means no SSO requirement.';
+
+-- Trigger function: migrate grants from old social user to new SSO user.
+create or replace function internal.on_sso_identity_insert()
+returns trigger
+language plpgsql
+security definer
+-- https://www.postgresql.org/docs/current/sql-createfunction.html#SQL-CREATEFUNCTION-SECURITY
+set search_path to ''
```
GregorShear, line 26: Claude recommends this `search_path to ''` as an extra security measure, but i don't see it anywhere else in the codebase

supabase/migrations/20260319120000_sso_grant_migration.sql
```diff
@@ -0,0 +1,121 @@
...
+              public.user_grants.capability,
+              excluded.capability
+            ),
+            updated_at = now(),
+            detail = excluded.detail;
+    end if;
+  end loop;
+
+  -- Clean up old account: delete grants, tokens, sessions.
+  -- delete from public.user_grants where user_id = old_user_id;
```
GregorShear, line 98: Okay to leave grants on old user? at least makes this reversable

supabase/migrations/20260319120000_sso_grant_migration.sql
```diff
@@ -0,0 +1,121 @@
...
+  delete from public.refresh_tokens where user_id = old_user_id;
+  delete from auth.sessions where user_id = old_user_id;
+
+  -- Ban old account in GoTrue so they can't sign in via social auth.
+  update auth.users
+    set banned_until = '2999-01-01'::timestamptz
+    where id = old_user_id;
+
+  -- Record suspension for our own tracking.
+  -- insert into internal.account_suspensions (user_id, reason)
```
GregorShear: Social users are banned after migrating to SSO - is this something we should record in `internal.account_suspensions`?

@skord

skord: `internal.account_suspensions` exists for a very particular reason: GDPR/CCPA right-to-be-forgotten requests as facilitated by the (dangerous as hell to use) deletion script. There is a bit of documentation in that repo/branch.

The expire command (which permanently deletes users) requires the suspension reason to be `voluntary_deletion` before it will proceed. As long as you don't use that reason when suspending a user, the expire command will refuse to delete them. (There is a `--force` flag that bypasses this, but it's meant for edge cases and requires intentional use.)

This table reflects current state, not audit history. Suspension records are cleaned up automatically — either by the script on unsuspend, or by an `ON DELETE CASCADE` when the `auth.users` row is removed during expire. The table was kept separate from `auth.users` to avoid modifying a schema managed by Supabase.

So feel free to use the table if you wish, I can't think of a reason to say no, but there's no hard compelling reason you _have_ to use it.

GregorShear: Thanks for the input. account_suspensions feels unrelated to banning migrated social users - we're just banning the social account to avoid the confusion of active users logging into the wrong account after they've set up SSO.

supabase/migrations/20260319120000_sso_grant_migration.sql
psFried: I did some searching, and my understanding is this is intended to mitigate the risk of someone either messing with `search_path` or creating tables in other schemas. It seems to me like if someone has enough access to our DB to do either of those, then we're kinda already hosed. But I can't think of anything it would hurt to leave this here.

supabase/migrations/20260319120000_sso_grant_migration.sql
```diff
@@ -0,0 +1,129 @@
...
+  -- the most specific tenant governing each grant's prefix.
+  for grant_row in
+    select ug.object_role, ug.capability, ug.detail,
+           t.sso_provider_id as tenant_sso_provider_id
+    from public.user_grants ug
+    left join lateral (
+      select t2.sso_provider_id
+      from public.tenants t2
+      where ug.object_role ^@ t2.tenant
+      order by length(t2.tenant::text) desc
```
psFried: Why the `desc` sort here? Supposing a user has grants to both `acmeCo/` and `acmeCo/nested/`, wouldn't we want to select the `acmeCo/` one?

supabase/migrations/20260319120000_sso_grant_migration.sql
psFried: seems safe enough for now. We can always cleanup grants from disabled users later if we need.

psFried (approved):
This looks good to me % the comment on the desc sort

supabase/migrations/20260319120000_sso_grant_migration.sql
GregorShear: good catch, thanks for drawing attention here - this was vestigial/over complicated, a regular left join will do what we need

PR #2797, opened 2026-03-20
Phase 4b: Invite SSO enforcement + SSO routing

GregorShear:
- Reject invite redemption when the invite's tenant has `sso_provider_id IS NOT NULL` and the redeeming user lacks a matching SSO identity (`auth.identities.provider_id = tenants.sso_provider_id`). 
- Keyed on `sso_provider_id IS NOT NULL`. 
- This gates **new** grant creation via invites as soon as a tenant has SSO configured.
- Fix invite link query ordering (order by created_at desc)

Additionally, route invite links directly into the SSO flow:

- `createInviteLink` looks up the tenant's `sso_provider_id` and includes it in the returned link/metadata
- Add `ssoProvider` ID to the invite link URL (e.g. `?grantToken=...&ssoProvider=...`)
- Dashboard reads `ssoProvider` from query params and calls GoTrue `POST /auth/v1/sso` directly

These ship together because rejecting non-SSO invite redemption without routing users to the correct SSO flow would be a dead end.

**Verify:**

- [ ] SSO user (matching provider) redeems invite for SSO tenant → success
- [ ] SSO user from Tenant A's provider tries to redeem invite for Tenant B (different `sso_provider_id`) → rejected with clear error
- [ ] Non-SSO user tries to redeem invite for SSO tenant → rejected with clear error
- [ ] Non-SSO user redeems invite for tenant without `sso_provider_id` → unaffected
- [ ] SSO user redeems invite for tenant without `sso_provider_id` → unaffected (no restriction)
- [ ] Invite for sub-prefix under SSO tenant → same enforcement applies
- [ ] Invite link for SSO tenant includes `ssoProvider` param → user routed directly into SSO flow

### DIFF
--- a/crates/control-plane-api/src/server/public/graphql/invite_links.rs
+++ b/crates/control-plane-api/src/server/public/graphql/invite_links.rs
@@ -1,4 +1,4 @@
-use super::filters;
+use super::{TimestampCursor, filters};
 use async_graphql::{Context, types::connection};
 
 /// An invite link that grants access to a catalog prefix.
@@ -16,6 +16,10 @@ pub struct InviteLink {
     pub detail: Option<String>,
     /// When this invite link was created.
     pub created_at: chrono::DateTime<chrono::Utc>,
+    /// The SSO provider ID for the invite's tenant, if any.
+    /// When present, the frontend should route the user directly into the SSO
+    /// flow using this provider ID (e.g. via `supabase.auth.signInWithSSO`).
+    pub sso_provider_id: Option<uuid::Uuid>,
 }
 
 /// Result of redeeming an invite link.
@@ -28,7 +32,7 @@ pub struct RedeemInviteLinkResult {
 }
 
 pub type PaginatedInviteLinks = connection::Connection<
-    String,
+    TimestampCursor,
     InviteLink,
     connection::EmptyFields,
     connection::EmptyFields,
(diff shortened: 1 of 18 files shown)


crates/control-plane-api/src/server/public/graphql/invite_links.rs
```diff
@@ -255,6 +288,49 @@ impl InviteLinksMutation {
...
+        // through SAML. If this assumption changes, we should check the JWT's
+        // amr claim to verify the current session used SSO.
+        let tenant = invite
+            .catalog_prefix
+            .split_at(
+                invite
+                    .catalog_prefix
+                    .find('/')
+                    .expect("prefix must contain '/'")
+                    + 1,
```
GregorShear: Is there a util function somewhere to extract tenant from prefix? not that it's complicated, but...

crates/control-plane-api/src/server/public/graphql/invite_links.rs
```diff
@@ -95,7 +99,7 @@ impl InviteLinksQuery {
             first,
             None,
             |after: Option<String>, _, first, _| async move {
-                let after_token: Option<uuid::Uuid> = match after {
+                let after_created_at: Option<chrono::DateTime<chrono::Utc>> = match after {
```
psFried: I didn't notice this on the first pr, but I think the better way to do this is to let `connection::query` handle parsing the pagination parameters by setting the first `Connection<T, ...>` type parameter. There's already a `TimestampCursor` type that we use [here](https://github.com/estuary/flow/blob/7e6328bfba26be0a3a336a4eb30faf0ce6b860b9/crates/control-plane-api/src/server/public/graphql/publication_history.rs#L117), and it would probably be appropriate to hoist that up into a parent module to share it.

crates/control-plane-api/src/server/public/graphql/invite_links.rs
psFried: There's [this](https://github.com/estuary/flow/blob/7e6328bfba26be0a3a336a4eb30faf0ce6b860b9/crates/control-plane-api/src/publications/quotas.rs#L76), though nothing I can recall that's in a shared module. Seems like it might be worth it, though

crates/models/src/references.rs
```diff
@@ -233,6 +233,15 @@ string_reference_types! {
...
 }
 
+/// Extract the tenant prefix (e.g. "acmeCo/") from a catalog name or prefix.
+/// The input must contain at least one '/'.
+pub fn tenant_from(prefix: &str) -> &str {
+    let idx = prefix
+        .find('/')
+        .expect("prefix must contain at least one /");
+    &prefix[..idx + 1]
+}
```
GregorShear, line 243: okay for this to live here?

crates/control-plane-api/src/server/public/graphql/invite_links.rs
GregorShear: made this change

PR #2798, opened 2026-03-20
Phase 4c: Intercept non-SSO logins for enforced domains

GregorShear:
- Add `enforce_sso` boolean column to tenants. 
- GoTrue `customize_access_token` hook (`check_sso_requirement`) blocks social login for users with email covered by tenant's SSO domains
- The hook checks `auth.users.is_sso_user` — SSO users always pass through (important for token refresh). 
- Blocked logins receive a 403 error with `message: 'sso_required:<domain>'` — the frontend parses the domain and calls `supabase.auth.signInWithSSO({ domain })` to redirect the user to their IdP.

Users whose email domain does NOT match an enforcing tenant's SSO domain (e.g., contractors with `@gmail.com`) are not blocked by the hook. Their grants on enforcing tenants are handled by grant filtering in 4d instead.

Activated by `enforce_sso = true`. Tenants configure SSO first, then communicate the transition to their users. `enforce_sso` is flipped on a hard cutoff date.

**Verify:**

- [x] Social user with matching email domain on enforcing tenant → blocked with `sso_required:<domain>`
- [x] Social user with matching domain, `enforce_sso = false` → not blocked
- [x] Social user with non-matching domain (e.g. `@gmail.com`) → not blocked (grants handled by 4d)
- [x] SSO user (`is_sso_user = true`) with matching domain → not blocked (token refresh works)
- [x] Blocked user redirected via `signInWithSSO({ domain })` → authenticates via IdP → grant migration trigger fires
- [x] Malformed hook input → exception caught, login IS blocked, 500 appears in the logs with message `check_sso_requirement failed`


mise/tasks/local/start-with-saml
GregorShear, line 1: 100% claude generated script for setting up mocksaml in a dev environment. it works on my machine... :)

.claude/skills/setup-mocksaml/SKILL.md
GregorShear, line 1: will probably remove this skill in favor of the `start-with-saml.sh` script below

supabase/config.toml
```diff
@@ -45,6 +45,10 @@ jwt_expiry = 604800
 # Allow/disallow new user signups to your project.
 enable_signup = true
 
+[auth.hook.custom_access_token]
```
GregorShear: for dev environment - will have to configure this via the supabase dashboard in production

supabase/migrations/20260320120000_check_sso_requirement_hook.sql
```diff
@@ -0,0 +1,84 @@
...
+  ) then
+    return jsonb_build_object(
+      'error', jsonb_build_object(
+        'http_code', 403,
+        'message', 'sso_required:' || user_domain
+      )
+    );
+  end if;
+
+  return event;
```
GregorShear, line 67: I've removed the exception handler so exceptions will bubble up to the caller - is this the right approach? is there any specific monitoring we need to set up or are all exceptions like readily visible? (where would i look for these anyway?)

.claude/skills/setup-mocksaml/SKILL.md
jshearer: agreed. I think it might be worth documenting any relevant nuances or complexities in a `README` or document somewhere, but the skill and bash script seem a bit redundant

supabase/migrations/20260320120000_check_sso_requirement_hook.sql
```diff
@@ -0,0 +1,84 @@
...
```
jshearer, line 67: > is this the right approach? is there any specific monitoring we need to set up or are all exceptions like readily visible?

You are now the expert in how GoTrue/Supabase auth/SSO works 😁. 

Can you verify what actually happens if this throws an exception? It would be nice to know ahead of time what that looks like in case something breaks later.

supabase/migrations/20260320120000_check_sso_requirement_hook.sql
```diff
@@ -0,0 +1,84 @@
+-- Access token hook that blocks social login for users whose email domain
+-- matches an SSO-enforcing tenant's configured domain.
+--
+-- When enforce_sso is true and the user's email domain appears in
+-- auth.sso_domains for that tenant's provider, the hook refuses to mint
+-- a token — returning an error that tells the frontend to redirect to SSO.
+--
+-- Users whose email domain does NOT match (e.g. contractors, partners)
+-- are not blocked here; their grants wil be handled separately
```
jshearer:
```suggestion
-- are not blocked here; their grants will be handled separately
```

supabase/config.toml
jshearer: One thought: this will break everyone's dev stack until they run `supabase db reset` or manually apply the above migration. That's... probably fine. Can you think of any reason the above migration will fail for people, for example if they aren't running supabase via `start-with-saml.sh` or something?

supabase/migrations/20260320120000_check_sso_requirement_hook.sql
jshearer, line 67: One more observation: if `target_user_id` doesn't match any `auth.users`, then this hook will pass (i.e it'll `return event`). Have you given this scenario any thought? It really ought to be handled by the rest of the auth stack.. nevertheless I'd rather trust but verify.

mise/tasks/local/start-with-saml
```diff
@@ -0,0 +1,150 @@
+#!/usr/bin/env bash
```
jshearer: IMHO this belongs in the `mise` scripts. Can you find the right place to put it, ideally as some sort of opt-in flag when mise is starting up supabase?

supabase/migrations/20260320120000_check_sso_requirement_hook.sql
GregorShear: this hook intercepts the flow post-authentication, so the user_id must already be in the table

supabase/config.toml
GregorShear: the start-with-saml script just enables SAML for the local supabase instance - it's not required to run the migration. 

that said, i'll move this config into the start-with-saml script so it doesn't get in the way of regular non-sso development

mise/tasks/local/start-with-saml
GregorShear: good idea - usage is now `mise run local:stack --saml`

supabase/migrations/20260320120000_check_sso_requirement_hook.sql
GregorShear: i put the exception handler back so we can throw our own 500 error with a message `check_sso_requirement failed` so we can at least filter on that string. 

we can keep an eye on it here, at least
https://supabase.com/dashboard/project/eyrcnmuzzyriypdajwdk/observability/auth

supabase/config.toml
GregorShear: moved this to the `local:stack --saml` script

mise/tasks/local/stack
```diff
@@ -3,6 +3,13 @@ set -euo pipefail
 
 #MISE description="Start a local stack (control plane and 'local-cluster' data plane)"
 #MISE depends=["build:connector-init", "build:rocksdb", "local:control-plane"]
+#USAGE flag "--saml" help="Enable SAML/SSO via MockSAML after stack starts"
+#USAGE flag "--saml-domain <domain>" help="Email domain for MockSAML provider" default="example.com"
+#USAGE flag "--saml-tenant <tenant>" help="Tenant to link to MockSAML provider (defaults to first non-ops tenant)"
+
+SAML="${usage_saml:-false}"
+SAML_DOMAIN="${usage_saml_domain:-example.com}"
+SAML_TENANT="${usage_saml_tenant:-}"
```
jshearer, line 12: ❤️

PR #2800, opened 2026-03-20
materialization triggers

williamhbaker:
**Description:**

Adds a new triggers configuration to materializations that fires webhook requests after each committed transaction. This enables users to notify downstream services (e.g. kick off a dbt job) when new data has been materialized.

See individual commit messages for specifics on the implementation, as well as [this design discussion](https://github.com/estuary/flow/discussions/2713) where higher-level points are described.

Testing:
- [x] Unit tests exercise the flow of persisting trigger variables and firing the triggers (webhooks)
- [x] Ran on a local stack: Used `flowctl` to publish a config testing payload delivery to webhook.site. Using `flowctl`, you can change the non-secret, non-HMAC protected values without decrypting the config.

These things are called "Triggers" as an overarching concept, even though they are webhooks, and the terminology of webhooks is inevitably mixed in. I wanted to use a more general term for the feature name, since it is perhaps conceivable that other kinds of triggers could be imagined in the future.

**Workflow steps:**

Create/edit materialization configs with the new `triggers` field, see the docs with this PR for examples. 

Some initial docs are included in this PR.

**Notes for reviewers:**

The additions to `flow-web` are somewhat speculative. I'm fairly confident the UI will need the triggers config schema for rendering a form eventually. The helpers `strip_trigger_hmac_excluded_fields` & `restore_trigger_hmac_excluded_fields` may provide equivalent behaviors as are built into `flowctl` to manipulate encrypted configs, but how this will actually be implemented in the UI form handling is TBD.

### DIFF
--- a/site/docs/concepts/materialization/materialization-triggers.md
+++ b/site/docs/concepts/materialization/materialization-triggers.md
@@ -1,0 +1,292 @@
+---
+slug: /concepts/materialization-triggers/
+---
+
+# Materialization Triggers
+
+Triggers let you fire webhook requests whenever a materialization commits a
+transaction. You can use them to notify downstream services that new data has
+been materialized — for example, to kick off a dbt run, send a Slack message,
+or call a custom API.
+
+Triggers are configured on the materialization itself and fire once per
+committed transaction. Each trigger sends an HTTP request whose URL, method,
+headers, and JSON body you define. The body is a
+[Handlebars](https://handlebarsjs.com/) template that can reference transaction
+metadata and secret header values.
+
+## How triggers work
+
+Estuary processes data in
+[transactions](/concepts/advanced/shards/#transactions). When a materialization
+transaction commits, the runtime renders each configured trigger's payload
+template with variables from the transaction, then sends the resulting JSON body
+to the trigger's URL.
+
+Triggers are delivered with **at-least-once** semantics. The trigger parameters
+are persisted to durable storage as part of the transaction commit, so if the
+materialization restarts before delivery completes, the triggers will be
+re-delivered on recovery.
+
+All configured triggers fire concurrently after each transaction is acknowledged
+by the connector. If a trigger receives a retryable error (5xx, 408, or 429), it
+is retried with exponential backoff for up to `maxAttempts` total attempts.
+Non-retryable client errors (other 4xx status codes) fail immediately without
+retry.
+
+:::note
+Triggers fire once per materialization transaction, not once per document. A
+single transaction may contain documents from multiple bindings. The
+`collection_names` template variable lists which collections contributed
+documents to the transaction.
+:::
+
+## Specification
+
+Triggers are defined under the `triggers` key of a materialization
+specification:
+
+```yaml
+materializations:
+  acmeCo/example/database-views:
+    endpoint:
+      connector:
+        image: ghcr.io/estuary/materialize-postgres:dev
+        config: path/to/connector-config.yaml
+    bindings:
+      - source: acmeCo/example/collection
+        resource: { table: example_table }
+
+    # Webhook triggers fired after each committed transaction.
+    # Optional, type: object
+    triggers:
+      config:
+        - # URL of the webhook endpoint.
+          # Required, type: string
+          url: "https://example.com/webhook"
+          # HTTP method for the request.
+          # Optional. Default: POST. One of: POST, PUT, PATCH.
+          method: POST
+          # HTTP headers to include in the request. Header values are
+          # encrypted at rest and can be referenced in the payload template
+          # as {{headers.HeaderName}}, making them useful for secrets like
+          # API keys that need to appear in the request body.
+          # Optional, type: object
+          headers:
+            Authorization: "Bearer my-secret-token"
+          # Handlebars template for the JSON request body. See "Template
+          # variables" below for available variables.
+          # Required, type: string
+          payloadTemplate: |
+            {
+              "materialization": "{{materialization_name}}",
+              "collections": [{{#each collection_names}}"{{this}}"{{#unless @last}}, {{/unless}}{{/each}}],
+              "publishedAtMin": "{{flow_published_at_min}}",
+              "publishedAtMax": "{{flow_published_at_max}}",
+              "runId": "{{run_id}}"
+            }
+          # Request timeout for each delivery attempt.
+          # Optional. Default: 30s.
+          timeout: 30s
+          # Maximum number of delivery attempts (including the initial attempt).
+          # Optional. Default: 3.
+          maxAttempts: 3
+```
+
+## Properties
+
+| Property | Title | Description | Type | Default |
+|---|---|---|---|---|
+| **`/triggers/config`** | Trigger Configurations | List of webhook triggers to fire when new data is materialized. | array | |
+| **`/triggers/config/*/url`** | URL | URL of the webhook endpoint. Must be a valid URL. | string | |
+| **`/triggers/config/*/method`** | HTTP Method | HTTP method for the request. One of `POST`, `PUT`, or `PATCH`. | string | `POST` |
+| **`/triggers/config/*/headers`** | Headers | HTTP headers to include in the request. Values are encrypted at rest. | object | |
+| **`/triggers/config/*/payloadTemplate`** | Payload Template | Handlebars template that renders to the JSON request body. | string | |
+| **`/triggers/config/*/timeout`** | Timeout | Request timeout for each delivery attempt. Must be greater than 0. The task is failed if all attempts are exhausted without a successful delivery. | string (duration) | `30s` |
+| **`/triggers/config/*/maxAttempts`** | Max Attempts | Maximum number of delivery attempts (including the initial attempt). | integer | `3` |
+
+## Template variables
+
+The payload template is rendered using [Handlebars](https://handlebarsjs.com/)
+with the following variables:
+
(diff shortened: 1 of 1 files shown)


crates/runtime/src/materialize/protocol.rs
```diff
@@ -247,14 +248,19 @@ pub fn recv_client_load_or_flush(
...
             if let Some((_, clock, _)) = binding.uuid_ptr.query(&doc).and_then(|node| match node {
                 doc::HeapNode::String(uuid) => proto_gazette::uuid::parse_str(uuid.as_str()).ok(),
                 _ => None,
             }) {
-                stats.3 = clock;
+                if stats.right.docs_total == 1 || clock < stats.first_source_clock {
+                    stats.first_source_clock = clock;
+                }
+                if clock > stats.last_source_clock {
+                    stats.last_source_clock = clock;
```
williamhbaker, line 262: A small behavior change here, since we'll now ensure the last_source_lock for a binding is the maximum observed.

crates/runtime/src/materialize/triggers.rs
```diff
@@ -0,0 +1,479 @@
...
+        .min()
+        .map(|c| clock_to_rfc3339(&c))
+        .unwrap_or_default();
+
+    let flow_published_at_max = last_clocks
+        .max()
+        .map(|c| clock_to_rfc3339(&c))
+        .unwrap_or_default();
+
+    let flow_run_id = uuid::Uuid::new_v4().to_string();
```
williamhbaker: I originally had a UUID here mostly as a placeholder, and then never came back around to thinking about it anymore. Any random string should do the trick, since it is durably persisted and recovered when retrying the webhook delivery.

But perhaps there is something more useful we could use here - I am considering a hex-encoded hash of the runtime checkpoint, for example.

crates/models/src/triggers.rs
```diff
@@ -0,0 +1,212 @@
...
+    #[serde(default)]
+    #[schemars(extend("default" = "POST"))]
+    pub method: HttpMethod,
+    /// # HTTP headers to include in the webhook request.
+    #[serde(default, skip_serializing_if = "BTreeMap::is_empty")]
+    #[schemars(schema_with = "secret_string_map_schema")]
+    pub headers: BTreeMap<String, String>,
+    /// # Handlebars template for the JSON payload body.
+    pub payload_template: String,
+    /// # Request timeout in seconds.
```
jgraettinger: nit: perhaps note that the task is failed if timeout_secs + max_attempts is violated.

Also, use `humantime` rather than a `_secs` suffix. See how min_txn_duration / max_txn_duration work as examples.

crates/models/src/triggers.rs
```diff
@@ -0,0 +1,212 @@
...
+/// Template variables for webhook trigger rendering, computed from transaction state.
+/// Persisted to RocksDB during StartCommit for at-least-once delivery guarantees.
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+pub struct TriggerVariables {
+    pub collection_names: Vec<String>,
+    pub connector_image: String,
+    pub materialization_name: String,
+    pub flow_published_at_min: String,
+    pub flow_published_at_max: String,
+    pub flow_run_id: String,
```
jgraettinger: nit: just `run_id` ? `flow_published_at` is a term of art we're pretty locked into, but in general we're moving away from "Estuary Flow" branding.

If this name has been discussed / agreed to already and I'm unaware, feel free to ignore.

crates/validation/src/materialization.rs
```diff
@@ -602,6 +603,21 @@ async fn walk_materialization<C: Connectors>(
         false, // Don't disable wait_for_ack.
         &network_ports,
     );
+    let triggers_json: bytes::Bytes = match &triggers {
+        None => bytes::Bytes::new(),
+        Some(t) => {
+            validate_triggers(scope.push_prop("triggers"), &t.config, errors);
+
+            // TODO(whb): SOPS HMAC integrity of encrypted triggers is currently
```
jgraettinger: We'd want to wire it into the Materialization Validate RPC, and have the `runtime` crate unseal as part of that call's implementation. That would happen within the data-plane that's intended to actually fire the triggers / has SOPs decrypt privilege.

Makes sense to do as a separate PR, though, and maybe not very high priority unless it's an issue in practice

crates/runtime/src/materialize/protocol.rs
```diff
@@ -362,10 +370,19 @@ pub async fn recv_connector_acked_or_loaded_or_flushed(
...
             if let Some(state) = state {
-                let mut wb = rocksdb::WriteBatch::default();
                 queue_connector_state_update(&state, &mut wb).context("invalid Acknowledged")?;
-                db.write_opt(wb, rocksdb::WriteOptions::default()).await?;
+            }
+            if let Some(compiled) = compiled_triggers
+                && let Some(variables) =
+                    db.load_trigger_params::<models::TriggerVariables>().await?
+            {
+                fire_pending_triggers(compiled, &variables, http_client).await?;
```
jgraettinger: 🤔 while this is fine for a first pass, it blocks the transaction loop and these triggers could block for a while. I think we're going to ultimately want to tokio::spawn this and service the JoinHandle in the main protocol loop.

This can wait for a for a follow-up

crates/runtime/src/materialize/serve.rs
```diff
@@ -86,6 +90,63 @@ async fn serve_session<L: LogHandler>(
 ) -> anyhow::Result<Option<Request>> {
     recv_client_open(&mut open, &db).await?;
 
+    // Extract trigger config and connector image from the spec.
+    let (triggers_json, connector_image) = {
```
jgraettinger: I think this should happen in runtime::materialize::connector, alongside where unseal::decrypt_sops() of the endpoint config lives for each of the image and local connector cases.

This would also more-naturally power eventual Validate-time validation, discussed elsewhere

crates/runtime/src/materialize/triggers.rs
jgraettinger: A thought: what about a wall-clock transaction time, such as the start time of the transaction from its own stats? More information bearing, monotonic, but also unique.

I don't think we'll _regret_ a v4 UUID, but I'm also not convinced it rises to the level of deserving "an ID", which elevates it to a heaviness of concept / gravity of thing-ness that doesn't quite seem warranted.

site/docs/concepts/materialization/materialization-triggers.md
```diff
@@ -0,0 +1,292 @@
...
+All configured triggers fire concurrently after each transaction is acknowledged
+by the connector. If a trigger receives a retryable error (5xx, 408, or 429), it
+is retried with exponential backoff for up to `maxAttempts` total attempts.
+Non-retryable client errors (other 4xx status codes) fail immediately without
+retry.
+
+:::note
+Triggers fire once per materialization transaction, not once per document. A
+single transaction may contain documents from multiple bindings. The
+`collection_names` template variable lists which collections contributed
```
jgraettinger, line 40: Right... for later, but reading this is making me realize that we may want resource paths to _also_ be exposed to templates. That relates more closely to what the user may actually care about (which _tables_ of mine were updated?).

crates/models/src/triggers.rs
williamhbaker: Added that note, and also switched to `humantime` for the timeout.

crates/models/src/triggers.rs
williamhbaker: Changed to `run_id`! I don't think anything's been set in stone here, so that should be fine.

crates/validation/src/materialization.rs
williamhbaker: Oh yeah, that makes sense. Based on your other comment, I moved this decrypting into `runtime::materialize::connector`.  It would be straightforward to include the trigger spec in the Validate RPC and attempt a decryption there...but yeah I'll wait to do that as a follow-up if its warranted.

crates/runtime/src/materialize/protocol.rs
williamhbaker: I've re-structured things a bit so that the trigger completion only blocks just prior to where the rocksdb state update that would remove the persisted trigger variables is actually made durable in `persist_max_keys`. I think this is the latest it makes sense to specifically await their completion, being after the Load phase is nearly complete and just before sending Stores for the next phase.

crates/runtime/src/materialize/serve.rs
williamhbaker: Done!

williamhbaker:
Addressed feedback in a couple of `fixup!` commits that I will squash in before merging.

Two comments I couldn't reply to inline:
> A thought: what about a wall-clock transaction time, such as the start time of the transaction from its own stats? More information bearing, monotonic, but also unique.

I have incorporated this, making the `run_id` the start time of the transaction. It'll be an RFC3339 timestamp string which seems as good as anything, though I did consider making it a Unix integer timestamp instead just to make it a little more opaque. But it should be something we can change later if there's a need to.

> Right... for later, but reading this is making me realize that we may want resource paths to also be exposed to templates. That relates more closely to what the user may actually care about (which tables of mine were updated?).

This is almost certainly right, and collection names aren't really what people are going to want to know. Resource paths should probably be made available too as you suggest, and I'd imagine some other things we'll find users want as well. So probably in a follow-up we'll add resource paths & more.

PR #2802, opened 2026-03-23
agent: expose `DISABLE_IDLE_TASKS` and `DISABLE_FAILING_TASKS` separately

jshearer:
## Summary

- Replaces `DISABLE_ABANDONED_TASKS` with two independent env vars: `DISABLE_IDLE_TASKS` and `DISABLE_FAILING_TASKS`, both defaulting to `false`
- Allows enabling auto-disable for idle tasks (30d no data movement) without also auto-disabling chronically failing tasks, and vice versa

## Context

Discussion about abandoned task tracking surfaced that many chronically failing tasks are still actively moving data, and the upshot is that we should be able to enable idle-task disablement independently while we work on improving shard failure alerting before acting on chronically failing tasks.

### DIFF
--- a/crates/agent/src/integration_tests/abandoned_tasks.rs
+++ b/crates/agent/src/integration_tests/abandoned_tasks.rs
@@ -131,8 +131,14 @@ async fn test_abandoned_task_detection_and_resolution() {
     test_idle_fires_when_no_data_and_old(&mut harness, CatalogType::Capture, "pandas/capture")
         .await;
 
-    // Auto-disable runs last because it leaves the capture disabled.
-    test_auto_disable_publishes_disable(&mut harness, CatalogType::Capture, "pandas/capture").await;
+    // Auto-disable runs last because it leaves tasks disabled.
+    test_auto_disable_idle(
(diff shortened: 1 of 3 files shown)


crates/agent/src/integration_tests/abandoned_tasks.rs
GregorShear, line 1: the concept of "abandoned tasks" isn't going away, right? `DISABLE_ABANDONED_TASKS` is splitting into two more specific flags, but we'll still use the term "abandoned" generally? (mostly just trying to get up to speed here...)

GregorShear (approved):
Consider testing isolation between the two flags, i.e. a chronically failing task is _not_ disabled when the `idle` flag is on. Not a blocker...

crates/agent/src/integration_tests/abandoned_tasks.rs
jshearer: yeah that's right. I tried to think of naming for these env vars that contained the word `ABANDONED` but the ones I came up with were too wordy imo so I went with what I have here

PR #2803, opened 2026-03-24
Don't ban social users when migrating grants

GregorShear:
Revising the grant migration trigger I merged yesterday because the "this user is banned" error message is confusing.

We'll lock users out in phase 4c using a supabase auth hook where we can give a better error message.

### DIFF
--- a/supabase/migrations/20260324120000_sso_grant_no_ban.sql
+++ b/supabase/migrations/20260324120000_sso_grant_no_ban.sql
@@ -1,0 +1,88 @@
+-- Remove the ban on old social-auth accounts during SSO grant migration.
+-- Users should be able to continue signing in via social auth after their
+-- grants have been transferred to a new SSO identity.
+
+begin;
+
+-- Replace the trigger function without the ban step.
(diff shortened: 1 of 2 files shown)


supabase/migrations/20260324120000_sso_grant_no_ban.sql
```diff
@@ -0,0 +1,88 @@
...
+            ),
+            updated_at = now(),
+            detail = case
+              when excluded.capability > public.user_grants.capability
+              then excluded.detail
+              else public.user_grants.detail
+            end;
+    end if;
+  end loop;
+
```
GregorShear, line 83: Removed a few lines related to deleting refresh tokens and sessions, and also banning users from [this function](https://github.com/estuary/flow/blob/c81ef2de33be61cd358bfeb457da12ea88929f50/supabase/migrations/20260319120000_sso_grant_migration.sql#L99), which is otherwise unchanged.

GregorShear: here's a diff
https://www.diffchecker.com/d7ah3QGc/

PR #2805, opened 2026-03-24
SSO Phase 4d: Omit sso-enforced grants from non-sso users

GregorShear:
When `enforce_sso = true`, filter grants so that only users with an SSO identity matching the tenant's `sso_provider_id` can see grants for that tenant. This handles users whose email domain doesn't match the SSO domain — they aren't blocked at login (4c) but can't access enforcing tenants' resources. Must cover both auth channels:

- **PostgREST**: `internal.user_roles()` SQL function (used by `auth_roles()`)
- **GraphQL**: snapshot query in `snapshot.rs` — must use `auth.identities.provider_id` check (NOT the blanket `is_sso_user` flag)
- - 
**Verify:**

- [ ] Non-SSO user with grant on SSO-enforced tenant → grant filtered out, no access via either PostgREST or GraphQL
- [ ] SSO user on SSO-enforced tenant (matching provider) → grant visible, full access
- [ ] SSO user from Tenant A's provider → cannot see grants for SSO-enforced Tenant B
- [ ] Non-SSO user with grants only on non-enforced tenants → unaffected
- [ ] flowctl user with grants on both enforced and non-enforced tenants → only non-enforced grants visible (not fully locked out)
- [ ] estuary support has access via role grant (not user grant) so they're not blocked

### DIFF
--- a/crates/control-plane-api/src/server/snapshot.rs
+++ b/crates/control-plane-api/src/server/snapshot.rs
@@ -496,11 +496,21 @@ pub async fn try_fetch(
             g.object_role AS "object_role: models::Prefix",
             g.capability AS "capability: models::Capability"
         FROM user_grants g
+        WHERE NOT EXISTS (
+            SELECT 1 FROM tenants t
+            WHERE g.object_role ^@ t.tenant
+              AND t.enforce_sso
+              AND NOT EXISTS (
+                SELECT 1 FROM auth.identities ai
+                WHERE ai.user_id = g.user_id
+                  AND ai.provider = 'sso:' || t.sso_provider_id::text
+              )
+        )
         "#,
     )
     .fetch_all(pg_pool)
     .await
-    .context("failed to fetch role_grants")?;
+    .context("failed to fetch user_grants")?;
 
     let tasks = sqlx::query_as!(
         SnapshotTask,
--- a/.claude/skills/setup-mocksaml/SKILL.md
+++ b/.claude/skills/setup-mocksaml/SKILL.md
@@ -2,7 +2,6 @@
 name: setup-mocksaml
 description: Set up MockSAML as a local SAML IdP for testing SSO flows (grant migration, invite enforcement). Run when setting up a fresh local env or after `supabase db reset`.
 disable-model-invocation: true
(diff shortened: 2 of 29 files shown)


crates/control-plane-api/src/server/snapshot.rs
```diff
@@ -496,11 +496,21 @@ pub async fn try_fetch(
...
+        WHERE NOT EXISTS (
+            SELECT 1 FROM tenants t
+            WHERE g.object_role ^@ t.tenant
+              AND t.enforce_sso
+              AND NOT EXISTS (
+                SELECT 1 FROM auth.identities ai
+                WHERE ai.user_id = g.user_id
+                  AND ai.provider = 'sso:' || t.sso_provider_id::text
+              )
+        )
```
GregorShear, line 508: explain analyze before adding this where clause:
```
Seq Scan on user_grants g  (cost=0.00..16.00 rows=600 width=52) (actual time=0.024..0.026 rows=3 loops=1)
Planning Time: 0.152 ms
Execution Time: 0.089 ms
```

and after:
```
Nested Loop Anti Join  (cost=0.00..969850.12 rows=598 width=52) (actual time=0.047..0.048 rows=3 loops=1)
  Join Filter: (((g.object_role)::text ^@ (t.tenant)::text) AND (NOT (hashed SubPlan 2)))
  ->  Seq Scan on user_grants g  (cost=0.00..16.00 rows=600 width=52) (actual time=0.030..0.030 rows=3 loops=1)
  ->  Materialize  (cost=0.00..14.25 rows=170 width=48) (actual time=0.005..0.005 rows=0 loops=3)
        ->  Seq Scan on tenants t  (cost=0.00..13.40 rows=170 width=48) (actual time=0.015..0.015 rows=0 loops=1)
              Filter: enforce_sso
              Rows Removed by Filter: 1
  SubPlan 2
    ->  Seq Scan on identities ai  (cost=0.00..13.80 rows=380 width=48) (never executed)
Planning Time: 1.175 ms
Execution Time: 0.135 ms
```

GregorShear:
stale, closing

Issue #2780, opened 2026-03-17
Updates to `TargetNaming` behavior


jshearer:
Materializations have two separate settings that control where destination tables end up:

1. Endpoint config `schema`: a connector-level setting that controls where tables are created (e.g., Snowflake's `schema: "prod"`). The control plane treats this as opaque JSON and never reads its value.
2. `source.targetNaming`: a model-level setting ("Default Naming Convention" in the UI) that controls how collection names map to destination schemas and table names.

`TargetNaming` lives inside `SourceDef`, which is the `source` field on `MaterializationDef`:

```rust
pub enum TargetNaming {
    NoSchema,                    // alias: leaveEmpty
    WithSchema,                  // alias: fromSourceName
    PrefixSchema,
    PrefixNonDefaultSchema,      // DEFAULT
}
```

When bindings are added (either via sourceCapture auto-add or manually via the UI), the control-plane splits the collection name on `/` to get the last two components (schema and table), then injects values into the resource config JSON at locations annotated with `x-collection-name` and `x-schema-name` in the connector's resource config schema.

Given collection `acmeCo/mySchema/myTable`:
- `NoSchema`: `x-collection-name` = `myTable`, `x-schema-name` = (not set; connector decides, typically using endpoint config `schema` or a built-in default like `public`)
- `WithSchema`: `x-collection-name` = `myTable`, `x-schema-name` = `mySchema`
- `PrefixSchema`: `x-collection-name` = `mySchema_myTable`, `x-schema-name` = (not set; same connector fallback)
- `PrefixNonDefaultSchema`: like PrefixSchema, but omits prefix for `public` or `dbo`. `x-schema-name` = (not set; same connector fallback)

The populated resource config is then sent to the connector's Validate RPC. The connector parses it and returns a `resource_path` as a vector of strings (e.g., `["mydb", "mySchema", "myTable"]`). The control plane stores this path in `/_meta/path` in the model's resource config JSON and uses it as the persistent identity for correlating bindings across publications, computing state keys and journal read suffixes.

When `x-schema-name` is left unset, the connector itself decides what schema to use, typically falling back to whatever `schema` is set in its endpoint config.

### Problems

1. Users set `schema: "prod"` in endpoint config, expect their data in `prod`, and are surprised when tables end up in source-mirrored schemas because `targetNaming` is `WithSchema`. The endpoint config `schema` then only affects system tables like `flow_checkpoints_v1`.

[...]

PR #2809, opened 2026-03-25
Target naming phase 1

### LINKED ISSUE #2234: models: support arbitrary prefixes and suffixes for table names in `targetNaming`
The original ask from @dyaffe was to support something like:

```
source:
  targetNaming:
    prefix: 'some_prefix_'
    suffix: '_some_suffix'
```

which given the collection `acmeCo/foo/bar` would result in table names like `some_prefix_bar_some_suffix`. I'm currently thinking that perhaps we should represent it as a `nameTemplate: 'some_prefix_{{name}}_some_suffix'`, because then it would allow the template to also reference the full collection name or the "schema" component. Then the question is whether we should also support a `schemaTemplate` for connectors that use `x-schema-name`. We should probably settle on a design before we proceed.

jshearer:
## Summary

Phase 1 of the target naming redesign ([#2780](https://github.com/estuary/flow/issues/2780)). This PR adds `TargetNamingStrategy` as a new top-level field on `MaterializationDef`, decoupling naming behavior from `SourceDef` so that materializations without a sourceCapture can express a naming preference. It also adds optional template fields for decorating generated table and schema names.

### What changed

`TargetNamingStrategy` is an enum with three variants:
- `matchSourceStructure { tableTemplate, schemaTemplate }`: schema from collection name, table is the final component. Both names are derived, so both support templates.
- `singleSchema { schema, tableTemplate }`: all tables in one named schema. Schema is user-provided, so only the table name supports a template.
- `prefixTableNames { schema, skipCommonDefaults, tableTemplate }`: schema component prefixed to table name. Schema is user-provided, so only the table name supports a template.

Template fields use mustache-style substitution (e.g. "staging_{{table}}" produces staging_myTable). tableTemplate uses {{table}} as the placeholder, schemaTemplate uses {{schema}}. When absent, names are used as-is.`singleSchema` and `prefixTableNames` have no schema template because their schema is already a user-provided string.

The existing `SourceDef.target_naming` enum is preserved as a fallback. When both `MaterializationDef.target_naming` and `source.target_naming` are present, the top-level field wins for naming; `delta_updates` is always read from `source`.

### Rollout strategy

This PR ships only the model and naming logic. There is no auto-promotion of existing specs in the validation layer. Clients (the UI) can adopt the new field progressively: present and edit `targetNaming` if present, otherwise fall back to the existing `source.targetNaming` field. A later phase will populate the remaining materializations' root `target_naming` based on knowledge of each connector's endpoint-config `schema` location and default.

### WASM API

The `update_materialization_resource_spec` WASM function now accepts `targetNaming` alongside `sourceCapture`, both optional. `sourceCapture` is optional because not all materializations have one, not because it's being replaced. Existing UI code that sends `sourceCapture` continues to work unchanged.

### Note on `flow.schema.json`

The regenerated schema includes `Triggers`, `TriggerConfig`, and `HttpMethod` types that were already in the models but missing from the previously-stale `flow.schema.json`. These are unrelated to this PR.

Closes https://github.com/estuary/flow/issues/2234

### DIFF
--- a/crates/validation/src/materialization.rs
+++ b/crates/validation/src/materialization.rs
@@ -114,6 +114,7 @@ async fn walk_materialization<C: Connectors>(
     let models::MaterializationDef {
         on_incompatible_schema_change,
         source: sources,
+        mut target_naming,
         endpoint,
         bindings: bindings_model,
         mut shards,
@@ -122,6 +123,20 @@ async fn walk_materialization<C: Connectors>(
         delete: _,
         reset,
     } = model;
+
+    // Model fix #1: Promote source.target_naming == WithSchema to top-level
+    // MatchSourceStructure. This is safe because MatchSourceStructure and WithSchema
+    // represent identical behavior.
+    if target_naming.is_none() {
+        if let Some(source) = &sources {
+            if source.to_normalized_def().target_naming == models::TargetNaming::WithSchema {
+                target_naming = Some(models::TargetNamingStrategy::MatchSourceStructure);
+                model_fixes.push(
+                    "promoted source.targetNaming 'withSchema' to top-level targetNaming 'matchSourceStructure'".to_string(),
+                );
+            }
+        }
+    }
 
     indexed::walk_name(
         scope,
@@ -624,6 +639,7 @@ async fn walk_materialization<C: Connectors>(
     };
     let model = models::MaterializationDef {
         source: sources,
+        target_naming,
         on_incompatible_schema_change,
         endpoint,
         bindings: bindings_model,
--- a/crates/validation/src/materialization.rs
+++ b/crates/validation/src/materialization.rs
@@ -114,7 +114,7 @@ async fn walk_materialization<C: Connectors>(
     let models::MaterializationDef {
         on_incompatible_schema_change,
         source: sources,
-        mut target_naming,
+        target_naming,
         endpoint,
         bindings: bindings_model,
         mut shards,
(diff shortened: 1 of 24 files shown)


crates/validation/src/materialization.rs
```diff
@@ -123,6 +124,20 @@ async fn walk_materialization<C: Connectors>(
...
+    if target_naming.is_none() {
+        if let Some(source) = &sources {
+            if source.to_normalized_def().target_naming == models::TargetNaming::WithSchema {
+                target_naming = Some(models::TargetNamingStrategy::MatchSourceStructure);
+                model_fixes.push(
+                    "promoted source.targetNaming 'withSchema' to top-level targetNaming 'matchSourceStructure'".to_string(),
+                );
+            }
+        }
+    }
```
mdibaiee: am I right that this is going to still retain `source.targetNaming` while adding the new top-level `targetNaming`? Can we get rid of the old one automatically as well?

jshearer: The intent is that it will leave `source.targetNaming` alone specifically in order to allow this to be progressively released. [For example](https://github.com/estuary/flow/pull/2809/changes#diff-7c92cad7a8f7fc6a517452cd8aa65126407e14179dc0c4f55494258b01c2671fR58-R77), in `update_materialization_resource_spec()` we first check if the new `targetNaming` exists, then fall back to the original check against `source.targetNaming`.

The intent is to allow this PR to be released, then the UI to be updated to use the new `targetNaming`, though I'm realizing as I write that that we actually need to split this up even further: we need to _introduce_ the new `targetNaming` without even this trivial upgrade, then the UI needs to be updated to handle it, _then_ we can introduce the model fix(es). Reason being, otherwise we would do this one-time upgrade, but the UI would continue to use `source.targetNaming`.

I will remove this model fix until the UI is updated

mdibaiee (approved):

jshearer:
Based on convos with with @travjenkins, @jgraettinger and @dyaffe, I added optional `tableTemplate` and `schemaTemplate` fields to the `TargetNamingStrategy` variants to enable prefixes/suffixes using placeholders that resemble Mustache templating:

```json
{
  "strategy": "matchSourceStructure",
  "tableTemplate": "staging_{{table}}",
  "schemaTemplate": "analytics_{{schema}}"
}
```

`{{table}}` and `{{schema}}` are replaced with the computed name. When the field is absent, names are used as-is (no change to existing behavior).

Which variants get which templates:
- `matchSourceStructure`: `tableTemplate` + `schemaTemplate` (both names are derived from the collection path)
- `singleSchema` / `prefixTableNames`: `tableTemplate` only (schema is already user-provided)

The intent behind matching mustache/handlebars is that if we decide to add full-fat templating in the future, we should be able to do that easily.

Even though this PR is approved, this isn't a trivial change so I'd like to get another ✅

mdibaiee (approved):
The templating change LGTM 👍🏽

Issue #2794, opened 2026-03-19
Expand the alert type metadata exposed via GraphQL


kiahna-tucker:
Update the alert type metadata returned by the GraphQL endpoint used to fetch the comprehensive set of possible alert types. The new model will include the following information for a given alert type:

* The canonical enumeration identifier (e.g., `shard_failed`)

* An indication of whether the alert type is a system alert (e.g., billing-related alert types, `free_trial`, etc.)

* An indication of whether the alert type is subscribed to by default

* A suitable, user-facing name for display

* A description

A sample of the updated model is as follows:

```ts
{
  description: string;
  displayName: string;      // The user-facing name for display
  isDefault: boolean;
  isSystemAlert: boolean;
  name: AlertType;          // The canonical enumeration identifier
}
```

PR #2811, opened 2026-03-26
Add alert type GraphQL query

GregorShear:
## Summary
- Add a new `alertTypes` GraphQL query that returns all possible alert types with user-facing metadata (title, description)
- Update subscription mutations to add system alerts to all subscriptions (i.e. system alerts are non-optional)

```gql
query {
  alertTypes {
    alertType,
    description,
    displayName,
    isDefault,
    isSystem
  }
}
```

**NOTE:** Pagination will not be implemented for this endpoint because of the simplicity of the data returned.

### Issues

The issues directly below are completely resolved by this PR:
https://github.com/estuary/flow/issues/2794


crates/control-plane-api/README.md
GregorShear, line 1: @kiahna-tucker @travjenkins

jshearer (changes requested):
- What are we doing about `ShardFailed`? Currently, we have been treating it as a beta feature that customers can opt into by asking us. This PR marks it as both default and system, meaning that not only is it on by default, you actually can't turn it off. That seems surprising. Did you talk about this w/ Dave?
- There are now two different sources of truth for the default alerts: this `AlertType::is_default()` and the [database column](https://github.com/estuary/flow/blob/1e778661a276180c691bf7f551e4fbcf40494f4e/supabase/migrations/01_compacted.sql#L3999-L4007) defaults. And in fact since `ShardFailed` is marked as default in the PR, those are divergent now. Today, the DB column default is what gets used when provisioning new tenants. I note that despite introducing a concept called `is_default`, this PR doesn't actually change the behavior of the system to use it anywhere.
- How are we thinking about the migration here now that we're introducing `is_system`/`is_default`? Once we merge this, there will be a lot of existing tenants with, for example, missing alert subscriptions that are marked as `is_system`. Will they be able to... add them in the UI, but just not remove them? Or what happens if someone modifies this in the DB, which does happen.
- There's an implicit relationship between system and default alerts: every system alert is also necessarily a default alert, but not every default alert is necessarily a system alert. It might be worth modeling this, or at least commenting on it? WDYT?
- I don't love how the description string can be/is different from the doc-comment for these alerts. I wonder if there's any way you/Claude can find to "reflect" the doc comment here? If not no big deal, but IMO it would be worth at least trying.

Put another way, I don't think this is coherent with the existing architecture, and I'm asking you to re-think how it should work so we don't introduce competing concepts and tech debt.

PR #2813, opened 2026-03-27
ci: add changed page links to docs PR preview comments

jwhartley:
## Summary
After deploying a docs preview, posts a comment listing up to 5 changed pages with direct links to the preview URL. Resolves slugs from frontmatter when present, falls back to file path.

Includes a trivial docs change to trigger the workflow for testing - you can see the '📄 Changed pages:' addition in the below comment

## How it works
- Filters changed files to `.md`/`.mdx` under `site/docs/`
- Reads `slug:` from frontmatter if present, otherwise derives path from filename
- Handles `README.md` → parent directory, deleted files, and slug edge cases
- Shows "... and N more" if >5 pages changed

### DIFF
--- a/.github/workflows/docs-ci-previews.yaml
+++ b/.github/workflows/docs-ci-previews.yaml
@@ -100,5 +100,5 @@ jobs:
         uses: marocchino/sticky-pull-request-comment@v2
         with:
           message: |
-            🚀 **Preview deployed** to ${{ steps.preview.outputs.preview-url }}
+            🚀 **Preview deployed** to https://docs.estuary.dev/pr-preview/pr-${{ github.event.number }}/
             ${{ steps.pages.outputs.result }}


.github/workflows/docs-ci-previews.yaml
```diff
@@ -39,3 +39,88 @@ jobs:
         uses: rossjrw/pr-preview-action@v1
         with:
           source-dir: ./site/build/
+      - name: Comment with changed pages
+        if: github.event.action != 'closed'
```
aeluce, line 48: Should more of the previous steps be conditional like this? (Are we currently trying to build and deploy when a PR gets closed? 🤔)

.github/workflows/docs-ci-previews.yaml
```diff
@@ -39,3 +39,88 @@ jobs:
...
+              execSync('git fetch origin master --depth=1');
+            } catch (e) {
+              core.warning(`Failed to fetch origin/master: ${e.message}`);
+              return;
+            }
+
+            // Get all changed docs files, filter to md/mdx, exclude deleted files
+            const allFiles = execSync('git diff --name-only origin/master -- site/docs/')
+              .toString().trim().split('\n')
+              .filter(f => /\.(md|mdx)$/.test(f))
```
aeluce: I don't believe we currently have any `.mdx` files in docs anymore, though that could always change again in the future.

.github/workflows/docs-ci-previews.yaml
```diff
@@ -39,3 +39,88 @@ jobs:
...
+              }
+              if (!slug.startsWith('/')) slug = '/' + slug;
+              return `- [${slug}](${base}${slug})`;
+            });
+
+            const suffix = total > 5 ? `\n- ... and ${total - 5} more` : '';
+            const pagesBlock = `\n\n📄 **Changed pages:**\n${links.join('\n')}${suffix}`;
+
+            // Append to the PR Preview Action comment to keep the PR tidy.
+            // Retry a few times since the preview action posts its comment asynchronously.
```
aeluce: It feels messy to find the right comment by searching for a hardcoded message. Is there a better way to do this?

We could potentially [manage the sticky comment](https://github.com/rossjrw/pr-preview-action?tab=readme-ov-file#customise-the-sticky-comment) ourselves so that we create the customized version all at once rather than create the generic comment, search for it, then edit it. I believe that would also mean we wouldn't have to worry about stripping the comment body on re-runs or defaulting to a regular comment if we can't find the PR preview one.

.github/workflows/docs-ci-previews.yaml
jwhartley: Good catch — added `if: github.event.action \!= 'closed'` to the build, install, and setup-node steps. No point building the site on PR close.

.github/workflows/docs-ci-previews.yaml
jwhartley: Removed — no .mdx files in docs currently.

.github/workflows/docs-ci-previews.yaml
jwhartley: Agreed — refactored to use `marocchino/sticky-pull-request-comment` with `comment: false` on the preview action. Now we build the full comment (preview link + changed pages) in one shot. No more string matching, retries, or regex stripping.

PR #2827, opened 2026-04-02
Remove mention of service accounts in the data plane dialog 

GregorShear:
Service account IDs are being removed from the dataplane dialog in favor of the new storage mapping flow


aeluce (approved):
LGTM.

Also FYI @jwhartley the link preview action didn't manage to construct the full slug: `/installation` should be `/getting-started/installation`.

PR #2833, opened 2026-04-07
Transfer refresh tokens to SSO users

GregorShear:
# Summary 

This is a bit of a grab bag of small updates related to the recent SSO work I've been doing.

- Modify `internal.on_sso_identity_insert()` to transfer refresh tokens from the old social account to the new SSO user, so existing CLI/API tokens keep working after SSO enforcement locks out social logins
- Change invite link redemption to gate SSO requirement on `tenants.enforce_sso` instead of `tenants.sso_provider_id IS NOT NULL`, aligning with the rest of the SSO enforcement logic
- Add unique constraint on `tenants.sso_provider_id` to prevent multiple tenants from sharing the same IdP
- Update `mise` scripts for configuring saml in a local dev environment


supabase/migrations/20260406120000_sso_transfer_refresh_tokens.sql
```diff
@@ -0,0 +1,96 @@
...
+  update public.refresh_tokens set user_id = NEW.user_id where user_id = old_user_id;
+  delete from auth.sessions where user_id = old_user_id;
+
+  return null;
+end;
+$$;
+
+-- Ensure each SSO provider is linked to at most one tenant.
+alter table public.tenants
+  add constraint tenants_sso_provider_id_unique unique (sso_provider_id);
```
GregorShear, line 88: adding a constraint here to make sure sso providers don't get mapped to multiple tenants by accident

supabase/migrations/20260406120000_sso_transfer_refresh_tokens.sql
GregorShear, line 1: here's a diff with the last version of this `on_sso_identity_insert` function.

also note the new unique constraint on `tenants.sso_provider_id` at the bottom.

https://www.diffchecker.com/wg6VSKmT/

supabase/migrations/20260406120000_sso_transfer_refresh_tokens.sql
```diff
@@ -0,0 +1,90 @@
...
+            detail = case
+              when excluded.capability > public.user_grants.capability
+              then excluded.detail
+              else public.user_grants.detail
+            end;
+    end if;
+  end loop;
+
+  -- Transfer estuary refresh tokens to new account
+  update public.refresh_tokens set user_id = NEW.user_id where user_id = old_user_id;
```
GregorShear, line 80: Any security concerns here? This will be invoked only when the customer's IdP authenticates an SSO user with an email that matches an existing social user.

Enforcing SSO will draw attention to the jankiness of these refresh tokens being tied to a specific user, so my hope is to streamline the process a bit, at least until we have proper service accounts.

mise/tasks/local/saml
GregorShear, line 1: This script is a work in progress - it needs testing on other machines, especially non-macos. I also feel like something breaks when restarting the mise stack, but i haven't looked into it much yet.

jshearer (approved):
LGTM

PR #2839, opened 2026-04-08
docs: prefix-based access control guide

### LINKED ISSUE #2759: docs: add practical guide for sub-prefix access control
## Summary

The [Authorizing Users](https://docs.estuary.dev/reference/authentication/) page explains the concepts of capabilities, prefixes, and inheritance, but doesn't provide a practical how-to for common access control patterns. Customers are asking how to set this up and getting confused by the two layers (user access vs. task access).

## What's missing

A guide covering common patterns like:

### 1. Restricting a user to a sub-prefix
- Grant user write/admin to `acmeCo/team-a/` instead of `acmeCo/`
- User can only create/modify tasks under that prefix

### 2. Isolating sub-prefixes from each other (e.g. staging vs. production)
- The distinction between **user grants** (what tasks you can create) and **role grants** (what collections a task can access)
- By default, `acmeCo/` has write to itself, so all tasks inherit access to everything — restricting the user alone is not enough
- To fully isolate: remove the default tenant-level write-to-self grant, replace with per-sub-prefix grants
- Example: `acmeCo/staging/` write to `acmeCo/staging/`, `acmeCo/prod/` write to `acmeCo/prod/`

### 3. Additional considerations
- Data plane grants may be needed for sub-prefixes
- Storage mapping inheritance — sub-prefixes inherit from parent, but the UI may not show available prefixes without explicit storage mappings
- The difference between what the UI shows vs. what validation actually enforces

## Context

A customer asked how to restrict materializations in a sub-prefix to only access collections within that same sub-prefix. The existing docs didn't answer this, and the answer is nuanced.

jwhartley:
## Summary

- Adds a new **Access Control** sidebar section, moving the existing "Authorizing Users" page there
- Adds a new guide: **Prefix-based access control** covering the four layers of sub-prefix isolation and three common scenarios

## Context

Customers frequently ask how to restrict users and tasks to sub-prefixes (e.g. staging vs. prod, regional isolation). The existing auth docs explain capabilities conceptually but don't cover practical setup or the interaction between the four layers (user grants, role grants, storage mappings, data plane access).

This is raw content for @aeluce to take over — location, structure, and format are all up for grabs.

## Scenarios covered

1. **Default** — single prefix, no isolation (what every new org starts with)
2. **Environment isolation with cross-read** — dev/prod where dev tasks can read prod collections
3. **Full isolation** — EU/US with separate storage, data planes, and per-user admin scope

## Notes

- Technically reviewed against internal Slack threads with Johnny (confirmed accuracy of the four-layer model, grant additivity, longest-prefix-match for storage/data planes)
- Data plane UI visibility bug (RLS → GraphQL migration) is mentioned in a note
- Private link limitation is called out in the Limitations section


aeluce (approved):
Ended up just going with some minor changes from my end for now. As noted in the [linked issue](https://github.com/estuary/flow/issues/2759), upcoming auth changes will likely see the new access control section rewritten and expanded. Some info on those upcoming changes available [here](https://github.com/estuary/flow/blob/greg/plans/plans/orthogonal-authz.md).

PR #2847, opened 2026-04-10
controller: clean up open alerts when a spec is deleted

### LINKED ISSUE #2846: The `active alerts` GQL query returns deleted tasks
There is an issue where a task that had acive alerts being deleted will still be returned when we look for all `active alerts` in a `prefix`. This ends up showing in the welcome page and users might click on the task to see more details but will be unable to see anything.

The UI is querying with these variables: `{active: true, prefix: "acme_co/"}`

This task has been deleted (reviewed calls made by user) and should not be shown.
Image

jshearer:
When a task with active alerts is deleted, the alert resolution actions were never applied, leaving orphaned rows in `alert_history` with `resolved_at = NULL`. The `alertHistory` query then returned these phantom alerts for tasks that no longer exist.

* Add `cleanup_open_alerts()` that sets `resolved_at` and deletes the associated notification tasks without sending resolution emails
* Call it in `Outcome::apply` before the hard delete when `live_spec_deleted` is true
* Add a one-time migration to resolve existing orphaned alerts and their notification tasks

Fixes #2846

### DIFF
--- a/.sqlx/query-36cf5f128777775f0a77bd1d1eb81f78ea62aeba6d29d1e02ee8d3fbef96a1ff.json
+++ b/.sqlx/query-36cf5f128777775f0a77bd1d1eb81f78ea62aeba6d29d1e02ee8d3fbef96a1ff.json
@@ -1,0 +1,28 @@
+{
+  "db_name": "PostgreSQL",
+  "query": "\n        delete from internal.tasks t\n        where t.task_id = any($1::flowid[])\n        ",
+  "describe": {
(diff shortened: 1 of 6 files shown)


travjenkins (approved):
lgtm - had claude take a look and sent 1 callout over to Joseph but that should be fine. It called out `catalog_name` _could_ come as an empty string if the `live_spec` was hard deleted before the controlle rran.

I would also prefer a rust focused person taking a look as well.

PR #2849, opened 2026-04-13
A couple of billing updates

jshearer:
* Finally got around to adding a real `--dry-run` mode. Running against the Stripe testmode API key was starting to cause too many issues. See more details in the commit message

  The invoice generator's trial-run workflow ran against a sandbox Stripe account, but produced inaccurate results because the sandbox lacked livemode invoice state (manual bills with existing open/paid invoices appeared as phantom creates). `--dry-run` runs against livemode Stripe in read-only mode, showing what would happen without creating invoices, customers, or modifying anything.
* Change the behavior for manual bills. Contract customers don't want us to use their payment method on file to pay for their contracted invoices, just for monthly overages. So from now on, all `Manual` bills will be forced to `charge_type: SendInvoice` during the final send/charge step.

### DIFF
--- a/crates/billing-integrations/src/publish.rs
+++ b/crates/billing-integrations/src/publish.rs
@@ -500,6 +500,14 @@ impl Invoice {
         }
 
         // Create or reuse the invoice
+        // Manual invoices should always be sent as invoices rather than
+        // charged to the customer's payment method.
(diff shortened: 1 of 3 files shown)


jshearer:
rolling into https://github.com/estuary/flow/pull/2883 is simpler

PR #2850, opened 2026-04-13
Insert role grants for private data planes (and tasks)

### LINKED ISSUE #2848: Storage mapping wizard private plane grants
When a storage_mapping is created at a sub-prefix (e.g. acmeCo/qa/staffing-solutions/) referencing the tenant's private data plane, no corresponding `role_grant` is created. The mapping saves fine, but subsequent `flowctl publish` runs will return:

> data plane ops/dp/private/acmeCo/aws-us-east-2-c1, referenced by storage mapping acmeCo/qa/staffing-solutions/, was not found; did you mean data plane ops/dp/public/noop?

The only role grant created at provisioning is `(acmeCo/, ops/dp/private/acmeCo/, read)`. Nothing creates the equivalent at the subprefix level, so the publish-time data plane filter in specs.rs cant bridge the subprefix catalog -> private data plane:
https://github.com/estuary/flow/blob/61635c47300f46d987560d5b60309a7addab5852/crates/control-plane-api/src/publications/specs.rs#L862-L900

Fix:
In [`create_storage_mapping`/`update_storage_mapping`](https://github.com/estuary/flow/blob/61635c47300f46d987560d5b60309a7addab5852/crates/control-plane-api/src/server/public/graphql/storage_mappings.rs#L165), when the spec references a private data plane, also insert into `role_grants` `(subject = catalog_prefix, object = plane_prefix, 'read')`

Workaround right now:
```
insert into role_grants (subject_role, object_role, capability, detail)
values ('<sub_prefix>', '<plane_prefix>', 'read', 'Sub-tenant access to private plane');
```

GregorShear:
## Summary

When private data planes are created, we also create a `role_grant` to give the tenant access to the data plane: `acmeCo/` -> `ops/dp/private/acmeCo/` : `read` 

This _should_ give read access to the entire subject (`acmeCo/` and all sub-prefixes), but there's a [limitation in how `internal.user_roles()` works](https://github.com/estuary/flow/blob/61635c47300f46d987560d5b60309a7addab5852/supabase/migrations/01_compacted.sql#L2245) and the result is someone with access to `acmeCo/nested/` won't be able to see their private data plane.

GQL operations will resolve these grants correctly using the snapshot, but the remaining postgrest API consumers (there are many) need an explicit grant from the sub-prefix to `ops/dp/private/acmeCo/` (and `ops/tasks/private/acmeCo/` for the same reason).

This PR preemeptively adds those role grants when a user redeems an invite link (with admin capabilities). It's true this will make a lot of redundant and unnecessary role grants (particularly for tenants with no private data planes) but this is a temporary measure with limited impact - fewer than 100 admin user grants point to a sub-prefix (meaning that if we had been creating these role grants from the beginning we'd have fewer than 100 additional role grants than we have today).

Closes #2848

### DIFF
--- a/.sqlx/query-e2c4cc44ea85f1fc17e1e610c51c51086db6117e08b189fef33923ea9532d3f0.json
+++ b/.sqlx/query-e2c4cc44ea85f1fc17e1e610c51c51086db6117e08b189fef33923ea9532d3f0.json
@@ -1,0 +1,15 @@
+{
+  "db_name": "PostgreSQL",
+  "query": "\n        INSERT INTO role_grants (subject_role, object_role, capability, detail)\n        SELECT $1::text, object, 'read', 'sub-prefix access to private data plane'\n        FROM UNNEST($2::text[]) AS t(object)\n        ON CONFLICT DO NOTHING\n        ",
+  "describe": {
+    "columns": [],
+    "parameters": {
+      "Left": [
+        "Text",
+        "TextArray"
+      ]
+    },
+    "nullable": []
+  },
+  "hash": "e2c4cc44ea85f1fc17e1e610c51c51086db6117e08b189fef33923ea9532d3f0"
+}
--- a/crates/control-plane-api/src/server/public/graphql/invite_links.rs
+++ b/crates/control-plane-api/src/server/public/graphql/invite_links.rs
@@ -322,6 +322,13 @@ impl InviteLinksMutation {
         )
         .await?;
 
+        // When the invite grants Admin, ensure the prefix has explicit read
+        // grants to its tenant's private data plane and ops-tasks prefixes.
+        // See `ensure_private_data_plane_grants` for why this workaround exists.
+        if invite.capability == models::Capability::Admin {
+            ensure_private_data_plane_grants(&mut *txn, &invite.catalog_prefix).await?;
+        }
+
         txn.commit().await?;
 
         tracing::info!(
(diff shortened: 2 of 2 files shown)


crates/control-plane-api/src/server/public/graphql/storage_mappings.rs
```diff
@@ -53,6 +53,82 @@ pub struct UpdateStorageMappingResult {
...
+    // Derive the tenant from the catalog prefix's first segment.
+    let Some((tenant, _)) = catalog_prefix.as_str().split_once('/') else {
+        return Ok(());
+    };
+    if tenant.is_empty() {
+        return Ok(());
+    }
+    let grant_objects = vec![
+        format!("ops/dp/private/{tenant}/"),
+        format!("ops/tasks/private/{tenant}/"),
```
GregorShear: @SeanWhelan your ticket was just about `ops/dp/...` but i think the same reasoning would apply to `ops/tasks/...` is this something you're familiar with? it's a bit unfamiliar to me, but i'm basing this on the grants inserted when provisioning a data plane [here](https://github.com/estuary/flow/blob/61635c47300f46d987560d5b60309a7addab5852/crates/control-plane-api/src/server/create_data_plane.rs#L156)

SeanWhelan: Yes, good catch. The exact same applies to `ops/tasks/`

jshearer (approved):
LGTM. We talked about this, and the tl;dr is: RLS only looks "down", not "up", so if you ONLY have a grant to `acmeCo/foo`, but `acmeCo/` has a role grant to a data-plane, you won't be able to see that plane. This checks to see if you're accepting an invite link to a sub-perfix, and if you are, it adds a fallback grant to let you access data-planes under `ops/dp/private/{your_tenant}`. 

One note is that private data-planes living under that prefix of `ops/dp/private/{tenant_name}` is purely by convention, and this workaround would _not_ automatically let you use a data-plane under a different sub-prefix of `ops/dp/private/`, even if your tenant has a grant to it.

NBD since the intent is to get rid of this once everything is migrated to GraphQL

PR #2853, opened 2026-04-14
docs: note QUOTED_IDENTIFIERS_IGNORE_CASE incompatibility for Snowflake materialization

jwhartley:
Adds a one-line prerequisite note that `QUOTED_IDENTIFIERS_IGNORE_CASE` must not be enabled for the Estuary user when using the Snowflake materialization connector.

Related connector fix: estuary/connectors#4243


danielnelson (approved):
I wouldn't object to going into more detail here about how to correct the issue here, check my comments on the associated PR.

PR #2864, opened 2026-04-15
Target naming migration

jshearer:
This is the migration for target naming that I talk about in #2780. I decided to frame it as a new, temporary `flowctl raw` subcommand that analyzes existing materializations and determines the correct `TargetNamingStrategy` based on its current `source.targetNaming`, endpoint config, and built resource paths. By default it prints a dry-run report. With `--execute`, it publishes the changes one materialization at a time.

## Migration classification

Every materialization is classified into one of:

- **MIGRATE**: `target_naming` and per-binding `x-schema-name` can be set automatically without causing unintended backfilling.
- **MANUAL**: Either we _can_ determine the correct settings, but applying them would cause resource paths to change from 1-element (`[table]`) to 2-element (`[schema, table]`), or we can't determine the correct schema automatically (no endpoint config schema, no consistent schema in resource paths).
- **SKIP_NO_SCHEMA**: connector doesn't support `x-schema-name` (no schema pointer in its resource spec).
- **SKIP_NOT_CONNECTOR**: Dekaf
- **SKIP_ALREADY_SET** / **SKIP_DISABLED_NO_BUILT_SPEC**: already migrated, or disabled with no built spec to analyze.

## Strategy selection rules

`TargetNamingStrategy` is derived from `source.targetNaming`:

| `source.targetNaming` | Proposed strategy |
|---|---|
| `WithSchema` | `MatchSourceStructure` |
| `PrefixSchema` | `PrefixTableNames { schema, skip_common_defaults: false }` |
| `PrefixNonDefaultSchema` | `PrefixTableNames { schema, skip_common_defaults: true }` |
| `NoSchema` | `SingleSchema { schema }` |
| No source capture | `MatchSourceStructure`, falling back to `SingleSchema` if existing bindings conflict |

For strategies that require a schema value, the tool resolves it from (in order):
* the endpoint config's schema/dataset/namespace field
* a unanimous schema detected from existing built resource paths
* the connector's well-known default (Snowflake: `PUBLIC`).

### Filling x-schema-name on existing bindings

`targetNaming` controls how future bindings get their schema and table names. Existing bindings need `x-schema-name` filled in separately to match where their data actually lives. Once the control plane can guarantee that `x-schema-name` is populated on every binding, connectors/#3977 can delete the fallback logic, and the endpoint-config `schema` shrinks to its real job (system tables like `flow_checkpoints_v1`). The migration is the prerequisite that lets that cleanup happen without changing any existing binding's resource path.

For many existing bindings, the `TargetNamingStrategy`-generated schema matches the actual schema in the resource path, and x-schema-name can be simply set to that value. But for some bindings, the two diverge: a binding created before `x-schema-name` existed, or via a code path that didn't populate it, would have been placed in whatever schema the endpoint config specified, which may not match what the strategy would derive from the collection name.

When the customer explicitly set `source.targetNaming`, the tool preserves their strategy for future bindings but fills in the existing schema (from the built resource path) on existing bindings where the strategy-derived value would conflict. The report flags these as `(actual; strategy would produce "..." for new bindings)`.

When no `source.targetNaming` was set and a binding's collection-derived schema doesn't match its resource path schema, the tool falls back to `SingleSchema` with the resolved endpoint schema. If the endpoint schema also doesn't match the resource path schemas, the task is marked as `MANUAL`.

## Snowflake compatibility mode handling

`materialize-snowflake` uniquely produces 1-element resource paths (`[table]`) when the binding's schema matches the endpoint-config's schema, and 2-element paths (`[schema, table]`) otherwise. The migration tool mirrors the connector's logic to determine whether setting `x-schema-name` would preserve or change the resource path. When the endpoint config has no explicit schema, the tool assumes Snowflake's default of `PUBLIC`.

## Disabled materializations

Disabled materializations with a built spec are analyzed normally. Disabled materializations without a built spec are skipped entirely, as they're old enough that re-enabling them at this point would almost certainly require a backfill anyway.

## Execute mode

With `--execute`, the tool publishes each `MIGRATE` materialization individually:
- Re-fetches the spec and verifies `last_pub_id` hasn't changed
- Sets `targetNaming` on the materialization
- Fills in `x-schema-name` on bindings that are missing it
- Publishes via `draft_specs`

### DIFF
--- a/crates/flowctl/src/migrate_target_naming.rs
+++ b/crates/flowctl/src/migrate_target_naming.rs
@@ -1,0 +1,1498 @@
+use anyhow::Context;
+use futures::stream::{self, StreamExt};
+use models::{MaterializationDef, MaterializationEndpoint, TargetNaming, TargetNamingStrategy};
+use proto_flow::flow::MaterializationSpec;
+use serde_json::Value;
+use std::collections::{BTreeMap, HashMap};
+
+/// Migrate existing materializations to use explicit targetNaming strategies.
+///
+/// Analyzes all materializations and determines the appropriate TargetNamingStrategy
+/// based on current source.targetNaming and endpoint configuration.
+/// Prints a detailed report of every decision. Without --execute, this is a dry run.
+#[derive(Debug, clap::Args)]
+#[clap(rename_all = "kebab-case")]
+pub struct MigrateTargetNaming {
+    /// Filter to materializations under a catalog prefix (e.g. "acmeCo/" or "acmeCo/prod/my-mat").
+    #[clap(long)]
+    prefix: Option<String>,
+    /// Actually publish the migration changes. Without this flag, only a dry-run report is printed.
+    #[clap(long)]
+    execute: bool,
+}
+
+#[derive(Debug)]
+enum Action {
+    /// Connector doesn't support schemas (no x-schema-name in resource spec).
+    SkipNoSchemaSupport,
+    /// targetNaming is already set on this materialization.
+    SkipAlreadySet,
+    /// Not a connector-based materialization (Dekaf or Local).
+    SkipNotConnector,
+    /// Task is disabled and has no built spec, so there's nothing to analyze.
+    SkipDisabledNoBuiltSpec,
+    /// Connector supports x-schema-name but isn't in the connector_info() map,
+    /// so we can't safely compute schema_idx or the compat-path behavior.
+    SkipUnknownConnector,
+    /// Can't determine schema automatically.
+    NeedsManualIntervention { reason: String },
+    /// Standard migration: set targetNaming + fill x-schema-name.
+    Migrate,
+}
+
+struct MaterializationAnalysis {
+    catalog_name: String,
+    connector_image: Option<String>,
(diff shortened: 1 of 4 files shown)


crates/flowctl/src/migrate_target_naming.rs
```diff
@@ -0,0 +1,1482 @@
...
+
+    // would_change_path: filling in x-schema-name would change the resource path
+    // (e.g. ["my_table"] becomes ["public", "my_table"]), but the data still goes
+    // to the same logical destination. Resource paths are used as state keys,
+    // so changing the path means the connector loses track of its checkpoint for that binding.
+    // That requires a backfill (via feature flags + bumping the backfill counter).
+    //
+    // Exception: disabled tasks whose path-changing bindings all have synthetic
+    // "binding-N" noop paths can be safely migrated. The noop paths aren't real
+    // connector state keys, and the task will backfill when re-enabled anyway.
```
jshearer: I _think_ this is true, and all of the cases I found where this applies have been disabled for over 6 months so it's probably fine, but I still wanted to check..

PR #2866, opened 2026-04-17
Remove retired grant directives

GregorShear:
Invite links are now managed via the graphql API and grant directives are no longer used by the frontend (and they were never used in flowctl)


jshearer (approved):
LGTM

PR #2868, opened 2026-04-17
Customizable alerting

### LINKED ISSUE #2845: RFC: customizable alert configuration
### Context

We've been building out the alert system over the past year: shard failure alerts, data movement stalled alerts, and more recently, abandoned task detection alerts (idle, chronically failing, auto-disabled). We want to keep going. Specifically:

1. Make `ShardFailed` alerts default-on for all users
2. Enable auto-disabling of abandoned tasks with confidence
3. Support bulk and programmatic alert configuration without requiring per-task spec changes

Goals 1 and 2 are blocked in part because a single global threshold doesn't work for all tasks. Tasks that fail frequently but still move data every hour would be classified as "chronically failing", but should not be disabled (though the connectors should be fixed to not do this). The 30-day idle threshold is too short for some tasks that sync monthly. Every time we try to tighten one scenario, we break another.. per-task tuning is needed to make these features safe to enable broadly.

Goal 3 is blocked because there is currently no place for alert configuration to live. The only per-task config today (`DataMovementStalled` interval) is in a side-table (`alert_data_processing`) that flowctl doesn't know about. All other thresholds are global env vars with no per-task or per-prefix override.

The solution: a new prefix-scoped database table for alert configuration, analogous to storage mappings. Alert config is only ever read by the control plane (not the data plane or reactor), so it doesn't need to be included in the task spec. A database table also allows bulk configuration by prefix and instant changes without publication.

### Proposed solution

Create a new `alert_configs` table that associates alert configuration with catalog prefixes or exact task names:

```sql
CREATE TABLE public.alert_configs (
    id                      public.flowid NOT NULL DEFAULT internal.id_generator(),
    catalog_prefix_or_name  text NOT NULL,
    config                  jsonb NOT NULL,
    detail                  text,
    created_at              timestamptz NOT NULL DEFAULT now(),
    updated_at              timestamptz NOT NULL DEFAULT now(),
    last_modified_by        uuid REFERENCES auth.users(id),
    UNIQUE (catalog_prefix_or_name)
);
```

A config row applies to all tasks whose catalog name matches:

```json
{
  "dataMovementStalled": { "threshold": "2h" },
  "shardFailed": { "failureThreshold": 3, "retentionWindow": "8h" },
  "taskChronicallyFailing": { "threshold": "30d" },
  "taskIdle": { "threshold": "60d" },
  "autoDisable": { "idle": true, "failing": false }
}
```

Example: a config at `acmeCo/prod/` applies to all tasks under that prefix. A config at `acmeCo/prod/source-postgres` (exact name, no trailing `/`) applies to only that task. An exact name match takes priority over a prefix match.

### Alert type categorization

**Configurable (in the `alert_configs` table):**

| Alert type | Parameter | Current source | Current default | Meaning |
| --- | --- | --- | --- | --- |
| `dataMovementStalled` | `threshold` | `alert_data_processing` table | 2h | How long without data movement before alerting |
| `shardFailed` | `failureThreshold` | `ALERT_AFTER_SHARD_FAILURES` env var | 3 | Number of failures within the retention window required to fire the alert |
| `shardFailed` | `retentionWindow` | `SHARD_FAILURE_RETENTION` env var | 8h | Time window within which failures are counted; older failures are discarded |
| `taskChronicallyFailing` | `threshold` | `CHRONICALLY_FAILING_THRESHOLD` env var | 30d | How long `ShardFailed` must be continuously active before firing |
| `taskIdle` | `threshold` | `IDLE_THRESHOLD` env var | 35d | How long without data movement before firing (bumped from 30d to accommodate monthly syncs) |
| (auto-disable) | `autoDisable.idle` | `DISABLE_IDLE_TASKS` env var | false | Whether idle tasks matching this config are auto-disabled |
| (auto-disable) | `autoDisable.failing` | `DISABLE_FAILING_TASKS` env var | false | Whether chronically failing tasks matching this config are auto-disabled |

Some notes:

* For tasks with unusual cadences (monthly syncs, seasonal data sources), setting a longer `taskIdle.threshold` is the intended mechanism for preventing false alerts and unwanted auto-disable.
* Auto-disable grace periods (`CHRONICALLY_FAILING_DISABLE_AFTER`, `IDLE_DISABLE_AFTER`, both default 7d) remain global. The threshold controls when an alert fires; the grace period is a system-level policy about how much notice to give before disabling, and doesn't need to vary per-prefix.
* `resolve_shard_failed_alert_after` (how long of healthy shard status before resolving a `ShardFailed` alert, default 2h) remains a global setting.
* `user_pub_threshold` (how recently a user must have published to suppress abandoned-task alerts, default 14d) remains a global setting. It represents a system-level definition of "active ownership" rather than a user-facing knob.

**Not configurable (remain global or tenant-level):**

- `freeTrial`, `freeTrialEnding`, `freeTrialStalled`, `missingPaymentMethod`: billing/tenant-level, not task-level
- `autoDiscoverFailed`, `backgroundPublicationFailed`: binary (failed or didn't), no meaningful threshold to configure
- `taskAutoDisabledFailing`, `taskAutoDisabledIdle`: these fire as consequences of the chronically-failing and idle alert chains, not independently configurable

#### Prefix resolution

Longest-prefix-match with exact-name override. For a task `acmeCo/prod/source-postgres`:

1. An exact-name config `acmeCo/prod/source-postgres` wins if it exists
2. Otherwise, the longest matching prefix wins (e.g., `acmeCo/prod/` over `acmeCo/`)
3. For fields not specified in the matching config, fall back to `ControllerConfig` env var defaults

#### Per-prefix-or-exact-name matching (discussion topic)

The `catalog_prefix_or_name` column accepts both prefixes (ending in `/`) and exact catalog names (no trailing `/`). A prefix applies to all matching tasks; an exact name applies to one task. This differs from `storage_mappings`, which uses the `catalog_prefix` domain type and only supports prefixes.

**This is a design decision open for discussion**. The alternative is prefix-only, which would mean per-task config requires a naming convention that puts the task in its own sub-prefix. Most customers manage tasks as cattle, but exact-name matching provides an escape hatch for the cases that need it.

#### Auto-disable

Currently `DISABLE_IDLE_TASKS` and `DISABLE_FAILING_TASKS` are global env var booleans. Moving them into `alert_configs` makes auto-disable prefix-scoped. A customer (or the support team) can enable auto-disable for `acmeCo/prod/` but leave it off for `acmeCo/dev/`. This addresses a real use case where a customer's shard failure alert subscriptions had to be restricted to prod because dev pipelines were noisy.

#### DataMovementStalled: move evaluation into the controller

The `DataMovementStalled` evaluation currently runs through a DB view (`alert_data_movement_stalled`) polled by an `AlertEvaluator` automation. This is a legacy pattern: the view joins `alert_data_processing.evaluation_interval` with `catalog_stats_hourly` to determine which tasks have stalled, and the evaluator diffs the results against `alert_history` to fire/resolve alerts.

Moving `DataMovementStalled` evaluation into the controller eliminates this indirection. The controller already evaluates a similar condition for `TaskIdle` via `fetch_last_data_movement_ts()` in `abandon.rs`. Adding `DataMovementStalled` evaluation is the same pattern: check `catalog_stats_hourly` for recent byte movement, compare against the threshold from `alert_configs`, fire or resolve the alert.

This replaces the `alert_data_movement_stalled` DB view and the `AlertEvaluator<DataMovementStalledAlerts>` automation. The `AlertEvaluator<TenantAlerts>` automation remains for now to limit scope; tenant-level billing alerts could move to controllers in the future.

A comment in `evaluator.rs` hints at this direction:

https://github.com/estuary/flow/blob/61635c47300f46d987560d5b60309a7addab5852/crates/agent/src/alerts/evaluator.rs#L201-L204

#### API and authorization

The table is accessed exclusively through GraphQL, with authorization evaluated in the resolver layer rather than through RLS.

Authorization follows the storage mappings pattern: admin capability on the prefix is required to create, update, or delete an `alert_configs` row. GraphQL queries support `exactPrefixes` and `underPrefix` filters.

#### Audit trail

Deferred. The `last_modified_by` column captures who made each change. Full change history can be derived from CDC events on the table in the future.

### Migration from `alert_data_processing`

The migration avoids a flag-day deploy by using a transitional fallback in the controller. The ordering is important to avoid stale data shadowing fresh writes:

1. Create the `alert_configs` table (empty). Deploy the controller reading from `alert_configs` with fallback to `alert_data_processing` when no matching `alert_configs` row exists. Since the table is empty, everything falls back to `alert_data_processing`. No behavioral change.
2. Deploy UI changes to read/write `alert_configs` via GraphQL instead of `alert_data_processing` via PostgREST. New user changes go to the new table. Old untouched values still resolve via fallback to `alert_data_processing`.
3. Run a data migration to copy remaining `alert_data_processing` rows (those not already overwritten by UI writes in step 2) into `alert_configs`. After this, `alert_configs` has all values.
4. Remove the fallback from the controller and drop `alert_data_processing`.

The UI must switch before the data migration. If the data migration ran first, subsequent UI writes to `alert_data_processing` would be shadowed by stale copies in `alert_configs` (since `alert_configs` takes priority in the fallback logic).

### Alternatives considered

**Alert config in the task spec.** Putting an `alerts` block directly on `CaptureSpec`/`MaterializationSpec`/`DerivationSpec` would make config version-controlled, automatically visible to flowctl, and preserved in publication history. We ended up not going down this path because:

- Changing config would require a full publication which is slow, can fail due to unrelated connector issues, and unneccesarily restarts the connector.
- There's no way to configure by prefix, meaning that every task must be tuned individually
- Alert config is only consumed by the control plane so it doesn't need to live in the spec.

jshearer:
This PR implements Phase 1 of configurable alert conditions (#2845).

## The `alert_configs` table

```sql
CREATE TABLE public.alert_configs (
    id                      public.flowid NOT NULL DEFAULT internal.id_generator(),
    catalog_prefix_or_name  text NOT NULL CHECK (catalog_prefix_or_name IS NFKC NORMALIZED),
    config                  jsonb NOT NULL,
    detail                  text,
    created_at              timestamptz NOT NULL DEFAULT now(),
    updated_at              timestamptz NOT NULL DEFAULT now(),
    last_modified_by        uuid REFERENCES auth.users(id),
    UNIQUE (catalog_prefix_or_name)
);
```

`catalog_prefix_or_name` accepts either a catalog prefix (ends in `/`) or an exact catalog name. The `config` JSONB is typed (`models::AlertConfig`, `deny_unknown_fields`) and carries optional per-alert-type subconfigs.

| Setting | Applies to | Default source |
|---|---|---|
| `shardFailed.failureThreshold` | `ShardFailed` | `ALERT_AFTER_SHARD_FAILURES` (default 3) |
| `shardFailed.retentionWindow` | `ShardFailed` | `SHARD_FAILURE_RETENTION` (default 8h) |
| `dataMovementStalled.threshold` | `DataMovementStalled` | `alert_data_processing.evaluation_interval` (fallback) |
| `taskChronicallyFailing.threshold` | `TaskChronicallyFailing` | `CHRONICALLY_FAILING_THRESHOLD` (default 30d) |
| `taskChronicallyFailing.autoDisable` | auto-disable after chronic-failing grace period | `false` |
| `taskIdle.threshold` | `TaskIdle` | `ABANDON_IDLE_THRESHOLD` (default 30d) |
| `taskIdle.autoDisable` | auto-disable after idle grace period | `false` |
| `{shardFailed, dataMovementStalled, taskIdle, taskChronicallyFailing}.enabled` | opt-in / opt-out lever | see below |

Grace periods (`CHRONICALLY_FAILING_DISABLE_AFTER`, `ABANDON_IDLE_DISABLE_AFTER`) and system-level knobs (`ABANDON_USER_PUB_RECENCY`, `RESOLVE_SHARD_FAILED_ALERT_AFTER`) remain global policies.

## Nested `alert_configs` are additive

Every covering `alert_configs` row contributes to the effective config for a task, in top-down order. This is to allow setting e.g tenant-wide defaults, while still allowing task-specific overrides. For example:

**Fleet-wide ShardFailed threshold tuning with per-task silencing:**
```yaml
acmeCo/                 : { shardFailed: { failureThreshold: 10 } }
acmeCo/prod/            : { shardFailed: { failureThreshold: 5 } }
acmeCo/prod/noisy-task  : { shardFailed: { enabled: false } }
```
- `acmeCo/any-task` fires at 10 failures (inherits threshold 10 from `acmeCo/`).
- `acmeCo/prod/task-a` fires at 5 failures (threshold overridden by `acmeCo/prod/`).
- `acmeCo/prod/noisy-task` silenced (merged: `{ enabled: false, failureThreshold: 5 }`).

**Dev-prefix idle silencing:**
```yaml
acmeCo/dev/ : { taskIdle: { enabled: false } }
```
Tasks under `acmeCo/dev/` don't fire `TaskIdle`. Tasks elsewhere continue firing on the 30d env default (baseline).

**Re-enabling a specific task under a silenced prefix:**
```yaml
acmeCo/dev/              : { taskIdle: { enabled: false } }
acmeCo/dev/critical-task : { taskIdle: { enabled: true } }
```
`acmeCo/dev/critical-task` fires (merged: `{ enabled: true }`); siblings stay silenced.

## Opt-in / opt-out via `enabled`

Each configurable alert has a fixed **baseline**. `enabled` is the sole opt-in / opt-out lever:

- Merged `enabled: Some(true)` → alert may fire, subject to its own conditions.
- Merged `enabled: Some(false)` → alert is silenced.
- Merged `enabled: None` → baseline applies.

| Alert | Baseline | Conditions (when enabled) |
|---|---|---|
| `ShardFailed` | firing | fires after `failureThreshold` failures within `retentionWindow` |
| `DataMovementStalled` | silent (requires a threshold) | fires when no byte movement for `threshold` |
| `TaskChronicallyFailing` | firing (cascades off `ShardFailed`) | fires when `ShardFailed` has been continuously active for `threshold` |
| `TaskIdle` | firing | fires when no data movement for `threshold` |
| `AutoDiscoverFailed`, `BackgroundPublicationFailed` | firing (binary; not configurable) | — |

**Note:** `ShardFailed`'s behavior is not changing from the way it works today. Presently, `ShardFailed` alerts are enabled universally (which is why everyone can see them in the "Alerts" tab in the UI), but we haven't added it to the list of alert subscriptions you get by default, so the actual shard failed _emails_ are (and will remain) opt-in.

## `DataMovementStalled` moves into the controller

`DataMovementStalled` evaluation now runs per-task inside the controller pipeline, alongside `TaskIdle` and `TaskChronicallyFailing`, reading `catalog_stats_hourly` directly to check recent byte movement. 

## API

`alert_configs` is accessed exclusively through GraphQL.

- **Query**: `alertConfigs(filter, after, first)` returns rows the caller has `Read` capability on, sorted by `catalog_prefix_or_name`. Results are automatically bounded by the caller's authorized prefixes (same pattern as `inviteLinks`); `filter.catalogPrefixOrName.startsWith` narrows further, and a full catalog name as the `startsWith` value targets a single row.
- **Mutation**: `updateAlertConfig(catalogPrefixOrName, config, detail)` requires `Admin` on the governing prefix (the row itself if it's a prefix, else the parent prefix). Returns a `created: Boolean` indicating insert vs update.

## `flowctl` 

`alert_configs` is exposed through a new `flowctl alerts` subcommand, with `subscriptions` and `configs` as sub-namespaces. The pre-existing top-level `flowctl alert-subscriptions` is moved under `flowctl alerts subscriptions` and retained as a hidden alias so existing scripts continue to work.

**`flowctl alerts configs list [--prefix <p>]`** paginates through the `alertConfigs` connection. `--prefix` maps to `filter.catalogPrefixOrName.startsWith`; a full catalog name targets a single row. Output respects the global `--output {table,json,yaml}` selection.

**`flowctl alerts configs update --prefix <prefix-or-name>`** has two mutually-exclusive modes:

- `--set <path>=<value>` and `--unset <path>` (both repeatable) for read-modify-write patches. If no row exists for `--prefix`, patches are applied to `{}` and a new row is created.
- `--config <file|->` for whole-row replace from YAML or JSON.

The merged JSON is deserialized through `models::AlertConfig` to surfac typos client-side rather than at the GraphQL boundary.

```
flowctl alerts configs update --prefix acmeCo/                  --set shardFailed.failureThreshold=10
flowctl alerts configs update --prefix acmeCo/dev/              --set taskIdle.enabled=false
flowctl alerts configs update --prefix acmeCo/dev/critical-task --set taskIdle.enabled=true
flowctl alerts configs update --prefix acmeCo/                  --unset taskIdle
```

## Migrations

`DataMovementStalled` is handled in four steps to avoid a flag-day:

1. **(this PR)** Deploy `alert_configs` (empty) and the controller reading from it, with fallback to `alert_data_processing` when no matching `alert_configs` row supplies a `dataMovementStalled.threshold`.
2. UI switches its PostgREST writes on `alert_data_processing` to GraphQL `updateAlertConfig` calls. New configuration flows into `alert_configs`; old untouched values still resolve via fallback.
3. A data migration copies remaining `alert_data_processing` rows into `alert_configs`. The UI switch must precede this so fresh UI writes aren't shadowed by stale copies.
4. Remove the controller fallback and drop `alert_data_processing`.

This PR covers step 1 only. `alert_data_processing` is unchanged and continues to accept UI writes.

`ShardFailed` keeps its default-firing baseline, so no migration is required. Existing `alert_subscriptions` rows that include `shard_failed` continue to deliver notifications exactly as before. Tenants who want to silence `ShardFailed` for a prefix or task can write an `alert_configs` row with `enabled: false`; tenants who want a non-default `failureThreshold` / `retentionWindow` can set those fields directly.


crates/control-plane-api/src/server/public/graphql/alert_configs.rs
```diff
@@ -0,0 +1,354 @@
...
+pub struct AlertConfigsFilter {
+    /// Filter on the `catalog_prefix_or_name` column.
+    pub catalog_prefix_or_name: Option<filters::PrefixFilter>,
+}
+
+/// A single row from `public.alert_configs`.
+#[derive(Debug, Clone, async_graphql::SimpleObject)]
+pub struct AlertConfigEntry {
+    pub id: models::Id,
+    pub catalog_prefix_or_name: String,
```
GregorShear, line 30: i wish we could just call this `scope`. 

But i also don't love how much weight the trailing slash carries - both `acmeCo/nested` and `acmeCo/nested/` are valid and they could have massively different outcomes.

GregorShear:
A query like this could be pretty useful:

```gql
resolvedAlertConfig(catalogPrefixOrName: String!): ResolvedAlertConfig!

type ResolvedAlertConfig {
  config: AlertConfig!      # the merged config applied at this scope
  provenance: JSON!         # dotted path → source prefix, e.g.
                            # { "shardFailed.enabled": "acmeCo/prod/noisy-task",
                            #   "shardFailed.failureThreshold": "acmeCo/prod/",
                            #   "taskIdle.threshold": "<estuary default>" }
} 
```

GregorShear:
```yaml
acmeCo/prod/noisy-task  : { shardFailed: { enabled: false } }
```

any thoughts on how we handle currently-firing alerts when this gets set?

GregorShear:
```yaml
shardFailed.failureThreshold: 3
taskIdle.threshold: 60d
```

maybe we can be even more descriptive, like 
```yaml
shardFailed.condition: { failures: 3, per: '8h' }
taskIdle.condition: { idleFor: '60d' }
```

kinda spitballing here... happy to chat about it

jshearer:
kk, @GregorShear what do you think? I added `effectiveAlertConfig` to show the resolved config, and `provenance` to show where each setting comes from. I also renamed the settings to be more user-facing.

### Condition nesting

Each alert's condition parameters now live under a `condition` sub-object. `enabled` and `autoDisable` stay at the top level:

```json
{
  "shardFailed": {
    "enabled": true,
    "condition": { "failures": 5, "per": "4h" }
  },
  "taskIdle": {
    "autoDisable": true,
    "condition": { "idleFor": "14days" }
  }
}
```

The condition field names are more descriptive than the originals:

| Before                             | After                                         |
| ---------------------------------- | --------------------------------------------- |
| `shardFailed.failureThreshold`     | `shardFailed.condition.failures`              |
| `shardFailed.retentionWindow`      | `shardFailed.condition.per`                   |
| `dataMovementStalled.threshold`    | `dataMovementStalled.condition.stalledFor`    |
| `taskChronicallyFailing.threshold` | `taskChronicallyFailing.condition.failingFor` |
| `taskIdle.threshold`               | `taskIdle.condition.idleFor`                  |

Top-level keys (`shardFailed`, `dataMovementStalled`, `taskChronicallyFailing`, `taskIdle`) are unchanged since they match the `AlertType` enum. Deep merge still works correctly since `condition` is an object that merges by key across layers.

### `effectiveAlertConfig` with provenance

Instead of a new root-level query, the resolved config is nested on existing types:

```graphql
type LiveSpec {
  effectiveAlertConfig: EffectiveAlertConfig!
}

type AlertConfigEntry {
  effective: EffectiveAlertConfig!
}

type EffectiveAlertConfig {
  config: AlertConfig!
  provenance: [FieldProvenance!]!
}

type FieldProvenance {
  path: String!
  source: String   # null = controller default
}
```

This lets you query the fully-merged effective config for a task along with which prefix/name contributed each setting.

GregorShear:
regarding `"condition": { "failures": 5, "per": "4h" }`:

"failures...per" makes sense if they're always defined together, but i assume we can set `"condition": { "per": "4h" }` and inherit `failures` from an ancestor. I think _either_ requiring the entire condition be defined as a whole, or renaming "per" to "window" or something that makes more sense in isolation would be slightly better, but not a blocker here

GregorShear (approved):

crates/control-plane-api/src/server/public/graphql/alert_configs.rs
jshearer: I ended up adding a check to prevent you from creating an `alert_config` for a task name (no trailing slash) that doesn't exist... it's better than nothing at least.

jshearer:
WRT failures per... I _think_ this is mostly a UX issue. We have `effectiveAlertConfig` so we can show the full alert config when editing.

Issue #2877, opened 2026-04-22
PostgREST surface area / GraphQL Migration


jshearer:
### 1. Catalog Lifecycle (drafts, publications, discovers, evolutions)

This is the largest and most complex domain. It covers the entire user workflow of creating, editing, validating, and publishing catalog specifications.

**PostgREST endpoints in use:**

| Endpoint | Type | UI | flowctl | Operations | GraphQL? |
|---|---|---|---|---|---|
| `drafts` | table | SELECT, INSERT, DELETE | SELECT, INSERT, DELETE | Full CRUD for change-sets | None |
| `drafts_ext` | view | SELECT | SELECT | List drafts with computed fields | None |
| `draft_specs` | table | SELECT, INSERT, UPDATE, DELETE, UPSERT | SELECT, INSERT, DELETE | Core spec editing | None |
| `draft_specs_ext` | view | SELECT | SELECT | Extended draft specs with authz | None |
| `draft_errors` | view | SELECT | SELECT | Validation error display | None |
| `publications` | table | INSERT, SELECT (polling) | INSERT | Trigger publish, poll for completion | None |
| `publication_specs_ext` | view | SELECT | SELECT | Publication history with user info | **Partial**: flowctl `catalog history` uses GraphQL `PublicationHistoryQuery`; UI still uses PostgREST |
| `discovers` | table | INSERT, SELECT (polling) | INSERT, SELECT | Auto-discovery of connector resources | None |
| `view_logs` | RPC | yes | yes | Stream job logs during publish/discover | None |

**Existing GraphQL coverage:** flowctl's `catalog history` command uses a `PublicationHistoryQuery` via GraphQL, which queries publication history for a specific live spec. The UI does not use this yet.

**RFC for this work** lives at #2782

**Migration notes:** This is the hardest domain to migrate because:
- Publications involve complex server-side orchestration (build, validate, authorize, commit)
- The proposed approach is a `createBuild` / `activateBuild` mutation pair
- Snapshot-based authorization (#2781) is a prerequisite for acceptable publication performance through GraphQL
- Need to figure out what to do about drafts, as well as migrating the UI; flowctl ought to be fairly straightforward
- `view_logs` is a polling-based log streaming mechanism that will need a GraphQL equivalent. Maybe polling, maybe SSE

**Estimated size:** XL. This is probably 40-50% of the total migration work.

---

### 2. Live Specs

The "read" side of the catalog: querying what's currently deployed.

[...]

PR #2883, opened 2026-04-23
Billing GraphQL

### LINKED ISSUE #2879: Migrate billing edge functions to GraphQL
## Context

The control plane currently exposes edge functions for Stripe setup intents, payment-method management, invoice listing, and customer/payment-method display. We want to expose those APIs through the GraphQL

## Current state

The edge function handles five operations: 
* `setup-intent` creates a Stripe SetupIntent (creating the Stripe customer first if needed)
* `get-tenant-payment-methods` lists attached payment methods plus the primary,
* `set-tenant-primary-payment-method` updates the default
* `delete-tenant-payment-method` detaches a method and auto-promotes another as primary if the deleted one was the default
* `get-tenant-invoice` fetches an invoice by date range and type (`Usage` or `Manual`). 

### Proposed GraphQL Schema

A top-level `tenant` query returns a `Tenant` type. Billing is a field on `Tenant` rather than a standalone top-level query, because tenant is a natural aggregation root and other facets (membership, usage, settings etc) will follow.

```graphql
type Query {
  tenant(name: String!): Tenant
}

type Tenant {
  name: String!
  billing: TenantBilling
}

type TenantBilling {
  paymentMethods: [PaymentMethod!]!
  primaryPaymentMethodId: String
  invoices(filter: InvoiceFilter, before: String, after: String, first: Int, last: Int): PaginatedInvoices!
}

input InvoiceFilter {
  dateStart: StringFilter
  dateEnd: StringFilter
  invoiceType: InvoiceTypeFilter
}
...
```

The schema above covers the fields the UI reads today. Additional fields from either the database view or the Stripe response should be added during implementation if the UI consumes them.

### Invoice resolution

The `Invoice` type merges two data sources. The database fields (`dateStart`, `dateEnd`, `invoiceType`, `subtotal`, `lineItems`, `extra`) come from the `invoices_ext` view, which the billing history table uses. The Stripe fields (`amountDue`, `status`, `invoicePdf`, `hostedInvoiceUrl`) come from the Stripe Invoices API, matched by `estuary.dev/period_start`, `estuary.dev/period_end`, and `estuary.dev/invoice_type` metadata on the Stripe invoice.

GraphQL field selection makes this efficient. The billing history table selects only the database fields and never hits Stripe. When a user drills into a specific invoice to get the PDF or payment status, the detail view selects the Stripe fields, and the resolver fetches from Stripe for that single row. The Stripe fields are nullable because a database invoice record may not yet have a corresponding Stripe invoice (e.g. the current billing period).

### Mutations

Three mutations cover the write operations the UI needs.

* `createBillingSetupIntent(tenant: String!): CreateBillingSetupIntentPayload` creates a Stripe setup intent and returns the `clientSecret` the browser needs to continue the Elements flow. If no Stripe customer exists for the tenant yet, this mutation creates one.
* `setBillingPaymentMethod(tenant: String!, paymentMethodId: String!): SetBillingPaymentMethodPayload` sets the tenant's default payment method for invoicing by updating `invoice_settings.default_payment_method` on the Stripe customer, and returns the updated payment method list so the UI can refresh without a separate query.
* `deleteBillingPaymentMethod(tenant: String!, paymentMethodId: String!): DeleteBillingPaymentMethodPayload` detaches a payment method from the tenant's Stripe customer. If the deleted method was the primary, another attached method is auto-promoted as the new default.

### Authorization

Billing resolvers keep the existing authorization behavior, which requires the caller to have Admin to the tenant. In the future as we build out more specific capabilities, it will shift to a Billing-specific role.

jshearer:
This PR moves the billing endpoints from edge functions into GraphQL.

## `Tenant.billing`

Billing is a field on `Tenant` rather than a standalone top-level query, because tenant is the natural aggregation root and other fields will follow.

```graphql
type Query {
  tenant(name: String!): Tenant
}

type Tenant {
  name: String!
  billing: TenantBilling!
}

type TenantBilling {
  paymentMethods: [PaymentMethod!]!
  primaryPaymentMethod: PaymentMethod
  invoices(
    filter: InvoiceFilter, 
    before: String, 
    after: String, 
    first: Int, 
    last: Int
  ): InvoiceConnection!
}
```

This is a design choice I went with as it felt like we were starting to pollute the global namespace with queries that ought to be scoped under their natural owner. It's possible that we should move some other queries as well, but that's out of scope. 

WRT permissions: `Tenant.billing` is non-nullable which means that a request that selects it without `Admin` on the tenant gets an error, rather than a partial response containing `null` and a path-scoped entry in `errors`.

## Invoices

`Invoice` merges two data sources. Database fields (`dateStart`, `dateEnd`, `invoiceType`, `subtotal`, `lineItems`, `extra`) come from the `invoices_ext` Postgres view. Stripe fields (`amountDue`, `status`, `invoicePdf`, `hostedInvoiceUrl`) are resolved directly from Stripe.

## Mutations

```graphql
type MutationRoot {
  createBillingSetupIntent(tenant: String!): CreateBillingSetupIntentPayload
  setBillingPaymentMethod(tenant: String!, paymentMethodId: String!): SetBillingPaymentMethodPayload
  deleteBillingPaymentMethod(tenant: String!, paymentMethodId: String!): DeleteBillingPaymentMethodPayload
}
```

- `createBillingSetupIntent` finds or creates a Stripe customer for the tenant, then returns the `clientSecret` the browser needs for the Stripe Elements flow. 
- `setBillingPaymentMethod` updates `invoice_settings.default_payment_method` on the customer.
- `deleteBillingPaymentMethod` detaches the method; if the deleted method was the primary, the first remaining method is auto-promoted as default.

All three require `Admin` on the tenant and validate that the payment method belongs to the tenant's Stripe customer before acting.

## `billing-types`

Shared Stripe types and search-query builders (`InvoiceType`, `InvoiceSearch`, `InvoiceMetadata`, `customer_search_query`, `stripe_search`) are extracted into a `billing-types` crate. `billing-integrations` now imports these instead of defining its own copies.

## CI

`deploy-agent-api.yaml` injects `STRIPE_API_KEY` from Cloud Run secret manager. `platform-test.yaml` conditionally runs the `graphql_billing_live_stripe` integration test when `STRIPE_TESTMODE_API_KEY` is available in the environment.

Closes https://github.com/estuary/flow/issues/2879
Part of #2877


crates/control-plane-api/src/billing/db.rs
```diff
@@ -0,0 +1,167 @@
...
+    pub billed_prefix: String,
+    pub line_items: sqlx::types::Json<serde_json::Value>,
+    pub subtotal: i32,
+    pub extra: sqlx::types::Json<serde_json::Value>,
+    pub invoice_type: InvoiceType,
+}
+
+/// Fetch invoices older than `cursor` (or the newest invoices when `cursor`
+/// is `None`). Returned rows are ordered newest-first.
+pub async fn fetch_invoice_rows_forward(
```
jshearer, line 43: I really tried to not have `fetch_invoice_rows_forward` and `fetch_invoice_rows_backward` since they're almost identical, but the alternative (dynamic query generation) was worse, so I kept it. If anyone has any better ideas I'm all ears

crates/control-plane-api/src/server/public/graphql/billing/invoices.rs
```diff
@@ -0,0 +1,510 @@
...
+
+/// Request-scoped loader that resolves Stripe-backed records through the
+/// shared `BillingProvider`.
+pub(in crate::server::public::graphql) struct StripeDataLoader(pub Arc<dyn BillingProvider>);
+
+impl Loader<StripeInvoiceKey> for StripeDataLoader {
+    type Value = stripe::Invoice;
+    type Error = async_graphql::Error;
+
+    async fn load(
```
jshearer: This could be parallelized.. wanted to KISS where possible tho. Happy to be convinced otherwise

crates/agent/src/main.rs
```diff
@@ -41,6 +42,12 @@ struct Args {
     /// The port to listen on for API requests.
     #[clap(long, default_value = "8080", env = "API_PORT")]
     api_port: u16,
+    /// Stripe secret API key. When provided, the billing GraphQL queries and
```
jshearer, line 44: This is `Option` so that a Stripe API key is not required to run locally.

crates/control-plane-api/src/server/mod.rs
```diff
@@ -34,6 +34,7 @@ pub enum Rejection {
 /// App is the wired application state of the control-plane API.
 pub struct App {
     pub _id_generator: std::sync::Mutex<models::IdGenerator>,
+    pub billing_provider: Option<Arc<dyn crate::billing::BillingProvider>>,
```
jshearer, line 37: I really wanted this to be generic, but it would have exploded the complexity of this PR even further and it wasn't worth it

crates/agent/src/integration_tests/harness.rs
```diff
@@ -538,45 +538,14 @@ impl TestHarness {
...
-        )
-        .await
-        .expect("failed to provision tenant");
-
-        // Remove the estuary_support/ role grant, which gets automatically
-        // added by a trigger whenever we create a new tenant. Removing it here
-        // ensures that things still work correctly without it.
-        sqlx::query!(r#"delete from role_grants where subject_role = 'estuary_support/';"#)
-            .execute(&mut *txn)
+        control_plane_api::test_support::provision_test_tenant(&self.pool, tenant, &email, meta)
```
jshearer: Moved this to `control_plane_api` so it could be shared, rather than duplicating it.

crates/control-plane-api/src/server/public/graphql/billing/tenant.rs
```diff
@@ -0,0 +1,173 @@
...
+            .await
+            .map_err(|err| async_graphql::Error::new(err.to_string()))?;
+        Ok(methods.iter().map(PaymentMethod::from).collect())
+    }
+
+    async fn primary_payment_method(&self) -> Result<Option<PaymentMethod>> {
+        let Some(customer) = self
+            .provider
+            .as_ref()
+            .find_customer(&self.tenant)
```
GregorShear: `find_customer()` is called both here and in `payment_methods` resolver - could we deduplicate/cache?

crates/agent/src/main.rs
```diff
@@ -17,6 +17,7 @@ use std::sync::Arc;
 #[derive(Derivative, Parser)]
 #[derivative(Debug)]
 #[clap(author, version, about, long_about = None)]
+#[clap()]
```
GregorShear: omit?

crates/control-plane-api/src/billing/db.rs
GregorShear: i struggled with this a while ago when working on invite links and i think this duplication is just the price of admission for the query cache. At least it's easy to reason about...

mise/tasks/local/control-plane
```diff
@@ -28,6 +28,16 @@ SSL_CERT_FILE=${FLOW_LOCAL}/ca.crt
...
 EOF
 
+# Billing backend: forward STRIPE_API_KEY from the invoking shell when set
+# (e.g. a testmode key pasted before `mise run`), otherwise fall back to the
+# in-memory stub. The agent's clap ArgGroup requires exactly one of the two,
+# so never write both.
+if [ -n "${STRIPE_API_KEY:-}" ]; then
+    echo "STRIPE_API_KEY=${STRIPE_API_KEY}" >> "${FLOW_LOCAL}/env/agent.env"
+else
+    echo "BILLING_IN_MEMORY=true" >> "${FLOW_LOCAL}/env/agent.env"
```
GregorShear: i can't find anything that reads `BILLING_IN_MEMORY` - could we remove it?

crates/control-plane-api/src/server/public/graphql/billing/invoices.rs
```diff
@@ -0,0 +1,599 @@
...
+                invoice_type: Some(key.invoice_type),
+                period_start: Some(&key.date_start),
+                period_end: Some(&key.date_end),
+                status: billing_types::StatusFilter::Exclude(stripe::InvoiceStatus::Draft),
+            }
+            .to_query();
+            let fetched = self
+                .0
+                .search_invoices(&query)
+                .await
```
GregorShear: if i'm reading this correctly, we're waiting for the current `search_invoices` call to respond before we make the next one - could we fire these off in parallel, or do we risk rate limiting?

crates/control-plane-api/src/server/public/graphql/billing/mutations.rs
```diff
@@ -0,0 +1,358 @@
...
+        .list_payment_methods(customer_id)
+        .await
+        .map_err(|err| async_graphql::Error::new(err.to_string()))?;
+
+    if methods
+        .iter()
+        .all(|method| method.id.as_str() != payment_method_id)
+    {
+        return Err(async_graphql::Error::new(
+            "payment method is not attached to the tenant's Stripe customer",
```
GregorShear: Might want to be a little more coy about this - could we just say "not found"?

crates/control-plane-api/src/server/public/graphql/billing/mutations.rs
```diff
@@ -0,0 +1,358 @@
...
+    }
+}
+
+#[derive(Debug, Clone, SimpleObject)]
+pub struct CreateBillingSetupIntentPayload {
+    client_secret: String,
+}
+
+#[derive(Debug, Clone, SimpleObject)]
+pub struct UpdateBillingPaymentMethodPayload {
```
GregorShear: since it's used for both delete and update operations:

```suggestion
pub struct BillingPaymentMethodPayload {
```

crates/billing-integrations/src/publish.rs
```diff
@@ -167,25 +155,18 @@ impl Invoice {
...
-            .as_str()
-            .expect("InvoiceType is serializable");
-
         let invoice_search = stripe_search::<stripe::Invoice>(
             &client,
             "invoices",
             SearchParams {
-                query: format!(
-                    r#"
-                    -status:"deleted" AND
```
GregorShear, line 182: this hardcoded param to omit deleted invoices is missing from the `InvoiceSearch` - does this change the behavior in a way that we want?

crates/billing-types/src/lib.rs
```diff
@@ -0,0 +1,249 @@
...
+
+/// Status clause to append to a Stripe invoice search query.
+///
+/// Stripe's search DSL accepts both positive (`status:"open"`) and negative
+/// (`-status:"draft"`) filters
+#[derive(Debug, Clone, Copy, Default)]
+pub enum StatusFilter {
+    /// No status clause.
+    #[default]
+    Any,
```
GregorShear, line 48: related to my comment in publish.rs about omitting deleted invoices - should we omit them by default here? i might be surprised to see deleted invoices in the results unless i explicitly asked for them

crates/billing-integrations/src/publish.rs
```diff
@@ -791,18 +769,18 @@ async fn get_or_create_customer_for_tenant(
...
-            query: format!("metadata[\"{TENANT_METADATA_KEY}\"]:\"{tenant}\""),
+            query: customer_search_query(&tenant),
             ..Default::default()
         },
     )
     .await
     .context(format!("Searching for tenant {tenant}"))?;
 
-    let customer = if let Some(customer) = customers.into_iter().next() {
+    let customer = if let Some(customer) = customers.drain(..).next() {
```
GregorShear: claude says:
>customers.into_iter().next() reads more naturally (`drain` was probably carried over from borrowing constraints that no longer apply).

.github/workflows/platform-test.yaml
```diff
@@ -125,6 +125,13 @@ jobs:
       - run: mise run build:flowctl-go
       - run: mise run ci:nextest-build
       - run: mise run ci:nextest-run
+
+      - name: Stripe integration test
+        if: env.STRIPE_API_KEY != ''
+        env:
+          STRIPE_API_KEY: ${{ secrets.STRIPE_TESTMODE_API_KEY }}
+        run: cargo nextest run --frozen --run-ignored ignored-only -E 'test(graphql_billing_live_stripe)'
```
GregorShear: does this run? claude says
>  The if: evaluates before the step's env: block takes effect, so env.STRIPE_API_KEY is always empty at  evaluation time.

crates/billing-integrations/src/publish.rs
jshearer: lol yeah I have no idea what I was thinking when I originally wrote that... AFAICT, `deleted` does not (or does no longer??) exist. So... no behavior change

crates/billing-types/src/lib.rs
jshearer: same as comment above, `deleted` is not a state that you can fetch invoices in

crates/billing-integrations/src/publish.rs
jshearer: fixed

crates/control-plane-api/src/server/public/graphql/billing/invoices.rs
jshearer: Stripe returns helpful back-off headers and `async-stripe` respects them so I parallelized all of the stripe dataloaders

crates/control-plane-api/src/server/public/graphql/billing/tenant.rs
jshearer: turned into stripe customer dataloader 👍

.github/workflows/platform-test.yaml
jshearer: It passes:

https://github.com/estuary/flow/actions/runs/25458269305/job/74693445092#step:22:1

crates/control-plane-api/src/server/public/graphql/billing/tenant.rs
```diff
@@ -0,0 +1,187 @@
...
+            if let Some(customer) = self
+                .0
+                .find_customer(tenant)
+                .await
+                .map_err(|err| async_graphql::Error::new(err.to_string()))?
+            {
+                out.insert(tenant.clone(), customer);
+            }
+        }
+        Ok(out)
```
GregorShear: same parallelization treatment instead of sequential? maybe for the sake of consistency with the other loaders more than a real performance concern

crates/control-plane-api/src/server/public/graphql/billing/invoices.rs
```diff
@@ -0,0 +1,628 @@
...
+            .await?
+            .and_then(|inv| inv.amount_due))
+    }
+
+    async fn status(&self, ctx: &Context<'_>) -> Result<Option<String>> {
+        Ok(self.stripe_invoice(ctx).await?.and_then(|inv| {
+            inv.status
+                .as_ref()
+                .and_then(|s| serde_json::to_value(s).ok())
+                .and_then(|v| v.as_str().map(str::to_string))
```
GregorShear: InvoiceStatus has an `as_str()` func

```suggestion
            inv.status.as_ref().map(|s| s.as_str().to_string())
```

crates/billing-types/src/lib.rs
```diff
@@ -0,0 +1,272 @@
...
+    }
+
+    pub fn from_str(s: &str) -> Option<Self> {
+        match s {
+            "final" => Some(InvoiceType::Final),
+            "preview" => Some(InvoiceType::Preview),
+            "manual" => Some(InvoiceType::Manual),
+            _ => None,
+        }
+    }
```
GregorShear: agent says:
>this shadows std::str::FromStr::from_str, so "final".parse::<InvoiceType>() won't work and the method signature is confusingly close to the trait's. Could either impl FromStr for InvoiceType (which gives you .parse() for free) or rename the inherent method to something like parse_str to avoid the collision.

crates/control-plane-api/src/billing/db.rs
GregorShear, line 43: also, do you mind mentioning in your comment that this is forward _pagination_? there's another gql api where pagination and ordering of the result set is conflated (and wrong) and i want to make sure we give the agent a chance to get it right if it follows this example

crates/control-plane-api/src/billing/db.rs
```diff
@@ -0,0 +1,167 @@
...
+    // past this batch; truncate it before returning.
+    let has_more = limit.is_some_and(|l| invoices.len() > l);
+    if let Some(l) = limit {
+        invoices.truncate(l);
+    }
+    Ok((invoices, has_more))
+}
+
+/// Fetch invoices newer than `cursor`. Returned rows are ordered newest-first.
+pub async fn fetch_invoice_rows_backward(
```
GregorShear, line 108: same here - add comment this is backward _pagination_

crates/control-plane-api/src/server/public/graphql/billing/tenant.rs
```diff
@@ -0,0 +1,187 @@
...
+
+    async fn primary_payment_method(&self, ctx: &Context<'_>) -> Result<Option<PaymentMethod>> {
+        let loader = ctx.data::<DataLoader<CustomerDataLoader>>()?;
+        let Some(customer) = loader.load_one(self.tenant.clone()).await? else {
+            return Ok(None);
+        };
+        let Some(primary_id) = billing::default_payment_method_id(&customer) else {
+            return Ok(None);
+        };
+        let pm = self
```
GregorShear, line 61: agent says:
>If the client selects both paymentMethods and primaryPaymentMethod in the same query (which the UI mocks suggest is the typical shape), this fetches primary_id via get_payment_method even though that PM is already in the list returned by list_payment_methods for the same customer. Worth either:
>- Looking the primary up in the cached list, or
>- Adding a PaymentMethodLoader keyed on (customer_id, payment_method_id) with a default method that batches via list_payment_methods.
>
>The simpler "find in the list" approach would mirror what set_billing_payment_method already does in the mutations file. Not a hot path so it's fine to defer, but flagging.

crates/control-plane-api/src/server/public/graphql/billing/invoices.rs
```diff
@@ -0,0 +1,628 @@
...
+            subtotal: row.subtotal,
+            line_items: async_graphql::Json(row.line_items.0),
+            extra: async_graphql::Json(row.extra.0),
+            tenant: row.billed_prefix,
+        }
+    }
+
+    async fn stripe_invoice(&self, ctx: &Context<'_>) -> Result<Option<stripe::Invoice>> {
+        let customer_loader = ctx.data::<DataLoader<super::tenant::CustomerDataLoader>>()?;
+        let Some(customer) = customer_loader.load_one(self.tenant.clone()).await? else {
```
GregorShear, line 195: agent says:
> Worth confirming the cache hits when both paymentDetails and TenantBilling.payment_methods are selected in the same query — the latter passes self.tenant from validate_tenant_name (a Prefix-validated string), and this one passes row.billed_prefix from the DB. If the two ever disagree on trailing-slash or case normalization, the per-request DataLoader silently issues two find_customer calls instead of one.

GregorShear (commented):
sorry to leave a couple of raw agent comments - i'm rushing out the door and wanted to get this up before i leave

GregorShear (approved):

crates/control-plane-api/src/server/public/graphql/billing/tenant.rs
jshearer: I'd rather keep these codepaths simpler than try to prematurely optimize this, I think I'll defer this one

crates/control-plane-api/src/server/public/graphql/billing/invoices.rs
jshearer: They should be the same, but the following isn't a big deal anyway:

> he per-request DataLoader silently issues two find_customer calls instead of one.

PR #2886, opened 2026-04-23
Add `detail` column to the `connectors` GQL endpoint

travjenkins:
**Description:**

Adding `detail` to the `connectors` query. It also works for the singular `connector` but that dashboard does not utilize that currently.

**Workflow steps:**

N/A

N/A

**Notes for reviewers:**

How this is consumed in the dashboard

image


jshearer (commented):
One question: are `shortDescription` and `longDescription` _and_ `detail` all load-bearing? I.e does the UI actually need all 3 for various things? Otherwise it'd be super cool to drop one or both of them if the UI isn't using them.

jshearer (approved):
LGTM, thanks for removing those unused fields @travjenkins! I think this needs a 

```
cargo sqlx prepare --workspace
```

and then it should be good to go

PR #2888, opened 2026-04-24
Prevent tenants from enabling enforce_sso without an sso_provider_id

GregorShear:
Adds a CHECK constraint on `public.tenants` preventing `enforce_sso = true` when `sso_provider_id is null` — a combination that locks matching-domain users out with no SSO path in. Caught after a tenant was set to `enforce_sso` without a provider.


jshearer (approved):
LGTM

PR #2889, opened 2026-04-24
doc: add ExtractorPlan with merge-join field extraction

### LINKED ISSUE #2872: performance: compose extracted fields into minimal traversals, to reduce AsNode::Field::get() calls

williamhbaker:
**Description:**

This PR optimizes field extraction by pre-compiling a `Vec<Extractor>` into **blocks** (runs of consecutive sibling-leaf extractors under a common parent). Each block navigates to the parent once and merge-joins the parent's fields against the block's leaf names. This replaces `N` root-to-leaf `Pointer::query` calls and `N` `Fields::get` calls with a single walk and a linear sweep.

On its own, the merge-join of many sorted extractor fields is a significant win. Avoiding repeated `Pointer::query` calls for common prefixes naturally extends this mechanism for further optimization.

### Output Equivalence

The plan produces byte-for-byte identical output to the reference Extractor path in the common case.

**Known edge case: duplicate property names at 16+ fields**. `Fields::get` uses a linear find below 16 fields (first match) and `binary_search_by` at or above (arbitrary match per Rust's documented semantics). The plan's merge-join always returns the first match, so the two paths agree below the threshold and can disagree at or above.  The plan's defined tiebreak is "first occurrence wins". See the `duplicate_keys_diverge_at_fields_get_threshold` unit test for the concrete construction.

### Performance Impact
The included benchmarks demonstrate a 2x to 5x+ improvement in typical cases:
* **Flat, wide objects:** See ~2x improvement, primarily benefiting from the linear merge-join.
* **Wide, deeply nested structures:** See 5x+ improvement, compounding the benefits of the linear sweep and the reduction in tree-walking.

### Edge Cases & Degradation
* **Parity:** If the selected fields are completely scattered throughout the object and form no blocks, the compiled plan essentially degrades gracefully to using single `Fields::get` calls, performing at parity with the unoptimized approach.
* **Worst-case scenario:** Performance can theoretically be worse than pure `Fields::get` in a wide object containing small clusters of fields that form isolated blocks scattered throughout it. This forces multiple linear scans instead of multiple binary searches. This is an adversarial and highly uncommon construction, but it serves as the primary motivator for not allowing *single* fields to form a block.

### Design Notes
In a preliminary implementation, I included a tuning parameter for the minimum number of eligible fields needed to form a block. It turned out that a threshold of just 2 was beneficial in plausible cases, making it easier to simply define a block as "any number of consecutive fields" (greater than one). The only role a tuning parameter would play is protecting against the adversarial case described above, which seems too rare in practice to justify the added complexity.

Closes https://github.com/estuary/flow/issues/2872

**Workflow steps:**

**Notes for reviewers:**

### DIFF
--- a/crates/doc/src/extractor.rs
+++ b/crates/doc/src/extractor.rs
@@ -80,48 +80,7 @@ impl Extractor {
         &'s self,
         doc: &'n N,
     ) -> Result<&'n N, Cow<'s, serde_json::Value>> {
-        let Some(node) = self.ptr.query(doc) else {
-            return Err(Cow::Borrowed(&self.default));
-        };
-
-        match self.magic {
-            None => { /* sorry, kid, I guess your parents aren't coming back */ }
-            Some(Magic::UuidV1DateTime) => {
-                if let Some(date_time) = match node.as_node() {
-                    json::Node::String(s) => Some(s),
-                    _ => None,
-                }
-                .and_then(|s| proto_gazette::uuid::parse_str(s).ok())
-                .and_then(|(_producer, clock, _flags)| {
-                    let (seconds, nanos) = clock.to_unix();
-                    time::OffsetDateTime::from_unix_timestamp_nanos(
-                        seconds as i128 * 1_000_000_000 + nanos as i128,
-                    )
-                    .ok()
-                }) {
-                    // Use a custom format and not time::format_description::well_known::Rfc3339,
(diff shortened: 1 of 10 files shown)


jgraettinger (approved):
Looks great. I cannot think of any downsides or issues. I'm very unconcerned about the change in handling for duplicate properties, our commitment here is only that you get an arbitrary one.

PR #2896, opened 2026-04-27
Fix missing User-Agent on postgrest client, encourage users to update flowctl if a new version is available

GregorShear:
## Summary

- Fixes a bug where the `flowctl-<version>` User-Agent was set on the agent-API HTTP client but never on the PostgREST client, so all PostgREST calls arrived with an empty UA
- flowctl checks github for newer versions, prompts user to upgrade

```
$ flowctl some_command
WARN You're running an outdated version of flowctl — please update to 0.6.6
```

### DIFF
--- a/crates/flow-client/src/client.rs
+++ b/crates/flow-client/src/client.rs
@@ -54,14 +54,15 @@ impl Client {
         );
 
         let http_client = reqwest::ClientBuilder::new()
-            .user_agent(user_agent)
+            .user_agent(user_agent.clone())
             .build()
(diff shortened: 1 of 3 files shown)


jshearer (approved):

willdonnelly:
The version check added in this PR appears to trigger on any dev builds of flowctl (or at least the ones I can fetch from GitHub `dev-next`). Not a big deal, but mildly annoying and it took me a minute to figure out why it was complaining.

PR #2899, opened 2026-04-28
shuffle, publisher: refactors in support of runtime-v2

jgraettinger:
A stack of self-contained refactors against `crates/shuffle` and
`crates/publisher`, extracted from the longer-lived `runtime-v2`
feature branch so they can land independently.

Themes:

- **AuthZ shape.** Journal client construction moves to a shared
  subject/object `ClientFactory` in `gazette::journal`, replacing the
  prior bespoke factories in shuffle and publisher. Bindings now carry
  an AuthZ object scope (a partitions-template prefix or specific
  journal name) and clients are built lazily per-binding.

- **Frontier progress.** Shuffle's frontier exposes partial
  ("peek") progress while hints are still resolving, letting the
  coordinator make forward progress (e.g. log segment release) ahead
  of full resolution.

- **ACK intents.** Publisher's transaction intents are pre-serialized
  to NDJSON `Bytes` at build time, and `write_intents` tolerates
  orphan journals by flushing them through an ephemeral client and
  ignoring `NotFound` — so recovered intents from a prior task
  generation drain cleanly across data resets.

- **Error ergonomics.** `Frontier::new` returns a typed `thiserror`
  enum in place of `anyhow::bail!` strings.

See individual commit messages for the mechanical details.


crates/flow-client-next/src/workflows/user_collection_auth.rs
```diff
@@ -30,6 +30,59 @@ pub fn new_journal_client(
...
+///
+/// The AuthZ object must be a collection partition template prefix.
+/// These have the form "acmeCo/path/to/collection/{generation_id}/",
+/// which is mapped to a collection by stripping off the "/{generation_id}/"
+/// suffix.
+///
+/// The partition template prefix of certain legacy collections were created
+/// without a "/{generation_id}/" suffix -- these are supported as well.
+///
+/// Panics if the partition template prefix is not in the expected format.
```
williamhbaker, line 48: Stale comment? I'm not seeing it panic, but rather take the `_` catch-all path on line 66.

crates/publisher/src/publisher.rs
```diff
@@ -115,46 +162,79 @@ impl Publisher {
...
+                    Err(status) => Err(status),
+                }
+            });
+        _ = futures::future::try_join_all(ephemeral_flushes).await?;
+
         Ok(())
     }
+
+    /// Access the lazy Client and partitions watch for the binding at `index`.
+    /// Primarily used by tests.
```
williamhbaker, line 230: Maybe this will change with later use, but from what I can tell this is _only_ used by tests right now, so maybe a `pub(crate)` instead of `pub` if it's going to stay that way.

williamhbaker (approved):
LGTM

crates/flow-client-next/src/workflows/user_collection_auth.rs
jgraettinger: Yes, stale. Will fix in a future PR.

crates/publisher/src/publisher.rs
jgraettinger: It's used in the `e2e-support` crate

jgraettinger:
Thanks!

PR #2902, opened 2026-04-29
billing: add tenant billing contact fields and per-tenant controller

jshearer:
## Summary

Customers cannot self-serve billing email changes today. Every request requires manual Stripe intervention. This adds admin-editable billing contact fields on `tenants`, a GraphQL mutation for managing them, and a per-tenant controller that reconciles the Stripe-backed subset to Stripe asynchronously.

- Adds `billing_email`, `billing_name`, and `billing_address` fields to `tenants`, letting admins self-serve billing contact changes through a new `setBillingContact` GraphQL mutation instead of requesting manual Stripe edits
- Introduces a per-tenant `TenantController` automation that reconciles DB-authoritative billing contact data to Stripe asynchronously
- Updates existing customer-creation paths (`createBillingSetupIntent`, `billing-integrations`) to prefer tenant-managed billing email over JWT claims when creating new Stripe customers

## How it works

The `setBillingContact` mutation writes to the DB. A trigger then wakes the tenant's controller task, which reads current DB state, compares against the Stripe customer, and calls `update_customer_billing_profile` to sync the changes to the Stripe customer record if they differ.

## Migration

`20260429120000_tenant_controller_billing_contact.sql`:
- Adds columns. New tenants get a controller task via an insert trigger. Existing tenants get one lazily: `wake_tenant_controller` creates the task on-demand if `controller_task_id` is null, so `setBillingContact` or customer-creation paths work without a pre-existing task row.
- Any source of change to these billing fields will trigger the automation to sync to stripe
- Backfills `billing_email` and `billing_address` from CDC-synced `stripe.customers` so existing data matches Stripe without triggering reconciliation. Only tenants that received billing data from this backfill get a controller task row; the rest get one on first use.

## Testing
I tested this e2e in a local stack with a testmode Stripe API key


crates/control-plane-api/src/server/public/graphql/billing/mutations.rs
```diff
@@ -103,6 +117,46 @@ impl BillingMutation {
         })
     }
 
+    async fn set_billing_contact(
+        &self,
+        ctx: &Context<'_>,
+        tenant: String,
+        email: Option<String>,
+        name: Option<String>,
```
GregorShear: AFAICT we don't send `name` to Stripe. Do we need to capture it at all?

crates/control-plane-api/src/server/public/graphql/billing/mutations.rs
```diff
@@ -103,6 +117,46 @@ impl BillingMutation {
...
+        address: Option<BillingAddressInput>,
+    ) -> Result<SetBillingContactPayload> {
+        let env = ctx.data::<crate::Envelope>()?;
+        let tenant = validate_tenant_name(&tenant)?;
+        verify_authorization(env, tenant.as_str(), models::Capability::Admin).await?;
```
GregorShear: How about rebasing so we have the new billing capabilities to choose from here

crates/control-plane-api/src/server/public/graphql/billing/mutations.rs
```diff
@@ -103,6 +117,46 @@ impl BillingMutation {
         })
     }
 
+    async fn set_billing_contact(
+        &self,
+        ctx: &Context<'_>,
+        tenant: String,
+        email: Option<String>,
```
GregorShear:
```gql 
setBillingContact(tenant: "gCo/") {
    ...
}
```

this succeeds, and deletes the contact info from the tenant, but the change doesn't propagate to stripe. do we even want to allow removing your billing email? 

i might at least make a case for no action if the field isn't provided, and if you do want to delete a field (in the event it's allowed) then explicitly set it to null: `email: null`

would need to make this change: 
```suggestion
        email: MaybeUndefined<String>,
```

supabase/migrations/20260429120000_tenant_controller_billing_contact.sql
```diff
@@ -0,0 +1,125 @@
+-- Add billing contact fields and per-tenant controller task infrastructure.
```
GregorShear, line 1: wrap the migration in a transaction

my understanding is that supabase would wrap for us if we were using their migration feature, but since we're just executing these directly...

supabase/migrations/20260429120000_tenant_controller_billing_contact.sql
```diff
@@ -0,0 +1,125 @@
...
+BEGIN
+    FOR r IN
+        SELECT tenant FROM public.tenants
+        WHERE controller_task_id IS NULL
+          AND (billing_email IS NOT NULL OR billing_address IS NOT NULL)
+    LOOP
+        PERFORM internal.create_tenant_controller(r.tenant);
+    END LOOP;
+END;
+$$;
```
GregorShear: > The backfill UPDATE fires the new AFTER UPDATE trigger, which makes the trailing DO block dead code — migration lines ~95–125
> 
> The tenant_billing_contact_update trigger is created before the backfill UPDATE. So when the backfill sets billing_email/billing_address (NULL → value, which IS DISTINCT FROM passes), the trigger fires per row → wake_tenant_controller → controller_task_id is NULL → it creates the controller right there. By the time the final DO block runs WHERE controller_task_id IS NULL AND (billing_email IS NOT NULL OR …), there's nothing left to match.
> 
> So the DO block never does anything, and as a side effect every backfilled tenant also gets a send_to_task wake queued during the migration (and a no-op Stripe find_customer on first poll). My sense is the intent was for the DO block to be the thing that seeds controllers — if so, either create the trigger after the backfill, or drop the DO block and lean on the trigger. Worth being explicit about which, because right now the behavior and the comment disagree. At tenant scale this is also a small thundering-herd of controller polls at deploy time.

crates/control-plane-api/src/server/public/graphql/billing/tenant.rs
```diff
@@ -37,9 +38,21 @@ impl TenantBilling {
...
+        contact::fetch_billing_contact(&env.pg_pool, &self.tenant)
+            .await
+            .map_err(|err| async_graphql::Error::new(err.to_string()))
+    }
+
+    async fn payment_methods(&self) -> Result<Vec<PaymentMethod>> {
+        let Some(customer) = self
+            .provider
+            .as_ref()
+            .find_customer(&self.tenant)
```
GregorShear: > payment_methods now bypasses the CustomerDataLoader while primary_payment_method still uses it — crates/control-plane-api/src/server/public/graphql/billing/tenant.rs:48 vs crates/control-plane-api/src/server/public/graphql/billing/tenant.rs:66
> 
> payment_methods was changed to call self.provider.find_customer(...) directly, but primary_payment_method (and invoices.rs) still go through the loader. A client selecting both fields on one billing { … } now does two separate customer lookups against Stripe for the same tenant, where the loader would've deduped them within the request. Is there a reason to diverge here? I'd either keep both on the loader or move both off it — the inconsistency is what'll bite us.

crates/control-plane-api/src/server/public/graphql/billing/mutations.rs
jshearer: 💡

crates/control-plane-api/src/server/public/graphql/billing/mutations.rs
jshearer: Good point. I don't think we should allow unsetting any of these fields, and in fact you should be required to set all of them at once. I'll make that change, which should eliminate this as a concern.

crates/control-plane-api/src/server/public/graphql/billing/mutations.rs
jshearer: Yes... we do want to capture it as part of the requirements for this feature. We can't set it as the customer name in Stripe quiet yet because there are some workflows that rely on stripe customer names being the tenant ID. In a future PR we'll update those (or honestly get rid of them when we do the tenant lifecycle work at some point), at which point we can update the Stripe customer name

jshearer: For now I updated the code to store it on a Customer metadata field `estuary.dev/customer_name`

crates/agent/src/tenant_controller/billing_contact.rs
```diff
@@ -0,0 +1,145 @@
...
+    let customer = provider
+        .find_customer(&tenant_row.tenant)
+        .await
+        .context("looking up Stripe customer")?;
+    let Some(customer) = customer else {
+        return Ok(());
+    };
+
+    // Desired state comes from the DB. A NULL column means "no desired value",
+    // so it never counts as a mismatch and is never pushed to Stripe.
```
GregorShear, line 91: add a test for this non-obvious behavior?

GregorShear (approved):
I could see a little more test coverage around the reconciliation logic, but i can't find anything else worth mentioning... lgtm

PR #2907, opened 2026-04-30
Inject controller defaults into `effectiveAlertConfig` resolution

jshearer:
## Summary

After merging #2868, `effectiveAlertConfig` only returned fields from explicit `alert_configs` DB rows. A task with no prefix-level overrides would return just the fields that happened to be set:

```json
"effectiveAlertConfig": {
  "config": { "dataMovementStalled": { "condition": { "stalledFor": "4h" } } },
  "provenance": [{ "path": "dataMovementStalled.condition.stalledFor", "source": "estuary/jshearer/" }]
}
```

It should include all alert types with their controller defaults (provenance `source: null`):

```json
"effectiveAlertConfig": {
  "config": {
    "shardFailed": { "enabled": true, "condition": { "failures": 3, "per": "8h" } },
    "taskChronicallyFailing": { "enabled": true, "autoDisable": false, "condition": { "failingFor": "30days" } },
    "taskIdle": { "enabled": true, "autoDisable": false, "condition": { "idleFor": "30days" } }
  },
  "provenance": [
    { "path": "shardFailed.condition.failures", "source": null },
    { "path": "shardFailed.enabled", "source": null },
    ...
  ]
}
```

This fixes it by providing the defaults from `ControllerConfig` into `App` so they can be injected into the GraphQL request context and rendered as the default layer for the effective alert config.

### DIFF
--- a/crates/agent/src/controllers/abandon.rs
+++ b/crates/agent/src/controllers/abandon.rs
@@ -1,4 +1,7 @@
-use super::{ControllerConfig, ControllerState, NextRun, alerts};
+use super::{
+    ControllerConfig, ControllerState, NextRun, ResolvedAlertConfig,
+    ResolvedTaskChronicallyFailingConfig, ResolvedTaskIdleConfig, alerts,
+};
 use crate::controllers::activation::has_task_shards;
 use crate::controllers::publication_status::PendingPublication;
 use crate::controlplane::ControlPlane;
@@ -33,16 +36,22 @@ impl std::fmt::Display for DisableReason {
     }
 }
 
-fn auto_disable_enabled(alert_config: Option<&models::AlertConfig>, reason: DisableReason) -> bool {
+fn auto_disable_enabled(alert_config: &ResolvedAlertConfig, reason: DisableReason) -> bool {
     match reason {
-        DisableReason::Idle => alert_config
-            .and_then(|a| a.task_idle.as_ref())
-            .and_then(|t| t.auto_disable)
-            .unwrap_or(false),
-        DisableReason::ChronicallyFailing => alert_config
-            .and_then(|a| a.task_chronically_failing.as_ref())
-            .and_then(|c| c.auto_disable)
-            .unwrap_or(false),
+        DisableReason::Idle => matches!(
+            &alert_config.task_idle,
+            ResolvedTaskIdleConfig::Enabled {
+                auto_disable: true,
+                ..
+            }
+        ),
+        DisableReason::ChronicallyFailing => matches!(
+            &alert_config.task_chronically_failing,
+            ResolvedTaskChronicallyFailingConfig::Enabled {
+                auto_disable: true,
+                ..
+            }
+        ),
     }
 }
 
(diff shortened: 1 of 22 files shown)


crates/agent/src/controllers/mod.rs
```diff
@@ -92,6 +92,46 @@ impl Default for ControllerConfig {
...
+            shard_failed: Some(models::ShardFailedConfig {
+                enabled: Some(true),
+                condition: Some(models::ShardFailedCondition {
+                    failures: Some(self.alert_after_shard_failures),
+                    per: Some(self.shard_failure_retention.to_std().unwrap_or_default()),
+                }),
+            }),
+            task_chronically_failing: Some(models::TaskChronicallyFailingConfig {
+                enabled: Some(true),
+                auto_disable: Some(false),
```
GregorShear, line 115: This is cosmetic for the API, right? And if we decide that `task_chronically_failing` should be disabled globally, we'd have to make that change somewhere else _and_ remember to update this value so the API reports the correct default val?

jshearer: Well, kind of cosmetic? There isn't a global config knob for these that are hard-coded in here. For the `auto_disable` flags, for example, that default is defined here (`unwrap_or(false)`):

https://github.com/estuary/flow/blob/0c3f388254b1d8e5c0bbe3d70759f7a240c2ad74/crates/agent/src/controllers/abandon.rs#L36-L47

Similarly, `enabled`'s default is encoded as:

https://github.com/estuary/flow/blob/0c3f388254b1d8e5c0bbe3d70759f7a240c2ad74/crates/agent/src/controllers/abandon.rs#L164-L170

---

> If we decide that task_chronically_failing should be disabled globally, we'd have to make that change somewhere else and remember to update this value so the API reports the correct default val

Yep, that's true

crates/agent/src/controllers/abandon.rs
```diff
@@ -33,16 +33,18 @@ impl std::fmt::Display for DisableReason {
...
-fn auto_disable_enabled(alert_config: Option<&models::AlertConfig>, reason: DisableReason) -> bool {
+fn auto_disable_enabled(alert_config: &models::AlertConfig, reason: DisableReason) -> bool {
     match reason {
         DisableReason::Idle => alert_config
-            .and_then(|a| a.task_idle.as_ref())
+            .task_idle
+            .as_ref()
             .and_then(|t| t.auto_disable)
-            .unwrap_or(false),
+            .expect("defaults layer populated task_idle.auto_disable"),
```
GregorShear: Does this now depend on the default being set in `pub fn alert_config_defaults()` ? Should we confirm the defaults are defined in some unit tests so we don't remove them and crash this by accident?

jshearer: Yeah, valid. I ended up refactoring this into a separate struct, `ResolvedAlertConfig` so there are no longer any `unwrap_or()`s _or_ `expect`s in the controllers at all.

crates/control-plane-api/src/server/mod.rs
```diff
@@ -34,6 +34,7 @@ pub enum Rejection {
 /// App is the wired application state of the control-plane API.
 pub struct App {
     pub _id_generator: std::sync::Mutex<models::IdGenerator>,
+    pub alert_config_defaults: Option<models::AlertConfig>,
```
GregorShear: does this need to be optional?

jshearer: I figured out how to get rid of this field entirely

GregorShear (approved):
I haven't found any more holes to poke in this...

We might consider taking a pass at the graphql schema doc comments to make sure they're useful for an agent (yes, this should have been a comment on the original PR... and we could also punt for now and fill it in later - up to you)

jshearer:
I would rather get this in and do a comprehensive pass over the graphql api doc comments, though I agree that we should do it.

PR #2912, opened 2026-05-01
shuffle(slice): require broker observation of journal create_revision

### LINKED ISSUE #2911: create and then read journal race

williamhbaker:
**Description:**

Set `ReadRequest.min_etcd_revision` to the journal's `CreateRevision` on the slice's metadata probe and streaming read. The broker's resolver treats this as a floor and waits for its KeySpace to catch up before answering, avoiding a spurious `JOURNAL_NOT_FOUND` when the read RPC reaches a broker whose Etcd watch hasn't yet observed the journal.

A couple of other options were considered, but this is ultimately a more principled approach:
- **Client-side retries based on known required revs vs. broker return revs** ❌: Self-contained in the client code and generally "do-able", but  bakes a new "is the broker caught up?" predicate into every retry path that cares, and adds additional network round-trips on retries.
- **Have the client thread through a complete `broker::Header` ❌**: The broker honors `proxyHeader.Etcd.Revision` as a min-revision floor, so we could repurpose it. But that field is server-to-server proxy machinery; using it from a client means manufacturing a credential (`cluster_id` sourced from a real `ListResponse`, route that passes validation, etc.) and threading it through every layer that produces a read request, all to communicate a single `int64`. A dedicated request field says what we mean.

See also: https://github.com/gazette/core/pull/469

Closes https://github.com/estuary/flow/issues/2911

**Workflow steps:**

**Notes for reviewers:**

### DIFF
--- a/crates/proto-gazette/src/protocol.rs
+++ b/crates/proto-gazette/src/protocol.rs
@@ -375,6 +375,11 @@ pub struct ReadRequest {
     /// to skip persisted Fragments having a modication time before the bound.
     #[prost(int64, tag = "8")]
     pub begin_mod_time: i64,
+    /// Minimum Etcd revision the broker must have read through before resolving
+    /// the request. If the broker is behind, it waits for its KeySpace to catch
+    /// up.
+    #[prost(int64, tag = "9")]
+    pub min_etcd_revision: i64,
 }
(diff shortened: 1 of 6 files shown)


jgraettinger (approved):
LGTM, that was easy

PR #2915, opened 2026-05-04
fix CI: add missing `build:gazette` step

jshearer:
0c3f388254 added `mise run build:gazette` to `mise/tasks/ci/platform-build` and `ln ${HOME}/go/bin/gazette` to `ci:package`, but didn't update `platform-build.yaml` to match. The `ci:package` step fails trying to hard-link a gazette binary that was never built.


williamhbaker (approved):
LGTM

PR #2922, opened 2026-05-06
docs: distinguish truncate from drop+recreate in backfill behavior

jwhartley:
## Summary

A routine materialization backfill or dataflow reset runs `TRUNCATE` against destination tables, preserving table-level DDL (partitioning, clustering, etc.). The drop-and-recreate path only fires when the backfill is coupled with an incompatible schema change in the same publication — an incompatible field type change, a primary key change, or the `always_drop_tables_on_backfill` feature flag.

Existing docs presented drop+recreate as the default behavior, which is wrong and pushes customers (especially BigQuery users with custom partitioning) toward unnecessary workarounds.

The decision logic lives in [materialize-boilerplate/materializer.go](https://github.com/estuary/connectors/blob/main/materialize-boilerplate/materializer.go#L520-L600) — same code path for every SQL/warehouse materialization.

## Files updated

- **`guides/backfilling-data.md`** — the highest-impact file. Rewrote the Materialization backfill and Dataflow reset sections to lead with the truncate behavior, with drop+recreate framed as the schema-change exception. Updated the consideration list and the summary table accordingly.
- **`concepts/advanced/evolutions.md`** — three lines that conflated truncate with recreate. Clarified that the `backfill` option truncates when possible and only drops+recreates when the schema change requires it.
- **`reference/Connectors/materialization-connectors/BigQuery.md`** — replaced the partitioning caution box with a corrected info block. Leads with the safe path, lists the three drop+recreate triggers, and adds `onIncompatibleSchemaChange` as a protection mechanism for customers with custom DDL.

## Test plan

- [ ] Render docs preview and verify all three pages read cleanly
- [ ] Confirm `[onIncompatibleSchemaChange](/concepts/advanced/evolutions/)` link resolves (no specific anchor exists for that field name on the evolutions page; landing on the page is acceptable)
- [ ] Confirm the existing intra-doc links in `BigQuery.md` (`#dataflow-reset`, `#materialization-backfill`, `#incremental-backfill`) still resolve after `backfilling-data.md` edits

### DIFF
--- a/site/docs/reference/Connectors/materialization-connectors/BigQuery.md
+++ b/site/docs/reference/Connectors/materialization-connectors/BigQuery.md
@@ -199,6 +199,20 @@ drop table <your_dataset>.<your_schema>.<your_table>_copy;
 ```
 6. Re-enable the materialization to continue materializing data to the now partitioned table.
 
-:::caution
-A [dataflow reset capture backfill](/reference/backfilling-data/#dataflow-reset) or [materialization backfill](/reference/backfilling-data/#materialization-backfill) will drop and recreate the table, removing any custom partitioning. If you need to backfill but don't need to drop the table, use an [incremental capture backfill](/reference/backfilling-data/#incremental-backfill) instead. Otherwise, you will need to re-apply the partitioning conversion steps above after the backfill.
+:::info
+A routine [dataflow reset](/reference/backfilling-data/#dataflow-reset) or [materialization backfill](/reference/backfilling-data/#materialization-backfill) with no schema changes runs `TRUNCATE TABLE`, which preserves partitioning, clustering, and all other table-level metadata. In that case you do **not** need to re-apply table partitioning.
+
+The table is dropped and recreated — losing custom partitioning and any other DDL applied outside of Estuary — only when the backfill is coupled with an incompatible schema change in the same publication. Cases that trigger drop + recreate:
+
+- A selected field's type changes incompatibly (e.g., string → integer)
+- The source collection's primary key changes
+- The materialization has the `always_drop_tables_on_backfill` feature flag set
+
+To protect custom partitioning when an incompatible schema change is detected, configure [`onIncompatibleSchemaChange`](/concepts/advanced/evolutions/) on the binding (or the materialization). The default is `backfill`, which is what triggers the drop + recreate above. Other options:
+
+- `abort` — fail the publication so the change is never applied
+- `disableBinding` — disable the affected binding instead of backfilling it
+- `disableTask` — disable the entire materialization
+
+If you need to keep existing rows during a backfill, use an [incremental capture backfill](/reference/backfilling-data/#incremental-backfill), or set the `retain_existing_data_on_backfill` feature flag on the materialization.
 :::
--- a/site/docs/reference/Connectors/materialization-connectors/BigQuery.md
+++ b/site/docs/reference/Connectors/materialization-connectors/BigQuery.md
@@ -200,19 +200,7 @@ drop table <your_dataset>.<your_schema>.<your_table>_copy;
 6. Re-enable the materialization to continue materializing data to the now partitioned table.
 
 :::info
-A routine [dataflow reset](/reference/backfilling-data/#dataflow-reset) or [materialization backfill](/reference/backfilling-data/#materialization-backfill) with no schema changes runs `TRUNCATE TABLE`, which preserves partitioning, clustering, and all other table-level metadata. In that case you do **not** need to re-apply table partitioning.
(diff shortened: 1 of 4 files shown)


jwhartley:
[Regarding the content](https://estuaryworkspace.slack.com/archives/C03Q2NRFKDL/p1778064793529069?thread_ts=1778041914.442299&cid=C03Q2NRFKDL):
> Mahdi Dibaiee  [8:53 PM]
> This looks correct to me :+1::skin-tone-3:

site/docs/reference/Connectors/materialization-connectors/BigQuery.md
```diff
@@ -199,6 +199,20 @@ drop table <your_dataset>.<your_schema>.<your_table>_copy;
 ```
 6. Re-enable the materialization to continue materializing data to the now partitioned table.
 
-:::caution
-A [dataflow reset capture backfill](/reference/backfilling-data/#dataflow-reset) or [materialization backfill](/reference/backfilling-data/#materialization-backfill) will drop and recreate the table, removing any custom partitioning. If you need to backfill but don't need to drop the table, use an [incremental capture backfill](/reference/backfilling-data/#incremental-backfill) instead. Otherwise, you will need to re-apply the partitioning conversion steps above after the backfill.
+:::info
```
aeluce, line 202: This becomes a fairly large info box. Consider turning this into its own sub-section or keeping more general info in the Backfilling Data page.

site/docs/reference/Connectors/materialization-connectors/BigQuery.md
```diff
@@ -199,6 +199,20 @@ drop table <your_dataset>.<your_schema>.<your_table>_copy;
...
+A routine [dataflow reset](/reference/backfilling-data/#dataflow-reset) or [materialization backfill](/reference/backfilling-data/#materialization-backfill) with no schema changes runs `TRUNCATE TABLE`, which preserves partitioning, clustering, and all other table-level metadata. In that case you do **not** need to re-apply table partitioning.
+
+The table is dropped and recreated — losing custom partitioning and any other DDL applied outside of Estuary — only when the backfill is coupled with an incompatible schema change in the same publication. Cases that trigger drop + recreate:
+
+- A selected field's type changes incompatibly (e.g., string → integer)
+- The source collection's primary key changes
+- The materialization has the `always_drop_tables_on_backfill` feature flag set
```
aeluce: Is this feature flag specific to BigQuery? If not, it may be useful to mention in the Backfilling Data page and/or the Feature Flags page (I see that `retain_existing_data_on_backfill`, mentioned later, is already in Feature Flags).

aeluce (approved):
Thanks for the updates. Should I modify the backfill types image as well? It currently indicates that destination tables get dropped. I could try to make it more general. Or if we want to show the variations that we now offer, it might be cleaner to split the overall image into a separate one for each backfill method and then incorporate a kind of decision tree split where applicable. Either way, it might be tricky to visually differentiate between empty/repopulate and drop/recreate…

site/docs/reference/Connectors/materialization-connectors/BigQuery.md
jwhartley: Good call. Trimmed this info box down to a 4-line pointer that just states "TRUNCATE preserves partitioning" and links out to a new `Schema changes during backfill` sub-section in `backfilling-data.md` for the full triggers + `onIncompatibleSchemaChange` details. Lets BigQuery focus on the partitioning concern and keeps the cross-connector behavior in one place.

site/docs/reference/Connectors/materialization-connectors/BigQuery.md
jwhartley: Not BigQuery-specific — it's gated in `materialize-boilerplate/materializer.go`, so it applies to every relational SQL/warehouse materialization. Added an entry for it on the Feature Flags page (next to `retain_existing_data_on_backfill`), and the new `Schema changes during backfill` section in `backfilling-data.md` lists it alongside the other drop+recreate triggers.

jwhartley:
Thanks for the review! Just pushed a commit addressing both inline comments — `always_drop_tables_on_backfill` is documented on the Feature Flags page, and the BigQuery info box now points out to a new generalized `Schema changes during backfill` section in `backfilling-data.md`.

@aeluce On the backfill types image: generalizing it sounds great. Empty/repopulate vs. drop/recreate is genuinely an implementation detail that most users don't need to know about — the user-facing concern is "data is replaced." If the image just shows the destination tables being refreshed (without committing to a specific mechanism), it'll match the new docs and won't need updating again the next time the underlying behavior shifts.

Issue #2831, opened 2026-04-06
runtime: tasks should noisily log & retry log append failures


williamhbaker:
A failure to append logs of a running task currently is silently ignored, and the task will continue running and not writing logs even if the underlying issue is fixed, until the task is restarted.

Logging is important for task operation, especially for things like inferred schema update logs. Tasks should noisily log failures to write to log journals and retry, as they do with data journal appends.

PR #2925, opened 2026-05-07
runtime-v2: introduce `runtime-next` with `flowctl raw preview-next`

jgraettinger:
Foundational work for the architecture in `plans/runtime-v2/plan.md`:
a per-shard Rust `TaskService` and a sidecar **Shuffle Leader** that
together replace the Go-driven transaction loop. Materialization is
implemented end-to-end (open / commit / acknowledge / trigger,
recovery, Frontier↔Checkpoint mapping). Derivations and captures are
not yet implemented.

**The Go runtime is unchanged in this branch.** It still drives the
existing `runtime` crate for every task. The new `runtime-next` code
is reachable only through `flowctl raw preview-next`, which is how
materialization is exercised end-to-end today. Wiring `runtime-next`
into the Go reactor is follow-on work, as is introduction of the sidecar binary.

Commits are arranged for narrative review:

- Design docs first (`plan.md`).
- Then small, self-contained protocol changes (`proto-gazette`,
  `shuffle` Frontier collapse, `materialize` state patches).
- Then a verbatim file copy from `runtime/` → `runtime-next/`, so the
  next commit's diff is reviewable.
- Then the substantive `runtime-next` implementation.
- Finally, `flowctl raw preview-next` as an E2E harness against the
  new stack (replacing the removed `raw shuffle`).

## Test plan

- [x] `cargo test -p runtime-next` (recovery + triggers snapshots)
- [x] `cargo test -p proto-flow`, `cargo test -p shuffle`
- [x] E2E materialization via `flowctl raw preview-next` —
      see `plans/runtime-v2/preview-harness.md`
- [x] Legacy runtime still green under `mise run ci:catalog-test`


AGENTS.md
```diff
@@ -0,0 +1,133 @@
+# CLAUDE.md
```
williamhbaker, line 1: Maybe have this file as a symlink to CLAUDE.md?

go/protocols/runtime/runtime.proto
```diff
@@ -334,3 +336,548 @@ message ConnectorProxyResponse {
...
+  int64 max_etcd_revision = 1;
+}
+
+// Task which is being processed by the runtime.
+// Sent from Controller to Shard, and from Shard zero (only) to Leader
+// after Joined. Other shards do not forward Task.
+message Task {
+  // Task specification (protobuf-encoded bytes).
+  bytes spec = 1;
+  // Collection journal partition to which task states are written.
```
williamhbaker, line 473: Typo? states -> stats

crates/runtime-next/src/publish.rs
```diff
@@ -0,0 +1,147 @@
+//! Publishing surface used by leader actors.
+//!
+//! `Publisher` is the unified entry point. Two variants:
+//!
+//! - `Publisher::Real` wraps a real `publisher::Publisher` and performs
+//!   Gazette journal IO (stats / logs / ACK intents / future capture &
```
williamhbaker, line 6: Might be a future thing, but I was looking for logs publishing here and didn't see it, in relation to https://github.com/estuary/flow/issues/2831

crates/runtime-next/src/shard/rocksdb.rs
```diff
@@ -61,186 +57,110 @@ impl RocksDB {
...
         opts.create_missing_column_families(true);
 
-        // The MANIFEST file is a WAL of database file state, including current live
-        // SST files and their begin & ending key ranges. A new MANIFEST-00XYZ is
-        // created at database start, where XYZ is the next available sequence number,
-        // and CURRENT is updated to point at the live MANIFEST. By default MANIFEST
-        // files may grow to 4GB, but they are typically written very slowly and thus
-        // artificially inflate the recovery log horizon. We use a much smaller limit
-        // to encourage more frequent compactions into new files.
-        opts.set_max_manifest_file_size(1 << 17); // 131072 bytes
```
williamhbaker, line 71: Is removing `set_max_manifest_file_size` intentional?

crates/runtime-next/src/leader/materialize/fsm.rs
```diff
@@ -0,0 +1,2165 @@
...
+                sourced_bytes_total,
+                sourced_docs_total,
+            } in bindings
+            {
+                let min_source_clock = uuid::Clock::from_u64(min_source_clock);
+                let max_source_clock = uuid::Clock::from_u64(max_source_clock);
+                let extent = self.extents.bindings.entry(index).or_default();
+
+                if !max_key_delta.is_empty() {
+                    extent.max_key_delta = max_key_delta;
```
williamhbaker, line 327: I think we need a `max(extent.max_key_delta vs. max_key_delta)` here, otherwise keys for multiple shards  getting recorded out of order would be a problem.

crates/runtime-next/src/leader/materialize/fsm.rs
```diff
@@ -0,0 +1,2165 @@
...
+    let mut policy_close = open_age > task.open_duration.start
+        && last_age > task.last_close_age.start
+        && (!policy_extend || max_combiner > task.combiner_usage_bytes.start)
+        && (!policy_extend || read_bytes > task.read_bytes.start)
+        && (!policy_extend || read_docs > task.read_docs.start);
+    policy_close |= idempotent_replay && !unresolved_hints;
+    policy_close |= close_requested;
+
+    let may_close = policy_close && !unresolved_hints && tail_done;
+    let may_extend =
```
williamhbaker, line 1172: Should we prevent extension (exception being: unresolved hints, I think) if `close_requested`? That way an extended stream of ready frontiers doesn't stop the shard from closing when requested.

crates/runtime-next/src/leader/materialize/fsm.rs
```diff
@@ -0,0 +1,2165 @@
...
+
+        // It's possible that multiple bindings source from the same collection.
+        // We accumulate when reporting by-collection.
+        entry.bytes_behind = entry.bytes_behind.saturating_add_signed(
+            binding_bytes_behind
+                .get(*binding_index as usize)
+                .copied()
+                .unwrap_or_default(),
+        );
+        entry.last_source_published_at = extents.max_source_clock.to_pb_json_timestamp();
```
williamhbaker, line 1216: Also for the case of multiple bindings sourcing from the same collection: I think it would be technically more correct to take the maximum of `entry.last_source_published_at` vs. `extents.max_source_clock`. Perhaps most significantly this would prevent any cases of the `lastSourcePublishedAt` from going backwards in subsequent transactions.

crates/runtime-next/src/leader/materialize/startup.rs
```diff
@@ -0,0 +1,834 @@
...
+            tracing::debug!(
+                ?committed_close,
+                ?hinted_close,
+                ?hinted_frontier,
+                "connector_checkpoint present and matches Recover::hinted_close; applying delta"
+            );
+            committed_close = hinted_close;
+            committed_frontier = committed_frontier.reduce(std::mem::take(&mut hinted_frontier));
+
+            pending_ack_intents = connector_checkpoint.ack_intents;
```
williamhbaker, line 245: Thinking through a couple of scenarios here, for remote authoritative materializations:

- We get the pending ack intents here from the connector's persisted checkpoint, but not the trigger params. Since the persist at `fsm.rs:828` didn't land (or we wouldn't be on this branch?) there won't be any trigger parameters loaded, and so that trigger firing would get skipped.
- Max keys is similar, but I think doesn't have the same problem, since it is persisted at `fsm.rs:523` (hinted), before StartCommit, so maybe trigger params could do the same?

crates/runtime-next/src/leader/materialize/fsm.rs
```diff
@@ -0,0 +1,2165 @@
...
+        // Fold deltas from the extents Frontier into per-binding "bytes behind" gauges.
+        for jf in &extents.frontier.journals {
+            let Some(entry) = binding_bytes_behind.get_mut(jf.binding as usize) else {
+                continue; // Reachable if shuffle service reports invalid binding indices.
+            };
+            *entry += jf.bytes_behind_delta;
+        }
+
+        // Compose the trigger payload now that we have a complete txn-wide view.
+        if task.triggers.is_some() {
```
williamhbaker, line 647: Should we gate this to compute `trigger_params` only if the transaction contained Stored docs? That's something I remember from the original implementation, to avoid firing triggers on empty transactions.

williamhbaker (commented):
A few comments below. Overall looks good; there's a lot here but the individual pieces were relatively easy to understand. How everything fits together in the complete end-to-end workings was more difficult to really internalize, and I'm not confident I 100% got there on everything, so let me know if any of these are off the mark.

AGENTS.md
jgraettinger: Yep, I'd attempted a hard-link first but that doesn't work. I have it as a symlink in the next PR.

go/protocols/runtime/runtime.proto
jgraettinger: This is removed altogether in the next PR (wasn't needed -- the `labeling` already carries the journal so this is duplicative).

crates/runtime-next/src/publish.rs
jgraettinger: This gets reworked a bit in the next PR, but yea: ops logs publishing is still deferred, and we let the Go runtime handle it. Doing otherwise would require a more invasive change to the Go runtime behavior so I deferred for now.

We do continue to get logs -- they're just bridged from Rust => Go and published from there over the current mechanism, rather than introducing a new Rust-only path.

crates/runtime-next/src/shard/rocksdb.rs
jgraettinger: Definitely not. Restored on the next PR.

crates/runtime-next/src/leader/materialize/fsm.rs
jgraettinger: Good catch, thanks. Adding more test coverage here.

crates/runtime-next/src/leader/materialize/fsm.rs
jgraettinger: Done

crates/runtime-next/src/leader/materialize/fsm.rs
jgraettinger: Yes, good call.  As implemented `stopping` has the semantics we want already (extend if we can't close, but as soon as we can close, stop extending), so I've generalized into a `let finishing = close_requested || stopping;`

crates/runtime-next/src/leader/materialize/fsm.rs
jgraettinger: I did consider this case, actually, and kind of shrugged.

The trouble with taking a max here is that it _still_ won't prevent flopping if changes arrive in non-overlapping transactions. Putting a max here is therefore false comfort. I'll add a comment to this effect.

The only scenario where this can meaningfully happen is if you have two bindings of the same collection _also with different priorities_ -- otherwise they'll be in the same cohort and process the exact same txn documents via the causal hint mechanism.

crates/runtime-next/src/leader/materialize/startup.rs
jgraettinger: Yea, I did think about this -- and believe the V1 runtime has the same issue. We persist trigger params _after_ StartedCommit, when the remote authoritative commit has already occurred.

I kinda stopped at "it's not any worse than V1", but agreed this could be better. I'm not entirely sure _how_ to make it better, though. It seems it'd require persisting _two_ sets of trigger params at both Persist points, a hinted set vs committed set, so we can compare the recovered remote-authoritative checkpoint clock and figure out which to use.

jgraettinger:
As discussed over VC, I've addressed all feedback in the next PR in this series, #2952. Thanks!

williamhbaker (approved):
LGTM

PR #2926, opened 2026-05-07
validation: cascade backfill on collection reset regardless of onIncompatibleSchemaChange

### LINKED ISSUE #2916: Data Flow Reset silently disables downstream materialization bindings when onIncompatibleSchemaChange: disableBinding is set
## Summary

Running a Data Flow Reset on a capture silently disables downstream materialization bindings whenever those bindings have `onIncompatibleSchemaChange: disableBinding` configured. Recovery requires manually re-enabling each binding and clicking Backfill on the materialization. We've now seen the same customer hit this twice on production pipelines, and the symptom is confusing because there's no obvious error — bindings just stop writing.

## Repro

1. Have a materialization with a binding configured `onIncompatibleSchemaChange: disableBinding` (or task-level default of the same).
2. Run a Data Flow Reset on the capture that feeds the binding's source collection.
3. Observe that the materialization binding is silently set to `disable: true` after the next publication. No new data flows; no error surfaces; backfill counter is unchanged.

Recovery: re-enable the binding and click Backfill on the materialization. Once `model.backfill > live_spec.backfill`, the next validation skips the reset path and the binding stays enabled.

## Root cause

The check that gates the `onIncompatibleSchemaChange` rules during a collection reset lives at [`crates/validation/src/materialization.rs:760-764`](https://github.com/estuary/flow/blob/main/crates/validation/src/materialization.rs#L760-L764):

```rust
// Was this binding's source collection reset under its current backfill count?
let was_reset = live_spec.is_some_and(|live_spec| {
    live_spec.backfill == model.backfill
        && super::collection_was_reset(&source_spec, &live_spec.collection)
});
```

`collection_was_reset` ([`crates/validation/src/lib.rs:706-727`](https://github.com/estuary/flow/blob/main/crates/validation/src/lib.rs#L706-L727)) returns true when the journal partition template name (which carries a generation ID) differs between live and built specs — exactly the case after a Data Flow Reset.

When `was_reset == true` and the binding has `disableBinding`, the binding is disabled at [`materialization.rs:791-794`](https://github.com/estuary/flow/blob/main/crates/validation/src/materialization.rs#L791-L794):

```rust
(true, models::OnIncompatibleSchemaChange::DisableBinding) => {
    model_fixes.push(format!("disabling binding of reset collection {source}"));
    model.disable = true;
    return (model_path, model, None, None);
}
```

The "user-initiated backfill bypass" that lets the workaround succeed is implicit — it's just the `live_spec.backfill == model.backfill` clause on line 762. When the user clicks Backfill on the materialization, `model.backfill` is incremented in the draft, the equality fails, `was_reset` becomes false, and the `disableBinding` arm is never reached.

The bug, then, is that **a Data Flow Reset on a capture does not propagate a backfill increment to dependent materialization bindings**, so the next materialization validation cannot tell that the user already initiated this reset upstream and treats the resulting collection regeneration as an unsolicited incompatible schema change.

## Why this is surprising

From a user's perspective, Data Flow Reset is itself a deliberate, user-initiated rebuild of the pipeline — the same intent that the manual Backfill click expresses. There is no way to opt into "reset the capture, please also reset the materialization" except by clicking through the materialization separately, and the default `disableBinding` policy actively works against that intent by silently disabling the downstream binding.

Customers who use `disableBinding` defensively (to guard against unsolicited schema drift from auto-discover) end up paying for that defense every time they run a legitimate Data Flow Reset, with no warning surfaced in the UI.

## Suggested fix directions

A few options, roughly in order of preference:

1. **Cascade the backfill on Data Flow Reset.** When a reset is published, increment `backfill` on every binding of every materialization that reads from the reset collection, in the same publication. This treats Reset as a whole-chain user-initiated backfill, which matches user intent. The condition on `materialization.rs:762` then naturally evaluates false and `was_reset` becomes false for the dependent bindings.

2. **Distinguish "reset" from "schema drift" in the rule evaluation.** Add a fourth match arm or a separate signal so that a collection that was reset (detected via partition template) doesn't fall through `onIncompatibleSchemaChange` at all — it always behaves as Backfill, regardless of per-binding policy. This is closer to the current implicit behavior of the Backfill arm at lines 787-789.

3. **At minimum, warn in the UI.** When the user kicks off Data Flow Reset, show that any downstream binding with `disableBinding` will be disabled and require manual recovery. This documents the rough edge without changing semantics, but doesn't fully fix it.

Option 1 is probably the cleanest — Reset is already a destructive, deliberate operation, and there's no scenario I can think of where the user wants the capture rebuilt but the materialization binding silently disabled.

## Customer impact

Two production resets so far on the same customer pipeline; both required us to walk them through the manual recovery in Slack. Not severe enough to be an outage, but it's surprising-by-default for any customer using `disableBinding`, and the recovery isn't discoverable without support help.

jshearer:
## Problem

When a collection is dataflow-reset, the control plane notifies downstream materializations, which go through re-validation. The existing logic applied `onIncompatibleSchemaChange` to those bindings: `disableBinding` disabled the binding, and `disableTask` disabled the entire materialization.

`onIncompatibleSchemaChange` was designed for system-detected schema incompatibilities (field type changes, group-by changes etc), not for user-initiated resets. Phil's original commit 4c10686ef3 routed resets through `onIncompatibleSchemaChange` as a pragmatic shortcut, noting "there's some further opportunities for refactoring, related to the `disableTask` value, but I'm kicking the can on that."

## Solution

When a collection is reset, downstream materialization bindings always cascade the backfill (increment `backfill`) instead of consulting `onIncompatibleSchemaChange`. The exception is `onIncompatibleSchemaChange: abort`, which fails the publication with a visible error.

`onIncompatibleSchemaChange` continues to govern non-reset incompatibilities (group-by changes, connector field conflicts, etc.) as before.

Fixes #2916


crates/validation/tests/materialization_collection_resets.rs
```diff
@@ -31,14 +31,14 @@ test://example/catalog.yaml:
           table: b
         "#,
     );
-    // Expect that the whole task gets disabled, along with the binding for b.
-    // The binding for a should remain enabled.
-    insta::assert_debug_snapshot!(outcome);
+    // Expect that both bindings are backfilled (resets cascade as
+    // backfills for all onIncompatibleSchemaChange variants except abort).
+    insta::assert_debug_snapshot!((selections, errors));
```
jgraettinger: nit: the change to (selections, errors) loses fidelity in the ramifications of the workflow -- for example, you've lost the certainty that the built binding's backfill counter actually increased.

also nit: the actions of the test no longer match its name

jgraettinger (approved):
LGTM % comments re tests

crates/validation/tests/materialization_collection_resets.rs
jshearer: Good points.. 

* updated the tests to snapshot `outcome.built_materializations` instead of `(selections, errors)` so they retain the built model/spec fidelity, including the incremented binding backfill values. 
* renamed/merged the reset behavior coverage into `test_reset_backfills_except_on_abort`

PR #2928, opened 2026-05-07
activate: preserve journal and shard splits across dataflow reset

### LINKED ISSUE #2881: Dataflow reset drops journal splitting, painful for large captures
### Problem

Dataflow reset on a collection removes the collection's journal splitting configuration. Journals that were previously split back to default (single journal per logical partition). For large, high-throughput captures that relied on splitting for parallelism, post-reset performance drops significantly until splits are re-applied.

Customers cannot split journals themselves — splitting is an Estuary-internal operation — so once splitting is lost on reset, throughput stays degraded until Estuary support intervenes to re-split.

### Why it matters

- Large captures (multi-TB, multi-million-row) frequently need journal splitting to keep up with source throughput or to process a backfill within an acceptable window.
- Dataflow reset is commonly used precisely for these captures (e.g. to recover from schema drift, or to restart from the current source state rather than replaying weeks of change data).
- Post-reset, the backfill is often the single largest throughput event the pipeline will ever do — exactly when split parallelism matters most.
- There is a real risk scenario: if the source's replication slot (Postgres, MySQL, etc.) is sensitive to replication lag, the slow post-reset catch-up can cause the slot to fall behind or get purged, escalating to an outage.

### Prior internal discussion

- [Dataflow reset removes journal splitting, flagged as a "pile on" pain alongside the `require`-field-validation issue (see #2880)](https://estuaryworkspace.slack.com/archives/C03QBN83GQ4/p1756919667130159)
  > "collection reset also removes journal splitting which is a bit of a problem for very large captures. Ex. [customer] has one that the replication slot fell behind on over the weekend. It would have no matter what because of under-provisioning, but if that weren't the case, it would have caused a real issue... since customers can't split journals"

### Proposed resolution directions

1. **Preserve journal splits across reset.** The split configuration is an operational concern separate from the collection data / inferred schema. It could be copied forward onto the new collection state as part of reset publication.

2. **Auto-re-apply splits after reset based on a recorded target.** If we can't carry the live split state forward, we could persist the split *intent* (e.g. "this collection should have N splits") on the spec/controller, and have the control plane re-apply it once the reset completes and the collection is receiving data again.

3. **Expose journal splitting to customers.** Out of scope for this issue, but would make this problem a self-service recovery rather than a support ticket. Tracked elsewhere if it isn't already.

Direction (1) is likely simplest if the splitting state is recorded in a place that survives the reset's journal turnover.

### Cross-links

- #2880 — sibling dataflow-reset pain: reset blocks on required fields that only exist in the inferred schema. Same root workflow, different failure mode.
- #2821 — truncations umbrella.

### Labels

`control-plane`, `enhance`

jshearer:
## Summary

* Preserve shard split count across dataflow resets by using the old-generation shard count (instead of the default 1) when creating new-generation shards in `apply_initial_splits`
* Preserve partition journal splits by mirroring old-generation partition topology (key ranges and logical partition field labels) into new-generation partitions via a new `apply_initial_partition_splits` function
* Both changes are no-ops for truly new tasks/collections (empty old-gen) and for tasks where the new-gen already exists

### Note on `dekaf::e2e::not_ready`

The `not_ready` test reached the no-journals state by resetting a populated collection, then racing to observe `LeaderNotAvailable` before journals were recreated. Now that `apply_initial_partition_splits` recreates partition journals at activation time, that state is no longer reachable via reset and the test was bailing every run with `Journals were created before we could test NotReady state`.

The relevant dekaf branch (`Collection::new` returning `NotReady` when `partitions.is_empty()`) is still load-bearing for fresh-published collections where the capture hasn't committed any data yet. Partition journals are created lazily by the runtime mapper on first commit, so dropping the reset cycle and asserting against a fresh `setup` (no inject) exercises the same codepath. 

Fixes #2881

### DIFF
--- a/crates/activate/src/lib.rs
+++ b/crates/activate/src/lib.rs
@@ -17,7 +17,7 @@ pub enum Change {
 }
 
 // JournalSplit describes a collection partition or a shard recovery log.
-#[derive(Debug, Default, Clone, serde::Serialize)]
+#[derive(Debug, Default, Clone, PartialEq, serde::Serialize)]
 pub struct JournalSplit {
     pub name: String,
     pub labels: LabelSet,
@@ -177,6 +177,7 @@ pub async fn activate_collection(
     )?;
 
     let shards = apply_initial_splits(task_template, initial_splits, shards)?;
+    let partitions = apply_initial_partition_splits(task_spec, partitions)?;
     let changes_1 = partition_changes(partition_template, partitions)?;
     let changes_2 = task_changes(task_template, shards, recovery, ops_logs, ops_stats)?;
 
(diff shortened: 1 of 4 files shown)


jgraettinger:
I think this needs to attempt only to preserve splits of an **unpartitioned** collection that **continues** to be unpartitioned. We should not attempt to preserve logical partitions across collection resets, for a couple of reasons:

1. Collection reset is how one _changes_ the logical partitioning being applied to a collection. The former splits may not even be logically partitioned fields of the post-reset spec.
2. Reset is also how one cleans up partition instances which may no longer be relevant or desired, and we provide no other means for removing partition instances today (aside from reset).

So, if a collection is or was partitioned, we should use initial splits of 1 (and not attempt to pre-create any logical partitions). Otherwise, in the common case of a collection without logical partitioning, preserving splits makes plenty of practical sense.

jshearer:
Ahhhh, didn't even think about logical partitioning. Good point, I'll carve that out 👍

jshearer:
Kk, updated so that `apply_initial_partition_splits` now preserves splits _only_ when both old and new generations are unpartitioned. If either is logically partitioned, the function returns its input unchanged, `partition_changes` deletes the old generation's journals, and the actual partitions are created at runtime.

Will squash before merging

jgraettinger (approved):
LGTM

PR #2931, opened 2026-05-07
Reject footguns in `redact` annotation placement

### LINKED ISSUE #2914: Validate redaction annotation placement to prevent silent non-functional or fragile redactions
### Problem

Redaction annotations only function when placed at the **top level of `writeSchema`, outside `$defs.flow://connector-schema`**. The platform does not validate placement, so misplaced annotations either silently fail to redact, work transiently then get clobbered, or appear correct but are undetectable when broken. The failure modes are all silent — customer adds an annotation, sees no error, assumes redaction is active.

### Specific misplacement cases not caught today

| Placement | Outcome | Caught by validation? |
|---|---|---|
| Top of `writeSchema`, outside `$defs.flow://connector-schema` | ✅ Functional and durable across reset/discover | N/A — supported placement |
| `readSchema` only | ❌ Non-functional. Redaction is write-time only | No warning/error |
| Inside `$defs.flow://connector-schema` | ⚠️ Functional today, but the next discover overwrites the connector-schema sub-tree | No warning |
| Single combined `schema:` block | ⚠️ Functional today, but auto-discover transitioning to `writeSchema`/`readSchema` initialises fresh schemas without the customer's annotations | No warning |

The `update_connector_schema` opt-in-marker check in `crates/control-plane-api/src/discovers/specs.rs:495-500` does not catch case 3 — schemas with redact inside `$defs.flow://connector-schema` *do* have the marker, so discover happily overwrites the sub-tree.

### Existing safeguards (correctness-only, not placement)

The validation crate today catches only:
- Redact on a key location → `KeyHasRedact` error (`crates/validation/src/errors.rs:141`, checked in `crates/validation/src/schema.rs:104` and `crates/validation/src/collection.rs:200`)
- Redact on `/_meta/uuid` → `UuidPtrHasRedact` error (`crates/validation/src/errors.rs:149`, checked in `crates/validation/src/collection.rs:791`)

Both are about runtime correctness (would break dedup or destroy the UUID placeholder). Neither flags placement that silently fails to redact at all.

### Why it matters

Redaction is the customer-facing PII / compliance mechanism. Silent loss of redaction means PII flows unredacted into materialized destinations — there has been at least one production incident where a customer had to retroactively delete unredacted data from cloud storage after a discover-driven schema rebuild dropped misplaced redactions.

When a customer hits this in the field, the entire customer-facing playbook becomes "make sure your redactions are at the top level of `writeSchema`, outside `$defs.flow://connector-schema`, not in `readSchema`, not in a single combined `schema:` block." That's a lot of conditions to verify by inspection, and there's no platform feedback when any one of them is violated.

### Prior internal discussion

Two related proposals were sketched in [the redact-loss incident thread](https://estuaryworkspace.slack.com/archives/C03QBN83GQ4/p1772136996315899) and [the original connector-schema design thread](https://estuaryworkspace.slack.com/archives/C07SEA0G40N/p1756415442607559):

1. *Treat schemas without `flow://connector-schema` as manually managed* — partially shipped, as the opt-in-marker check in `update_connector_schema`. Doesn't address the "redact inside connector-schema def" case.
2. *Auto-deselect `redact: block` fields from materialization defaults* — proposed but not shipped. Would prevent redacted fields from being projected even if the redaction itself fails or migrates locations.

The [docs](https://docs.estuary.dev/features/redaction/) describe the supported placement, but rely on the customer reading and following them — there is no validation backstop.

### Proposed resolution directions

1. **Validation-time error or warning for non-functional placement.** Add to `crates/validation/src/collection.rs`:
   - `redact` annotation present in `readSchema` only (no corresponding annotation in `writeSchema` at the same location) → warning recommending move to `writeSchema`. Could be elevated to error after a deprecation window.
   - `redact` annotation inside `$defs.flow://connector-schema` → warning (annotation will be lost on the next discover).
   - `redact` annotation inside a single `schema:` block → warning recommending split `writeSchema`/`readSchema` (annotation will be dropped if the connector emits a sourced schema and auto-discover transitions to split).
2. **Auto-deselect `redact: block` fields from `fields.recommended`.** Per the prior proposal, treat the presence of a `redact` annotation as a sound reason not to include the field in default materialization recommendations. Belt-and-braces against any future placement bug.
3. **Optional UI guidance.** Surface the same checks in the dashboard's schema editor at edit time, before publication.

Direction (1) is the lowest-risk: it's diagnostic-only, doesn't change runtime behavior, and would have prevented every redact-loss incident I'm aware of. Direction (2) is a useful correctness backstop and aligns with existing precedent for handling annotated fields. Direction (3) is a nice-to-have but out of scope for the control-plane work.

### Labels

`control-plane`, `enhance`

jshearer:
## Background

Redaction runs at capture time (or pedantically, at any time that data is written to a collection, which also includes derivations), at the moment documents enter the system. The runtime applies it according to the `redact` annotations present in the collection's write schema. For the annotation to take effect, it has to survive the publication pipeline and end up in the write schema. The platform currently _accepts_ `redact` in several locations where one or both of those conditions fail silently: in `readSchema`, which only materializations consult, and anywhere inside a `flow://...` subtree (`flow://write-schema`, `flow://inferred-schema`, `flow://relaxed-write-schema`, `flow://connector-schema`), each of which the publication pipeline overwrites on every build.

Two categories of misplacement are accepted today with no signal to the user:

* **`redact` only in the read schema.** The read schema is used at materialization time, not at capture time. A redaction annotation in this location _will_ prevent sensitive data from being written to the destination system, but it _won't_ prevent it from being stored in the customer's collection in the first place. 
* **`redact` inside `$defs/flow://*`.** These are virtual schemas that get replaced before publication completes. Discover overwrites `$defs/flow://connector-schema` with `overwrite: true` on every run, and validation inlines `flow://write-schema`, `flow://relaxed-write-schema`, and `flow://inferred-schema` from other sources. Annotations placed inside any of them are erased before they ever reach the runtime.
  * **Concept:** would a better solution be to error if a publication contains _any_ changes to a `flow://` schema? Not actually sure if that's a trivial thing to check though..

This PR adds validation checks that block these misplacements at publication time.

## Changes

Two new errors, modeled on the existing `KeyHasRedact` and `UuidPtrHasRedact`:

* `ReadSchemaRedactNotInWriteSchema`: walks the read schema's locations and errors if any `redact` annotation has no corresponding `redact` at the same pointer in the write schema.
  * **Note:** I suppose technically this doesn't have to be an error... but it feels like we should err on the side of caution and forbid the narrow use-case of writing unredacted data to the collection, and only redacting at materialize time. Any objections?
* `RedactInsideManagedDefs`: scans the user-authored schema model for `redact` annotations inside any `$defs/flow://...` entry and reports the offending JSON pointer. The scan reads the pre-inlining model, so the verbatim copy of the write schema that validation places into `$defs/flow://write-schema` is not misreported.

Closes #2914.

### DIFF
--- a/crates/validation/src/collection.rs
+++ b/crates/validation/src/collection.rs
@@ -191,6 +191,48 @@ fn walk_collection(
         check_uuid_ptr_not_redacted(scope.push_prop("schema"), &write_spec.spec, errors);
     }
 
+    // Redaction runs at capture time using the write schema. Annotations
+    // placed only in the read schema do not protect stored documents.
+    if let Some(read) = &read_spec {
+        check_read_schema_redact_in_write_schema(
+            scope.push_prop("readSchema"),
+            &read.spec,
+            &write_spec.spec,
+            errors,
+        );
+    }
+
+    // Annotations placed inside `$defs` entries whose `$id` is a system-managed
+    // schema URI are silently overwritten by discover or validation-time
+    // inlining. Scan the user-authored models (not the post-inlining
+    // `read_spec.model`, which carries a verbatim copy of the write schema).
+    if let Some(read) = read_model.as_ref() {
+        let write = write_model
+            .as_ref()
+            .expect("split schemas have write_model");
+        check_redact_in_managed_defs(
+            scope.push_prop("writeSchema"),
+            write,
+            &write_spec.spec.curi,
+            errors,
+        );
+        if let Some(read_spec) = &read_spec {
+            check_redact_in_managed_defs(
+                scope.push_prop("readSchema"),
+                read,
+                &read_spec.spec.curi,
+                errors,
+            );
+        }
+    } else if let Some(schema) = schema_model.as_ref() {
+        check_redact_in_managed_defs(
+            scope.push_prop("schema"),
+            schema,
+            &write_spec.spec.curi,
+            errors,
+        );
+    }
+
     let effective_read_spec = read_spec
         .as_ref()
         .map(|Schema { spec, .. }| spec)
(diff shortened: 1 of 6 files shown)


crates/validation/src/collection.rs
```diff
@@ -807,6 +849,116 @@ fn check_uuid_ptr_not_redacted(scope: Scope, spec: &schema::Schema, errors: &mut
...
+) {
+    let raw: serde_json::Value = match serde_json::from_str(model.get()) {
+        Ok(v) => v,
+        Err(_) => return, // Parse errors are reported elsewhere.
+    };
+    let Some(defs) = raw.get("$defs").and_then(|v| v.as_object()) else {
+        return;
+    };
+    for (def_id, def_value) in defs {
+        if !def_id.starts_with("flow://") {
```
jgraettinger: nit: it'd be more precise to inspect `$id` of the map value schema, which is the canonical URI of the definition and the active driver of $ref behavior -- technically, the key doesn't matter and can be swapped out with something else. We happen to use flow:// by convention.

crates/validation/src/collection.rs
```diff
@@ -807,6 +849,116 @@ fn check_uuid_ptr_not_redacted(scope: Scope, spec: &schema::Schema, errors: &mut
...
+            continue;
+        }
+        let mut findings = Vec::new();
+        find_keyword_in_json(def_value, "redact", &mut String::new(), &mut findings);
```
jgraettinger: nit: It feels weird to use `shape.locations()` above to search for `redact`, and a raw custom search over serde_json::Value here, to do essentially the same thing. Why not build the $defs sub-schema into a doc::Shape and search via locations()?

jgraettinger (approved):
LGTM % comments

crates/validation/src/collection.rs
```diff
@@ -807,6 +849,109 @@ fn check_uuid_ptr_not_redacted(scope: Scope, spec: &schema::Schema, errors: &mut
...
+// lost on the next publication. We discriminate by `$id` (the canonical URI
+// that drives `$ref` resolution), not by the `$defs` key (which is convention
+// only and can be renamed).
+fn check_redact_in_managed_defs(
+    scope: Scope,
+    model: &models::Schema,
+    curi: &url::Url,
+    errors: &mut tables::Errors,
+) {
+    const MANAGED_IDS: &[&str] = &[
```
jshearer, line 905: @jgraettinger WRT checking `$id`... did you mean like this? I didn't see a way to blanket check if an ID is one of these well-known URIs or not. Doing it this way vs checking for `flow://` does mean that this is one more place we have to remember to update

jgraettinger: I had meant only checking if `$id`'s value starts with `flow://`, rather than the $defs map key. I'm cool with this too though :shrug:

jgraettinger (approved):
Still LGTM

PR #2932, opened 2026-05-08
Docs: Update backfill image and other fixes

aeluce:
**Description:**

Collects a small assortment of docs fixes, including:
* Follow-up for #2922. Generalizes the backfill image to show destinations are refreshed rather than always dropped.
* Adds example YAML config for sync schedule usage.
* Clearly marks old HubSpot connector as deprecated.

Let me know if the backfill options image is clear now or if it could use any further tweaking.


site/docs/guides/customize-dataflows/materialization-sync-schedule.md
jwhartley: It's not something you changed in this PR, but reading this page I wanted to see that the non-fast sync time is 4 hours (not just 'more slowly') - could you update these three cells to say 'every 4 hours' instead of 'more slowly'? It's a question we get sometimes

jwhartley (approved):
LGTM % 1 optional request

site/docs/guides/customize-dataflows/materialization-sync-schedule.md
aeluce: Yep, good point. Specificity is better here.

PR #2933, opened 2026-05-08
docs/source-ga4-bigquery: New connector documentation

willdonnelly:
**Description:**

Adds new connector documentation for source-ga4-bigquery, a dedicated connector for GA4 BigQuery export datasets. This replaces the old "Google Analytics 4 BigQuery Native" capture documentation, which amounted to "here's how to set up the whole GA4->BigQuery linkage and how the BigQuery batch connector works, good luck!"

The problem with that approach is that it didn't work very well, and that's the reason we needed to implement this as a custom connector in the first place, so the old documentation should probably just be deleted.

Part of https://github.com/estuary/connectors/issues/3628


site/docs/reference/Connectors/capture-connectors/google-analytics-4-bigquery-exports.md
```diff
@@ -0,0 +1,170 @@
...
+3. Locate and select the **Google Analytics 4 BigQuery** connector.
+4. Enter a name and optional description for the capture task.
+5. Enter the Project ID for the project containing the GA4 export dataset, and provide the Service Account credentials. Optionally specify a Dataset to restrict discovery to a single dataset, or leave it unset to discover GA4 exports across every dataset in the project.
+6. Click **Next** and wait for the connector to discover available bindings. The connector emits one binding per (dataset, stream type) pair that has matching tables.
+7. Select the bindings you want to capture and adjust per-binding settings if needed.
+8. Once you are satisfied with your binding selection, click **Save and Publish**.
+
+## Usage
+
+The connector polls daily by default, at noon UTC. On each polling cycle it inspects the dataset's available daily tables and decides what to read:
```
aeluce:
```suggestion
The connector polls daily by default, at noon UTC. There is also an advanced option to set your own polling schedule. On each polling cycle it inspects the dataset's available daily tables and decides what to read:
```

aeluce: It may be worth explicitly calling out the polling schedule option here to let users know they can change the default.

aeluce (approved):
Thanks! One minor suggestion. And I updated the docs redirect to point to the new connector so anyone looking to use GA4 BigQuery using the old `/bigquery-ga4-native` link will get directed to the improved process (the `/google-analytics-4-bigquery-native` slug was in use for such a short time I don't think we need to add another redirect for it; it had only been changed to group it with other Google connectors).

Other than that, LGTM!

PR #2934, opened 2026-05-08
data-plane-controller: let vpc exist

qaoj:
**Description:**

Extremely small PR to let the `vpc` field be passed from Pulumi to Ansible.

**Workflow steps:**

Automatic. VPC populated from Pulumi.

N/A

**Notes for reviewers:**

N/A

### DIFF
--- a/crates/data-plane-controller/src/shared/stack.rs
+++ b/crates/data-plane-controller/src/shared/stack.rs
@@ -286,6 +286,8 @@ pub struct AnsibleHost {
     pub zone: String,
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub geo_region: Option<String>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub vpc: Option<String>,
 }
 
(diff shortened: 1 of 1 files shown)


crates/data-plane-controller/src/shared/stack.rs
```diff
@@ -286,6 +286,8 @@ pub struct AnsibleHost {
     pub zone: String,
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub geo_region: Option<String>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
```
jgraettinger, line 289: Seeing this diff, I'm wondering what `geo_region` is exactly and whether it plays the same role? If the vendor and region roll into geo_region, then the VPC will as well (because we create one VPC per region)

qaoj: That's a tailscale tag @skord added a while ago- they are not unique in any sense, they have values like "us" or "ap"

jgraettinger (approved):
LGTM

PR #2936, opened 2026-05-08
Authz refactor w/ orthogonal capabilities

GregorShear:
## Summary

Adds an orthogonal capability model that coexists with the legacy `read`/`write`/`admin` hierarchy. Existing authorization paths (PostgREST, transitive-roles BFS) are untouched. New GraphQL authorization checks can opt into orthogonal capabilities, and we'll migrate existing GraphQL checks over one at a time. Once the GraphQL API covers everything PostgREST does, PostgREST and the legacy capability path can be retired together.

## What changes

A new `OrthogonalCapability` enum and `role_grants.capabilities` / `user_grants.capabilities` columns let a grant carry an independent set of capabilities, rather than a single level in a hierarchy. This is finer-grained than the legacy roles: instead of `admin` implying everything, a grant lists exactly which capabilities it confers.

Special capabilities:

- `delegate` — a grant carrying `delegate` can propagate **its own** capabilities to the next hop. The next hop's effective set is `node.capabilities ∩ edge.capabilities` — you can only pass on capabilities you actually hold. Without `delegate`, the capabilities apply at the object, but cannot chain further.
- `assume` — a grant carrying `assume` is a **trust root**: the next hop inherits the full capability set declared on the edge, with no intersection against the parent's caps. Used when delegating complete authority (e.g. a user grant that says "this user fully impersonates `tenantA/groups/editors/`"), and as the BFS seed marker so user_grants get their declared capabilities through unfiltered.

In short: `delegate` carries your own permissions forward; `assume` carries the edge's permissions forward.

## Coexistence with legacy

`AnyCapability` wraps either a single legacy `Capability` or a `Vec<OrthogonalCapability>`. `RoleGrant::is_authorized` and `UserGrant::is_authorized` dispatch on the variant: the legacy arm runs the existing `transitive_roles` BFS unchanged; the orthogonal arm runs a new `reachable_nodes` BFS that respects the `delegate` / `assume` rules above. Call sites pick which model they want.

## Migration path

The two systems live side-by-side indefinitely. GraphQL authorization checks get migrated to orthogonal capabilities one at a time as we gain confidence. When the GraphQL API has full coverage of what PostgREST does today, PostgREST is retired and the legacy `capability` column / BFS can be dropped.

## Test plan

- [ ] `supabase db reset` applies cleanly
- [ ] `cargo sqlx prepare --workspace` is up to date
- [ ] `cargo check -p control-plane-api` and `cargo test -p tables` pass
- [ ] New unit tests cover `delegate` propagation, `assume` trust-root semantics, terminal nodes, multi-path capability union, and `RoleGrant` reachability
- [ ] Existing legacy authorization tests still pass


crates/tables/src/behaviors.rs
```diff
@@ -68,17 +68,57 @@ impl super::RoleGrant {
...
+                })
+            }
+            models::AnyCapability::Orthogonal(required) => {
+                if required.is_empty() {
+                    debug_assert!(
+                        false,
+                        "is_authorized called with empty orthogonal capabilities"
+                    );
+                    return false;
+                }
```
GregorShear: asking `is_authorized()`? with empty capabilities returns false

crates/tables/src/behaviors.rs
```diff
@@ -608,7 +827,671 @@ mod test {
...
+        );
+        assert_not_authorized(&rg, &ug, uid, "acmeCo/", vec![Read]);
+
+        // Assume does not add capabilities to the following edge
+        let (rg, ug, uid) = build_orthogonal_scenario(
+            vec![("acmeCo/", vec![Write, Assume])],
+            vec![("acmeCo/", "bobCo/shared/", vec![Read, Billing, TeamAdmin])],
+        );
+        assert_authorized(&rg, &ug, uid, "acmeCo/", vec![Write]);
+        assert_not_authorized(&rg, &ug, uid, "bobCo/shared/", vec![Write]);
```
GregorShear: note potentially non-obvious behavior

crates/control-plane-api/src/server/public/graphql/invite_links.rs
```diff
@@ -180,6 +180,12 @@ impl InviteLinksMutation {
         let env = ctx.data::<crate::Envelope>()?;
         let claims = env.claims()?;
 
+        if capability == models::Capability::None {
```
jshearer, line 183: I introduced the `None` capability in order to allow us to represent e.g a `Billing`-only user grant, but we shouldn't let people actually create invite links with that capability. This is unlikely to ever be an issue, and we're going to be refactoring this API to support bundles soon, but just in case.

jshearer:
Alright @GregorShear take a look at the most recent commit. A few worthy call-outs:

### This leaves all existing authorization checks alone

Every existing authorization call-site stays exactly as it was. `evaluate_names_authorization`, `is_authorized`, `authorized_prefixes`, and friends all continue to receive `models::Capability::Read`/`Write`/`Admin` from their callers, return the same errors, and behave identically against current grants. The only thing that changes underneath is what those calls do internally: each legacy capability now lowers into a `CapabilitySet` (an `EnumSet<authz::Capability>` of fine-grained bits) via `bits_for_legacy`, and the BFS evaluates over bits. So the call site says `Capability::Admin` and what actually gets checked is the Admin bundle's bit set — but no caller has to know that yet.

This means the commit doesn't migrate any enforcement site to fine-grained bits. That's intentional: a future PR can deliberately design new bits and move specific gates onto them (e.g., `CreateInviteLink` for invite link operations) without conflating that work with the model introduction here.

### `Capability::None` enables bundles-only grants

We added `None = 0` to the legacy `Capability` enum, mapped onto the renamed `none` value of the DB `grant_capability` enum (formerly the `x_00` placeholder, which already sorted below `read`). It exists for one reason: to allow a grant whose authorization comes entirely from the new `bundles` array column, with no legacy capability at all. A grant like `capability: none, bundles: [billing]` confers exactly the billing bits and nothing else.

### Authorization is multi-path union, not any-path

The previous `is_authorized()` for legacy capabilities used "any one path satisfies" semantics. With the unified BFS, both legacy and bit-shaped queries go through the same algorithm (multi-path bit union), and `bits_for_legacy` lowers the legacy input to a bit-shaped question. For current grants (`bundles=[]`), the two rules return identical answers, so the existing snapshot test passes unchanged.

They diverge once bundle-bearing grants exist. The reason is the algebra: `Bundle::Admin` is *defined* as `Editor | TeamAdmin | Billing`. Once Admin is the union of those bit sets, asking `is_authorized(_, Admin)` is asking "does the user's accumulated bit set cover Editor's ∪ TeamAdmin's ∪ Billing's bits?" Any-path evaluation answers a different question, "did a single grant happen to carry all of those bits at once," which the model has stopped using as a primitive.

The deeper point: the end state isn't "check that the user has Admin's bits at the call site," it's "the call site checks only the specific bits the operation requires." `create_invite_link` will eventually ask for `CreateInviteLink`, not for `Admin`. Once enforcement is bit-shaped at every gate, `Bundle::Admin` becomes a storage and UI affordance for how humans grant capabilities in administrative-role-sized chunks, not an evaluation predicate. Multi-path union is the evaluation rule that survives this transition.

### Naming

`Capability` is the right name for the fine-grained bit enum. In Rust we can lean into that and have both enums named `Capability`, namespaced under their respective modules: `models::Capability` for the legacy read/write/admin enum, `models::authz::Capability` for the bits. They never collide at use sites because we can fully qualify them.

Type names in GraphQL are a single flat namespace though, so I ended up keeping the `OrthogonalCapability` name for the bit enum on the GraphQL side, leaving the legacy `Capability` GraphQL type unchanged for the deployed UI.

The deployed UI's `CreateInviteLink` mutation references the legacy `Capability` GraphQL type by name (`$capability: Capability!`), so we can't rename it in the schema without coordinated UI work. We can rename it back to just `Capability` once that mutation is updated and the legacy enum is removed entirely.

crates/flow-client/control-plane-api.graphql
```diff
@@ -1081,24 +1082,16 @@ type MutationRoot {
...
-Unified capability enum covering both the legacy hierarchical capabilities
-(read/write/admin) and the new orthogonal capabilities (billing, team_admin).
-
-The legacy variants mirror `Capability` and exist so that authorization
-checks can be written against a single type. Today, read/write/admin are
-checked via BFS traversal of the old `capability` column, while
-billing/team_admin are checked via the `capabilities` array on
-`user_grants`. A future migration will consolidate both into the array.
-"""
 enum OrthogonalCapability {
```
GregorShear: I think we talked about "orthogonal" naming here to be temporary, yes?

crates/models/src/authz.rs
```diff
@@ -0,0 +1,77 @@
...
+        match self {
+            Self::Viewer => CatalogRead | JournalRead,
+            Self::Writer => Self::Viewer.capabilities() | JournalAppend,
+            Self::Editor => Self::Writer.capabilities() | SpecEdit | Delegate,
+            Self::Admin => {
+                Self::Editor.capabilities()
+                    | Self::TeamAdmin.capabilities()
+                    | Self::Billing.capabilities()
+            }
+            Self::Billing => EnumSet::empty(),
```
GregorShear, line 89: any reason not to add the `billing` capability above and use it here? or are you waiting to add it with your billing PR?

crates/models/src/authz.rs
```diff
@@ -0,0 +1,77 @@
...
+    Writer,
+    Editor,
+    Admin,
+    Billing,
+    TeamAdmin,
+    Delegate,
+    Assume,
+}
+
+impl Bundle {
```
GregorShear: i forget if we've talked about what the user-facing term would be - are you thinking of using "bundle" in the UI, too?

crates/tables/src/behaviors.rs
```diff
@@ -51,343 +53,224 @@ impl super::Import {
...
+    /// Returns each prefix reachable from `user_id` mapped to the union
+    /// of capability bits granted at that prefix across every path
+    /// through the grant graph, paired with the max legacy `capability`
+    /// column value among grants directly emitting that prefix.
+    ///
+    /// Bits compose additively (multi-path union); the legacy column is
+    /// a literal pass-through from storage, max'd across same-prefix
+    /// arrivals. Applying a min-capability filter to the bit set agrees
+    /// with `is_authorized` on the same inputs.
+    pub fn reachable_prefixes<'a>(
```
GregorShear, line 153: might be nice to omit redundant results here, particularly because the frontend will get this list via the prefixes query.

e.g. this'll return `acmeCo/ : [read,write]` and `acmeCo/child/ : [read]` if there are explicit grants to both of those prefixes

crates/models/src/authz.rs
```diff
@@ -0,0 +1,77 @@
...
+    pub fn capabilities(&self) -> CapabilitySet {
+        use Capability::*;
+        match self {
+            Self::Viewer => CatalogRead | JournalRead,
+            Self::Writer => Self::Viewer.capabilities() | JournalAppend,
+            Self::Editor => Self::Writer.capabilities() | SpecEdit | Delegate,
```
GregorShear: this is the place to add a big comment on the relationship between SpecEdit, Delegate, JournalAppend, etc... important for both humans and agents

might also be a reason to list the capabilities explicitly (instead of covered bundles) so you can comment that specEdit, delegate, journalAppend, and journalRead are related because "reasons" and then we're just adding catalogRead because it makes sense to but isn't inherently related in the same way as the others are

crates/flow-client/control-plane-api.graphql
jshearer: Ok so I renamed this `CapabilityBit` -- I had _wanted_ to rename the existing enum `LegacyCapability`, but `flowctl` also uses `Capability` in a graphql query and I didn't want to ship a new flowctl and wait for people to upgrade before merging/deploying this.

Yes, it's temporary: once everything stops referring to the existing `Capability`, we can re-introduce it as an alias of `CapabilityBit`, move everything over to that, and get rid of `CapabilityBit` nomenclature.

crates/models/src/authz.rs
jshearer: We talked about this -- I think we should probably call it something a bit more qualified in the GraphQL schema, like maybe `CapabilityBundle`? What we call it in the UI is a different question, and I still lean towards "role" despite that term being badly overloaded in the backend

jshearer: I renamed it to `CapabilityBundle`

crates/models/src/authz.rs
jshearer: Added

jshearer:
Ok @GregorShear, updated. I'll squash the fixup commits once we're ready to merge. Thoughts?

GregorShear (commented):
i've reviewed and done a bunch of e2e testing on my local. lgtm

jshearer (approved):
LGTM! We did some manual testing live as well to confirm that this doesn't break existing access, and it's all good. There's a substantial amount of UI work to migrate over to GraphQL, but all of the existing capability logic remains functional, so that migration work can happen at our leisure.

PR #2938, opened 2026-05-11
data-plane-controller: periodic converge

qaoj:
**Description:**

Historically, we've relied on `PulumiRefresh` to detect drift and trigger a convergence. This holds for AWS and GCP, but not Azure. Because of that, we need some way to ensure that converges are still happening regularly. This is a small change to allow that to happen. 4 hours was chosen because that seems to be the average amount of time the other providers go before some remote drift triggers a converge.

**Workflow steps:**

Automatic. Detects time since last successful converge for a plane and if greater than 4 hours, "signals" for the DPC to run a convergence on it.

N/A

**Notes for reviewers:**

N/A

### DIFF
--- a/crates/data-plane-controller/src/job/executor.rs
+++ b/crates/data-plane-controller/src/job/executor.rs
@@ -281,6 +281,11 @@ impl Executor {
         if state.last_refresh + REFRESH_INTERVAL < chrono::Utc::now() {
             state.pending_refresh = true;
         }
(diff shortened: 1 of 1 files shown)


mwillman-estuary:
LGTM, but this very much needs someone who knows more Rust than me to approve.

mdibaiee (approved):

PR #2944, opened 2026-05-14
Self-service private-link configuration

### LINKED ISSUE #2773: Allow customers to create privatelinks
**Background**
Currently privatelinks (AWS, Azure) and PSC (Google) are created by our solutions team manually for customers.  This is time consuming for us and them and customers.

**Proposal**
We should add a API, CLI command or capability in the UI to create them.  Whatever path we choose should be available to customers directly.

jshearer:
This PR exposes the `data_planes.private_links` configuration through the GraphQL API. Today private data-plane customers file a support request and we edit the column by hand.

## `dataPlanes` query: private-networking fields

* `privateLinks` returns the configured endpoints as a typed union (`AWSPrivateLink | AzurePrivateLink | GCPPrivateServiceConnect`).
* `awsLinkEndpoints` / `azureLinkEndpoints` / `gcpPscEndpoints` expose the provisioning results the data-plane controller writes back, as opaque JSON.
* All four fields are gated on the `ViewDataPlanePrivateNetworking` capability and fail closed to an empty list.

## `updateDataPlanePrivateLinks` mutation

* Takes `[PrivateLinkInput!]!`, a oneof input mirroring the union, and replaces the entire `private_links` array on a private data plane; partial updates are intentionally not supported. The DPC converges to the new configuration on its next poll.
* Requires `ModifyDataPlanePrivateNetworking` on the data-plane name, after a structural check that the name is under `ops/dp/private/<tenant>/`.

## Authorization model

* Two new capability bits: `ViewDataPlanePrivateNetworking` and `ModifyDataPlanePrivateNetworking`.
* The `Viewer` bundle (and therefore legacy `read`) carries the View bit: `read` on a data-plane prefix already conveys deploy-level trust, so viewing the plane's networking configuration comes with it.
* `Modify` reaches a tenant admin through the conjunction of two grants: the `ops/dp/private/<tenant>/` role_grant must carry the new `ManageDataPlane` bundle (`create_data_plane` installs it at provisioning time; a migration backfills existing private data planes), and the user's own bits at the tenant must include the bit, which today means `admin` since `Admin` folds `ManageDataPlane` in. Bits intersect along the delegation chain, so neither grant alone conveys `Modify`.
* To grant `ModifyDataPlanePrivateNetworking` to a non-admin, create a user_grant on the tenant carrying `{manage_data_plane, delegate}` (`Editor` already carries `Delegate`, so editors need only `manage_data_plane`).

## Shared models

* The `PrivateLink` types moved from `data-plane-controller::shared::stack` to `models`, with feature-gated async-graphql derives so one set of structs backs the DB column, the GraphQL schema, and the DPC (which re-exports them).
* Azure's `dns_name` / `resource_type` are now `Option<String>`. The previous shape was `String` with `skip_serializing_if = "String::is_empty"`, so `""` and absent were already indistinguishable on the wire: both serialized as a missing field. The new deserializer normalizes an incoming `""` to `None`, and serialization stays field-absent, so the stored bytes and the stack config est-dry-dock reads are identical for every historical row shape (absent, `""`, or a value); a round-trip test in `models::private_links` pins this.
* est-dry-dock cannot observe the difference either: its Pydantic models default these fields to `""` and use truthiness checks. The visible effect is confined to GraphQL, where the fields are nullable and rows that stored `""` surface as `null`.

## Reviewer notes

Trade-offs discussed during review, recorded so they don't need to be re-litigated:

* The four private-networking resolvers are deliberately not deduplicated further. The non-trivial shared piece, the capability check, is the `authorized()` helper in `graphql/mod.rs`; what remains per resolver is the gate call plus a clone-and-wrap body. A `gated_endpoints` helper bundling the gate and the JSON mapping was tried and dropped as trivial-body indirection under a vague name. Keeping the gate visible inside each resolver keeps the authorization audit local; the cost is that a change to the denial behavior must be edited in four places.
* Azure `Option<String>` was chosen over two alternatives: keeping `String` (zero glue, but the `""` sentinel becomes part of the public GraphQL contract as `String!`), and mapping `""` to `null` at the GraphQL layer only (DPC untouched, but it requires an Azure-only split of input and output types while the model keeps the sentinel). Normalizing in `models` costs a small serde adapter plus a round-trip test, and makes the nullable contract hold everywhere the type is used.
* `ViewDataPlanePrivateNetworking` is kept as its own bit even though `read` currently implies it via `Viewer`. The rejected alternative was to drop the bit and let DP-node visibility be the gate, which would have deleted the field gates entirely. Keeping the bit means each field gate names its capability explicitly, and the View / Modify split survives future changes to what `read` implies.

## Deployment note

Migrations are applied manually; the sequencing matters:

* Apply the enum migration before deploying: the new `create_data_plane` binds the `manage_data_plane` value, so provisioning a private data plane on the new code errors if the value is missing. (Two migration files because `alter type add value` cannot be referenced in the same transaction.)
* Run the backfill only after the rollout completes: binaries built before this PR decode `bundles` strictly in the snapshot loader, so any row carrying `manage_data_plane` breaks their `role_grants` fetch. For the same reason, do not provision a private data plane mid-rollout while old pods are live.
* In the window between deploy and backfill, tenant admins of existing private data planes can view private links but not modify them. The backfill is idempotent and skips rows that already carry the bundle, so it is safe to run any time after the rollout, and planes provisioned on the new code are unaffected by it.

Fixes https://github.com/estuary/flow/issues/2773


crates/control-plane-api/src/server/public/graphql/data_planes.rs
```diff
@@ -45,6 +46,116 @@ pub struct DataPlane {
...
+        )? {
+            return Ok(Vec::new());
+        }
+        self.raw_private_links
+            .iter()
+            .enumerate()
+            .map(|(idx, raw)| {
+                serde_json::from_value::<models::PrivateLink>(raw.clone()).map_err(|err| {
+                    async_graphql::Error::new(format!(
+                        "failed to parse private_links[{idx}] for data plane {}: {err}",
```
GregorShear, line 88: okay to fail the entire request if one of these links is bad? my interpretation is that a query for multiple data planes would fail if one dataplane had one bad private link. 

...seems unlikely to happen given type-checking on the write mutation, but i don't have a clear understanding of the stakes. how many things depend (or will depend in the future) on the query?

crates/control-plane-api/src/server/public/graphql/data_planes.rs
```diff
@@ -312,6 +441,79 @@ impl DataPlanesQuery {
...
+            )));
+        }
+
+        tracing::info!(
+            %data_plane_name,
+            link_count = private_links.len(),
+            %claims.sub,
+            "updated data plane private links",
+        );
+        Ok(true)
```
GregorShear: low conviction atm, thinking out loud here...

can we repeat the new state in the mutation response to at least give the client the option of merging the result into its cache? i understand the client will still need to poll in this instance (DPC writes back to the data plane row on the order of 60 seconds later or something?) 

but as a general pattern for our gql mutations i like the idea of repeating the new state back.

crates/control-plane-api/src/server/public/graphql/data_planes.rs
```diff
@@ -364,6 +566,429 @@ mod tests {
         insta::assert_json_snapshot!(response);
     }
 
+    #[sqlx::test(
```
GregorShear, line 580: consider testing for...

> 1. **Admin + role_grant lacking `manage_data_plane` → Modify denied.** This is the strongest ask. I confirmed the mechanics the feedback assumes: the BFS intersects bundle bits on *both* sides — the user_grant's effective bits and each role_grant edge's bits (`effective_bits(self.capability, &self.bundles) & delegatable` at `crates/tables/src/behaviors.rs:121` and `:204`). Every currently-authorized test path runs as Alice through the fixture's role_grant that already carries `{manage_data_plane}` (`private_links.sql:79`), so a regression where edge bundles got ignored would pass the whole suite. The test is ~10 lines: `UPDATE role_grants SET bundles = '{}'` before taking the snapshot, then assert Alice's mutation is denied while her `privateLinks` read still works (her `read` edge still carries `Viewer` → `ViewDataPlanePrivateNetworking`). This pins the exact backfill-window invariant the PR's deployment note relies on.
> 
> 2. **Non-admin delegation path.** Also confirmed feasible — `user_grants` got a `bundles` column in the same `20260511120000_orthogonal_capabilities.sql` migration. Insert bob with a user_grant on `aliceCo/` carrying `read` + `{manage_data_plane, delegate}`, and the existing fixture role_grant carries him to the dp prefix with Modify intact. This is the one path a real operator would actually use, and it has never executed.
>

crates/control-plane-api/src/server/public/graphql/invite_links.rs
```diff
@@ -53,6 +53,28 @@ pub struct InviteLinksQuery;
...
+        .iter()
+        .any(|p| catalog_prefix.starts_with(p))
+    {
+        return Err(async_graphql::Error::new(format!(
+            "{catalog_prefix} is an internal prefix and cannot be managed via invite links"
+        )));
+    }
+    Ok(())
+}
+
```
GregorShear: we should ban invite links (i.e. user grants) to `ops/` entirely, don't you think?

and my hunch is that we're already naturally protected from this action so long as no role_grants to `ops/` carry the "CreateInviteLink" capability (which i don't think any do... except for the estuary_support/ -> ops/ role grant

crates/control-plane-api/src/server/public/graphql/mod.rs
```diff
@@ -41,11 +41,46 @@ pub mod status;
...
+    Ok(tables::UserGrant::is_authorized(
+        &snapshot.role_grants,
+        &snapshot.user_grants,
+        env.claims()?.sub,
+        name,
+        capability,
+    ))
+}
+
+/// Errors unless the current user holds `capability` on `prefix`.
```
GregorShear: Love the substantive distinction, but we should talk about how to name these two functions (`authorized` vs `verify_authorization`) to make it clear when to use each

crates/models/src/private_links.rs
```diff
@@ -0,0 +1,168 @@
...
+{
+    let opt = Option::<String>::deserialize(deserializer)?;
+    Ok(opt.filter(|s| !s.is_empty()))
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
```
GregorShear, line 110: Leaning on agent review for tests here...

> `service_region`: `""` persists as-is. Unlike the Azure optionals, `service_region` has no empty-string normalization, so a client submitting an empty optional field stores `""`, which flows through DPC into est-dry-dock as the cross-region service region. Whether that breaks depends on Python falsiness in dry-dock. Suggest giving it the same deserialize_empty_string_as_none / is_none_or_empty treatment as the Azure fields, plus a round-trip test (none exists for this field).

crates/control-plane-api/src/server/public/graphql/data_planes.rs
jshearer: We should fail the whole request for a couple of reasons:
* This is indeed unlikely to happen, so if it _does_ happen (someone fat-fingered a direct DB update is the only way I can think of), we should fail as noisily as we can
* Because the only way we expose to update private-links is to post the entire list at once, if we silently filtered out malformed entries, we'd just silently overwrite them without knowing they ever existed.

crates/control-plane-api/src/server/public/graphql/data_planes.rs
jshearer: Yeahhh... also low conviction. I had it that way originally and it made the code more complex so I got rid of it since we're only ever going to return exactly what you submitted. But I think I can see a way to add it back appropriately.

crates/control-plane-api/src/server/public/graphql/invite_links.rs
jshearer: Yes.. but I ended up double-checking, and there's no pathway to creating an invite link to `ops/` (unless you're already an admin of `ops/`, which I believe either nobody should be, or just `estuary_support` while that role still exists) in the first place, so let's not muddy the waters unneccesarily.

crates/control-plane-api/src/server/public/graphql/mod.rs
jshearer: I ended up renaming `authorized` to `may_access`

crates/control-plane-api/src/server/public/graphql/data_planes.rs
jshearer: Added scenario 1 as `test_modify_denied_when_role_grant_lacks_manage_data_plane`: it clears the `manage_data_plane` bundle from the `aliceCo/ -> ops/dp/private/aliceCo/` role-grant edge, then asserts the `privateLinks` read still resolves while `updateDataPlanePrivateLinks` is denied, pinning the deploy-before-backfill state.

I left scenario 2 (non-admin via a `{manage_data_plane, delegate}` user_grant) out. It's feasible (the snapshot loader does read `user_grants.bundles`), but it pins no new invariant: the user_grant -> role_grant -> Modify chain at the private DP is already exercised by Alice's happy path, and bundle delegation/intersection through a role edge is already unit-tested in `behaviors.rs`.

crates/models/src/private_links.rs
jshearer: Done: `service_region` now gets the same treatment as the Azure optionals: `deserialize_empty_string_as_none` + `skip_serializing_if = "is_none_or_empty"`, with a round-trip test (`aws_service_region_round_trip`) covering absent / `""` / value, so all three optional string fields behave identically and a client-submitted `""` no longer persists.

GregorShear (approved):

PR #2945, opened 2026-05-14
doc/shape: add top-level contentMediaType to all shapes

mdibaiee:
**Description:**

- Previouly we had `content_type` for string shapes specifically (this matches JSONSchema's spec), which referred to JSONSchema's `contentMediaType`
- With this pull-request we are lifting `contentMediaType` to apply to all types, and the naming is now consistent with JSONSchema. This allows schemas such as:
```
{
  "type": "object",
  "contentMediaType": "application/vnd.estuary.psql.jsonb+json",
  "properties": { ... }
}
```

Which we intend to use for conveying more information about type of fields from captures to materializations (See example: https://github.com/estuary/connectors/pull/4400
- Regenerating the proto files using prost-build lends some formatting differences which are in a separate commit

**Workflow steps:**

**Notes for reviewers:**


crates/doc/src/shape/inference.rs
```diff
@@ -476,6 +476,13 @@ mod test {
                 - reduce: {strategy: firstWriteWins}
                 - redact: {strategy: sha256}
                 - default: john.doe@gmail.com
+                # `contentMediaType` is now a top-level Shape field, so the
+                # union of `{contentMediaType: x}` with `{type: null}` drops
```
jgraettinger: nit: this behavior looks right but the comment doesn't make sense to me.

Rather, when intersecting schemas, if one schema doesn't have an opinion (doesn't set contentMediaType), we allow a Some on another schema to take effect -- but then remove it if there's disagreement.

When unioning, we require all unioned schemas have the same Some to carry it forward into the union'd Shape.

jgraettinger (approved):
LGTM, small comment nit

PR #2946, opened 2026-05-14
docs: source-sage-intacct permission required for deletions

Alex-Bair:
**Description:**

Documentation updates for https://github.com/estuary/connectors/pull/4465. Mainly just specifying that the connector only captures deletions if it can query the `AUDITHISTORY` object.


nicolaslazo (approved):
👍

PR #2950, opened 2026-05-15
docs/redaction: document hash salting and redactSalt override

jwhartley:
## Summary

- Update the **Hash** strategy bullet to make clear that SHA-256 hashes are salted, and link to a new section explaining why.
- Add a **Hashing salt** section documenting that Estuary generates and manages a per-task salt automatically, with stable hashing across publications.
- Document the top-level `redactSalt` field on captures and derivations as the escape hatch for customers who need to share hashed values across tasks or supply their own salt.

## Why

A customer asked in Slack ([thread](https://estuaryworkspace.slack.com/archives/C08LJ60MDBQ/p1778857590338679)) whether Estuary supports only unsalted SHA-256 hashing, citing rainbow-table risk. The behavior in [#2383](https://github.com/estuary/flow/pull/2383) actually salts every hash and exposes a `redactSalt` override — but [features/redaction.md](https://docs.estuary.dev/features/redaction/) doesn't mention any of that today.

## Test plan

- [ ] Visual review on the deploy preview — confirm the new section renders, the in-page link from the **Hash** bullet works, and the YAML examples are readable.

### DIFF
--- a/site/docs/features/redaction.md
+++ b/site/docs/features/redaction.md
@@ -24,11 +24,13 @@ Redaction strategies include:
 
 * **Hash**
 
-   Replaces values with SHA-256 hashed versions of the values.
+   Replaces values with salted SHA-256 hashes of the values.
 
    Hashing fields can be useful when you would like to include stand-in values for sensitive fields in downstream systems but don't want those systems or system users to have access to the unhashed value.
 
    For example, hashing a user email so analysts can still compile information about a user journey without seeing the user's PII.
+
+   Estuary salts every hash to mitigate dictionary and rainbow-table attacks on low-entropy values such as emails or phone numbers. See [Hashing salt](#hashing-salt) for details on how the salt is managed and how to provide your own.
 
 ## How to Use Redaction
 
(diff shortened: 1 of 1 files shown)


site/docs/features/redaction.md
```diff
@@ -91,3 +93,43 @@ An example collection specification would therefore look like:
...
+  acmeCo/my-derived-collection:
+    schema: {...}
+    key: [/id]
+    derive:
+      using: {...}
+      transforms: [...]
+      redactSalt: "c29tZS1zZWNyZXQtc2FsdC12YWx1ZQ=="
+```
+
+When `redactSalt` is set on a specification, Estuary uses your value instead of the generated one. Treat the salt as sensitive — anyone who knows both the salt and a candidate plaintext value can compute the corresponding hash.
```
aeluce, line 135: It may be worth expanding slightly on "treat the salt as sensitive." I believe our `flowctl` auto-encryption is currently restricted to connector endpoint fields (or, at least, a quick test didn't automatically encrypt a new `redactSalt` field at the endpoint/binding level).

If we're not easily able to expand auto-encryption behavior to sensitive fields outside the endpoint config, we may want to direct users to manually encrypt with `sops`.

jwhartley: Hey @aeluce, I don't think either of these would be supported today, and I don't actually know of anybody specifying their own salt today. I'd say salt config encryption is a feature request to bring to eng when the ask comes up from a customer, and for now we leave it vague

aeluce: All right. Just seems difficult to treat a value as sensitive if it's stored as plain text. 🥲

aeluce (approved):

site/docs/features/redaction.md
jwhartley: Confirmed — not supported today. The `flowctl encrypt_configs` pass only walks endpoint configs and (for materializations) triggers, and the runtime reads `redact_salt.to_vec()` straight off the spec, so there's no decryption boundary either. Filed [estuary/flow#2963](https://github.com/estuary/flow/issues/2963) as a backlog feature request for when the ask comes up. Leaving the docs vague on encryption guidance for now.

PR #2951, opened 2026-05-15
docs: instruct users to clear min/max_backfill_xid after backfill

jwhartley:
## Summary
- Adds a step 7 to the PostgreSQL Capture backfill procedure in `backfilling-data.md` telling users to clear the Minimum/Maximum Backfill XID field after the backfill completes
- Explains the failure mode: the XID is only applied when a backfill is triggered, so it has no effect during normal CDC, but leaving it in place silently limits any future backfill (intentional or otherwise) to rows above/below the stale transaction ID — potentially missing data

## Context
Two customers in one day forgot about an old `min_backfill_xid` and triggered a dataflow reset backfill, only capturing a fraction of their data. The dedicated troubleshooting page (`postgres-replication-slot-recovery.md`) already has a "Clear the XMIN value" step, but the main `backfilling-data.md` guide stopped at "save and publish" with no removal instruction. This closes that gap.

## Test plan
- [ ] Docs preview renders the new step 7 correctly

### DIFF
--- a/site/docs/guides/backfilling-data.md
+++ b/site/docs/guides/backfilling-data.md
@@ -379,4 +379,6 @@ To configure this option:
 
 6. Save and publish your changes.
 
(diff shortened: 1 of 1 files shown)


SeanWhelan (approved):
LGTM, thanks!

PR #2952, opened 2026-05-16
runtime-v2 — E2E Go runtime wiring, runtime-sidecar, and observability

jgraettinger:
## Description

This branch completes the end-to-end wiring of the V2 materialization
runtime: the Go runtime controller
that drives the V2 shard RPC lifecycle, the `runtime-sidecar` process that
hosts the Shuffle and Shuffle Leader gRPC services, and a `service-kit`
observability crate backing both.

The new code ships **deployed inert**: no task uses it unless the per-task
`enable-runtime-v2` feature flag is set on its shard labels.

**Status:** functional on local stacks with initial testing workloads.
**Not yet production-ready** — needs further QA before production tasks
are migrated.

### What's included

- **Go runtime wiring** — the flagged V2 materialization app and
  controller, which drive the `Join`/`Joined` => `Opened` shard RPC
  lifecycle; the CGO `TaskService` v2 entry point; a `RestoreCheckpoint`
  guard that restarts a shard when its `enable-runtime-v2` flag is toggled
  mid-run; and lazy `aws-lc-rs` rustls install for the CGO path.
- **runtime-sidecar** — `crates/runtime-sidecar/`, the per-machine Rust
  process hosting the Shuffle and Shuffle Leader gRPC services, plus its
  systemd unit and local data-plane wiring (`base_port+60`).
- **Observability** — `crates/service-kit/`, a service-agnostic leaf crate
  (handler `Registry`, loopback admin surface, runtime trace-level
  control, `event!` macro, Prometheus metrics), wired into both sidecar
  gRPC services.
- **Publishing & auth** — split publisher `Binding` into `Mapped` and
  `Fixed` variants; grant V2 journal clients a joint `APPEND|APPLY|LIST`
  capability for on-demand partition creation.
- **Frontier pruning & V1↔V2 reconciliation** — conservative pruning of
  stale committed-frontier entries during recovery scan; a cutover cleanup
  `Persist` that reconciles V1's complete checkpoint against V2's delta
  `FC:` keys; rollback-marker stripping; and the `drop-runtime-v1-rollback`
  flag.
- **Smaller fixes** — `runtime.proto` cleanups, leader close-policy and
  stats-orientation fixes, a gazette `file://` fragment proxy fix that
  was hot-looping local stacks, and assorted serde/label fixes.
- **Local stack & test infra** — `dekaf-e2e` test-tenant onboarding,
  `FLOW_AUTH_TOKEN` unification, and new local-stack / preview-harness
  docs.

See individual commit messages for per-change rationale.


crates/runtime-next/src/leader/materialize/fsm.rs
```diff
@@ -554,16 +596,16 @@ impl HeadPersist {
...
             },
         )) = shard_rx
-            && *nonce == self.nonce
+            && *seq_no == self.seq_no
         {
             shard_rx.take();
 
             let Self {
-                nonce: _,
+                seq_no: _,
```
williamhbaker, line 608: Can't comment on the line specifically, but down on line 689 - `if task.triggers.is_some()` the triggers fire, even on empty transactions. It was a comment from the previous runtime-next PR.

crates/runtime/src/materialize/protocol.rs
```diff
@@ -590,6 +590,17 @@ pub fn recv_client_start_commit(
...
+    // The V2 leader stamps a synthetic "committed-close" source into the
+    // consumer.Checkpoint on each commit, recording the V2 RocksDB epoch.
+    // If V1 inherits a checkpoint from a prior V2 run, the marker is
+    // preserved verbatim across V1 commits. A subsequent V2 rollforward
+    // would then mistake the stale marker for an in-sync RocksDB state
+    // and ignore the legacy_checkpoint, resuming from V2's stale frontier
+    // and re-processing whatever V1 had advanced past. Strip the marker
+    // so V2 startup treats V1's advanced sources as authoritative.
+    let mut runtime_checkpoint = runtime_checkpoint.clone();
+    runtime_checkpoint.sources.remove("committed-close");
```
williamhbaker, line 601: Should this strip be applied to the returned `request` as well?

I'm thinking about a scenario where a V2-flagged task reverts to a V1 and hits this stripping, but then the unstripped checkpoint from `request` is persisted in a remote-authoritative materialization. Then if the task crashes at an inopportune time when it is flipped back to V2, I think we could end up re-reading data based on the unstripped field from the connector recovered checkpoint?

This seems like a pretty unlikely case but the fix at least on the surface seems like it would be simple.

crates/gazette/src/journal/append.rs
```diff
@@ -67,16 +69,19 @@ impl Client {
...
                     // It's technically possible to get an empty set of bytes when reading
                     // from the input stream. Filter these out as otherwise they would look
                     // like EOFs to the append RPC and cause confusion.
                     Ok(content) if content.len() == 0 => None,
-                    Ok(content) => Some(Ok(broker::AppendRequest {
-                        content,
-                        ..Default::default()
-                    })),
+                    Ok(content) => {
+                        metrics.append.increment(content.len() as u64);
```
williamhbaker, line 79: I think this will over-count the metric on retries. Unless that is intentional...maybe we could increment the counter after the append has succeeded, a little further down?

go/runtime/materialize_v2.go
```diff
@@ -0,0 +1,418 @@
...
+		}
+		var join, rev, err = m.buildJoin()
+		if err != nil {
+			return fmt.Errorf("building Join: %w", err)
+		}
+		if join == nil {
+			waitForRevision = rev + 1
+			continue
+		}
+		if err := m.client.Send(&pr.Materialize{Join: join}); err != nil {
```
williamhbaker: In the V1 runtime, we have [this doSend](https://github.com/estuary/flow/blob/5b215cfe9f491fcb49d6942b0a922e9bfeb59b43/go/runtime/task.go#L366-L385) helper for checking `io.EOF` and then reading the actual error.

Do we need to do that here too, to avoid returning the less useful `io.EOF` error instead of the real causal one? Same applies to the Task send below at L223.

williamhbaker (approved):
LGTM! Just a few comments.

crates/runtime-next/src/leader/materialize/fsm.rs
jgraettinger: Done -- not sure what happened there. I double checked that all other feedback from the last PR is still accounted for.

crates/gazette/src/journal/append.rs
jgraettinger: It was something I thought about, and left this way intentionally.

The reason is that we get transactional read & write volume through our reporting. We have non-transactional reads through other `gazette` metrics, which is important because it highlights read rate undertaken by doomed transactions (as well as recovery logs).

This is similarly an operational measure rather than a reporting measure. We still need to know the volume of data _attempted_ to be appended, even if it failed. Generally they're 1:1, and reporting also tells us transactional collection append volume.

(Also, tracking the total amount of byte volume to do a deferred increment only on success is actually quite tricky here, given how ownership passes into the stream).

crates/runtime/src/materialize/protocol.rs
jgraettinger: Yes, it definitely should. Great catch, thanks.

go/runtime/materialize_v2.go
jgraettinger: We don't want to use doSend directly, because it races a Recv() with the receive pump, but tracing through here I'm thinking we should just ignore Send errors entirely and treat them as infallible, because we'll get the causal error a bit later in the protocol.

jgraettinger (commented):
~PTAL~ ~I borked this up with a rebase, fixing...~ Fixed, PTAL

williamhbaker (approved):
LGTM

PR #2955, opened 2026-05-18
ops-catalog: materialize stats to Bigtable

williamhbaker:
**Description:**

- Create a minimalist Bigtable rust read client for reading catalog stats. As of these changes it's not wired up to anything, but it was developed in conjunction with prototyping various other usages of catalog stats in the code base: Data movement stalled alerts & abandoned task detection (control plane agent), OpenMetrics API, billing calculations, and speculative UI capabilities.
- Run a Bigtable emulator locally for local stats, and materialize stats into it in parallel with the Postgres materialization. Some basic instructions are included for verifying the two databases are equivalent.
- Add the new L2 stats derivation (now partitioned by grain) to the `update_l2_reporting` control plane API. This is needed for the local Bigtable stats materialization to work. Once the control plane API is deployed on this new code, it will also cause it to publish the new L2 derivation in production.

Per the above, prior to deploying the new control plane API on this code, we should create the new reporting-dedicated data plane. Then just after the control plane API is deployed, call the `update_l2_reporting` endpoint with the default data plane parameter set to that reporting data plane so that the new L2 derivation is created there.

**Workflow steps:**

**Notes for reviewers:**


crates/catalog-stats/src/lib.rs
```diff
@@ -0,0 +1,1027 @@
...
+    ) -> impl futures_core::Stream<Item = anyhow::Result<CatalogStats>> + '_ {
+        let row_set = bt::RowSet {
+            row_keys: names.iter().map(|name| pack_row_key(name, ts)).collect(),
+            row_ranges: vec![],
+        };
+
+        self.read_rows(grain, row_set, vec![])
+    }
+
+    /// Streams every `catalog_stats_<grain>` row whose `catalog_name` stats
```
jgraettinger:
```suggestion
    /// Streams every `catalog_stats_<grain>` row whose `catalog_name` starts
```

crates/catalog-stats/src/lib.rs
```diff
@@ -0,0 +1,1027 @@
...
+
+                while let Some(res) = stream.try_next().await? {
+                    match res {
+                        ReadResult::Yield(stats_doc) => {
+                            // Hand off the row; success resets the retry budget.
+                            () = co.yield_(stats_doc).await;
+                            retries = 0;
+                        }
+                        ReadResult::Retry(status) => {
+                            if retries >= MAX_RETRIES {
```
jgraettinger: Consider using the `gazette` crate strategy of having the caller be responsible for retry policy, with a wrapping result type that indicates number of attempts: https://github.com/estuary/flow/blob/master/crates/gazette/src/journal/read/mod.rs#L55-L58
This also lets the caller decide what to do with logging of the actual error.

crates/catalog-stats/src/lib.rs
```diff
@@ -0,0 +1,1027 @@
...
+            self.project, self.instance,
+        )
+    }
+
+    fn read_rows(
+        &self,
+        grain: Grain,
+        row_set: bt::RowSet,
+        additional_filters: Vec<bt::RowFilter>,
+    ) -> impl futures_core::Stream<Item = anyhow::Result<CatalogStats>> + '_ {
```
jgraettinger: Crates like `gazette`, `tokens`, and the V2 runtime use tonic::Result over anyhow::Result consistently for all streamed results and IPC (the V1 runtime does as well now, IIRC, with a couple of exceptions like derive-sqlite).
I've found over time that tonic::Status is nicer to work with as a fundamental type for protocol errors -- for one, it's Clone, where anyhow::Error isn't.
The rule of thumb I've adhered to is anyhow::Result for errors _within_ a current application call stack, converted to/from tonic::Status when communicating that error from/to other systems and inter-process services.

crates/catalog-stats/src/lib.rs
```diff
@@ -0,0 +1,1027 @@
...
+    Done,
+}
+
+impl ReadRows {
+    fn new(
+        table_name: String,
+        row_set: bt::RowSet,
+        additional_filters: Vec<bt::RowFilter>,
+    ) -> Self {
+        // `CellsPerColumnLimitFilter(1)`: materialize-bigtable writes
```
jgraettinger, line 362: Oh, fantastic that it has MVCC built in! So we don't even need to represent multiple versions ourselves...

crates/control-plane-api/src/server/update_l2_reporting.rs
```diff
@@ -115,15 +132,31 @@ pub async fn update_l2_reporting(
         );
     };
 
+    let models::DeriveUsing::Typescript(models::DeriveUsingTypescript {
```
jgraettinger, line 135: Good, I'm glad this is using TypeScript (I was thinking about it the other day and couldn't recall which derivation runtime it used).

TypeScript should be easy(*) to scale using the V2 runtime. We'll be able to support SQLite derivations in V2 and run them as we do today, but not scale them out for the time being (until we meaningfully tackle block devices).

jgraettinger (approved):

crates/catalog-stats/src/lib.rs
williamhbaker: 👍 that makes sense to do here as well I think. Updated in the latest commit.

crates/catalog-stats/src/lib.rs
williamhbaker: Addressed with the above caller retry policy refactor.

jgraettinger (approved):
Still LGTM

As a call-out, idiomatic Rust would say to use `crate::Error` variants with `thiserror` to represent the fine-grain underlying errors that can happen, and to not have `anyhow` as a dependency because this is a library crate and not a top-level binary crate. I'm fine with it as-is and that can be tackled later as it's integrated and actually used :shrug:

PR #2959, opened 2026-05-20
protocols/catalog: skip NULL-spec rows in LoadAll* queries

SeanWhelan:
**Description:**

`flowctl catalog delete --name <test>` (and any publication that deletes a test) currently fails with:
```
test:N> FATA fatal error err="extracting from build: validating spec : missing Name"
Error: failed with status: testFailed
```

The Rust build layer represents a deletion as a `built_*` row with `spec` NULL (per `BuiltRow::is_delete` in [built.rs](https://github.com/estuary/flow/blob/8553fa84846058d449726012842cacb707209eec/crates/tables/src/built.rs#L64-L66)). The Go-side `LoadAll*` queries in [build_load.go](https://github.com/estuary/flow/blob/8553fa84846058d449726012842cacb707209eec/go/protocols/catalog/build_load.go) did not filter those rows, so `flowctl-go api test` would `Scan` a NULL `spec` into an empty byte slice, `Unmarshal` it into a zero-value `TestSpec`, and trip `TestSpec.Validate()` with `missing Name`.

Fix: add `WHERE spec IS NOT NULL` to all four LoadAll* queries (collections, captures, materializations, tests). The same defect was latent in the other three loaders even though the visible failure was only in tests, since the test phase is the one that per-row-validates inside loadSpecs.

**Workflow steps:**

Repo before fix:
```
cat > flow.yaml <<'YAML'
collections:
  acmeCo/widgets:
    schema: { type: object, properties: { id: { type: string } }, required: [id] }
    key: [/id]
tests:
  acmeCo/widget-test:
    - ingest: { collection: acmeCo/widgets, documents: [{ id: "w1" }] }
    - verify: { collection: acmeCo/widgets, documents: [{ id: "w1" }] }
YAML
flowctl catalog publish --source flow.yaml --auto-approve
flowctl catalog delete --name acmeCo/widget-test
# → testFailed: validating spec : missing Name
```

**Notes for reviewers:**

### DIFF
--- a/go/protocols/catalog/build_load.go
+++ b/go/protocols/catalog/build_load.go
@@ -64,11 +64,13 @@ func LoadAllErrors(db *sql.DB) ([]Error, error) {
 	return out, err
 }
 
-// LoadAllCollections loads all collections.
+// LoadAllCollections loads all collections, excluding tombstone rows
+// whose spec is NULL because the collection is being deleted by the
+// current publication.
 func LoadAllCollections(db *sql.DB) ([]*pf.CollectionSpec, error) {
 	var out []*pf.CollectionSpec
 	var err = loadSpecs(db,
-		`SELECT spec FROM built_collections ORDER BY collection ASC;`,
+		`SELECT spec FROM built_collections WHERE spec IS NOT NULL ORDER BY collection ASC;`,
 		func() loadableSpec { return new(pf.CollectionSpec) },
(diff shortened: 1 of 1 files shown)


jgraettinger (approved):
LGTM

As a nit, LLMs tend to overly-extend comments to document the conditions of a change, including discussion of historical conditions that are no longer true, or handling which is obvious just by reading the code itself. In this case, I would have left the function comments as they are, and just added the WHERE guard.

PR #2965, opened 2026-05-22
Docs: Refresh schema evolution info

aeluce:
**Description:**

Different updates have streamlined schema evolution so there aren't as many opportunities for schemas to break. This PR includes docs refreshes to account for the following:

* Position automatic discovery and inference as the default to make it more clear that Estuary handles a lot of common evolution scenarios automatically.
* Changing the collection key is no longer a truly incompatible update for collections. Group-by keys push handling down to materializations.
* We no longer increment collections (automatically suffixing with "_v2"); if schema changes are truly incompatible, dataflow resets allow collections to be refreshed. This is especially noted for changes to collection logical partitions.

Mainly updates schema evolution guide and concepts pages

**Notes for reviewers:**

Thanks for reviewing! Suggestions/comments welcome.


site/docs/concepts/advanced/evolutions.md
```diff
@@ -32,10 +32,11 @@ Each collection and its data are defined by a **collection specification**, or s
 The spec serves as a formal contract between the capture and the materialization, ensuring that data is correctly shaped and moves through the Data Flow without error.
 
 The spec includes the collection's `key`, its `schema`, and [logical partitions](./projections.md#logical-partitions) of the collection, if any.
+Any of these parts may change:
 
-When any of these parts change, any capture or materialization writing to or reading from the collection must be updated to approve of the change, otherwise, the Data Flow will fail with an error.
-
-You can use Estuary's **schema evolutions** feature to quickly and simultaneously update other parts of a Data Flow so you're able to re-start it without error when you introduce a collection change.
+* When a collection **key** changes, you can separately manage materialization [group-by keys](/guides/customize-materialization-fields/#group-by-keys) during field selection.
```
jwhartley, line 37: This sentence feels a bit random - I'd still mention that collection key changes will trigger an automatic backfill with default settings (onIncompatibleSchemaChange: backfill) unless you have specifically set a materialization group by

site/docs/concepts/advanced/evolutions.md
```diff
@@ -84,33 +75,11 @@ This field can be set at the top level of a materialization spec or within each
...
-  properties:
-    id: { type: integer }
-    foo: { type: string, format: date-time }
-  required: [id]
-key: [/id]
-```
-
-If you materialized that collection into a relational database table, the table would look something like `my_table (id integer primary key, foo timestamptz)`.
-
-Now, say you edit the collection spec to remove `format: date-time` from `foo`. You'd expect the materialized database table to then look like `(id integer primary key, foo text)`. But since the column type of `foo` has changed, this will fail. An easy solution in this case would be to change the name of the table that the collection is materialized into. Evolutions do this by appending a suffix to the original table name. In this case, you'd end up with `my_table_v2 (id integer primary key, foo text)`.
```
jwhartley, line 116: Aren't there still cases where we can't automatically migrate type changes, and thus will do an automatic backfill? e.g. a column type narrows from string to integer.

Wondering why you've removed this section

site/docs/concepts/captures.md
```diff
@@ -89,7 +89,7 @@ The toggles in the web app correspond directly to the properties above:
 
 - "Automatically keep schemas up to date" enables `autoDiscover`
 - "Automatically add new collections" corresponds to `addNewBindings`
-- "Breaking changes re-versions collections" corresponds to `evolveIncompatibleCollections`
+- "Changing primary keys re-versions collections" corresponds to `evolveIncompatibleCollections`
```
jwhartley, line 92: I believe we just do a collection reset rather than re-version (v2, v3, etc) now

site/docs/guides/schema-evolution.md
```diff
@@ -1,114 +1,94 @@
...
+This comprises several related functions:
+* Automatically keep schemas up to date
+* Automatically add new collections
+* Changing primary keys resets collections
+
+Estuary accounts for incompatible schema changes (such as changes to a column's data type) by _widening_ the schema.
 
 :::info
-There are a variety of reasons why these properties may change, and also different mechanisms for detecting changes in source data. In general, it doesn't matter why the collection spec has changed, only _what_ has changed. However, [AutoDiscovers](../concepts/captures.md#automatically-update-captures) are able to handle some of these scenarios automatically. Where applicable, AutoDiscover behavior will be called out under each section.
+Automatic schema evolution will only ever widen data types.
```
jwhartley, line 38: Useful to mention this is to ensure historical data is still valid under the evolved schema?

site/docs/guides/schema-evolution.md
```diff
@@ -1,114 +1,94 @@
...
-        image: ghcr.io/estuary/source-postgres:v1
-        config: encrypted-pg-config.sops.yaml
-    bindings:
-      - resource:
-          namespace: public
-          stream: anvils
-          mode: Normal
-        backfill: 1
-        target: acmeCo/inventory/anvils_v2
+Note that materializations can specify separate [group-by keys](/guides/customize-materialization-fields/#group-by-keys) from the collection key structure.
```
jwhartley, line 76: If there's a source key change coming, you can:
1) preventatively set the materialization group by to avoid propagating key changes
2) set the group by after the fact to migrate back to the previous/original key
3) do nothing and let us change the key with a dataflow reset

jwhartley: With ^, this seems backwards: "Ensure the materialization group-by key is updated if you wish to propogate key changes downstream."

site/docs/guides/schema-evolution.md
jwhartley: edit: I see this is mentioned below:
"This works because the type is broadened, so existing values will still validate against the new schema. "

jwhartley (commented):
Had a couple of questions here @aeluce

jwhartley:

as-is. For this PR:

- `site/docs/concepts/advanced/evolutions.md` becomes `content/concepts/advanced/evolutions.md`
- `site/docs/concepts/materialization/materialization.md` becomes `content/concepts/materialization/materialization.md`
- `site/docs/guides/schema-evolution.md` becomes `content/guides/schema-evolution.md`

The three images move with them, same relative paths under `content/`.

aeluce:
Addressed in https://github.com/estuary/docs/pull/55

PR #2975, opened 2026-05-26
runtime-v2: implement captures and derivations

jgraettinger:
Materializations already run on the Rust-driven `runtime-next`; this branch
adds the remaining two task types — **captures** and **derivations** — plus
the protocol additions, publishing primitives, and shared leader refactors
they need.

As before, everything ships **deployed inert**: nothing drives the new paths
until a shard carries `estuary.dev/flag/enable-runtime-v2`.
`FlowConsumer.NewStore` routes all three task types to their V2 application
when the flag is set, and the V1 apps fail their shard if the flag is flipped
on under them, forcing a restart that re-selects the runtime.

## Summarized Changes

- **Captures** — a leaderless model: each shard runs an independent
  transaction loop with its own connector, RocksDB, and recovery log, and
  reuses the single `Persist` path by synthesizing `Persist` locally.
- **Derivations** — reuse the full Shuffle Leader machinery, with one
  structural difference from materializations: every shard publishes
  documents, so a shard picks its own producer plus a leader producer that
  shard zero forwards. derive-sqlite runs in-process as a
  remote-authoritative connector on an ephemeral shard-zero RocksDB, with V1
  rollback retained.
- **Protocol** — state-carrying capture Apply; iterative multi-shard derive
  Flush (`state_patches_json` / `more`); public `Opened.runtime_checkpoint`;
  the `Shard.Capture` / `Shard.Derive` RPCs; `Transform.state_key`; and the
  materialize `state_patches_json` rename so both task types name the concept
  identically. derive-typescript/python codegen is now V2-only (existing V1 `:dev` tags left alone).
- **Shared refactors** — owned-document publishing (`enqueue_owned`), the
  `leader::close_policy` extraction, HeadIdle/HeadExtend unification across
  all three FSMs, `frontier_mapping` promoted to `leader/`, and explicit
  producer selection carried on `Task`.
- **flowctl** — `raw preview-next` now hosts and drives captures and
  derivations in-process for end-to-end exercise.
- **Local/QA** — per-tenant `test-tenant-<tenant>.env` files, and
  `local:test-tenant` waits for a public data plane to register before
  onboarding.

## Rollout

The complete runtime ships alongside the legacy one
with no tasks using it. Migration is per-task via the `enable-runtime-v2`
shard label, with rollback by flipping the flag back; the legacy
`consumer.Checkpoint` is maintained until a task sets
`drop-runtime-v1-rollback`.

## Testing

* New unit / integration / fuzz tests.
* Initial, non-exhaustive QA using both `flowctl raw preview-next` and a local stack.


crates/runtime-next/src/publish.rs
```diff
@@ -140,3 +210,46 @@ impl Publisher {
         }
     }
 }
+
+/// Patch a document UUID placeholder in-place after the publisher has assigned
+/// the transaction UUID.
+fn patch_document_uuid(doc: &mut doc::OwnedNode, uuid_ptr: &json::Pointer, uuid: uuid::Uuid) {
+    if uuid_ptr.0.is_empty() {
```
williamhbaker, line 217: This should be impossible right? I think we'd want to fail with an error here instead.

crates/runtime-next/src/publish.rs
```diff
@@ -140,3 +210,46 @@ impl Publisher {
...
+
+    // SAFETY: We have sole ownership of doc::OwnedNode.
+    let cell = unsafe {
+        std::slice::from_raw_parts_mut(
+            cell.start as *mut u8,
+            cell.end.offset_from_unsigned(cell.start),
+        )
+    };
+    if cell.len() != ::uuid::fmt::Hyphenated::LENGTH {
+        return; // Return, rather than panic-ing if the length is somehow wrong.
```
williamhbaker, line 252: Similar to the comment above...I think we'd want to fail noisily here.

crates/runtime-next/src/leader/derive/fsm.rs
```diff
@@ -0,0 +1,1501 @@
...
+        let Self {
+            extents,
+            publisher_commits,
+            ..
+        } = self;
+
+        // Fold deltas from the extents Frontier into per-binding "bytes behind" gauges.
+        for jf in &extents.frontier.journals {
+            let Some(entry) = binding_bytes_behind.get_mut(jf.binding as usize) else {
+                continue; // Reachable if shuffle service reports invalid binding indices.
```
williamhbaker, line 623: Should we error here instead?

crates/runtime-next/src/leader/derive/fsm.rs
```diff
@@ -0,0 +1,1501 @@
...
+    task: &Task,
+    extents: &Extents,
+    binding_bytes_behind: &[i64],
+) -> ops::proto::Stats {
+    let mut transforms: BTreeMap<String, ops::proto::stats::derive::Transform> = BTreeMap::new();
+    let mut max_published_at: uuid::Clock = uuid::Clock::default();
+
+    for (binding_index, binding) in &extents.bindings {
+        let Some(transform_name) = task.binding_transform_names.get(*binding_index as usize) else {
+            continue; // Reachable if shards report invalid binding indices.
```
williamhbaker, line 845: The `continue` here and the on line 849 also seem like they should be errors instead of silent fallbacks.

Also may apply to materialize FSM, around here: https://github.com/estuary/flow/blob/master/crates/runtime-next/src/leader/materialize/fsm.rs#L1252

crates/runtime-next/src/leader/derive/fsm.rs
williamhbaker: Similar thing in the materialize FSM, that I had not considered before: https://github.com/estuary/flow/blob/8f1a3e4676c69dff2a729f2d6d9554117c2a9c41/crates/runtime-next/src/leader/materialize/fsm.rs#L683

crates/runtime-next/src/leader/materialize/fsm.rs
```diff
@@ -257,88 +249,132 @@ pub struct HeadIdle {
...
+            close_requested: *close_requested,
+            idempotent_replay: self.idempotent_replay,
+            last_age: uuid::Clock::delta(now, self.last_close),
+            combiner_bytes,
+            open_age,
+            read_bytes,
+            read_docs,
+            stopping,
+            tail_done,
+            unresolved_hints: self.extents.frontier.unresolved_hints != 0,
```
williamhbaker, line 299: At first when a task restarts after crashing mid-commit it may be in `idempotent_replay` mode. But the FSM `HeadIdle` [is built with default extents](https://github.com/estuary/flow/pull/2975/changes#diff-b7fc2d56ab1ed8cf0031eacf6b3e73cc926dd7466914cacb38da76c1b4b1b69fR205). So on the first tick of the FSM, I believe `unresolved_hints` will always be false. Then, the close policy is evaluated on `idempotent_replay && !unresolved_hints`. So `may_extend` will always be `false` in that case, and we'd end up returning `(Action::Idle, Head::Idle(self))` on line 321.

This seems like it would end up in a loop, with the actor loop calling step again with the same inputs?

go/runtime/derive.go
```diff
@@ -136,7 +146,14 @@ func (d *deriveApp) RestoreCheckpoint(shard consumer.Shard) (_ pf.Checkpoint, _e
 	var openedExt = pr.FromInternal[pr.DeriveResponseExt](opened.Internal)
 	d.container.Store(openedExt.Container)
 
-	return *openedExt.Opened.RuntimeCheckpoint, nil
+	if opened.Opened == nil {
+		return pf.Checkpoint{}, fmt.Errorf("expected Opened, but got %#v", opened)
+	}
+	var runtimeCheckpoint = opened.Opened.RuntimeCheckpoint
+	if runtimeCheckpoint == nil {
```
williamhbaker, line 153: Is this guard necessary? On first glance it looks wrong to not return an error like the `open.Opened` one above. It seems impossible to have this be `nil` per `recv_connector_opened` in `crates/runtime/src/derive/protocol.rs` so maybe it would be fine to panic? That is still how `go/runtime/capture.go` and `go/runtime/materialize` are set up in their versions of `RestoreCheckpoint`.

williamhbaker (approved):
LGTM % a few comments

go/runtime/derive.go
jgraettinger: No, this is cruft -- removing.

crates/runtime-next/src/publish.rs
jgraettinger: It _should_ be but the value comes from the control plane. There's nothing structurally unsafe about an empty pointer, though, so removing this guard (it'll fail later).

crates/runtime-next/src/publish.rs
jgraettinger: I updated the publisher interface to allow this routine to return a tonic::Result, and threaded in errors throughout.

crates/runtime-next/src/leader/derive/fsm.rs
jgraettinger: Fixing these by introducing an Action::Error variant which can carry / surface proper errors.

jgraettinger (commented):
Thanks! Github's UI is spazzing on my reply to the FSM comment, but indeed, that was a great catch. Fixing that too.

I'll land this, given its size and because it's inert today, and bring fixes in the next PR.

PR #2976, opened 2026-05-26
Improve Docker daemon error for flowctl runtime

### LINKED ISSUE #2969: flowctl: clearer error when Docker daemon isn't running
## Problem

When the Docker daemon isn't running, `flowctl preview` and `flowctl catalog test` (against TS / Python derivations, captures, or materializations) fail with a misleading error that looks like a corrupt image:

```
failed to parse `docker inspect` output: []
```

This points the user in the wrong direction (image / network / spec) when the actual cause is that Docker isn't running.

## Root cause

`crates/runtime/src/container.rs:379-394` (`docker_cmd`) treats only non-zero exit status as a failure. With Docker Desktop stopped on macOS, every docker subcommand exits **0** and writes its real error to stderr:

```
Cannot connect to the Docker daemon at unix:///Users/.../docker.sock. Is the docker daemon running?
```

So `docker_pull()` and `docker inspect` "succeed", execution reaches `parse_image_inspection(&[])` (line 561), and serde fails to destructure the empty array — surfacing the parse error above.

The same code is duplicated in `crates/runtime-next/src/container.rs`, so a fix there gets both paths.

Related: estuary/flow#2968 (docs for the per-language Docker requirement).

## Suggested implementation

Add stderr capture to `docker_cmd` in both `runtime/src/container.rs:379-394` and `runtime-next/src/container.rs`. When stderr contains `"Cannot connect to the Docker daemon"`, return a tailored error like:

> Docker daemon is not running. Start Docker Desktop (or your container runtime — set `DOCKER_CLI=podman` if using podman) and retry. `flowctl preview` and `flowctl catalog test` require Docker for TypeScript and Python derivations, captures, and materializations.

Optional complement: a `docker version --format '{{.Server.Version}}'` probe in `crates/flowctl/src/preview/mod.rs:106` (right before `runtime::Runtime::new`) to surface the error one step earlier with no spec-parse work wasted.

## Why this matters

New users writing their first Python or TS derivation hit this and have no signal that Docker is the missing prerequisite. The current error string suggests an image / spec problem, sending them down the wrong debugging path. SQLite walkthroughs work fine without Docker, so customers don't expect Docker to be needed.

RitwijParmar:
## Summary
- detect Docker daemon/connectivity errors even when the docker CLI exits successfully with the real failure on stderr
- return a clear flowctl-oriented error before empty docker inspect output reaches JSON parsing
- apply the fix to both runtime and runtime-next container paths with focused unit coverage

Fixes #2969.

## Tests
- cargo fmt --check
- cargo test -p runtime detects_docker_daemon_unavailable_from_stderr --lib
- cargo test -p runtime ignores_unrelated_docker_stderr --lib
- cargo test -p runtime-next detects_docker_daemon_unavailable_from_stderr --lib
- cargo test -p runtime-next ignores_unrelated_docker_stderr --lib
- git diff --check

### DIFF
--- a/crates/runtime-next/src/container.rs
+++ b/crates/runtime-next/src/container.rs
@@ -384,6 +384,10 @@ where
         .await
         .with_context(|| format!("failed to run docker command {args:?}"))?;
 
+    if let Some(err) = docker_daemon_unavailable_error(&output.stderr) {
+        return Err(err).with_context(|| format!("docker command {args:?} failed"));
+    }
+
     if !output.status.success() {
         anyhow::bail!(
             "docker command {args:?} failed: {}",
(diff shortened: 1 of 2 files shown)


RitwijParmar:
Closing this PR because it has been inactive for several months and is no longer merge-ready. Thanks for the consideration.

PR #2977, opened 2026-05-27
validation: allow multiple constraints per projection

mdibaiee:
**Description:**

- Materialization connectors report field constraints through a map<string, Constraint> in Validated.Binding. Because it's a map, only one constraint per projection field is possible.

- This caused a bug we saw with a customer recently: when a new binding is attached to an existing table whose `flow_document` column has an incompatible type, the connector emits `flow_document: INCOMPATIBLE`. The control plane drops bare `INCOMPATIBLE` on new bindings (no live spec) because there's no way to know whether the field is required — so `FieldSelection.Document` is set to "" and the materialization silently degrades to
  delta updates writes since there is no `flow_document`, creating duplicate rows on every update.

- The connector can't fix this unilaterally: emitting `LOCATION_REQUIRED` instead of `INCOMPATIBLE` hides the incompatibility signal: they can publish but they will get an error during connector runtime that complains about an incompatible `flow_document` column.

- This PR allows connectors to provide multiple constraints per field. This allows a connector to emit both `LOCATION_REQUIRED` and `INCOMPATIBLE`, which results in control plane throwing a clean error about why the field cannot be materialized as-is.

- Existing tasks are not affected, there is full backwards compatibility. Once we merge and land this, I will work on migrating all materializations to use the new protocol field, and later we can remove the old deprecated field.

**Workflow steps:**

**Notes for reviewers:**


crates/validation/src/field_selection.rs
```diff
@@ -246,49 +273,73 @@ pub fn extract_constraints<'a>(
...
+                    },
+                )),
+                Ok(ConstraintType::FieldForbidden) => rejects.push((
+                    field,
+                    Reject::ConnectorForbids {
+                        reason: constraint.reason.clone(),
+                    },
+                )),
+                Ok(ConstraintType::Incompatible | ConstraintType::Unsatisfiable)
+                    if live_field_selection.is_some() || has_required =>
```
jgraettinger: This is a little unsatisfying, because we could proceed if:
- there are multiple fields of the same pointer that are LocationRequired
- at least _one_ of them is !Incompatible, even if another is

It's been a while since I was in here so this recommendation is vibe-y, but it makes me think we should emit the reject here unconditionally, and filter it later (when we're walking in grouped+prioritized order, and understand the status of required locations).

jgraettinger (approved):
LGTM, this is strictly better than what we have today, but a comment below on (what I suspect would be) a better factoring for reconciling the constraint conflict.

PR #2943, opened 2026-05-13
Service Accounts 

GregorShear:
## Summary

- Adds **service accounts**: non-login identities (backed by `auth.users` rows) that authenticate via API keys and are authorized through the existing `user_grants` / `role_grants` system. Two new tables: `internal.service_accounts` and `internal.api_keys`.
- Adds GraphQL mutations `createServiceAccount`, `addServiceAccountGrant`, `removeServiceAccountGrant`, `createApiKey`, `revokeApiKey`, and a paginated `serviceAccounts` query.
- Extends `POST /api/v1/auth/token` with an `api_key` grant type that exchanges a `flow_sa_`-prefixed API key for a short-lived access token. 
- An API key may also be presented directly as an `Authorization: Bearer` credential.
- Adds a fine-grained `ManageServiceAccounts` capability (included in the `TeamAdmin` bundle) that gates service-account management.

### Key design decisions

- **Service accounts are `auth.users` rows.** All existing RLS policies, PostgREST authorization, `user_roles()` resolution, and `role_grants` traversal work unchanged. Each account gets a synthetic, non-login address (`<catalog_name>@service.estuary.dev`) and stores its catalog name as `full_name`.
- **`catalog_name` is a management anchor** It is unique and determines *who* may manage the account (admins of a covering prefix, via `ManageServiceAccounts`) and how the account is addressed.
- Access is governed solely by the service_account's `user_grants`, which may span multiple prefixes. 
- **Authorization is split to prevent privilege escalation.** Managing an account (create, mint/revoke keys, add/remove grants) requires `ManageServiceAccounts` on the catalog name. *Adding* a grant additionally requires `CreateGrant` on the granted prefix, so a caller can't hand a service account reach they couldn't grant anyone. *Removing* a grant requires only management capability, since narrowing access is always safe.
- **API keys are hashed with SHA-256. (Refresh tokens hash using bcrypt which normally protects against brute-force attacks - unnecessary with a high-entropy random secret) 
- **API keys use fixed expiry.** `validFor` is an ISO-8601 duration (e.g. `P90D`, `P1Y`), bounded to `(0, 1 year]`, converted to `now() + interval`. This is distinct from the sliding window used by refresh tokens.
- **Service accounts cannot create refresh tokens.** `createRefreshToken` rejects SA principals, since a refresh token would bypass the expiring, revocable API-key model.

## Test plan

- [ ] Create service account with grants → mint API key → exchange for an access token via `POST /api/v1/auth/token`
[...]

PR #2983, opened 2026-05-29
Migrate flowctl to gql operations and token-exchange rest endpoints

GregorShear:
## Migrate flowctl auth off PostgREST onto GraphQL + token-exchange endpoint

Updates flowctl to use the new application-layer auth surfaces introduced in [phase 1](https://github.com/estuary/flow/pull/2943), retiring the dependence on the PostgREST RPCs `create_refresh_token` and `generate_access_token`.

### What changed

**Refresh-token creation** now goes through the `createRefreshToken` GraphQL mutation instead of the `create_refresh_token` PostgREST RPC.

**Access-token exchange** now `POST`s to `/api/v1/auth/token` with `grant_type: refresh_token` (unauthenticated) instead of the `generate_access_token` RPC.

Both client implementations are updated in lockstep:
- `flow-client` (legacy, used by flowctl) — `refresh_authorizations` in `client.rs`
- `flow-client-next` — `UserTokenSource` / `create_refresh_token` / `exchange_refresh_token` in `user_auth.rs`

`UserTokenSource` swaps its `pg_client: postgrest::Postgrest` field for `rest_client: crate::rest::Client`.

### New helpers

A small GraphQL request helper is added to each crate so callers can issue a query/mutation and deserialize the `data` field directly:
- `flow_client::graphql(...)` — `anyhow`-based.
- `flow_client_next::rest::Client::graphql(...)` — typed errors via a new `GraphqlError` enum (adds the `thiserror` dependency).

### Notes for review

- `valid_for` moves from the PostgREST-style `"90d"` to ISO-8601 `"P90D"`, matching the GraphQL mutation's contract.
- The `multi_use: true` argument is dropped — it's no longer part of the `createRefreshToken` signature.
- The token-exchange call is deliberately unauthenticated; the refresh token's id + secret are the only credentials needed, and any stale/expired access token on the client would otherwise cause a spurious rejection.


GregorShear:
changed direction here, can close for now

PR #2984, opened 2026-05-29
Remove mention of connectors tab from docs

GregorShear:
It was removed recently: https://github.com/estuary/ui/pull/1982


aeluce (approved):
Good catch, thanks!

PR #2986, opened 2026-05-30
V2 runtime soak test, and punch-list fixes & features it surfaced

jgraettinger:
**Description:**

This branch stands up a continuous, self-checking soak test for the V2 runtime,
and a slew of punch-list items:
* Fixes from prior PR code-review, and deep code reviews of the `runtime-next` and `shuffle` crates
* Implementation gaps identified during development of the soak test
* Fixes for actual bugs identified by the soak test.

The test (`tests/soak/`) drives the complete V2 chain under one prefix — a
capture, three event collections, the `accounts` TypeScript derivation, and two
materializations: `materialize-postgres` for at-rest verification plus a
standalone Python `ledger` connector that re-checks the same invariants in flight
while driving the materialize Load / max-keys / connector-state scatter-gather
paths. Every task carries `enable-runtime-v2`. The workload is a double-entry
transfer ledger engineered so that any lost, duplicated, reordered, or
causally-torn document surfaces as a downstream contradiction — giving us a
repeatable harness that exercises exactly-once delivery, in-order (associative,
non-commutative) reduction, cross-collection transactional atomicity, shard
splits, and global conservation invariants. The soak connectors are
standalone Pydantic-only Python (no estuary-cdk), launched from an in-project
poetry venv.

The remaining commits are runtime hardening, grouped roughly as:

- **Connector codec & extraction.** Negotiate the connector wire codec
  (protobuf vs JSON) from the image at startup, and unify key/value extraction
  behind a single `doc::Encoding` that emits either FoundationDB packed tuples
  or JSON arrays — then populate the conditional JSON tuple fields of the derive
  and materialize protocols accordingly.
- **V2 gRPC authentication & authorization.** Front the Leader and Shuffle
  services with inbound auth: a per-service capability floor (`LEAD` / `SHUFFLE`)
  plus a per-handler scope check scoped to a task's shard-id prefix, and
  self-signed data-plane bearers when dialing peers. Local / `preview` paths stay
  trusted-local and unauthenticated by construction.
- **Shuffle correctness & observability.** Close several checkpoint-visibility
  and ordering gaps (ACK self-hints across bindings sharing a journal,
  `flushed_lsn` non-regression, prefix-aware partition filtering, the
  `u16::MAX` block edge) and make head-of-line read stalls observable via tracked
  stalled reads, a `stall` trace event, and a new gauge.
- **Runtime-next fixes.** A materialize idempotent-replay recovery livelock, the
  connector state-patch wire framing, build-time resolution of the built-in
  derive image (so Validate and the runtime compile against the same connector),
  and converting several silent "impossible state" skips into hard errors.
- **preview-next & packaging.** Clean Ctrl-C drain, correct multi-shard stdout
  interleaving and per-shard materialize Task initialization, local-connector
  absolute-path resolution, and packaging `runtime-sidecar` into the reactor
  image.

Per-fix mechanism and rationale live in the individual commit messages.

**Workflow steps:**

Primarily internal / runtime-facing. Any soak task can be run in isolation with
`flowctl raw preview-next <task>/flow.yaml` (each file imports only the upstream
specs it sources), or the whole chain published via `tests/soak/flow.yaml` on a
local stack. Operators gain a `shuffle_slice_stalled_reads` gauge and a `stall`
trace event for observing which journals are blocking a Slice's drain.

None external. In-repo READMEs updated under `tests/soak/`, `crates/shuffle`, and
`crates/runtime-sidecar`.


jgraettinger:
I see the test flake here, and turns out it's actually a race in Gazette that needs a separate fix (Etcd txns that delete a journal, and then re-create a journal, are being coalesced into an effective update; shards actually handle this correctly by flagging a changed CreateRevision as a delete-then-re-create).

crates/shuffle/src/slice/state.rs
```diff
@@ -419,22 +419,42 @@ pub fn sequence_document(
...
         let mut index: Vec<(&str, u16, u32, PartitionFilter)> = entries.collect();
 
         index.sort_by(|a, b| a.0.cmp(&b.0).then(a.1.cmp(&b.1)));
 
+        // A cohort reads a journal under multiple bindings iff two entries share a
+        // prefix and cohort. Entries are sorted by (prefix, binding), so any such
+        // pair is adjacent. This is conservative — `lookup`'s partition filter
+        // still scopes self-hints to bindings that actually read a given journal —
+        // but it cheaply clears the common 1:1 case.
+        let cohort_shares_journal = index
```
williamhbaker, line 448: The adjacency check compares cohort but the sort above is by binding index, so aren't same-(prefix, cohort) entries only adjacent if nothing else on that prefix falls between them? For example three transforms on one source with priorities `[0,1,0]` → cohorts `[0,1,0]`, where neither windows(2) pair matches, so this comes out false even though cohort 0 reads the journal under bindings 0 and 2 — which would skip the self-hint projection on line 549.

crates/runtime-next/src/patches.rs
```diff
@@ -57,10 +65,10 @@ pub fn encode_connector_state(state: Option<proto_flow::flow::ConnectorState>) -
     b.push(b'[');
 
     if !merge_patch {
-        b.extend_from_slice(b"null\n,");
+        b.extend_from_slice(b"null\t,");
     }
     b.extend_from_slice(&updated_json);
-    b.extend_from_slice(b"\n]");
+    b.extend_from_slice(b"\t]");
```
williamhbaker, line 90: If a connector emits state with tabs (like from `json.MarshalIndent(_, "", "\t")` or something), is that going to a cause a problem? I don't know of any connectors that do that nor should they really, but it doesn't seem impossible. I wouldn't think this is a blocker, but maybe a constraint to be aware of for connectors.

crates/doc/src/extractor.rs
```diff
@@ -300,38 +324,142 @@ impl Extractor {
...
+        indicator: &AtomicBool,
+    ) -> std::io::Result<()> {
+        // Our SerNode values never produce a serde serialization error, so any
+        // failure is an IO error surfaced from `w`.
+        match self.value_from_resolved(resolved) {
+            Ok(node) => serde_json::to_writer(
+                &mut *w,
+                &self.policy.with_truncation_indicator(node, indicator),
+            ),
+            Err(value) => serde_json::to_writer(&mut *w, value.as_ref()),
```
williamhbaker: This Err branch skips `SerPolicy`, but the Ok branch above and the packed path both apply it. Since Err is the default for a missing field, would a really large default value get written untruncated here, unlike the packed encoding of the same doc?

williamhbaker (approved):
LGTM % a few comments

crates/shuffle/src/slice/state.rs
jgraettinger: Indeed! Fixed.

crates/runtime-next/src/patches.rs
jgraettinger: I added a defensive normalization pass. `memchr` is very fast and I don't think the cost will be material.

crates/doc/src/extractor.rs
jgraettinger: Yep, fixed.

Issue #2890, opened 2026-04-24
Prevent integer schema inference from over-widening past `max(i64)`


williamhbaker:
### Context
Currently, schema inference for integer fields scales by powers of 10 to minimize schema churn. For example, an observed value of `3` sets the inferred maximum to `10`, and an observed value of `12` sets the maximum to `100`. This broader granularity is by design, as updating inferred schemas too frequently is undesirable.

### The Problem
The tension with this approach is that materializations rely on these inferred values to determine what kind of column to create. If the power-of-10 rounding pushes an inferred maximum beyond the limits of an `i64` (64-bit signed integer), the system will allocate a needlessly larger column type. 

At these specific type boundaries, exactness is critical. We only want an inferred maximum to exceed `max(i64)` if the actual data strictly requires it. Other useful cutoffs to consider are the 128-bit and 256-bit points, though these are less commonly handled separately for materializations.

### Proposed Solution
Maintain the power-of-10 granularity for the vast majority of cases to prevent schema churn, but implement specific bounds testing for critical numeric thresholds (e.g., `i64` limits). This will prevent the inference engine from artificially over-widening column types during materialization.

dgreer-dev:
Completed, merged

PR #2987, opened 2026-06-01
#2890 Prevent integer schema inference from over-widening

dgreer-dev:
Description:

Schema inference expands inferred minimum/maximum bounds by powers of ten as new values arrive. i64::MAX is a meaningful boundary, though: downstream materializations read these bounds to choose a column type, and a maximum above i64::MAX pushes them from bigint to a wider numeric type. Many columns hold values that will always fit in an i64 but, under plain power-of-ten rounding, would have their maximum rounded past i64::MAX and needlessly widen the column.

This change snaps the bound to exactly i64::MAX (and i64::MIN) instead of rounding past it — a bound only crosses the boundary when an observed value genuinely requires it.

The clamp is integer-only: it applies to values that arrive as JSON integers (PosInt/NegInt). Values encoded as floats — anything with a decimal point or exponent, e.g. 2e18 — are intentionally left on the plain power-of-ten path and are not clamped.

Example (values arriving as JSON integers):

100000000000000000 (1e17) arrives. Its power-of-ten bound is 1e18, which fits within i64::MAX, so the new maximum is 1e18.
2000000000000000000 (2e18) arrives. The power-of-ten bound would be 1e19, which exceeds i64::MAX — but the value itself fits, so the maximum clamps to exactly i64::MAX.
9223372036854775808 (i64::MAX + 1) arrives. It does not fit in i64::MAX, so the maximum widens to 1e19 as normal.

### DIFF
--- a/crates/doc/src/shape/widen.rs
+++ b/crates/doc/src/shape/widen.rs
@@ -305,7 +305,8 @@ impl Shape {
     ///
     /// The approximate lengths of arrays and strings are attached to widened Shapes at
     /// power-of-two boundaries. Numeric ranges are also attached, at order-of-magnitude
-    /// (10x) boundaries.
+    /// (10x) boundaries, except that they snap to the i64 limits rather than crossing
+    /// them, so that materializations don't widen a column past i64 needlessly.
     pub fn widen<'n, N>(&mut self, node: &'n N) -> bool
     where
         N: AsNode,
@@ -430,7 +431,8 @@ fn length_bounds(l: usize) -> (u32, u32) {
     }
 }
 
-// Compute a lower and upper order-of-magnitude bound for the given number.
+// Compute a lower and upper order-of-magnitude bound for the given number,
+// respecting the i64 limits.
 #[cold]
 fn number_bounds(num: json::Number) -> (json::Number, json::Number) {
     use json::Number;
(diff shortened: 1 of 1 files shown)


crates/doc/src/shape/widen.rs
```diff
@@ -439,11 +453,9 @@ fn number_bounds(num: json::Number) -> (json::Number, json::Number) {
         Number::Float(f) => {
             if f.total_cmp(&-1.0).is_le() {
                 let e = (-f).log10() as i32;
-                let l = -(10f64.powi(e + 1));
-                (
-                    Number::Float(if l.is_finite() { l } else { f64::MIN }),
-                    Number::Float(-(10f64.powi(e))),
-                )
+                // Order-of-magnitude bound, saturating at f64::MIN on overflow.
```
williamhbaker: For the sake of minimizing risk of unintended consequences, it looks like these `Number::Float(f)` cases are just cosmetic changes? If so, can we revert those to minimize the diff to the required change on the `PosInt` case?

crates/doc/src/shape/widen.rs
```diff
@@ -431,6 +432,19 @@ fn length_bounds(l: usize) -> (u32, u32) {
 }
 
 // Compute a lower and upper order-of-magnitude bound for the given number.
+//
+// Integer inputs snap to the i64 / u64 boundary: materializations choose a column
```
williamhbaker: Stylistic nit: I'd generally try to trim these comments down quite a bit. Comments are certainly great for explanation the "why's" of things and non-obvious behavior, but I find that LLM-generated code in particular often goes pretty overboard, and kind of miss the over-arching narrative. For example, I see the same mention the rationale for materializations not wanting to widen a column past i64 needlessly in this comment, the one above, and also in comments in the test code. It probably makes sense to mention that somewhere, but not in all 3 places. Tactically...and I doubt this would ever happen, BUT...if that downstream requirement were to ever change, we'd have to be diligent about updating all these comments too, and probably end up with stale / conflicting comments.

Along those same lines, the last paragraph of this comment is basically repeated in the code itself on line ~491.

williamhbaker (commented):
Looking good! One diff-minimization request, and also a comment about comments.

williamhbaker (approved):
LGTM

PR #2988, opened 2026-06-01
Guard against stale GQL schema

GregorShear:
## Guard against committed GraphQL SDL drift

`crates/flow-client/control-plane-api.graphql` is generated from the control-plane schema and committed to the repo.
Nothing currently verifies it stays in sync, so a schema change that lands without regenerating leaves the checked-in SDL stale until someone notices.


jgraettinger (approved):
LGTM

PR #2990, opened 2026-06-01
grant authenticated SELECT on bundles so PostgREST representation works

GregorShear:
## Problem

Removing a user_grant in the dashboard failed with `42501 permission denied for table user_grants`. Migration `20260511120000` replaced the table-level grant with column-level grants that exclude `bundles` (part of our recent authorization refactor), but supabase mutations use `Prefer: return=representation`, which PostgREST runs as `RETURNING *` — selecting `bundles`, which `authenticated` can't read. Same break applies to grant insert/update, and it all applies to role_grants, too.

## Fix

Grant `SELECT (bundles)` to `authenticated` on `user_grants` and `role_grants`. Reads only — writes to `bundles` stay restricted to `service_role`.

### DIFF
--- a/supabase/migrations/20260601120000_grant_bundles_select.sql
+++ b/supabase/migrations/20260601120000_grant_bundles_select.sql
@@ -1,0 +1,20 @@
+begin;
+
+-- Migration 20260511120000 added the `bundles` column to user_grants and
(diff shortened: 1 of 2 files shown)


jgraettinger (approved):
LGTM

PR #2994, opened 2026-06-03
data-plane-controller: extra route

qaoj:
**Description:**

Value passthrough in DPC for some extra route config options. Needs to be in-place before IaC changes.

**Workflow steps:**

Set `extra_private_routes` on the config (a CIDR block and a gateway ID) and it should populate the needful from there. Non-destructive change to routes.

N/A

**Notes for reviewers:**

N/A


mwillman-estuary (approved):
LGTM

PR #2995, opened 2026-06-03
locate-bin: don't resolve a sibling directory as the binary

jgraettinger:
`locate()` joined the binary name onto the running program's directory and used `Path::exists()` to decide whether to use it or fall back to the $PATH. `exists()` is true for any entry, so a sibling directory sharing the binary's name (e.g. a `go/` source tree next to the program) was canonicalized and returned in place of the real `go` executable.

Gate on `is_file()` instead, which is false for both directories and missing paths. Factor the resolution policy into a private `locate_inner` that takes the program path and $PATH lookup as parameters, decoupling it from the process environment so it can be unit-tested, and add tests covering sibling-file resolution, the directory regression, and the not-found error.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5119,6 +5119,7 @@ name = "locate-bin"
 version = "0.0.0"
 dependencies = [
  "anyhow",
+ "tempfile",
  "tracing",
  "which",
 ]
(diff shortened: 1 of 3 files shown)


Alex-Bair (approved):
LGTM

PR #2996, opened 2026-06-04
flowctl: modernize the client stack and route ad-hoc reads through shuffle

jgraettinger:
## Summary

This branch sands down two long-standing rough edges in flowctl:

1. **Auth/client modernization** — flowctl moves off the legacy `flow-client`
   onto `flow-client-next` + the `tokens` crate, gaining a live, auto-refreshing
   user-token watch that survives rotation of both the user and data-plane token
   layers.

2. **Reads through shuffle** — `collections read`, `logs`, and `raw stats` no
   longer issue direct, single-journal, uncommitted journal reads. They run an
   ephemeral loopback shuffle Service, so reads are committed, span partitioned
   collections, and stop cleanly once caught up to every journal's head.

The remaining commits are the supporting work those two arcs needed: a shuffle
false-stall fix that the new read path depends on, an ops-catalog split that lets
flowctl build just the ops collections offline, and a small version-check polish.

The commits are arranged to be read in order, each building on the last.

## User-visible changes

- **`--uncommitted` is removed.** Reads are now always committed (transaction ACKs
  are skipped), so the flag no longer has meaning.
- **Single-journal limitation is gone.** `collections read` now spans partitioned
  collections.
- **Reads are non-blocking by default.** They snapshot the current journal heads
  and exit once caught up; pass **`--follow`** to tail live content.
- **New `--not-before <RFC-3339>`** absolute lower-bound flag alongside the
  existing relative `--since`, across the read commands.
- **`FLOW_AUTH_TOKEN`** is honored for a run but never persisted, and is rejected
  when establishing stored credentials (so it can't silently shadow them on the
  next invocation).
- **Version-check warning** now reports both current and latest versions, and is
  suppressed for `dev` builds.

## Documentation links affected

All updates flow from `collections read` / `logs` / `stats` dropping `--uncommitted`
and gaining `--not-before` / `--follow`:

- `site/docs/concepts/collections.md` — rewrote the "Using the flowctl CLI" read
  section: removed the Beta-limitations block (`--uncommitted` required,
  single-partition only, `--output` unusable), documented `--not-before` and
  `--follow`, and renamed `--include-partition` to `--partitions`.
- `site/docs/features/working-logs-stats.md` — dropped the `--uncommitted` Beta
  note for `flowctl stats`; documented read-then-exit plus `--follow`.
- `site/docs/guides/customize-dataflows/customize-materialization-fields.md` —
  removed `--uncommitted` from the example command.
- `site/docs/transformations/derivation-tutorial-{python,sql,typescript}.md` —
  swapped `--uncommitted` for `--follow` in the tutorial read examples.
- `ops-catalog/README.md` (internal) — points contributors at the new
  `mise run build:ops-bundles` task for regenerating `*.bundle.json`.

## Notes for reviewers

- The data-plane-template bundle is unchanged in content despite the ops-catalog
  split — the new template is inlined on import. The new build task does, however,
  normalize the machine-specific repo-root prefix in schema `$id`s to a stable
  `file:///flow/` placeholder, which rewrites the existing checked-in bundles.


jgraettinger:
Tagging @GregorShear for `flowctl` bits related to UX, their work, and the control plane. @williamhbaker for lower-level bits (e.x. the `shuffle` false-stall bug).

williamhbaker (approved):
LGTM

Issue #2895, opened 2026-04-27
Introduce x-str-minimum and x-str-maximum validation keywords for numeric-formatted strings


williamhbaker:
Currently, materialization connectors utilize the `minimum` and `maximum` JSON Schema attributes to determine the optimal underlying data types for fields mapped as `type: integer` and `type: number`. This works well for native numeric fields. 

However, for fields mapped as `type: string` with `format: integer` or `format: number`, we currently lack equivalent boundary attributes. Without min/max ranges for these fields, materialization connectors cannot make informed decisions about the best database types to use for numbers that are presented as strings.

## Problem
The initial idea was to apply the existing `minimum` and `maximum` JSON Schema attributes directly to numeric-formatted string fields. However, because these are strict validation keywords (not just annotations) within JSON Schema, reusing them for strings introduces two major issues:

1. **Loss of precision (Lossiness):** Reusing the same keyword mixes the bounds of native numbers and strings. A string representation's bound might exceed a 32-bit or 64-bit limit while the native numeric bound does not, making it difficult to differentiate and act upon.
2. **Breaking backfills:** Validation is critical to the end-to-end schema inference process. If we started enforcing standard `maximum` validation against strings, established collections with existing strings larger than the inferred `maximum` would permanently break during materialization backfills. Connectors rely on validation failures to safely halt, restart, and apply required DDL changes *before* processing widened documents.

## Proposed Solution
Introduce two new custom validation keywords specifically for string fields:
* `x-str-minimum`
* `x-str-maximum`

## Implementation Details / Requirements

* **Schema Inference Update:** Schema inference should begin populating `x-str-minimum` and `x-str-maximum` for *new* locations where `type: string` and `format: integer/number` are detected. 
* **Backward Compatibility:** These new bounds should *not* be added to established string locations that do not already have them (matching our existing behavior for standard `minimum`/`maximum` to prevent breaking existing streams).
* **Validation Enforcement:** These **must** be implemented as strictly enforced validations in the validation crate, not merely as data-plane annotations. 
* **Connector Guarantee:** This enforcement guarantees that a connector will never receive a document that violates an `x-str-maximum` or `x-str-minimum` setting it was provided at startup, giving it the opportunity to halt and apply DDL updates if the schema widens.

PR #2997, opened 2026-06-04
Schema/Proto support for `x-str-minimum`/`x-str-maximum`

dgreer-dev:
**Description:**

Adds support for the `x-str-minimum` / `x-str-maximum` JSON-Schema annotation
keywords, which bound the value of `format: integer`/`format: number` values.
These exist because such strings can carry values outside 64-bit ranges, so the
bounds are represented as canonical decimal text rather than numbers. (Issue #2895.)

This PR is part 1 of 2, and covers everything *except* exposing the bounds
through the protocol:

- **Schema/shape side** (`crates/json`, `crates/doc`): a landmark `Bound` type
  (canonical decimal `Display`), validation enforcement of the two keywords,
  shape inference, widening, intersect/union, and serialization of the bounds
  back out as `x-str-minimum` / `x-str-maximum` in inferred schemas.
- **Protocol holders** (`go/protocols/flow`, regenerated Go + Rust bindings):
  adds `Inference.String.str_minimum` / `str_maximum` (field numbers 8 and 9) as
  string fields, empty when unset. These are *holders only* — `assemble` writes
  them empty, so there is **zero change to serialized protobuf on the wire**.
  Populating them from the inferred shape is deferred to a stacked follow-up PR,
  so the regenerated structs roll out to all readers before any spec carries a
  value.

**Workflow steps:**

As of this PR the platform validates documents against these bounds and carries
them through schema inference (including inferred-schema write-back). The bounds
are not yet visible to connectors via projection inference.

**Notes for reviewers:**

- **No protobuf wire-byte change in this PR**: the new fields are plain proto3
  scalars left empty, and the generated marshalers skip empty values
  (`len() > 0` in Go, `!is_empty()` in Rust), so nothing is emitted.


crates/doc/src/annotation.rs
```diff
@@ -51,6 +51,9 @@ impl schema::Annotation for Annotation {
 
     fn uses_keyword(keyword: &str) -> bool {
         match keyword {
+            // `x-str-minimum` / `x-str-maximum` are enforced validation keywords not annotations
+            // so we need to pre-empty the starts_witH("x-") check.
```
williamhbaker: Typo in `starts_witH`...also `pre-empty` -> `pre-empt`?

crates/json/src/number.rs
```diff
@@ -240,11 +240,547 @@ impl From<&f64> for Number {
...
+        // Order-of-magnitude of abs(v), clamped at zero so sub-unit values map to 10^0.
+        let e = ((v.digits() as i64 - 1) - v.fractional_digit_count()).max(0);
+
+        let edges = IntKind::ALL.into_iter().flat_map(|kind| {
+            [
+                Bound::IntEdge(kind, Edge::Min),
+                Bound::IntEdge(kind, Edge::Max),
+            ]
+        });
+        let powers = ((e - 1).max(0)..=(e + 1)).filter_map(move |k| {
```
williamhbaker: Similar to the above: A value of `1e9223372036854775807` has an `e` of `i64::MAX`, which would overflow here in the `e+1`.

crates/json/src/number.rs
```diff
@@ -240,11 +240,547 @@ impl From<&f64> for Number {
...
+            .min()
+    }
+
+    /// Landmark values that could be the tightest enclosing bound of `v`:
+    /// zero, every integer-type edge, and the powers of ten bracketing `abs(v)`.
+    /// Yielded lazily so `floor`/`ceil` consume them without a heap allocation.
+    fn candidates(v: &BigDecimal) -> impl Iterator<Item = Bound> {
+        let sign = Sign::of(v.sign());
+        // Order-of-magnitude of abs(v), clamped at zero so sub-unit values map to 10^0.
+        let e = ((v.digits() as i64 - 1) - v.fractional_digit_count()).max(0);
```
williamhbaker: This would admittedly be a pretty crazy input, but maybe not entirely impossible: A value of `1e9223372036854775808` ends up being `0 - i64::MIN` here, which overflows.

crates/doc/src/shape/widen.rs
```diff
@@ -25,6 +25,15 @@ impl StringShape {
             self.max_length = Some(max);
             self.min_length = min;
 
+            if is_numeric_format(&self.format) {
+                // Seed numeric-string bounds for new `format: integer`/`number`
+                // locations by snapping to the tightest enclosing landmarks.
+                if let Ok(num) = val.parse::<bigdecimal::BigDecimal>() {
+                    self.str_minimum = json::Bound::floor(&num);
+                    self.str_maximum = json::Bound::ceil(&num);
```
williamhbaker: When the first observed value is a landmark, floor and ceil both return the same thing, so for example a value of `100` we'd get a min / max of `100`. That's different than for example `number_bounds(100)` which gives `100` / `1000`. A similar case for `i64::MAX`, which is `i64::MAX` / `i64::MAX` for these str_min/max, but would be a power-of-10 min / `i64::MAX` for the `number_bounds` case due to the recently added snapping.

williamhbaker (commented):
I started this review and left a few comments. Generally though I am wondering if we could have a much simpler implementation.

A goal that I recommend would be the to have the bounds computations be as consistent as they can be with the regular integer and number bounds for the their respective `format: integer` and `format: number`. So that could help...for `format: integer` I don't think it's worth us adding boundaries other than powers-of-10 and the i64::MIN/MAX. And for `format: number` just powers of 10. Unless adding those boundaries is effectively ~free in terms of code required, but if we don't have them for regular integer/numbers I don't see much justification for their `x-str` counterparts.

I'm also (still) curious about this: We seem to be trying really hard to avoid parsing strings into a `BigDecimal` for the new comparisons. That certainly makes sense. But we actually already do that anyway for values that are known to be `format: integer/number`, in [visit_string](https://github.com/estuary/flow/blob/b8f36de733afe116ccab0f7ad01615af39bdabfa/crates/json/src/validator/validation.rs#L511-L555), and that's right next to where we need to compute the outcome for the `KeyWord::XStrMinimum` / `KeyWord::XStrMaximum`. So I'd hope we could do some kind of caching of the computed `BigDecimal` (we have to do it either way) and then use _that_ for the comparison.

crates/doc/src/shape/mod.rs
```diff
@@ -60,6 +60,10 @@ pub struct StringShape {
     pub format: Option<Format>,
     pub max_length: Option<u32>,
     pub min_length: u32,
+    /// `x-str-minimum` value:  Minimum numeric value for number-ish formats
+    pub str_minimum: Option<json::Bound>,
```
williamhbaker: What if we used `bigdecimal::num_bigint::BigInt` or perhaps `BigDecimal` here? As an `Option<Box<>>`. It would push the shape size slightly larger, but I think it may have two benefits that could make it worth it:

1) Further simplify the comparison logic in `number.rs`, perhaps significantly, and to the point where the widening logic could live alongside `number_bounds` in `doc/widen.rs` (maybe as `str_number_bounds`?)
2) Eliminate the allocation of a `BigInt` that would otherwise falls out from `cmp_decimal` on common cases

dgreer-dev: Yep, seems reasonable, and easy enough to implement

crates/doc/src/shape/mod.rs
```diff
@@ -291,8 +297,10 @@ mod test {
     fn shape_size_regression() {
         use super::{ArrayShape, ObjShape, Shape, StringShape};
         assert_eq!(std::mem::size_of::<ObjShape>(), 56);
-        assert_eq!(std::mem::size_of::<StringShape>(), 32);
+        // Two `Option<Box<BigDecimal>>` numeric-string bounds (the x-str-min/max
```
williamhbaker: This comment reads as if it is referencing a discarded prior strategy?

crates/doc/src/shape/intersect.rs
```diff
@@ -42,10 +42,24 @@ impl StringShape {
...
 }
 
+// Combine two optional numeric-string bounds for an intersection
+fn intersect_bound(
+    lhs: Option<Box<bigdecimal::BigDecimal>>,
+    rhs: Option<Box<bigdecimal::BigDecimal>>,
+    take_greater: bool,
+) -> Option<Box<bigdecimal::BigDecimal>> {
+    match (lhs, rhs) {
+        (Some(l), Some(r)) => Some(if (l > r) == take_greater { l } else { r }),
```
williamhbaker: nit: This is kind of breaking my brain, vs. `(Some(l), Some(r)) => Some(if take_greater { l.max(r) } else { l.min(r) })`

crates/doc/src/shape/union.rs
```diff
@@ -70,6 +72,18 @@ impl StringShape {
...
 }
 
+// Combine two optional numeric-string bounds for a union, taking the looser one
+fn union_bound(
+    lhs: Option<Box<bigdecimal::BigDecimal>>,
+    rhs: Option<Box<bigdecimal::BigDecimal>>,
+    take_greater: bool,
+) -> Option<Box<bigdecimal::BigDecimal>> {
+    match (lhs, rhs) {
+        (Some(l), Some(r)) => Some(if (l > r) == take_greater { l } else { r }),
```
williamhbaker: nit: Same as the comparison above...I think `(Some(l), Some(r)) => Some(if take_greater { l.max(r) } else { l.min(r) })` is more self-explanatory.

crates/json/src/schema/formats.rs
```diff
@@ -161,6 +154,32 @@ impl Format {
...
+    /// (and `bigdecimal` would otherwise accept it)
+    pub(crate) fn parse_numeric(val: &str) -> Option<bigdecimal::BigDecimal> {
+        if val.contains('_') {
+            return None;
+        }
+        bigdecimal::BigDecimal::from_str(val).ok()
+    }
+
+    /// Whether `val` satisfies this numeric format, given its already-parsed
+    /// value from [`Format::parse_numeric`]. Split out of [`Format::validate`]
```
williamhbaker: I'm thinking we could drop the `Split out of [`Format::validate`]` part from the comment.

crates/json/src/validator/validation.rs
```diff
@@ -1127,6 +1181,99 @@ mod tests {
...
+        // These aren't orderable as decimals, so the bounds don't apply to them
+        // (the bound only constrains finite numeric magnitudes).
+        let schema = serde_json::json!({
+            "type": "string",
+            "format": "number",
+            "x-str-minimum": "0",
+            "x-str-maximum": "100",
+        });
+
+        assert!(validates(schema.clone(), serde_json::json!("NaN")));
```
williamhbaker: This is interesting.

We do see these non-finite values show up from time to time. Materialization connectors have special handling for their values, even. What they do with these values is destination-dependent, but ranges from "just insert a `NUL`", to to "use the actual destinations support representation of such a value", to "just use a 0".

Although I don't know of any specific current cases where this would be a problem offhand, I think it could be surprising for a schema with a declared maximum of say `"1000"` to allow a `+Infinity` value through to the materialization. Infinity is greater than 1000, after all. If a materialization created a specific kind of column based on the x-str bounds, it might create one that can't hold these non-finite values, whereas in their absence it could have used a less-specific (but more accomodating) column.

I think it would be a little better to clear the bounds on the presence of a non-finite value, and strictly validate schemas as failing for non-finite values if a x-str-minimum/maximum is present. We'd be saying that positive or negative infinite does _not_ fit within the specified range, and for `NaN` that it does not fit in a range of actual numbers.

One other thing this would improve: Right now there is some non-determinism based on the ordering of values: `NaN` followed by `"1000"` would get inferred with no bounds, since `NaN` can't see the schema and so it carries forward forever without a bound. But in reverse order you'd get `"1000"`...the later `NaN` would not clear the bounds. If we always clear the bounds on a non-finite value, it would consistently not have bounds if one were present.

crates/doc/src/shape/widen.rs
```diff
@@ -491,6 +528,63 @@ fn number_bounds(num: json::Number) -> (json::Number, json::Number) {
...
+    const MAX_STR_BOUND_EXP: i64 = u16::MAX as i64; // don't bother bounding pass this exponent
+
+    // floor(log10(|num|)): mantissa digit count, less one, less the fractional
+    // digits. Saturating so a pathological scale (e.g. "1e9999999999") can't
+    // overflow the subtraction.
+    let e = (num.digits() as i64)
+        .saturating_sub(1)
+        .saturating_sub(num.fractional_digit_count());
+
+    let dec = |n: BigInt| Some(Box::new(bigdecimal::BigDecimal::from(n)));
```
williamhbaker, line 562: Are we going to end up possibly writing out huge numbers of 0's for really large numbers? I'm not 100% sure here, but I think `BigDecimal::from(BigInt)` gives scale 0 with every zero in the mantissa, and would lead to full-decimal printing later when we do `to_string()`.

williamhbaker (commented):
A few more comments, having looked at this holistically, in detail now.

crates/doc/src/shape/widen.rs
dgreer-dev: I've been thinking about this one -- yeah under huge numbers that gets really really unwieldy to print it all out, but if we collapse it to like, 1e1000 we end up inexact again because that is likely getting parsed on the other side as a float.

What about this, we just lower the max we track.  We don't really care about bounding 1000 digit numbers, let's just cap it at like, 1e64 or something and we don't widen/enforce bounds past that?

crates/json/src/validator/validation.rs
dgreer-dev: Great point on this.  I hadn't considered the determinism issues but yeah I think your solution here is right.

crates/doc/src/shape/widen.rs
williamhbaker: That sounds good, just clear the bounds on seeing a really large number. Once we start getting beyond values that can reasonably represented in the full text form we're into the "effectively unbounded" territory anyway.

crates/doc/src/annotation.rs
```diff
@@ -51,6 +51,9 @@ impl schema::Annotation for Annotation {
 
     fn uses_keyword(keyword: &str) -> bool {
         match keyword {
+            // `x-str-minimum` / `x-str-maximum` are enforced validation keywords
+            // not annotations so we need to preempt the starts_wih("x-") check.
```
williamhbaker: Still a typo in this comment: `starts_wih`

crates/doc/src/shape/widen.rs
```diff
@@ -1361,6 +1600,64 @@ mod test {
...
+            bounds("-9223372036854775808"), // i64::MIN -> [i64::MIN, -10^18]
+            some2(i64::MIN.to_string(), format!("-{}", ten(18))),
+        );
+        // Unlike `number_bounds`, magnitudes beyond u64 round-trip exactly.
+        assert_eq!(bounds("18446744073709551616000"), some2(ten(22), ten(23)));
+
+        // The exponent cap: at the limit bounds still materialize (as short
+        // decimal strings), while one power beyond is dropped on both sides
+        // rather than serializing ever-larger runs of zeros.
+        assert_eq!(bounds("1e64"), some2(ten(64), ten(65)));
```
williamhbaker: This test is currently failing, I believe due to the new 40 zero's cap

williamhbaker (commented):
A couple more small things, and then one other more important item I hadn't thought about before (caught by Fable's code review, which ran for over an hour and burned 75% of my daily limit lol): These also need added to the relaxed write schema's list of fields it removes, see [here](https://github.com/estuary/flow/blob/8bacd0c5af6d5e2c811723905267854ea64ff27b/crates/models/src/schemas.rs#L203-L219). There's some additional background on what this "relaxed write schema" is in [this discussion](https://github.com/estuary/flow/discussions/1988).

A concrete example:
1. A postgres capture has schema inference enabled
2. Initially, a column is just text, so it has `type: string`. Some data is captured into this column that is not numeric, maybe a mix of garbage like `"n/a"` and some numerics like `"10.5"`
3. Later, the source column type is changed to `NUMERIC(4,2)`. The capture _could_ (it's not quite smart enough to do this, yet) reasonably emit a write schema with a `{ type: string, format: number, x-str-minimum: "-99.99", x-str-maximum: "99.99" }` for that field.
4. Unless those `x-str-` values are stripped from the relaxed write schema, they would intersect with the inferred schema and still be enforced, even for data from earlier in the collection that might have the non-numeric string values. This would cause schema violation errors when re-reading collection data from the beginning, which is exactly the thing to relaxed write schema is meant to prevent.

williamhbaker (approved):
LGTM

PR #3003, opened 2026-06-05
runtime: drain shuffle topology on graceful leader shutdown

jgraettinger:
A back-pressured shuffle Log RPC could wedge graceful shutdown. When a Log engages disk back-pressure it stops draining a Slice's Appends, which parks that Slice's request stream so the Log no longer polls it and never observes its EOF. On shutdown the back-pressure never releases on its own (nothing is consuming the local log to reclaim segments), so the Log — and the Session/Slice actors draining it, and the leader awaiting them — hang, leaking tasks, fds, and on-disk segments.

Break the wedge by relieving the back-pressure out-of-band, reusing the existing sealed-segment reclaim path:

- shuffle: add `log::remove_shard_segments`, which unlinks a shard's segment files (scoped to the `mem-{idx:03}-seg-*.flog` naming; sibling shards and `.compress-*` temp files untouched; missing dir / per-file NotFound tolerated). The co-located Log's reclaim observes the unlinks, drops its backlog below threshold, and resumes draining to EOF.
- shuffle: split `SessionClient::next_checkpoint` into a sync `request_checkpoint` + cancel-safe `recv_checkpoint`, letting the leader own the SessionClient directly (instead of via `stream::unfold`) so it can `close()` it.
- runtime-next leaders: after broadcasting L:Stopped, synchronously `SessionClient::close()` — blocking until the whole topology drains — before EOFing shards. Intentionally un-timed: a Session that never drains is a liveness bug to surface, not mask.
- runtime-next shards: on L:Stopped (a transaction boundary, so we're idle and own the shuffle Reader), remove our segment files before awaiting leader EOF.

### DIFF
--- a/crates/runtime-next/src/leader/materialize/actor.rs
+++ b/crates/runtime-next/src/leader/materialize/actor.rs
@@ -64,7 +64,7 @@ impl Actor {
         &mut self,
         mut head: fsm::Head,
         mut tail: fsm::Tail,
-        session: shuffle::SessionClient,
+        mut session: shuffle::SessionClient,
         shard_rx: Vec<BoxStream<'static, tonic::Result<proto::Materialize>>>,
     ) -> anyhow::Result<()> {
         service_kit::event!(
@@ -83,15 +83,10 @@ impl Actor {
             .map(next_shard_rx)
             .collect();
 
-        // Build a stream of frontier checkpoints from the shuffle Session.
-        let frontier_rx = futures::stream::unfold(session, |mut session| async {
-            let result = session.next_checkpoint().await;
-            Some((result, session))
-        });
-        let mut frontier_rx = std::pin::pin!(frontier_rx);
-
         // Per-binding absolute measure, into which deltas are reduced.
         let mut binding_bytes_behind = vec![0; self.task.binding_collection_names.len()];
+        // We keep exactly one NextCheckpoint request in flight while idle.
+        let mut checkpoint_requested = false;
         // When true, Head should close its current open transaction ASAP.
         let mut close_requested = false;
         // Iteration counter for the per-loop trace event.
@@ -211,6 +206,12 @@ impl Actor {
                 );
             }
 
+            // Keep one NextCheckpoint in flight whenever we can accept a frontier.
+            if ready_frontier.is_none() && !checkpoint_requested {
+                session.request_checkpoint();
+                checkpoint_requested = true;
+            }
+
             tokio::select! {
                 biased;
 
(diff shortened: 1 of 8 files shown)


crates/runtime-next/src/leader/materialize/actor.rs
```diff
@@ -211,6 +206,12 @@ impl Actor {
                 );
             }
 
+            // Keep one NextCheckpoint in flight whenever we can accept a frontier.
+            if ready_frontier.is_none() && !checkpoint_requested {
```
williamhbaker, line 210: I'm wondering if we need to gate this somehow on if the task is closing, otherwise it seems like it could request a checkpoint, and then go into `session.close()` without getting the response back, get the response then instead of the expected EOF. Maybe in `session.close()` it needs to read responses until it gets EOF, since a checkpoint requested a couple iterations earlier could still be in flight at exit.

crates/runtime-next/src/shard/materialize/actor.rs
```diff
@@ -247,7 +247,24 @@ impl Actor {
...
+        // log segment files now, before blocking on the leader's EOF below. The
+        // leader is concurrently closing its shuffle SessionClient; deleting these
+        // segments releases any disk back-pressure held by the co-located Log RPC,
+        // letting the shuffle topology drain to EOF so the leader's close() can
+        // complete. Only then does the leader drop our channel, delivering the EOF
+        // we await next.
+        let Phase::Idle { shuffle_reader, .. } = &phase else {
+            anyhow::bail!("leader sent Stopped while shard was not idle");
+        };
+        shuffle::log::remove_shard_segments(
```
williamhbaker, line 261: Could there be enough data still in flight that, once back-pressure releases here, the Log seals new segments and crosses its back-pressure limit again? If so I'm not entirely sure what we could do differently and it's probably not a blocker, since we'd be doing better than we are now at least.

williamhbaker (approved):
LGTM

crates/runtime-next/src/leader/materialize/actor.rs
jgraettinger: Yep we do, but that's implemented in this commit on the flowctl PR: https://github.com/estuary/flow/pull/2996/changes/2dff75c6ed2e37709008ef04f3f548e36fcc0104

Planning to land this one, and then that one.

crates/runtime-next/src/shard/materialize/actor.rs
jgraettinger: I don't think so. The only date in-flight is RPC's stream H/2 window plus a deliberately-tiny channel in the shuffle Slice actor. Practical production values of disk back-pressure are 2GB _compressed_.

jgraettinger: If I had more appetite for a larger change, I might say that the log actor should flow back-pressure signals back up to the Slice actor instead of relying on stream-level flow control. That'd involve new shuffle protocol changes and a re-working of the slice and log actors.

PR #3004, opened 2026-06-06
runtime-next: resync publisher clocks

jgraettinger:
## Summary

The primary change resyncs the runtime-next **publisher clock** at each
transaction's publish, so stamped document UUIDs track transaction write time
instead of drifting behind wall-clock. The remaining commits are cleanups and
local-stack improvements.

## Publisher clock resync

The publisher clock previously advanced only by `Clock::tick`'s ~1µs per
document from session start, so over a long-lived session `flow_published_at`
and the journal clocks drifted progressively behind wall-clock. Now
`Publisher::update_clock()` is called once at the start of every transaction's
stream of published documents:

- the capture and derive shard **drains** (before their `publish_doc` loops), and
- the derive and materialize **leader actors** (before `publish_stats` and the ACK intents).

The clock is monotonic, so resyncing never regresses a prior-written clock.

## Cleanups / local-stack improvements

- **local/test-tenant**: after provisioning, block until the agent's in-memory
  authorization snapshot reflects the new grants — drive one
  `/authorize/user/prefix` request and follow its 307 redirects, which forces a
  snapshot refresh and blocks until the grants are visible — so an immediate
  `flowctl` call no longer races the refresh. Fails on any non-200.
- **local/README & tests/soak/README**: correct the connector-networking model
  (the agent proxies connectors to the data plane via `ProxyConnectors`; only
  the reactor sets `FLOW_NETWORK`), establish local/README as the primary
  stack-wiring reference, and trim the soak README to defer to it.


williamhbaker (approved):
LGTM

PR #3007, opened 2026-06-08
gazette: only mint pre-signed fragment URLs when actively polled

williamhbaker:
**Description:**

`read()` looped straight from one fragment into the next `read_some`, whose metadata RPC may mint a pre-signed fragment URL with a short (~60s) TTL. A consumer that then parked the read before the GET ran let the URL expire, so the GET failed non-retryably (GCS returns 400 for an expired V2 signature) and the read failed.

Yield before re-entering `read_some` so its metadata RPC, and thus any pre-signed URL, fires only once we're polled again — when the consumer is actually ready to pull the next fragment — keeping the sign-to-GET window inside a single active read.

**Workflow steps:**

**Notes for reviewers:**

### DIFF
--- a/crates/gazette/src/journal/read/mod.rs
+++ b/crates/gazette/src/journal/read/mod.rs
@@ -40,6 +40,14 @@ impl Client {
                 {
                     Ok(()) => {
                         attempt = 0;
+
+                        // Yield now rather than looping immediately back into
+                        // `read_some`, which issues a request for metadata &
+                        // maybe a pre-signed fragment URL. Callers may not
+                        // immediately consume the resulting stream, and the
+                        // pre-signed URL might expire in that case. Wait until
+                        // actively polled again to mint any pre-signed URLs.
+                        () = tokio::task::yield_now().await;
                         continue;
                     }
                     Err(err) => err,


dgreer-dev (approved):
I think the issue is having a `co.yield()` *between* the mint and the GET -- the yield is when the consumer can park the whole thing and cause a timeout.

Ideally there'd be, like, an "atomic" way to mint and GET.  If you could re-work it so there's not that yield between the mint and send that would probably be super clean.  You could also maybe spawn a seperate task that does the mint/get so that any kinds of yields don't affect it (might make cancellation hard)?

In any case, I think this PR improves the chances of this not happening, but I think it's not  100% closed while there is a yield splitting those things.

williamhbaker:
Definitely a good point. The `read()` API still feels a bit leaky here because raw callers can observe metadata with a signed URL before the direct GET is started. For the shuffle v2 slice reader, though, I think this is closed in practice since it goes through `ReadLines`. This PR defers fetching the next fragment’s metadata until the actor polls the read again; if `ReadLines` has any aligned buffered content, that poll returns the buffered batch before minting a new signed URL. Once the buffer is drained and the next metadata is fetched, `ReadLines` consumes contiguous metadata internally and immediately polls the inner stream again in the same `poll_next`, so the actor doesn’t get a chance to park the read between minting the URL and starting the GET.

PR #3008, opened 2026-06-09
docs: document associative: false for disabling key deduplication

jwhartley:
## What

Documents the `associative: false` reduction annotation, which lets advanced users disable deduplication on the collection key — and removes the now-inaccurate `EXPERIMENTAL` note from the keyword's code comment.

- **firstwritewins-and-lastwritewins.md** — new "Disabling associative reduction" section explaining the flag, when same-key documents are held back vs. combined, and how to pair it with a delta-updates materialization for a fully un-reduced stream.
- **README.md** (the Reduction Strategies landing page) — one-line pointer from the "Reduction guarantees" section to the new content.
- **crates/doc/src/reduce/strategy.rs** — drops `EXPERIMENTAL: This keyword may be removed in the future.` from the `associative` field comment (comment-only, no behavior change).

## Why

Recurring support question: "can a collection be configured to not deduplicate at all on the key?" The mechanism (`associative: false` + delta updates) was undocumented. @jgraettinger confirmed in [Slack](https://estuaryworkspace.slack.com/archives/C08LJ60MDBQ/p1780492513240229) the flag is "very stable" and reasonable to document for advanced users — which also makes the in-code "may be removed" note inaccurate.

The wording follows Johnny's description of the actual mechanism: `associative: false` tells the reduction machinery that same-key documents cannot be safely combined incrementally, so they're held back until a full reduction against the base document can be performed (or never, under delta updates).

@jgraettinger — flagging the comment removal for your approval since you confirmed the stability.

## Test plan

- [ ] Render preview builds cleanly
- [ ] "Disabling associative reduction" anchor resolves from the README link

### DIFF
--- a/site/docs/concepts/advanced/reduction-strategies/firstwritewins-and-lastwritewins.md
+++ b/site/docs/concepts/advanced/reduction-strategies/firstwritewins-and-lastwritewins.md
@@ -33,3 +33,54 @@ tests:
         documents:
           - { key: "key", fww: "one", lww: "two" }
 ```
+
+## Disabling associative reduction
+
+By default, Estuary assumes `lastWriteWins` reductions are **associative** — that
+documents sharing a collection key can be combined incrementally, in any grouping,
+without changing the final result (see [Reduction guarantees](./#reduction-guarantees)).
+This is what produces one row per key: same-key documents are progressively combined
+down to a single value.
+
+Setting `associative: false` tells the runtime that same-key documents **cannot** be
+safely combined incrementally. Instead of collapsing them, the runtime holds each
+document back until a full reduction against the base document can be performed — or,
+if the materialization uses [delta updates](/concepts/materialization/#delta-updates),
+combines them never.
+
+```yaml
+collections:
+  - name: example/reductions/no-dedup
+    schema:
+      type: object
+      # Applied at the document root, this disables incremental combining
+      # on the collection key.
+      reduce:
+        strategy: lastWriteWins
+        associative: false
+      properties:
+        id: { type: integer }
+      required: [id]
+    key: [/id]
+```
+
+To preserve **every** document for a key end-to-end (no deduplication at any stage),
+pair `associative: false` on the collection with a **delta-updates** materialization:
+
+* `associative: false` stops same-key documents from being combined as they're
+  written to and read from the collection.
+* Delta updates skips the destination's load-reduce-store cycle, so the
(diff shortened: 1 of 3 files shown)


site/docs/concepts/advanced/reduction-strategies/firstwritewins-and-lastwritewins.md
```diff
@@ -33,3 +33,62 @@ tests:
...
+document back until a full reduction against the base document can be performed — or,
+if the materialization uses [delta updates](/concepts/materialization/#delta-updates),
+combines them never.
+
+```yaml
+collections:
+  - name: example/reductions/no-dedup
+    schema:
+      type: object
+      # Applied at the document root, this disables incremental combining
```
jgraettinger: There's nothing special about the document root. Any location may be non-associative, and abort reduction of the entire document.

site/docs/concepts/advanced/reduction-strategies/firstwritewins-and-lastwritewins.md
```diff
@@ -33,3 +33,62 @@ tests:
...
+        id: { type: integer }
+      required: [id]
+    key: [/id]
+```
+
+To preserve **every** document for a key end-to-end (no deduplication at any stage),
+you need both `associative: false` on the collection **and** a **delta-updates**
+materialization, because reductions happen at more than one place:
+
+* **At capture time**, documents are combined with a *partial* (associative)
```
jgraettinger: Also true of derivations, not just captures. Derivations combine over multiple documents having the same key, in the same derivation transaction.

jgraettinger (approved):

site/docs/concepts/advanced/reduction-strategies/firstwritewins-and-lastwritewins.md
jwhartley: Good point — dropped the "document root" framing. Added a note that the annotation works at any location and that a non-associative reduction anywhere holds back combining of the whole document, with the root example as the simplest way to cover the collection key. (f8d935f)

site/docs/concepts/advanced/reduction-strategies/firstwritewins-and-lastwritewins.md
jwhartley: Added — bullet now reads "At capture or derivation time" and notes derivations combine same-key docs within a transaction. (f8d935f)

PR #3009, opened 2026-06-09
doc: drop EXPERIMENTAL note from associative reduction keyword

jwhartley:
## What

Removes the `EXPERIMENTAL: This keyword may be removed in the future.` line from the doc comment on the `associative` field of `LastWriteWins` (`crates/doc/src/reduce/strategy.rs`).

No behavior change — comment only.

## Why

Per @jgraettinger in [Slack](https://estuaryworkspace.slack.com/archives/C08LJ60MDBQ/p1780492513240229), the `associative` keyword is "very stable." We're documenting `associative: false` for advanced users (estuary/flow#3008), so the in-code "may be removed" note now contradicts both the guidance and the public docs.

@jgraettinger — flagging for your approval since you confirmed the stability.

### DIFF
--- a/crates/doc/src/reduce/strategy.rs
+++ b/crates/doc/src/reduce/strategy.rs
@@ -107,7 +107,6 @@ pub struct LastWriteWins {
     /// The default is true. When set to false, then unequal left- and right-hand
     /// values may not reduce associatively and both documents must be retained
     /// until a full reduction can be performed.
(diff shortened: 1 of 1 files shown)


jwhartley:
Folding the one-line comment change into the docs PR #3008 — they're tightly related and easier to review together.

PR #3013, opened 2026-06-09
supabase: scoped refresh tokens so CI writes via PostgREST instead of direct psql

skord:
**Description:**

Adds a Supabase-native path for CI to write the control-plane database over PostgREST (HTTPS) instead of a direct psql connection on port 5432, ahead of the database network allowlist. A CIDR allowlist cannot cover GitHub-hosted runner IPs, which are a large, weekly-rotating pool of Azure addresses shared by every GitHub Actions tenant.

A `refresh_tokens` row may now carry an optional `pg_role`. `generate_access_token` stamps it into the access token's `role` claim, and PostgREST runs under that role. Two write RPCs expose exactly the operations CI performs today, each restricted to one role:

- `requeue_connector_tag(image_name, image_tags[])` for `github_action_connector_refresh`
- `replace_data_plane_releases(payload jsonb)` for `data_plane_releases_ci`

The credential is a single revocable `refresh_tokens` row (delete to revoke), rather than a long-lived JWT signed with the shared JWT secret, which is effectively unrotatable because it also signs every anon key, the service_role key, and all user sessions.

**Workflow steps:**

An `ops/` admin mints a token once per role:

```sql
select create_scoped_refresh_token('github_action_connector_refresh', '100 years', 'CI connector tag refresh');
select create_scoped_refresh_token('data_plane_releases_ci', '100 years', 'ops data_plane_releases');
```

CI exchanges it for a one-hour access token via `POST /rest/v1/rpc/generate_access_token`, then calls `requeue_connector_tag` / `replace_data_plane_releases` with that token as the bearer. The matching workflow changes ship as separate PRs in connectors, airbyte, source-salesforce-next, and ops. This PR is the additive database side and is safe to deploy on its own: nothing calls the new functions until those PRs merge.

None.

**Notes for reviewers:**

- The `generate_access_token` change is a single line: `role` becomes `coalesce(rt.pg_role, 'authenticated')`. When `pg_role` is null the emitted claims are byte-identical, and a pgTAP test asserts that no-regression explicitly. This is the only edit to a shared auth-path function.
- `create_scoped_refresh_token` is admin-gated (`ops/` admin, the same check used by `create_data_plane` and `update_l2_reporting`) and hard-allowlists the two target roles. Without the allowlist it would be a privilege-escalation path.
- The two scoped roles are granted to `authenticator` so PostgREST can `SET ROLE` to them, the same wiring `dekaf` already uses in production.
- Tokens are owned by a dedicated grant-less `ci-bot@estuary.dev` `auth.users` row, so an access token is inert if presented to agent-api.
- pgTAP coverage: null-`pg_role` unchanged, scoped round trip (token role claim then `requeue_connector_tag` flips `job_status` to queued), admin and allowlist guards, and `authenticator` membership of both roles. Full `ci:sql-tap` suite passes.
- Deploy ordering: this should be live in prod before the workflow PRs merge; additive and safe alone.

### DIFF
--- a/supabase/migrations/20260608120000_scoped_ci_tokens.sql
+++ b/supabase/migrations/20260608120000_scoped_ci_tokens.sql
@@ -1,0 +1,218 @@
+begin;
+
+-- Lets specific database roles be reached over PostgREST using a revocable refresh
+-- token, instead of a direct psql/5432 connection. This is so that CI writes which
+-- currently connect directly to the database survive the upcoming network allowlist,
+-- which cannot cover GitHub-hosted runner IPs.
+--
+-- A refresh token may carry an optional `pg_role`. generate_access_token stamps it
+-- into the access token's `role` claim, and PostgREST uses that claim for SET ROLE.
+-- The credential is revoked by deleting the single refresh_tokens row, which is why
+-- this is preferred over a long-lived JWT signed with the (effectively unrotatable)
+-- shared JWT secret.
+
+-- Null preserves all existing behavior; only machine credentials set this.
+alter table public.refresh_tokens
+  add column pg_role text;
+
+comment on column public.refresh_tokens.pg_role is
+  'Optional Postgres role stamped into the access token `role` claim by generate_access_token. '
+  'Null yields the default ''authenticated'' role. Used for scoped, revocable machine credentials.';
+
+-- Identical to the prior definition except the `role` claim, which now honors a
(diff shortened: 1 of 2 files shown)


jgraettinger (approved):
LGTM

@GregorShear FYI -- we should merge this first, but it has bearing on the GraphQL migration you're working on.

PR #3019, opened 2026-06-11
Runtime V2: Emit valid stats for empty txs in derivations

### LINKED ISSUE #3017: Runtime V2: Derivations emit invalid stats documents on empty transactions
### Summary

Every Flow task writes a stats document (`ops-catalog/stats.schema.yaml`) per
transaction into its tenant's `ops/tasks/...` stats journal. The v2 runtime
(`crates/runtime-next`) serializes a **derivation's** stats incorrectly when a
transaction commits with zero output documents: the doc contains `"derive":
{"published": {}, "out": {}}`, but the schema's `docsAndBytes` definition
requires `docsTotal` and `bytesTotal` whenever the object is present
(`stats.schema.yaml:155-157`).

Stats docs are written unvalidated, so nothing fails at the source. The failure
shows up downstream: the `ops/rollups/L1/.../catalog-stats` derivation validates
its source documents on read, and one invalid doc permanently crash-loops it
(`source document failed validation ... Missing required property: bytesTotal`).
Since the bad doc is already committed to the journal, there's no recovery short
of manual intervention — catalog stats stop updating for the whole data plane.

Example poison doc, observed locally:

```json
{"ts":"2026-06-10T23:45:09.719...","shard":{"name":"test/soak/accounts",...},"txnCount":1,"derive":{"out":{}}}
```

### Root cause

`crates/runtime-next/src/leader/derive/fsm.rs` (`build_stats_doc`, ~line 890)
sets the docsAndBytes fields unconditionally:

```rust
published: Some(extents.published),
out: Some(extents.drained),
```

Proto3 JSON serialization skips zero-valued scalars, so a zero-valued
`DocsAndBytes` message serializes as `{}`. Capture and materialize don't have
this bug — their builders route every docsAndBytes field through
`ops::merge_docs_and_bytes` (`crates/ops/src/lib.rs:165`), which leaves the
field unset (omitted, which is schema-valid) when counts are zero. Derive uses
that helper for per-transform `input` but not for the top-level
`published`/`out`.

Latent second hazard in the same function: if a shard reports a binding with
zero sourced docs, the transform entry serializes without `input`, violating
`transformStats`'s `required: [input]` (`stats.schema.yaml:178-179`). Not yet
observed, but should be fixed in the same pass.

### Reproduction

Deterministic: publish a v2 derivation whose transform reads documents but
publishes nothing (e.g. a sqlite transform with `SELECT ... WHERE false`), then
write one document to its source collection. The transaction commits (the
checkpoint advances on the input), output extents are zero, and the resulting
stats doc is invalid. Verify by validating the doc against `stats.schema.yaml`,
or watch the local catalog-stats rollup shard start failing in reactor logs.

No-catalog-change variant (how it was first observed): restart the shard of a
capture feeding a v2 derivation. On recovery the capture re-writes its ACK
intents into the source journals; the derivation reads the ACK-only batch and
commits a zero-input, zero-output transaction.

### Suggested fix

Make derive symmetric with capture/materialize: route `published` and `out`
through `ops::merge_docs_and_bytes` so zero means omitted, and either skip
transform entries with zero sourced docs or emit an explicit `input: {docsTotal:
0, bytesTotal: 0}`. Add a snapshot test for the zero-output stats doc, asserting
it validates against the schema.

dgreer-dev:
Derivations in the v2 runtime emit stats docs with `"published": {}` / `"out": {}` on
zero-output transactions, violating the stats schema's `docsAndBytes` and permanently
crash-looping the downstream catalog-stats rollup once one is committed.

Route the top-level extents through `ops::merge_docs_and_bytes` (zero ⇒ omitted),
matching capture/materialize and v1 derive, and skip transform entries with zero
sourced docs so the schema-required `input` can never be absent.

Adds a regression test which validates the empty-transaction stats doc against the
actual `ops-catalog/stats.schema.yaml`. Fixes #3017.

### DIFF
--- a/crates/runtime-next/src/leader/derive/fsm.rs
+++ b/crates/runtime-next/src/leader/derive/fsm.rs
@@ -848,6 +848,11 @@ fn build_stats_doc(
     let mut max_published_at: uuid::Clock = uuid::Clock::default();
 
     for (binding_index, binding) in &extents.bindings {
+        // Skip bindings which sourced no documents: the stats schema requires
+        // `input` of each transform entry
+        if binding.sourced.docs_total == 0 {
+            continue;
+        }
         let transform_name = task
             .binding_transform_names
             .get(*binding_index as usize)
@@ -879,6 +884,12 @@ fn build_stats_doc(
 
     let open_seconds_total = uuid::Clock::delta(extents.close, extents.open).as_secs_f64();
 
+    // Leave zero-valued extents unset: proto3 JSON serialization would emit
+    // them as `{}`, which violates the stats schema's `docsAndBytes`
+    let (mut published, mut out) = (None, None);
+    ops::merge_docs_and_bytes(&extents.published, &mut published);
+    ops::merge_docs_and_bytes(&extents.drained, &mut out);
+
     Ok(ops::proto::Stats {
         meta: Some(ops::proto::Meta {
             uuid: String::new(), // Stamped by Publisher::enqueue()
@@ -889,8 +900,8 @@ fn build_stats_doc(
         txn_count: 1,
         derive: Some(ops::proto::stats::Derive {
             transforms,
-            published: Some(extents.published),
-            out: Some(extents.drained),
+            published,
+            out,
             last_published_at: if max_published_at != uuid::Clock::default() {
                 max_published_at.to_pb_json_timestamp()
             } else {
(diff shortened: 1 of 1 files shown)


crates/runtime-next/src/leader/derive/fsm.rs
```diff
@@ -1510,4 +1521,65 @@ mod tests {
...
+            "build": "00aabbccddeeff00"
+          },
+          "ts": "2023-11-14T22:13:20+00:00",
+          "openSecondsTotal": 1.0,
+          "txnCount": 1,
+          "derive": {}
+        }
+        "#);
+
+        // Bundle the shard schema at the canonical $id which the stats
```
williamhbaker: I'd probably take this last part of the test out of the committed changes. It's good as a local validation but the relative file paths to the schemas seem fragile, and we could drop the new dev dependency. The snapshot assertion is sufficient IMO.

williamhbaker (approved):
LGTM, thanks! See comment about the schema validation part of the test.

crates/runtime-next/src/leader/derive/fsm.rs
dgreer-dev: yep okay you got it.  I liked doing it against the real schema but yeah this is more robust.

PR #3020, opened 2026-06-11
Refresh token GraphQL operations and token exchange endpoint

GregorShear:
Adds GraphQL operations for managing refresh tokens (refreshTokens query, createRefreshToken / revokeRefreshToken mutations) and a POST /api/v1/auth/token REST endpoint that exchanges a refresh token for a signed access token. The endpoint transitionally delegates to the SQL generate_access_token function, which existing PostgREST clients still depend on; the plan is to migrate those callers here and then retire the SQL function.

Also reworks how the Envelope extractor handles refresh tokens presented as bearer credentials. Previously it called generate_access_token to mint a signed JWT and then immediately verified that JWT — a sign/verify round trip that crosses no trust boundary, since the token never leaves the process. The envelope now validates the credential directly against the refresh_tokens table and constructs claims from the verified row, via a new Verified::assert_stateful_authentication seam in the tokens crate that keeps 'holding a Verified<T>' as a type-level proof of authentication.

One edge case behavior change: single-use refresh tokens are no longer accepted as direct bearer credentials. The previous path consumed them on first use and discarded the rotated secret, silently bricking the token; they now fail authentication outright. Single-use tokens remain exchangeable via /api/v1/auth/token, which returns the rotation to the caller.

Because the bearer path no longer depends on pgjwt signing (absent from the sqlx::test polyfill), its happy path is now covered by an integration test: create a token via GraphQL, authenticate with it as a bearer, and observe identity and usage stamping.


crates/tokens/src/lib.rs
```diff
@@ -212,3 +263,84 @@ pub fn now() -> DateTime {
...
+// SECURITY GUARD: a census of every `Verified::assert_authenticity` call site.
+//
+// Each call site is an authentication entry point (see the docs on that
+// function): code past it runs with the request treated as authenticated. This
+// test enumerates the authorized callers and fails if the set — or the number
+// of calls in any file — changes. A new way to authenticate a request thus
+// cannot land without a human editing the allowlist below, which is the cue to
+// get security review. An agent copying an existing call site trips this.
+#[cfg(test)]
+mod authenticity_census {
```
GregorShear: Maybe this whole test is a little extra but i'm trying to get a future human's attention in case an agent adds more callers to the `assert_authenticity()` function

crates/control-plane-api/src/server/public/graphql/refresh_tokens.rs
```diff
@@ -0,0 +1,465 @@
...
+            default_with = "String::from(\"P90D\")"
+        )]
+        valid_for: String,
+        #[graphql(default = true)] multi_use: bool,
+        #[graphql(default)] detail: Option<String>,
+    ) -> async_graphql::Result<RefreshTokenResult> {
+        let env = ctx.data::<crate::Envelope>()?;
+        let claims = env.claims()?;
+
+        let row = sqlx::query!(
```
jshearer, line 140: The `valid_for: String` is passed straight into `$3::text::interval` with no positivity or range check. Three reachable bad outcomes for an authenticated caller:
- `validFor: "PT0S"` stores `interval '0'`, which is the exact sentinel `revokeRefreshToken` uses (`refresh_tokens.rs:187`). The token is then filtered out of the listing (`AND valid_for <> interval '0'`, line 73), can never authenticate (`(updated_at + valid_for) > now()` is false), and `revokeRefreshToken` returns "not found". The user holds a real `id`+`secret` for a token they cannot see, use, or revoke.
- A negative interval (`"P-1D"`) is accepted and yields a token that is born expired.
- An absurd interval (`"P300000Y"`) inserts fine, then `now() > updated_at + valid_for` (list query, line 70) and the bearer-auth predicate overflow with Postgres "timestamp out of range", returning a 500 on the user's own `refreshTokens` listing until the row is removed directly.
Validate at the resolver boundary (parse and assert a positive, bounded interval) and return a domain error.

crates/control-plane-api/src/server/public/token_exchange.rs
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    let response = sqlx::query!(
+        "select generate_access_token($1, $2) as token",
+        refresh_token_id as models::Id,
+        secret,
+    )
+    .fetch_one(&app.pg_pool)
+    .await
+    .map_err(|err| {
```
jshearer, line 61: Returning every error as 401 is probably not quite right, nor should we be returning the actual DB error string I don't think, right?

jshearer (commented):
I wanted to get these out, more to come in a bit

crates/control-plane-api/src/server/public/graphql/refresh_tokens.rs
```diff
@@ -0,0 +1,465 @@
...
+        let claims = env.claims()?;
+
+        connection::query_with::<TimestampCursor, _, _, _, async_graphql::Error>(
+            after,
+            None,
+            first,
+            None,
+            |after, _, first, _| async move {
+                let after_created_at = after.map(|c| c.0);
+                let limit = first.unwrap_or(DEFAULT_PAGE_SIZE);
```
jshearer, line 57: I think this may be pre-existing and you just copied it from somewhere else, but:

`refreshTokens(first: 0)` returns empty edges with `hasNextPage: true`. `limit = first.unwrap_or(25)` becomes 0, so `LIMIT $3 + 1` fetches 1 row, `has_next = rows.len() > limit` is `1 > 0` (true), and `.take(0)` yields no edges. A client paging with `first: 0` sees "more pages exist" with no cursor to advance and stalls.

crates/control-plane-api/src/server/public/token_exchange.rs
```diff
@@ -0,0 +1,84 @@
...
+// and then retire the SQL function, folding refresh-token minting into an
+// application-layer path. New clients should target this endpoint rather
+// than PostgREST.
+async fn exchange_refresh_token(
+    app: &crate::App,
+    refresh_token_id: models::Id,
+    secret: &str,
+) -> Result<axum::Json<TokenResponse>, crate::ApiError> {
+    #[derive(Debug, serde::Deserialize)]
+    struct SqlResponse {
```
jshearer: `SqlResponse` is a structural duplicate of `TokenResponse`. Both carry `access_token: String` and `refresh_token: Option<RefreshTokenResponse>`; the handler deserializes into `SqlResponse` then field-copies into `TokenResponse`. `SqlResponse` exists only because `TokenResponse` derives `Serialize` but not `Deserialize`. Add `serde::Deserialize` to `TokenResponse` and deserialize the SQL JSON straight into it, dropping the intermediate struct and the copy.

crates/control-plane-api/src/server/public/graphql/refresh_tokens.rs
```diff
@@ -0,0 +1,465 @@
...
+
+                let rows = sqlx::query!(
+                    r#"
+                    SELECT
+                        id AS "id!: models::Id",
+                        detail,
+                        created_at AS "created_at!: chrono::DateTime<chrono::Utc>",
+                        updated_at AS "updated_at!: chrono::DateTime<chrono::Utc>",
+                        multi_use AS "multi_use!: bool",
+                        valid_for::text AS "valid_for!: String",
```
jshearer: The listing returns `valid_for::text` (line 68), which renders in the session's `IntervalStyle`. The default (`postgres`) produces `90 days`, not the `P90D` this field's own doc string advertises (line 126, "ISO 8601 duration ... e.g. P90D"), so the value a client reads back from `validFor` does not match the format it submitted, and the rendered string depends on session config we do not set anywhere (pool, role, or database).

Minor, and there is no clean inline fix: `IntervalStyle` is a session GUC, so it cannot be cast per-expression, `SET` cannot ride along on the prepared statement, and `to_char` cannot emit ISO 8601 durations. The options:
 
- Format the interval as ISO 8601 in Rust from the parsed value. This is the only fix local to this query and the only one that guarantees the documented format.
- Set `IntervalStyle = 'iso_8601'` at the pool/`after_connect` or role/database level. This also changes how `connectors/tags.rs:58` (`default_capture_interval::text`) and every other `::text` interval render, so it is a deliberate codebase-wide choice rather than a local one.
- Leave the rendering as-is and document that `validFor` is normalized to the server's interval style, pinning it with a round-trip assertion in the existing test.

I don't like any of these options TBH. Can you comment on how, if anywhere, this return value is actually consumed? Maybe it's just presented as a string in the frontend, in which case it's probably fine to leave as is. It's mainly if there's anything programatic trying to read it that it gets sketchy...

crates/tokens/src/lib.rs
```diff
@@ -212,3 +263,84 @@ pub fn now() -> DateTime {
...
+    use std::collections::BTreeMap;
+    use std::path::{Path, PathBuf};
+
+    // Authorized call sites, as (workspace-relative path, number of calls).
+    // Editing this list is a security-relevant change — review before you do,
+    // do not adjust it merely to make the test pass.
+    const AUTHORIZED_CALLERS: &[(&str, usize)] = &[
```
jshearer: We talked over huddle, let's back out this whole change and put refresh token authentication back to minting a throw-away JWT. I agree there's design smell here, but let's decouple fixing it from this PR

crates/control-plane-api/src/server/public/graphql/refresh_tokens.rs
GregorShear: Looks like we never even display the value in the frontend - i'll just remove it.

crates/control-plane-api/src/server/public/graphql/refresh_tokens.rs
GregorShear: ah thanks - i'm enforcing iso 8601 and a min/max for api_keys, but forgot to do it here.

crates/control-plane-api/src/server/public/token_exchange.rs
GregorShear: good catch, thank you

crates/control-plane-api/src/server/public/graphql/refresh_tokens.rs
GregorShear: i have to think about this one - i don't know what a client would expect if it asks for zero rows and then tries paging... 

not sure what the fix would be here so i'm inclined to leave it as is

crates/tokens/src/lib.rs
GregorShear: yep we'll come back to this. reverted for now

crates/control-plane-api/src/server/public/graphql/refresh_tokens.rs
```diff
@@ -0,0 +1,452 @@
...
+            detail.as_deref(),
+        )
+        .fetch_optional(&env.pg_pool)
+        .await
+        .map_err(|err| {
+            // Postgres raises SQLSTATE 22007 for malformed interval input,
+            // which is a client error rather than an internal fault.
+            if err.as_database_error().and_then(|e| e.code()).as_deref() == Some("22007") {
+                async_graphql::Error::new("validFor must be a valid ISO 8601 duration (e.g. P90D)")
+            } else {
```
jshearer, line 172: Why do we log the db error and then return a sanitized response [over here](https://github.com/estuary/flow/pull/3020/changes#diff-a9a21d7e32e0406deb932bdf7cdfcc0005dc14edc923b561c93cdd5a09e36f53R74-R77) but return the full error string here? I feel like we should pick one or the other, and the warning + sanitized user-facing error seems like the right one

crates/control-plane-api/src/server/public/graphql/refresh_tokens.rs
```diff
@@ -0,0 +1,452 @@
...
+            multi_use,
+            valid_for,
+            detail.as_deref(),
+        )
+        .fetch_optional(&env.pg_pool)
+        .await
+        .map_err(|err| {
+            // Postgres raises SQLSTATE 22007 for malformed interval input,
+            // which is a client error rather than an internal fault.
+            if err.as_database_error().and_then(|e| e.code()).as_deref() == Some("22007") {
```
jshearer: lol this is a bit ridiculous at this point and is totally a nit but:

> The `map_err` only matches SQLSTATE `22007`. Inputs extreme enough to overflow interval *parsing* (e.g. `P300000000000Y`) raise `22015` (interval_field_overflow), which falls into that leaky `else` branch. Note your original example `P300000Y` is fully fixed — it parses fine and the `366 days` cap rejects it; this is a strictly-more-extreme variant. The fix is one line: match `22015` alongside `22007`.

jshearer (approved):
Couple nitpicks, otherwise LGTM!

crates/control-plane-api/src/server/public/graphql/refresh_tokens.rs
jshearer: Yeah I'm fine with that. I think you might explicitly handle this somewhere else, but "play stupid games, win stupid prizes" is good enough for me here

PR #3022, opened 2026-06-11
data-plane-controller: dns and s3 passthroughs

qaoj:
**Description:**

New passthroughs for some aws options

**Workflow steps:**

Set them

N/A

**Notes for reviewers:**

N/A


crates/data-plane-controller/src/shared/stack.rs
```diff
@@ -122,6 +122,19 @@ pub struct DataPlane {
...
+    // When true, sets `enableDnsHostnames` on the AWS VPC. Required for
+    // Route 53 Private Hosted Zones associated with this VPC to resolve.
+    // Side effect: AWS publishes externally resolvable
+    // `ec2-<ip>.compute.amazonaws.com` hostnames for every public-IP
+    // instance. Opt in only when a PHZ is actually required.
+    #[serde(default, skip_serializing_if = "is_false")]
+    pub enable_dns_hostnames: bool,
+    // When true, attaches the VPC's S3 gateway endpoint to each zone's
+    // private subnet route table so connector containers reach S3 through
+    // the gateway endpoint instead of the NAT gateway. Trade-off: source
```
mwillman-estuary, line 134: Minor - doesn't the destination IP change?  S3 isn't the "source" in this conversation, is it?  Or am I thinking about this wrong?

mwillman-estuary (approved):
Looks minor, approved.

crates/data-plane-controller/src/shared/stack.rs
qaoj: I guess I should have said "Source IP from the perspective of S3"- basically this is for customers that have "Must come from a certain VPC"-type filters on their S3 buckets. This makes it come over the internal VPC rather than over the open internet. Has interactions with the route stuff I've been working on otherwise, but it's a bit much to explain in a comment here lol

PR #3023, opened 2026-06-12
Migrate refresh-token secret hashing from bcrypt to SHA-256

GregorShear:
Refresh-token secrets are high-entropy random values, so bcrypt's slow hashing adds no protection against offline brute force — that only matters for low-entropy (user-provided) passwords — while costing ~1ms of CPU on every verification. The bearer-credential path validates on every request (e.g. OpenMetrics scrapers), so verification cost matters. API key secrets will use the same approach in a future PR.

Existing bcrypt hashes cannot be rewritten in place (we don't hold the plaintexts), so both validators — the SQL generate_access_token function and the Rust bearer-authentication path — dual-read by hash format: bcrypt hashes start with '$2', hex SHA-256 digests are 64 hex characters. A successful bcrypt verification rewrites the row to SHA-256 while the plaintext is in hand. This rehash-on-use is load-bearing for the migration's completion: multi-use tokens re-arm their validity window on every use, so an actively-used bcrypt token would otherwise never expire and the bcrypt branch could never be removed. With it, bcrypt rows drain via rehash or expiry, after which the bcrypt branches can be dropped — planned alongside retiring generate_access_token.

All three writers (the SQL create_refresh_token function, single-use rotation inside generate_access_token, and the GraphQL createRefreshToken mutation) now write SHA-256. Secrets are also upgraded from random UUIDs (~122 bits) to 256-bit random hex via gen_random_bytes: the fast-hash rationale rests on secrets being high-entropy, the bump is free, and hex reads as an opaque secret rather than a UUID-shaped identifier. The change is invisible to clients, which treat the secret as an opaque string.

Covered by a new pgTAP case (legacy bcrypt row exchanges, is rewritten to SHA-256, and exchanges again) and a Rust integration test exercising the same migration through the bearer path.


GregorShear:
I still think this is a good idea but i'm holding it until after we retire `generate_access_token` and `create_refresh_token` sql functions

PR #3024, opened 2026-06-12
docs: add Stripe account API version prerequisite to stripe-realtime

jwhartley:
The source-stripe-native connector lists subscriptions with `status=all` so canceled subscriptions are captured too. Stripe accounts whose default API version is older than [2016-07-06](https://docs.stripe.com/changelog/2016-07-06/filter-canceled-subscriptions-retrieve-individually) reject that filter with `Invalid status: must be one of active, past_due, unpaid, ...`, which crash-loops the Subscriptions, SubscriptionItems, and UsageRecords backfills. The connector doesn't pin a `Stripe-Version` header, so the account default applies (same failure mode as stripe/stripe-python#687).

This hit a user in support this week and the prerequisites only mentioned the API key, so this adds the minimum account API version, the symptom to recognize it by, and where to upgrade in the Stripe dashboard.


aeluce:
Is there anything still unresolved for this following #3055?

jwhartley:
#3055 covers adding the option and info, all set here.

PR #3027, opened 2026-06-12
runtime: periodically store primary FSMHints of V2 shards

williamhbaker:
**Description:**

Gazette's consumer transaction loop stores primary hints in-between transactions, but V2 shards bypass that loop entirely: their hints were written only at graceful session cycles, so recovery logs of long-lived tasks grew without bound and could never be pruned.

Stand in with a `taskBase` loop which transcribes gazette's unexported `consumer.storeRecordedHints` on its same five-minute interval.

**Workflow steps:**

**Notes for reviewers:**

### DIFF
--- a/go/runtime/task.go
+++ b/go/runtime/task.go
@@ -3,6 +3,7 @@ package runtime
 import (
 	"context"
 	"database/sql"
+	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


jgraettinger (approved):
LGTM

Totally forgot about this. We'll do it better when we re-do recovery log recording.

PR #3028, opened 2026-06-12
control plane: tenant filtering for GraphQL listings

GregorShear:
Stacked on #2943 (`greg/service-accounts-phase1`).

## Summary

Adds a reusable **tenant filter** for GraphQL listings and wires it into the `serviceAccounts` query: `serviceAccounts(filter: { tenant: { eq: "acmeCo/" } })`.

## Design

`authorized_prefixes` gains a `tenant` parameter backed by a new `tenant_prefixes` helper. `tenant_prefixes` computes the tenant's own reach through the role-grant graph: the tenant prefix itself, plus every prefix reachable from it at the requested capability, using the same BFS as authorization (`RoleGrant::reachable_nodes`) — admin grants delegate and extend the chain, while plain read/write grants are leaves.

The caller's authorized prefixes are then intersected with the tenant's reach, keeping the narrower prefix of each overlapping pair. Every returned prefix lies within both sets, so the filter can never widen the caller's access, and the listing never extends past the tenant's scope — including the case where a caller's broad grant (e.g. all of `betaCo/`) is clamped to the tenant's narrower reach (`betaCo/nested/`).

The existing prefix-overlap filter (`catalogPrefix` / `catalogPrefixOrName` on invite links and alert configs) is marked **deprecated** in favor of the tenant filter, with TODOs noting the migration path for those callers.

## Testing

Unit tests cover `tenant_prefixes` (transitive admin chains, leaf grants, fine-grained capability filtering, cross-tenant exclusion, bare tenants) and the tenant clamping in `authorized_prefixes` (narrows to tenant scope, cannot widen access, keeps shared reach, clamps broad grants). The `serviceAccounts` lifecycle test exercises the filter end-to-end.


GregorShear:
putting this off for a while, closing

PR #3029, opened 2026-06-12
runtime-next: automatically split journals under sustained append-rate throttling

dgreer-dev:
## What

Gazette rate-limits appends per journal, so a hot collection partition
eventually throttles all of its writers. Splitting a partition journal in two
is additive — each half gets its own budget — but today that's a manual
`flowctl` operation. This makes it automatic: the V2 runtime observes append
throttling and splits hot partition journals on its own.

## How it works

- Each shard watches per-journal append throttling (the broker's
  `delayed_chunks` signal) and tracks which journals stay throttled.
- When a partition journal is persistently throttled, the shard splits it at its
  key-range midpoint — keeping the original (lower half) and adding one new
  journal (upper half).
- A minimum-width floor caps automatic splitting at ~8 even journals per
  partition, so a low-cardinality key set can't split forever; manual splitting
  remains available beyond that.
- Always on for tasks on the V2 runtime — no separate feature flag. Reuses the
  runtime's existing journal `APPLY` grant and the publisher's partition
  machinery; no new privileges.

## Scope

Targets collection partition journals (capture- and derivation-produced).
Recovery-log / shard-level pressure is out of scope.

## Commits

1. Observe-only throttle pipeline.
2. Split action + trigger.
3. Integration tests, logging, README.

## Testing

Unit tests cover the throttle policy, the min-width gate, and outcome handling.
End-to-end tests run against a real etcd + gazette broker (`e2e-support`):

- **`journal_auto_split_converges_to_floor`** — sustained over-rate load splits a
  partition repeatedly until it converges to an even ~8-way tiling and stops at
  the floor; below-rate load never splits; new journals take writes.
- **`concurrent_journal_split_loses_cas_race`** — two shards racing to split the
  same journal: etcd admits exactly one Apply, the loser observes `Lost` and then
  proceeds against the split layout its watch re-observes.
- **`journal_split_lost_cas_is_non_destructive`** — a `Lost` from a stale-revision
  CAS leaves the parent journal untouched and still writable (the two-change
  Apply is atomic — nothing partial lands).
- **`journal_auto_split_preserves_document_completeness`** — every document
  written before, during, and after a cascade is present exactly once across the
  resulting journals (no loss, no duplication).
- **`journal_auto_split_respects_logical_partitions`** — throttling one
  partition-field value splits only that logical partition's journal (field
  labels preserved); other partitions are untouched.


crates/runtime-next/src/shard/split_policy.rs
```diff
@@ -0,0 +1,456 @@
+//! Pure throttle policy that decides when a collection's journals are
```
williamhbaker, line 1: Might make sense to move this under `shard/`, since it is used at that level

williamhbaker (commented):
Look a quick look, seems pretty good so far.

crates/runtime-next/src/shard/capture/actor.rs
```diff
@@ -278,6 +298,25 @@ impl Actor {
         Ok((db, shapes))
     }
 
+    /// Drain this transaction's per-journal throttle samples from the publisher
+    /// and feed them into the long-lived split policy, then start a split of
+    /// at most one persistently-throttled journal — off the hot path, parked
+    /// as `split_fut`.
+    fn observe_throttle(&mut self) {
+        let Some(publisher) = self.publisher.as_mut() else {
+            return;
```
williamhbaker: This state should be impossible, so I'd do an `expect` here rather than a silent return.

crates/runtime-next/src/shard/derive/actor.rs
```diff
@@ -416,6 +435,25 @@ impl Actor {
         Ok((phase, false))
     }
 
+    /// Drain this transaction's per-journal throttle samples from the publisher
+    /// and feed them into the long-lived split policy, then start a split of
+    /// at most one persistently-throttled journal — off the hot path, parked
+    /// as `split_fut`.
+    fn observe_throttle(&mut self) {
+        let Some(publisher) = self.publisher.as_mut() else {
+            return;
```
williamhbaker: This state should be impossible, so I'd do an `expect` here rather than a silent return.

crates/runtime-next/tests/split_e2e.rs
```diff
@@ -0,0 +1,1042 @@
+//! End-to-end integration test of automatic journal splitting (design doc
+//! "stream G2"): a real etcd + gazette broker, a real `Publisher`, and real
```
williamhbaker: Looks like a stale comment here (?)

crates/runtime-next/src/shard/split_policy.rs
```diff
@@ -0,0 +1,496 @@
...
+        // never fed, so they can never accumulate pressure or become due.
+        if self.ignore.contains(journal) {
+            return;
+        }
+
+        let tau_secs = self.config.tau.as_secs_f64();
+        let max_dt = self.config.max_sample_dt;
+        let min_span = self.config.min_observation_span;
+
+        let throttle = self
```
williamhbaker: This isn't on the _hottest_ path, but since it is for every journal on every transaction, perhaps we could save the `journal.to_owned()` allocation here with something like this, for the common case where the journal has already been recorded:

```rust
      let throttle = if let Some(throttle) = self.journals.get_mut(journal) {
          throttle
      } else {
          self.journals
              .entry(journal.to_owned())
              .or_insert_with(|| JournalThrottle {
                  ewma: 0.0,
                  last_ts: now,
                  not_before: now + min_span,
              })
      };
```

williamhbaker: Also something to consider: Maybe that `else` could be `else if throttled` so that assignment to the map is lazy, and we only pay the overhead for journals that actually get throttled at some point? Since most never will. And a final `else { return }` for the common case.

crates/publisher/src/appender.rs
```diff
@@ -21,6 +22,11 @@ pub struct Appender {
     /// Optional registers which must be present for the Append to succeed.
     /// This may be used with cooperative fencing of journals.
     check_registers: Option<Box<broker::LabelSelector>>,
+    /// Accumulated `delayed_chunks` summed across every append RPC issued to
+    /// this journal since the last `take_throttle_samples()`.
+    delayed_chunks: Arc<AtomicI64>,
```
williamhbaker: I think we can simplify here to just use `i64`'s. We'd need a `AppendState::InFlight(tokio::task::JoinHandle<(UpdateFn, i64, i64)>)` and the various mechanical changes that arise from that, but I think it could be a better approach than the `Arc<AtomicI64>>`.

williamhbaker (commented):
Looking really good, just a few comments!

crates/runtime-next/tests/split_e2e.rs
dgreer-dev: whups, yep

crates/runtime-next/src/shard/split_policy.rs
dgreer-dev: Oh I love the `else if throttled` idea.  Not that this was a huge memory cost, but yeah that essentially zeroes the cost of this for the vast vast majority of journals.

williamhbaker (approved):
LGTM

PR #3035, opened 2026-06-15
flowctl: normalize --prefix trailing slash

### LINKED ISSUE #2608: Confusing error message when prefix missing trailing slash
## Problem

When using `flowctl catalog list --prefix Maverick_Automation` (without trailing slash), the error message is:

```
user is not authorized to Maverick_Automation
```

This is confusing because the user IS authorized - the issue is just the missing trailing slash.

## Expected Behavior

Either:
1. Accept the prefix without trailing slash, OR
2. Provide a clearer error message like: "Prefix must end with '/' - did you mean 'Maverick_Automation/'?"

## Steps to Reproduce

```bash
# This fails with confusing auth error:
flowctl catalog list --prefix Maverick_Automation

# This works:
flowctl catalog list --prefix Maverick_Automation/
```

### LINKED ISSUE #2834: flowctl catalog: missing trailing slash on --prefix gives misleading PermissionDenied error
When a user passes `--prefix <tenant>` without a trailing slash, the GraphQL API returns a `PermissionDenied` error even if the user has full admin access. The actual cause is that the `Prefix` type requires a trailing slash — the value is silently passed as-is to the API, which rejects it as an unrecognized/unauthorized prefix name.

**Steps to reproduce:**
```
flowctl catalog pull-specs --prefix AcmeCo
# Error: PermissionDenied: johnsmith@acme.co is not authorized to access prefix or name 'AcmeCo' with required capability read
```

**Expected:**
Either auto-normalize (`AcmeCo` → `AcmeCo/`) or return a clear error: `prefix must end with '/': did you mean 'AcmeCo/'?`

**Workaround:**
```
flowctl catalog pull-specs --prefix AcmeCo/
```

**Location:** `crates/flowctl/src/catalog/list/mod.rs` line ~272, where `models::Prefix::new(prefix)` is called without validating/normalizing the trailing slash. The same fix should apply to any other command using `NameSelector` with `--prefix`.

**Labels:** bug

telcharr:
**Description:**

`flowctl` now normalizes a user-supplied `--prefix` to end with `/`, which `models::Prefix` requires. Previously, passing a prefix without a trailing slash was sent to the API verbatim and rejected as an unauthorized name, causing a misleading `PermissionDenied` error even when the user had full access. Applies across `catalog list`/`pull-specs`/`delete`, `raw gazctl-env`, and `alerts subscriptions`.

**Workflow steps:**

`flowctl catalog pull-specs --prefix AcmeCo` now behaves the same as `--prefix AcmeCo/`. The trailing slash is added automatically with a stderr warning noting the change. Prefixes that already end in `/` are unchanged, so existing usage is unaffected. 

None. The CLI `--prefix` help text already describes prefix semantics, the behavior is just more forgiving now. No difference in contract. 

**Notes for reviewers:**

Closes #2834. Closes #2608

### DIFF
--- a/crates/flowctl/src/alerts/subscriptions/mod.rs
+++ b/crates/flowctl/src/alerts/subscriptions/mod.rs
@@ -103,7 +103,7 @@ struct ListAlertSubscriptions;
 
 async fn do_list(list: &List, ctx: &mut crate::CliContext) -> anyhow::Result<()> {
     let prefix = if let Some(pre) = list.prefix.clone() {
-        pre
+        crate::normalize_prefix(&pre)
     } else {
         let prefixes =
             crate::get_default_prefix_arguments(ctx, models::Capability::Read, 1).await?;
@@ -181,7 +181,7 @@ async fn do_subscribe(create: &SubscribeArgs, ctx: &mut crate::CliContext) -> an
         anyhow::anyhow!("no explicit email argument provided, and unable to determine the user email from the available authentication information. Please try logging in again, or pass an explicit email argument."))?;
 
     let prefix = if let Some(pre) = create.prefix.clone() {
-        pre
+        crate::normalize_prefix(&pre)
     } else {
         let prefix_vec =
             crate::get_default_prefix_arguments(ctx, models::Capability::Read, 1).await?;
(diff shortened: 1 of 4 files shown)


telcharr:
Closing, #3075 covers the `NameSelector` path.

Issue #3032, opened 2026-06-15
Auto-discover removes manual bindings on filesource captures despite addNewBindings: false


jwhartley:
> Priority: Low (control-plane triage, 2026-06-18)
>
> Very few captures use this pattern today, and the multiple-file-source-bindings guide has been updated to steer users away from the broken setup (estuary/flow#3037).

---

## Summary

`addNewBindings: false` is documented and intended to stop auto-discover from changing a capture's binding set, but it does not prevent binding *removal*. The discover-merge reconciler rewrites the binding list to exactly what discovery returns and drops any existing binding whose resource path isn't in that output, regardless of `addNewBindings`. The flag only controls whether *newly discovered* bindings are added enabled vs disabled (`update_only`).

This breaks acutely on filesource captures (`source-s3`, `source-gcs`, `source-azure-blob-storage`, `source-google-drive`), where discovery always returns exactly one binding: the bucket+prefix root. A user who manually splits one bucket into multiple sub-prefix bindings, which is the workflow in the official guide that explicitly tells them to set `addNewBindings: false` to protect those bindings, has every manual binding treated as unmatched and removed on each ~2h auto-discover run, then a single disabled root binding re-added. The pipeline silently stops feeding their collections until they restore the bindings, after which the next run reverts them again. A real capture configured exactly per that guide was clobbered repeatedly.

Guide: https://docs.estuary.dev/guides/flowctl/multiple-file-source-bindings/

## Reproduction

1. Create an S3 capture over a bucket with sub-folders (e.g. Hive-style `key=value/` partitions).
2. Manually configure two bindings, each with a sub-prefix `stream` (e.g. `bucket/a=x/` &rarr; collection A, `bucket/a=y/` &rarr; collection B), set `autoDiscover.addNewBindings: false`, and delete the auto-created root binding.
3. Wait for the next auto-discover run. Both sub-prefix bindings are removed and the root is re-added disabled. History shows `auto-discover changes (0 added, 0 modified, 2 removed, 1 added (disabled))`.

## Root cause

- The removal pass is unconditional and not gated by `update_only`: `crates/control-plane-api/src/discovers/specs.rs:213-227` (`model.bindings = next_bindings` drops any existing binding whose resource path isn't in the discovery output). `addNewBindings: false` sets `update_only = true` (`crates/agent/src/controllers/capture/auto_discover.rs:251`), which is consumed only at `specs.rs:181` to disable newly-added bindings.
[...]

PR #3037, opened 2026-06-16
docs: disable auto-discover for multi-binding file source captures

jwhartley:
Fixes the docs guidance flagged in #3032.

The "Capture Multiple Paths with File Source Connectors" guide tells users to set `addNewBindings: false` so auto-discover "won't overwrite your manual bindings." That flag does not prevent auto-discover from *removing* bindings; it only controls whether newly discovered bindings are added. File source discovery returns a single binding (the bucket/prefix root), so auto-discover removes every manually-added binding on its next run. A capture configured exactly per this guide was clobbered repeatedly.

The guide also incorrectly states that keeping `evolveIncompatibleCollections: true` "preserves schema inference." Schema inference is driven by the collection's read schema (the `flow://inferred-schema` reference) and is independent of auto-discover; that flag only controls collection reset on incompatible key changes.

Changes:
- Recommend disabling auto-discover entirely (`autoDiscover: null`) instead of `addNewBindings: false`.
- Remove the incorrect schema-inference justification.
- Update the web-app tip (uncheck "Automatically keep schemas up to date") and remove the `autoDiscover` block from both worked examples.

This guidance is correct today and stays correct even if the control-plane fix in #3032 lands, at which point disabling auto-discover for file sources becomes the default.

### DIFF
--- a/site/docs/guides/flowctl/multiple-file-source-bindings.md
+++ b/site/docs/guides/flowctl/multiple-file-source-bindings.md
@@ -37,7 +37,10 @@ Open the capture YAML file. Auto-discover periodically re-runs discovery and rew
 Setting `addNewBindings` to `false` does **not** prevent this — it only stops new bindings from being added, not existing ones from being removed. Remove the `autoDiscover` block entirely (or set it to `null`) so auto-discover never runs:
 
 ```yaml
-autoDiscover: null
+captures:
+  acmeCo/sftp-capture:
+    autoDiscover: null
+    endpoint: {...}
 ```
 
 Disabling auto-discover does not affect schema inference, which is driven by your collection's read schema and continues regardless.
--- a/site/docs/guides/flowctl/multiple-file-source-bindings.md
+++ b/site/docs/guides/flowctl/multiple-file-source-bindings.md
@@ -30,18 +30,20 @@ flowctl catalog pull-specs --name your-org/your-capture
 
 This creates a local `flow.yaml` file and subdirectories with your capture's specification.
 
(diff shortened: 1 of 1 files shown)


site/docs/guides/flowctl/multiple-file-source-bindings.md
```diff
@@ -30,18 +30,20 @@ flowctl catalog pull-specs --name your-org/your-capture
...
-Open the capture YAML file. Set `addNewBindings` to `false` so auto-discover won't overwrite your manual bindings. Keep `evolveIncompatibleCollections` enabled to preserve schema inference:
+Open the capture YAML file. Auto-discover periodically re-runs discovery and rewrites the capture's bindings to match what the connector reports. File source connectors only ever discover a single binding (the bucket/prefix root), so on its next run auto-discover will remove the extra bindings you add below.
+
+Setting `addNewBindings` to `false` does **not** prevent this — it only stops new bindings from being added, not existing ones from being removed. Remove the `autoDiscover` block entirely (or set it to `null`) so auto-discover never runs:
 
 ```yaml
-autoDiscover:
-  addNewBindings: false
-  evolveIncompatibleCollections: true
+autoDiscover: null
```
aeluce:
```suggestion
captures:
  acmeCo/sftp-capture:
    autoDiscover: null
    endpoint: {...}
```

Since the `autoDiscover` blocks are removed from the example specs later, it might be useful to expand this snippet to show where `autoDiscover` falls in the spec structure.

aeluce (approved):

PR #3040, opened 2026-06-16
docs: Describe SQL Server replica suport

willdonnelly:
**Description:**

Adds a note to the source-sqlserver-ct docs noting that it can only capture from the primary DB instance.

Also adds a "Read Replica Support" column to the comparison table for all three SQL Server captures since that seems like probably a pretty important dimension of comparison between the three.


aeluce (approved):
🎉

PR #3045, opened 2026-06-17
docs: update Bigtable connector permissions example

mwillman-estuary:
**Description:**

Update the permissions example to reflect the fact that roles/bigtable.admin is a superset of roles/bigtable.user, so adding roles/bigtable.user is redundant

**Workflow steps:**

N/A

N/A

**Notes for reviewers:**

Cross check my logic here -> https://docs.cloud.google.com/iam/docs/roles-permissions/bigtable

Pinging @williamhbaker specifically as he's the original author.


mwillman-estuary:
Don't trust the preview - I force pushed another change, and it doesn't look like that forces a refresh of the preview, so it's out of date.

williamhbaker (approved):

PR #3058, opened 2026-06-18
Service Accounts

GregorShear:
## Service accounts

Non-login identities for CI/CD and automation. A service account is a real `auth.users` row (with a synthetic `@service_accounts.estuary.dev` email) whose access comes entirely from ordinary `user_grants`, so RLS, `user_roles()` resolution, and role traversal work unchanged. It's addressed by a unique catalog name (e.g. `acmeCo/ci-bot`) that acts as a management anchor: holders of the new service-account capabilities on a prefix covering that name manage the account. The account's grants may span any prefixes, independent of the anchor.

Schema: a new `internal.service_accounts` table (unique `catalog_name`, SP-GiST-indexed for prefix scoping) and a nullable `refresh_tokens.created_by` column recording the admin who minted a credential on the account's behalf.

## API keys are refresh tokens

There is no bespoke credential type. `createApiKey` mints a multi-use refresh token owned by the account, bcrypt-hashed at rest, and returns the bearer secret exactly once, in the base64 `{id, secret}` form that `Authorization: Bearer` and `POST /api/v1/auth/token` already accept. `validFor` must be an ISO 8601 duration, positive and at most one year; like any refresh token, the validity window slides with use. Revoking zeroes `valid_for`: the key becomes inert immediately (every request re-verifies against the database) while the row survives for audit. `revokeAllApiKeys` is the credential kill switch.

Service accounts cannot self-manage credentials: `createRefreshToken` and `revokeRefreshToken` reject service-account callers, so a leaked key can't mint its own replacements or revoke outside the admin flow. (A residual self-service surface exists via PostgREST's `refresh_tokens` grants; it disappears with PostgREST.)

## Authorization

Four new fine-grained capabilities (`QueryServiceAccounts`, `CreateServiceAccount`, `CreateApiKey`, `RevokeApiKey`), bundled as `manage_service_account` and folded into `team_admin`:

| Operation | Requires |
|---|---|
| `serviceAccounts` query | `QueryServiceAccounts` on a covering prefix |
| `createServiceAccount` | `CreateServiceAccount` on the catalog name, plus `CreateGrant` on each seeded grant prefix |
| `addServiceAccountGrant` | `CreateGrant` on both the catalog name and the granted prefix |
| `removeServiceAccountGrant` / `removeAllServiceAccountGrants` | `CreateServiceAccount` on the catalog name |
| `createApiKey` | `CreateApiKey` on the catalog name |
| `revokeApiKey` / `revokeAllApiKeys` | `RevokeApiKey` on the catalog name (`revokeApiKey` resolves the account from the key id) |

The per-prefix `CreateGrant` check is the anti-escalation guard: a manager can't hand an account reach they couldn't grant anyone. Removal has no per-prefix requirement since it only narrows access. Adding a grant that already exists overwrites its capability in place (new `grants::overwrite_user_grant`), so downgrades work in one call.

Mutations return the account in its post-mutation state so clients reconcile without a refetch, and removals and revocations are idempotent. The `serviceAccounts` query derives `lastUsedAt` from the account's keys (max `updated_at` across tokens with `uses > 0`) and batch-loads grants and keys for the whole page.

## Also in this PR

- New `Sensitive` GraphQL scalar for one-time secrets: it serializes as a plain string but is a distinct scalar in the schema so client tooling can recognize and redact it, and its `Debug` impl never prints the value. Applied to API-key, refresh-token, and Stripe setup-intent secrets.
- The bearer-credential authentication path (a refresh token presented directly as `Authorization: Bearer`) and `POST /api/v1/auth/token` now share one `generate_access_token`, so credential-error sanitization lives in one place. Both return a generic "invalid, expired, or unknown credential" on 401.
- Stripe billing-contact selection excludes service accounts, whose synthetic addresses must never become a tenant's billing contact.
- CI: the flaky Stripe integration test runs last in platform-test so it can't skip the steps behind it (it still gates the job).

## Testing

An sqlx test covers the lifecycle end to end: create with multi-grant seeding, duplicate catalog names, input validation (including each `validFor` rejection branch), mint, list with derived `lastUsedAt`, wrong-secret and revoked-key 401s, revoke and kill switches with idempotency, grant add/downgrade/remove semantics, self-management rejection, and non-manager denials throughout. A second test verifies a `team_admin` holder without full `Admin` can manage accounts while `CreateGrant` still bounds escalation. The JWT-signing exchange itself is covered by existing pgTAP tests, since the sqlx test DB lacks the vault secret and pgjwt.


crates/control-plane-api/src/server/public/graphql/service_accounts.rs
```diff
@@ -0,0 +1,1457 @@
...
+            sa_user_id,
+            catalog_name.as_str(),
+            claims.sub,
+        )
+        .fetch_one(&mut *txn)
+        .await
+        .map_err(duplicate_err)?;
+
+        for grant in &grants {
+            crate::grants::overwrite_user_grant(
```
GregorShear, line 306: I have mixed feelings about doing this in the create mutation - maybe we need an "update grant" mutation that lets us downgrade capabilities, and the create mutation just upserts (upgrade capabilities only)?

crates/models/src/authz.rs
```diff
@@ -108,7 +109,7 @@ impl CapabilityBundle {
                     | Self::ManageDataPlane.capabilities()
             }
             Self::Billing => EnumSet::empty(),
-            Self::TeamAdmin => CreateGrant | DeleteGrant | CreateInviteLink,
+            Self::TeamAdmin => CreateGrant | DeleteGrant | CreateInviteLink | ManageServiceAccount,
```
GregorShear: let's talk about this

.github/workflows/platform-test.yaml
```diff
@@ -131,13 +131,15 @@ jobs:
...
       - run: mise run ci:nextest-run
 
+      - run: mise run ci:doctest
+      - run: mise run ci:gotest
+      - run: mise run ci:catalog-test
+      - run: mise run build:flow-schema
+
+      # Run last so its flakiness can't skip the test steps above. It still gates
+      # the job on failure, surfacing genuine Stripe regressions.
       - name: Stripe integration test
```
GregorShear, line 141: moving this flaky test to the end of the line, at least for now

crates/control-plane-api/src/server/public/graphql/service_accounts.rs
```diff
@@ -0,0 +1,2066 @@
...
+}
+
+/// Verify that `user_id` is not a service-account identity, erroring if it is.
+///
+/// Service-account credentials are administered through createApiKey /
+/// revokeApiKey. The self-service refresh-token mutations reject a
+/// service-account caller: a valid key could otherwise mint replacement
+/// credentials for its own account — sidestepping the CreateApiKey gate and
+/// the admin-chosen expiry — or revoke keys outside the admin-facing flow.
+pub(super) async fn verify_not_service_account(
```
jshearer, line 744: `verify_not_service_account` only guards the GraphQL `createRefreshToken`/`revokeRefreshToken` mutations. It doesn't cover the PostgREST path, which reaches the same table and isn't protected.

I think this is what you call out:
> (A residual self-service surface exists via PostgREST's refresh_tokens grants; it disappears with PostgREST.)

IIRC the UI already moved over to GraphQL. If we get flowctl moved over then we can just disable the postgrest api for refresh tokens all together.

That being said, it wouldn't be that hard to close this hole today:

```sql
-- in public.create_refresh_token, before the insert:
if exists (select 1 from internal.service_accounts where user_id = auth_uid()) then
  raise 'service accounts cannot mint their own refresh tokens';
end if;
```

crates/control-plane-api/src/server/public/graphql/service_accounts.rs
```diff
@@ -0,0 +1,2066 @@
...
+                        (
+                            SELECT max(rt.updated_at)
+                            FROM public.refresh_tokens rt
+                            WHERE rt.user_id = sa.user_id
+                              AND rt.uses > 0
+                        ) AS "last_used_at: chrono::DateTime<chrono::Utc>"
+                    FROM internal.service_accounts sa
+                    LEFT JOIN auth.users creator ON creator.id = sa.created_by
+                    WHERE sa.catalog_name::text ^@ ANY($1)
+                      AND ($2::timestamptz IS NULL OR sa.created_at < $2)
```
jshearer, line 134: nit: cursor is `created_at` and the page predicate is `sa.created_at < $2`. Two accounts sharing a `created_at` straddling a page boundary results in the later being dropped during pagination. Low likelihood (needs identical timestamps) and it mirrors the pre-existing `refresh_tokens` pattern, but a `(created_at, user_id)` pagination key would close it.

crates/flow-client/control-plane-api.graphql
```diff
@@ -1706,6 +1826,88 @@ type RepublishRequested {
...
+	"""
+	cursor: String!
+}
+
+"""
+A user_grant held by a service account: the prefix it may act on and the
+capability it holds there. An account's access is the union of its grants,
+which may span multiple prefixes independent of its catalog_name anchor.
+"""
+type ServiceAccountGrant {
```
GregorShear: This should just be called UserGrant (service accounts and users both use the same user_grant mechanism for authz)

GregorShear:
My agent review finds an apparent inconsistency where we ban SAs from creating refresh (personal) tokens for themselves, but given `Admin` capability they're free to create API keys for themselves. 

API key creation has a specific capability bit which is indeed bundled with the `Admin` capability, but once unbundled customers will be able to choose whether or not to grant this power to SAs.

I'm inclined to accept this inconsistency during the migration/unbundling period, but I'm open to a temporary ban on SAs creating API keys even if they have sufficient authz that we would remove once the capability is unbundled. 

@jshearer

crates/control-plane-api/src/server/public/graphql/service_accounts.rs
```diff
@@ -0,0 +1,2063 @@
...
+
+#[cfg(test)]
+mod test {
+    use crate::test_server;
+
+    #[sqlx::test(
+        migrations = "../../supabase/migrations",
+        fixtures(path = "../../../fixtures", scripts("data_planes", "alice"))
+    )]
+    async fn test_service_account_lifecycle(pool: sqlx::PgPool) {
```
jshearer, line 968: This is the test that's timing out in CI (nextest kills it at 120s). I believe the root cause is in the test harness rather than the code under test: `test_server::snapshot` shifts the snapshot's `taken` only 2 seconds forward, and a denial is only terminal while `snapshot.taken_after(request.started)` holds. Once the test has been running for more than ~2s of wall clock, every expected-denial request below (bob's unauthorized attempts, the foreign-grant rejections) takes the provisional path in `authorization_outcome`, cancels the snapshot, and awaits a refresh that `GatedSnapshot` never serves (its source panics if refreshed a third time). Fast local machines finish inside the window; CI doesn't.

Suggested fix: bump the artificial `taken` shift in `test_server::snapshot` to something like an hour so denials stay terminal for the whole test, or teach `GatedSnapshot` to re-serve the snapshot with a bumped `taken`. Splitting this test up would also get each half under the window, but the shift fix addresses the class of problem.

crates/control-plane-api/src/server/public/graphql/service_accounts.rs
```diff
@@ -0,0 +1,2063 @@
...
+        let owner = match owner {
+            Some(owner) => owner,
+            None => {
+                return Err(async_graphql::Error::new(
+                    "service account API key not found",
+                ));
+            }
+        };
+
+        super::verify_authorization(
```
jshearer, line 665: Nit of the vanishingly-unlikely variety, since token ids are effectively unguessable: because the owner is resolved before authorizing (necessarily, since the account comes from the token id), a denial here surfaces `evaluate_names_authorization`'s message, which embeds the object name. So an authenticated caller holding a token id they aren't authorized for learns which catalog name owns it. Trivial fix though: map `permission_denied` to the same "service account API key not found" error so existence and denial are indistinguishable.

supabase/migrations/20260706120000_service_accounts.sql
```diff
@@ -0,0 +1,34 @@
+begin;
+
+-- Service accounts: non-login identities used for programmatic access
+-- (CI/CD, automation) and scoped, time-limited access grants.
+
+-- The `manage_service_account` bundle confers the fine-grained service-account
+-- capabilities (query, create, mint/revoke API keys) without full team-admin.
+alter type capability_bundle add value if not exists 'manage_service_account';
```
jshearer: nit: #3144 (deferred until after this PR lands) relabels the bundle enum to action-oriented, pluralized names: `manage_data_plane` -> `manage_data_planes`, `team_admin` -> `manage_users`, etc. Landing this as singular `manage_service_account` means 3144 grows yet another relabel. Since the value is brand new and nothing stores it yet, let's name it `manage_service_accounts` from the start (and `ManageServiceAccounts` on the Rust enum) so it's already in its final form.

crates/control-plane-api/src/server/public/graphql/service_accounts.rs
```diff
@@ -0,0 +1,2063 @@
...
+            &snapshot.user_grants,
+            env.claims()?.sub,
+            models::authz::Capability::QueryServiceAccounts,
+            None,
+        );
+
+        if user_accessible_prefixes.is_empty() {
+            return Ok(PaginatedServiceAccounts::new(false, false));
+        }
+        if user_accessible_prefixes.len() > MAX_PREFIXES {
```
jshearer, line 101: `alertConfigs` pairs this same `MAX_PREFIXES` cap with a prefix filter so a caller over the cap can narrow below it. Here there's a cap but no filter, so a caller whose grants reach more than 20 prefixes (support-shaped users especially) gets a hard error with no recourse. `authorized_prefixes` already takes a `prefix_filter`; we're just passing `None`. I think the UI will want this anyway: the dashboard scopes to the currently selected tenant, so `serviceAccounts` should take a prefix argument for that flow regardless.

jshearer (commented):
Got some feedback for ya! Nothing structural, feel free to push back/ignore on the nits

GregorShear:
Here's the diff on `public.create_refresh_token` that's getting overwritten in the migration:

```diff
CREATE OR REPLACE FUNCTION public.create_refresh_token(multi_use boolean, valid_for interval, detail text DEFAULT NULL::text) RETURNS json
    LANGUAGE plpgsql SECURITY DEFINER
    AS $$
declare
  secret text;
  refresh_token_row refresh_tokens;
begin
+ if exists (select 1 from internal.service_accounts where user_id = auth_uid()) then
+   raise 'service accounts cannot mint their own refresh tokens';
+ end if;

  secret = gen_random_uuid();

  insert into refresh_tokens (detail, user_id, multi_use, valid_for, hash)
  values (
    detail,
    auth_uid(),
    multi_use,
    valid_for,
    crypt(secret, gen_salt('bf'))
  ) returning * into refresh_token_row;

  return json_build_object(
    'id', refresh_token_row.id,
    'secret', secret
  );
commit;
end
$$;
```

crates/control-plane-api/src/server/public/graphql/service_accounts.rs
GregorShear: i'm leaving this alone for now - risk is truly negligible, and i think the correct fix is an opaque cursor, which i'm inclined to apply across all paginated queries sometime in the future.

crates/control-plane-api/src/server/public/graphql/service_accounts.rs
GregorShear: i'd like to fix this with a tenant filter (that I'd also like to introduce to all other queries) but pulling that thread now expands the scope in a way i'd rather avoid. 

i acknowledge support users will hit this in the short term, but i'd also like to do this tenant filter thing in the short term (it's on the critical path for the dashboard agent)

jshearer (approved):
LGTM! 🚀

PR #3059, opened 2026-06-18
proto-flow: tolerate unknown JSON fields in connector protocols

williamhbaker:
**Description:**

Generate the .flow, .capture, .derive, and .materialize serde impls with pbjson's `ignore_unknown_fields`, so a reader built against an older protocol revision skips fields it doesn't recognize instead of failing to deserialize.

`.flow` is included because connector Request/Response messages embed flow.CollectionSpec and friends, which deserialize via their own impls. The internal .ops, .runtime, and .shuffle protocols stay strict.

**Workflow steps:**

**Notes for reviewers:**

### DIFF
--- a/crates/proto-flow/build.rs
+++ b/crates/proto-flow/build.rs
@@ -22,6 +22,7 @@ fn main() {
         .register_descriptors(&std::fs::read(b.descriptor_path).expect("read descriptors"))
         .unwrap()
         .btree_map(["."]) // Make ordering stable for snapshots.
+        .ignore_unknown_fields()
         .build(&[
             ".flow",
             ".capture",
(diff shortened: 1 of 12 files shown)


dgreer-dev (approved):
LGTM.  My only comment is if there's any useful way to test this?

williamhbaker:
Thought about this a bit more over the weekend, and I decided we might as well go with the simpler route of just using `ignore_unknown_fields` on `proto-flow` in general, rather than trying to break that up into lenient / strict flavors. At best, the previous "strict" ones should always have their deployed code in lock-step with the protocol definitions since they deploy together, so the extra complexity wasn't really buying anything.

Maybe there's an argument to be made for `proto-gazette` to have this handling too, but I don't think that has ever caused us problems, so I'm not going to worry about it now.

williamhbaker:
On second-second thought, I've added the `ignore_unknown_fields` to `proto-gazette` now as well. I can't think of any real risk of doing this, and it makes the protocol story simpler, since the Gazette entities are sub-structs of `proton-flow` so now all of that will just be lenient.

PR #3065, opened 2026-06-20
docs: clarify BYOK KMS is decrypt-only and identity may not be in dashboard

jwhartley:
## What

Three clarifications to the secret-encryption section of `concepts/flowctl`:

1. **Decrypt-only.** The data plane identity needs decrypt access on your KMS key, not encrypt. On GCP KMS that is `roles/cloudkms.cryptoKeyDecrypter`. Estuary only decrypts at connector run time and never encrypts with your key.
2. **Auto-encryption uses Estuary's key.** `flowctl` 0.5.18+ and the dashboard auto-encrypt plain-text configs with Estuary's managed key, not a bring-your-own key. To use your own key you must encrypt manually before publishing, otherwise your key is never exercised.
3. **Cross-cloud identity not in dashboard.** When the KMS provider differs from the data plane's cloud (for example a GCP KMS key with an AWS data plane), the GCP service account to grant is not surfaced under Admin > Settings > Data Planes. Added a note to contact support.

## Why

A customer asked whether the data plane service account needs Encrypt/Decrypt on their GCP KMS key. The page said to grant "access to decrypt" but never stated decrypt-only, did not warn that auto-encryption bypasses bring-your-own keys, and pointed at a dashboard panel that does not display the cross-cloud identity.

Verified against the source:
- The config-encryption service sends `config` + `schema` with no keychain and falls back to its single default key, so plain-text publishes use Estuary's key (`crates/config-encryption`, `crates/flowctl/src/draft/encrypt.rs`).
- Data plane provisioning grants `roles/cloudkms.cryptoKeyDecrypter` to the data plane service account (est-dry-dock).

### DIFF
--- a/site/docs/concepts/flowctl.md
+++ b/site/docs/concepts/flowctl.md
@@ -249,18 +249,22 @@ been encrypted.
 Estuary supports customers encrypting secrets with keys from AWS Key Management Service, Google Cloud Platform KMS, and Azure Key Vault.
 
 :::important
-When deploying catalogs onto the managed Estuary runtime, you must grant **decrypt**
-access on your KMS key to the appropriate identity for your data plane. On GCP KMS,
-this is the `roles/cloudkms.cryptoKeyDecrypter` role. Estuary only decrypts your
-configuration at connector run time and never encrypts with your key, so the
-identity does not need encrypt access.
+When deploying catalogs onto the managed Estuary runtime, you must grant **decrypt-only**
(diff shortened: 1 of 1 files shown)


aeluce (approved):
Rewrote things a bit, including some slight reordering to try to present a straightforward flow. Let me know if you think this version is an improvement.

Also directed users to support until https://github.com/estuary/ui/issues/2011 is addressed.

PR #3074, opened 2026-06-23
notifications: trial bucket retention is 20 days, not 30

jwhartley:
## What

The missing-payment-method notification email told customers their data is stored in an Estuary-owned trial bucket "with a 30 day retention policy." The trial bucket (`estuary-trial`) retention is **20 days**, not 30.

```
crates/notifications/src/missing_payment_method.rs:28
- ...cloud storage bucket with a 30 day retention policy.
+ ...cloud storage bucket with a 20 day retention policy.
```

## Why

20 days is the correct value and matches what the docs already say:
- [concepts/storage-mappings](https://docs.estuary.dev/concepts/storage-mappings/): "Data in Estuary's cloud storage bucket is deleted 20 days after collection."
- [getting-started/quickstart](https://docs.estuary.dev/getting-started/quickstart/) and the installation page say the same.

The retention was reduced from 30 to 20 a while back; the docs were updated but this email string was missed.

## Scope

Only the storage-retention sentence changed. The 30-day references in `free_trial.rs` and the UI register/billing copy are the **account free-trial period**, which is genuinely 30 days, and are intentionally left untouched.

### DIFF
--- a/crates/notifications/src/missing_payment_method.rs
+++ b/crates/notifications/src/missing_payment_method.rs
@@ -25,7 +25,7 @@ pub fn register_templates<'a>(registry: &mut handlebars::Handlebars<'a>) -> anyh
 <div style="margin-top: 15px;">
     <p style="font-weight: bold; font-size: 19px; margin-bottom: 5px;">Where is my data stored?</p>
     <div style="border-left: 4px solid lightgrey; padding-left: 12px; margin-bottom: 15px;">
-        <p class="body-text" style="margin-bottom: 0;">By default, all collection data is stored in an Estuary-owned cloud storage bucket with a 30 day retention policy. Now that you have a paid account, you can update this to store data in your own cloud storage bucket. We support GCS, S3, and Azure Blob storage.</p>
+        <p class="body-text" style="margin-bottom: 0;">By default, all collection data is stored in an Estuary-owned cloud storage bucket with a 20 day retention policy. Now that you have a paid account, you can update this to store data in your own cloud storage bucket. We support GCS, S3, and Azure Blob storage.</p>
     </div>
 </div>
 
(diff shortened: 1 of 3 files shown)


crates/notifications/src/missing_payment_method.rs
```diff
@@ -25,7 +25,7 @@ pub fn register_templates<'a>(registry: &mut handlebars::Handlebars<'a>) -> anyh
 <div style="margin-top: 15px;">
     <p style="font-weight: bold; font-size: 19px; margin-bottom: 5px;">Where is my data stored?</p>
     <div style="border-left: 4px solid lightgrey; padding-left: 12px; margin-bottom: 15px;">
-        <p class="body-text" style="margin-bottom: 0;">By default, all collection data is stored in an Estuary-owned cloud storage bucket with a 30 day retention policy. Now that you have a paid account, you can update this to store data in your own cloud storage bucket. We support GCS, S3, and Azure Blob storage.</p>
+        <p class="body-text" style="margin-bottom: 0;">By default, all collection data is stored in an Estuary-owned cloud storage bucket with a 20 day retention policy. Now that you have a paid account, you can update this to store data in your own cloud storage bucket. We support GCS, S3, and Azure Blob storage.</p>
```
aeluce:
```suggestion
        <p class="body-text" style="margin-bottom: 0;">By default, all collection data is stored in an Estuary-owned cloud storage bucket with a 20 day retention policy. If you haven't already, now is a great time to update your <a href="https://docs.estuary.dev/getting-started/installation/">storage mappings</a> to use your own cloud storage bucket. We support GCS, S3, and Azure Blob storage.</p>
```
The assertion in the next sentence that storage mappings are only available for paid accounts is also false. We could update the language to something like this.

aeluce (commented):
I'm not super familiar yet with updating crates, so take this with a grain of salt (and perhaps a once-over from control plane). Whatever language we go with may need to be propagated to the `missing_payment_method` notification snapshots as well. As is, the discrepancy in expected/actual language seems to cause test failures.

…And now that I'm looking at this, there are other updates I want to make to the language, but that's increasingly tangential to this specific issue. I should give the alert emails a once-over myself sometime…

jwhartley:
Good catch on the snapshots. Updated both `missing_payment_method` snapshots (paid + free_trial) to the new copy in a8e63cd1dd, full notifications suite passes (25/25).

Separately, I noticed a stale `alert_notifications__tests__*` snapshot set left over from when the crate was renamed to `notifications`. No current test reads them, so they aren't part of the failures here. I'll clean those up in a separate PR rather than widen this one.

aeluce (approved):

PR #3075, opened 2026-06-23
flowctl: normalize --prefix trailing slash to avoid misleading PermissionDenied

### LINKED ISSUE #2834: flowctl catalog: missing trailing slash on --prefix gives misleading PermissionDenied error

GregorShear:
## Problem

`flowctl catalog pull-specs --prefix AcmeCo` (no trailing slash) returns a misleading error even for fully-authorized users:

```
PermissionDenied: johnsmith@acme.co is not authorized to access prefix or name 'AcmeCo' with required capability read
```

The value is passed verbatim to `models::Prefix::new`. Estuary prefixes must end in `/`, so the control plane treats `AcmeCo` as an unrecognized catalog *name* rather than a prefix and rejects it.

## Fix

Auto-normalize the prefix by appending a trailing `/` when it's missing. The fix lives in `to_vars` — the single point through which every `NameSelector`-based command routes its `--prefix` values:

- `catalog list`
- `catalog pull-specs`
- `catalog delete`

An **empty** prefix is deliberately left untouched: `""` is a valid selector meaning "everything the user can access", whereas `"/"` is invalid, so blindly appending would introduce a new bug.

## Testing

- New unit test `test_normalize_prefix` covers the missing-slash case, multi-segment prefixes, already-normalized input, and the empty edge case.
- `cargo test -p flowctl normalize_prefix` passes.

## Out of scope

`flowctl alerts subscriptions` and `flowctl raw` accept their own `--prefix` args that don't go through `NameSelector`, so they retain the same foot-gun. The issue scoped the fix to `NameSelector`; happy to fold those in as a follow-up if desired.

Closes #2834

### DIFF
--- a/crates/flowctl/src/catalog/list/mod.rs
+++ b/crates/flowctl/src/catalog/list/mod.rs
@@ -270,10 +270,19 @@ fn to_vars(list: &List) -> Vec<list_live_specs_query::LiveSpecsBy> {
     let catalog_type = list.type_selector.get_single_type_selection();
 
     let mut vars = Vec::new();
+    let mut seen_prefixes = std::collections::HashSet::new();
     for prefix in list.name_selector.prefix.iter() {
+        let prefix = normalize_prefix(prefix);
+        // Skip prefixes that normalize to an already-queued value. Because we
+        // append a trailing '/', a bare `acmeCo` and an explicit `acmeCo/`
+        // collapse to the same prefix; issuing both queries would list every
+        // matching spec twice.
+        if !seen_prefixes.insert(prefix.as_str().to_string()) {
+            continue;
+        }
         vars.push(list_live_specs_query::LiveSpecsBy {
             names: None,
-            prefix: Some(normalize_prefix(prefix)),
+            prefix: Some(prefix),
             catalog_type,
             data_plane_name: data_plane_name.clone(),
         });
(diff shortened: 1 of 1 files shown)


crates/flowctl/src/catalog/list/mod.rs
```diff
@@ -273,7 +273,7 @@ fn to_vars(list: &List) -> Vec<list_live_specs_query::LiveSpecsBy> {
     for prefix in list.name_selector.prefix.iter() {
         vars.push(list_live_specs_query::LiveSpecsBy {
             names: None,
-            prefix: Some(models::Prefix::new(prefix)),
+            prefix: Some(normalize_prefix(prefix)),
```
jshearer: Nit: Overlapping/duplicate prefixes that previously failed fast now normalize to the same value and produce duplicate rows. This is pre-existing, just slightly more confusing now if you do something unusual.

---

Ex:`flowctl catalog list --prefix acmeCo --prefix acmeCo/` (or the same bare prefix twice) 

Before this change the bare `acmeCo` format errored (`PermissionDenied`), so the command failed fast. Now, both entries become `acmeCo/`, `to_vars` pushes two identical `LiveSpecsBy` queries, and `fetch_paginated_live_specs` concatenates both result sets with no deduplication — every spec under `acmeCo/` is listed twice.

jshearer (dismissed):
LGTM % one nit that was mostly pre-existing anyway

crates/flowctl/src/catalog/list/mod.rs
GregorShear: fixed - deduping prefixes now

jshearer (approved):
LGTM

telcharr:
For the record, #3035 which was opened June 15, linked to this same issue implemented this same normalization, including the empty-prefix handling and equivalent tests, plus the alerts subscriptions and raw coverage you've listed as "out-of-scope follow-up". Happy to rebase that remainder as the follow-up, but I would've appreciated a review or at least a heads-up on the original.

PR #3076, opened 2026-06-23
supabase: qualify replace_data_plane_releases DELETE for PostgREST

skord:
## Summary
The ops "Release Data Planes" workflow was cut over to call
`replace_data_plane_releases` over PostgREST instead of a direct psql
connection. The first master run failed:

    {"code":"21000","message":"DELETE requires a WHERE clause"}

The function clears the table with an unqualified `delete from
data_plane_releases`. The PostgREST execution context runs under
`sql_safe_updates`, which rejects an unqualified DELETE; the direct psql path
never enforced this, so it only surfaced after the cutover.

## Change
New migration that `create or replace`s the function, changing the delete to
`delete from data_plane_releases where true`. The always-true predicate
satisfies the safe-update guard while still removing every row, and the
full-table replace stays atomic (delete + insert in one call). `create or
replace` on the same signature preserves the existing execute grant to
`data_plane_releases_ci`; nothing else changes.

## Rollout
Takes effect once the migration is applied to the production database; the
"Release Data Planes" workflow passes on its next run after that.

### DIFF
--- a/supabase/migrations/20260623120000_replace_data_plane_releases_safe_delete.sql
+++ b/supabase/migrations/20260623120000_replace_data_plane_releases_safe_delete.sql
@@ -1,0 +1,31 @@
+begin;
+
+-- replace_data_plane_releases cleared the table with an unqualified DELETE. Postgres
+-- rejects that under sql_safe_updates, which the PostgREST execution context enables,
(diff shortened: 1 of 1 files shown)


qaoj (approved):
Odd but I get why they do that

PR #3079, opened 2026-06-25
notifications: remove stale alert_notifications snapshot files

jwhartley:
## What

Removes the 16 `crates/notifications/src/snapshots/alert_notifications__tests__*.snap` files.

## Why

The crate was renamed from `alert_notifications` to `notifications` (see `crates/notifications/Cargo.toml`). insta names snapshot files `{crate}__{module}__{name}.snap`, so it now writes and reads under the `notifications__tests__` prefix. Every test calls `insta::assert_snapshot!("<explicit_name>", ...)`, and the live set lives under `notifications__tests__`. The `alert_notifications__tests__*` files are dead duplicates left over from before the rename; no current test can reference them.

## Verification

```
cargo test --manifest-path crates/notifications/Cargo.toml
```

- Before removal: 25 passed.
- After `git rm` of the 16 files: 25 passed (a missing referenced snapshot would have failed the run).

The 19 live `notifications__tests__*.snap` files are untouched.


jwhartley:
should be a quick one @jshearer, just some cleanup

jshearer (approved):
LGTM, love net negative PRs 😍

PR #3080, opened 2026-06-26
Docs: Remove an IP address from GCP allowed list

jwhartley:
Removed an IP address entry from the GCP section.

**Description:**

**Workflow steps:**

**Notes for reviewers:**

### DIFF
--- a/site/docs/security/allow-ip-addresses.md
+++ b/site/docs/security/allow-ip-addresses.md
@@ -45,7 +45,6 @@ While your dashboard is the best location to find accurate, up-to-date IP addres
 
 **GCP `us-central1 c2`:**
 
-- `34.121.207.128/32`
 - `35.226.75.135/32`
 
 **AWS `us-west-2 c1`:**


skord:
My vote here is to remove this entirely. It puts a burden on us to provide updates to clients we may not be even sure are using it and is difficult for customers to resolve what is "correct"

aeluce:
The IP address this PR removes from docs still shows up in the dashboard for the `us-central1 c2` data plane. Is that being adjusted?

I'd also be amenable to removing the explicit IP addresses from docs, as the dashboard should be the source of truth. Unless, @jwhartley, you see users find a quick reference like this useful?

jwhartley:
There is value in being able to reference these publicly (e.g. customer's security team gets the link, vs needing to log in to verify despite not having an estuary account), so I'd prefer to leave them up, although it's not a strong opinion.

We should adjust the dashboard as well. 

Could I a get an approve to move this forward?

skord (approved):
The removal checks.

jwhartley:
Note: removed all the hardcoded IPs in https://github.com/estuary/flow/pull/3097

PR #3081, opened 2026-06-26
runtime-next: durably seed initial connector state as {} to match V1

williamhbaker:
**Description:**

Connector state is only updated via RocksDB Merge, and the merge operator treats the first operand on an absent key as the base document -- where a JSON null is a literal value, not a deletion. So a task first published on Runtime-V2 stores a connector's first checkpoint of e.g. `{"stateKey": null}` verbatim instead of reducing it to {}, corrupting its state from the first patch onward.

Mirror V1's `load_connector_state`: when the recovery scan finds no connector state, durably seed `{}` as the base and reflect it in the recovered `Recover`, so the first checkpoint reduces as a genuine merge patch. Seed only on the state-bearing shard -- the capture shard, or shard zero of a derivation or materialization, whose state the leader broadcasts to peers; other shards must recover an empty Recover.

This replaces a `non_empty_state` helper that substituted `{}` only on the connector-facing `Open`/`Apply` requests: it changed what the connector saw, not what the runtime stored, so the mis-store still happened.

**Workflow steps:**

**Notes for reviewers:**

### DIFF
--- a/crates/runtime-next/src/shard/capture/handler.rs
+++ b/crates/runtime-next/src/shard/capture/handler.rs
@@ -295,10 +295,11 @@ where
         .map(|(i, b)| (b.state_key.clone(), i as u32))
         .collect();
     sorted_state_keys.sort();
-    let (mut db, recover) = db
+    let (mut db, mut recover) = db
         .scan(sorted_state_keys)
         .await
         .context("scanning RocksDB")?;
+    db = db.seed_connector_state(&mut recover).await?;
     let proto::Recover {
         ack_intents,
         connector_state_json,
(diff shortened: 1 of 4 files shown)


jgraettinger (approved):
LGTM

williamhbaker:
Updated based on further discussions: We need to seed the initial persisted state, so that the connector's _first_ state update has something to merge into. In doing this, we're able to get of the `non_empty_state` helper entirely.

I did some more thorough manual testing on this change as well, with a one-off custom capture and materialization connector and a local stack. Confirmed that pre-change (and also, my initial attempt at this PR) the "bug" exists, and `{"stateKey": null}` kinds of initial state updates round-trip, but after the change a `{}` comes back to the connector.

dgreer-dev (approved):
LGTM

PR #3085, opened 2026-06-29
docs: clarify column-level SELECT grants unsupported for MySQL/MariaDB CDC

jwhartley:
## What

Adds a `Column-level permissions` section to the MySQL and MariaDB CDC capture connector docs, and a one-line pointer to it from the Setup steps.

## Why

A customer tried to limit what leaves their database by granting column-scoped `SELECT` (`GRANT SELECT (col1, col2) ...`) to the capture user. This does not work and actively breaks the capture, but the docs only said `SELECT` could be restricted "to just the tables," which implies column scoping is fine.

Confirmed in `estuary/connectors` `source-mysql`:
- Backfill builds `SELECT *` (`backfill.go` `keylessScanQuery` / `buildScanQuery`), and the per-table prerequisite check runs `SELECT * ... LIMIT 0` (`prerequisites.go`). A column-scoped grant fails both with `cannot read from table`.
- CDC reads the binary log, authorized by the global `REPLICATION CLIENT` / `REPLICATION SLAVE` privileges, and decodes every column in the row image (`replication.go`). Column grants have no effect on the replication stream.

The new section explains both failure modes and points users to redaction to keep specific columns out of the pipeline.

## Scope

Covers the two canonical pages (`MySQL.md`, `MariaDB.md`). The RDS/Aurora/Cloud SQL variant pages were left for a follow-up; happy to extend if preferred.

### DIFF
--- a/site/docs/reference/Connectors/capture-connectors/MariaDB/MariaDB.md
+++ b/site/docs/reference/Connectors/capture-connectors/MariaDB/MariaDB.md
@@ -42,6 +42,8 @@ To meet these requirements, do the following:
 
 The `SELECT` permission can be restricted to just the tables that need to be
 captured, but automatic discovery requires `information_schema` access as well.
+It cannot be restricted to a subset of columns within a table; see
+[Column-level permissions](#column-level-permissions) for details.
 
 ```sql
 CREATE USER IF NOT EXISTS flow_capture IDENTIFIED BY 'secret';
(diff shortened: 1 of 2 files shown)


aeluce (approved):
This may go into more depth than is needed. Having a special section makes it seem like table-level `SELECT` is an outlier that needs to be noted for these particular connectors. It may be sufficient to simply reword the existing info about `SELECT` permissions.

Ex.
```
1. Create the `flow_capture` user with permissions for replication and 
reading required tables.

Grant `SELECT` on all tables or restrict it to tables to be captured. 
Note that `SELECT` permissions must be at the table level rather 
than the column level. Automatic discovery will also require 
`information_schema` access.
```

Approving as-is, though, in case you've found this to be a recurring point of confusion for users.

jwhartley:
Good call, incorporated. Dropped the standalone section and reworded the permissions line in each setup step: `SELECT` at the table level not column level, with a one-line pointer to redaction for keeping specific columns out. It has come up with a customer, so I kept that redaction breadcrumb, but agreed the dedicated section over-weighted it.

PR #3086, opened 2026-06-29
docs: avoiding backfills during a database failover or host change

jwhartley:
## What

Documents how to avoid a full backfill when a CDC source moves to a new writer, including failovers, standby promotions, instance migrations, and Amazon Aurora Global Database regional switchovers.

The existing "Preventing backfills during database upgrades" section already covers the pause-writes / "Only Changes" procedure, but it was framed entirely around in-place upgrades (same host). It never mentioned failover or a host change, and had no MySQL specifics. A user searching for failover behavior landed nowhere, even though the procedure is the right answer.

## Approach

Rather than add a parallel section, this generalizes the existing one. The new-host case is a step variant, not a separate procedure.

- `guides/backfilling-data.md`: renamed the heading to "Preventing backfills during database upgrades and failovers"; generalized the intro to cover failovers/promotions/migrations (Aurora Global Database as one example); split step 4 into 4a (same host, no extra change) and 4b (new host: also repoint the capture `address`, because the new writer does not carry the old CDC position). Updated the internal anchor reference.
- `reference/Connectors/.../PostgreSQL.md` and `.../MySQL.md`: short pointers into that section. MySQL had no failover guidance at all.
- `private-byoc/privatelink.md`: note in the cross-region section about pre-registering another region's endpoint ahead of a host change, cross-linked back.

## Why these specifics are correct

- Postgres: logical replication slots are not replicated to a promoted standby or Aurora Global Database secondary cluster; a new slot can only start at the current WAL position. The publication and watermarks table do carry over with the storage volume.
- MySQL: binlog coordinates are per-server, so the stored file+offset is invalid on the new writer (`ERROR 1236`); capture must use the writer endpoint (readers report `log_bin = OFF`).
- Neither connector validates server identity on reconnect, so editing `address` alone resumes the stale cursor and fails. Verified against the connector source.

## Notes

No new page; this extends and cross-links existing content. Prompted by a customer asking for a repeatable regional-failover pattern for an Aurora Global Database.

### DIFF
--- a/site/docs/guides/backfilling-data.md
+++ b/site/docs/guides/backfilling-data.md
@@ -283,7 +283,7 @@ In this case, many connectors allow you to turn off backfilling on a per-stream 
 
 ### Preventing backfills during database upgrades and failovers
 
-During an upgrade, some databases invalidate a replication slot, binlog position, CDC tables, or similar. As Estuary relies on these methods to keep its place, upgrades will disrupt the Estuary pipeline in these cases. The same is true of a failover or promotion that moves the capture onto a new writer - for example a standby promotion, a migration to a new instance, or switching an [Amazon Aurora Global Database](https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/aurora-global-database.html) over to a secondary region - because the new writer does not carry the old writer's CDC position.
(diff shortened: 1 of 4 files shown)


aeluce (approved):
Thanks! I see the PR description notes that the Postgres publication and watermarks table will carry over with the storage volume on failover. I didn't see that specified in the docs; would it be useful to include, or is it extraneous?

jwhartley:
Yeah I trimmed those out as they were adding unnecessary bulk

PR #3090, opened 2026-06-29
runtime-next: don't abandon the final capture transaction on connector EOF

jgraettinger:
The capture HeadFSM stepped straight to `Head::Stop` when it observed connector EOF with no open transaction, without checking whether the Tail had finished committing. Because a sweep's final transaction is rotated into the Tail in the iteration just before EOF is observed, the actor loop (`while !Head::Stop`) could exit with that transaction still mid-commit — abandoning its documents (no WriteIntents) and its connector checkpoint (no committing Persist). Each restart then re-swept and re-dropped the same final batch.

This is V2-specific: exit-after-sweep makes connector EOF the normal session-end path. It was intermittent because `restart = session_start + interval` masked it — a sweep finishing faster than the poll interval took the Sleep arm and the Tail committed during the wait. Only sweeps running longer than the interval hit the immediate-Stop arm and lost data.

Guard the EOF Stop on `tail_done`, mirroring the sibling `stopping` branch. When the Tail is still committing, `ready` is Eof so the extend guard can't match, and we fall through to hold Idle until the Tail reaches Done.

### DIFF
--- a/crates/runtime-next/src/leader/capture/fsm.rs
+++ b/crates/runtime-next/src/leader/capture/fsm.rs
@@ -281,8 +281,9 @@ impl HeadIdle {
                 return (Action::Idle, Head::Idle(self));
             }
         }
-        // Restart condition: hold until `task.restart`, and then Stop this session.
-        if matches!(ready, ConnectorRx::Eof) && !is_open {
+        // Restart condition: once the Tail has drained, hold until `task.restart`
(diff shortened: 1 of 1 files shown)


williamhbaker (approved):
LGTM

dgreer-dev (approved):
LGTM as well

PR #3091, opened 2026-06-29
runtime/container: use ops::decode::Decoder for log lines

williamhbaker:
**Description:**

`ops::decode::Decoder` transmutes log lines into ones that are compatible with our ops logs schema, inferring as much as it can from the source log when it isn't well-formed and ensuring it has all the required fields we need.

This is already used in `connector-init`, and that path covers everything coming from the connector process itself. But it misses logs emitted by the container runtime, and although these are rare, they are not in conformance with the ops logs schema and are not valid logs collection documents.

There is no separate schema validation check of documents written to logs journals, so it's really important that what we write there is valid per the schema.

This does duplicate the `decoder` work done in `connector-init`, and that may be an area for future optimization.

Manually tested by writing some non-conforming logs to `/proc/1/fd/2` in a running connector. Pre-fix, this resulted in a document being written to the logs journal missing a `ts` field, among other things. Post-fix, the logs are processed through `ops::decode::Decoder` as written in a a conforming way, as expected. This simulates a stderr write from the container runtime, which is otherwise difficult to directly induce.

**Workflow steps:**

**Notes for reviewers:**

### DIFF
--- a/crates/ops/src/decode.rs
+++ b/crates/ops/src/decode.rs
@@ -147,7 +147,7 @@ where
     }
 
     fn rawline_to_log(&self, line: &str, lookahead: &[u8]) -> (Log, usize) {
-        let mut message = line[..line.len() - 1].to_string(); // Strip single trailing \n.
+        let mut message = line.strip_suffix('\n').unwrap_or(line).to_string();
         let mut consumed: usize = 0;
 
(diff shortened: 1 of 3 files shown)


jgraettinger (approved):
LGTM

Issue #3092, opened 2026-06-30
dekaf: advertised high-water-mark regresses below served reads during journal primary hand-off (aborts failOnDataLoss consumers)


jwhartley:
## 1. Priority

Low - a production Spark consumer was fully aborted, but no data was lost (records all present, journal self-recovered within minutes) and clean workarounds exist (`failOnDataLoss=false` + dedup, or auto-retry on the exception). The trigger (broker primary hand-off during allocator convergence) is rare.

## 2. Scope and prevalence

Theoretical blast radius: any Kafka client reading Dekaf with `failOnDataLoss=true` (the librdkafka/Spark default). Worst on low-traffic journals with a long `flushInterval`, where the unpersisted spool window is large, so the persisted-head fallback lags live reads by a lot.

Observed prevalence: first observed occurrence. One journal, one consumer, single event. Triggered only when a consumer has read ahead into the live (unpersisted) spool AND the journal's primary hands off during that window.

## 3. Trigger (testable)

A Dekaf consumer reads recent data that is still in the broker's in-memory spool (not yet flushed to the fragment store), committing offset `H_live`. The journal's primary then hands off / the allocator reconverges. A `ListOffsets`/high-water-mark query during the churn returns the last persisted fragment boundary `H_persisted < H_live`. A `failOnDataLoss=true` consumer sees the latest offset move backward and aborts.

Observed values: committed `169994417`, advertised-latest regressed to `169328047` (a delta of 666,370 bytes). Both are exact fragment boundaries of the journal.

## 4. Root cause + confidence

Confirmed (broker logs + fragment metadata + code). During a broker primary hand-off, the high-water-mark Dekaf advertises falls back to the last *persisted* fragment boundary, which lags the live spool the consumer had already read through. It is transient and self-heals once the new primary re-advertises the true head and the spool flushes.

## 5. Investigation steps / reproduction (testable)

Consumer error (scrubbed):
```
StreamingQueryException: [STREAM_FAILED]
Partition <topic>-0's offset was changed from 169994417 to 169328047, some data may have been missed.
```
Failure at 2026-06-29 12:50:54 UTC. Consumer: `startingOffsets=earliest`, `failOnDataLoss=true`, no `maxOffsetsPerTrigger`.

Fragment listing (`flowctl collections list-fragments`) on the journal `.../<collection>/<gen>/pivot=00`:
```
fragment [168909134, 169328047]  persisted 2026-06-29 00:00:59 UTC
[...]

PR #3094, opened 2026-06-30
docs: Dekaf consumer behaviors + Spark Structured Streaming guidance

jwhartley:
## What

Extends `using-dekaf.md` with consumer guidance that has come up repeatedly in support, in two layers:

1. **Consumer behaviors to know** (any client): offsets are journal byte positions; the advertised latest offset can transiently move backward during a broker hand-off and is not data loss (with how to confirm via `flowctl collections read`); Avro `logicalType` decoding (e.g. uuid -> UUID object); parallelism via journal splits.
2. **Reading from Apache Spark Structured Streaming**: avoid `maxOffsetsPerTrigger` (byte-budget cap drops partial records), handle `failOnDataLoss` (it aborts on the transient backward-offset case), and set `spark.sql.avro.datetimeRebaseModeInRead` explicitly (PERMISSIVE silently nulls pre-Gregorian dates, SPARK-31404). Plus an example reader config.

## Why

These are recurring, non-obvious Dekaf consumer issues. The byte-offset model, the transient latest-offset regression, and the Avro decoding traps each surfaced as "missing data" reports that turned out to be consumer-side or transient. The transient-latest behavior is tracked in estuary/flow#3092.

Part 1 lives with the general consumer guidance so non-Spark consumers (Flink, librdkafka, kcat) benefit too; Part 2 is the Spark-specific config that builds on it.

## Notes

- Single file changed, no code. Generic content, no customer specifics.

### DIFF
--- a/site/docs/features/using-dekaf.md
+++ b/site/docs/features/using-dekaf.md
@@ -226,17 +226,6 @@ Spark's `kafka` source reads a Dekaf topic directly. Read [Consumer
 behaviors](#consumer-behaviors) first; the items below are Spark-specific configuration on
 top of those behaviors.
 
-### Do not set `maxOffsetsPerTrigger`
-
-Because offsets are journal byte positions, Spark treats `maxOffsetsPerTrigger` as a
-per-micro-batch byte budget, not a record count. A capped batch can end in the middle of a
-document; the next batch resumes at that mid-document offset, which is not a clean record
-boundary, so Dekaf skips the partial record. The result is one silently dropped record per
-capped batch boundary.
-
-For backfill pressure, give the executors more cores and memory and/or split the collection
-into more journals for parallel reads, rather than capping the batch.
-
 ### Handle `failOnDataLoss`
 
 Spark's default `failOnDataLoss=true` aborts the query whenever a partition's latest offset
--- a/site/docs/features/using-dekaf.md
+++ b/site/docs/features/using-dekaf.md
@@ -222,9 +222,9 @@ journals (non-breaking). Contact Estuary support before splitting a production c
 
 ## Reading from Apache Spark Structured Streaming
 
-Spark's `kafka` source reads a Dekaf topic directly. Read [Consumer behaviors to
-know](#consumer-behaviors-to-know) first; the items below are Spark-specific configuration
-on top of those behaviors.
+Spark's `kafka` source reads a Dekaf topic directly. Read [Consumer
+behaviors](#consumer-behaviors) first; the items below are Spark-specific configuration on
+top of those behaviors.
 
 ### Do not set `maxOffsetsPerTrigger`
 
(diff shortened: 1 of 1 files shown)


site/docs/features/using-dekaf.md
```diff
@@ -169,3 +169,107 @@ If a record shows up here but not in your application, the problem is client-sid
 your consumer-group offsets and your deserializer. Permissive Avro/JSON decoders can
 silently drop records that fail to decode — switch to a strict or fail-fast mode to surface
 the error rather than discarding the record.
+
+## Consumer behaviors to know
```
aeluce:
```suggestion
## Consumer behaviors
```

Minor nit that anything in docs is something we think someone should know so we could leave that part off. Would affect anchor links further down the doc.

site/docs/features/using-dekaf.md
```diff
@@ -169,3 +169,107 @@ If a record shows up here but not in your application, the problem is client-sid
...
+silently never match and look like missing records — normalize the type (for example,
+`str(value)`) before comparing. This is a client-side decoding concern, not an Estuary one.
+
+### Read partitions in parallel by splitting the collection
+
+Because each journal is one Kafka partition, you parallelize reads by splitting the
+collection into more journals. A split only distributes data written after it; to spread an
+existing backlog across the new journals you also need to re-backfill from the source. A
+split is a collection-level change, so every materialization on the collection sees the new
+journals (non-breaking). Contact Estuary support before splitting a production collection.
```
aeluce, line 221: Will journal auto-splitting for v2 affect this advice?

site/docs/features/using-dekaf.md
```diff
@@ -169,3 +169,107 @@ If a record shows up here but not in your application, the problem is client-sid
...
+  offset recovers, so the transient case does not skip data.
+- Keep `failOnDataLoss=true` and wrap the streaming query in an auto-restart that retries on
+  this specific exception.
+
+If you hit the error, the records are almost always still present: verify with
+`flowctl collections read`, then restart (Spark resumes from its checkpoint).
+
+### Set the Avro datetime rebase mode explicitly
+
+Spark's `PERMISSIVE` Avro mode silently nulls values it cannot parse, including dates before
```
aeluce: This section doesn't scan super easily and may be difficult to parse. What do you think of reordering it like this?

```
Spark's `PERMISSIVE` Avro mode silently nulls values it cannot parse, so affected records
look empty or missing. Use `FAILFAST` while debugging to surface the real error. You can
then choose how you'd like to handle these values.

For example, dates before the Gregorian cutover, like `1582-10-15`, cannot be parsed with
permissive null-ing. With `FAILFAST`, the underlying
`INCONSISTENT_BEHAVIOR_CROSS_VERSION.READ_ANCIENT_DATETIME` error is exposed,
and you can set `spark.sql.avro.datetimeRebaseModeInRead` to handle old datetime values:
* Use `CORRECTED` to read values as-is
* Use `LEGACY` to rebase across the calendar difference
```

aeluce (dismissed):
Super helpful to collect these from real support cases, thanks!

Had some minor suggestions, though fine from my end to continue without them.

site/docs/features/using-dekaf.md
jwhartley: yeah... it'll create automatic splits which is a bit worrying, I'll raise it in slack.

jwhartley:
Agree with your edits - please re-review @aeluce, and I asked re auto journal splitting here: https://estuaryworkspace.slack.com/archives/C03Q2NRFKDL/p1784009210748589

aeluce (dismissed):

site/docs/features/using-dekaf.md
```diff
@@ -169,3 +169,111 @@ If a record shows up here but not in your application, the problem is client-sid
...
+Spark's `kafka` source reads a Dekaf topic directly. Read [Consumer
+behaviors](#consumer-behaviors) first; the items below are Spark-specific configuration on
+top of those behaviors.
+
+### Do not set `maxOffsetsPerTrigger`
+
+Because offsets are journal byte positions, Spark treats `maxOffsetsPerTrigger` as a
+per-micro-batch byte budget, not a record count. A capped batch can end in the middle of a
+document; the next batch resumes at that mid-document offset, which is not a clean record
+boundary, so Dekaf skips the partial record. The result is one silently dropped record per
```
aeluce: Sorry, one more note. Dekaf now handles mid-document offsets by using a readback method (#3150), so we may be able to remove this section.

jwhartley:
Confirmed and cut in b521fedca19 — PR #3150's readback fix covers the maxOffsetsPerTrigger case exactly (fetch lands mid-document -> serves the containing document instead of skipping it), so the warning no longer holds.

aeluce (approved):

PR #3096, opened 2026-07-01
shuffle: make the shuffle disk limit configurable per-task

jgraettinger:
The shuffle Log actor engages back-pressure once its aggregate on-disk backlog of sealed segment files crosses a limit. That limit was a single Service-wide value (2 GiB) applied uniformly to every task served by a runtime-sidecar.

Expose it per-task via `shards.shuffleDiskLimit` (bytes) in the task model. The value is carried as a data-plane label `estuary.dev/shuffle-disk-limit` on the shard spec, decoded into `ops.ShardLabeling`, threaded through the runtime-next leader into a new `shuffle.Shard.shuffle_disk_limit_bytes` field, and consumed by the Log handler -- falling back to the Service-wide default when a task doesn't set it. Only runtime-next materializations and derivations shuffle; captures carry the label harmlessly but never use it.

The Service-wide default lives in `DEFAULT_SHUFFLE_DISK_LIMIT_BYTES`, exposed by the runtime-sidecar as the `--shuffle-disk-limit-bytes` flag / `SHUFFLE_DISK_LIMIT_BYTES` env (both previously named `disk-backlog-threshold`; renamed since nothing deployed references them yet). The `disk_backlog_bytes` gauge / `shuffle_log_disk_backlog_bytes` metric are left as-is: they measure the live backlog, not the limit.

The `_bytes` suffix marks the bare-integer carriers (proto fields, Service / Topology fields, the default const, the sidecar flag/env); the user-facing model field and label are left unit-less, with "in bytes" documented on each.


jgraettinger:
Verified on a local stack: a runtime-next derivation with the field set surfaces shuffle_backlog_threshold=134217728 on its shuffle.log handler, while an unset task falls back to the 2 GiB Service default.

williamhbaker (approved):
LGTM

PR #3097, opened 2026-07-01
docs: remove static data plane IP list, reference dashboard

jwhartley:
Follow-up to #3080. Removes the static per-data-plane IP list from the allowlist reference page. The dashboard (Admin -> Settings -> Data Planes) is the source of truth for current data plane IPs, and the page already walks users through finding them there.


skord (approved):
LGTM

PR #3099, opened 2026-07-01
billing: add `--dry-run` mode and always send manual invoices

jshearer:
## `--dry-run`

The trial-run workflow used testmode Stripe, which lacks livemode invoice state, so already-invoiced manual bills showed up as phantom creates. `--dry-run` makes the run read-only so it can be pointed at livemode.

Per-invoice work is split into read-only `classify` and write-only `execute`; a dry run stops after `classify`, so both modes share one decision path. The preview also mirrors the billing-email resolution of a real run and reports tenants that would fail. It cannot validate the final invoice total, which requires creating the invoice.

## Manual invoices always use `send_invoice`

Contracts and one-off charges are sent for the customer to pay rather than auto-charged, regardless of `--charge-type`: at creation, when refreshing an existing draft, and in the send phase. Nothing ever switches a draft back to `charge_automatically`.

## Hardening

- `--recreate-finalized` voids open invoices instead of attempting a delete Stripe rejects, and recovers a run that voided but failed to recreate. Replacements are reported as a distinct `Replaced` result.
- Manual bills that already have an open, paid, void, or uncollectible invoice are skipped as `AlreadyProcessed` instead of erroring, and are excluded from `--recreate-finalized`.
- Payment-method state is read from Stripe directly


crates/billing-integrations/src/publish.rs
```diff
@@ -40,7 +39,8 @@ pub struct PublishInvoice {
     /// The month to generate invoices for, in format "YYYY-MM-DD"
     #[clap(long, value_parser = parse_date)]
```
GregorShear, line 40: nit/suggestion: while we're in here, might be nice to accept the `--month` input as `YYYY-MM`

i'm assuming there's no reason to ever expect `--month 2026-08-15` right?

crates/billing-integrations/src/publish.rs
```diff
@@ -40,7 +39,8 @@ pub struct PublishInvoice {
...
     month: NaiveDate,
-    /// Whether to delete and recreate finalized invoices
+    /// Whether to recreate existing invoices: drafts are deleted and open
+    /// invoices are voided before the replacement is created.
     #[clap(long)]
     recreate_finalized: bool,
```
GregorShear, line 45: nit: `recreate_finalized` sounds like a misnomer to me - this would delete and recreate draft invoices (which i wouldn't think are considered "finalized")?

crates/billing-integrations/src/send.rs
GregorShear, line 147: Could we have better separation of concerns here? IMO `update_draft_collection_methods` should only happen in the `publish-invoices` command and `send-invoices` should validate and skip invalid invoices instead of trying to fix them at the last minute. The `send` command could report back which invoices were skipped and need to be fixed by running `publish` again. 

My interpretation is that `update_draft_collection_methods` could fail but the send operation would continue anyway.

PR #3104, opened 2026-07-02
materialize: remove deprecated `constraints` from materialize protocol

mdibaiee:
**Description:**

- We switched from `constraints` to `projection_constraints` in #2977 and https://github.com/estuary/connectors/pull/4714 and https://github.com/estuary/connectors/pull/4648
- now we can remove the now-deprecated `constraints` from the protocol
- I'm not sure if Dekaf needs to do something about this as well? @jshearer 

**Workflow steps:**

**Notes for reviewers:**


crates/dekaf-connector/src/lib.rs
```diff
@@ -135,25 +134,25 @@ where
...
-                                )
+                            .map(|projection| validated::ProjectionConstraint {
+                                field: projection.field.clone(),
+                                constraint: Some(constraint_for_projection(
+                                    &projection,
+                                    &parsed_inner_config,
+                                )),
                             })
-                            .collect::<BTreeMap<_, _>>();
+                            .collect::<Vec<_>>();
```
williamhbaker, line 149: nit: I think a bare `collect()` would work here.

williamhbaker (commented):
For dekaf, I believe we need to first deploy it everywhere (all data planes + the remaining k8s deployment, IIRC) with its updated code to only emit `ProjectionConstraint`'s. This needs to happen before the control plane agent is deployed with the changes that remove the old `constraints`. I think it would probably be best to do a separate PR / deploy for the dekaf changes first, much like has already been done for all the materialization connectors.

Separately...the `materialize_soak` connector from the `tests/soak` test will also need to be updated. That is a soak test that is manually run.

williamhbaker (approved):
LGTM

PR #3105, opened 2026-07-02
data-plane-controller: fix db timeout

qaoj:
**Description:**

Changes DB timeout from 30s to 2m, which should be enough for it to find an open connection

We actually had to do this for the OIDC server too, stands to reason it needs to happen here: https://github.com/estuary/flow/pull/3062/changes

**Workflow steps:**

N/A

N/A

**Notes for reviewers:**

N/A


mwillman-estuary (approved):
Should fix it.

PR #3107, opened 2026-07-02
agent: flag to create new captures as runtime-v2

williamhbaker:
**Description:**

Adds a `--runtime-v2-new-captures` flag (env `RUNTIME_V2_NEW_CAPTURES`) to the agent. When enabled, any newly-created capture that doesn't already set `enable-runtime-v2` is published with that shard flag set to `true`, placing it on the V2 runtime. Existing captures are never touched, and an explicit per-task flag always takes precedence.

This can be deployed inert: the flag defaults to false, so it only begins to take effect once the agent is configured to enable it.

Manually tested on a local stack: Many scenarios with the `RUNTIME_V2_NEW_CAPTURES` agent config on and off, with new & existing captures being published / re-published. Also, sanity checking of derivation and materialization publishing (unchanged).

**Workflow steps:**

**Notes for reviewers:**

I ended up far from where I started. I was initially considering a mechanism in the runtime itself to decide to run a task on the V1 or V2 runtime, or some kind of computed shard label value. Both of these were complex and insufficient for various reasons: In the runtime, we don't have the creation time of the task readily available, computing the shard label value in all the places we'd need to was not trivial, and I don't think either way would readily extend to derivations since they need to make a decision at build time anyway about which image to run.

While these challenges are not intractable, I think this final "just add the spec flag to the spec" mechanism is the most obviously correct and least intrusive approach, and makes it very easy to tell which tasks are running V2.

### DIFF
--- a/.sqlx/query-9a35831f768dc6a9315d2bfe0b462969292e8065ca2889430e482768a0e36e72.json
+++ b/.sqlx/query-9a35831f768dc6a9315d2bfe0b462969292e8065ca2889430e482768a0e36e72.json
@@ -1,0 +1,22 @@
+{
+  "db_name": "PostgreSQL",
+  "query": "select catalog_name\n               from live_specs\n               where catalog_name = any($1::text[]) and spec is not null",
+  "describe": {
+    "columns": [
+      {
+        "ordinal": 0,
+        "name": "catalog_name",
+        "type_info": "Text"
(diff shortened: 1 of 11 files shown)


dgreer-dev (approved):
LGTM, functionally.  No comment on when to deploy

PR #3110, opened 2026-07-02
runtime-next: add minimum interval for post-txn triggers

### LINKED ISSUE #3060: triggers: add a debounce / minimum fire interval
## Summary

Runtime webhook triggers fire **once per committed materialization transaction that
materializes data**, with no rate-limiting. When several transactions commit in close
succession, this produces a burst of webhook deliveries. The older connector-side
dbt-job trigger avoided this with a **debounce** that collapses repeated triggers into
at most one per configured interval. That debounce was never ported to the more general
runtime triggers. We should add an equivalent.

## Background

Runtime triggers are the general replacement for the connector-specific dbt-job trigger:
any materialization can POST/PUT/PATCH a templated payload to a webhook when it
materializes new data (e.g. to kick off a downstream dbt Cloud run). Config is
`models::TriggerConfig` — `url`, `method`, `headers`, `payload_template`, `timeout`,
`max_attempts`. There is **no** interval / debounce field today
(`crates/models/src/triggers.rs:24`).

### Current firing behavior

A trigger fires for every committed transaction that materialized at least one document:

- **V1 (current production):** trigger variables are computed only when the transaction
  had data — `compiled_triggers.is_some() && !txn.stats.is_empty()`
  (`crates/runtime/src/materialize/serve.rs:253`), persisted to RocksDB at StartCommit
  for at-least-once delivery, then delivered when the connector sends `Acknowledged`
  (`crates/runtime/src/materialize/protocol.rs:385`).
- **V2 (runtime-next):** same gate —
  `task.triggers.is_some() && !extents.bindings.is_empty()`
  (`crates/runtime-next/src/leader/materialize/fsm.rs:642`), delivered in
  `crates/runtime-next/src/leader/materialize/actor.rs:451`.

There is no debounce, batching, or minimum interval anywhere on this path:
**N qualifying transactions → N webhook deliveries.**

## Problem

The canonical use of a trigger is to start an expensive downstream job (a dbt run). One
delivery per transaction means that anything producing several transactions in a short
window — catch-up after a skipped/paused sync interval, a frequent sync cadence, or a
backfill — produces a burst of webhook calls and a burst of downstream job runs. That's
the user-reported symptom ("multiple triggers when a sync schedule is skipped"): the
debounce is the general fix regardless of the exact sequence that produced the burst.

## Prior art: the connector-side dbt debounce

The dbt-job trigger already solved this. After each `Acknowledged` it signals a
background handler that enforces a minimum interval between actual job triggers
(`connectors/go/materialize/transactions_stream.go:154`, default `interval` = 30m,
"only triggered if data has been materialized"):

```go
// dbtJobTriggerHandler: interval-based debounce
//   lastRun          time.Time   // when a job last actually fired
//   nextRunScheduled atomic.Bool // a deferred fire is already queued
//
// On each Acknowledged signal:
//   - if a run is already scheduled        -> do nothing (collapses the burst)
//   - else if now - lastRun >= interval    -> fire immediately
//   - else                                 -> schedule one fire at lastRun + interval
```

Multiple acknowledgements within the interval window collapse into a single eventual
trigger. This logic is connector-local and was never generalized to runtime triggers.

## Proposed solution

Add a minimum-interval / debounce to runtime triggers so that, per trigger config, at
most one delivery occurs per interval, collapsing bursts of qualifying transactions —
while preserving the existing semantics (fire only when data was materialized;
at-least-once delivery).

We only need to implement for Runtime V2.

Likely the smallest viable shape:
- Add an optional `interval` (minimum fire interval) field to `TriggerConfig`.
- On the firing path, suppress a delivery if one fired within `interval`; ensure a
  pending fire still happens eventually (don't silently drop it).

## References

- Trigger config (no interval field today): `crates/models/src/triggers.rs:24`
- V2 fire gate / delivery: `crates/runtime-next/src/leader/materialize/fsm.rs:642`,
  `crates/runtime-next/src/leader/materialize/actor.rs:451`
- Connector dbt debounce (prior art): `connectors/go/materialize/transactions_stream.go:154`,
  `connectors/go/dbt/trigger.go`

dgreer-dev:
**Description:**

Adds an optional minimum fire interval (debounce) to materialization webhook triggers, for the V2 runtime. Fixes #3060.

Triggers previously fired once per committed transaction that materialized data, so a burst of transactions produced a burst of webhook deliveries (potentially costly). With an `interval` set, a burst collapses into at most one delivery per interval, whose payload covers the union of the collapsed transactions.  The old connector-side dbt trigger had an equivalent debounce, this generalizes it to runtime triggers.

Mechanically: each trigger config accumulates a pending window in a leader-lifetime accumulator, keyed by the config's method+URL. At commit time the accumulator is persisted and configs whose interval has elapsed fire with their accumulated window. The rest keep accumulating and fire either at the first commit at-or-after their interval elapses (fast path) or, if the task goes quiet, from the leader's idle loop once the deadline passes (so a debounced window never hangs waiting for a next transaction).

**Workflow steps:**

Add `interval` to a trigger config on a materialization:

```yaml
triggers:
  config:
    - url: "https://cloud.getdbt.com/api/v2/accounts/123/jobs/456/run/"
      headers:
        Authorization: "Token dbt-token"
      payloadTemplate: |
        {"cause": "Estuary trigger {{run_id}}"}
      interval: 30m
```

Unset preserves today's fire-every-transaction behavior; existing specs are unchanged. The field is per-config, so e.g. a Slack notification can fire every transaction while a dbt run on the same materialization is debounced to 30m.

- `site/docs/concepts/materialization/materialization-triggers.md` — updated in this PR: `interval` added to the spec example, properties table, and SOPS-excluded-fields list.

**Notes for reviewers:**

- **Two fire paths, one decision function.** A due window normally fires with the transaction that made it due (commit path, no added latency). If the task goes quiet — frontiers only arrive when journal content advances — `HeadIdle` fires due windows from idle and otherwise sleeps until the earliest pending deadline, so a burst's tail is delivered within its interval even with no further commits. The idle fire requires `Tail::Done` and rotates the Tail into its normal `Trigger` → `Persist` → `Done` sequence, so it can't race a commit-path fire.
- **Persisted format change + compatibility.** The durable "trigger-params" blob changes from a single `TriggerVariables` to a `{method+url → TriggerVariables}` map. Both runtimes decode through the new serde-untagged `models::triggers::PersistedTriggerParams` (the formats are mutually unambiguous), so: V2 recovering a pre-upgrade blob fans it out to all configs (its original fire-all semantics), and V1 after a V2→V1 rollback merges a map blob into one window (its normal fire-all behavior). Residual: a *binary* downgrade to a release before this PR has neither fallback — don't downgrade past this release while a trigger is mid-debounce.
- **Debounce state is in-memory** (`last_fire`): a leader failover resets the window, worst case one early fire — same as pre-debounce behavior. Recovery re-fires the persisted accumulator verbatim (at-least-once), which can also mean one early fire post-recovery.
- **Duplicate configs** (same method+URL) aren't supported since the key is the debounce identity: first config wins, later duplicates WARN at task startup and never fire. A pending window whose config was removed on republish is dropped with a WARN rather than failing the task (failing would crash-loop, since the config stays gone for the build's life).
- **V1 is otherwise untouched**: it accepts but ignores `interval` (no debounce), and its only code change is the tolerant `PersistedTriggerParams` decode above.

### DIFF
--- a/crates/models/src/snapshots/models__triggers__test__trigger-config-schema.snap
+++ b/crates/models/src/snapshots/models__triggers__test__trigger-config-schema.snap
@@ -15,6 +15,15 @@ expression: schema
       "items": {
         "$ref": "#/$defs/TriggerConfig"
       }
+    },
+    "interval": {
+      "title": "Minimum interval between trigger deliveries.",
+      "description": "A burst of transactions within the interval collapses into a single\ndelivery covering the full span.",
+      "type": [
+        "string",
+        "null"
+      ],
+      "pattern": "^\\d+(s|m|h|d)$"
     }
   },
   "additionalProperties": false,
--- a/crates/models/src/triggers.rs
+++ b/crates/models/src/triggers.rs
@@ -8,6 +8,16 @@ use std::time::Duration;
 #[derive(Serialize, Deserialize, Clone, Debug, JsonSchema, PartialEq)]
 #[serde(deny_unknown_fields, rename_all = "camelCase")]
 pub struct Triggers {
+    /// # Minimum interval between trigger deliveries.
+    /// A burst of transactions within the interval collapses into a single
+    /// delivery covering the full span.
+    #[schemars(schema_with = "super::duration_schema")]
+    #[serde(
+        default,
+        with = "humantime_serde",
+        skip_serializing_if = "Option::is_none"
+    )]
+    pub interval: Option<Duration>,
     /// # Trigger Configurations
     /// List of webhook triggers to fire when new data is materialized.
     pub config: Vec<TriggerConfig>,
@@ -92,6 +102,24 @@ impl TriggerVariables {
             run_id: "2024-01-01T00:00:00.000Z".to_string(),
         }
     }
+
+    /// Merge `other` into `self`, widening this window to cover both. Used to
+    /// collapse a burst of debounced transactions into a single delivery.
+    pub fn merge(&mut self, other: &TriggerVariables) {
+        for name in &other.collection_names {
+            if !self.collection_names.contains(name) {
+                self.collection_names.push(name.clone());
+            }
+        }
+        self.collection_names.sort();
+
+        if other.flow_published_at_min < self.flow_published_at_min {
+            self.flow_published_at_min = other.flow_published_at_min.clone();
+        }
+        if other.flow_published_at_max > self.flow_published_at_max {
+            self.flow_published_at_max = other.flow_published_at_max.clone();
+        }
+    }
 }
 
 /// Render a single payload template string with the given context.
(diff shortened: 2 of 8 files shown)


dgreer-dev:
Reworked per feedback, should be much simpler now:

- interval is now a single top-level triggers, shared by all configs, rather than per-config. Debounce state collapses to one pending window + one last-fire clock.
- format in rocksdb is the same for the v1/v2, so no complexity there

williamhbaker (approved):
LGTM

PR #3116, opened 2026-07-03
json: require RFC3339 'T' separator in date-time format validator

### LINKED ISSUE #3108: json: DateTime format validator silently drifted lenient (accepts space separator) via dependency bump, contradicts platform's RFC3339 stance
## 1. Priority
Medium — not user-facing on its own, but it's the root enabler of a bug that's already crash-looped 3 customers' materializations (estuary/connectors#4725) and blocks the product decision from the 2026-07-02 daily sync ("Estuary should strictly reject non-conforming alleged-RFC3339 timestamps").

## 2. Root cause (confirmed)
`Format::DateTime::validate()` (`crates/json/src/schema/formats.rs:88-90`) delegates directly to `time::OffsetDateTime::parse(val, &time::format_description::well_known::Rfc3339)` with no separator check of its own:
```rust
Self::DateTime => {
    time::OffsetDateTime::parse(val, &time::format_description::well_known::Rfc3339)
        .is_ok()
}
```
The `time` crate started accepting a space in place of `T` in v0.3.37 (time-rs/time#700, merged 2024-08-27; see CHANGELOG). Flow was pinned at `time` 0.3.36 (predates the change) until `estuary/flow#2302` ("update Rust and Go dependencies", merged 2025-10-08, commit `302d0d17105da4c2933f5e696e15db9adb71ea40`) bumped it to 0.3.44 — crossing that boundary. PR #2302's description states "No changes to function / behavior," so this was an unreviewed, unintended side effect, not a deliberate platform decision.

This function is used both for format *detection* during schema inference (`crates/doc/src/shape/widen.rs:15`, via `Format::detect()`) and for format *validation* against a schema's declared `format` keyword (`crates/json/src/validator/validation.rs:519-522` → `Outcome::FormatNotMatched`, which is a real, enforced validation error, not advisory — confirmed via `crates/doc/src/validation.rs:187`'s `error_filter`, which passes through anything but a bare `Annotation`).

## 3. Why it hasn't been caught by enforcement yet
Runtime ingestion validates documents against a collection's **write schema** (`crates/runtime/src/capture/task.rs:168-171`), not the inferred read schema. In the observed incidents, `format: date-time` exists only in the inferred read schema — the write schema doesn't constrain the field — so no `Keyword::Format` exists to violate at ingestion time. The validator's enforcement path is real and already wired; it simply isn't reached for these fields today.

## 4. Proposed fix
Harden `DateTime::validate()` to require a literal `T` (or whatever separator the platform decides on) instead of delegating blindly to the `time` crate's current behavior — e.g. an explicit check on the separator byte before/alongside the `time::OffsetDateTime::parse` call, plus a regression test locking in the space-separator-is-invalid case (none currently exists in `formats.rs`'s test table). This is a small, contained change:
- Stops schema inference from mistagging mixed-format fields as `date-time` going forward.
- Automatically strengthens enforcement anywhere a write schema already declares `format: date-time`, with no additional plumbing — the rejection path already exists end-to-end.

Out of scope here: whether inferred (read-schema) formats should ever feed into write-schema enforcement is a separate, larger architectural question not addressed by this fix.

## 5. References
- estuary/connectors#4725 — materialize-sql converter fix (liberal/normalizing side, per product guidance)
- estuary/connectors#4726 — source-klaviyo emits non-RFC3339 consent timestamps
- estuary/connectors#4743 — materialize-sql PR implementing the liberal-converter side
- estuary/flow#2302 — the dependency bump that introduced the drift
- time-rs/time#700 — upstream change that made `time` accept space separators
- Slack thread (#integrations): https://estuaryworkspace.slack.com/archives/C03Q2NRFKDL/p1782706029810379

jacobmarble:
**Description:**

Fixes #3108. The `date-time` format validator silently drifted lenient when `time` was bumped 0.3.36 → 0.3.44 (#2302): since time-rs/time#700 the crate's RFC3339 parser accepts *any* byte as the date/time separator, not just a space. JSON Schema defines `format: date-time` as the RFC3339 ABNF rule, which permits only `T` (case-insensitive), so the validator now checks the separator byte explicitly before delegating to `time`.

**Workflow steps:**

No workflow change. Strings like `2022-09-11 10:31:25Z` are no longer detected or validated as `date-time` — schema inference stops mistagging mixed-format fields, and existing write schemas declaring `format: date-time` reject them.

None.

**Notes for reviewers:**

- Byte index 10 is safe because RFC3339 `full-date` is fixed-width; lowercase `t` stays accepted per ABNF case-insensitivity (consistent with existing lowercase `z` handling).
- Regression tests added to the `test_format_cases` table; `cargo test -p json` and `cargo test -p doc --features combine` pass with no snapshot fallout.

### DIFF
--- a/crates/json/src/schema/formats.rs
+++ b/crates/json/src/schema/formats.rs
@@ -86,6 +86,13 @@ impl Format {
                 .is_ok()
             }
             Self::DateTime => {
+                // Since 0.3.37, `time`'s RFC3339 parser accepts any byte as the
+                // date/time separator (time-rs/time#700). JSON Schema's `date-time`
+                // is the RFC3339 ABNF rule, which permits only `T` (case-insensitive),
+                // so check the separator ourselves.
+                if !matches!(val.as_bytes().get(10), Some(b'T' | b't')) {
+                    return false;
+                }
                 time::OffsetDateTime::parse(val, &time::format_description::well_known::Rfc3339)
                     .is_ok()
             }
@@ -214,6 +221,10 @@ mod test {
             ("date-time", "2022-09-11T10:31:25+00:00", true),
             ("date-time", "2022-09-11T10:31:25-00:00", true),
             ("datetime", "2022-09-11T10:31:25.123Z", true), // Accepted alias.
+            ("date-time", "2022-09-11 10:31:25.123Z", false), // Space separator (regression: time-rs/time#700).
+            ("date-time", "2022-09-11x10:31:25.123Z", false), // Arbitrary separator.
+            ("date-time", "2022-09-11t10:31:25.123Z", true), // Lowercase `t` is valid per RFC3339 ABNF.
+            ("date-time", "2022-09-11", false),              // Date only, no separator position.
             ("date-time", "10:31:25.123Z", false),
             ("time", "10:31:25.123Z", true),
             ("time", "10:31:25.123z", true),


williamhbaker (approved):
LGTM

PR #3117, opened 2026-07-04
flowctl: re-tool `preview` on the runtime-next stack

jgraettinger:
## What & why

`flowctl preview` now runs on the **runtime-next + shuffle** stack instead of the
legacy V1 `runtime::harness`. It was developed on this branch as `flowctl raw
preview-next`, then promoted to be *the* `preview`; the V1 preview harness — the
last consumer of `runtime::harness` — is deleted along with the now-dead
`runtime/src/harness/*` and its private `exchange` combinator.

The user-facing contract is unchanged: the CLI flags, stdout/stderr line shapes,
and exit codes that `estuary/connectors` CI depends on. `--shards` and
`--debug-port` are new, additive flags.

To get there, runtime-next's host-specific behavior — how a task publishes
documents, reads its sources, and emits logs — was factored into three **factory
seams**. Production installs journal-backed factories; `preview` installs
output-capturing / fixture-replay factories from `flowctl`. runtime-next is now
preview-agnostic, so the lone `Task.preview` flag — the "write documents to
output instead of collections" wire control — is removed from the runtime proto
(field 2 reserved); *how* a task publishes and logs is now a host concern of the
installed factories, not a wire flag. (`max_transactions` stays as the
transaction bound, re-documented as a general preview/harness control.)

## The three host seams (`runtime-next`)

Each seam is a factory the leader and shard `Service`s are generic over (static
RPITIT, not `dyn`):

- **publish** — a `PublisherFactory` / `JournalPublisher` seam replaces the
  `Publisher` enum (a `Real` variant plus a `Preview` variant that baked
  NDJSON-to-stdout output into the crate); the publisher now carries only
  document output, with its former observation calls moved to the logger seam.
- **shuffle** (leader) — a new `leader/shuffle.rs` provides `ShuffleSession(Factory)`
  + `ShuffleServiceFactory`, factoring out the leader's previously-inline
  shuffle-`Session` opening and checkpoint sourcing.
- **log** — new `Logger` / `LoggerFactory` seam sinking both connector logs
  (`log`) and structured `LogEvent`s (`event`) — a `#[non_exhaustive]` enum
  (persist / applied / inferred-schema / container lifecycle) carrying borrowed,
  verbatim payloads so a host can intercept them structurally. `LogEvent::to_log`
  flattens each event into the ops-log line it replaced, and `event` defaults to
  that flattening, so a host overrides only the events it cares about. Production
  shards install `FnLoggerFactory`; leaders and tests install
  `TracingLoggerFactory`.

New harness helpers: `seed_initial_connector_state` / `read_final_connector_state`
seed and read shard zero's RocksDB by path; `shuffle::log::BlockMeta` is
re-exposed for segment writers.

## Preview on the seams (`flowctl`)

- **publish**: `PreviewPublisherFactory` writes captured/derived documents to
  stdout as NDJSON.
- **log**: `PreviewLoggerFactory` sinks the task log and powers `--output-state`
  / `--output-apply` (the legacy `connectorState` / `applied` lines) by
  intercepting `LogEvent::Persist` / `LogEvent::Applied` structurally and
  flattening everything else through `LogEvent::to_log`. The run-end final reduced
  state comes from re-opening shard zero's RocksDB once the session loop releases
  it.
- **shuffle**: `PreviewShuffleFactory` selects per run between a live in-process
  journal-reading `Session` and a `--fixture` replay (`FixtureOpener`).
- `--initial-state` seeds shard zero's RocksDB via `seed_initial_connector_state`.
  `--delay`, fixtures, and `--sessions` planning are unchanged.

## Tests / soak

- Deleted `plans/runtime-v2/preview-harness.md`, a closed bring-up doc for `raw
  preview-next`.
- Soak README updated to the new command name.

## Notes for reviewers

- Depends on the incidental fixes in #3121 only insofar as they were split off
  this branch; the preview work does not require them, so the two can land in
  either order.
- Behavioral compatibility with legacy preview was verified before the cutover;
  the promotion commit is mostly a module move (`raw/preview_next/` → `preview/`)
  plus deletion of the V1 harness.


crates/runtime-next/src/logger.rs
```diff
@@ -0,0 +1,434 @@
...
+                fields.push(("actionDescription", json_field(&action)));
+                (ops::LogLevel::Info, "connector applied")
+            }
+            LogEvent::InferredSchema {
+                collection_name,
+                binding,
+                schema,
+                ..
+            } => {
+                fields.push(("collection", json_field(collection_name)));
```
williamhbaker: Wanting to confirm this field name...should it be `collection_name`? Since the L1 inferred schemas rollup is looking for that: https://github.com/estuary/flow/blob/b98d27ceaba17bf9368434480d59943f4d753553/ops-catalog/data-plane-template.flow.yaml#L60

crates/runtime-next/src/logger.rs
```diff
@@ -0,0 +1,434 @@
...
+
+        let (level, message) = match self {
+            LogEvent::Persist { persist, .. } => {
+                if persist.connector_patches_json.is_empty() {
+                    return None;
+                }
+                // The patch payload is valid JSON (a tab-delimited JSON array;
+                // see `crate::patches`), so it embeds verbatim.
+                fields.push(("patches", persist.connector_patches_json.clone()));
+                (ops::LogLevel::Debug, "persisted connector-state delta")
```
williamhbaker, line 150: Will these get filtered out for `info` and above level tasks, since they don't go through Tokio's tracing subscriber?

williamhbaker (commented):
A couple of questions

crates/runtime-next/src/logger.rs
jgraettinger: Yes good catch. I'll fix, but I'd also like to work schema inference into the soak test so it provides a positive verification it's working.

crates/runtime-next/src/logger.rs
jgraettinger: Yes. We're paying a bit of cost that `tracing` avoids, in that we're actually building the structured event to have it then be filtered by `tracing` at point of emission. I wasn't concerned though (Bytes ref-count clone, or bare &ref)

jgraettinger:
PTAL -- fixed the schema inference regression, and added new coverage in the soak test to provide E2E verification.

williamhbaker (dismissed):
LGTM

williamhbaker (approved):
LGTM

Issue #2985, opened 2026-05-29
Nonsensitive Field Overlays for Endpoint Configs


willdonnelly:
We keep wishing that we had a mechanism to allow certain parts of connector endpoint configs to be modified without requiring decryption or re-encryption of the config.

This has been coming up every so often for years in the context of "why do I have to re-enter my password to change the backfill chunk size" questions, and now it's come up again in a new guise as part of the feature-flags pinning work (because we'd like connectors to edit their own `/advanced/feature_flags` field without needing to re-encrypt the whole config).

The fundamental issue here is that encrypted endpoint configs verify integrity with a MAC so that even plaintext fields can'd be modified. This is necessary in some cases, for instance it definitely shouldn't be possible to go edit the hostname or username fields of an endpoint config while retaining the existing encrypted password field. But there are a lot of other fields which simply are not security relevant in that manner, and so it would be nice if we could mark those ones as safe to modify outside of MAC validation.

## Design

To start with, we'll need a JSON schema annotation which indicates that a given config field is not security-relevant and should be freely modifiable without a user needing to re-enter credentials. I propose `"nonsensitive": true`.

We will eventually need a human to vet all properties of all connector configs, but the explicit "okay to not MAC" whitelisting nature of these annotations means that any connectors we haven't gotten around to will simply treat all fields as sensitive until we do, which is exactly the right fail-safe behavior for this feature.

### Overlay Representation

We introduce a new property `sops.overlay` within the SOPS metadata stanza of an encrypted endpoint config, which carries a JSON object structured to mirror the overall endpoint config.

At decrypt time, the overlay is merged into the SOPS-decrypted document using JSON Merge Patch (RFC 7396) semantics, which we already use all over. The conceptual operation is:

```
decrypted = merge_patch(sops_decrypt(encrypted), encrypted.sops.overlay)
```

(Note: We don't have to strip the `sops.overlay` property before decryption, empirically SOPS will ignore unrecognized properties within the SOPS stanza)

A nonexistent overlay implies skipping the merge or merging with the no-op patch object `{}`, which means that this handling of overlays is backwards compatible with all existing endpoint configs in production.

For example, consider a `source-postgres` endpoint config that marks fields like `feature_flags` and `backfill_chunk_size` as `nonsensitive` but errs on the side of caution by leaving `publicationName` and `slotName` covered by the MAC. Its encrypted form would look like:

```json
{
  "address": "db.example.com:5432",
  "user": "flow_capture",
  "database": "postgres",
[...]

PR #3119, opened 2026-07-06
capture+materialize: Support Nonsensitive Field Overlays

willdonnelly:
**Description:**

This PR:
- Adds a new `nonsensitive: true` JSON schema annotation for endpoint configs.
- Adds an `unseal::decrypt_with_overlay` helper function which validates overlay non-sensitivity and does the extract-decrypt-merge dance around overlay properties and the main config.
- Adds `sealed_config_json` fields to the `Open` RPCs in the capture and materialization protocols, which will be used to provide the encrypted endpoint config to connectors. Currently they only receive the decrypted config, but in order to produce `configUpdate`s which modify their own configuration via the overlays they'll need access to the raw encrypted one too.
- Refactors the capture and materialization task startup process (in both the old and new runtimes) so that endpoint config decryption is deferred until after the initial Spec RPC. This is necessary because in order to properly validate that an overlay only modifies nonsensitive fields we need the config schema from the spec response. This is mostly a behavior-preserving refactor except that invalid configs would now produce an error _after_ the initial spec RPC where previously the error would occur before the connector was launched at all.
- Swaps out `unseal::decrypt_sops` for `unseal::decrypt_with_overlay` in those places. This is basically the followup to the previous refactor, I just found it cleaner to make them separate commits.
- Populates the `sealed_config_json` property of Open RPCs.

Since I'm not super familiar with the runtime codebase I've broken this down into fairly small, atomic commits so that _I_ could review them individually and be pretty confident they're correct.

Part of https://github.com/estuary/flow/issues/2985

**Workflow steps:**

Nothing really changes. In theory someone could manually produce a task with an endpoint overlay but nothing will be generating them automatically yet, because we need the runtime to fully support this before we start using the feature.


willdonnelly:
I believe this is ready for review now. I've tested locally that with these changes a source-hello-world->materialize-postgres pipeline can be set up as normal when not using an overlay, and that an overlay can be added without re-encrypting a task config and (with a connector modified to mark those fields as nonsensitive) will be applied to the effective task config.

I have not specifically tested this with both the v1 and v2 runtimes because I don't know how to toggle between those, but the actual code changes are basically identical between the two so I think that's probably sufficient testing just to make sure I didn't break something dumb but subtle.

willdonnelly:
Hmm, the "Dekaf Test" CI failure appears to be because the `source-http-ingest` connector it uses hasn't been bumped to have https://github.com/estuary/flow/pull/3059 yet, I think?

crates/unseal/src/overlay.rs
```diff
@@ -0,0 +1,248 @@
...
+    // Otherwise iterate object fields and check each subproperty
+    for (property, child) in fields {
+        ptr.push(Token::Property(property.clone()));
+        check_location(root, ptr, child)?;
+        ptr.0.pop();
+    }
+    Ok(())
+}
+
+fn display_ptr(ptr: &json::Pointer) -> String {
```
williamhbaker: Any reason not use the `Display` already implemented by `json::Pointer`?

crates/unseal/src/overlay.rs
```diff
@@ -0,0 +1,248 @@
...
+        // Anything within a `nonsensitive` subtree is permitted, including new
+        // properties and nested objects the schema does not enumerate.
+        check(json!({"tunables": {"parallelism": 8, "extra": {"deep": true}}})).unwrap();
+    }
+
+    #[test]
+    fn empty_overlay_is_a_noop() {
+        check(json!({})).unwrap();
+        // An empty object writes nothing even at an otherwise-sensitive location.
+        check(json!({"credentials": {}})).unwrap();
```
williamhbaker, line 168: I can't think of a reason why this would need fixing, but if I'm understanding this correctly an empty object would _kind of_ write something to a sensitive location if it were not of type object (like a string), replacing it instead with an object. In any sane world the resulting config would fail to parse in the connector, or otherwise not have any kind of exposure risk.

williamhbaker: The worst I can think of is a connector that had something like `tls: true` as a sensitive field and then somehow it thought `tls: {}` meant `false` 🤷

williamhbaker (dismissed):
LGTM, no blockers, a couple of things to consider as comments, and also concurring deployment:

As is, this PR will start to populate the new sealed config field right away. We've been bitten numerous times by new field additions where consumers use the strict pbjson structs that disallow unknown fields. As you noted with `source-http-ingest`, it hadn't yet been updated with the recent move to more lenient pbjson structs that do allow unknown fields. Anything else that hasn't yet been updated will similarly crash if it sees these populated fields.

Off the top of my head, that would be the other rust connectors: source/materialize-kafka, and all the ATF connectors (at least, I don't remember seeing us having updated those recently, but I might have missed it).

Since the new field only shows up on `Open` messages it is just limited to consumers of `Open` messages, which is only connectors. 

I'm betting this would also break `flowctl preview` and `flowctl raw materialize-fixture` unless an up-to-date `flowctl` is used, built from at least this commit or later. This is a practical concern for connectors repo integration tests, unless they are always pulling/building the latest `flowctl`.

crates/unseal/src/overlay.rs
willdonnelly: The only functional difference I can see is that using `Display` an overlay which attempts to modify the root object would produce an error message like `overlay modifies location , which is not marked nonsensitive` (since the RFC 6901 representation of the root is the empty string) whereas this helper function special-cased the root as `/` for readability in the error message.

That probably isn't sufficient reason to reimplement a hackier version of JSON Pointer stringification so I've removed this helper.

willdonnelly: Actually that's a dumb objection anyway, I've added single-quotes so that error would be `overlay modifies location '', which is not marked nonsensitive` which makes it pretty unambiguous to me.

crates/unseal/src/overlay.rs
willdonnelly: Good catch, yeah, an empty object (or even a nested chain of objects as long as the final one is empty) passes the current nonsensitive-property validation but it probably shouldn't since `merge(scalar, object) = merge({}, object)`.

I think to close that gap the validation rule needs to be something like "we only recurse into an object if the target schema permits an object at this location" and an object in a scalar location needs to be rejected.

mdibaiee:
Am I right in my understanding that this allows connectors to emit an update to a non-encrypted field (e.g. `advanced.feature_flags` without having to call the decryption endpoint? I love the feature

willdonnelly:
> A couple of things to consider as comments

Thanks, I believe those should be addressed now.

> As is, this PR will start to populate the new sealed config field right away. We've been bitten numerous times by new field additions where consumers use the strict pbjson structs that disallow unknown fields. As you noted with `source-http-ingest`, it hadn't yet been updated with the recent move to more lenient pbjson structs that do allow unknown fields. Anything else that hasn't yet been updated will similarly crash if it sees these populated fields.
> 
> Off the top of my head, that would be the other rust connectors: source/materialize-kafka, and all the ATF connectors (at least, I don't remember seeing us having updated those recently, but I might have missed it).

Is your recommendation to split this up into separate pieces, adding the field and making sure the connectors pick that up and then populating it in a separate PR once that's done, or are you just calling out that we should make sure everything's picked up the recent allow-unknown-fields change before merging this?

If it's the latter, I think the current state of things is:
- All Go connectors would be fine, because they've never forbidden unknown fields
- All Python connectors would be fine for the same reason
- Rust Connectors
  - `source-http-ingest` was just updated yesterday to have the allow-unknown-fields change
  - `source-kafka` was updated a week or two ago
  - `materialize-kafka` uses protobuf rather than JSON so it _should_ be unaffected AFAIK
- ATF Connectors: Current Flow pin before the allow-unknown-fields change, so still vulnerable

TL;DR it looks like only ATF connectors still need any work, we would need to bump their Flow version pin and make sure they're all released before merging this.

> I'm betting this would also break `flowctl preview` and `flowctl raw materialize-fixture` unless an up-to-date `flowctl` is used, built from at least this commit or later. This is a practical concern for connectors repo integration tests, unless they are always pulling/building the latest `flowctl`.

The integration tests do actually pull down the latest flowctl ([`fetch-flow.sh`](https://github.com/estuary/connectors/blob/main/fetch-flow.sh) pulls it from the latest `dev-next` release at the start of each CI run). But I'm not sure I understand the mechanism here though, what exactly would break when using an older `flowctl` build?

williamhbaker:
> Is your recommendation to split this up into separate pieces, adding the field and making sure the connectors pick that up and then populating it in a separate PR once that's done, or are you just calling out that we should make sure everything's picked up the recent allow-unknown-fields change before merging this?

Yeah I think making sure everything's picked up the allow-unknown-fields change would be the practical way to go about it. Doesn't seem like there's much additional needed for that.

>  I'm not sure I understand the mechanism here though

I was probably wrong about this actually. I was thinking that since `flowctl preview` etc. read these values from connector RPCs, that they would be at risk of not having them in their pbjson structs. But it is `flowctl` itself that would have added them, so they should actually be in-sync by construction.

williamhbaker (dismissed):

willdonnelly:
> Yeah I think making sure everything's picked up the allow-unknown-fields change would be the practical way to go about it. Doesn't seem like there's much additional needed for that.

Okay, thanks. I've uploaded https://github.com/estuary/airbyte/pull/360 for the ATF connectors and will make sure that's merged and pushed to all 40-something of them before this gets merged.

williamhbaker (approved):

willdonnelly:
I updated the ATF connectors on Friday and have confirmed that the CI build updated the built image for them all, so I don't think there are any remaining obstacles to merging this now.

PR #3120, opened 2026-07-06
dekaf: log fetch frame when zero offset

danielnelson:
**Description:**

Additional logging to validate the frame parsing to troubleshoot an ongoing issue with Singlestore.  Occasional frames are parsed with a zero offset but on the Singlestore side frames are logged with the correct offset.

**Workflow steps:**

Enable debug logging.

None

**Notes for reviewers:**

### DIFF
--- a/crates/dekaf/src/lib.rs
+++ b/crates/dekaf/src/lib.rs
@@ -464,7 +464,20 @@ async fn handle_api(
         }
 
         ApiKey::Fetch => {
-            let (header, request) = dec_request(frame, version)?;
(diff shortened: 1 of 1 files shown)


jshearer (approved):
This is fine and will unconditionally log any time there's a `Fetch` request that matches your condition. I might have added the logging at some point in https://github.com/estuary/flow/blob/master/crates/dekaf/src/session.rs#L590 instead, which is the "real" Fetch handler and might give you some more detail to include in the log. Here is fine too though

danielnelson:
I'm going to close this one, I thought I would need the raw request frame to demonstrate the 0 offset but it doesn't seem like the evidence will be required.

PR #3121, opened 2026-07-06
Incidental fixes: gazette Append retry, connector-init image, materialize tear-down

jgraettinger:
## What & why

A handful of small, independent fixes that were originally carried along on the
`flowctl preview` re-tooling branch (#3117) but don't depend on — and aren't
required by — that work. Splitting them out so they can land on their own.

- **gazette: retry `Append` flow-control underflow instead of hard-failing.** A
  broker aborts an Append when the client can't sustain `MinAppendRate`; the
  rollback is clean and safe to retry (gazette's Go `AppendService` retries
  indefinitely). It surfaces as an Unknown-coded gRPC status that
  `Error::is_transient()` treated as terminal. Classify the underflow as
  transient so it rides the existing retry + backoff in `journal::Client::append`.

- **runtime: source `flow-connector-init` from the reactor image.** The pinned
  extraction image `ghcr.io/estuary/flow` is no longer published; its newest tag
  predates the `projection_constraints` Validated field by ~6 months. An old
  `flow-connector-init` silently drops that field, so materializations emitting
  list-form constraints fail validation with a spurious "group-by key field has no
  constraint" error. Repoint both the V1 `runtime` and `runtime-next` pins at the
  reactor image (same statically-linked binary, tracks current flow).

- **runtime-next: fix materialize 60s tear-down stall.** The materialize leader
  FSM returned `Action::Idle` at a clean transaction boundary while stopping,
  parking the actor for a full 60s `ACTOR_TICK_INTERVAL` before it noticed
  `Head::Stop`. Return `Action::PollAgain` to re-poll immediately, matching the
  capture / derive FSMs.

- **tests/soak: materialize `views` on `materialize-postgres:dev`.** The
  synchronous `StartedCommit` change the V2 remote-authoritative checkpoint path
  needs has merged into `:dev`, so drop the branch-build pin.

### DIFF
--- a/crates/gazette/src/lib.rs
+++ b/crates/gazette/src/lib.rs
@@ -94,6 +94,21 @@ impl Error {
                     true
                 }
 
+                // A broker aborts an Append RPC when its client fails to sustain
+                // the minimum data rate (gazette's MinAppendRate flow-control
+                // policing). This sheds a slow or contended client so others may
+                // proceed; the append rolled back cleanly and is safe to retry.
+                // Gazette surfaces it as an Unknown status wrapping
+                // ErrFlowControlUnderflow, so match on the stable core phrase
+                // rather than the wrapped-context prefix.
+                tonic::Code::Unknown
+                    if status
+                        .message()
+                        .contains("didn't sustain the minimum append data rate") =>
+                {
+                    true
+                }
+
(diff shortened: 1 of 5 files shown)


williamhbaker (approved):
LGTM

PR #3126, opened 2026-07-07
docs: explain why exclusiveCollectionFilter needs few enabled bindings

jwhartley:
The `exclusiveCollectionFilter` description said only that it 'should only be used if a small number of bindings are enabled', with no explanation of why or what to use otherwise. A recent support case enabled it on a capture with several hundred enabled bindings: the generated `$or` (one clause per enabled binding, evaluated against every oplog event) slowed change stream reads below the database's write rate and lag grew instead of shrinking. Reverting to `excludeCollections` resolved it.

One-line change: the property description now states the mechanism and points to `excludeCollections` as the alternative when many bindings are enabled.


aeluce (dismissed):
LGTM, thanks!

Non-blocking, but optional considerations:
* Do we want to add this explainer to MongoDB variants as well (DocumentDB and CosmosDB pages)?
* Cell size as-is is okay for the table. If we wanted to extend the information about `exclusiveCollectionFilter` any further, though, we might want to break exclusion filtering out into its own section further down the page and link out to it from the table.

jwhartley:
Good idea to add it to the variants - could you please re-review

aeluce (approved):

Issue #2675, opened 2026-02-13
Increase tenant quotas after user enters payment method


psFried:
The goal is to automatically increase the task and collections quotas after a user enters a credit card.

jshearer:
## Problem

Tenant task and collection quotas live on the `tenants` table (`tasks_quota`, `collections_quota`, defaulting to 10 and 500). When a tenant becomes paid, either by setting a default payment method or by being marked as an external-payment customer (I think), these quotas are not raised automatically. The increase is applied by hand with a direct SQL update, so a customer who has started paying stays capped at trial limits until someone runs that update.

## Goal

Raise a tenant's quotas to the paid floor automatically when it becomes paid, where "paid" means either of:

- the tenant has a non-null default payment method in Stripe, or
- the tenant's `payment_provider` is set to `external`.

The raise is one-way: it only ever increases a quota, never decreases it, and it never lowers a quota that has been set higher by hand. Stripe is the source of truth for whether a default payment method exists.

## Approach

Drive this from the per-tenant controller added in #2902. That controller already runs a reconcile loop per tenant with a wake primitive and retry/backoff, and it already talks to Stripe. Add a quota reconcile that:

1. Determines paid status as `(Stripe reports a non-null default payment method) OR (payment_provider = 'external')`. The Stripe side is read live from Stripe at reconcile time, so the decision is sourced from Stripe rather than a cached flag.
2. Computes the desired quota. If paid, `tasks_quota` and `collections_quota` are each raised to at least the paid floor, applied as `GREATEST(current, floor)` so the change ratchets up and leaves higher manual values untouched.
3. Is idempotent, so it can run repeatedly and converge.

## Triggering without webhooks

We are not adding Stripe webhooks, so the controller is woken from the paths we already control, with a periodic reconcile as the backstop:

- Default payment method set: wake the tenant controller from the `setBillingPaymentMethod` GraphQL mutation, right after the Stripe default-payment-method write it already performs. The dashboard's payment-method actions are being routed through that mutation (estuary/ui#2004), so it is the right home for the wake. 
- Marked external: a trigger on `tenants.payment_provider` wakes the controller, the same pattern as the billing-contact trigger in #2902.
- Backstop: the controller's scheduled reconcile covers tenants that are not yet paid and catches default-payment-method changes made outside the dashboard, for example a card set directly in Stripe. Because each tenant reconciles on its own schedule, these reads spread out over time rather than arriving as a burst.

## Open questions

[...]

PR #3127, opened 2026-07-07
Adding support for updating the quotas when payment info changes

bbartman:
**Description:**

Update the `TenantController` to support also checking for and updating the tasks and collections quotas when we detect new payment information.

**Workflow steps:**

This makes other workflows easier by removing the need for manual updates to the DB when new payment info is detected.

N/A?

**Notes for reviewers:**

Please let me know if I'm missing anything. I'm not sure how to verify behavior in a larger context of the system.

**Other conversations not held here**

**James Hartley  [8:10 PM]**

> 1. Do we want to keep checking on an interval if we check the payment_provider within our system is set to stripe, and we don’t have payment information inside of stripe.Does this happen in practice? If yes, then probably yes, although I’d like to understand when this happens before we bake in more complexity to the solution

>2. Once a quote is raised is there ever a reason to raise it again? Do we need to check a 2nd time after it’s already been done once? Would there ever be a case where it would be lowered and need to be raised again?I think further increases (customer who uses a huge number of tasks/collections) can be left to manual (current status quo). No further checks. If we manually lower it there’s probably a good reason and would want to leave it lower (edited) 

**Joseph Shearer  [8:18 PM]**

>1 is asking about a pretty rare situation: if a payment method is set as default in Stripe outside of an API call to agent-api, such as if we manually set it for some reason. It probably has never happened and I can't think of a good reason for why it would, but it could

**James Hartley  [8:24 PM]**

>that’s helpful, I’d say skip repeated checking then

** Things learned **

It is possible not to have a customer within strip. 

The min values for quotas is 100 and the min value for collections is 10000.

There isn't parity with waking the tenant controller when a payment method is deleted. The quotas stay even if the payment information is removed.


crates/agent/src/tenant_controller/quotas.rs
```diff
@@ -0,0 +1,330 @@
...
+
+    let Some(customer) = customer else {
+        return Ok(());
+    };
+    // There are two ways to set the default payment method:
+    // 1. customer.default_source.is_some() (old)
+    // 2. customer.invoice_settings.default_payment_method.is_some() (new)
+
+    // This is the old way of checking for a default payment method according to AI.
+    let has_default_payment_configured = if customer.default_source.is_some() {
```
jshearer: `billing::default_payment_method_id(&customer).is_some()` already encodes this, and it's used a few lines away in the same mutations file this PR edits. That helper checks only `invoice_settings.default_payment_method`, so hand-rolling the extra `default_source` branch makes 'paid' mean something broader for quotas than it does everywhere else in billing. Reusing the helper keeps the definition consistent and drops the `default_source` / 'according to AI' guesswork. The in-memory provider sets `invoice_settings.default_payment_method`, so the tests stay green.

crates/agent/src/tenant_controller/mod.rs
```diff
@@ -85,16 +116,45 @@ impl Executor for TenantController {
...
         if woken_by_message {
             billing_status.failures = 0;
             billing_status.next_retry = None;
+            quota_status.failures = 0;
+            quota_status.next_retry = None;
+        }
+
+        // Processing all of the different operations, and recording their name
+        // so we can emit better error messages.
+        let mut results = vec![];
```
jshearer: Both `reconcile` and `update_quotas` already turn operational failures into `Ok(WaitForRetry(backoff))` (storing `last_error` and logging via `tracing::warn!`). So `results`/`error_messages` collect a case that never fires, while real errors flow through the internal path. I'd fold the two outcomes with `?` on each call and drop the vec plus the 'better error messages' comment.

crates/agent/src/tenant_controller/outcome.rs
```diff
@@ -0,0 +1,22 @@
...
+    WaitForRetry(std::time::Duration),
+}
+
+impl Outcome {
+    /// This function determines the next action given two outcomes.
+    /// The following rules apply when combining outcomes.
+    ///
+    /// WaitForRetry is always chosen, before Idle, and the WaitForRetry
+    /// with the smallest duration is used.
+    pub fn next_action(self, other: Outcome) -> Outcome {
```
jshearer: nit: this returns an `Outcome`, not an action, so `next_action` reads a little oddly. `merge` or `combine` would say what it does.

crates/agent/src/tenant_controller/quotas.rs
```diff
@@ -0,0 +1,330 @@
...
+        );
+    }
+    Ok(())
+}
+
+const PAID_TASKS_QUOTA_MIN: i32 = 100;
+const COLLECTIONS_QUOTAS_MIN: i32 = 10000;
+
+/// Update the quotas to the necessary amount of 100 tasks and 10000 collections.
+async fn update_tenant_quotas_by_name(
```
jshearer, line 118: `UPDATE ... WHERE tenant = $1` returns Ok when it matches no rows, and the caller then sets `updated_quotas = true` regardless (the `quota_update_no_tenant_in_db` test documents exactly this). Safe today because the controller only runs for a tenant that already exists, but it's a silent no-op-treated-as-success. A rows-affected check would harden it.

crates/agent/src/tenant_controller/mod.rs
```diff
@@ -28,21 +33,47 @@ impl TenantController {
     }
 }
 
-pub(crate) struct TenantRow {
+// NOTE(BB): This was copied from the crates/billing-integrations/src/publish.rs
```
jshearer: There's already a `billing-types` crate that was intentionally created as a home for these shared, well, billing types. Let's move this over there and reference it from both places?

crates/agent/src/tenant_controller/quotas.rs
```diff
@@ -0,0 +1,330 @@
...
+    pub updated_quotas: bool,
+    #[serde(default)]
+    pub failures: u32,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub next_retry: Option<chrono::DateTime<chrono::Utc>>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub last_error: Option<String>,
+}
+
+fn retry_backoff(failures: u32) -> std::time::Duration {
```
jshearer: `retry_backoff`, the `failures`/`next_retry`/`last_error` fields, and the check-next_retry / call-do / reset-or-backoff wrapper are duplicated verbatim from `billing_contact.rs`. Worth pulling it out to keep things DRY IMO

jshearer (changes requested):

jshearer:
Overall the design matches what we agreed: one-way `GREATEST` raise, wake-driven re-checks, no polling. The `quota_update_success_pre_configured` test proving a hand-set-higher value isn't lowered is a nice touch, and declining the AI's 'wake on payment-method removal' suggestion was the right call.

Most substantive items are inline, a couple nit-level items:

- Comment typos: 'Check for error sand return', 'Processing all error', 'None or stipe', 'both the has has_default_payment_configured', 'Verifing no change', and the warn log 'failed while updating tenant quotas failed' (doubled 'failed').
- `mod test` -> `mod tests` to match billing_contact.rs.
- `let _ = sqlx::query!(...)` binding is unnecessary; `?` already covers it.

crates/agent/src/tenant_controller/quotas.rs
bbartman: Fixed.

crates/agent/src/tenant_controller/mod.rs
bbartman: Fixed.

crates/agent/src/tenant_controller/outcome.rs
bbartman: Fixed.

crates/agent/src/tenant_controller/quotas.rs
bbartman: Fixed.

crates/agent/src/tenant_controller/mod.rs
bbartman: Fixed.

crates/agent/src/tenant_controller/quotas.rs
bbartman: Fixed.

crates/agent/src/tenant_controller/quotas.rs
```diff
@@ -0,0 +1,311 @@
...
+    tenant: &Tenant,
+    billing_provider: &Option<Arc<dyn BillingProvider>>,
+) -> anyhow::Result<()> {
+    if status.updated_quotas {
+        return Ok(());
+    }
+    // Check this first because its possible for us to error out while communicating with
+    // the billing provider, and it's possible for the customer to not exist within
+    // stripe as well.
+    if tenant.payment_provider == Some(PaymentProvider::External) {
```
jshearer, line 65: This branch only runs if something else happens to wake the controller — but `payment_provider` is only ever set by a manual DB update currently, and the one wake we added is the `setBillingPaymentMethod` mutation. So external tenants never reach here on their own.

In #2675 we talked about also bumping the quotas when `tenants.payment_provider` => `external`, and James confirmed this in slack. Can you add that trigger?

crates/agent/src/tenant_controller/quotas.rs
```diff
@@ -0,0 +1,311 @@
...
+async fn do_update_quotas(
+    status: &mut QuotaUpdateStatus,
+    pool: &sqlx::PgPool,
+    tenant: &Tenant,
+    billing_provider: &Option<Arc<dyn BillingProvider>>,
+) -> anyhow::Result<()> {
+    if status.updated_quotas {
+        return Ok(());
+    }
+    // Check this first because its possible for us to error out while communicating with
```
jshearer, line 63: nit: the `quota_update_success_set_in_db` test comment below refers to `has_default_payment_configured`, which isn't a real symbol anymore.

jshearer (changes requested):
One more relevant change to make around `external` handling and a nit leftover name in a comment, then we should be good! Looking great :)

crates/agent/src/tenant_controller/quotas.rs
bbartman: I added triggers, and tests for the trigger.

crates/agent/src/tenant_controller/quotas.rs
bbartman: Fixed removed a reference to an extra variable.

jshearer (approved):
Nice, thanks for addressing that stuff. LGTM!

PR #3129, opened 2026-07-07
control-plane-api: add unauthenticated publicDataPlanes GraphQL query

GregorShear:
**Description:**

The new-account flow needs to offer a data-plane selection before the user has signed up or holds any grants, so the control plane must answer unauthenticated requests for the list of public data planes ([Slack thread](https://estuaryworkspace.slack.com/archives/C03QBN83GQ4/p1783441390746979?thread_ts=1783025188.655459&cid=C03QBN83GQ4)).

This adds a `publicDataPlanes` GraphQL query that requires no bearer token:

```graphql
query {
  publicDataPlanes(first: 10) {
    pageInfo { hasNextPage endCursor }
    edges {
      node {
        name
        cloudProvider
        region
      }
    }
  }
}
```

**Workflow steps:**

No user-facing workflow changes yet; the UI account-creation flow can adopt this query when data-plane selection lands there.

None.

**Notes for reviewers:**

- GraphQL auth is per-resolver (`Envelope` admits tokenless requests; resolvers gate via `env.claims()`), so the new resolver simply never consults claims. The existing `dataPlanes` query is untouched and still requires authentication — the test asserts both behaviors.
- Results come from the in-memory authorization Snapshot, filtered to names parsing as public via `parse_data_plane_name` and sorted by name. No database access.
- Only `name`, `cloudProvider`, and `region` are exposed, per the Slack discussion. Everything else stays behind the authenticated `dataPlanes` query.
- `PublicDataPlane` declares its `name`/`cloudProvider`/`region` independently of `DataPlane`. We explored sharing them — via a `PublicDataPlaneInfo` interface and via a `#[graphql(flatten)]` embed, both visible in the branch history — and decided plain duplication of three stable fields is simpler than either mechanism. An interface can be reintroduced compatibly if a client ever wants a cross-type fragment.
- Local data planes (`ops/dp/local/...`) parse as public and therefore appear, which makes the local dev stack's plane selectable in the account flow.
- Results are a Relay connection cursored on name, anticipating a public plane per provider/region pair someday. The cursor logic that was inline in the `dataPlanes` resolver is extracted into a `paginate_by_name` helper shared by both resolvers. That logic infers `hasNextPage` from a full page (`rows.len() == limit`, no +1 lookahead), so a page that ends exactly at the last row reports `hasNextPage: true` and the next fetch comes back empty — pre-existing `dataPlanes` behavior, now shared.
- Regenerated `crates/flow-client/control-plane-api.graphql`. Tested via `cargo test -p control-plane-api data_planes` (all pass; the usual parallel-migration flake aside).

### DIFF
--- a/crates/flow-client/control-plane-api.graphql
+++ b/crates/flow-client/control-plane-api.graphql
@@ -1465,6 +1465,50 @@ input ProtocolFilter {
 }
 
 """
+A public data plane, as visible to unauthenticated callers.
+"""
+type PublicDataPlane {
+	"""
+	Name of this data-plane under the catalog namespace.
+	"""
+	name: String!
+	"""
+	Cloud provider where this data-plane is hosted.
+	"""
+	cloudProvider: DataPlaneCloudProvider!
+	"""
+	Cloud region where this data-plane is hosted.
+	For example: "us-east-1" (AWS), "us-central1" (GCP), "eastus" (Azure).
+	"""
+	region: String!
+}
+
+type PublicDataPlaneConnection {
+	"""
+	Information to aid in pagination.
+	"""
+	pageInfo: PageInfo!
+	"""
+	A list of edges.
+	"""
+	edges: [PublicDataPlaneEdge!]!
+}
+
+"""
+An edge in a connection.
+"""
+type PublicDataPlaneEdge {
+	"""
+	The item at the end of the edge
+	"""
+	node: PublicDataPlane!
+	"""
+	A cursor for use in pagination
+	"""
+	cursor: String!
+}
+
+"""
 Summary of a publication that was attempted by a controller.
 """
 type PublicationInfo {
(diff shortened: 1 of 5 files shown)


crates/flow-client/control-plane-api.graphql
```diff
@@ -1464,6 +1464,71 @@ input ProtocolFilter {
...
+	name: String!
+	"""
+	Cloud provider where this data-plane is hosted.
+	"""
+	cloudProvider: DataPlaneCloudProvider!
+	"""
+	Cloud region where this data-plane is hosted.
+	"""
+	region: String!
+}
```
GregorShear: @jshearer is this any better? This at least guarantees name/cloudProvider/region stay in sync across both types, and lets clients share one fragment for them

jshearer (approved):
LGTM 👍

PR #3130, opened 2026-07-07
Restart Materializations sessions before IAM token expiry

dgreer-dev:
IAM tokens are minted once when a connector session starts and are never refreshed. Materializations hold a single long-lived session, so tasks using. GCP or Azure IAM auth (~1 hour tokens) fail once the token expires mid-session.  This PR makes materializations do a graceful shutdown just prior to their token expiration and also ensures we're using the longest token possible given the cloud and cloud settings.

### Cloud Refresher
Amazon: 12 hour tokens, always
Azure: 1 hour token default.  If set to allow longer tokens Azure will automatically deliver the longer tokens.
Google: 1 hour token, unless settings allow longer.  Note that even if the settings allow we still have to *request* the longer time.

### Changes

- **iam-auth**: all three providers now report the minted token's `expires_at`.
  - **GCP**: request a 12-hour token first, falling back to 1 hour when the customer's org policy doesn't allow lifetime extension. Enabling `iam.allowServiceAccountCredentialLifetimeExtension` is picked up automatically on the next session.
- **runtime-next**: when tokens are injected, the materialize shard computes a restart deadline (5 minutes before expiry for short-lived tokens, 30 minutes for 12-hour tokens) and asks the leader for a graceful stop when it elapses.  The session drains at a transaction boundary and its restart mints fresh tokens. Tasks without IAM auth are unchanged.
- **docs**: the GCP and Azure IAM guides gained "Token Lifetime" sections covering the customer-side policies above.

Because every transaction now starts with most of the token's lifetime ahead of it, a transaction is only at risk if it runs longer than the lifetime itself.  In those cases the remedy there is the cloud-side policy for 12-hour tokens.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4428,6 +4428,7 @@ dependencies = [
  "serde",
  "serde_json",
  "tokio",
+ "tracing",
  "zeroize",
 ]
 
--- a/crates/iam-auth/Cargo.toml
+++ b/crates/iam-auth/Cargo.toml
@@ -21,6 +21,7 @@ reqwest = { workspace = true }
 serde = { workspace = true }
 serde_json = { workspace = true }
 tokio = { workspace = true }
+tracing = { workspace = true }
 zeroize = { workspace = true, features = ["derive"] }
 
 [dev-dependencies]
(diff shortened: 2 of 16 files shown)


dgreer-dev:
@williamhbaker pointed out it'd probably be nice to have captures working the same way, so that has been added

mdibaiee (approved):
The parts that I had worked on make sense to me 👍🏽 thank you!

williamhbaker (approved):
LGTM

PR #3132, opened 2026-07-07
feat: end-to-end truncation

### LINKED ISSUE #2821: End-to-end Captured -> Materialized Truncations
Briefly, this feature work will enable captures (and perhaps derivations as well) to indicate that a captured resource is being truncated for ex. as part of a full-refresh backfill, and for the indication to propagate through to downstream readers (materializations) so that they can truncate resources produced from the captured data.

williamhbaker:
**Description:**

When a capture re-backfills a binding it re-snapshots the source, so the previously-captured documents in the collection become stale. This adds an end-to-end truncation boundary so downstream consumers skip or reset that superseded data instead of carrying it forward.

A capture connector signals `BackfillBegin` / `BackfillComplete` per binding. The capture publishes each as a `CONTROL` document into the collection's journals and stamps an `estuary.dev/truncated-at` label (the begin clock) on them. Every reader (V1 and V2 alike) treats `CONTROL` docs as metadata rather than content, and a fresh reader skips the stale pre-boundary prefix via the label. Shuffle carries the begin/complete clocks through the checkpoint frontier, and the runtime persists them so the boundary survives transaction rotation and restart. Materializations use the boundary to drop stale source documents, reset stale stored rows, and notify the connector so it can delete superseded destination rows.

Scoped to single-shard, full-key-range captures; multi-shard synchronized backfills are future work.

Tested on a local stack through various scenarios with a single shard capture and materialization processing sequences of `BackfillBegin` / `BackfillComplete` messages.

Closes https://github.com/estuary/flow/issues/2821

**Workflow steps:**

**Notes for reviewers:**

There's a constant stream of merge conflicting things landing on master. I've rebased and fixed these more times than I care to count at this point. I'll do one final pass just before this is ready to merge (and the requisite re-approvals).


go/protocols/shuffle/shuffle.proto
```diff
@@ -111,6 +111,32 @@ message Frontier {
   repeated JournalFrontier journals = 1;
   // Per-shard flushed LSN, indexed by shard_index.
   repeated uint64 flushed_lsn = 2;
+
+  // BackfillBegin is a binding's latest backfill-begin clock, keyed by the
+  // binding's stable journal_read_suffix.
+  message BackfillBegin {
+    // Binding's journal read suffix, stable across task versions.
+    string journal_read_suffix = 1;
```
jgraettinger: I've been seeking thus far to push read suffixes down to the very edge, and prefer binding indices everywhere. This is why `decode_recover_key_value` takes `binding_state_keys`, for example, doing translation from suffix => binding index as it scans, and we do the reverse translation in `encode_frontier`. See also `JournalFrontier.binding`.

Is there a particular reason this needs to be a read suffix, or could it be a binding index instead?

crates/runtime-next/src/shard/materialize/actor.rs
```diff
@@ -21,9 +21,24 @@ pub(super) enum Phase {
...
+    // Cumulative backfill-begin clock per binding.
+    backfill_begin: Vec<u64>,
+    // Truncation boundary (begin clock) of the latest completed backfill per
+    // binding.
+    backfill_complete: Vec<u64>,
+    // Backfill-begin clock the connector has been notified of per binding.
+    notified_backfill_begin: Vec<u64>,
+    // Backfill-complete clock the connector has been notified of per binding.
+    notified_backfill_complete: Vec<u64>,
+    // True if this is shard zero, the only shard that emits backfill
```
jgraettinger: While this is _probably_ true, I don't think it's the runtime's job to make a determination about whether a connector instance wants a notification delivered or not: rather, we deliver the signal and the connector self-selects whether it will act on it (conditioned on its key range).

Similarly, we deliver Acknowledge to _every_ shard, even if only shard zero will act on it.

crates/doc/src/combine/memtable.rs
```diff
@@ -234,6 +235,25 @@ impl MemTable {
...
     pub fn add<'s>(&'s self, binding: u16, root: HeapNode<'s>, front: bool) -> Result<(), Error> {
+        self.add_with_flags(binding, root, front, false, false)
+    }
+
+    /// Add a document from a prior generation to the MemTable.
+    /// A prior-generation document is not reduced with current-generation
+    /// documents and is never drained: it transfers front() onto a following
+    /// current-generation document of its same key, and is otherwise discarded.
+    pub fn add_prior_gen<'s>(&'s self, binding: u16, root: HeapNode<'s>) -> Result<(), Error> {
+        self.add_with_flags(binding, root, true, true, true)
```
jgraettinger: Reading how this is called from runtime-next's materialize actor, we actually _don't_ know that it's `is_valid`, right?

crates/runtime-next/src/shard/capture/actor.rs
```diff
@@ -32,6 +32,7 @@ pub(super) struct Actor {
     metrics: super::Metrics,
     // When Some, a deadline at which we begin a graceful session stop.
     token_restart_at: Option<tokio::time::Instant>,
+    is_control_producer: bool,
```
jgraettinger: nit: comment on what this is? It wasn't immediately obvious to me.

(I now understand it's a guard w.r.t. whether the shard is allowed to produce backfill control messages)

jgraettinger (commented):
Great! We talked over VC about re-structuring the journal carrier (folding into ACKs with causal hints). A few other comments below.

go/protocols/shuffle/shuffle.proto
williamhbaker: Binding index works here! Originally I was thinking about using read suffixes instead of state keys so that things were stable regardless of backfill counter increments on the reader side, but that was not a valid line of reasoning, so I think this was just a holdover from that that I didn't go back and clean up. Updated.

crates/runtime-next/src/shard/materialize/actor.rs
williamhbaker: That makes sense; updated to send the messages to all shards.

crates/doc/src/combine/memtable.rs
williamhbaker: That is correct. My thought process here that it would save us from validating a doc we know we'd drop anyway, but upon inspection even with `known_valid = false` we'll still avoid the extra validation anyway and it's more correct, so I made that change.

crates/runtime-next/src/shard/capture/actor.rs
williamhbaker: I changed the name to `is_single_shard` which I think is more comprehensible, and also added a comment.

williamhbaker:
Addressed the comments so far, and also rebased & pushed a couple other substantial changes:

1. Having the backfill signals as part of ACKs, rather than a separate message. Overall a net simplification, and a better model I think. That got me thinking a lot about causal hints and the reader side, which led to...
2. A fix for the pre-backfill document filtering logic. I'd had a scan filter based on the `backfillBegin` threaded through on the Frontier, but that was incomplete since it didn't account for the prior frontiers that may have already been received and scanned unfiltered. The only tractable way I could come up with for getting this fully correct was to double down on the filtering in the combiner, and actually add clock metadata for every document, and a filter when draining. This is a pretty significant change, and there are performance implications, but it seems like an unavoidable cost.
3. Less of a code change, and more of a behavior change: Materializations will be required to produce a `_meta/uuid` field with their `Loaded` responses. As a compatibility mechanism for pre-existing "no flow document" cases that might not have this field, this is only enforced when the transaction contains an active `backfillBegin`-based filter. Practically I think this will shake out as "no flow document" materializations needing to require some representation of the timestamp in their field selection.

Ran a fresh claude review and manual E2E tests on these pushed changes as well.

williamhbaker:
Ran the combiner benchmarks before and after this change.

- For larger docs, like the github/citi benchmark, there is no measurable change.
- The worst case is tiny docs with small keys and zero reduction, which when synthesized and run in the combiner perf test shows a 7-8% overhead from the extra 8 bytes added to `Meta`.

crates/shuffle/src/slice/state.rs
```diff
@@ -370,15 +378,32 @@ pub fn sequence_document(
...
 
-    // A `notBefore` or `notAfter` suppresses document append, but doesn't impact
-    // the propagation of flush and progress reporting.
-    let is_append = is_append && *clock >= binding.not_before && *clock < binding.not_after;
+    // A `notBefore`/`notAfter` window and the journal's `truncated_at`
+    // truncation boundary all suppress document append, but don't impact the
+    // propagation of flush and progress reporting.
+    let is_append = is_append
+        && *clock >= binding.not_before
+        && *clock >= read_state.truncated_at
```
jgraettinger: confirming my understanding: `binding.not_before` is (only) the built spec's not-before, and does not updated with the journals' truncation label (carried only in read_state.truncated_at). Thus we must check both.

crates/shuffle/README.md
```diff
@@ -336,6 +336,16 @@ The same "did `unresolved` make progress?" signal disarms the `on_tick`
 stall timeout: it fires only when no progress at all occurs between
 two consecutive ticks.
 
+A peek also carries `latest_backfill_begin` eagerly (cloned from
```
jgraettinger, line 346: nice, makes sense

crates/publisher/src/publisher.rs
```diff
@@ -359,6 +393,66 @@ impl Publisher {
...
+            // Re-list and retry until a full pass applies without losing a CAS
+            // race. Partitions already at the target are skipped on the retry
+            // and the loop converges once any racing writer (a concurrent split
+            // or activation) settles. Bounded at 10 attempts so a persistently-
+            // racing writer (or a bug that keeps losing the CAS) fails the
+            // transaction rather than spinning forever and silently wedging the
+            // caller.
+            for cas_retries in 0..10 {
+                let listing = retry_transient("list partitions", || {
+                    client.list(broker::ListRequest {
```
jgraettinger: It'd be more idiomatic to use a list_watch or list_watch_with here, which can be polled again to retry, already handles back-off internally, and will return a restatement of the journals after every Ok change. You'd use its items to drive convergence retries.

crates/publisher/Cargo.toml
```diff
@@ -24,6 +24,7 @@ tokens = { path = "../tokens" }
 anyhow = { workspace = true }
 base64 = { workspace = true }
 bytes = { workspace = true }
+chrono = { workspace = true }
```
jgraettinger: nit: I'm not actually seeing `chrono` in use in `publisher` changes ?

crates/runtime-next/src/shard/capture/actor.rs
```diff
@@ -52,15 +57,43 @@ pub(super) struct Actor<P: crate::Publisher, L: crate::Logger> {
     acknowledge_fut: Option<BoxFuture<'static, anyhow::Result<()>>>,
     drain_fut: Option<BoxFuture<'static, anyhow::Result<drain::Output<P>>>>,
     intents_write_fut: Option<BoxFuture<'static, tonic::Result<P>>>,
-    persist_fut: Option<BoxFuture<'static, anyhow::Result<(crate::shard::RocksDB, Vec<String>)>>>,
+    labels_apply_fut: Option<BoxFuture<'static, (P, BTreeMap<u32, u64>, tonic::Result<()>)>>,
+    persist_fut: Option<
+        BoxFuture<
```
jgraettinger: nit: factor out a type alias here and below?

crates/runtime-next/src/shard/capture/actor.rs
```diff
@@ -515,6 +588,29 @@ impl<P: crate::Publisher, L: crate::Logger> Actor<P, L> {
                 true
             }
 
+            fsm::Action::ApplyTruncatedLabels => {
+                // Only a single-shard capture manages truncated-at labels. A
+                // sub-range shard that inherited `active_backfills` (e.g. a split
+                // mid-backfill) must not re-apply labels it can never clear — it
+                // never receives the BackfillComplete that removes them.
+                if !self.is_single_shard || !self.labels_dirty || self.active_backfills.is_empty() {
```
jgraettinger: I'm wondering if we should make the rule that shard zero, alone, manages truncation signals. that would allow a CDC capture to scale out by having shard zero manage the WAL, while other shards perform backfill fetches.

crates/runtime-next/src/shard/materialize/actor.rs
```diff
@@ -492,16 +544,38 @@ impl Actor {
...
-                .ok_or_else(|| anyhow::anyhow!("Loaded binding {binding_index} out of range"))?;
+            let binding_spec = &self.bindings[binding_index];
 
             let (memtable, _alloc, doc) =
                 accumulator.parse_json_doc(&doc_json).with_context(|| {
                     format!("parsing loaded doc for {}", binding_spec.collection_name)
                 })?;
-            memtable.add(binding_index as u16, doc, true)?;
+
+            // Classify by the loaded document's UUID clock. As a fall-back for
```
jgraettinger: Should we make this fail faster? A concern is that materialization tasks will appear to work fine, and then blow up on a first source truncation, right?

jgraettinger (commented):
A few minor comments below (plus our Slack thread re: the combiner, which is my only real concern). Everything else makes sense!

williamhbaker:
Pushed a commit that includes a different strategy for the combiner changes, instead of adding document clock to the meta. It adds a `truncate` method which handles memtable documents directly, and fences off already spilled segments. It's a bit more machinery but avoids the very real performance hit of the extra 8 bytes to the meta. PTAL.

crates/shuffle/src/slice/state.rs
williamhbaker: Yep that's right. They are separate, with the `not_before` being a property of the binding and the `truncated_at` being a proper of the journal.

crates/runtime-next/src/shard/materialize/actor.rs
williamhbaker: Yeah, we probably should, I was thinking not right away though. With it like this, pre-existing "no flow document" materializations that don't load a `_meta/uuid` will keep working until captures start to emit truncations. I think the implementation on the connector side will need to do at least materializations first, with the "no flow document" cases always emitting a clock, and then we can harden this.

crates/publisher/src/publisher.rs
williamhbaker: Switched to using `list_watch`, much better.

crates/publisher/Cargo.toml
williamhbaker: Left over from some prior iterations...removed.

crates/runtime-next/src/shard/capture/actor.rs
williamhbaker: Done

crates/runtime-next/src/shard/capture/actor.rs
williamhbaker: Seems reasonable, I could see that working out. Updated.

williamhbaker:
Also pushed an additional commit to incorporate the review feedback. I'll aim to do one final rebase prior to merging that will amend the two `fixup!` commits into the cohesive history.

crates/doc/src/combine/memtable.rs
```diff
@@ -276,6 +294,30 @@ impl MemTable {
...
+        // Safety: mutable borrow does not escape this function.
+        let entries = unsafe { &mut *self.entries.get() };
+
+        let retain = |entry: &mut HeapEntry| -> bool {
+            if entry.meta.binding() != binding as usize {
+                return true;
+            } else if !entry.meta.front() {
+                return false;
+            }
+            entry.meta.set_stale();
```
jgraettinger, line 311: Nice, I hadn't fully considered the implications of a connector that streams Loaded responses on demand, which you caught here and have handled 👍 

I'm expecting this means we need to still surface front() documents of cut-off segments for purposes of the Exists flag. (confirmed)

jgraettinger (dismissed):
Changes look great! All of my comments are resolved. Ready to stamp after you rebase.

jgraettinger:
re: the agents.md change, fyi I added cargo fmt to the build:rust-protobufs task a bit ago

jgraettinger (approved):

PR #3136, opened 2026-07-08
runtime: opt-in relaxation of inferred date-time formats on read-side validation

### LINKED ISSUE #3133: runtime: materialize/derive read-side schema validation retroactively rejects historical documents after date-time format strictness fix (#3116)
## 1. Summary

PR #3116 (fixing #3108) correctly tightened `Format::DateTime::validate()` in `crates/json/src/schema/formats.rs` to require a literal `T`/`t` date-time separator, matching RFC3339's ABNF. This was the right fix for schema *inference* (stop mistagging mixed-format fields going forward) and for *capture-time write-schema* enforcement (reject new non-conforming values at the door).

However, it has an unintended and more severe side effect: **materializations (and likely derivations) that read pre-existing documents out of a collection now hard-fail validation on historical data that was tagged `format: date-time` by continuous inference *before* #3116 landed**, even though the actual bytes of those documents were never RFC3339-conformant to begin with (e.g. a space-separated timestamp like `2026-06-17 12:46:17.375663+00:00` instead of `2026-06-17T12:46:17.375663+00:00`).

This is a regression discovered in production after #3116 rolled out. It re-breaks tasks that the connector-side liberal fix (#4725, landed in #4743, which made `materialize-sql`'s date-time converters tolerate near-RFC3339 variants) was specifically supposed to keep working. The liberal connector-side fix never gets a chance to run, because the runtime now rejects the document before it reaches the connector at all.

Customer/task names are intentionally omitted from this issue; ask in the originating Slack thread (#integrations, July 2026 "materialize-bigquery date-time" thread) if reproduction against real production data is needed.

## 2. Why this matters / priority

This directly contradicts explicit product guidance from a 2026-07 daily sync: *"materialization connectors need to be liberal in this case, so that existing collection data can be handled"* combined with *"Estuary should strictly reject non-conforming alleged-RFC3339 timestamps"* — i.e., the intended design was strict rejection at the write/ingest boundary, liberal handling of already-admitted historical data. #3116 accidentally made the *read* boundary strict too, which defeats the liberal connector fix for any collection that has legacy non-conforming data under a field that inference had already tagged `format: date-time`.

There is **no available workaround**. Excluding the offending field from a materialization binding's field selection does *not* help — see section 4 for why.

## 3. Reproduction / observed symptom

A materialization shard that was previously healthy (after the connector-side fix in #4743 landed) starts failing again with an error of this shape:

```
runTransactions: txnStartCommit: app.FinalizeTxn: materialized collection <tenant>/<collection> (binding N) document failed validation against its collection JSON Schema

Caused by:
    {
      "basic_output": [
        {
          "absoluteKeywordLocation": "flow://inferred-schema#/properties/attributes/properties/properties/properties/$consent_timestamp",
          "detail": "Format mismatch: expected a DateTime",
          "instanceLocation": "/attributes/properties/$consent_timestamp",
          "instanceValue": "2026-06-17 12:46:17.375663+00:00"
        }
      ],
      "document": { "_meta": { "op": ... } ...
    }
```

Note `absoluteKeywordLocation` is scoped under `flow://inferred-schema#...` — the constraint originates from Flow's own continuously-inferred schema, not from anything the source connector or user explicitly authored in a write schema.

This was observed on a task whose materialization binding had the offending field (`$consent_timestamp`-style, a Klaviyo-sourced consent-timestamp property under a free-form `attributes/properties` object) **explicitly excluded** from the binding's field selection as a prior workaround (see #4725/#4726 for the original incident) — confirming the validation happens on the *whole document* before any field projection, so field exclusion offers no protection.

The binding in the observed case had a nonzero backfill count, meaning the task had recently reprocessed (backfilled) its full historical range — this is presumably how/why the runtime re-read and re-validated old, pre-fix documents. Any binding backfill (a routine, common operation — triggered by key changes, schema changes, disable/re-enable cycles, etc.) is a landmine for any collection that ever had non-conforming `format: date-time` data inferred before #3116.

### 4.1 Where the failing validator comes from

Materialization tasks build their document-combine validator from the collection's **read schema**, not the write schema:

```rust
// crates/runtime/src/materialize/task.rs:197-203
pub fn combiner_spec(
    &self,
) -> anyhow::Result<(bool, Vec<doc::Extractor>, String, doc::Validator)> {
    let built_schema = doc::validation::build_bundle(&self.read_schema_json)
        .context("collection read_schema_json is not a JSON schema")?;
    let validator =
        doc::Validator::new(built_schema).context("could not build a schema validator")?;
    ...
```

Compare with capture ingestion, which validates against the narrower **write schema**:

```rust
// crates/runtime/src/capture/task.rs:168-172
let built_schema = doc::validation::build_bundle(&write_schema_json)
    .context("collection write_schema_json is not a JSON schema")?;
let validator =
    doc::Validator::new(built_schema).context("could not build a schema validator")?;
```

The read schema embeds the continuously-inferred schema as a `$defs` entry keyed by the URI `flow://inferred-schema` (see `crates/validation/src/collection.rs:126-161`, and the observed error's `absoluteKeywordLocation` prefix). This is baked in at **publish/build time** in the control plane — it is not re-resolved dynamically at runtime; a collection needs a (possibly automatic "touch") publish to pick up newly-inferred shapes.

### 4.2 Why the fix can't be "just filter out this outcome after validation"

`crates/json/src/validator/validation.rs`'s `invalidate()` sets the document-invalid flag **unconditionally**, before any Rust-level outcome filter runs:

```rust
// crates/json/src/validator/validation.rs:1004-1022
fn invalidate<'s, A, F>(
    filter: &F,
    frame: &mut Frame<'s, A>,
    invalid: Outcome<'s, A>,
    tape_index: i32,
) where
    A: Annotation,
    F: Fn(Outcome<'s, A>) -> Option<Outcome<'s, A>>,
{
    frame.flags |= FLAG_INVALID;

    if let Some(outcome) = filter(invalid) {
        frame.outcomes.push(ScopedOutcome {
            outcome,
            schema_curi: schema::get_curi(frame.keywords),
            tape_index,
        });
    }
}
```

`FLAG_INVALID` is set regardless of what the `filter` closure decides to keep. The filter (e.g. `crates/doc/src/validation.rs`'s `error_filter`) only controls what's *collected for reporting* — it cannot rescue an already-invalid document. So any fix must prevent the offending `Keyword::Format` from being compiled as an *enforcing* keyword in the first place for this case; post-hoc filtering by `schema_curi` (e.g. "ignore FormatNotMatched when it comes from `flow://inferred-schema`") is architecturally impossible with the current validator design.

### 4.3 Where the inferred schema's `format` keyword gets serialized

```rust
// crates/doc/src/shape/schema.rs (to_sub_schema, ~line 160-165)
if let Some(f) = format {
    out.insert("format".to_string(), serde_json::json!(f.to_string()));
}
```

This unconditionally emits `format` as a plain JSON Schema keyword whenever a `Shape`'s `string.format` is set — with no distinction between "this was authored/declared" and "this was merely detected by best-effort continuous inference." Once compiled by `crates/json/src/schema/build.rs` (`FORMAT => keywords.push(Keyword::Format { format })`, unconditional, no vocabulary/mode gating), it becomes a hard-enforced keyword in *any* validator built from a schema bundle that includes it — including the read-side combine validator above.

### 4.4 What's *not* affected (verified, to save the next investigator time)

Materialize connectors' DDL/column-type-mapping decisions (e.g. mapping a `format: date-time` field to a `TIMESTAMP` column) are **not** driven by the serialized inferred-schema JSON. They come from a separate code path that reads the in-memory `Shape` struct directly:

```rust
// crates/assemble/src/lib.rs:8
pub fn inference(shape: &Shape, exists: Exists) -> flow::Inference { ... }
```
called from `crates/validation/src/collection.rs:603,616,651,678`, populating the `Projection.Inference` proto sent to connectors. This is structurally independent of `crates/doc/src/shape/schema.rs`'s `to_schema()` / the `read_schema_json` blob. **Any fix that only changes how the inferred schema is serialized/compiled for read-side document validation will not affect column typing.** Confirmed by tracing both call chains; no shared code between them.

## 5. Existing precedent to extend

Flow already has a mechanism for exactly this class of problem, just not applied to the piece that's actually causing this bug:

```rust
// crates/models/src/schemas.rs:96-102
/// Transform this Schema into a relaxed form which removes all `type`,
/// `require`, and `format` keywords of most (but not all) recursive
/// sub-schemas, while preserving other keywords.
///
/// The primary purpose of relaxed schemas is to transform a collection
/// write-schema into a relaxation which is likely to function well when
/// intersected with the collection's inferred schema.
pub fn to_relaxed_schema(&self) -> serde_json::Result<Self> { ... }
```

This is used **today only for the write-schema side** of the read-schema intersection:

```rust
// crates/validation/src/collection.rs:376-393 (approximate)
if read_model.references_relaxed_write_schema() {
    let has_inferred_schema = ...;
    relaxed = has_inferred_schema
        .then(|| write_model.to_relaxed_schema())
        ...
    defs.push(models::schemas::AddDef {
        id: models::Schema::REF_RELAXED_WRITE_SCHEMA_URL, // flow://relaxed-write-schema
        schema: relaxed.as_ref().unwrap_or(&write_model),
        overwrite: true,
    });
}
```

i.e., Flow's existing design principle is *"don't let the write schema's strict keywords retroactively reject data once inference has weighed in."* There is currently no symmetric `flow://relaxed-inferred-schema` — nothing relaxes the *inferred* schema's own `format`/`type`/`require` keywords, which is exactly where our failing keyword lives (see 4.1's `absoluteKeywordLocation`).

## 6. Proposed fix direction (not yet implemented — this issue is for scoping/implementation)

Introduce a `flow://relaxed-inferred-schema` analogous to the existing `flow://relaxed-write-schema`: apply `to_relaxed_schema()` (or a purpose-built variant) to the inferred schema before/when it's inlined into `$defs` in `crates/validation/src/collection.rs`, and have the read-schema template reference the relaxed form for whatever purpose currently causes hard validation failures on read, while:
- Leaving schema *inference/detection* logic (`crates/doc/src/shape/widen.rs`, `Format::detect()`) untouched and strict — it operates on in-memory `Shape`, not the serialized schema, and stopping mistagging of mixed-format fields going forward is correct and already working via #3116.
- Leaving *capture-time write-schema* validation untouched and strict (`crates/runtime/src/capture/task.rs:168-172`) — new incoming data should still be rejected at the door per product guidance.
- Not affecting DDL/column-type mapping (see 4.4 — already verified independent).

## 7. Open questions for whoever implements this (please resolve/verify, don't assume)

1. **Does `crates/runtime/src/derive` have the identical exposure?** Derivations also read documents from a source collection and likely build a combine validator from a read schema the same way materializations do. This has not yet been checked. If so, it needs the same fix in the same PR.
2. **Exact wiring for `flow://relaxed-inferred-schema`:** does this require new plumbing in `crates/models` (e.g. a `references_relaxed_inferred_schema()` counterpart to the existing `references_relaxed_write_schema()`), or can the control plane substitute it transparently in `crates/validation/src/collection.rs`'s read-schema assembly regardless of what a collection's authored `readSchema` explicitly references? Need to trace how/whether `flow://inferred-schema` is referenced in the *default*, system-generated read schema template (for collections without a custom-authored read schema) versus a user-authored one — the fix may need to handle both cases.
3. **Scope of relaxation:** `to_relaxed_schema()` strips `type`, `required`, *and* `format` together. Is that the right granularity here, or should only `format` be relaxed for the inferred schema (to avoid loosening other useful inferred constraints, e.g. numeric bounds, that aren't subject to the same "upstream dependency silently changed its opinion" failure mode as `format`)? Consider whether a narrower, format-only relaxation function is warranted versus reusing `to_relaxed_schema()` wholesale.
4. **Backward compatibility / rollout:** does this require existing collections to be republished (a "touch" publish) to pick up the new schema composition, the same way inferred-schema updates already require one? If so, is that automatic or does it need to be triggered?
5. **Testing:** add coverage in `crates/validation` for the new schema-assembly logic (unit test asserting the composed read schema tolerates a legacy non-conforming value under a field the inferred schema tags `format: date-time`), and ideally a `crates/runtime` or integration-level test that reproduces the exact failure mode in section 3 (a document with a non-conforming `format: date-time` value, previously accepted, now read back for materialization) and confirms it no longer hard-fails post-fix while a genuinely malformed *new* document at capture time still does.
6. **Should this generalize beyond `date-time`?** The triggering event here was specifically the `time` crate's RFC3339 parser drifting lenient-then-strict (see #3108's history), but the same retroactive-rejection risk exists for *any* format whose validator's definition of "valid" could change over time (e.g. `email`, `uri`, `ipv4`) relative to already-inferred, already-persisted data. Consider whether the relaxed-inferred-schema treatment should cover all formats uniformly (simplest, matches `to_relaxed_schema()`'s existing all-or-nothing behavior) rather than being date-time-specific.

## 8. References

- #3108 — original issue: `DateTime` format validator silently drifted lenient via a dependency bump
- #3116 — the fix for #3108 that introduced this regression (merged 2026-07-03)
- estuary/connectors#4725 — companion connector-side (liberal materialize-sql converters) issue
- estuary/connectors#4743 — the PR implementing #4725, whose intended protection is defeated by this issue
- estuary/connectors#4726, estuary/connectors#4742 — source-side (Klaviyo) normalization, unaffected by this issue but related context

jacobmarble:
**Description:**

Fixes #3133.

#3116 correctly tightened the RFC3339 `date-time` format validator, but read-side (materialize/derive) document validation builds its validators from the collection read schema, which inlines the continuously-inferred schema — so historical documents whose date-time values were admitted before the tightening (e.g. space-separated `2026-06-17 12:46:17+00:00` under a field inference had tagged `format: date-time`) are now retroactively rejected on read. Field exclusion doesn't help (validation is whole-document, pre-projection), and any binding backfill re-reads old documents, so previously-healthy tasks re-break.

This adds an **opt-in, per-task shard feature flag** `estuary.dev/flag/relax-inferred-datetime-formats`. When set to `"true"` on a task's `shards.flags`, the task's read-side validators are built from a read schema in which `date`/`date-time`/`time` `format` keywords are stripped from the inlined `flow://inferred-schema` `$defs` entry — and from nothing else. All four read-side validator-build sites are covered:

| Runtime | Site |
|---|---|
| V1 materialize | `Binding::new` (`crates/runtime/src/materialize/task.rs`) |
| V1 derive | `Transform::new` (`crates/runtime/src/derive/task.rs`) |
| V2 materialize | `build_binding` (`crates/runtime-next/src/shard/materialize/task.rs`) |
| V2 derive | `build_transform` (`crates/runtime-next/src/shard/derive/task.rs`) |

The relaxation logic is single-sourced in `models::Schema::relax_inferred_datetime_formats()` — a new, additive function (`to_relaxed_schema()` is untouched). The flag reader is single-sourced in `labels::shard_flag_enabled()`.

Deliberately narrow, per the scoping in #3133:
- **`format` only** — inferred `type`/`required`/bounds remain enforced (they're structural guarantees connectors rely on).
- **`date`/`date-time`/`time` only** — the three formats whose validator delegates to the `time` crate, the dependency whose RFC3339 interpretation drifted (time-rs/time#700). Other formats stay strict.
- **Inferred schema only** — author-declared formats anywhere in the read schema, and the write-schema `$def`, remain strictly enforced.
- **Read-side only** — schema inference/detection and capture-time write-schema enforcement are unchanged and stay strict.
- **Off by default** — zero behavior change for tasks without the flag.

Because the flag is read at the runtime task-build layer from the task's own `shard_template` labels, rollout is per-task and requires no collection republish: enable the flag on an affected task's spec and it takes effect on the next shard assignment.

**Workflow steps:**

For a task failing with `Format mismatch: expected a DateTime` scoped under `flow://inferred-schema` (see #3133 section 3), add to the task's model:

```yaml
shards:
  flags:
    relax-inferred-datetime-formats: "true"
```

and publish. The flag is emitted as the `estuary.dev/flag/relax-inferred-datetime-formats` shard label and read by the runtime when building the task's read-side validators. Remove the flag to restore strict enforcement.

None yet — this is an ops-facing escape hatch, not (yet) documented user-facing behavior. If we later decide to document `shards.flags` values, this one should be included.

**Notes for reviewers:**

- Reviewed commit-by-commit is easiest: V1 materialize → V1 derive → shared `labels::shard_flag_enabled` refactor → V2 (runtime-next).
- **V2 has a two-stage validation path**; the flag lands on the authoritative stage. The shuffle read pipeline's fast-path validator (`crates/shuffle/src/slice/read.rs`) is intentionally left strict — it only sets `FLAGS_SCHEMA_VALID`, it never faults. Documents without that flag are re-validated downstream against the (relaxed) binding/transform validator: derive in `shard/derive/scan.rs`, materialize in the combiner's handling of `!known_valid` docs (`doc/src/combine/memtable.rs`). Consequence: on flag-enabled tasks, legacy non-conforming docs always take the re-validation path — correct, with a small re-validation cost limited to those documents.
- `relax_datetime_formats` descends only through schema-bearing keywords, so `date-time` strings appearing as *data* (`enum`/`const`/`default`, or a property literally named `format`) are never touched — covered by tests. Its traversal list was cross-checked against everything `doc::shape::schema::to_schema()` can emit into an inferred schema.
- Tests at every site assert both flag states: OFF preserves today's strict behavior exactly; ON tolerates the #3133 reproduction value `2026-06-17 12:46:17.375663+00:00` while RFC3339-conformant values continue to validate.
- Per #3133, this touches a core validation path and should get review from owners of `crates/validation` / `crates/runtime` schema semantics.

### DIFF
--- a/crates/labels/src/lib.rs
+++ b/crates/labels/src/lib.rs
@@ -13,6 +13,11 @@ pub const FLAG_PREFIX: &str = "estuary.dev/flag/";
 // Flag which enables the V2 task runtime.
 // TODO(whb): remove once the runtime-v2 migration is complete.
 pub const RUNTIME_V2_FLAG: &str = "estuary.dev/flag/enable-runtime-v2";
+// Flag which relaxes read-side enforcement of `date`, `date-time`, and `time`
+// `format` keywords originating from a collection's inferred schema. It lets a
+// task tolerate historical documents whose date-time values predate a
+// tightening of the RFC3339 `format` validator. Opt-in, per-task. See #3133.
+pub const RELAX_INFERRED_DATETIME_FLAG: &str = "estuary.dev/flag/relax-inferred-datetime-formats";
 pub const KEY_BEGIN: &str = "estuary.dev/key-begin";
 pub const KEY_BEGIN_MIN: &str = "00000000";
 pub const KEY_END: &str = "estuary.dev/key-end";
--- a/crates/labels/src/lib.rs
+++ b/crates/labels/src/lib.rs
@@ -16,7 +16,7 @@ pub const RUNTIME_V2_FLAG: &str = "estuary.dev/flag/enable-runtime-v2";
 // Flag which relaxes read-side enforcement of `date`, `date-time`, and `time`
 // `format` keywords originating from a collection's inferred schema. It lets a
 // task tolerate historical documents whose date-time values predate a
-// tightening of the RFC3339 `format` validator. Opt-in, per-task. See #3133.
+// tightening of the RFC3339 `format` validator. Opt-in, per-task.
 pub const RELAX_INFERRED_DATETIME_FLAG: &str = "estuary.dev/flag/relax-inferred-datetime-formats";
 pub const KEY_BEGIN: &str = "estuary.dev/key-begin";
 pub const KEY_BEGIN_MIN: &str = "00000000";
--- a/crates/labels/src/lib.rs
+++ b/crates/labels/src/lib.rs
@@ -249,6 +249,22 @@ pub fn maybe_one<'s>(set: &'s LabelSet, name: &str) -> Result<&'s str, Error> {
     }
 }
 
+/// Return whether shard feature-flag `flag` (a fully-qualified flag label such
+/// as [`RELAX_INFERRED_DATETIME_FLAG`]) is set to "true" within a task's
+/// `shard_template` labels. Task runtimes call this to gate per-task behavior
+/// at the point they build read-side validators. A missing shard template,
+/// missing labels, or any non-"true" value all read as disabled.
+pub fn shard_flag_enabled(
+    shard_template: Option<&proto_gazette::consumer::ShardSpec>,
+    flag: &str,
+) -> bool {
+    shard_template
+        .and_then(|shard| shard.labels.as_ref())
+        .and_then(|set| maybe_one(set, flag).ok())
+        .map(|value| value == "true")
+        .unwrap_or(false)
+}
+
 /// Returns whether `set` is matched by `selector`, faithfully porting gazette's
 /// `LabelSelector.Matches` (`go.gazette.dev/core/broker/protocol`):
 /// an excluded label matching any of `set` is a non-match; otherwise every
@@ -386,6 +402,33 @@ impl<'a> LabelJoin<'a> {
 #[cfg(test)]
 mod test {
     use crate::*;
+
+    #[test]
+    fn shard_flag_enabled_cases() {
+        let flag = RELAX_INFERRED_DATETIME_FLAG;
+
+        let shard_with = |value: &str| proto_gazette::consumer::ShardSpec {
+            labels: Some(build_set([(flag, value)])),
+            ..Default::default()
+        };
+
+        assert!(shard_flag_enabled(Some(&shard_with("true")), flag));
+
+        // Any value other than "true" reads as disabled.
+        assert!(!shard_flag_enabled(Some(&shard_with("false")), flag));
+
+        // Flag absent, no labels, or no shard template at all.
+        let no_flag = proto_gazette::consumer::ShardSpec {
+            labels: Some(build_set([(COLLECTION, "acmeCo/foo")])),
+            ..Default::default()
+        };
+        assert!(!shard_flag_enabled(Some(&no_flag), flag));
+        assert!(!shard_flag_enabled(
+            Some(&proto_gazette::consumer::ShardSpec::default()),
+            flag
+        ));
+        assert!(!shard_flag_enabled(None, flag));
+    }
 
     #[test]
     fn label_range_cases() {
--- a/crates/models/src/schemas.rs
+++ b/crates/models/src/schemas.rs
@@ -118,7 +118,7 @@ impl Schema {
     /// (materialize / derive) document validation does not retroactively reject
     /// historical documents whose date-time values predate a tightening of the
     /// `format` validator, while leaving capture-time write-schema enforcement
-    /// and schema inference strict. See estuary/flow#3133.
+    /// and schema inference strict.
     pub fn relax_inferred_datetime_formats(&self) -> serde_json::Result<Self> {
         let mut bundle = serde_json::from_str::<serde_json::Value>(self.get())?;
 
@@ -191,7 +191,7 @@ pub struct AddDef<'a> {
 /// `format` values whose read-side enforcement is relaxed by
 /// [`Schema::relax_inferred_datetime_formats`]. These are exactly the formats
 /// whose validator delegates to the `time` crate, whose interpretation of
-/// RFC3339 has drifted over time (see estuary/flow#3133, #3108, #3116).
+/// RFC3339 has drifted over time.
 const RELAXED_DATETIME_FORMATS: [&str; 3] = ["date", "date-time", "time"];
 
 /// Recursively remove `date`/`date-time`/`time` `format` keywords from `node`
--- a/crates/models/src/schemas.rs
+++ b/crates/models/src/schemas.rs
@@ -104,6 +104,33 @@ impl Schema {
         Ok(Self(serde_json::value::to_raw_value(&relaxed)?.into()))
     }
 
+    /// Return a copy of this read-schema bundle in which `date`, `date-time`,
+    /// and `time` `format` keywords are removed from the inlined inferred
+    /// schema — the `$defs` entry keyed by [`Self::REF_INFERRED_SCHEMA_URL`].
+    /// Every other keyword, every other `$defs` entry, and the remainder of the
+    /// bundle are left exactly as-is. If the bundle does not inline an inferred
+    /// schema (e.g. a single-schema collection), this is a no-op copy.
+    ///
+    /// This is deliberately much narrower than [`Self::to_relaxed_schema`]: it
+    /// touches only `format`, only for the three formats whose validator
+    /// delegates to the `time` crate (whose RFC3339 interpretation has drifted
+    /// over time), and only within the inferred schema. It exists so read-side
+    /// (materialize / derive) document validation does not retroactively reject
+    /// historical documents whose date-time values predate a tightening of the
+    /// `format` validator, while leaving capture-time write-schema enforcement
+    /// and schema inference strict. See estuary/flow#3133.
+    pub fn relax_inferred_datetime_formats(&self) -> serde_json::Result<Self> {
+        let mut bundle = serde_json::from_str::<serde_json::Value>(self.get())?;
+
+        if let Some(inferred) = bundle
+            .get_mut("$defs")
+            .and_then(|defs| defs.get_mut(Self::REF_INFERRED_SCHEMA_URL))
+        {
+            relax_datetime_formats(inferred);
+        }
+        Ok(Self(RawValue::from_value(&bundle)))
+    }
+
     /// Extend this Schema with added $defs definitions.
     /// The $id keyword of definition is set to its id, and its id is also
     /// used to key the property of $defs under which the sub-schema lives.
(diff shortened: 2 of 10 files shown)


jacobmarble:
Closing per team decision — handling currently-known cases via support (dataflow reset or direct inferred-schema edit, see estuary/flow#3133) rather than a platform code change. Leaving #3133 open as a record of the underlying gap in case it needs to be revisited.

PR #3137, opened 2026-07-08
local: per-checkout stacks — every checkout runs its own isolated stack

jgraettinger:
## Why

Until now the local platform stack was effectively a singleton: one database, one
set of ports, one build directory, one flowctl profile per machine. That's fine
when you're the only thing running, but it makes parallel work painful. Spin up a
second checkout to QA a fix while a soak test runs in the first, and the two
stacks fight over the same ports and the same Postgres. In practice you had to
tear one down before bringing the other up — and with multiple people (or agents)
sharing a dev VM, you were constantly stepping on each other.

This change makes a local stack belong to a *checkout* rather than to the machine.
Every clone and every git worktree gets its own fully isolated platform — its own
database, etcd, brokers, reactors, ports, build artifacts, and flowctl profile —
so any number of them can run side by side without coordination.

## What it feels like to use

The whole point is that you don't think about isolation. You work in a checkout,
and everything scopes to it automatically:

```bash
mise run local:stack        # bring up this checkout's stack
mise run local:stack-info   # ports, units, and ready-to-paste commands for *this* stack
```

`stack-info` is the one thing to remember: it prints the ports and commands for
wherever you happen to be, so you never compute a port or pick a name. Under the
hood each checkout claims a small integer index and derives everything from it,
but as a developer you mostly just see that things don't collide.

A few niceties fall out of this:

- **`flowctl` just works.** Inside a checkout, bare `flowctl` targets that
  checkout's stack — no `--profile` flag to remember, and no chance of
  accidentally driving the wrong stack.
- **Common variables are ambient.** `FLOW_PG_URL`, `FLOW_CLUSTER`, the
  `FLOW_PORT_*` set, and friends are already in your environment when you run
  through mise, so `psql "$FLOW_PG_URL"` and the like do the right thing.
- **Builds are shared per stack, not globally.** Each checkout has its own target
  directory, so rebuilds in one worktree don't invalidate another's cache.
- **Teardown is scoped.** `local:stop` tears down just the stack you're in and
  leaves everyone else running.

## Lifecycle: stop is cheap, deleting the worktree is the release

Stopping and releasing are deliberately different actions:

- **`local:stop`** is a stack-scoped guillotine — it stops this stack's units and
  wipes its runtime state, but **keeps its registry slot, its build cache, and
  its Supabase volumes**, so a later start is instant.
- **Releasing a slot happens when you delete the checkout.** A deleted checkout
  directory is the *one and only* signal that frees a slot and its artifacts.
  Just remove the worktree when you're done:

  ```bash
  git worktree remove ../feature-x   # (or plain rm -rf) — that's the release
  ```

A garbage collector reclaims deleted checkouts in the background — stopping their
orphaned units, then freeing the slot, build target, state dir, and Supabase
volumes. It **runs automatically at natural joints** (`local:stack` and
`local:stop`), so starting or stopping a stack anywhere cleans up worktrees you've
removed. You can also run it on demand:

```bash
mise run local:stack-prune   # reclaim slots of deleted checkouts (idempotent)
```

This also means a registry full of removed checkouts heals itself instead of
dead-ending on "no free index": the new-checkout allocation path reaps gone
entries before it looks for a free slot.

## A note on mise

Because stack identity now comes from the checkout, mise is the mandatory entry
point — it's what makes the per-checkout variables ambient. Running tasks and
tests through mise (`mise run ...`, `mise exec -- ...`) always has the right
environment. Outside mise, wrap once with `eval "$(mise run local:stack-env)"`.
Scripts and tests fail loudly rather than silently defaulting if the environment
is missing, so you find out immediately instead of debugging a mystery stack.

In the same spirit, `dekaf --local` now fails loud when no endpoint is given:
local-stack ports are dynamic so there is no compiled local default, and clap
would otherwise fall through to the *production* defaults — pointing a local anon
token at prod. `mise run local:data-plane` supplies the right endpoints; run it
that way, or pass `API_ENDPOINT`/`AGENT_ENDPOINT` explicitly.

`local/README.md` is the full reference for how this is wired together.

## Keeping the disk in check on dev VMs

Per-checkout isolation has a cost: every checkout gets its own
`~/cargo-target/<name>/`, each 20–60G, and they overlap heavily — the same crates
compiled in worktree after worktree. On a shared dev VM with several checkouts
that adds up fast.

So on dev VMs, `~/cargo-target` is now backed by a **compressed,
self-deduplicating btrfs filesystem** on a sparse loopback image sized to 50% of
the disk:

- **zstd:3 transparent compression** (~3.2× measured on real artifacts).
- **A 10-minute `duperemove` + `fstrim` timer** that collapses extents shared
  across checkouts and punches the freed space back into the sparse image —
  near-total dedupe for same-commit worktrees, and ~26% of logical bytes even
  across divergent branches (extent-level dedupe shares blocks of differing files
  too). Net: many divergent checkouts fit in a fraction of their raw footprint.
- The Go build cache moves onto the filesystem, since Go stores it uncompressed
  and it compresses well. sccache stays at its default `~/.cache/sccache`: it
  already compresses its own entries, so it would gain nothing under the
  `compress-force` mount (and can't be excluded from it).

It's set up by `bootstrap:cargo-target-fs`, invoked from `vm:create-post` before
anything builds. The task is idempotent and Linux-only, and it never migrates an
existing plain-dir `~/cargo-target` (VMs are disposable, make a new one).

## QA

Exercised end to end: full bring-up, tenant provisioning, a soak test with all
verification checks passing, isolation spot-checks across concurrent stacks, and
scoped teardown.

The slot-reclaim lifecycle was driven directly against a live stack:

- **Reaper unit coverage** — its guards (empty/short name), prefix-safety
  (`alpha` doesn't touch `alpha-2`), exact-suffix volume matching (`flow` doesn't
  reap `flow-2`), and registry rewrite (including the all-gone empty case) all
  hold.
- **Reclaim, end to end** — allocated a throwaway worktree a slot, planted its
  build target, state dir, and a Supabase volume, deleted the worktree, and
  confirmed `local:stack-prune` reclaimed exactly that entry (slot, target, state,
  volume) while a concurrently-running stack was left completely untouched. The
  freed index was then immediately reused by a new checkout.
- **Auto-reap at joints** — confirmed `local:stack` and `local:stop` reclaim
  deleted checkouts, and that `local:stop` retains the current stack's slot, build
  cache, and Supabase volumes for an instant restart.
- **dekaf `--local` guard** — verified it bails with an actionable message when
  `API_ENDPOINT`/`AGENT_ENDPOINT` are missing, and starts cleanly on the
  `local:data-plane` happy path.


crates/dekaf/src/main.rs
```diff
@@ -250,11 +257,11 @@ async fn async_main(cli: Cli) -> anyhow::Result<()> {
...
-    let (api_endpoint, api_key) = if cli.local {
-        (LOCAL_PG_URL.to_owned(), LOCAL_PG_PUBLIC_TOKEN.to_string())
-    } else {
-        (cli.api_endpoint, cli.api_key)
-    };
+    // Endpoints are resolved by clap from flags/env (with `--local` selecting the
+    // local anon api_key default). Local-stack ports are dynamic, so a local
+    // deployment must pass API_ENDPOINT/AGENT_ENDPOINT explicitly — the
+    // local:data-plane task does this in the generated dekaf env file.
+    let (api_endpoint, api_key) = (cli.api_endpoint, cli.api_key);
```
skord, line 288: `dekaf --local` with no endpoint now silently targets production.

Dropping the `default_value_if("local", ...)` from `api_endpoint` (and the `if cli.local` override that used to live here) means `--local` still forces the local anon `api_key`, but `api_endpoint` falls through to `DEFAULT_PG_URL`, which is production. So a hand-run `dekaf --local` points a local anon token at the prod PostgREST endpoint, where on master it used `LOCAL_PG_URL`.

The managed path is fine: `local:data-plane` writes `API_ENDPOINT`/`AGENT_ENDPOINT` into the generated dekaf env file, so `mise run local:stack --dekaf` always has them. The gap is a bare `dekaf --local` (or any invocation outside that env file), which fails silently against prod instead of failing loud. That is at odds with the fail-loud stance the rest of this PR takes.

Suggested guard: if `--local` is set but the endpoints are still the prod defaults (never right for a local stack), bail with something actionable.

```rust
// `--local` selects the local anon api_key, but local-stack ports are dynamic
// so there is no compiled local endpoint default. Guard the surprising case of
// `--local` with no endpoint: clap would otherwise fall back to the *production*
// defaults, pointing a local token at prod. local:data-plane supplies these in
// dekaf's generated env file; a hand-run must pass them too.
if cli.local {
    if cli.api_endpoint == *DEFAULT_PG_URL {
        anyhow::bail!(
            "--local requires an explicit API_ENDPOINT (or --api-endpoint): \
             local-stack ports are dynamic, so there is no local default. Run \
             dekaf via `mise run local:data-plane`, or set API_ENDPOINT to your \
             stack's PostgREST URL (see `mise run local:stack-info`)."
        );
    }
    if cli.agent_endpoint == *DEFAULT_AGENT_URL {
        anyhow::bail!(
            "--local requires an explicit AGENT_ENDPOINT (or --agent-endpoint): \
             local-stack ports are dynamic, so there is no local default. Run \
             dekaf via `mise run local:data-plane`, or set AGENT_ENDPOINT to your \
             stack's agent URL (see `mise run local:stack-info`)."
        );
    }
}
```

Comparing against the prod default `Url` is reliable here, since a real local stack's endpoint is never the prod URL, and passing the prod URL alongside `--local` is nonsensical anyway. (`DEFAULT_PG_URL`/`DEFAULT_AGENT_URL` are already imported.) The tidier but larger alternative is making both `Option<Url>` and resolving explicitly, but for a one-file dev-tooling guard the equality check is the smaller diff.

Non-blocking: the managed flow is unaffected; this closes a footgun for hand-run invocations.

skord (commented):
A couple of non-blocking observations on the per-checkout local tooling. Both are about failure modes rather than the happy path, which works well. Details inline.

mise/tasks/local/stack-env
```diff
@@ -0,0 +1,223 @@
...
+
+_se_index=""
+_se_name=""
+_se_new=""
+_se_alloc="$(_se_lookup_or_allocate)"
+_se_alloc_rc=$?
+case "${_se_alloc_rc}" in
+    0) IFS=$'\t' read -r _se_index _se_name _se_new <<<"${_se_alloc}" ;;
+    2) _se_die "cannot lock ${_se_lock} — check permissions" ;;
+    3) _se_die "no free stack index (0..15 all taken); run 'mise run local:stack-release' in an unused checkout" ;;
```
skord: Stale registry entries pin their index forever, with no way to reclaim once the checkout is gone.

If a worktree is removed with a plain `rm -rf` (rather than `local:stack-release`), its line stays in `stacks.tsv` and its index stays claimed. After enough churn you hit this `no free stack index (0..15)` error, but the remedy it suggests (running `stack-release` in that checkout) is impossible because the checkout no longer exists. There is currently no escape hatch short of hand-editing `stacks.tsv`.

The reason this is not simply "reap entries whose root is missing during allocation" is a collision hazard: deleting a worktree's directory does not stop its `systemctl --user` units, which keep holding their ports. A naive auto-reap could free an index and hand it to a new checkout while the old stack's services are still live, which is exactly the port collision the sticky-registry design avoids. Any reap has to gate on root gone AND units idle.

Suggested: a dedicated `local:stack-prune` task that reaps dead-root slots (and their artifacts, guarded like `stack-release`) while keeping any entry whose services are still active. And point this error message at it:

```bash
3) _se_die "no free stack index (0..15 all taken); free one with 'mise run local:stack-release' in an unused checkout, or 'mise run local:stack-prune' to reap deleted checkouts" ;;
```

<details><summary>mise/tasks/local/stack-prune (sketch)</summary>

```bash
#!/usr/bin/env bash
set -euo pipefail
#MISE description="Reap registry slots of checkouts that were deleted without stack-release"

FLOW_LOCAL="${HOME}/flow-local"
REGISTRY="${FLOW_LOCAL}/stacks.tsv"
LOCKFILE="${FLOW_LOCAL}/stacks.lock"
[ -f "${REGISTRY}" ] || { echo "No registry present; nothing to prune."; exit 0; }

(
    flock 9
    tmp="$(mktemp "${FLOW_LOCAL}/stacks.tsv.XXXXXX")"
    while IFS=$'\t' read -r idx name root; do
        [ -n "${root}" ] || continue
        if [ -d "${root}" ]; then                      # live checkout: keep
            printf '%s\t%s\t%s\n' "${idx}" "${name}" "${root}" >> "${tmp}"; continue
        fi
        cluster="local-${name}-cluster"
        if systemctl --user list-units --no-legend --state=active \
              "flow-*@${name}.service" "flow-*@${name}.target" \
              "flow-*@${cluster}.service" "flow-*@${cluster}.target" \
              "flow-*@${cluster}-*.service" "flow-*@${cluster}-*.target" 2>/dev/null \
              | grep -q .; then
            echo "prune: KEEPING '${name}' (idx ${idx}): ${root} is gone but its units still run." >&2
            echo "prune:   stop first: systemctl --user stop 'flow-*@${name}.*' 'flow-*@${cluster}*'" >&2
            printf '%s\t%s\t%s\n' "${idx}" "${name}" "${root}" >> "${tmp}"; continue
        fi
        target="${HOME}/cargo-target/${name}"
        [ -n "${name}" ] && [ "${target}" != "${HOME}/cargo-target" ] && [ -d "${target}" ] && rm -rf "${target}"
        statedir="${FLOW_LOCAL}/${name}"
        [ -n "${name}" ] && [ "${statedir}" != "${FLOW_LOCAL}" ] && [ -d "${statedir}" ] && rm -rf "${statedir}"
        if command -v docker >/dev/null 2>&1; then
            while IFS= read -r vol; do
                [ "${vol##*_}" = "${name}" ] || continue
                docker volume rm "${vol}" >/dev/null 2>&1 || true
            done < <(docker volume ls --format '{{.Name}}' 2>/dev/null | grep '^supabase_' || true)
        fi
        echo "prune: reaped '${name}' (idx ${idx}); checkout ${root} is gone."
    done < "${REGISTRY}"
    mv "${tmp}" "${REGISTRY}"
) 9>"${LOCKFILE}"
```
</details>

Non-blocking: the workflow works today; this is about the failure mode after checkouts accumulate and get deleted casually.

crates/dekaf/src/main.rs
jgraettinger: Added this guard

mise/tasks/local/stack-env
jgraettinger: I re-worked stack release to be lazy. Deleting a worktree is the functional release mechanism, and then an ambient mise task (which can also be run explicitly) detects the deletion and removes everything else.

jgraettinger:
PTAL

skord (dismissed):
LGTM

skord (approved):
LGTM

PR #3138, opened 2026-07-09
docs: document connecting Azure Private Link to native Azure resources

jwhartley:
## Summary
- The Azure Private Link page only documented building a customer-managed Private Link Service + Load Balancer. In practice, several data planes link directly to native Azure PaaS resources (SQL Managed Instance, Storage Accounts, SQL servers) via Private Link's sub-resource type ("group ID"), which was undocumented and could lead customers to build unnecessary Load Balancer infrastructure for resources that support Private Link natively.
- Also documents the optional DNS name field: if a customer provides one, Estuary provisions a private DNS zone and points it at the endpoint automatically, instead of handing back a bare IP.

## Test plan
- [ ] Docs-only change, no code paths affected
- [ ] Preview the rendered page locally (`npm run start` in `site/`) and confirm the new "Connecting to a native Azure resource" subsection renders correctly under Azure Private Link

### DIFF
--- a/site/docs/private-byoc/privatelink.md
+++ b/site/docs/private-byoc/privatelink.md
@@ -134,35 +134,31 @@ You can find this URL as you would any other Atlas hostname:
 
 ## Azure Private Link
 
-For Azure private or BYOC deployments, we can establish connections to your endpoints using Azure Private Link or PaaS resource.
+For Azure private or BYOC deployments, we can establish connections to your endpoints using Azure Private Link.
 
 You will need to follow different steps depending on your resource type:
 
 * [The resource requires an explicit Azure Private Link Service](#connecting-with-azure-private-link-service)
 * [The resource already supports Private Link directly](#connecting-to-a-native-azure-resource)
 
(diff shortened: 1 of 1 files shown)


jwhartley:
Note for whoever picks this up next: we're hoping to ship self-service private link setup this quarter (Q3 2026). Once that lands, this page (and the customer-facing PrivateLink instructions in general) should be revisited since the manual request flow described here may no longer be the primary path.

aeluce (dismissed):
The structure of the section may make it confusing on what's required for one method or another, since there's no indication when the PaaS resource section ends and the two threads come together again.

I pushed a possible solution that reorders the information. If that doesn't seem quite right, another option would be to use tabs.

Though as you say, self-service is coming soon and that sounds great.

jwhartley:
Hey @aeluce I tweaked it slightly to make it a little more succinct and readable if you wouldn't mind re-reviewing.
FYI this question (Azure storage private linking) came up today again for second customer this week - it's come up maybe twice in all history before this week but we're seeing some more POCs on Azure

aeluce (approved):

PR #3140, opened 2026-07-09
runtime-next: remove hardcoded materialization ser_policy fallback

dgreer-dev:
Remove the last vestiges of the ser_policy defaults.  All but three materializations already have the correct behavior, and [a PR](https://github.com/estuary/connectors/pull/4787) has been raised.

Marking this as a draft because we shouldn't merge until that PR has been deployed and all tasks using those connectors have been republished.


williamhbaker (approved):
LGTM

PR #3146, opened 2026-07-10
Updating migrations to better support testing.

bbartman:
**Description:**

The migrations were causing test failure due to conflicting writes to the database. The migrations were talking directly to the DB to change permissions and configurations, this was causing conflicting write problems between tests using the `sqlx::test` functionality which re-applies migrations to the DB every time. The solution here is to guard against those migration changes by preventing them from running without the appropriate database name of `postgres`. Sqlx's test framework doesn't use the `postgres` name specifically, it appends a hash to the end of the test database name. 

**Workflow steps:**

N/A

N/A

**Notes for reviewers:**

This work was reviewed, and understood by me, but I don't fully understand our entire DB structure yet. Claude seems to think that we are not enforcing checks on migrations when we doing migrations, only file names I have been unable to verify if this is the case. If it is the case then this will work fine and only improve testing, if it's not we need to reject this MR and find another solution.

### DIFF
--- a/supabase/migrations/00_polyfill.sql
+++ b/supabase/migrations/00_polyfill.sql
@@ -72,7 +72,18 @@ END
 $$;
 
 -- Required for postgres to give ownership of catalog_stats to stats_loader.
-grant stats_loader to postgres;
+-- Guarded to the primary `postgres` database: role membership is cluster-global
+-- (shared pg_auth_members). Re-applying it concurrently across the isolated
+-- `#[sqlx::test]` databases risks colliding on the same shared tuple. The
+-- membership is established once by the `postgres` run and inherited cluster-wide
+-- by the test databases, so they don't need to re-grant it.
+do $$
+begin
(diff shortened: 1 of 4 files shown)


skord (approved):
The concern only bites if our migration tracking checksums file contents (editing an applied file would then error). Prod doesn't run a migrator at all though: migrations are hand-pasted into the Supabase SQL console, nothing tracks filenames or checksums, and old ones never get re-pasted. So editing an applied migration is a no-op for prod, and new pastes run against the `postgres` DB where the `current_database() = 'postgres'` guards evaluate true anyway. Safe to merge.

Why the guards are needed: role grants, `alter role`, and `alter database` write to Postgres cluster-global catalogs shared across every database, so parallel tests collide on the same tuple. Gating to `postgres` runs them once and the test DBs inherit the roles. Same pattern `00_polyfill.sql` already uses for `create role`.

PR #3147, opened 2026-07-10
validation: surface write-schema redact annotations on projections of split-schema collections

GregorShear:
Fixes redaction bug reported [in slack](https://estuaryworkspace.slack.com/archives/C02G25WJCNS/p1783627117013839).

A `redact` annotation reachable only through the write schema is in force — redaction runs at write time — but projections of collections with a standalone `readSchema` reported `redact: UNSET` for it, both from full build validation and from `skim_projections` (the dashboard path via flow-web's `skim_collection_projections`).

`walk_collection_projections` now fills in `inference.redact` from the write schema for any projection whose read-schema inference has no strategy of its own; the read schema wins when both are annotated. Extracts `assemble::inference_redact` for reuse.

A redact strategy on a key location is still checked against the read schema only: connector-managed write schemas may annotate a key which `flow://relaxed-write-schema` then strips from the read path (6e58bdd4d94).

Tests: snapshot tests over `skim_projections` (single schema, ref-style read schema, standalone read schema, error cases) and a WASM acceptance test of the flow-web entry point.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3346,7 +3346,7 @@ dependencies = [
 
 [[package]]
 name = "flow-web"
-version = "0.6.10"
+version = "0.6.11"
 dependencies = [
  "console_error_panic_hook",
  "doc",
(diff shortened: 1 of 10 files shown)


jgraettinger (approved):
LGTM

A bit more new test SLOC than the change warrants IMO 🤷. As a general tip try asking for table driven cases.

PR #3154, opened 2026-07-13
flowctl: multi-shard --fixture for raw preview-next

mdibaiee:
Lifts the `--shards 1` restriction on `flowctl raw preview-next --fixture`, so fixture previews exercise real multi-shard topologies.

**What changed:**

- The fixture writer opens one shuffle-log writer per shard within the shared session directory (a shard's stream is distinguished by its `mem-{shard:03}-…` segment file names), and hash-routes each document to its shard using the same packed-key hash (`doc::Extractor::packed_hash`) and range routing (`shuffle::slice::routing`, now exported) as the live slice.
- Synthesized checkpoint Frontiers carry a per-shard `flushed_lsn` vector; a shard receiving no documents in a transaction carries its prior session-local barrier forward.
- Document clocks now mirror the legacy preview fixture harness: the ordinal-th transaction's documents are stamped `3600 * ordinal + line` seconds, ticked once per fixture line (shared by every binding the line feeds, and consumed even by lines whose collection isn't sourced). A fixture therefore yields identical document UUIDs and publication timestamps for every task it drives, across both the legacy and runtime-v2 preview stacks.
- `--output-state` / `--output-apply` continue to require `--shards 1`.

**Why:** groundwork for the connectors repository's runtime-v2 integration tests, which drive scale-out materialization connectors through preview-next with `--shards 2` against the same fixtures — and the same expected table contents — as their legacy-preview integration tests.

## Test plan

- [x] `cargo test -p flowctl` and `cargo test -p shuffle`
- [x] E2E from the connectors repo: `materialize-postgres` fixture preview at `--shards 1` reproduces its legacy-preview integration snapshot byte-for-byte (including `--output-apply` / `--output-state` lines; the only addition is the second session's no-op apply). `materialize-databricks` at `--shards 2` (with its scale-out feature flag) reproduces identical table contents and schemas, with documents genuinely split across both shards and cross-session recovery applying the final staged transaction.

### DIFF
--- a/crates/flowctl/src/raw/preview_next/fixture.rs
+++ b/crates/flowctl/src/raw/preview_next/fixture.rs
@@ -199,6 +199,7 @@ pub fn build(
     path: &std::path::Path,
     base_dir: &std::path::Path,
     requested_targets: &[u32],
+    n_shards: u32,
 ) -> anyhow::Result<FixturePlan> {
     let (bindings, mut validators, collection_bindings) = task_bindings(task)?;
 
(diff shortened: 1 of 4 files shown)


jgraettinger (approved):
LGTM

PR #3156, opened 2026-07-13
runtime: fix recursive read-lock deadlock in capture-v2 buildJoin

jgraettinger:
## Summary

Fixes a recursive read-lock deadlock in the V2 capture runtime, traced from a production goroutine dump (`reactor005`) where tens of thousands of goroutines were parked on `KS.Mu` read-lock acquisition.

`captureAppV2.buildJoin` held `KS.Mu.RLock` and then called `shard.Spec()` several times. Gazette's `shard.Spec()` re-acquires `s.resolved.RLock`, which **aliases the same `KS.Mu`**. When the Etcd watcher's `KeySpace.Apply` writer queued between the outer and recursive read locks, `sync.RWMutex` — which blocks new readers once a writer is waiting — deadlocked:

- the recursive `shard.Spec()` `RLock` could not proceed past the queued writer, while
- the writer could not proceed until the outer `RLock` released, which only happens when `buildJoin` returns.

That stalled the keyspace lock for every subsequent reader (`ShardList`, `Resolver.Resolve`, metrics, shuffle reads, …), amplified by proxy goroutines waiting on `ShardList` RPCs.

## Fix

Read from the term's already-captured `c.term.shardSpec` and `c.term.labels` instead of calling `shard.Spec()` under the lock. This mirrors `buildLeaderfulJoin` (already used by the derive and materialize V2 paths) and is also more correct: the term is a coherent topology snapshot rather than a possibly-newer spec read mid-session. The now-unused `shard` parameter and `go/labels` import are dropped.

The derive and materialize V2 paths do **not** have this bug — they already route through `buildLeaderfulJoin`, which uses the term-captured spec and never calls `shard.Spec()` inside its lock region.

## Test plan

- [x] `go build -tags libsqlite3 ./go/runtime/`

### DIFF
--- a/go/runtime/capture_v2.go
+++ b/go/runtime/capture_v2.go
@@ -6,7 +6,6 @@ import (
 
 	"github.com/estuary/flow/go/bindings"
 	"github.com/estuary/flow/go/flow"
-	"github.com/estuary/flow/go/labels"
 	pf "github.com/estuary/flow/go/protocols/flow"
 	pr "github.com/estuary/flow/go/protocols/runtime"
 	"github.com/estuary/flow/go/shuffle"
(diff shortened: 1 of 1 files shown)


williamhbaker (approved):
LGTM

jgraettinger:
Sanity checked using the soak test on a local stack

Issue #3102, opened 2026-07-02
billing: store Stripe customer and invoice IDs instead of matching on metadata


jshearer:
We never store the ID of the Stripe objects we create. Every code path that needs "the Stripe customer for this tenant" or "the Stripe invoice for this bill row" re-derives it at runtime by searching Stripe metadata:

- **Customers** are found by searching `metadata["estuary.dev/tenant_name"]` (`billing_types::customer_search_query`).
- **Invoices** are found by searching the compound key `(customer, estuary.dev/invoice_type, estuary.dev/period_start, estuary.dev/period_end)` (`billing_types::InvoiceMetadata` / `InvoiceSearch`). The doc comment on `InvoiceMetadata` already flags this as debt: neither `internal.manual_bills` nor `internal.billing_historicals` models a stable primary key, so identity is reconstructed from mutable fields.

This identity model is fragile in three independent ways — it's **mutable** (bill dates get corrected, which silently changes the identity), **erasable** (deleting a Stripe-side object deletes the only record it ever existed), and **eventually consistent** (Stripe's Search API lags writes by seconds and is heavily rate-limited).

Consequences:

* `publish-invoices` re-selects a multi-month `manual_bills` row on every monthly run in its date range, and relies on the metadata search to find the invoice it created previously. When the match fails (dates edited, invoice created by hand in Stripe, stray draft deleted) it creates a duplicate; when it succeeds against a finalized invoice it errors instead.
* The GraphQL `StripeInvoiceLoader` issues one Search call per invoice and hits 429s. #3054 works around this and notes storing the invoice ID as the real fix.
* Customer lookups race Stripe's search-index lag. This is why `Customer::create` needs a deterministic idempotency key, and why the live-Stripe integration test needs `wait_for_customer_searchable` and still flakes in CI.

## Fix

Add nullable ID columns and retrieve by ID everywhere. Metadata search remains only as a one-time adoption path: on a hit, write the ID back; on a miss, create and write back.

* `tenants.stripe_customer_id`
* `internal.manual_bills.stripe_invoice_id`
* `internal.billing_historicals.stripe_invoice_id`, exposed through `invoices_ext`

In `publish-invoices`, with a stored ID: update the invoice if it's still a draft, skip it if it's open or paid, and warn without recreating if it's voided or deleted. A manual bill then produces exactly one invoice, regardless of how many months it spans or how its dates are edited.

PR #3157, opened 2026-07-13
control-plane-api: retry Stripe customer search on index-lag misses

jgraettinger:
## Problem

The **Stripe integration test** step of the Platform Test CI job flakes frequently. Every recent failure I pulled shares one identical error — the mutations abort with `no Stripe customer exists for tenant`:

| Run | Assertion |
|-----|-----------|
| [29248258820](https://github.com/estuary/flow/actions/runs/29248258820) | `card_a should be set as primary` |
| [29223924809](https://github.com/estuary/flow/actions/runs/29223924809) | `card_a should be set as primary` |
| [29126176839](https://github.com/estuary/flow/actions/runs/29126176839) | `card_b should become primary…` |
| [29113243430](https://github.com/estuary/flow/actions/runs/29113243430) | `card_a should be set as primary` |

## Root cause

`setBillingPaymentMethod` / `deleteBillingPaymentMethod` resolve the customer via `require_customer` → `find_customer` → Stripe's **customer search API**. That index is eventually consistent *and non-monotonic*: a customer that matched a moment earlier can transiently drop out of a later query. The test proves it — `wait_for_customer_searchable` succeeds (cards attach to the returned `customer.id`), yet the mutation's search seconds later misses. A single miss aborts the whole mutation.

This is also a real production fragility, not just a test artifact: a live user whose customer exists but is momentarily absent from search would hit the same spurious error.

## Fix

Override `require_customer` in `StripeBillingProvider` with bounded retry-on-miss + exponential backoff (6 attempts, capped at 3s, ~9.5s worst case). Every caller only reaches it *after* a SetupIntent created the customer, so a miss is virtually always index lag. The trait's single-shot default still applies to the in-memory mock, keeping unit tests instant and deterministic.

## On splitting the step into a parallel CI job

Considered and recommend against a fully separate job: the test runs in ~20s but needs the compiled test binary + RocksDB + musl/gnu toolchains + a running Supabase — a standalone job would repeat that multi-minute build for a 20s test. Inline it reuses everything Platform Test already built. The real pain was the flake failing the required job, which this fix addresses at the root. If further isolation is still wanted, the cheap options are moving the step to the end of the job (today it runs before doctest/gotest/catalog-test and cancels them on failure) and/or marking it `continue-on-error`.

## Testing

Verified via compile + code-path review. The live Stripe path can't be exercised locally without a testmode key; it'll get its real workout on CI.

### DIFF
--- a/crates/control-plane-api/src/billing/stripe_impl.rs
+++ b/crates/control-plane-api/src/billing/stripe_impl.rs
@@ -43,6 +43,36 @@ impl BillingProvider for StripeBillingProvider {
         .await
     }
 
+    /// Stripe's customer-search index is eventually consistent and can even be
+    /// non-monotonic: a customer that matched a moment ago may transiently drop
+    /// out of a later search. Every caller of `require_customer` (setting or
+    /// deleting a payment method) only reaches it after a SetupIntent has
+    /// already created the customer, so a search miss is virtually always index
+    /// lag rather than a true absence. Retry a bounded number of times with
+    /// exponential backoff before surfacing the error, overriding the trait's
+    /// single-shot default.
+    async fn require_customer(&self, tenant: &str) -> anyhow::Result<stripe::Customer> {
+        const MAX_ATTEMPTS: u32 = 6;
+        const MAX_DELAY: std::time::Duration = std::time::Duration::from_secs(3);
+
+        let mut delay = std::time::Duration::from_millis(500);
+        for attempt in 1..=MAX_ATTEMPTS {
+            if let Some(customer) = self.find_customer(tenant).await? {
+                return Ok(customer);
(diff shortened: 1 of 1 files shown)


jshearer (approved):
LGTM thanks... I've been meaning to get back around to this. I believe the real fix is https://github.com/estuary/flow/issues/3102 -- the fetch-by-id apis don't have the same eventual consistency problem that the search API has, and associating customers/invoices by compound keys composed of their metadata fields has already caused other problems (like when you change the date range of an invoice, for example). 

So IMO as part of a future billing cleanup saga, we should start storing stripe IDs directly. But I haven't gotten around to this yet, so the retry here should at least solve the flaky test.

Issue #3131, opened 2026-07-07
created at date in built specs / connector protocol


dgreer-dev:
some related discussion here: https://estuaryworkspace.slack.com/archives/C03Q2NRFKDL/p1781133294558569

that thread ends inconclusively, but later VC reached consensus that threading the creation date through the connector protocol is the simplest way to go

dgreer-dev:
## Proposal: connector compatibility dates (control-plane stamped)

A task carries a **compatibility date**, a UTC date meaning "resolve my unset
feature flags to their defaults as of this date." New tasks get current
defaults; existing tasks keep theirs; behavior is always determined by visible,
human-modifiable state.

Resolution precedence, each layer overriding the next:

```
explicit flag in endpoint config  >  model compatibilityDate  >  build-time stamped date
```

This preserves the tri-state we want for support: a flag a human explicitly
set, a date a human explicitly bumped, and a system-stamped default.

### Mechanism

1. **Model**: optional `compatibilityDate` field on task defs. Set by humans
   only.
2. **Build time**: validation resolves it (model value if set, else inherit
   the prior value from the live built spec, else stamp the current date) and
   emits it as a label on the task's shard template (e.g.
   `estuary.dev/compatibility-date`). Labels are data, not schema: **no
   flow.proto changes**, no codegen, no reader/writer version-ordering
   concerns; every component at every version round-trips unknown labels
   untouched.
3. **Delivery**: runtime-next reads the label from its shard labeling (parsing
   already exists — see `crates/runtime-next/src/leader/mod.rs` flag handling)
   and injects the date into the decrypted `config_json` handed to the
   connector.  v2-only; v1 ignores the label and needs no changes.
4. **Interpretation**: connectors that flip a default maintain a
   `flag -> (old default, new default, since-date)` table and resolve flags as a
   pure function of (explicit flags, compatibility date, table). Connectors
   that never flip a default need zero changes, ever.

Ownership split: the control plane owns the date's **existence** (stamp it,
never interpret it); the data plane owns its **meaning**. Introducing a new
flag is a connector-side table entry — nothing is pushed to the fleet.

### Existing tasks

Covered with existing patterns: controllers already touch-publish every
enabled task at least every 20 days (`crates/agent/src/controllers/periodic.rs`),
and a rebuild stamps any task without a date. The whole fleet converges within
one periodic cycle using publications that already happen. Roll out ordering
only requires that all tasks that may need a compat flag are stamped prior
to the new flag existing.

[...]

PR #3160, opened 2026-07-13
Provide capture/materializations created_at date

dgreer-dev:
Implements #3131: built capture and materialization specs carry the task's
creation date (`createdAt`, UTC YYYY-MM-DD), reaching connectors via Apply
and Open so they can resolve feature-flag defaults as-of task creation —
new tasks get current defaults, existing tasks keep theirs.

- Dates are derived from flowid
- Value can be empty, should just read that as the epoch or similar
- The existing fleet backfills itself: periodic touch republishes rebuild
  every enabled task within 20 days, stamping from its live Id.
- Discover and drafts are deliberately out of scope for this step. The
  eventual ideal is created_at as a user-visible model field (carried by
  pull-specs; delete + republish to re-mint) — this is forward-compatible
  with that.
- Derivations are deferred until fully on runtime V2.

### Deploy notes

- Breaking change with `flowctl` (new protobuf entries, issue is fixed
   in code but it might still be working its way towards deployment)
- Connectors shouldn't declare a flag cutoff date until ~20 days after
  this deploys, so the fleet is stamped first.


williamhbaker:
I think we need to add _something_ for `Discover` RPCs too. That's where a lot of the complexity we've talked about is, but I think it's fine to keep it as simple as possible for now. The motivating case is Discovers run for pre-existing tasks, like auto-discovers, where the feature flags at the time of the task creation are important to maintain for those subsequent discovers as well. I think it can be absent for a **new** task's Discover, since there won't be a date available to use, in which case the connector processing the Discover would need to assume something...probably the current time. But it should reflect the tasks actual `created_at` (maybe from the `pub_id`) for pre-existing tasks.

crates/validation/src/lib.rs
```diff
@@ -615,6 +615,26 @@ where
...
 }
 
+// The task creation date (UTC, YYYY-MM-DD) which is stamped into built task
+// specs, resolving connector feature-flag defaults as-of task creation.
+// An existing task's creation time is embedded in its control-plane Id.
+// A task being created has no control-plane Id yet — it's assigned only as
+// the publication commits — and its creation time is that of the publication
+// creating it. An empty date means the creation time is unknown, for which
+// connectors resolve their oldest ("epoch zero") feature-flag defaults.
+fn created_at_date(control_id: models::Id, pub_id: models::Id) -> String {
```
williamhbaker: Any chance of this causing test flakes? The vast majority of the times these will end up being the same, but for a build right around midnight...

dgreer-dev: good call out, fixed because we now only use the pub id ever

dgreer-dev:
> I think we need to add _something_ for `Discover` RPCs too. That's where a lot of the complexity we've talked about is, but I think it's fine to keep it as simple as possible for now.

Done. Discover carries the live task's date (auto-discovers included — the controller already holds the task's Id), absent for new tasks. And I took your suggestion on first builds: no pub_id fallback, first build is empty, absence uniformly means "assume a current date."

crates/proto-flow/src/lib.rs
```diff
@@ -101,6 +101,28 @@ impl ops::log::Level {
     }
 }
 
+/// Parse a `created_at` wire value: a UTC date in RFC 3339 "full-date"
+/// format (YYYY-MM-DD). Empty (the proto default) and malformed values
+/// are both None: the task's creation date isn't yet known, and a
+/// connector should assume a current date.
+fn parse_created_at(created_at: &str) -> Option<chrono::NaiveDate> {
```
williamhbaker: The new functions here appear to be unused. If that's correct, I'd vote we drop them for now, and add something like this back if the need emerges elsewhere.

williamhbaker (dismissed):
LGTM, with one suggestion.

Merging/deploying something like this where new fields are added to the protocols has historically been problematic with the strict unknown fields rules. I think most everything has been released with the change to make that lenient, but probably not _everything_ (flow-web? gitpod? etc.). So I'd also suggest spending some time studying recent releases and incidents to try to make sure we don't repeat history with this.

williamhbaker (approved):

crates/proto-flow/src/lib.rs
dgreer-dev: Yep, good spot

PR #3163, opened 2026-07-13
mise: remove Lima VM host share and other VM tweaks

jgraettinger:
Lima VMs now check out the flow repo to their VM-internal disk. This plays better with git worktrees: git gets confused when operating on volumes shared from the Mac host.

Also fix:
* A mis-ordering which prevented the ~/cargo-target btrfs mount from being created correctly
* A session restart issue with Lima impacting groups (docker) on first boot
* Add drata exclusions for GCP dev VMs
* Use a larger GCP disk
* Enable nested virtualization

### DIFF
--- a/mise/README.md
+++ b/mise/README.md
@@ -52,11 +52,14 @@ source ~/.zshrc
 
 Choose based on your needs:
 
+Both VM types clone Flow using your forwarded SSH agent. Ensure a key is loaded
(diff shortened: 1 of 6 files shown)


jgraettinger:
Verified through a full end-to-end bring-up and bring-down using Lima and GCP VMs.

skord (approved):
LGTM

PR #3165, opened 2026-07-14
runtime-next: discard hinted frontier when committed frontier is rebuilt

jgraettinger:
## Problem

When the materialization leader rebuilds a task's **committed frontier** from an authoritative recovered checkpoint — a legacy V1 checkpoint during a V1→V2 migration, or a remote-authoritative connector checkpoint — the recovered **hinted frontier** may reference producers that the rebuilt committed frontier no longer carries.

`project_hinted` zeroes each hinted producer's read offset and relies on a matching committed-frontier producer to restore it. An orphaned hint (one with no committed counterpart) therefore resolves to **journal offset 0**, forcing a full re-read of the journal from the beginning to replay a hint whose data the authoritative checkpoint already reflects. On large journals this can prevent a task from completing startup.

## Fix

- Discard the recovered hinted frontier whenever the committed frontier is rebuilt from an authoritative checkpoint. The checkpoint is itself a complete committed frontier, so there is nothing to replay.
- Clear the hinted frontier from RocksDB via the startup cleanup `Persist` (`delete_hinted_frontier`), so stale hints can't resurface on a later recovery.

This applies to both rebuild paths (legacy V1 checkpoint and connector-authoritative checkpoint). The no-rebuild path, the connector hinted-close delta path, and the drop-rollback path are unchanged.

## Refactor + tests

Extracts the startup recovery reconciliation out of the async `run()` path into a pure `reconcile_recovered` function (no IO), making the reconciliation policy unit-testable. Adds unit tests covering:

- the discard on the legacy-rebuild branch,
- the discard on the connector-authoritative branch,
- hint preservation (replay from the committed offset) when there is no rebuild,
- the connector hinted-close delta fold,
- the drop-rollback path deleting only the legacy checkpoint.

All `runtime-next` `leader::materialize::startup` tests pass.

### DIFF
--- a/crates/runtime-next/src/leader/materialize/startup.rs
+++ b/crates/runtime-next/src/leader/materialize/startup.rs
@@ -6,7 +6,7 @@ use bytes::Bytes;
 use futures::{StreamExt, stream::BoxStream};
 use prost::Message;
 use proto_flow::flow;
-use proto_gazette::uuid;
+use proto_gazette::{consumer, uuid};
 use std::collections::BTreeMap;
 use tokio::sync::mpsc;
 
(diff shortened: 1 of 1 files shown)


williamhbaker (approved):
LGTM

PR #3170, opened 2026-07-15
docs: reorganize agent skills page by plugin, add derivations and schema

jwhartley:
## Summary

`site/docs/guides/agent-skills.md` had gone stale: it only listed the original three plugins (captures, materializations, operations) with a full table of every individual skill. Two new plugins shipped in `estuary/agent-skills` (`estuary-derivations` PR #7, `estuary-schema` PR #8) with no doc update, and the per-skill table format means this will keep happening every time a skill is added or renamed.

## Changes

- Replace the three per-skill tables with a short paragraph per plugin (what it covers, not an exhaustive skill list), so new skills within an existing plugin don't require a doc edit.
- Reorder plugin sections operations-first (useful on every task regardless of source/destination), then captures, derivations, materializations, schema.
- Add the `estuary-derivations` and `estuary-schema` sections.
- Add an info callout noting that new skills/plugins may exist beyond what's summarized here, pointing to the `agent-skills` repo and the `/plugin` Discover tab as the live source of truth.
- Update the Claude Code install command block to include `/plugin install estuary-derivations@estuary` and `/plugin install estuary-schema@estuary`, reordered to match the new section order.

## Test plan

- [ ] Docs preview build renders the new `:::info` callout and section order correctly

### DIFF
--- a/site/docs/guides/agent-skills.md
+++ b/site/docs/guides/agent-skills.md
@@ -20,48 +20,31 @@ Before installing the skills, make sure you have:
 
 ## What's included
 
-The skills are grouped into three plugins: captures, materializations, and operations.
+Skills are grouped into plugins by workflow area, including operations, captures, derivations, materializations, and schema.
 
-### Captures (sources)
-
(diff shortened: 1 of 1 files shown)


aeluce (approved):
Thanks! Wasn't a huge fan of reiterating what was already in the repo readme. This is better.

PR #3171, opened 2026-07-15
docs: add Processing order section for materialization binding priority

jwhartley:
## Summary
- `priority` on materialization bindings was only documented as an inline spec comment (no heading, no anchor, no use-case explanation).
- Adds a `## Processing order` section mirroring the `### Read priority` treatment on the derivations page, so the concept is discoverable and linkable on its own.
- Updates the spec example's `priority` from `0` to `40` so it reads as a real value being set, not the default.


aeluce (approved):
Cool, thanks!

PR #3172, opened 2026-07-15
fix(flowctl): honor -q short flag on raw get (#3084)

### LINKED ISSUE #3084: flowctl: raw get help documents -q short flag but only --query is accepted
## Summary

`flowctl raw get --help` documents a `-q` short flag in its own usage example, but the parser only accepts the long form `--query`. The help text is self-contradictory: it tells you to use `-q`, and `-q` errors.

## Repro

The `-q` short flag shown in the command's own help example does not exist; it errors as an unknown argument. Using `-q` alone:

```
flowctl raw get --table data_planes -q select=data_plane_name
error: unexpected argument '-q' found
```

The equivalent with the long flag works (repeat `--query` for multiple params):

```
flowctl raw get --table data_planes --query select=data_plane_name --query id=eq.<some-id>
[{"data_plane_name":"..."}]
```

So this is not a "can't combine `-q` and `--query`" issue; `-q` is simply undefined. Workaround: always use `--query`.

## Root cause (confirmed in source)

`crates/flowctl/src/raw/mod.rs`

The doc comment on the `Get` variant (which clap renders as the `raw get` long help) shows `-q` (lines 43-44):

```rust
/// Pass query arguments as multiple `-q key=value` arguments.
/// For example: `-q select=col1,col2 -q col3=eq.MyValue`
Get(Get),
```

But the `query` arg on the `Get` struct is defined long-only, with no `short = 'q'` (lines 92-94):

```rust
/// Optional query parameters.
#[clap(long, value_parser = parse_key_val::<String, String>, number_of_values = 1)]
query: Vec<(String, String)>,
```

So the documented `-q` example can never parse. The same long-only `query` pattern is used by `Update` (lines 103-105) and `Rpc` (lines 117-119), whose help refers back to `get` for query usage, so the inconsistency propagates to those commands too.

## Broader observation (open question for maintainers)

This is one instance of a more general thing: flowctl's short-flag conventions vary across subcommands, and help examples don't always match what the parser accepts. A quick scan found only two real short flags in the whole CLI: the global `-o`/`--output` (`crates/flowctl/src/output.rs:8`) and `-f`/`--flows` (`crates/flowctl/src/catalog/list/mod.rs:10`). The one place a short flag appears in a help *example* is the `-q` above, which doesn't exist.

Tools and AI agents take help text literally, so a documented-but-unparseable flag costs them disproportionately (they'll emit `-q` because the help told them to, then fail). Two reasonable directions, leaving the call to you:

1. Add `short = 'q'` to the `query` args so the documented example works as written.
2. Or change the help example to use `--query` so it matches the parser.

No strong opinion on whether flowctl should standardize short aliases more broadly; flagging it as a DX/predictability question for whoever owns the CLI.

## Priority

Low. DX papercut with an easy workaround (`--query`). Noting it mainly because it's a cheap fix and it trips up scripts and agents more than humans.

0xfandom:
**Description:**

`flowctl raw get` documents a `-q` short flag in its own long help, but the parser only ever accepted `--query`, so the documented invocation fails:

```
$ flowctl raw get --table data_planes -q select=data_plane_name
error: unexpected argument '-q' found
```

The help text comes from the `Get` variant's doc comment ("Pass query arguments as multiple `-q key=value` arguments. For example: `-q select=col1,col2 -q col3=eq.MyValue`"), while `Get::query` was declared `#[clap(long, ...)]` — long-only. This adds `short = 'q'`, making the parser match what the help has always promised.

I took the add-the-flag route rather than amending the help text: the documented spelling is the friendlier one, and `-q` is unused across flowctl — the only short flags are the global `-o/--output` and `catalog list`'s `-f/--flows` — so it was free to take.

A second commit extends `-q` to `raw rpc` and `raw update`, which take the same query parameters and whose help defers to `get` ("As with `get`, you may pass optional query parameters"). It's split out so it can be dropped if you'd rather keep this strictly to #3084.

Fixes #3084

**Workflow steps:**

`-q` now works anywhere `--query` did, and is repeatable:

```
# Previously failed; now works:
flowctl raw get --table data_planes -q select=data_plane_name

# Repeatable, and mixable with the global -o:
flowctl raw get --table data_planes -o json -q select=id,name -q id=eq.42

# Same short flag now on rpc and update:
flowctl raw rpc --function my_fn -q key=value --body '{}'
flowctl raw update --table my_table -q id=eq.42 --body '{}'
```

`--query` is unchanged and keeps working; this is purely additive, so no existing invocation breaks. `raw get --help` now renders `-q, --query <QUERY>`, matching the prose above it.

None. The only place `-q` is documented is the command's own long help, which this change makes accurate rather than requiring an edit. Nothing under `site/docs/` references `raw get`'s query flag.

**Notes for reviewers:**

- Two commits: the first is the `get` fix for #3084, the second extends `-q` to `rpc`/`update` for consistency. Drop the second if it's out of scope for you.
- On verification: I couldn't complete `cargo build -p flowctl` locally — `simd-doc`'s C++ build fails on my machine with `rust::Slice<const unsigned char>::iterator` failing the `contiguous_iterator` concept check against my SDK's libc++ headers. That's a pre-existing local toolchain issue on a clean tree, unrelated to this change, and I didn't want to bundle a toolchain fix into a one-line CLI fix.
- So I verified the parsing behavior against clap 4.5.48 with the `Get`/`Rpc`/`Update` structs replicated exactly: the issue's invocation is rejected before the change and accepted after; `-q` and `--query` produce identical parses; repeated `-q` accumulates; `-q` doesn't collide with the global `-o`; and invalid values still fail through `parse_key_val` (`-q no-equals` → ``invalid KEY=value: no `=` found in `no-equals` ``). Relying on CI for the full build.
- No tests or snapshots currently cover `raw`'s arg parsing, so I didn't add any — happy to if you'd like coverage here.

### DIFF
--- a/crates/flowctl/src/raw/mod.rs
+++ b/crates/flowctl/src/raw/mod.rs
@@ -90,7 +90,7 @@ pub struct Get {
     #[clap(long)]
     table: String,
     /// Optional query parameters.
-    #[clap(long, value_parser = parse_key_val::<String, String>, number_of_values = 1)]
+    #[clap(short = 'q', long, value_parser = parse_key_val::<String, String>, number_of_values = 1)]
     query: Vec<(String, String)>,
 }
 
--- a/crates/flowctl/src/raw/mod.rs
+++ b/crates/flowctl/src/raw/mod.rs
@@ -101,7 +101,7 @@ pub struct Update {
     #[clap(long)]
     table: String,
     /// Optional query parameters.
-    #[clap(long, value_parser = parse_key_val::<String, String>, number_of_values = 1)]
+    #[clap(short = 'q', long, value_parser = parse_key_val::<String, String>, number_of_values = 1)]
     query: Vec<(String, String)>,
     /// Serialized JSON argument of the request.
     #[clap(long)]
@@ -115,7 +115,7 @@ pub struct Rpc {
     #[clap(long)]
     function: String,
     /// Optional query parameters.
-    #[clap(long, value_parser = parse_key_val::<String, String>, number_of_values = 1)]
+    #[clap(short = 'q', long, value_parser = parse_key_val::<String, String>, number_of_values = 1)]
     query: Vec<(String, String)>,
     /// Serialized JSON argument of the request.
     #[clap(long)]


0xfandom:
Friendly ping on this one — it's a one-line clap change (`raw get`'s help documents `-q` but only `--query` parses) and has been sitting a couple of weeks. Happy to rebase or adjust anything if that helps it land. No rush, just flagging it's ready whenever someone has a moment.

PR #3173, opened 2026-07-15
notifications: use "Estuary" in alert email subjects (#3000)

### LINKED ISSUE #3000: Update product name in alert emails
Basically all of the alert types in the `notifications` crate seem to be configured to include "Estuary Flow" in the subject line of the notification email. For cohesive messaging, we should change this to just "Estuary" as we have across other parts of the product.

0xfandom:
**Description:**

Alert emails prefixed their subject with `Estuary Flow:`, while the rest of the product says just "Estuary". This drops "Flow" from the subject of every alert type, and from the remaining body prose.

This mostly finishes a migration that was already underway — `FreeTrial`'s fired subject was already `Estuary Free Trial`, `MissingPaymentMethod`'s was already `Estuary: …`, and several bodies already read "Your Estuary account" / "Your Estuary {{spec_type}}".

Three commits, so the last two can be dropped independently if you'd rather scope this tightly:

1. **Subjects** — `Estuary Flow:` → `Estuary:` across all 13 subject templates. This is what #3000 asks for.
2. **Bodies** — the free-trial and missing-payment-method bodies still read "We hope you are enjoying Estuary Flow". Prose rather than subject lines, but the same "cohesive messaging" point.
3. **Tests** — closes a coverage gap the rename exposed (see below).

Fixes #3000

**Workflow steps:**

No API or CLI surface changes; this only affects the subject and body text of outbound alert emails. For example, a failing capture previously sent:

```
Estuary Flow: Task failure detected for capture acmeCo/test/capture
```

and now sends:

```
Estuary: Task failure detected for capture acmeCo/test/capture
```

Affected alert types: `AutoDiscoverFailed`, `BackgroundPublicationFailed`, `DataMovementStalled`, `FreeTrial`, `FreeTrialEnding`, `ShardFailed`, `TaskAutoDisabledFailing`, `TaskAutoDisabledIdle`, `TaskChronicallyFailing`, `TaskIdle`.

None. The subject strings live only in the `notifications` crate's handlebars templates; nothing under `site/` documents them.

**Notes for reviewers:**

- **A coverage gap turned up while doing this, and it's the part most worth a look.** The subject templates for `TaskIdle`, `TaskAutoDisabledFailing` and `TaskAutoDisabledIdle` had no test asserting them — I changed one to `BROKEN:` and the suite stayed green, so the rename was touching three lines nothing guarded. The third commit adds a fired-state test per type, following the existing `assert_email` + snapshot pattern; the same deliberate break now fails on `subject mismatch`. All three are fired-only, so there's no resolved case. Happy to split that commit into its own PR if you'd prefer it separate from a rename.
- **Snapshots:** only the `notifications__*` files are updated. The parallel `alert_notifications__*` set appears orphaned — no test renders them, and they don't regenerate — and #3079 is already removing them, so I left them alone rather than create a conflict. Each of the 5 updated snapshots is a one-line diff.
- `free_trial`'s resolved subject is a conditional; its `has_credit_card` branch becomes `Estuary: Paid Tier`, which lines up with the existing `else` branch (`Estuary Paid Tier: Enter Payment Info…`).
- Verified with `cargo test -p notifications` — 28 passing (25 existing + 3 new). `cargo fmt` clean.

### DIFF
--- a/crates/notifications/src/auto_discover_failed.rs
+++ b/crates/notifications/src/auto_discover_failed.rs
@@ -7,7 +7,7 @@ pub fn register_templates<'a>(registry: &mut handlebars::Handlebars<'a>) -> anyh
     registry
         .register_template_string(
             &fired_subject,
-            r#"Estuary Flow: Auto-discover failed for {{arguments.spec_type}} {{catalog_name}}"#,
+            r#"Estuary: Auto-discover failed for {{arguments.spec_type}} {{catalog_name}}"#,
         )
         .context("registering auto_discover_failed-fired-subject template")?;
 
@@ -32,7 +32,7 @@ pub fn register_templates<'a>(registry: &mut handlebars::Handlebars<'a>) -> anyh
     registry
         .register_template_string(
             &resolved_subject,
-            r#"Estuary Flow: Auto-discover resolved for {{arguments.spec_type}} {{catalog_name}}"#,
+            r#"Estuary: Auto-discover resolved for {{arguments.spec_type}} {{catalog_name}}"#,
         )
         .context("registering auto_discover_failed-resolved-subject template")?;
 
--- a/crates/notifications/src/background_publication_failed.rs
+++ b/crates/notifications/src/background_publication_failed.rs
@@ -9,7 +9,7 @@ pub fn register_templates<'a>(registry: &mut handlebars::Handlebars<'a>) -> anyh
     registry
         .register_template_string(
             &fired_subject,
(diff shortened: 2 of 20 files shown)


aeluce:
Hey, thanks for the interest in contributing to Estuary!

I ended up wanting to change some other aspects of notification emails beyond the "Estuary vs. Estuary Flow" issue, so I opened my own PR to address #3000. Appreciate the thought, though!

PR #3175, opened 2026-07-15
control-plane: add `closed` flag to data planes

GregorShear:
## What

Adds a `closed` boolean to data planes and lets the `dataPlanes` query filter on it.

- **DB:** new `data_planes.closed` column, `boolean NOT NULL DEFAULT false` (existing planes stay open), with a `SELECT (closed)` grant to `authenticated`.
- **GraphQL:** `closed: Boolean!` on the `DataPlane` type, plus a `DataPlanesFilter` argument on `dataPlanes` carrying `closed: BoolFilter`. `filter.closed.eq` narrows to matching planes; omitting it returns both open and closed planes.

## Why

Lets operators retire a data plane from new selection - a closed plane keeps serving existing tasks but drops out of listings. "Closed" reflects that the plane is still fully live; it's only withheld from new selection.

## Notes

- `closed` rides on the authorization snapshot, so both the filter and the per-node value resolve in-memory with no database round-trip. The filter is applied before pagination, keeping page sizes and cursors correct. The snapshot lags the DB by up to one refresh interval, which is acceptable for a listing filter.
- The unauthenticated `publicDataPlanes` query (pre-signup selection) excludes closed planes unconditionally - a retired plane should never be offered to a new signup. It takes no filter argument.


jshearer (approved):
LGTM thanks

PR #3181, opened 2026-07-15
control-plane-api: surface per-store health diagnostics in storage-mapping mutations

GregorShear:
Storage mapping mutations run a health check to verify reachability between dataplanes and storage buckets, but until recently the journal client was swallowing diagnostic info on failed tests.

This PR plumbs the new Error::FragmentStoreUnhealthy through to the gql mutation errors so the UI can surface it to users.


jshearer (approved):
LGTM!

PR #3182, opened 2026-07-15
Bound too-large error messages over RPC

### LINKED ISSUE #3177: runtime-v2: oversized causal-hint-timeout error masks the real failure (swallowed as `too_many_continuations`)
## Summary

When runtime-v2 hits a causal-hint-resolution timeout, the error it produces embeds the
**entire serialized frontier** — one line per unresolved hint/producer. In the real
incidents this was ~1400 hints across ~540 journals, i.e. a multi-thousand-line error
string. That string is too large to survive either path out of the process:

1. **gRPC propagation (leader → shard):** the error is serialized into a `tonic::Status`
   message and transmitted in HTTP/2 trailers. The oversized trailer exceeds the `h2`
   CONTINUATION-frame limit, so the connection aborts with `too_many_continuations` — and
   that transport error *replaces* the real status text. The operator sees an opaque
   `Some resource has been exhausted` / `h2 protocol error`, never the actual cause.
2. **Log pipeline:** the same string, logged leader-side, is truncated by the log
   pipeline, so it's unreadable there too.

Net effect: the true root cause (`causal hint resolution timed out: N hints unresolved
across M journals`) is invisible in both the surfaced task error and the logs, and had to
be reverse-engineered from the recovery-log DB. This masking happened in two separate
incidents and each time cost real time before someone recognized the pattern.

## Evidence

What operators saw as the terminal task error:

```
Materialize error (expected leader message) from leader

Caused by:
  0: status: 'Some resource has been exhausted', self: "h2 protocol error: error reading a body from connection"
  1: error reading a body from connection
  2: connection error detected: detected excessive load generating behavior (b"too_many_continuations")
```

The actual root cause (only recovered by other means):

```
SessionResponse error (expected next_checkpoint) from (in-process)@0: causal hint resolution timed out: 1413 hint(s) unresolved across 537 journal(s) journal "acmeCo/core_data/some_collection/<...>/pivot=00" binding=1 producer=Producer(<...>) last_commit=Clock(0s 0ns) hinted_commit=Clock(<...>) journal "acmeCo/core_data/some_collection/<...>/pivot=00" binding=1 producer=Producer(<...>) last_commit=Clock(0s 0ns) hinted_commit=Clock(<...>) journal "acmeCo/core_data/some_collection/<...>/pivot=80000000" ...  [continues for hundreds of journals]
```

## Mechanism / code references

- The unbounded string is built in `crates/shuffle/src/session/state.rs:511-538`
  (`SessionState::on_tick`): on stall it loops over **every** `self.unresolved.journals`
  and **every** hinted `producer`, appending a
  `\n  journal … binding=… producer=… last_commit=… hinted_commit=…` line with **no cap**,
  then `anyhow::bail!`s the whole thing.
- That error bubbles up as the leader/shuffle-session terminal error and crosses
  leader → shard over gRPC. `anyhow_to_status` (`crates/runtime-next/src/lib.rs:117-122`)
  maps it via `tonic::Status::unknown(format!("{err:?}"))`, so the *entire* error chain —
  including the giant detail dump — becomes the gRPC status message. gRPC status messages
  ride in HTTP/2 trailers; an oversized trailer exceeds the `h2` CONTINUATION-frame limit
  and the transport aborts with `too_many_continuations`, dropping the real message.

## Acceptance criteria

- A causal-hint timeout with thousands of unresolved hints surfaces as a readable task
  error whose first line is the real cause (`causal hint resolution timed out: N hints
  unresolved across M journals`), not `too_many_continuations`.
- The error string is bounded regardless of hint/journal count.
- No gRPC status message emitted by the leader can exceed the h2 trailer limit (or, if it
  does, it is truncated rather than dropped).
- (Nice to have) a test that constructs an oversized unresolved frontier and asserts the
  resulting error / `Status` message is bounded and still contains the root-cause prefix.

dgreer-dev:
**Description:**

Fixes #3177: an oversized runtime-v2 task error that masked the real failure. In the real case, we had a huge number of frontiers which ended up with a multi-thousand  line string. That string was too big to cross gRPC (the oversized HTTP/2 trailer tripped `h2`'s `too_many_continuations` guard, which aborted the connection and replaced the status text with an opaque `Some resource has been exhausted` / `h2 protocol error`) and was also truncated in the log pipeline. Net effect: the true cause was invisible on both paths and had to be reverse-engineered from the recovery-log DB.

Two independent fixes:

- **Bound gRPC status messages at the boundary** (the load-bearing fix). Any status message runtime-next emits is capped to `MAX_STATUS_MESSAGE_LEN` so it fits within a single HTTP/2 frame and never needs a CONTINUATION frame. Truncation is from the tail, preserving the root-cause prefix the anyhow format leads with. This covers both the errors we format ourselves *and* pass-through statuses (round-tripped from a peer or produced by a connector), so a misbehaving connector emitting a huge message can't produce a status that's dropped in transit.
- **Bound the causal-hint error at the source.** The per-producer detail is capped (first N lines, then `… and M more unresolved hint(s)`), which is what keeps it readable in the logs — the boundary cap only helps what crosses gRPC.

**Notes for reviewers:**

- Two commits — (1) bounds every gRPC status message runtime-next emits (errors we format *and* pass-through statuses from connectors/peers, with code/details/metadata preserved) so an oversized trailer can't be dropped as `too_many_continuations`, and (2) caps the causal-hint-timeout detail at the source so it stays readable in logs; the only un-capped statuses left are hand-built ones that are bounded by construction. 

- `MAX_STATUS_MESSAGE_LEN = 4096` stays inside one 16 KiB `h2` frame even under worst-case 3× `grpc-message` percent-encoding 

(I also want to point out that the short hash for the first commit is `8222222` which is a lot of 2's and so hopefully auspicious)

### DIFF
--- a/crates/runtime-next/src/leader/derive/handler.rs
+++ b/crates/runtime-next/src/leader/derive/handler.rs
@@ -246,8 +246,8 @@ where
 
     // Best-effort broadcast of terminal error to all shards.
     let status = match err.downcast_ref::<tonic::Status>() {
-        Some(status) => status.clone(),
-        None => tonic::Status::unknown(format!("{err:?}")),
+        Some(status) => crate::bound_status(status.clone()),
+        None => crate::bounded_unknown_status(format!("{err:?}")),
     };
     for tx in error_tx {
         let _ = tx.send(Err(status.clone()));
--- a/crates/runtime-next/src/leader/materialize/handler.rs
+++ b/crates/runtime-next/src/leader/materialize/handler.rs
@@ -262,8 +262,8 @@ where
 
     // Best-effort broadcast of terminal error to all shards.
     let status = match err.downcast_ref::<tonic::Status>() {
-        Some(status) => status.clone(),
-        None => tonic::Status::unknown(format!("{err:?}")),
+        Some(status) => crate::bound_status(status.clone()),
+        None => crate::bounded_unknown_status(format!("{err:?}")),
     };
     for tx in error_tx {
         let _ = tx.send(Err(status.clone()));
--- a/crates/runtime-next/src/lib.rs
+++ b/crates/runtime-next/src/lib.rs
@@ -111,14 +111,71 @@ impl RuntimeProtocol {
     }
 }
 
+/// Maximum byte length of a `tonic::Status` message that we build from a
+/// formatted error. gRPC status text rides in an HTTP/2 trailer; an oversized
+/// trailer forces the header block across many CONTINUATION frames, tripping
+/// `h2`'s `too_many_continuations` guard, which aborts the connection and
+/// *replaces* the real status text with an opaque transport error
+///
+/// The ceiling exists so that *any* error, anticipated or not, stays within a
+/// single HTTP/2 frame (default `max_frame_size` is 16 KiB) and never needs a
+/// CONTINUATION frame at all. The `grpc-message` trailer is percent-encoded, so
+/// a byte can expand up to 3x (`%XX`); 4 KiB raw is ≤ 12 KiB encoded, fitting
+/// one frame even in the worst case.
+pub const MAX_STATUS_MESSAGE_LEN: usize = 4096;
+
 // Map an anyhow::Error into a tonic::Status.
-// If the error is already a Status, it's downcast.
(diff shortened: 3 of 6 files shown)


williamhbaker (approved):
LGTM

PR #3186, opened 2026-07-16
control-plane-api: seed a 12h DataMovementStalled alert config for new tenants

jshearer:
**Description:**

Tenant provisioning now seeds an `alert_configs` row at the new tenant's prefix:

```json
{ "dataMovementStalled": { "enabled": true, "condition": { "stalledFor": "12h" } } }
```

With this row in place, every task in a new tenant is covered by the `DataMovementStalled` alert: a task that has moved no data for 12 hours fires an alert to the tenant's default subscription (new tenants are already subscribed to `data_movement_stalled`), and the alert resolves automatically once bytes flow. The controller's age gate means a task must be at least 12 hours old before it can fire, so freshly created tasks get that long to move their first byte.

The row is seeded at the tenant prefix — the broadest layer of the `alert_configs` hierarchy — so it behaves exactly like a tenant-wide setting the user configured themselves: more-specific prefix or per-task rows override it field by field, it can be edited or deleted through the GraphQL API and `flowctl alerts configs`, and `enabled: false` at any narrower scope opts out. `on conflict do nothing` preserves any config that already exists at the prefix.

Existing tenants are unchanged: no backfill, no change to controller defaults, and the legacy `alert_data_processing` fallback continues to apply to tenants without any `alert_configs` coverage.


jwhartley:
doc files no longer do.

at build time.

Suggested split:

  - `site/docs/guides/notifications/customize-alerts.md` becomes `content/guides/notifications/customize-alerts.md`

PR #3187, opened 2026-07-16
runtime-next: persist committed_close on ahead-remote-authoritative recovery

### LINKED ISSUE #3176: runtime-v2: ahead-remote-authoritative recovery promotes committed_close in memory only
### Summary
In the runtime-next materialize leader startup, when a remote-authoritative
connector checkpoint is recovered *ahead* of the local committed frontier (its
embedded close clock matches the recovered `hinted_close`), the leader promotes
the hinted transaction to committed **in memory only**: `committed_close` and
`committed_frontier` are advanced but never reconciled to RocksDB. A later crash
can then leave persisted state that matches neither the connector's authoritative
checkpoint nor the recovered hint, and startup `bail!`s — wedging the shard.

### Impact
High: stuck shard / crash loop. Once wedged, the leader `bail!`s during startup
before running any transaction, so the connector never receives another commit and
its checkpoint stays put — the shard can't recover on its own. Requires a
remote-authoritative connector (one whose checkpoint the runtime treats as
authoritative) plus the crash sequence below.

### Where
`crates/runtime-next/src/leader/materialize/startup.rs`

- The ahead-remote-authoritative arm (`clock == Some(hinted_close)` →
  `committed_close = hinted_close; committed_frontier = committed_frontier.reduce(hinted_frontier)`)
  does **not** set `committed_frontier_rebuilt`.
- The startup reconcile Persist fires only on
  `committed_frontier_rebuilt || delete_legacy_checkpoint`, and even then never
  writes `committed_close_clock`.
- So the promotion stays in memory; RocksDB keeps the stale `committed_close`.
- On the next recovery, `frontier_mapping::extract_committed_close(&connector_checkpoint)`
  matching neither the recovered `committed_close` nor `hinted_close` hits the
  `anyhow::bail!("...doesn't match Recover's committed_close ... or hinted_close ...")` arm.

### Reproduction (two crashes)
Let C0 = last durable committed close, H1 = the hinted txn the connector commits,
H2 = a later hint.

1. A txn hint-Persists at H1 but crashes before its commit Persist →
   RocksDB `{committed_close: C0, hinted_close: H1}`.
2. Recovery: connector checkpoint reports H1 == recovered `hinted_close` →
   **promote in memory** (`committed_close = H1`); RocksDB untouched.
3. Leader runs; the first txn hint-Persists at H2 →
   RocksDB `{committed_close: C0, hinted_close: H2}`.
   RocksDB `{committed_close: C0, hinted_close: H2}`.
4. Crash before the connector commits (its checkpoint still reports H1).
5. Recovery: RocksDB `{committed_close: C0, hinted_close: H2}`, connector reports
   H1 → matches neither → `bail!` → shard wedged.

dgreer-dev:
When a remote-authoritative connector checkpoint is recovered ahead of the local committed frontier (its embedded close clock matches the recovered hinted_close), the materialize leader promotes the hinted transaction to committed. That promotion advanced committed_close and folded the hinted frontier into committed in memory only -- it was never reconciled to RocksDB, because the startup cleanup Persist fired solely on committed_frontier_rebuilt || delete_legacy_checkpoint and never wrote committed_close_clock.

A later crash could then leave persisted state matching neither the connector's authoritative checkpoint nor the recovered hint, and the next startup would bail! -- wedging the shard in a crash loop.

Fire the cleanup Persist on promotion too, rewriting the folded committed frontier, clearing the consumed hinted frontier, and writing committed_close_clock. The write is gated to the promotion arm; the two authoritative rebuild paths leave committed_close as recovered and keep writing the 0 "unchanged" sentinel.

Fixes #3176.

### DIFF
--- a/crates/runtime-next/src/leader/materialize/startup.rs
+++ b/crates/runtime-next/src/leader/materialize/startup.rs
@@ -317,6 +317,20 @@ fn reconcile_recovered(
     // checkpoint nor the recovered hint, and the next startup `bail!`s — see #3176.
     let mut committed_close_promoted = false;
 
+    // Set when an authoritative checkpoint (legacy V1, connector rebuild, or
+    // connector promotion) replaces `pending_ack_intents`. The cleanup Persist must
+    // then durably swap RocksDB's ACK intents to match (`delete_ack_intents` + the
+    // new set); otherwise a crash before the first go-forward commit re-recovers the
+    // stale ACK intents and the checkpoint's transaction is never acknowledged.
+    let mut ack_intents_replaced = false;
+
+    // A promotion that advances `committed_close` past a maintained V1-rollback
+    // legacy checkpoint leaves that checkpoint embedding the old close, which the
+    // next recovery would `bail!` on (legacy-close mismatch). Set to the connector's
+    // authoritative checkpoint — which, in legacy mode, IS the full checkpoint at the
+    // promoted close — so the cleanup Persist re-persists a consistent legacy image.
+    let mut reconciled_legacy_checkpoint: Option<consumer::Checkpoint> = None;
+
     // Handle migration from `legacy_checkpoint`.
     let legacy_checkpoint_present = legacy_checkpoint.is_some();
     if let Some(legacy_checkpoint) = legacy_checkpoint {
@@ -351,6 +365,7 @@ fn reconcile_recovered(
             committed_frontier_rebuilt = true;
 
             pending_ack_intents = legacy_checkpoint.ack_intents;
+            ack_intents_replaced = true;
         }
     } else {
         service_kit::event!(
@@ -385,7 +400,15 @@ fn reconcile_recovered(
             committed_close = hinted_close;
             committed_frontier = committed_frontier.reduce(std::mem::take(&mut hinted_frontier));
             committed_close_promoted = true;
+            ack_intents_replaced = true;
 
+            // In legacy-rollback mode the connector stores the *full* checkpoint
+            // (frontier + committed-close key + ACK intents), so its checkpoint at
+            // the promoted close IS the legacy checkpoint we must re-persist. Clone
+            // before moving `ack_intents` out below.
+            if legacy_checkpoint_present && !drop_v1_rollback {
+                reconciled_legacy_checkpoint = Some(connector_checkpoint.clone());
+            }
             pending_ack_intents = connector_checkpoint.ack_intents;
         } else if let Some(clock) = clock {
             // Implementation error: these update together and should always sync.
@@ -411,6 +434,7 @@ fn reconcile_recovered(
             committed_frontier_rebuilt = true;
 
             pending_ack_intents = connector_checkpoint.ack_intents;
+            ack_intents_replaced = true;
         }
     } else {
         service_kit::event!(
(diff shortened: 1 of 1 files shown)


crates/runtime-next/src/leader/materialize/startup.rs
```diff
@@ -410,24 +421,39 @@ fn reconcile_recovered(
...
+            committed_close_promoted,
             delete_legacy_checkpoint,
             "reconciling recovered checkpoint state",
         );
         proto::Persist {
             seq_no: 0,
-            delete_committed_frontier: committed_frontier_rebuilt,
-            delete_hinted_frontier: committed_frontier_rebuilt,
-            committed_frontier: committed_frontier_rebuilt
+            committed_close_clock: if committed_close_promoted {
```
williamhbaker, line 493: Adding the `committed_close_clock` makes sense, as that's the whole issue as identified, but beyond that I'm also not sure how this works without `ack_intents`/`delete_ack_intents` and also the `legacy_checkpoint` itself. As best I can tell we need those too in order to prevent different flavors of the same issue that just adding the `committed_close_clock` solves.

I'm not feeling super confident about this though. @jgraettinger can you weigh in?

dgreer-dev: Ah yeah, that's a good point on both.  Let me spend a few brain cells on this.

dgreer-dev: Yeah okay, I was able to write a few red tests based on your thoughts.  Let me fix those new tests and I'll push up hopefully a solution

dgreer-dev: Pushed `06cba43c359`. You were right on both — a partial image just relocates the wedge, so the promotion's startup cleanup Persist now writes a complete image of the promoted commit, mirroring a normal commit Persist.

The two red tests pin down each flavor (both round-trip through the real `encode_persist`/`decode_recover_key_value` codecs):

- **legacy checkpoint** — after the promotion advances `committed_close` to H1, the recovered legacy checkpoint still embedded the old close, so the next recovery bailed on the mismatch — the same wedge, moved to the legacy arm. My first commit actually introduced this one.
- **ACK intents** — the promotion adopts the connector's ACK intents in memory but never made them durable, so a crash before the first commit replayed the stale ones and never acknowledged the promoted txn's output. Silent, same root cause.

On a promotion the Persist now writes: `committed_close_clock` + folded committed frontier (clearing the hinted frontier), swapped ACK intents (`delete_ack_intents` + the connector's set), and the legacy checkpoint re-persisted at the promoted close. In rollback mode the connector stores the full checkpoint, so its checkpoint at the promoted close *is* the legacy checkpoint to re-persist. I also applied the ACK-intent swap to the two rebuild arms, which had the same gap.

jgraettinger:
Oooof. A PR in this reconciliation logic could be perfect, but I'll still have a hard time stamping it. It's the most inscrutable bit of V2, much of it due to the V1 migration mechanics, and it's also been our biggest source of V2 bugs.

That said, we don't need to worry about ACKs because given a premise of a second hinted frontier write, then by construction we've already completed ACK intent writes. The hinted write happens only after the Flush state, which requires that the Tail FSM (including ACK intent writes) has already completed. 

One ... okay ... factoring is adding a `Reconcilled::hint_promote_persist: Option<Persist>` alongside `cleanup_persist`, which is applied after, and which has `Persist::committed_frontier = hinted_frontier` and `committed_close = hinted_close`.

Another thought that occurs is we could perhaps begin staging the Persist of the hinted frontier _here_, by setting committed_frontier / committed_close, so it commits atomically alongside a _next_ hinted_frontier.

I may need to think on this more.

dgreer-dev:
OBE

PR #3191, opened 2026-07-17
runtime: persist the max-key fail-safe sentinel at adoption

jgraettinger:
## Summary

Persist the max-key fail-safe sentinel at the moment of adoption, so the persisted state once again matches V2's native invariant that a missing `MK-v2:` key means "no keys were ever stored."

## Background

The V1 runtime can commit checkpoint progress for a binding without ever observing one of its keys (e.g. via ACK-only journal reads), in which case no `MK-v2:` max-key is persisted. On the next recovery the binding is active in the checkpoint without a persisted max-key, and recovery adopts a `[0xff]` fail-safe sentinel **in memory only**. Because no packed key compares greater than `[0xff]`, the binding's max-key never ratchets again and is never persisted — permanently.

This is end-to-end safe within V1 (the sentinel conservatively disables the load-skip optimization), but it breaks V1 → V2 migration semantics: the V2 runtime interprets a missing max-key as "no keys were ever stored", which enables the load-skip optimization for a binding that may hold extensive stored data.

## Change

- Durably persist the sentinel at the moment of adoption, synchronized to the recovery log before any Load or Store of the session. A missing `MK-v2:` key then once again means "no keys ever stored", matching V2's native invariant, so migration requires no V2-side changes.
- Recovery accepts the persisted sentinel: it is intentionally not a valid packed tuple (`0xff` is `tuple::ESCAPE`, above every type code), so it is matched exactly and used verbatim rather than unpacked. Any other unparseable value still fails as corruption. This same relaxation makes a V2 → V1 rollback of a task carrying the sentinel recover cleanly.

### DIFF
--- a/crates/runtime/src/materialize/protocol.rs
+++ b/crates/runtime/src/materialize/protocol.rs
@@ -166,21 +166,57 @@ pub async fn recv_connector_opened(
         .filter_map(|(journal, _source)| journal.split(';').skip(1).next())
         .collect();
 
-    // Fetch a persisted max-key value for each active binding.
-    let max_keys: Vec<String> = task.bindings.iter().map(max_key_key).collect();
-    let max_keys = db.multi_get_opt(max_keys, Default::default()).await;
+    let max_keys = recover_max_keys(db, &task, &active_read_suffixes).await?;
+
+    Ok((
+        task,
+        accumulator,
+        checkpoint,
+        opened,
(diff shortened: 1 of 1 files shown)


williamhbaker (approved):
LGTM

jgraettinger:
Sanity checked on a local stack.

PR #3192, opened 2026-07-17
docs: add Azure BYOC subscription provisioning prerequisites

jwhartley:
## What

Adds a "Prepare your subscription for provisioning" section to the Azure BYOC setup page covering the subscription-level limits a customer should raise before Estuary provisions a data plane:

- **vCPU quota** — new Azure subscriptions often default to a low or zero vCPU limit per family. Includes a table of the default VM families (and Intel v5 fallbacks) with vCPU counts at 2x steady-state for rolling upgrades.
- **Public IP quota** — request ~40, since rolling upgrades briefly double the instance count.
- **Region choice** — flags that East US / East US 2 are frequently capacity-constrained, and explains that Estuary provisions regionally (no pinned AZ) by default to maximize allocation success.

## Why

These prerequisites aren't documented anywhere today, so they surface as a back-and-forth during onboarding. Two recent Azure BYOC bring-ups hit the quota/capacity question. Sourcing them up front turns a multi-day round-trip into a single request.

## Notes

- The quota families/counts are pinned to est-dry-dock's current defaults; there's an MDX source comment (invisible on the rendered page) pointing the next editor at `est_dry_dock/constants.py` so the table can be refreshed if the fleet's default SKUs change.
- Regional-vs-zonal default was verified against est-dry-dock (`ask_cloud_zone` presents regional as the default choice; no proximity placement groups or zonal disks that regional scatter would break). Cross-AZ data transfer within a region is free as of Azure's June 2024 change, so regional placement carries no extra data cost.

### DIFF
--- a/site/docs/private-byoc/byoc-deployments/azure-byoc.md
+++ b/site/docs/private-byoc/byoc-deployments/azure-byoc.md
@@ -47,3 +47,50 @@ Finally, provide the following information to your Estuary point of contact:
  - Subscription ID (found in Subscriptions -> Overview)
  - Tenant ID (found in Tenant Properties)
  - Azure region
+
+## Prepare your subscription for provisioning
+
+After you have shared the details above, you may need to raise a few
+subscription limits so Estuary can provision the data plane. Depending on your
+subscription's existing limits, some of these may already be sufficient. You
+can do this in parallel with the steps above.
+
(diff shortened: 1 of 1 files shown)


jwhartley:

as-is. For this PR:

- `site/docs/private-byoc/byoc-deployments/azure-byoc.md` becomes `content/private-byoc/byoc-deployments/azure-byoc.md`

PR #3193, opened 2026-07-17
shuffle, gazette: guard against wedged reads

jgraettinger:
Two defensive measures for the class of V2 materialization stalls where a read hangs indefinitely on broker/storage I/O.

## gazette: bound fragment-fetch connection establishment

Add a 30s `connect_timeout` to the HTTP/1 fragment-fetch client. Without it, a storage endpoint that accepts the socket but never completes the TCP+TLS handshake wedges a fragment fetch forever — TCP keepalive can't help while the peer is still ACKing, and reqwest sets no connect timeout by default. We deliberately avoid `read_timeout` (it's heavier — a boxed future on every read — and biases toward timeout over completion of a ready read given we may poll infrequently).

## shuffle: stall diagnostics for wedged reads

Emit periodic `WARN` diagnostics for non-tailing pending reads that make no progress across a ticker interval (60s). The mechanism arms on tick and disarms on any resolved read, so a warning requires a full inter-tick interval with zero reads resolving before firing. It's coarse — any read completion (including tailing bindings that don't unblock the heap) disarms it — but cheap, and gives noisy optics into truly wedged tasks. Each warning carries read_id, binding, journal, read/write offsets, and fragment metadata.

### DIFF
--- a/crates/shuffle/src/slice/actor.rs
+++ b/crates/shuffle/src/slice/actor.rs
@@ -124,6 +124,12 @@ impl SliceActor {
         let mut ticker = tokio::time::interval(crate::ACTOR_TICKER_INTERVAL);
         ticker.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Delay);
 
+        // Warning mechanism which is armed on tick and disarmed on read.
+        // This is a coarse mechanism as it's disarmed by any read, including
+        // completions of tailing bindings that don't unblock the heap,
+        // but it's cheap and gives us noisy optics into truly wedged tasks.
+        let mut stall_armed = false;
+
         let mut loop_count: u64 = 0;
         loop {
             loop_count += 1;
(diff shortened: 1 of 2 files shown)


crates/shuffle/src/slice/actor.rs
```diff
@@ -472,6 +484,39 @@ impl SliceActor {
...
+            };
+
+            service_kit::event!(
+                tracing::Level::WARN,
+                "stall",
+                read_id = read.id(),
+                binding = read_state.binding_index,
+                journal = read_state.journal.to_string(),
+                read_offset = read_state.read_offset,
+                write_head = read.write_head(),
```
dgreer-dev, line 511: Should this be `read_state.write_head` to match the previous line?  Is there a risk the live value from `write_head()` could diverge from what's in `read_state`?

dgreer-dev (approved):
Talked over the issue, looks good

crates/shuffle/src/slice/actor.rs
jgraettinger: discussed over VC

PR #3195, opened 2026-07-17
flowctl: preview-next appends a final drain session

mdibaiee:
A materialization session under runtime-v2 halts after its final commit without running its post-commit Acknowledge. Without an additional session, `flowctl raw preview-next` exits reporting the final transaction as committed while its `Acknowledge` has not been run, and for post-commit apply connectors, it means their transactions are not actually applied to the destination.

Every materialization preview now appends one final **empty drain session**: so `--sessions 2` actually ends up running 3 sessions, but it matches the expectation of having two sessions, driven up to their `Acknowledge`.

## Test plan

- [x] `cargo test -p flowctl`
- [x] E2E from the connectors repo, whose test harness dropped its appended-drain-transaction workaround in favor of this: `materialize-postgres` (1 shard, `--output-apply`/`--output-state`) passes deterministically — the drain session contributes exactly one no-op `["applied.actionDescription", ""]` line and, no longer committing a transaction, drops the previous workaround's `["connectorState",{}]` filler. `materialize-databricks` (2 shards, scale-out) passes with an unchanged snapshot, its final staged MERGE now applied by the drain session's recovery.

### DIFF
--- a/crates/flowctl/src/raw/preview_next/driver.rs
+++ b/crates/flowctl/src/raw/preview_next/driver.rs
@@ -213,8 +213,14 @@ async fn drive_one_shard(
             }))
             .map_err(|_| anyhow::anyhow!("serve task closed before Task"))?;
 
-        drive_session_responses(&request_tx, &mut response_rx, session_index, &stop_token, drain)
-            .await?;
+        drive_session_responses(
+            &request_tx,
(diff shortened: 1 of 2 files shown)


jgraettinger (approved):
LGTM, that was straigt forward!

PR #3200, opened 2026-07-18
Improve storage mapping GraphQL results

GregorShear:
Returns the affected storage mapping from create and update mutations. Also deprecates the optional query filter and scopes unfiltered results to catalog-readable prefixes.


crates/control-plane-api/src/server/public/graphql/storage_mappings.rs
```diff
@@ -558,59 +578,77 @@ pub type PaginatedStorageMappings = Connection<
...
+            None => (Vec::new(), None),
+            Some(StorageMappingsBy {
+                exact_prefixes,
+                under_prefix,
+            }) => {
+                let exact_prefixes = exact_prefixes.unwrap_or_default();
+                match (exact_prefixes.is_empty(), under_prefix.is_none()) {
+                    (true, true) => {
+                        return Err(
+                            "must provide exactly one of `exactPrefixes` or `underPrefix`".into(),
```
jshearer: nit: omitting `by` entirely is also valid now:

```suggestion
                            "provide exactly one of `exactPrefixes` or `underPrefix`, or omit `by` entirely".into(),
```

crates/control-plane-api/src/server/public/graphql/storage_mappings.rs
```diff
@@ -558,59 +578,77 @@ pub type PaginatedStorageMappings = Connection<
...
-            exact_prefixes.iter().collect()
         };
 
-        let policy_result = crate::server::evaluate_names_authorization(
-            &env.snapshot(),
-            claims,
-            models::Capability::Read,
-            prefixes_to_check,
+        let snapshot = env.snapshot();
+        let mut read_prefixes = super::authorized_prefixes::authorized_prefixes(
```
jshearer, line 627: `authorized_prefixes` admits a grant whose bits come only from `bundles` (legacy `capability = None`), but `get_user_capability` below only considers legacy capabilities and filters `None` out. So a row that's readable only through such a grant fails the whole request with "missing capability for catalog prefix".

Pre-existing (the sibling queries share it), but unfiltered mode makes it much easier to hit: one bundle-only grant anywhere in the caller's graph now breaks their whole unfiltered listing. Can bundle-only user grants exist yet? If not, happy for this to be a follow-up.

crates/control-plane-api/src/server/public/graphql/storage_mappings.rs
```diff
@@ -558,59 +578,77 @@ pub type PaginatedStorageMappings = Connection<
...
+            &snapshot.role_grants,
+            &snapshot.user_grants,
+            env.claims()?.sub,
+            models::authz::Capability::CatalogRead,
+            under_prefix.as_ref().map(|prefix| prefix.as_str()),
         );
-        env.authorization_outcome(policy_result).await?;
+
+        if !exact_prefixes.is_empty() {
+            read_prefixes.retain(|authorized| {
```
jshearer, line 636: This flips the contract for unauthorized prefixes: an `exactPrefixes` entry the caller can't read used to fail the whole query, and now it's silently dropped (the `narrowed` snapshot pins this). Same for an `underPrefix` broader than the caller's grants.

Silent scoping seems like the right shape for a listing API, but a consumer can no longer tell "no access" from "no mapping exists" -- is the dashboard relying on the old error anywhere?

crates/control-plane-api/src/server/public/graphql/storage_mappings.rs
```diff
@@ -558,59 +578,77 @@ pub type PaginatedStorageMappings = Connection<
...
+                    exact.as_str().starts_with(authorized.as_str())
+                        || authorized.starts_with(exact.as_str())
+                })
+            });
+        }
+
+        if read_prefixes.is_empty() {
+            return Ok(PaginatedStorageMappings::new(false, false));
+        }
+        if read_prefixes.len() > MAX_PREFIXES {
```
jshearer, line 647: nit: this counts surviving grants, not requested prefixes, so `exactPrefixes` spanning more than 20 distinct grants now errors where the old per-name check succeeded. Recourse exists (smaller batches) and it matches the sibling queries' cap, so fine with it -- just flagging the behavior change.

jshearer (dismissed):
All nits/pre-existing, do with them what you will

crates/control-plane-api/src/server/public/graphql/storage_mappings.rs
GregorShear: good callout, i created a ticket for this
https://github.com/estuary/flow/issues/3232

crates/control-plane-api/src/server/public/graphql/storage_mappings.rs
GregorShear: Dashboard doesn't rely on the distinction so we're safe

crates/control-plane-api/src/server/public/graphql/storage_mappings.rs
GregorShear: IMO it's imperfect, but slightly better in that it's consistent, and my hope is that it goes away soon anyway with tenant filtering

jshearer (approved):
LGTM thanks

PR #3202, opened 2026-07-19
shuffle: replay gapped producers on restart

jgraettinger:
## What

This changes *when* shuffle pays the cost of recovering an uncommitted producer span on
restart. Both the old and new behaviors are correct — the difference is an always-on cost
vs. a pay-only-when-needed one.

**Before:** a `(binding, journal)` read restarts from the *most conservative* offset — far
enough back to re-observe every producer's open span. This unconditionally replays the
content of **all** producers, even producers that are most likely dead and never coming
back. It pins restart to very early journal offsets and forces deep re-reads before the
task can resume processing — an always-on cost paid on the off chance that a seemingly-dead
producer is not, in fact, dead.

**After:** the read starts fast — resuming at the furthest position its checkpoint
justifies, the maximum offset magnitude `M` across producer entries — and replays a span
only when it actually turns out to be needed. A producer whose uncommitted span begins at
`F < M` is *gapped*: the main read skips `[F, M)`. Rather than rewind the whole read to
cover that span (the old always-on cost) or silently drop it, the gapped span is recovered
on demand, only if that producer proves it is still alive:

- `resolve_checkpoint` marks a gapped producer with the in-memory
  `max_continue == last_commit` (non-zero) sentinel (`ProducerState::is_gapped`), pinning its
  `offset` at `F`. `sequence_producer` reports the producer's first newer CONTINUE or ACK
  as a replay trigger — the signal that it is alive and its span must be recovered. A
  producer that never re-appears is simply never replayed (no cost).
- On a trigger the actor reconstructs the open span, then opens a single bounded historical
  read of `[F, trigger.begin)` (the new `Replay`, `slice/replay.rs`). Its sequenced
  documents append before everything in the heap (including the trigger); the replay
  head-of-line-blocks the Slice until it completes.
- Checkpoint durability underpins this: `build_flush_frontier` floors every read-derived
  `last_commit` at `OBSERVED_COMMIT_FLOOR`, `ProducerFrontier::reduce` breaks an
  equal-magnitude offset tie toward the uncommitted begin, and `resolve_hints` flips a
  hint-resolved span's offset to `-M` so a settled span recovers un-gapped and never
  re-replays. `max_continue` is never persisted.

Adds `shuffle_slice_replays_started/stopped` metrics and a `Drop` for `SliceActor` that
balances the started/stopped counters on teardown. New fixture scenarios (ACK- and
CONTINUE-triggered replays, cross-journal blocking, the hint-elevated offset flip, the
gapped-OUTSIDE protocol violation) and fuzz-harness open-span actions.

## Commits

- `gazette: terminate a bounded read when a fragment hole spans end_offset`
- `service-kit: accept a runtime Level in event!`
- `shuffle: factor read-failure classification and batch parsing out of the read loop`
- `shuffle: replay gapped producers on restart`

## QA — local-stack integration test (PASS)

Exercised end-to-end on a local data plane: three high-rate `source-hello-world` captures
(distinct message prefixes) all writing one collection keyed on `/message`, materialized
to Postgres on **runtime-v2**. Committed docs therefore form a contiguous gap-free cursor
sequence per capture, so a dropped document surfaces as a hole. Drove churn by deleting
the materialization's etcd shard assignment ~30× (catching it mid-transaction) plus
killing capture shards mid-flight.

- **Replays fired and did real work:** `replays_started == replays_stopped` (balanced, no
  leaks); captured `replay complete` tracks all `span_empty=false` (real documents
  recovered) with `trigger_kind=ack`; multiple distinct producers replayed, some restarts
  recovered two gapped producers at once.
- **Data integrity held exactly:** after quiescing, the materialized table equals the
  source collection per prefix (same count, same max, **zero missing cursors**); the
  source itself had zero duplicate committed cursors despite mid-transaction capture
  kills.
- **"Never re-appears" path:** a final restart with captures stopped left counts
  unchanged and produced no new replay — a gapped producer with no forthcoming trigger is
  correctly frozen, not dropped or double-counted (exactly the always-on cost this PR
  avoids).
- **Stability:** zero ERROR logs, zero panics, zero protocol violations
  (`unexpected ACK` / `OutsideWithPrecedingContinue`), zero shard crashes throughout.


crates/shuffle/src/slice/state.rs
```diff
@@ -265,15 +272,21 @@ pub fn resolve_checkpoint(
...
-            // Offset begins an uncommitted producer span.
-            min_uncommitted_begin = min_uncommitted_begin.min(offset);
+        // TODO(johnny): transitional — remove once pre-floor checkpoints have
+        // aged out post-release. Checkpoints written before `build_flush_frontier`
+        // floored read-derived entries at OBSERVED_COMMIT_FLOOR can carry
+        // `last_commit == 0` on an open span. A positive `offset` proves the
+        // entry is read-derived, so re-establish the writer's floor here;
+        // otherwise the second pass installs a zero (non-)sentinel and the
+        // gap is silently missed.
+        let last_commit = uuid::Clock::from_u64(if offset > 0 {
```
williamhbaker, line 283: Noting / wondering if this might drop spans that actually start of offset 0? Although probably rare and just a concern for the transition case.

crates/shuffle/src/frontier.rs
```diff
@@ -26,14 +46,26 @@ impl ProducerFrontier {
...
+        // clobber an accurate (ahead) `offset` of a lesser `last_commit` by
+        // replacing it with a greater hinted-and-resolved `last_commit` having
+        // a lesser, conservative `offset`.
+        //
+        // On equal magnitude, prefer the non-negative (uncommitted begin) side.
+        // A producer's next span begins exactly at its previous transaction's
+        // committed end offset whenever no other producer appended in between,
+        // so a `+F` span begin routinely ties with the prior committed `-O`
+        // (F == O) — and the span begin is the strictly newer state.
+        let offset = if self.offset.abs() != other.offset.abs() {
```
williamhbaker, line 61: Nit: I think this could be condensed down to `let offset = std::cmp::max_by_key(self.offset, other.offset, |&o| (o.abs(), o))`

williamhbaker (approved):
LGTM

crates/shuffle/src/slice/state.rs
jgraettinger: Yes, you're right, and it can happen for existing open spans at the time of transition.

Post-floor this can't happen: build_flush_frontier floors any read-derived last_commit to OBSERVED_COMMIT_FLOOR, so a genuine span beginning at offset 0 always carries last_commit == 1 and the second-pass is_span_at_head check picks it up.

The residual is purely a pre-floor checkpoint carrying {offset: 0, last_commit: 0} on a real open span — that reads as hint-only, so its [0, M) span gets skipped. It needs all of: a pre-release checkpoint, a producer whose open span begins at exactly offset 0, and another producer read past it — within the window before that task's next checkpoint flush re-derives entries with the floor.

I don't think we can close it generally: an offset == 0 entry with last_commit == 0 is genuinely ambiguous between hint-only and a real head span, and the only drop-proof handling is to treat every all-zero entry as gapped — which resurrects the always-on replay-from-0 cost for hint-only producers that this PR exists to remove. So: I'm thinking we treat it as accepted transition-window limitation, something that's exceedingly unlikely to be an issue in practice (captures are single-shard and single-producer-at-a-time), but something we'd like resolved.

PR #3203, opened 2026-07-20
runtime-next: force txn close at any usage ceiling, bypassing min duration

jgraettinger:
## What

Reworks the V2 close policy (`crates/runtime-next/src/leader/close_policy.rs`) so that breaching **any** usage ceiling forces the open transaction to close immediately, even when the minimum transaction duration has not yet elapsed.

## Why

A saturated usage measure — e.g. combiner disk usage, which captures cap at 64 MiB to favor small transactions — means an open transaction can make no further useful progress. Previously, a usage-max breach still respected the time floor (`open_duration.start`), so the transaction would idle up to `min_txn_duration` while holding its full combiner. That's a doomed transaction we've already decided to end; closing promptly reduces peak resource residency and latency.

## How

`evaluate()` is restructured around symmetric time/usage min/max predicates. The rule is now uniform: **any `.end` breach forces close and waives all `.start` floors** (both time and usage). Previously that waiver only applied when a *time* max was breached; a *usage* max still gated on the time floor. `policy_extend`, `may_close`, and `may_extend` are otherwise logically unchanged.

Note on scope: `min_txn_duration` becomes a soft batching target rather than a hard floor — a resource ceiling can now produce sub-`min_txn_duration` transactions under load. This is the intended priority (resource protection > batching preference).

## Tests

- Rewrote the `evaluate()` doc comment to explain batching intent rather than restate the logic.
- Extended the table: per-dimension usage ceilings, the last-close time floor, the max-duration ceiling with usage below floor, and a `wake_after` assertion — all previously uncovered.
- Verified capture FSM tests (the real consumer) still pass.

### DIFF
--- a/crates/runtime-next/src/leader/close_policy.rs
+++ b/crates/runtime-next/src/leader/close_policy.rs
@@ -56,22 +56,32 @@ impl Policy {
         }
     }
 
-    /// Evaluate whether an open transaction may extend, close, or hold.
+    /// Decide whether an open transaction may extend, may close, or must hold —
+    /// the batching heuristic at the heart of a V2 task.
     ///
-    /// Threshold policy with a hysteresis band per dimension:
-    /// - `policy_extend` while every measure is below its `range.end`.
-    /// - `policy_close` once every measure is above its `range.start`.
-    ///   Usage-based measures saturate below `start` once `policy_extend` is false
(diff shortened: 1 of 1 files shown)


williamhbaker (approved):
LGTM

PR #3205, opened 2026-07-20
dekaf: only read back when a fetch offset lands inside a document

mdibaiee:
**Description:**

#3150 fixed lost documents on mid-document fetch offsets by starting every new journal read up to `OFFSET_READBACK` (64MB) before the requested offset, scanning and discarding documents until reaching it. That cost is paid per read start, and consumers which frequently start new reads (e.g. SingleStore pipelines, which reconnect per batch) pay it on nearly every fetch: Dekaf's broker-side read volume per fetch request jumped from tens of KB to many MB, and delivery stalled behind the suppression scans.

This change makes the readback conditional. Collection journals are newline-delimited JSON — documents are written compactly with exactly one terminating newline, and JSON forbids unescaped control characters — so a raw newline byte occurs in journal content only as a document terminator. An offset is therefore a document boundary exactly when the preceding byte is a newline. Before opening a read, a single-byte non-blocking point read inspects that byte:

* `\n` → boundary: the read begins exactly at the fetch offset. This is the overwhelmingly common case (sequential consumers always fetch at a served document's end).
* Anything else → mid-document landing (commonly a fetch at a document's final byte, which is precisely the Kafka offset assigned to the document as a record): the read begins up to `OFFSET_READBACK` earlier, with predecessors suppressed, as before.
* Byte unavailable (offset zero, at/beyond the write head, expired fragments, journal suspended or deleted) → no containing document can be served regardless; the read begins at the fetch offset.

Stream rebuilds on token refresh run the same check. Each readback increments a new `dekaf_readback_reads{task_name, journal_name}` metric, which is also the signal to watch after deploy.

**Workflow steps:**

No user-facing changes.

None.

**Notes for reviewers:**

* An earlier revision probed by parsing (starting the stream one byte early and classifying the first parse outcome, with a UUID check to disambiguate suffixes that parse as valid JSON); review feedback simplified this to inspecting the byte directly, which costs one extra intra-data-plane roundtrip per read start but removes the probe state machine from `next_batch` entirely.
* Testing: `cargo test -p dekaf --lib` passes, and a full local `ci:dekaf-e2e` run is green — 25 passed / 0 failed. That includes #3150's tests (mid-document fetch serves the containing document; write-head fetch serves nothing), a new assertion that the mid-document fetch records a readback read, and a new `test_fetch_at_document_boundary_avoids_readback` asserting a boundary fetch serves the correct next record with zero readbacks.

### DIFF
--- a/crates/dekaf/src/read.rs
+++ b/crates/dekaf/src/read.rs
@@ -19,7 +19,7 @@ pub struct Read {
     /// Journal offset to be served by this Read.
     /// (Actual next offset may be larger if a fragment was removed).
     /// If this offset lands inside a document, the underlying journal read
-    /// restarts up to OFFSET_READBACK bytes earlier, and documents which end
+    /// begins up to OFFSET_READBACK bytes earlier, and documents which end
     /// at-or-before this offset are suppressed rather than served.
     pub(crate) offset: i64,
     /// Most-recent journal write head observed by this Read.
@@ -51,12 +51,6 @@ pub struct Read {
     deletes: DeletionMode,
 
     pub(crate) rewrite_offsets_from: Option<i64>,
-
-    // Whether the current stream's starting position is not yet verified to
-    // be a document boundary. While set, the stream's first parsed document
-    // (or parse failure) decides whether the read landed inside a document,
-    // in which case it restarts with readback. See OFFSET_READBACK.
-    probe_boundary: bool,
 
     // Leader epoch for this partition, used in RecordBatch headers.
     // Must match the epoch reported in OffsetFetch's committed_leader_epoch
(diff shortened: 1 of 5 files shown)


mdibaiee:
Verified on a local stack via `mise run ci:dekaf-e2e`: **25 passed / 0 failed**.

The first full run failed `test_fetch_mid_document_returns_containing_document` deterministically and caught a real hole in the original probe design: probing from the fetch offset itself cannot distinguish a document boundary from a fetch at a document's *final byte* (its trailing newline — exactly the Kafka record offset), because the parser silently skips separator whitespace and cleanly parses the next document. Fixed in the second commit by probing from `offset - 1`: boundary probes skip the prior document's newline and verify on the next document, while final-byte fetches read an unparseable suffix and correctly restart with readback.

crates/dekaf/src/read.rs
```diff
@@ -300,19 +364,37 @@ impl Read {
...
+                        if self.probe_boundary {
+                            // A mid-document landing may still parse, when the
+                            // document's suffix happens to be valid JSON, but
+                            // it cannot carry a valid UUID at the collection's
+                            // UUID pointer.
+                            if matches!(
+                                self.uuid_ptr.query(root.get()),
+                                Some(doc::ArchivedNode::String(uuid))
+                                    if gazette::uuid::parse_str(uuid.as_str()).is_ok()
+                            ) {
```
jshearer: I think we're making this more indirect than it needs to be. Collection journals are newline-delimited JSON (right? I think they are..), so because this read starts at `offset - 1`, can we look at that byte? `\n` means the requested offset is a document boundary, anything else means we landed inside a doc and need the readback. I suppose offset 0 is a special case as it won't begin with a newline.

jshearer (commented):
WDYT about using newlines to detect doc boundaries instead of uuid_ptr presence?

jshearer (approved):
LGTM

PR #3206, opened 2026-07-20
Allow localhost:3000 origin for the local control-plane agent

GregorShear:
Restore `http://localhost:3000` as an allowed CORS origin for the local control-plane agent. The multi-stack refactor (d1ebe24) swapped it for the per-stack `${FLOW_PORT_DASHBOARD}`, which blocks GraphQL from a UI served on the conventional port 3000. Deployed still allows 3000; the dashboard-port origin is kept too.


jgraettinger (approved):
LGTM

PR #3208, opened 2026-07-20
mise: dev-VM zone configurability (create fallback, zone-agnostic SSH, vm:move-gcp)

skord:
## Description

Dev VMs are pinned to a single GCP zone twice over: `vm:create-gcp` defaults to `us-central1-a`, and each VM's generated SSH `ProxyCommand` hardcodes that zone for the restart-on-connect path. When that zone runs out of `c4-highcpu-8` capacity (`ZONE_RESOURCE_POOL_EXHAUSTED`), idle-stopped VMs can't start again and their owners are stranded, as several folks hit this week.

Three changes:

1. **`vm:create-gcp --zone` takes an ordered comma-separated list** (default `us-central1-b,us-central1-c,us-east5-a,us-east4-a`, all verified to offer c4) and falls through to the next zone on capacity errors only. Typos, quota errors, and everything else still fail fast.
2. **The generated SSH `ProxyCommand` discovers the VM's zone at connect time** instead of baking it in, so configs never go stale when a VM moves. Costs one ~1s `gcloud instances list` per connection, amortized by the existing 20m ControlPersist.
3. **New `vm:move-gcp <name> --zone <target>`** relocates an existing VM, preserving its disk (repo checkout, bootstrap state). Machine images don't support Hyperdisk (all C4 can attach), so it follows GCP's documented snapshot -> disk -> recreate procedure. The original instance is *renamed*, not deleted, until the replacement passes an SSH check; every failure path either restores the original state or leaves both copies with printed recovery steps. Someone stuck on an exhausted zone today runs `mise run vm:move-gcp <name> --zone us-east4-a` and their existing VM (regenerated config included) follows them out.

Shared helpers live in `mise/tasks/vm/gcp-lib.sh`, sourced by both tasks (same pattern as `mise/tasks/local/reap-lib.sh`).

## Testing

Live end-to-end in `estuary-theatre` with a scratch VM (all resources deleted afterwards):

- `vm:create-gcp scratch` landed in `us-central1-b` (first zone in the new default), SSH'd via the zone-discovering ProxyCommand, full bootstrap completed.
- Rollback: planted a colliding disk in the target zone to force a mid-move failure *after* the rename; the trap renamed the instance back, restored the boot disk's auto-delete flag, deleted the snapshot, and exited non-zero with the VM intact.
- Full move `us-central1-b -> us-east4-a -> us-central1-b`; second leg completed hands-off in ~5 minutes with contents verified after each leg.
- The first move leg caught a real bug, fixed in the second commit: cloud-init regenerates SSH host keys on instance-id change, so the strict known-hosts check rejected the moved VM. Both tasks now drop the stale known-hosts entry before first connect (`accept-new` re-records over the IAP tunnel). The failure also proved the phase-4 safety: nothing was deleted, and the printed recovery manifest was followed verbatim to finish that move by hand.
- Unit-level: capacity-error regex checked against the documented GCP message variants plus quota/not-found/auth negatives; jq property extraction checked against full and sparse `instances describe` payloads; config writer checked for idempotency and write-time escaping.

Existing VMs keep working untouched; their configs heal to the zone-agnostic form on first `vm:move-gcp`.

### DIFF
--- a/mise/tasks/vm/create-gcp
+++ b/mise/tasks/vm/create-gcp
@@ -4,9 +4,11 @@ set -euo pipefail
 #MISE description="Create a development VM within GCP for a project"
 #USAGE arg "<project_name>" help="Short name of the development project (for example, 'one' or '$feature-name')"
 #USAGE flag "--project <project>" help="GCP project ID" default="estuary-theatre"
-#USAGE flag "--zone <zone>" help="Zone in which to create the VM" default="us-central1-a"
+#USAGE flag "--zone <zone>" help="Zone in which to create the VM, or an ordered comma-separated list tried in turn when zones are out of capacity" default="us-central1-b,us-central1-c,us-east5-a,us-east4-a"
 #USAGE flag "--machine-type <machine_type>" help="GCP machine type" default="c4-highcpu-8"
 #USAGE flag "--image-family <image_family>" help="GCP machine image" default="ubuntu-2404-lts-amd64"
+
+source "$(dirname "$0")/gcp-lib.sh"
 
 ACCOUNT=$(gcloud auth list --filter=status:ACTIVE --format="value(account)")
 USERNAME="${ACCOUNT%%@*}"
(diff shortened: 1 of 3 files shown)


dgreer-dev (approved):
I was one of the users affected by this -- the new process worked perfect first time, and I can confirm I have my same disk on a new machine in a new zone.

I can't swear I'm a GCP expert, but LGTM

PR #3215, opened 2026-07-21
docs/postgres: Discovery Filters

willdonnelly:
**Description:**

Copying over the docs edit from https://github.com/estuary/connectors/pull/4856


aeluce (approved):
Thanks!

PR #3216, opened 2026-07-21
runtime, flow-web, flowctl: store trigger configs using secrets overlay system

dgreer-dev:
## Summary

Moves materialization **trigger configs** onto the general `sops.overlay` secrets system and removes the bespoke HMAC field-exclusion scheme that predated it. Includes a backward-compatibility shim so existing sealed configs keep working across the deploy, making this a **non-breaking** change.

Two commits:
- **base** — the overlay migration + shape change
- **shim** — backward compatibility for pre-overlay configs (revertable on its own; see Removal below)

## The shape change

**`triggers.config` goes from a list to a name-keyed map:**

```yaml
# before
triggers:
  config:
    - url: https://example.com/hook
      headers: { Authorization: "Bearer …" }
      payloadTemplate: "{…}"

# after
triggers:
  config:
    myHook:                       # user-assigned trigger name
      url: https://example.com/hook
      headers: { Authorization: "Bearer …" }
      payloadTemplate: "{…}"
```

The map is required because the overlay merges via JSON Merge Patch (RFC 7396), which can't address array elements — a named key gives each trigger a stable, addressable identity (and nicer logs/errors: `trigger 'myHook' failed` rather than `trigger 1`).

**Secret handling now rides the shared mechanism:**
- Non-secret, user-tunable fields — `interval`, `payloadTemplate`, `timeout`, `maxAttempts` — are annotated `nonsensitive: true` and can be edited after encryption via a plaintext `sops.overlay`, without re-supplying secrets.
- Header **values** stay encrypted; `url`, `method`, and header **names** stay MAC-protected (an overlay can't redirect a request carrying a secret header).
- `url` is intentionally left MAC-protected but **not encrypted**. The only real exposure is a credential-in-URL-path webhook (e.g. Slack); anything authenticating via a header is already fully covered. Deferred as a deliberate, low-risk gap.

**Removed:** the `strip_hmac_excluded_fields` / `restore_hmac_excluded_fields` machinery from the live path, across `models`, `flowctl`, the `flow-web` WASM bindings, and both runtimes. Both runtimes decrypt new configs through the same `unseal::overlay::decrypt_with_overlay` path used for endpoint configs. (The deleted `flow-web` trigger bindings had no consumer — the dashboard has no trigger UI — so there is no frontend change.)

## Backward compatibility

Pre-overlay configs stored `config` as a list *and* used the old HMAC-exclusion MAC scheme, so they can't be decrypted by the overlay path and can't be structurally reshaped in place (SOPS binds ciphertext to its path). The shim handles them:

- `Triggers.config` is a shape-preserving `TriggerConfigs` enum. A legacy list round-trips **byte-faithfully** through the control plane, preserving its sealed SOPS structure and MAC; its JSON Schema still advertises only the map form.
- The runtimes detect a list-shaped `config` and decrypt it via the retained strip/restore path (isolated behind `LegacyTriggers`); new configs use the overlay path.
- Validation runs uniformly over both shapes.

New publishes always produce the current (map) form — the shim is read-only compatibility.

## Operations & deployment

Because of the shim, this is **safe to deploy without coordinating config migrations first** — existing configs keep decrypting and firing on updated reactors, for as long as needed.

**1. Deploy anytime.** No hard ordering against the config migration; legacy configs tolerate the (unknown-length) reactor rollout window.

**2. Migrate existing configs as post-deploy cleanup (no customer action).** Once reactors are confirmed updated, re-publish the pre-overlay configs into the map form. Identify them with:

```sql
SELECT catalog_name, updated_at, data_plane_id
FROM live_specs
WHERE spec_type = 'materialization' AND spec->'triggers' IS NOT NULL;
```

Re-publish is done on our side (we hold the decryption keys); the one hard rule is not to publish a new-format config until all reactors are updated (old reactors can't parse it).

**3. Remove the shim once no legacy configs remain.** Confirm via the census above, then revert the shim commit (the in-code `// Remove once no legacy configs remain` markers are the fallback if the revert conflicts).

**Stale `flowctl` clients** emit the old list shape, which is still accepted, so they continue to publish successfully.


dgreer-dev:
Moving to a draft as we re-consider this as a breaking change

dgreer-dev:
On further chat w/ Will Baker we opted to leave this change until a more holistic change in triggers (especially wrt config) can be done.

PR #3218, opened 2026-07-21
runtime-next: converge startup reconciliation via rescan Persists

### LINKED ISSUE #3176: runtime-v2: ahead-remote-authoritative recovery promotes committed_close in memory only

jgraettinger:
## Summary

Reworks runtime-next leader **startup reconciliation** from a one-shot cleanup Persist into an iterative convergence loop, so a leader's startup state is a projection of the actually-scanned durable state rather than of an in-memory reconstruction — the leader can never diverge from disk.

The reconcile policy is factored into a pure, IO-free `reconcile` function (one each for derive/materialize) expressed as an ordered sequence of triggered, self-clearing steps:
- convert a V1-written legacy checkpoint,
- drop a V1 rollback,
- fold a hinted commit the connector confirms,
- adopt a connector checkpoint that committed ahead.

`reconcile_loop` repeatedly reconciles the scanned RocksDB `Baseline` against its authoritative checkpoints, applying each step as a `rescan` Persist and re-reading the freshly written durable state before the next iteration.

### Bug fix: crash-loop in the remote-authoritative connector case

The old materialize reconciliation applied the hinted-commit fold purely in memory — it advanced the session's `committed_close`/committed Frontier but emitted no Persist, leaving RocksDB's `committed_close` at its pre-fold value. A session could fold a hinted commit (in memory), write the *next* hint via StartCommit (bumping on-disk `hinted_close`), then crash before that transaction committed. Recovery then saw a stale `committed_close`, a newer `hinted_close`, and an endpoint clock matching neither — hitting the `reconcile_recovered` mismatch `bail!` on every recovery, crash-looping the shard.

Routing the fold through `reconcile_loop` as a `rescan` Persist makes the advance durable, so a later next-hint crash recovers on the in-sync branch and never bails. Covered by the `reconcile_fold_then_next_hint_matches_committed_close` regression test.

## Commits

- **flow.schema.json: regenerate and enforce up-to-date in CI** — fix drifted schema; add `ci:flow-schema-check` gating future model changes.
- **runtime-next: RocksDB::scan takes binding-index-ordered state keys** — `scan` builds its own sorted index internally; becomes a sync fn returning a lifetime-free Future.
- **runtime: add Persist.rescan field** — when set, shard zero re-scans RocksDB after applying the WriteBatch and replies with a fresh `Recover` instead of a `Persisted` echo. Regenerates Go/Rust protobuf.
- **runtime-next: extract shared leader test fixtures** — pure refactor pulling leader test builders into a `leader::fixtures` module.
- **runtime-next: converge startup reconciliation via rescan Persists** — the reconcile loop and crash-loop fix above.

## QA: local-stack soak

Ran a chaos soak on a local stack (runtime built from this branch) using the `tests/soak/` harness — capture → `accounts` derivation → `views` (materialize-postgres, **remote-authoritative**, the fold target) + `ledger` (recovery-log-authoritative). Shards were force-recovered by repeatedly deleting their Etcd assignment keys under continuous load (graceful cancel → reassign → fresh recovery + startup reconciliation).

**Reproduced the crash-loop scenario and confirmed the fix.** With defaults the fold window is too narrow to sample (65 random kills → 0 folds, all in-sync recoveries clean). Capping `maxTxnDuration` to force frequent commits then made the leader log land in the fold window repeatedly:

```
leader.materialize … connector checkpoint matches hinted_close; folding the hinted delta as committed
    committed_close=…:07.909Z  hinted_close=…:07.980Z      # endpoint committed the hint, then crashed
leader … persisting reconciled startup baseline and re-scanning  iteration=1
```

That second line is the fix: the fold is now persisted durably via a `rescan` Persist rather than applied in memory.

**Results across the run:**
- **22 fold events, 22 matching rescan-persists** — every fold recovered cleanly; **zero** `did not converge` / `committed_close`-mismatch bails / crash-loops.
- **309 `views` recoveries** (plus `ledger`/`accounts`/`source` recoveries); all shards stayed live.
- **All 6 `views` at-rest SQL oracles = 0** (conservation std+delta, oracle-match std+delta, sentinels, std-vs-delta), full account population materialized.
- `ledger` in-flight self-checks: **no** conservation/oracle/seq/sentinel violations (its only ERRORs were `expected Stop or CloseNow … unexpected EOF`, i.e. the kill severing the gRPC stream). `accounts`: only the expected schema-widening churn.

**Unrelated finding (not this change):** aggressive shard cancellation can panic `crates/async-process/src/lib.rs:49` (`handle.await.unwrap()` → `JoinError::Cancelled`) reaping a connector subprocess — rare (1/135 kills), non-fatal (shard recovers). Pre-existing connector-lifecycle plumbing; being fixed separately.

## Test plan

Covered by PR CI (passing):
- [x] `cargo test -p runtime-next` (includes the `reconcile_*` unit tests / `reconcile_fold_then_next_hint_matches_committed_close` regression)
- [x] `cargo test -p proto-flow`
- [x] CI `ci:flow-schema-check`
- [x] Local-stack chaos soak (above)


williamhbaker (approved):
LGTM

Something I realized when thinking about this: A remote authoritative connector that also uses the driver checkpoint mechanism for persisting state in the recovery log might lose its driver checkpoint state if it commits the runtime checkpoint in the endpoint and then crashes before the recovery log commit. Then upon recovery, it would resume from its own endpoint checkpoint. I think this is necessarily the way it has to be, and existed in runtime V1 as well.

PR #3225, opened 2026-07-22
async-process: handle reaping-task JoinError in Child::wait() without panicking

jgraettinger:
## Problem

`Child::wait()` did `self.status = Err(handle.await.unwrap())`, which panics whenever the reaping `spawn_blocking` task's join resolves to an error — `JoinError::Cancelled` (the task's runtime was torn down) or `JoinError::Panic`.

This was observed once during an aggressive shard-cancellation soak (repeatedly deleting a materialize shard's Etcd assignment key under load): a runtime-next shard context was cancelled while a `Child::wait()` was outstanding, so the reaping `spawn_blocking` task was cancelled and the `.unwrap()` panicked:

```
panicked at crates/async-process/src/lib.rs:49:44:
called `Result::unwrap()` on an `Err` value: JoinError::Cancelled(Id(...))
```

It was non-fatal — the reactor process survived and the shard recovered immediately — but a panic on a routine cancellation path is a latent robustness bug. This is pre-existing connector-subprocess lifecycle code, unrelated to the runtime-next frontier-persist work under which it surfaced.

## Fix

Replace the ad-hoc `Result<JoinHandle, Child>` status field with an explicit three-state machine:

- **`Reaping(handle)`** — the reaping task is running.
- **`Reaped(child)`** — the child was reaped; exit status is queryable.
- **`Abandoned`** — the reaping task did not complete (join cancelled or panicked).

On a `JoinError`, `wait()` transitions to `Abandoned` and returns an `io::Error` instead of unwrapping; it is idempotent on repeated calls. An `Abandoned` child was **not** reaped by us, so `is_finished()` reports `false` and `Drop` still `SIGKILL`s it via its pid. (`Reaped` reports `true`, correctly suppressing the kill — reaping the pid makes it eligible for reuse, so blindly killing it would risk an unrelated process.)

## Tests

Two regression tests, one per `JoinError` arm. Note that runtime teardown does **not** reliably fail a running `spawn_blocking` handle (it is detached and runs to completion, not cancelled), so the tests instead drive a real (still-running) `sleep` child through a hand-built reaping `JoinHandle` forced to fail each way — an aborted pending task for `Cancelled`, a panicking closure for `Panic`. Each verifies `wait()` errors gracefully (no panic), is idempotent, reports `!is_finished()`, and that `Drop` actually `SIGKILL`s the process (confirmed and reaped via `waitpid`, leaving no zombie).

`cargo test -p async-process`, `cargo fmt`, and `cargo clippy -p async-process --tests` are all clean.

### DIFF
--- a/crates/async-process/src/lib.rs
+++ b/crates/async-process/src/lib.rs
@@ -8,11 +8,28 @@ use std::os::fd::OwnedHandle as OwnedImpl;
 
 pub struct Child {
     pid: libc::pid_t,
-    status: Result<tokio::task::JoinHandle<std::process::Child>, std::process::Child>,
+    status: Status,
 
     pub stdin: Option<ChildStdio>,
     pub stdout: Option<ChildStdio>,
     pub stderr: Option<ChildStdio>,
+}
+
+/// State machine for reaping the child process.
+///
+/// A `spawn_blocking` task reaps the child by blocking in its `wait()`.
+/// `Status` tracks whether that task is still running, has completed (yielding
+/// the reaped `std::process::Child`), or was abandoned because its runtime was
+/// torn down before it could reap.
+enum Status {
+    /// The `spawn_blocking` reaping task is running; it yields the reaped child.
+    Reaping(tokio::task::JoinHandle<std::process::Child>),
+    /// The child has been reaped and its exit status can be queried.
+    Reaped(std::process::Child),
+    /// The reaping task did not complete: its runtime was torn down (join was
+    /// cancelled) or it panicked. The child was not reaped by us, so `Drop`
+    /// must SIGKILL *and* reap it (nobody else will) to avoid a zombie.
+    Abandoned,
 }
 
 pub type ChildStdio = tokio::fs::File;
@@ -24,7 +41,7 @@ impl From<std::process::Child> for Child {
         let stderr = map_stdio(inner.stderr.take());
 
         let pid = inner.id() as libc::pid_t;
-        let status = Ok(tokio::runtime::Handle::current().spawn_blocking(move || {
+        let status = Status::Reaping(tokio::runtime::Handle::current().spawn_blocking(move || {
             _ = inner.wait();
             inner
         }));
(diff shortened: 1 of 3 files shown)


crates/async-process/src/lib.rs
```diff
@@ -8,13 +8,30 @@ use std::os::fd::OwnedHandle as OwnedImpl;
...
+/// torn down before it could reap.
+enum Status {
+    /// The `spawn_blocking` reaping task is running; it yields the reaped child.
+    Reaping(tokio::task::JoinHandle<std::process::Child>),
+    /// The child has been reaped and its exit status can be queried.
+    Reaped(std::process::Child),
+    /// The reaping task did not complete: its runtime was torn down (join was
+    /// cancelled) or it panicked. The child was not reaped by us, so `Drop`
+    /// must still SIGKILL it.
+    Abandoned,
```
dgreer-dev, line 32: Do we need to `waitpid` in these cases or something?  This case seems like it could lead to a zombie process

crates/async-process/src/lib.rs
```diff
@@ -45,19 +62,36 @@ impl Child {
...
-            unreachable!()
-        };
-        inner.wait()
     }
 
     pub fn is_finished(&mut self) -> bool {
         match &self.status {
-            Ok(handle) => handle.is_finished(),
-            Err(_inner) => true,
+            Status::Reaping(handle) => handle.is_finished(),
```
dgreer-dev: `handle.is_finished()` is true for a cancelled task too, I think?  It just means "terminated," not "successfully reaped." So if the reaping task is *cancelled* and this Child is dropped without calling wait(), `Drop` concludes the child already exited and skips the SIGKILL, leaking a live process.

dgreer-dev (dismissed):
Looks good, although I have a few small comments.  Good tests, and I think a good improvement for sure.

crates/async-process/src/lib.rs
jgraettinger: Good catch — yes. Fixed in 2fb0c69: on the abandoned path `Drop` now `waitpid`s to reap the child, not just `SIGKILL`. Since `SIGKILL` is uncatchable the reap blocks only briefly. `ECHILD` (a racing reaping task already reaped it) is treated as success. When the reaping task is still running I still let *it* do the `waitpid` rather than racing it.

crates/async-process/src/lib.rs
jgraettinger: Exactly right, and this is the more serious of the two — the runtime/runtime-next container guards drop the `Child` without ever calling `wait()`, so a cancelled reaping task would leak a live process. Fixed in 2fb0c69: the reaping task sets an explicit `reaped: Arc<AtomicBool>` after its `wait()` returns, and `is_finished()`/`Drop` consult that instead of `handle.is_finished()`, so "terminated but never reaped" is no longer mistaken for "exited". Added `test_drop_without_wait_reaps_cancelled_child` covering this path. One honest caveat: a pre-existing narrow race remains (the pid can be reaped inside `inner.wait()` a moment before the flag/handle observably reflects it); this change narrows rather than widens it, and I did not attempt to fully close it here.

dgreer-dev (dismissed):
LGTM

jgraettinger:
Force-pushed a reworked version. The two commits (original + the review follow-up that added `kill_and_reap` in `Drop`) are now squashed into a single commit, and the mechanism has been simplified:

- **Dropped the `reaped: Arc<AtomicBool>` signal.** The previous approach had the reaping `spawn_blocking` task set an atomic flag after `wait()` returned, so `Drop`/`is_finished()` could tell "already reaped" from "terminated without reaping" without being able to `.await` the handle.

- **Replaced it with `try_resolve()`**, which polls the `JoinHandle` non-blockingly via `futures::FutureExt::now_or_never`. A `JoinHandle` is pollable outside a runtime, so this is safe to call from `Drop`. It advances `Reaping` → `Reaped` (task yielded the reaped child) or `Reaping` → `Abandoned` (task cancelled/panicked without reaping). Both `is_finished()` and `Drop` now just call `try_resolve()` and match on the resulting state, so there's a single source of truth for the child's reaping state rather than a handle plus a side-channel flag. Adds a `futures` dependency for `now_or_never`.

- **`Drop` dispatch is unchanged in behavior:** `Reaped` → leave the pid alone (may be recycled); `Reaping` (task still blocked in `wait()`) → SIGKILL only and let it reap; `Abandoned` → SIGKILL *and* `waitpid` here, since nobody else will.

- **Tests:** extracted the duplicated cancelled-reaper setup into a `spawn_cancelled_reaper()` helper; the drop-without-wait test now asserts via `is_finished()` instead of reaching into the removed `reaped` field. Coverage is otherwise the same (both `JoinError` arms, plus the drop-without-wait path).

No change to the final observable behavior — this is a cleaner internal state model for the same fix.

dgreer-dev (approved):
No real changes so still LGTM!

PR #3227, opened 2026-07-22
control-plane-api: add a root effectiveAlertConfig(catalogPrefixOrName:) query

GregorShear:
## Summary

Adds a top-level `effectiveAlertConfig(catalogPrefixOrName:)` GraphQL query that resolves the effective alert config at a single prefix or catalog name — merging all ancestor prefix layers and controller defaults, with per-field provenance.

## Why a root query

Master already exposes effective config as *fields* (`AlertConfigEntry.effective`, `LiveSpec.effectiveAlertConfig`), which only answer for scopes that have an entry or a live spec. This root query takes an **arbitrary** `catalogPrefixOrName` and does not require an `alert_configs` row to exist there, so it answers for prefixes and names that carry no explicit config and are therefore absent from the `alertConfigs` listing.

It reuses the existing `resolve_effective_alert_config` helper and its provenance output; auth requires `CatalogRead` on the scope, matching the field-level accessors.

## Input validation

`validate_prefix_or_name` now rejects a slash-less token like `aliceCo`. Such a token is neither a prefix (no trailing `/`) nor a real catalog name (names are always hierarchical), so it previously validated as a bare `Name` and then failed authorization with a confusing `PermissionDenied` — a grant at `aliceCo/` never covers the bare string `aliceCo`. It now returns a validation error pointing at the trailing-slash prefix form, matching the slash requirement the mutation path already enforces via `governing_prefix`.

## Testing

`test_effective_alert_config_query` covers:
- defaults-only resolution (no row present; provenance `source: null`)
- an inherited prefix override with provenance attribution
- the exact-catalog-name layer (`aliceCo/in/capture-foo`)
- the slash-less validation rejection
- read-denied

SDL regenerated.

### DIFF
--- a/crates/control-plane-api/src/server/public/graphql/alert_configs.rs
+++ b/crates/control-plane-api/src/server/public/graphql/alert_configs.rs
@@ -207,6 +207,32 @@ impl AlertConfigsQuery {
         )
         .await
     }
+
+    /// Resolves the effective alert config at a single prefix or catalog name.
+    pub async fn effective_alert_config(
+        &self,
+        ctx: &Context<'_>,
+        catalog_prefix_or_name: String,
+    ) -> async_graphql::Result<EffectiveAlertConfig> {
+        let env = ctx.data::<crate::Envelope>()?;
+        let claims = env.claims()?;
+
+        validate_prefix_or_name(&catalog_prefix_or_name)?;
+
+        // Reading a scope's effective config requires catalog-read access at
+        // that scope. Ancestor layers merged into the result are visible to
+        // anyone who can read the scope, matching the `effective` field on
+        // AlertConfigEntry and `effectiveAlertConfig` on liveSpec.
+        let policy_result = crate::server::evaluate_names_authorization(
+            env.snapshot(),
+            claims,
+            models::authz::Capability::CatalogRead,
+            [catalog_prefix_or_name.as_str()],
+        );
+        env.authorization_outcome(policy_result).await?;
+
+        resolve_effective_alert_config(ctx, &catalog_prefix_or_name).await
+    }
 }
 
 #[derive(Debug, Default)]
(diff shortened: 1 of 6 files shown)


crates/control-plane-api/src/server/public/graphql/filters.rs
```diff
@@ -13,5 +13,8 @@ pub struct DateFilter {
 
 #[derive(Debug, Clone, Default, async_graphql::InputObject)]
 pub struct PrefixFilter {
+    /// Match rows whose value is prefixed by this string.
     pub starts_with: Option<String>,
+    /// Match the single row whose value equals this string exactly.
```
GregorShear: i'm not sure why this has to be a _single_ row - could be multiple rows that have the same prefix, right?

crates/control-plane-api/src/server/public/graphql/alert_configs.rs
```diff
@@ -218,6 +218,39 @@ impl AlertConfigsQuery {
...
+        validate_prefix_or_name(&catalog_prefix_or_name)?;
+
+        // Reading a scope's effective config requires read access at that
+        // scope. Ancestor layers merged into the result are visible to anyone
+        // who can read the scope, matching the `effective` field on
+        // AlertConfigEntry and `effectiveAlertConfig` on liveSpec.
+        let policy_result = crate::server::evaluate_names_authorization(
+            env.snapshot(),
+            claims,
+            models::Capability::Read,
```
GregorShear: could we use the capability_bit instead of legacy capability?

crates/flow-client/control-plane-api.graphql
```diff
@@ -1798,12 +1810,23 @@ type QueryRoot {
...
 	alertConfigs(filter: AlertConfigsFilter, after: String, first: Int): AlertConfigEntryConnection!
 	"""
+	Resolves the effective alert config at a single prefix or catalog name,
+	merging all ancestor prefix layers and controller defaults.
+	
+	This does not require an `alert_configs` row to exist at the scope, so
+	it works for prefixes and names that carry no explicit config and are
+	therefore absent from the `alertConfigs` listing. `catalogPrefixOrName`
+	is a prefix ending in `/` or an exact catalog name; the caller must have
+	`Read` capability on it.
```
jshearer: Stale SDL -- the Rust doc comment says `CatalogRead`, so this was generated before that edit. Regenerating (`cargo build -p flow-client --features generate`) produces a one-line diff here, and `ci:graphql-schema-check` will fail once checks run on this branch (none have yet).

crates/control-plane-api/src/server/public/graphql/alert_configs.rs
```diff
@@ -204,6 +218,39 @@ impl AlertConfigsQuery {
...
+    /// `CatalogRead` capability on it.
+    pub async fn effective_alert_config(
+        &self,
+        ctx: &Context<'_>,
+        catalog_prefix_or_name: String,
+    ) -> async_graphql::Result<EffectiveAlertConfig> {
+        let env = ctx.data::<crate::Envelope>()?;
+        let claims = env.claims()?;
+
+        validate_prefix_or_name(&catalog_prefix_or_name)?;
```
jshearer, line 220: nit: A single-token input like `aliceCo` (no trailing slash) passes validation as a `Name`, then fails authorization -- a grant at `aliceCo/` doesn't cover the bare string `aliceCo` -- so a caller who drops the trailing slash gets `PermissionDenied` even as tenant admin, when the actual problem is the input. 

Real catalog names always contain a `/`, and `governing_prefix` below already rejects slash-less names on the mutation path; doing the same here would turn that denial into a validation error.

crates/control-plane-api/src/server/public/graphql/alert_configs.rs
```diff
@@ -578,4 +625,107 @@ mod test {
             .await;
         insta::assert_json_snapshot!("effective_defaults_plus_prefix_override", response);
     }
+
+    #[sqlx::test(
+        migrations = "../../supabase/migrations",
+        fixtures(path = "../../../fixtures", scripts("data_planes", "alice"))
+    )]
+    async fn test_effective_alert_config_query(pool: sqlx::PgPool) {
```
jshearer, line 894: Coverage here is prefix-only. A catalog *name* is the other half of the argument's domain and exercises the exact-name layer of `ancestor_prefixes_and_name` -- the alice fixture already has `aliceCo/in/capture-foo`, so one more query + snapshot pins it. An invalid-input case would also pin the `validate_prefix_or_name` rejection path, which no test hits.

jshearer (dismissed):
LGTM, left a couple nits -- need to rebase ontop of master once the base branch is merged, regenerate graphql schema etc. Plus a test coverage nit and an error message nit

Let me know once you rebase/regen gql and I'll re-approve

crates/control-plane-api/src/server/public/graphql/alert_configs.rs
GregorShear: returning a more descriptive error with "did you mean `aliceCo/` ?" now

jshearer (approved):

PR #3233, opened 2026-07-23
docs: clarify sync schedule connector scope and backfill heuristic

jwhartley:
## What

Clarifies the [Materialization Sync Schedule](https://docs.estuary.dev/reference/materialization-sync-schedule/) page on two points that currently cause confusion:

1. **Which connectors support a sync schedule.** The page previously implied this is a warehouse-only feature. In fact the same acknowledgement-delay mechanism is also used by the file and object-store connectors (S3/GCS/Azure Blob file materializations and Iceberg), which route their batching/upload interval through the same `syncFrequency`. Transactional-database materializations (Postgres, MySQL, SQL Server) and streaming/API destinations do **not** use it. The page now lists all three groups.

2. **How the "backfills always run as fast as possible" behavior actually works.** Added a short paragraph describing the heuristic: if any of the last several transactions stored a large number of documents (on the order of a million), the connector treats itself as still backfilling and skips the sync-schedule delay; once several consecutive transactions are all below that size it is treated as caught up. This is the same across every connector that supports a sync schedule.

## Why

A support question asked whether the "1M record skip" applies to all materializations or only database/warehouse ones. The docs didn't answer it, and an inference from the current wording is that file-based materializations use a separate mechanism, which is wrong: file sinks share the exact same code path.

## Source of truth

`go/materialize/transactions_stream.go` in `estuary/connectors` (`storeBackfillThreshold = 1_000_000`, `storedHistorySize = 5`), and the `AckSchedule` wiring in each connector's driver. The connectors that set `AckSchedule` are exactly: Snowflake, BigQuery, Databricks, Redshift, ClickHouse, MotherDuck, Azure Fabric Warehouse, filesink (S3/GCS file sinks), S3 Iceberg, and Iceberg.

Docs owner: Emily (@aeluce).

### DIFF
--- a/site/docs/guides/customize-dataflows/materialization-sync-schedule.md
+++ b/site/docs/guides/customize-dataflows/materialization-sync-schedule.md
@@ -18,7 +18,12 @@ reduce the costs incurred in the destination from the actions the connector
 takes to load data to it.
 :::
 
-Most warehouse materialization connectors (such as [Snowflake](/reference/Connectors/materialization-connectors/Snowflake), [Databricks](/reference/Connectors/materialization-connectors/databricks), and [BigQuery](/reference/Connectors/materialization-connectors/BigQuery)) support configuring a sync schedule.
+Sync schedules are supported by two kinds of materialization connectors:
+
+- Warehouse connectors, such as [Snowflake](/reference/Connectors/materialization-connectors/Snowflake), [Databricks](/reference/Connectors/materialization-connectors/databricks), and [BigQuery](/reference/Connectors/materialization-connectors/BigQuery), as well as Amazon Redshift, MotherDuck, ClickHouse, and Azure Fabric Warehouse.
+- File and object-store connectors that write data in batches, such as the Amazon S3, Google Cloud Storage, and Azure Blob Storage file materializations, and the Apache Iceberg materializations. For these connectors the batching interval (for example, an upload interval) is the sync frequency.
+
+Transactional database materializations (such as PostgreSQL, MySQL, and SQL Server) and streaming or API destinations (such as Elasticsearch, MongoDB, DynamoDB, and Pinecone) do _not_ use a sync schedule; they apply updates as they arrive.
 Check the connector reference docs to determine if a specific connector supports sync schedules.
 
 ## How transactions are used to sync data to a destination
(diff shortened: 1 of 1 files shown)


site/docs/guides/customize-dataflows/materialization-sync-schedule.md
```diff
@@ -18,7 +18,12 @@ reduce the costs incurred in the destination from the actions the connector
 takes to load data to it.
 :::
 
-Most warehouse materialization connectors (such as [Snowflake](/reference/Connectors/materialization-connectors/Snowflake), [Databricks](/reference/Connectors/materialization-connectors/databricks), and [BigQuery](/reference/Connectors/materialization-connectors/BigQuery)) support configuring a sync schedule.
+Sync schedules are supported by two kinds of materialization connectors:
+
+- Warehouse connectors, such as [Snowflake](/reference/Connectors/materialization-connectors/Snowflake), [Databricks](/reference/Connectors/materialization-connectors/databricks), and [BigQuery](/reference/Connectors/materialization-connectors/BigQuery), as well as Amazon Redshift, MotherDuck, ClickHouse, and Azure Fabric Warehouse.
+- File and object-store connectors that write data in batches, such as the Amazon S3, Google Cloud Storage, and Azure Blob Storage file materializations, and the Apache Iceberg materializations. For these connectors the batching interval (for example, an upload interval) is the sync frequency.
```
aeluce: It may be how we name things, but the intention was to note which connectors support the sync schedule _stanza_ since we're not including all the bundled properties (`syncFrequency`, `fastSyncEnabledDays`, etc) in every connector reference page that supports it. The sole `uploadInterval` property, in comparison, is included on the file store connector reference pages, and these connectors do not support `syncSchedule`.

So, this proposed update may be leaning more on how we treat the implementation on the backend rather than the user-facing experience. I'm happy to take a stab at rewriting if you agree with that assessment.

jwhartley: Right, yeah it would be good to target how to configure the sync behaviour you want, vs what does/doesn't support syncFrequency specifically. Have at it @aeluce !

aeluce (approved):
Whoops, sorry for the delay. This look good to you?

As a side note, sounds like this page will require a larger overhaul very soon as we move sync schedules to the runtime rather than connector-level, so I'm keeping my eye on that.

PR #3234, opened 2026-07-23
Gate invite-link private-DP grants on an existing private data plane

GregorShear:
Redeeming an admin invite link called `ensure_private_data_plane_grants`, which inserted `ops/dp/private/<tenant>/` and `ops/tasks/private/<tenant>/` read grants **unconditionally** — it never checked whether the tenant actually owns a private data plane. Provisioning gates the identical grants behind `if let Some(private)`; the invite path did not.

For a tenant that never had a private data plane, this fabricated grants to private-DP/ops-tasks prefixes that don't exist, and because the insert runs on every admin redeem with `ON CONFLICT DO NOTHING`, the grants reappeared after being deleted (the "zombie grants" reported by affected tenants).

**Fix:** gate the insert on the existence of a `data_planes` row named `ops/dp/private/<tenant>/<region>`, matched with the `^@` prefix operator (not `LIKE`, since tenant names may contain `_`). When no private data plane exists, install nothing.

**Tests:** the positive regression test now loads the `private_links` fixture (its source of a private DP for `aliceCo`); added a negative test asserting an admin redeem installs no grants for a tenant without a private data plane.


jshearer (approved):
LGTM!

PR #3239, opened 2026-07-23
control-plane-api: fix storage-mapping suffix strip to respect path-segment boundaries

GregorShear:
The `collection-data/` suffix transform in `collection_and_recovery_spec_from` used raw string matching (`strip_suffix` / `ends_with`), so a user-chosen store prefix that happens to end with the characters `collection-data/` **without** a preceding `/` boundary — e.g. `estuary-collection-data/` — was mishandled in both directions:

- **Recovery spec (strip):** the entire tail was chopped, mangling `estuary-collection-data/` into `estuary-`, which is not a valid prefix. This is the customer-reported symptom.
- **Collection spec (append):** the managed segment was skipped, so collection data would be written directly under the user's prefix instead of under a nested `collection-data/`.

## Fix

Recognize the segment only on a `/` boundary — the prefix is exactly `collection-data/`, or the matched suffix is preceded by `/` — via a new `ends_with_collection_data_segment` helper used by both the append and strip paths.

## Tests

- `test_split_ignores_non_boundary_suffix_match` — regression for `estuary-collection-data/` (collection gets the segment appended; recovery is left intact).
- `test_split_strips_nested_boundary_suffix` — segment recognized when nested under a further prefix.
- `test_ends_with_collection_data_segment` — direct unit coverage of the boundary helper.

All `control-plane-api` `storage_mappings` tests pass (8 passed).


skord (approved):
LGTM

PR #3240, opened 2026-07-23
control-plane-api: add `id` and an `id` filter to the `dataPlanes` GraphQL query

GregorShear:
Return `dataPlane.id`, and support filtering by id in the data plane list query.

```graphql
dataPlanes(filter: { id: { in: [$dataPlaneId] } }) {
  edges { node { id name reactorAddress region cloudProvider isPublic } }
}
```


jgraettinger (approved):
LGTM

PR #3241, opened 2026-07-23
control-plane-api: coalesce onboard storage-mapping data_planes to empty array

GregorShear:
## Problem

Tenant onboarding builds the tenant-root storage mapping's `data_planes` from `json_agg` over public data planes. `json_agg` returns SQL `NULL` over an empty set, so a tenant provisioned when no `ops/dp/public/` data plane matches gets a spec of `{"stores": [...], "data_planes": null}`.

That value fails to deserialize into `StorageDef`: its `data_planes: Vec<String>` field is `#[serde(default)]`, which tolerates a *missing* key but not an explicit `null`. The result is that `storageMappings` (and any other read of that spec) errors with `invalid type: null, expected a sequence` for the affected prefix.

This reliably shows up in local stacks, where the public data-plane binding isn't present at onboarding time.

## Fix

Coalesce the aggregate to `'[]'::json` so `data_planes` is always a JSON array.

Existing rows already written with `data_planes: null` still need a one-time `UPDATE ... SET spec = jsonb_set(...)` (or re-onboard); this change prevents new occurrences.


GregorShear:
@jshearer ding dong - testing notifications

PR #3242, opened 2026-07-23
Extend `PrefixFilter` with `in` and add storage mapping query as a consumer

GregorShear:
## Summary

Extends the shared `PrefixFilter` GraphQL input with an exact-set `in` predicate — an alternative to the existing `startsWith` subtree match (the two are mutually exclusive), letting prefix-scoped queries filter by an explicit set of prefixes. `storageMappings` gains a composable `filter` argument built on it, matching the pattern already used by `inviteLinks` (and the in-flight `dataPlanes` filter).

Groundwork for moving the UI's storage-mappings table (estuary/ui#2031) and flowctl onto the shared filter pattern.

## `PrefixFilter`

Two scoping modes, mutually exclusive across every consumer:

```graphql
input PrefixFilter {
  startsWith: String     # subtree match: acmeCo/ matches acmeCo/, acmeCo/team/, ...
  in: [String!]          # exact-set match: only the listed prefixes
}
```

For example, scope a query to an explicit set of prefixes (intersected with what the caller can read):

```graphql
storageMappings(filter: { catalogPrefix: { in: ["acmeCo/", "betaCo/"] } })
```

- A prefix scope is always either a subtree (`startsWith`) or an exact set (`in`), never a mix; supplying both is a request error.
- An empty `in: []` is rejected at input validation, rather than ambiguously meaning "nothing" or "everything".
- All three consumers — `storageMappings`, `alertConfigs`, `inviteLinks` — resolve their scope through one combinator, `filtered_authorized_prefixes`, which chains the mutual-exclusion check (`PrefixFilter::into_parts`), the authorized-prefix lookup, and the exact-set narrowing (`PrefixFilter::narrow_to_exact_set`). The narrow-only invariant — a filter can only remove authorized prefixes, never add them — has a single owner.

## Deprecating `storageMappings(by:)`

`storageMappings` gains `filter: { catalogPrefix }`, deprecating `by` in its favor (`underPrefix → startsWith`, `exactPrefixes → in`). `by` maps onto the same `PrefixFilter` internally and stays until flowctl migrates (#3238).

## Stack / follow-ups

- flowctl (#3238): replace `by: { exactPrefixes }` with `filter: { catalogPrefix: { in } }`, stacked on this PR.
- UI (estuary/ui#2031): switch the storage-mappings table to `filter: { catalogPrefix: { startsWith } }` after a codegen refresh.

### DIFF
--- a/crates/control-plane-api/src/server/public/graphql/alert_configs.rs
+++ b/crates/control-plane-api/src/server/public/graphql/alert_configs.rs
@@ -15,7 +15,9 @@ const DEFAULT_PAGE_SIZE: usize = 50;
 const MAX_PREFIXES: usize = 20;
 
 /// Optional filter for the `alertConfigs` query. When omitted, all accessible
-/// rows are returned.
+/// rows are returned. A filter only narrows those results; the caller's
+/// catalog-read scope is enforced independently, so it can never widen what a
+/// caller may see.
 #[derive(Debug, Clone, Default, async_graphql::InputObject)]
 pub struct AlertConfigsFilter {
     /// Filter on the `catalog_prefix_or_name` column.
@@ -108,9 +110,9 @@ impl AlertConfigsQuery {
     /// Lists alert-config rows visible to the caller.
     ///
     /// Results are limited to readable prefixes and sorted by
-    /// `catalog_prefix_or_name`. `filter.catalogPrefixOrName.startsWith` can
-    /// narrow the results further. Passing a full catalog name returns at
-    /// most one exact-name row.
+    /// `catalog_prefix_or_name`. `filter.catalogPrefixOrName` narrows further,
+    /// by subtree (`startsWith`) or an exact set (`in`) — not both. Passing a
+    /// full catalog name returns at most one exact-name row.
     pub async fn alert_configs(
         &self,
         ctx: &Context<'_>,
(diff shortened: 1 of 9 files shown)


crates/control-plane-api/src/server/public/graphql/filters.rs
```diff
@@ -13,5 +13,130 @@ pub struct DateFilter {
...
+    /// Match values that start with this prefix — a subtree match, e.g.
+    /// `acmeCo/` matches `acmeCo/`, `acmeCo/team/`, and so on.
     pub starts_with: Option<String>,
+    /// Match values exactly equal to any entry in this set. When provided the
+    /// set must be non-empty; an empty `in` is rejected during input
+    /// validation rather than silently matching nothing (or everything).
+    /// `startsWith` and `in` are mutually exclusive: a resolver rejects a
+    /// filter that sets both, so a prefix scope is always either a subtree
+    /// (`startsWith`) or an exact set (`in`), never a mix.
+    #[graphql(validator(min_items = 1))]
```
jshearer: `min_items = 1` puts a floor on this but no ceiling -- `narrow_to_exact_set` is O(grants × entries) and runs before the `MAX_PREFIXES` guard, and the whole array binds into the `= ANY($n)`. `MAX_PREFIXES` bounds the side the caller doesn't control; this is the side they do. Can we cap it?

```suggestion
    #[graphql(validator(min_items = 1, max_items = 20))]
```

crates/control-plane-api/src/server/public/graphql/alert_configs.rs
```diff
@@ -578,4 +581,151 @@ mod test {
...
+            &server,
+            &alice_token,
+            serde_json::json!({ "catalogPrefixOrName": { "startsWith": "aliceCo/team/" } }),
+        )
+        .await;
+        assert_eq!(subtree, vec!["aliceCo/team/"]);
+
+        // `in` matches an exact set. Alice can read two prefixes, so this also
+        // exercises the retain() that narrows the authorized set down before
+        // the MAX_PREFIXES guard.
```
jshearer: Alice reads two prefixes against a cap of twenty, so this doesn't exercise anything `MAX_PREFIXES`-related -- the retain() runs, but the guard would pass either way. As far as I can tell nothing in any of the four resolvers covers the guard, so the behavior `narrow_to_exact_set` exists for (an `in` filter rescuing a caller with more than `MAX_PREFIXES` readable prefixes) is untested in either direction. Fine as a follow-up, but the comment shouldn't claim the coverage.

jshearer (dismissed):
LGTM % couple nits

jshearer (approved):

PR #3243, opened 2026-07-23
data-plane-controller: add GitHub App git auth to replace SSH machine user

skord:
## Why

DPC's service authenticates to GitHub to clone estuary/est-dry-dock at
runtime. Today it does so as the `estuary-dpc` machine user via an SSH key
loaded into ssh-agent by entrypoint.sh. We're enabling SAML SSO enforcement
at the org level, which locks out non-identity-tied accounts like this
machine user. GitHub App identities are exempt from SSO enforcement, so we're
moving DPC's git auth to a GitHub App before the cutover, then retiring the
machine user and reclaiming its Enterprise seat.

## Approach

Keep authentication ambient, exactly where it lives today (entrypoint.sh),
so the controller's worker code does not change. The one real difference from
an SSH key is that App installation tokens expire hourly, so a static
credential loaded once at startup is not enough. Git's built-in answer to
"a credential that needs periodic refresh" is a credential helper, so that's
what this adds.

- New `git-credential` subcommand acts as a git credential helper. On `get`
  it signs an RS256 App JWT, exchanges it for an installation access token,
  caches the token to disk until shortly before expiry, and hands it to git
  as the password for `x-access-token`. The token only ever crosses the
  stdin/stdout pipe with git; it never appears on a command line or in an
  on-disk git config, so it cannot leak into the user-visible command logs.
- entrypoint.sh gains a `DPC_GIT_AUTH_MODE` switch (default `ssh`). In
  `github-app` mode it rewrites the SSH-style remote to HTTPS and registers
  the helper, instead of setting up ssh-agent.
- The deploy workflow (service only; the job does no git) gains the App key
  secret and the App/Installation IDs, with the mode defaulting to `ssh` so
  merging and deploying changes nothing until we deliberately flip it.

worker.rs, the protocol, ControllerConfig, and the tests are untouched.

## Scope

Only estuary/est-dry-dock is affected. An audit-log export over a 3-day
window (~9,500 authenticated events) shows `estuary-dpc` clones/fetches
est-dry-dock and nothing else, all read-only. estuary/flow is a public repo
(no auth needed to clone), and estuary/ops is dead config in DPC (its
checkout was removed in 149b151a27b). So the App needs Contents: Read-only on
est-dry-dock alone.

## Verification

Built and ran the credential helper locally against the real App and the key
in Secret Manager:
- Minted a valid installation token (App ID 4126016, Installation ID
  148554392).
- Performed a real HTTPS clone of est-dry-dock through the helper using the
  same insteadOf rewrite entrypoint.sh applies in github-app mode. Clone
  succeeded.

## Cutover (after merge)

1. Deploy as-is (`ssh` mode). Ships the new binary plus the App key/IDs to the
   service; behavior unchanged.
2. Flip `DPC_GIT_AUTH_MODE=github-app` in the workflow, redeploy the service,
   confirm a clean clone/fetch cycle over HTTPS in the logs.
3. Remove the SSH key from `estuary-dpc` (reversible), wait a full cycle to be
   safe, then delete the user and reclaim the seat.

Note: the deploy action uses `env_vars_update_strategy: overwrite`, so
`DPC_GIT_AUTH_MODE` must be flipped in this workflow file, not in the Cloud
Run console (a console edit would be reverted on the next deploy).

## Rollback

Rollback is a single change: set `DPC_GIT_AUTH_MODE=ssh` in this workflow and
redeploy the service. entrypoint.sh falls back to the existing ssh-agent path,
which is unchanged and still fully wired (the `DPC_GITHUB_SSH_KEY` secret
remains in place throughout cutover). No code revert, no rebuild.

This is why the retire step is deliberately last and ordered as it is: keep the
SSH key and the `estuary-dpc` user until github-app mode has run cleanly for a
full cycle. Do not remove the SSH key until you are ready to give up the
instant rollback, since removing it is what makes the ssh path stop working.

## Prerequisites already done

- GitHub App created, Contents: Read-only, installed on estuary/est-dry-dock.
- Private key stored in estuary-control Secret Manager as `DPC_GITHUB_APP_KEY`.

### DIFF
--- a/.github/workflows/deploy-data-plane-controller.yaml
+++ b/.github/workflows/deploy-data-plane-controller.yaml
@@ -58,6 +58,14 @@ jobs:
             DPC_DATABASE_CA=/etc/db-ca.crt
             DPC_DATABASE_URL=postgresql://postgres@db.eyrcnmuzzyriypdajwdk.supabase.co:5432/postgres
             NO_COLOR=1
+            # Git auth for cloning private repos. Leave `ssh` until the GitHub
+            # App is verified, then flip to `github-app` to cut over. See
+            # crates/data-plane-controller/entrypoint.sh.
+            DPC_GIT_AUTH_MODE=ssh
+            # GitHub App identifiers (non-secret). Installation ID is available
+            # once the App is installed on estuary/est-dry-dock.
+            DPC_GITHUB_APP_ID=4126016
+            DPC_GITHUB_INSTALLATION_ID=148554392
 
           secrets: |-
             CONTROL_PLANE_DB_CA_CERT=CONTROL_PLANE_DB_CA_CERT:latest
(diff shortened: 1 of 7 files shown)


mwillman-estuary (approved):
Not a Rust guy, but the explanation makes sense to me, and the code looks cleanish.  Also, the downside to this as I understand it is that dry-dock basically fails on checkout, which does not seem like a end of the world scenario to me.  Approving so this can move forward.

PR #3250, opened 2026-07-23
data-plane-controller: fix broken deploy (comments in env/secrets block scalars)

skord:
## Problem

The data-plane-controller deploy is failing on master (first failure: https://github.com/estuary/flow/actions/runs/30048846203). `env_vars` and `secrets` for `deploy-cloudrun` are literal YAML block scalars (`|-`), so the explanatory `#` comment lines added in #3243 were passed verbatim to the action, which tried to parse a comment as a secret mapping:

```
ERROR: (gcloud.run.deploy) No secret version specified for
# GitHub App private key (PEM). Use # GitHub App private key (PEM):latest ...
```

No outage: a failed Cloud Run deploy leaves the previous revision serving, so DPC keeps running on its prior revision. But the deploy pipeline is broken until this lands, and the new binary is not rolled out.

## Fix

Remove the inline `#` comments from both block scalars in the deploy-service step and move the operational note to real YAML comments above `env_vars:`. No functional change: `DPC_GIT_AUTH_MODE` stays `ssh`, and the App key/IDs remain wired for the later cutover.

## Follow-up

#3249 (the github-app cutover flip) branched before this fix and carries the same bad comments; it needs a rebase on top of this before it can deploy.

### DIFF
--- a/.github/workflows/deploy-data-plane-controller.yaml
+++ b/.github/workflows/deploy-data-plane-controller.yaml
@@ -54,16 +54,16 @@ jobs:
           # Declare the VPC egress (previously set out-of-band) in code so it survives a full replace.
           flags: '--args=service --no-allow-unauthenticated --service-account=data-plane-controller@estuary-control.iam.gserviceaccount.com --network=data-plane-controller --subnet=data-plane-controller-sn1 --vpc-egress=all-traffic'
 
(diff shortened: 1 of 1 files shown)


mwillman-estuary (approved):
it's the sneaky ones that get ya

PR #3252, opened 2026-07-24
docs: document connector-managed schemas and auto-discover behavior

jwhartley:
## What

Documents how auto-discover interacts with collection schemas, which is currently undocumented (`flow://connector-schema` appears only incidentally on the redaction page).

- **`concepts/schemas.md`** — new subsection "Connector-managed schemas and auto-discover": the `flow://connector-schema` `$defs` marker that makes a collection connector-managed, why a hand-authored/inline schema is left untouched (the `not managed by auto-discover` error), and how to opt one in. Notes that the read/write split for inference is created automatically once a collection is managed.
- **`concepts/captures.md`** — extends the `autoDiscover` options with the management prerequisite and how new / dropped / renamed source tables are reconciled.

## Why

A customer hit repeated auto-discover failures because their collections were published from a local `flowctl` project with inline schemas and so were never connector-managed. Nothing in the docs explained the requirement or the fix.

## Verification

Behavior confirmed against source: `crates/control-plane-api/src/discovers/specs.rs` (management check + binding reconcile) and `crates/models/src/schemas.rs` (the `flow://` reference constants).

### DIFF
--- a/site/docs/guides/flowctl/ci-cd.md
+++ b/site/docs/guides/flowctl/ci-cd.md
@@ -44,6 +44,18 @@ To create a [capture](../../concepts/captures.md), start with a local `flow.yaml
 You will need to use the capture connector's [reference](../../reference/Connectors/capture-connectors/README.md) for details on the available settings, authorization methods, and required fields for the configuration.
 The connector's reference page will include an example specification you can use to get started.
 
+:::tip Recommended: generate bindings and collections with `flowctl discover`
+Rather than hand-writing every binding and its collection, start from a minimal capture spec with an empty `bindings: []` list, then run [`flowctl discover`](../../concepts/captures.md#discovery-through-flowctl):
+
+```bash
+flowctl discover --source flow.yaml
+```
+
+This invokes the connector from your data plane, enumerates the source's resources, and writes the `bindings` and their target `collections` back into your local files. The generated collections are **connector-managed** and inference-ready, so [`autoDiscover`](../../concepts/captures.md#automatically-update-captures) can keep their schemas in sync and the read schema evolves automatically. This is the recommended starting point for most captures.
+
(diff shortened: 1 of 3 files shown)


jwhartley:
Added a third file to this PR: `guides/flowctl/ci-cd.md`.

The guide previously told CI/CD users to hand-author collection schemas and "be very aware of your source system's schema in order to replicate it accurately" — which produces exactly the unmanaged, frozen collections that motivated this PR. The changes:

- Lead the capture section with `flowctl discover` as the recommended way to generate bindings and connector-managed, inference-ready collections.
- Reframe hand-authoring toward a permissive `writeSchema` + `readSchema: allOf[flow://write-schema, flow://inferred-schema]` (so inference handles evolution) instead of a fully-specified frozen schema, and note this is compatible with a version-controlled catalog.
- Add a note on `autoDiscover` republishing out-of-band vs. Git-managed specs.

Happy to split this into its own PR if you'd rather review the `concepts/` changes and the guide change separately.

site/docs/guides/flowctl/ci-cd.md
```diff
@@ -44,6 +44,18 @@ To create a [capture](../../concepts/captures.md), start with a local `flow.yaml
 You will need to use the capture connector's [reference](../../reference/Connectors/capture-connectors/README.md) for details on the available settings, authorization methods, and required fields for the configuration.
 The connector's reference page will include an example specification you can use to get started.
 
+:::tip Recommended: generate bindings and collections with `flowctl discover`
```
aeluce, line 47: I'd like the main described path in the guide to be our recommended path (so, by default, the guide would describe using auto-discover for the bindings rather than leaving the recommended route in a tip box). If you'd like to get the concepts changes out first and you want to split the PR, I can approve those pages.

jwhartley:

as-is. For this PR:

- `site/docs/concepts/schemas.md` becomes `content/concepts/schemas.md`
- `site/docs/guides/flowctl/ci-cd.md` becomes `content/guides/flowctl/ci-cd.md`

PR #3260, opened 2026-07-24
docs: fix broken anchor, enforce anchor checking, document macOS build gotcha

jwhartley:
## What

1. Fixes the one broken anchor found site-wide: `firstwritewins-and-lastwritewins.md` linked `./#reduction-guarantees` (itself) instead of `../#reduction-guarantees` (the reduction-strategies overview page, where that heading lives).
2. Flips `onBrokenAnchors` from `warn` to `throw` in `docusaurus.config.js`, now that the only offender is fixed, so future broken anchors fail the build instead of shipping silently.
3. Documents a pre-existing `hubspot-real-time` redirect/page collision in `site/README.md`: the redirect's `from` path differs from the real `HubSpot-real-time` page only by case, so `npm run build` fails on macOS's case-insensitive filesystem with "the redirect plugin is not supposed to override existing files." Confirmed this does not affect production (Linux CI is case-sensitive), so it's a local-dev note, not a code change.

## Why

Found while investigating stale-anchor feedback on #3094 — a heading rename there broke an anchor link silently since `onBrokenAnchors` only warns. Verified locally that flipping it to `throw` catches that class of bug without any other backlog (confirmed via full `npm run build` before and after).

## Testing

- `npm run build` locally: only failure is the pre-existing hubspot-real-time macOS collision (unrelated, documented, not present in CI).
- Confirmed via grep this is the only `reduction-guarantees` broken anchor pattern in the docs tree.

### DIFF
--- a/.gitleaksignore
+++ b/.gitleaksignore
@@ -1,0 +1,2 @@
+# Example private key placeholder in documentation (Snowflake.md line 177)
+site/docs/reference/Connectors/materialization-connectors/Snowflake.md
--- a/site/docs/LEARNINGS.md
+++ b/site/docs/LEARNINGS.md
@@ -1,0 +1,40 @@
+# Learnings
+
+## 2026-02-02 | Documentation | Internal links break when moving files
+
+**Issue:** After moving docs from `features/` to `guides/advanced-usage/`, internal links using absolute paths (`/features/feature-flags/`) broke.
+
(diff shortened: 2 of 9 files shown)


jwhartley:
Closing: this PR was opened against the wrong branch by mistake (a stray branch on this repo that happens to share a name with an unrelated local branch elsewhere). Correct PR: see replacement linked shortly.

PR #3261, opened 2026-07-24
shuffle: fix open-phase deadlock via EOF-cascade teardown

### LINKED ISSUE #3245: runtime-v2: multi-shard shuffle open deadlocks when a peer sidecar is briefly unreachable
Observed in production on a two-shard `runtime-next` (V2) materialization: the task wedged at startup for 25+ minutes with no error, no retry, and no teardown — a single frozen `shuffle.slice` "received Open from Session" track entry, and nothing after two WARN `Connection refused` lines. Confirmed by source analysis (line references as of 8ddd4d44b21). The bug is latent in **every multi-shard V2 task** (materializations and derivations): it triggers whenever one shard's runtime-sidecar is briefly unreachable while another shard's shuffle Session is opening — a routine cold-start / rolling-restart race. In the observed incident the peer's sidecar began listening ~49 seconds *after* the dial failed, but nothing ever re-attempts, so the task hung indefinitely.

## Deadlock chain

```
leader startup → SessionClient::open (awaits Opened, no timeout)              [HUNG]
 └ Session: join_all[ open_slice_rpc(0), open_slice_rpc(1) ]                  [HUNG]
     ├ open_slice_rpc(1) = Err (connection refused, peer sidecar not up)      ← resolved, trapped
     └ open_slice_rpc(0) → in-process Slice 0, awaits Opened                  [HUNG]
         └ Slice 0: join_all[ open_log_rpc(0), open_log_rpc(1) ]              [HUNG]
             ├ open_log_rpc(1) = Err (connection refused)                     ← resolved, trapped
             └ open_log_rpc(0) → in-process Log 0, awaits Opened              [HUNG]
                 └ Log 0 rendezvous needs ALL slices to connect. Slice 1 was
                   never created (its open failed at the Session), so it
                   never dials in.                                            [HUNG FOREVER]
```

## Root cause

1. **`futures::future::join_all` (not `try_join_all`) in both fan-out open paths** — `crates/shuffle/src/session/handler.rs:80` and `crates/shuffle/src/slice/handler.rs:85`. `join_all` does not short-circuit on `Err`; the `result?` walk after it is unreachable while any sibling future is parked. The peer's terminal error occurs and resolves, but is trapped.

2. **The in-process open parks on the Log rendezvous, permanently.** On Open, a Log handler registers its `(request_rx, response_tx)` into the process-global `Service::log_joins` map keyed by `(directory, log_shard_index)` and, if not all slices have connected, returns `Ok(())` immediately (`crates/shuffle/src/log/handler.rs:113-131`). Because `response_tx` is *moved into the map*, the Slice awaiting `Opened` (`crates/shuffle/src/slice/handler.rs:238`) gets neither a message **nor an EOF** — the channel never closes. `Opened` is only ever sent by the invocation that completes the rendezvous, which requires the exact peer whose open already failed.

3. **No timeout at any layer.** `SessionClient::open` (`crates/shuffle/src/client.rs:45`), the leader call site (`crates/runtime-next/src/leader/materialize/startup.rs:248`), and the Go controller's wait for `Opened` (`go/runtime/materialize_v2.go:230`) all await without deadlines. Per-actor tick/stall machinery only runs inside the actor `serve` loops, i.e. after open completes. HTTP/2 keepalives can't help — every parked leg is an in-process mpsc channel.

Net effect: the crate's documented fail-fast recovery model (README) is defeated — the terminal error never propagates, so teardown/rebuild never begins.

## Aggravating factors

- **Teardown doesn't clean up.** Session/Slice/Log handlers are detached `tokio::spawn` tasks; while parked in open they never poll their own request streams, so dropping the `SessionClient` (Stop, shard reassignment, task deletion) is not observed. The wedged tasks plus the `LogJoin` map entry leak in the sidecar until process restart.

- **Retry poisoning.** `log_joins` entries are removed only when a rendezvous completes (`log/handler.rs:134`). The join key's `directory` is the per-`NewStore` shuffle tempdir (`go/runtime/materialize_v2.go:57`; same pattern in `derive_v2.go`), which is stable across in-tenure session retries. Any aborted partial rendezvous therefore makes the retry's Slice→Log open fail with `received duplicate Slice connection` (`log/handler.rs:108`). A naive `try_join_all` fix converts the hang into a crash-loop unless this is also addressed.

- **`try_join_all` alone is insufficient in a nearby variant.** If the remote Slice open fails while the peer sidecar *is* reachable (e.g. authz error), the local Slice's remote Log open succeeds in dialing and parks in the *peer's* rendezvous — neither of its log opens errors, so a slice-level `try_join_all` has nothing to short-circuit on. Session-level `try_join_all` restores error propagation, but dropping the open future doesn't cancel the detached in-process Slice task, which stays parked holding live rendezvous slots in two sidecars. The open phase itself must be cancellable (raced against request-stream closure, or bounded by a deadline).

## Suggested fix directions

1. Replace `join_all` with `try_join_all` in `session/handler.rs` and `slice/handler.rs` so a peer open error short-circuits and propagates → documented fail-fast teardown engages.
2. Companion (required): make partial `LogJoin` rendezvous state abort-safe — e.g. the incomplete-arrival Log handler parks awaiting rendezvous-complete OR its own request-stream abort, removing its slot on abort — so aborted sessions neither poison retries nor leak.
3. Bound the shuffle open path with a deadline (Session and/or Slice). Given the variant above, this is load-bearing for full cleanup, not just defense-in-depth.
4. Confirm Gazette consumer-framework Store lifecycle on a hard `runOneSession` error return (fresh Store+tempdir vs reuse) to size how much of (2) is reachable in practice.

jgraettinger:
**Description:**

Fixes #3245: a multi-shard V2 task wedged forever at startup whenever one shard's runtime-sidecar was briefly unreachable while another shard's shuffle Session was opening — a routine cold-start / rolling-restart race.

The Session→Slice→Log open fan-outs used `futures::future::join_all`, which does not short-circuit on `Err`: a peer's connection-refused error resolved but was trapped while sibling in-process opens parked forever on the Log rendezvous (which completes only once ALL Slices connect). The parked handlers are detached tasks not yet polling their request streams, so teardown was never observed, and partial rendezvous state in the process-global `Service::log_joins` map poisoned in-tenure retries with `received duplicate Slice connection`.

The fix is structural and entirely within `crates/shuffle`: the opening phase now participates in the same request-stream-EOF teardown cascade the running phase already uses (per the crate README's shutdown model).

- **Session & Slice handlers**: `join_all` → `try_join_all`, raced via `tokio::select!` against the handler's own request stream. The first downstream open error short-circuits, dropping sibling open futures and cascading EOF teardown downward. Any EOF/error/message on the request stream during opening (all illegal before `Opened`) likewise aborts.
- **Log rendezvous is abort-safe**: a non-completing invocation no longer returns immediately after registering. It parks on a per-slot `oneshot` completion signal raced against `response_tx.closed()`, and on abort reaps its own slot under the `log_joins` mutex (removing the entry when its last slot goes). Aborted partial rendezvous neither poison retries nor leak.
- **`LogJoin` key is session-scoped**: `(directory, log_shard_index)` → `(directory, session_id, log_shard_index)`, making retries collision-free by construction; abort cleanup is then leak-prevention rather than correctness-critical. The duplicate-slice check remains as an intra-session protocol assertion.

**Deliberately no deadlines or stall watchdogs.** Every transient failure now errors and self-heals through error propagation plus the Go retry loop. The remaining non-erroring wedge classes are rare and likely persistent — a stable wedge is far easier to debug than one flapping through teardown/retry. Rationale is recorded in the crate README so it isn't "fixed" later.

**Workflow steps:**

None — internal runtime bug fix with no user-facing surface. Multi-shard V2 tasks now recover from a briefly-unreachable peer sidecar instead of hanging.

None (internal `crates/shuffle/README.md` updated in-repo).

**Notes for reviewers:**

- Two commits, TDD: the first adds three red regression tests (`crates/shuffle/src/open_deadlock_tests.rs`, each with a 30s hang guard) verified failing — hanging — against the unfixed code; the second lands the fix and README update, turning them green in <0.1s.
- No Go, protocol, or runtime-next changes; `session_id` already rides in every Open. `log_joins` is process-local, so there is no mixed-version concern. Derivations get the fix alongside materializations.
- Full `cargo test -p shuffle` passes (99 lib tests + scenario fixture/fuzz tests).

### DIFF
--- a/crates/shuffle/README.md
+++ b/crates/shuffle/README.md
@@ -91,6 +91,32 @@ consumed, a coordinator does it only once it will request no further checkpoints
 then calls `SessionClient::close()` (which blocks until the Session→Slice→Log
 topology has fully drained).
 
+#### Teardown during the opening phase
+
+The opening phase participates in the same EOF cascade. Each handler opens its
+downstream peers and races that fan-out against its own request stream. It
+short-circuits if any downstream open errors — dropping the sibling open futures,
+which drops their request channels and cascades EOF down to those peers and
+their peers in turn. Racing the request stream catches the *upstream* going
+away mid-open.
+
+The Log rendezvous is the subtle case. `M` Slices connect to each Log shard; the
+invocation that connects last runs the `LogActor`, while the earlier `M-1` park.
+A parked invocation must remain cancel-observant, so it selects its own
+completion signal against `response_tx.closed()` (fired when its Slice
+client drops its receiver) and reaps *only its own* slot under the `log_joins`
+mutex (dropping the whole entry when it removed the last live slot),
+so a stale partial rendezvous can't wedge the map or poison
+a retry.
+
+#### No deadlines, by design
+
+There are deliberately **no** open-phase deadlines. Every transient failure
+surfaces as an error that self-heals via `try_join_all` and the Go retry loop,
+so a timeout would add flapping that would only paper over a true regression,
+and makes the ensemble more difficult to debug (a stable wedge is far easier
+to inspect).
+
 ### Authorization
 
 When the `Service` is built with a `proto_grpc::Signer` (the sidecar; `None` in
--- a/crates/shuffle/src/lib.rs
+++ b/crates/shuffle/src/lib.rs
@@ -119,6 +119,27 @@ fn verify_send<T>(tx: &mpsc::Sender<T>, value: T) -> anyhow::Result<()> {
     }
 }
 
+/// Build the error for a request stream that closed or spoke out of turn while
+/// a handler was still opening. No legitimate request arrives before a handler
+/// emits its `Opened` (the coordinator sends `resume_checkpoint`, and the
+/// Session sends `Start`, only after observing `Opened`). So EOF, an error, or
+/// any message means the client (or its RPC) went away: the handler aborts,
+/// dropping its downstream open fan-out so that EOF teardown cascades to peers.
+/// Deliberately involves no deadline — see the crate README "Shutdown" notes.
+fn opening_aborted<T>(source: &str, peer: &str, msg: Option<tonic::Result<T>>) -> anyhow::Error {
+    match msg {
+        None => {
+            anyhow::format_err!("{source} request stream from {peer} reached EOF during opening")
+        }
+        Some(Err(status)) => status_to_anyhow(status).context(format!(
+            "{source} request stream from {peer} errored during opening"
+        )),
+        Some(Ok(_)) => anyhow::format_err!(
+            "{source} received an unexpected request from {peer} during opening (before Opened was sent)"
+        ),
+    }
+}
+
 // Map an anyhow::Error into a tonic::Status.
 #[inline]
 fn anyhow_to_status(err: anyhow::Error) -> tonic::Status {
--- a/crates/shuffle/src/lib.rs
+++ b/crates/shuffle/src/lib.rs
@@ -54,6 +54,8 @@ mod session;
 pub mod slice;
 
 #[cfg(test)]
+mod open_deadlock_tests;
+#[cfg(test)]
 pub(crate) mod testing;
 
 /// Document passed JSON Schema validation.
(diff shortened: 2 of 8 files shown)


williamhbaker (approved):
LGTM

PR #3264, opened 2026-07-24
runtime-v2: bind task terms to ShardSpec bytes and restart gracefully from Join-wait

### LINKED ISSUE #3247: runtime-v2: multi-shard task enters self-sustaining fail-stop restart loop after any single shard drop
## Summary

A leaderful runtime-v2 task (materialization or derivation) that has been split into N > 1 shards can enter a **self-sustaining restart loop** after any single shard drop. Once entered, all N shards cycle FAILED → restart → FAILED with **zero forward progress**, on a period that grows with the control-plane's detection backoff. The loop was never observed to self-terminate; in every observed case it recovered only via manual intervention (deleting all of the task's Etcd assignment keys in one transaction, which recovers the task in ~3 s from any point in the cycle).

The Join/Joined consensus mechanism itself is **not** at fault: given a stable topology where all N shards hold PRIMARY simultaneously, consensus completes in 2–7 s, every time, including under aggressive assignment churn and reactor kills. The defect is that the control-plane recovery path structurally never *produces* that topology.

Reproduced on an isolated local stack against `ops/views/catalog-stats` split to 4 shards; minimal repro is: healthy 4-shard V2 task → delete one shard's assignment key → observe stall (no `leader.materialize` handler on the sidecar, frozen `runtime_leader_transactions`) → recover with whole-task assignment delete.

## Observed timeline (single `del_one` from a healthy 4-shard baseline)

- **T+0:** deleted shard's stream drops; leader fail-stops the session; all 4 shards go FAILED (`Materialize error (expected leader message) … unexpected EOF`).
- **T+8 s:** the allocator re-serves 3 shards; they reach PRIMARY and park in the Join-wait loop, correctly waiting for the 4th. The 4th stays FAILED (a FAILED slot-0 assignment is "consistent" to the allocator, which therefore never restarts it on its own).
- **T+8 s → T+119 s:** stall. The 3 waiters are stable and healthy — not busy-failing.
- **T+119 s:** the agent detects the failure via periodic health-check (3 consecutive failed checks) and restarts the task. The missing shard comes back **and at the same instant the 3 waiters fail** with `servePrimary failed err="runTransactions: readMessage: awaiting Etcd revision N: context canceled"`.
- The cycle repeats with roles swapped, on a growing period (~105–111 s observed, capping at the health-check backoff).

## Root cause: three composing behaviors

**1. Term-context cancellation is fatal during Join-wait but graceful in steady state.**
The Phase-A1 Join-wait loop blocks in `ks.WaitForRevision(m.term.ctx, …)` and returns any error (`go/runtime/materialize_v2.go:186-218`); per the consumer framework, any non-nil `runOneSession` return → shard FAILED (the only in-process retry is on a nil return). After `Opened`, the *same* term cancellation instead sends `Stop` and returns nil — a graceful in-process session restart (`materialize_v2.go:240-267`). This contradicts the documented term contract at `go/runtime/task.go:219-220`: *"A cancellation of the term's Context doesn't invalidate the shard."* `derive_v2.go` has the identical structure. Phase A2 (parked in `recv()` awaiting `Joined`) has the same exposure.

**2. Every activation re-PUTs every ShardSpec, and a byte-identical PUT still cancels all terms.**
The restart path is `do_activate` → `data_plane_activate` → `activate::activate_materialization` → `task_changes`, which emits a ShardSpec upsert for **every** shard of the task unconditionally — no diff against the current value (`crates/activate/src/lib.rs:607-612`). Gazette's `ShardApply` is a bare `OpPut` with no value comparison, so a byte-identical spec (the normal case for a restart — same build) still bumps mod-revision and produces a watch event. The KeySpace re-decodes the item, and `signalOnSpecUpdate` compares **pointer identity** (`go/runtime/task.go:498`), so the re-minted pointer cancels the term of every live shard of the task — healthy Join-waiters included. Combined with (1), the control-plane's restart of the failed shard simultaneously kills every shard that was correctly waiting for it.

**3. `apply_changes` unassigns failed shards *after* the spec upserts, and loses the race with the failures those upserts cause.**
`apply_changes` orders: upsert all ShardSpecs, then `Unassign(only_failed: true)` over the task's shards (`crates/activate/src/lib.rs:243-314`). `ShardUnassign` evaluates FAILED status from the consumer's KeySpace at RPC time — but the waiters killed by the spec upserts reach FAILED only after an async multi-step path (watch delivery → term cancel → error propagation → Etcd status write), while the Unassign follows the Apply almost immediately. So each activation's unassign sees and restarts only the *previously*-failed set, while the spec PUTs manufacture a *new* failed set it cannot see.

**The loop:** the FAILED subset alternates ({straggler} → {the other N−1} → …) and is never empty at activation time; the allocator never restarts FAILED shards on its own; `PendingJoin` has no timeout (`crates/runtime-next/src/leader/join.rs`). All-N-simultaneously-PRIMARY is structurally unreachable through this recovery path — it isn't bad luck, it's the ordering.

**Why manual whole-task assignment deletion works:** it deletes assignment keys only — no ShardSpec write — so all N shards restart via the silent shard-context path with fresh terms, come up together at one revision, and Join within seconds.

## Scope / severity

- Affects **any multi-shard runtime-v2 task**; any transient single-shard drop (reactor OOM, node drain, assignment churn) converts into an indefinite zero-progress crash loop that the control plane's own recovery perpetuates.
- The reproduction used an ops-catalog task, but **the defect is not ops-specific** — ops-catalog only changes the detection mechanism (periodic health checks, cycle capped at ~5 min by `shard_health_check_interval`, `crates/agent/src/controllers/activation.rs:568-602`). User tasks restart through the same `do_activate` spec-re-PUT path via `ShardFailed` events and `get_next_retry_time` backoff, so the same kill-the-waiters dynamic applies unless the restart happens to land before the allocator re-serves survivors into Join-wait (~8 s window). Not yet exercised empirically for non-ops tasks.
- Single-shard V2 tasks pass through Join-wait near-instantly and are effectively unaffected. V1 tasks have no Join-wait phase and handle term cancellation gracefully everywhere they block.

Line references are at master `8ddd4d44b21`; Gazette references are the pinned `go.gazette.dev/core v0.103.1-0.20260715211535-42c2cc160b2c`.

jgraettinger:
**Description:**

Fixes #3247 by addressing causes (1) and (2). In brief:

1. A term was bound to the decoded `*pf.ShardSpec` pointer, which any re-decode re-mints. Terms are now bound to the raw etcd value bytes: `newTaskTerm` captures the decoded spec and its `Raw.Value` coherently from one `KeyValue` under a single read lock, and `signalOnSpecUpdate` compares `bytes.Equal` on each wakeup, so only a real value change or a deletion ends a term. The `taskSpec`-reuse optimization compares bytes as well, so a no-op re-PUT no longer re-opens the build DB.

2. Term cancellation was fatal during Join-wait but graceful post-`Opened`, contradicting the term contract in `task.go`. A Join-wait cancellation with a still-live shard context now logs at INFO and returns nil, so the framework loops back through `RestoreCheckpoint` with a fresh term. This is safe pre-consensus: the Rust session loop is parked awaiting a Join and re-Joins on the same long-lived stream. Applied identically to materialize and derive.

Root cause (3), the `apply_changes` upsert-before-`Unassign` ordering, is not addressed. With (1) fixed those upserts no longer manufacture a failed set, and the race evaporates.

**Workflow steps:**

No user-facing change. The surviving shards of a multi-shard V2 task now stay parked in Join-wait across an activation's spec re-PUT instead of failing, so all N reach consensus together.

None.

**Notes for reviewers:**

- The first commit affects V1 shards too, since `newTaskTerm` / `signalOnSpecUpdate` are shared. The pointer-identity bug was equally wrong there; V1 just tolerated the spurious term restart.
- Retaining `kv.Raw.Value` uncopied is safe: `KeySpace.Apply` rebuilds `KeyValues` wholesale on each watch update and never mutates value bytes in place.
- `newTaskTerm` now reads the ShardSpec from the KeySpace rather than `shard.Spec()`, and errors if the key is absent — where before the term was instead cancelled immediately by the signal goroutine. In that narrow window (STANDBY→PRIMARY as the spec is deleted) a deletion now surfaces as a shard failure. Reading the current spec rather than the transition-time one also means a new term binds to the latest spec, avoiding an immediate self-cancellation.
- `go/runtime/task_signal_test.go` is new, and adds a `TestMain` so `etcdtest` can host an etcd for the package. The identity-re-PUT test double-PUTs and awaits each revision: the second awaited round-trip gives the signal goroutine a full etcd RPC of scheduling grace to wrongly react to the first. It then PUTs different bytes and requires the term to end, proving the goroutine was watching throughout rather than parked on a revision it never received.

### DIFF
--- a/go/runtime/derive_v2.go
+++ b/go/runtime/derive_v2.go
@@ -198,8 +198,11 @@ func (m *deriveAppV2) StartReadingMessages(shard consumer.Shard, _ pc.Checkpoint
 // materializeAppV2.runOneSession: build & send Join until topology consensus,
 // send Task and await Opened (which carries the connector Container), then run
 // the steady-state select loop until term cancellation, shard cancellation, or
-// a stream error. On term cancellation (spec update) we send Stop and await
-// Stopped, then return nil so the framework loops back through RestoreCheckpoint.
+// a stream error. Term cancellation (a spec update, per the term contract in
+// task.go) is always a graceful in-process restart, never a shard failure, both
+// before and after Opened: pre-Opened we return nil directly, and post-Opened
+// we first send Stop and await Stopped. Either way the framework loops back
+// through RestoreCheckpoint with a fresh term.
 func (m *deriveAppV2) runOneSession(shard consumer.Shard, ch chan<- consumer.EnvelopeOrError) (err error) {
 	defer func() {
 		if err != nil {
@@ -221,6 +224,15 @@ func (m *deriveAppV2) runOneSession(shard consumer.Shard, ch chan<- consumer.Env
 		err = ks.WaitForRevision(m.term.ctx, waitForRevision)
 		ks.Mu.RUnlock()
 		if err != nil {
+			// A term cancellation is a spec update, not a failure. The Rust
+			// session loop is still parked awaiting a Join, so returning nil
+			// restarts the session in-process. Only a shard-context
+			// cancellation is a genuine shutdown.
+			if shard.Context().Err() == nil {
+				logrus.WithField("shardId", shard.Spec().Id).
+					Info("task term ended during Join-wait; restarting session")
+				return nil
+			}
 			return fmt.Errorf("awaiting Etcd revision %d: %w", waitForRevision, err)
 		}
 
--- a/go/runtime/materialize_v2.go
+++ b/go/runtime/materialize_v2.go
@@ -165,10 +165,12 @@ func (m *materializeAppV2) StartReadingMessages(shard consumer.Shard, _ pc.Check
 //  4. Run the steady-state select loop until term cancellation, shard
 //     cancellation, or a stream error.
 //
-// On term cancellation (spec update) we send Stop and wait for Stopped, then
-// return nil so the framework loops back through RestoreCheckpoint. Any
-// non-nil return is forwarded onto `ch` by the deferred sentinel before
-// the channel is closed.
+// Term cancellation (a spec update, per the term contract in task.go) is always
+// a graceful in-process restart, never a shard failure, both before and after
+// Opened: pre-Opened we return nil directly, and post-Opened we first send Stop
+// and wait for Stopped. Either way the framework loops back through
+// RestoreCheckpoint with a fresh term. Any non-nil return is forwarded onto
+// `ch` by the deferred sentinel before the channel is closed.
 func (m *materializeAppV2) runOneSession(shard consumer.Shard, ch chan<- consumer.EnvelopeOrError) (err error) {
 	defer func() {
 		if err != nil {
@@ -190,6 +192,15 @@ func (m *materializeAppV2) runOneSession(shard consumer.Shard, ch chan<- consume
 		err = ks.WaitForRevision(m.term.ctx, waitForRevision)
 		ks.Mu.RUnlock()
 		if err != nil {
+			// A term cancellation is a spec update, not a failure. The Rust
+			// session loop is still parked awaiting a Join, so returning nil
+			// restarts the session in-process. Only a shard-context
+			// cancellation is a genuine shutdown.
+			if shard.Context().Err() == nil {
+				logrus.WithField("shardId", shard.Spec().Id).
+					Info("task term ended during Join-wait; restarting session")
+				return nil
+			}
 			return fmt.Errorf("awaiting Etcd revision %d: %w", waitForRevision, err)
 		}
 
--- a/go/runtime/task.go
+++ b/go/runtime/task.go
@@ -1,6 +1,7 @@
 package runtime
 
 import (
+	"bytes"
 	"context"
 	"database/sql"
 	"encoding/json"
@@ -62,7 +63,10 @@ type taskTerm[TaskSpec pf.Task] struct {
 	ctx       context.Context    // Context for the task term.
 	labels    ops.ShardLabeling  // Shard labels of this task term.
 	shardSpec *pf.ShardSpec      // Term ShardSpec of the task.
-	taskSpec  TaskSpec           // Term TaskSpec of the task.
+	// Raw etcd value of shardSpec, which defines this term: it ends when these
+	// bytes change or the key is deleted, but not on a byte-identical re-PUT.
+	specBytes []byte
+	taskSpec  TaskSpec // Term TaskSpec of the task.
 }
 
 type taskReader[TaskSpec pf.Task] struct {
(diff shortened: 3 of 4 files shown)


jgraettinger:
## QA on a local stack: both root causes verified, with a master control

Ran this branch against an isolated local stack, side by side with a reactor rebuilt from `master`, and both commits behave as described.

### Setup

A runtime-v2 materialization (`materialize-bigtable`) split to 4 shards via `flowctl raw split-shards`, healthy at `leader.materialize phase=running shards=4`.

Activations were driven deterministically rather than by waiting out the health-check backoff: reset `activation.last_activated` below `last_build_id` and wake the controller task. Same `do_activate` path, same byte-identical ShardSpec upserts (verified: `mod_revision` advances, value sha256 unchanged).

For the control, note the reactor unit rebuilds `flowctl-go` from the worktree on every start, so the checkout — not just the binary — had to be swapped to `master`.

### Cause (1) — terms bound to spec bytes

A byte-identical re-PUT of all 4 ShardSpecs:

| | result |
|---|---|
| `master` | leader handler id 108 → **134**, phase age reset to 23s — every term cancelled, whole session restarted |
| this branch | handler id **34 unchanged**, age 273s → 303s monotonic — no restart at all |

### Cause (2) — the restart loop

Deleted one shard's assignment key to produce the stall from #3247 (3 FAILED, 1 healthy shard parked in Join-wait), then forced an activation:

- **`master`** — the waiter dies with the issue's verbatim signature, `runTransactions: readMessage: awaiting Etcd revision 738: context canceled`, and the FAILED set alternates `{c0000000}` → `{the other three}` → … across **4 consecutive activations**, never reaching all-4-PRIMARY. Self-sustaining, zero forward progress.
- **this branch** — one activation recovered all 4 to consensus; three further activations left the same leader session running (age 39 → 64 → 89 → 114s), 4/4 healthy throughout.

Swapping the fixed binary back in while the task was mid-loop converged it immediately.

### The Join-wait graceful-restart path, exercised directly

With a shard parked in Join-wait, publishing a *genuine* spec change (a `logLevel` edit) hits the new path — INFO instead of a shard failure — and the task converged to 4/4:

```
level=info msg="task term ended during Join-wait; restarting session" shardId=materialize/…/c0000000-00000000
```

### Unit tests

`go test -tags libsqlite3 ./go/runtime/` passes, including the three new `task_signal_test.go` cases.

williamhbaker (approved):
LGTM

PR #3268, opened 2026-07-27
publisher: discard unauthorized recovered ACK intents

### LINKED ISSUE #3263: runtime-v2: data-plane migration crash-loops on recovered ACK intents to the prior plane's ops journal
A task migrated between data planes crash-loops on startup with a
`PermissionDenied` error naming its **old** data plane's ops stats journal:

```
runTransactions: readMessage: rpc error: code = PermissionDenied desc = task shard
materialize/acmeCo/widgets/materialize-snowflake/0011223344556677/00000000-00000000
is not authorized to
ops/tasks/private/acmeCo/aws-us-east-1-c1/stats/1122334455667788/kind=materialization/name=acmeCo%2Fwidgets%2Fmaterialize-snowflake/pivot=00
for Write
```

The failure is persistent: the shard fails, restarts, recovers the same state, and
fails again. Observed after a migration from a private data plane to a public one,
but nothing about the failure is specific to that direction.

## This is not a stale ShardSpec label

The natural first suspicion is that the post-migration ShardSpec carries a stale
`estuary.dev/stats-journal`. It does not:

- Nothing in `assemble` / `build` emits `estuary.dev/logs-journal` or
  `estuary.dev/stats-journal`, so the label never rides along inside a built spec.
- `activate::task_changes` sets both unconditionally on every shard upsert
  (`crates/activate/src/lib.rs`), from the ops journal names derived from the
  templates passed by the caller.
- Neither label is in `labels::is_data_plane_label`, so the value observed on the
  source-plane shard is discarded rather than merged forward.
- `migrate` passes the *target* plane's ops templates everywhere it authors
  target-plane state (phase one's `task_changes`, and phase three's un-cordon).
  Source templates are only used for changes applied back to the source plane.

## Failure chain

The write is a **recovered ACK intent**, not a fresh stats publish.

1. Migration copies recovery logs and primary hints into the target plane. The
   runtime-v2 leader therefore recovers RocksDB state that includes
   `pending_ack_intents` (`crates/runtime-next/src/leader/materialize/startup.rs`).
   A materialization leader publishes exactly one journal — its ops stats
   partition — so the only recovered intent is keyed by the **source** plane's
   stats journal name.

2. The first transaction dispatches `Action::WriteIntents` carrying it
   (`crates/runtime-next/src/leader/materialize/fsm.rs`).

3. In `publisher::Publisher::write_intents` (`crates/publisher/src/publisher.rs`),
   that journal matches no current binding — the fixed ops-stats binding is the
   *new* journal, matched by exact name — so it takes the ephemeral-client path,
   authorizing against the old journal name as the AuthZ object.

4. `/authorize/task` denies it. The `is_ops` special case compares the journal's
   collection against `task_data_plane.ops_stats_name`, and the task's data plane
   is now the target, so the names differ. It falls through to the RBAC check,
   where the task holds no write grant on the source plane's ops prefix
   (`crates/control-plane-api/src/server/authorize_task.rs`).

5. `write_intents` tolerates only `NotFound` on the ephemeral path.
   `PermissionDenied` propagates, failing the leader session. The intent is never
   cleared, so every restart reproduces it exactly.

The error reaches the log in that shape because a runtime-v2 materialization
forwards any leader session error onto the consumer read channel
(`go/runtime/materialize_v2.go`), which `runTransactions` reports as
`readMessage:`. `pf.UnwrapGRPCError` passes `PermissionDenied` through verbatim,
preserving the `rpc error: code = ...` form.

## Why runtime-v1 does not hit this

`go/runtime/authorizer.go` carries an explicit accommodation for precisely this
scenario: an APPEND to an `ops/tasks/` journal that returns 403 is converted to
`client.ErrJournalNotFound`, which trips logic in the Gazette `consumer` module
that discards the recovered ACK intent and continues. Its comment describes the
newly-migrated-task case directly. There is no runtime-v2 counterpart — the Rust
runtime authorizes through its own journal client factory
(`crates/flow-client-next/src/workflows/task_collection_auth.rs`) and surfaces the
denial as a terminal `tonic::Status`.

## Why the graceful path does not engage

The control plane's "black hole" response
(`crates/control-plane-api/src/server/authorize_task.rs`) extends the claims with a
match-nothing selector when the requested journal maps to no known collection,
which turns an Append into `JOURNAL_NOT_FOUND` — exactly the status
`write_intents` already tolerates and discards. That only applies when the old
collection is *gone*. After a migration between two live data planes the source
ops collection still exists, so the request reaches the RBAC check and is denied
instead.

## Scope

Any runtime-v2 task migrated between data planes should be able to reproduce this,
whenever a recovered ACK intent names a journal in the previous plane that still
exists. Materializations are the clearest case because the stats partition is the
only journal their leader publishes, so a recovered intent for it is near-certain.

jgraettinger:
Fixes #3263.

`Publisher::write_intents` now tolerates `PermissionDenied` on its ephemeral path, alongside the `NotFound` it already discards.

A task holding a recovered ACK intent for a journal it is no longer authorized to crash-loops forever. Nothing clears an intent but writing it, and the task cannot heal itself — updating it to no longer publish there is precisely what stranded the intent. Fail-open is the only terminating choice; the cost is that documents behind the discarded ACK stay PENDING and are never released to readers.

This is safe because only recovered intents reach the ephemeral path: a live transaction's intents are keyed on journals it just wrote, so they always match a binding and still fail strictly.

Observed after a data-plane migration — ops journal names embed a plane, so they alone stop matching a binding — but the same follows any revoked grant on a collection the task published to under an earlier specification. `authorize_task` is correct and unchanged. Runtime-v1's narrower accommodation is left alone: it cannot express the "matches no current binding" predicate from the authorizer layer, and v1 is being retired.

Two hermetic tests over a token watch fixed at `PermissionDenied`: an unbound journal yields `Ok`, a bound one yields `Err`. The second pins the tolerance to the ephemeral path.

### DIFF
--- a/crates/publisher/src/publisher.rs
+++ b/crates/publisher/src/publisher.rs
@@ -282,8 +282,9 @@ impl Publisher {
     /// - Otherwise, build an ephemeral client. This supports recovered ACK
     ///   intents that may reference journals no longer bound to the current
     ///   task (e.g. from a prior published task). For this class of journals,
-    ///   a `NotFound` status is tolerated and handled by discarding the ACK
-    ///   (it was intentionally deleted, for example due to a data reset)
+    ///   `NotFound` is tolerated and handled by discarding the ACK (the journal
+    ///   was intentionally deleted, for example due to a data reset), and so is
+    ///   `PermissionDenied` (the task's grant to that collection was revoked).
     ///
     /// On Ok return, all ACKs have been durably appended (no need to flush).
     pub async fn write_intents<I>(&mut self, journal_intents: I) -> tonic::Result<()>
@@ -322,6 +323,25 @@ impl Publisher {
         self.appenders.flush().await?;
         self.appenders.sweep();
 
+        // Ephemeral appenders may fail-open, discarding their ACK.
+        //
+        // Only *recovered* intents reach this path: a live transaction's
+        // intents are keyed on journals it just wrote, and every such journal
+        // came from a binding. An ephemeral journal is therefore always one
+        // this task published to under an earlier specification.
+        //
+        // Fail-open is the only terminating choice for these. Nothing clears an
+        // intent but writing it, so failing isn't merely noisy: the session
+        // fails, recovers the identical intent, and fails again, forever. Nor
+        // can the task heal itself, because updating it to no longer publish to
+        // the journal is precisely what stranded the intent. We knowingly
+        // trade the documents behind a discarded ACK — which remain PENDING
+        // and are never released to readers — against a permanently dead task.
+        //
+        // `PermissionDenied` alone, because it alone denies *this object*: the
+        // task is known and its request well-formed, but its grant to that
+        // collection is gone (revoked, or left behind by a data-plane
+        // migration).
         let ephemeral_flushes = ephemeral
             .active_set()
             .map(|(journal, appender)| async move {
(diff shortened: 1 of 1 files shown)


williamhbaker (approved):
LGTM

PR #3269, opened 2026-07-27
Sync Schedule moved to runtime v2

dgreer-dev:
## Summary

Adds materialization sync schedules as a first-class model field enforced by
runtime-next, parallel to the connector-side feature. It's a faithful port of the
connector's `schedule` package, so the user-visible semantics (sync frequency,
fast-sync window/enabled-days, timezone, 4h fallback) are unchanged. Migrating
existing connector configs onto the model field is out of scope here.

Note this brings in the `chrono-tz` crate which is net-new for flow  in order to
support looking up by timezone name (ie, `America/New_York`)

Ticket: 3229

## What's here (two commits)

**1. `materialize sync schedule: model, built spec, and validation`**
- New `SyncSchedule` model type on `MaterializationDef`, with dependency-free
  shape validation and the shared day/offset parsers.
- `sync_schedule_json` on `MaterializationSpec` (proto field 12), mirroring
  `triggers_json`; regenerated Go/Rust protobufs and `flow.schema.json`.
- Validation serializes it into the built spec, runs shape validation, and
  rejects a publication that sets a schedule *both* as the model field and in
  the connector config.

**2. `runtime-next: evaluate and enforce sync schedules`**
- The schedule evaluator (`next_fire_after`) and the `close_policy` hold that
  pauses an open transaction until the next scheduled commit instant.
- `Task::new` wires it up, validates the IANA zone name, and derives the jitter
  seed from the sync group (else the task name).

## Design notes

- **`chrono-tz` is confined to `runtime-next`.** The model type + shape
  validation live in `models` with no timezone-database dependency, so the ~28
  crates that depend on `models` (including the flow-web WASM bundle) are
  unaffected. IANA zone *names* are validated at task startup instead of build.
- **No backfill detection.** A configured schedule is taken at face value. A
  large backfill still drains because the combiner usage ceiling forces commits
  under memory pressure, so we don't need to detect it -- and that avoids the
  connector heuristic's dependence on document size / reduction.
- **A schedule overrides `min`/`max_txn_duration`**, leaving the schedule and the
  usage ceiling as the only close drivers.
- **V2-only.** The field is enforced by runtime-next; on the v1 runtime it's a
  dormant no-op (forward-compatible: you can set it ahead of a task moving to V2).

## Testing

- Unit tests: the evaluator (grid/jitter/window/DST), the `close_policy`
  schedule gate (hold, usage-ceiling override, close-request override, wake).
- An FSM integration test driving a scheduled transaction held-then-released
  through the materialize Head FSM.
- Manual validation against the local stack.

## Out of scope / follow-ups

- Migrating existing connector-config sync schedules onto the model field.
- User-facing docs (deferred until we have an upgrade path for v1 connectors.)


crates/runtime-next/src/leader/materialize/task.rs
```diff
@@ -63,6 +64,31 @@ impl Task {
...
+            None
+        } else {
+            Some(
+                serde_json::from_slice(&sync_schedule_json)
+                    .context("parsing sync_schedule_json")?,
+            )
+        };
+        // Validate the IANA timezone name here (models validation only checked
+        // offset timezones, to avoid a timezone-database dependency there).
+        if let Some(tz) = sync_schedule.as_ref().and_then(|s| s.timezone.as_deref())
```
williamhbaker: I think we're going to want this validation to happen at spec build time rather than task run time.

crates/runtime-next/src/leader/materialize/task.rs
```diff
@@ -90,10 +123,24 @@ impl Task {
...
+            sync_schedule,
+            sync_seed,
         })
     }
 }
 
+/// FNV-1a hash of `bytes` to a stable, process-independent `u64` seed. Used only
+/// to phase the sync schedule's jitter, so any stable hash suffices; FNV avoids
+/// a dependency and never changes across builds.
+fn fnv1a(bytes: &[u8]) -> u64 {
```
williamhbaker: We generally use xxhash for hashing like this elsewhere.

crates/sources/tests/snapshots/schema_generation__catalog_schema_snapshot.snap
```diff
@@ -1429,6 +1440,63 @@ expression: "&schema"
...
+          "title": "Fast Sync Enabled Days",
+          "description": "Days of the week the sync frequency is effective (e.g. `M-F`,\n`Su-T,Th-S`). Outside these days, transactions commit every four hours.",
+          "type": [
+            "string",
+            "null"
+          ]
+        },
+        "fastSyncStartTime": {
+          "title": "Fast Sync Start Time",
+          "description": "Time of day (`HH:MM`) at which the sync frequency becomes effective.\nOutside the window, transactions commit every four hours.",
```
williamhbaker: As a general conceptual comment, I think we wan to the "slow sync" time period to be configurable as well. We default to 4 hours in connectors out of necessity, but it would be fine to stretch that out more with the runtime impl, and there is demand for allowing that I believe.

crates/sources/tests/snapshots/schema_generation__catalog_schema_snapshot.snap
```diff
@@ -1429,6 +1440,63 @@ expression: "&schema"
...
+          "title": "Fast Sync Stop Time",
+          "description": "Time of day (`HH:MM`) at which the sync frequency stops being effective.",
+          "type": [
+            "string",
+            "null"
+          ]
+        },
+        "syncFrequency": {
+          "title": "Sync Frequency",
+          "description": "Frequency at which transactions commit when the materialization is\ncaught up. Effective 24/7 unless a fast-sync window is configured below.\nA value of zero commits as fast as possible. Defaults to 30 minutes.",
```
williamhbaker: The 30 minute default thing is a historical artifact from the connector implementation.  I think we should require an input for these, rather than taking no input assumed to be 30 minutes.

crates/runtime-next/src/leader/sync_schedule.rs
```diff
@@ -0,0 +1,485 @@
...
+        || sched.fast_sync_start_time.is_some()
+        || sched.fast_sync_stop_time.is_some()
+        || sched.fast_sync_enabled_days.is_some();
+
+    if !alternating {
+        return fire(after, freq, jitter);
+    }
+
+    // Malformed alternating config is rejected during validation; degrade to
+    // the fixed frequency rather than panicking if it slips through.
```
williamhbaker: Erroring out if an invalid config somehow gets here seems like it would be better than silently falling back to a fixed frequency. It be easier for us to debug if such a thing were a loud failure.

crates/runtime-next/src/leader/materialize/sync_schedule.rs
```diff
@@ -0,0 +1,485 @@
+//! Evaluation of a materialization's [`models::SyncSchedule`]: the wall-clock
```
williamhbaker, line 1: Maybe move this to `leader/materialize/sync_schedule.rs`? Since it is not shared by all the task types, only materializations. It would be alongside the triggers module.

crates/runtime-next/src/leader/materialize/sync_schedule.rs
```diff
@@ -0,0 +1,485 @@
...
+        if alt_active(zone, days, start, end, midnight_tomorrow)
+            && (midnight_tomorrow < window_start
+                || !alt_active(zone, days, start, end, window_start))
+        {
+            return midnight_tomorrow;
+        } else if alt_active(zone, days, start, end, window_start) {
+            return window_start;
+        }
+    }
+    unreachable!("at least one day of the week must be active")
```
williamhbaker: I'm thinking there are cases involving DST where a day is skipped. Claude says:
>America/New_York, 09:00–17:00, fastSyncEnabledDays: Su, any anchor in a 780-minute band around 2026-03-08. A one-year sweep found 35 realistic (zone, window, day) combos that panic — including Europe/London+Su and Pacific/Auckland+Su.

This is probably a gap in the connector version as well.

crates/models/src/sync_schedule.rs
```diff
@@ -0,0 +1,259 @@
...
+        };
+
+        if (tz.starts_with('+') || tz.starts_with('-')) && parse_offset(tz).is_none() {
+            return Err(format!("invalid timezone offset {tz:?} (expected +HH:MM)"));
+        }
+        parse_clock(start)
+            .ok_or_else(|| format!("invalid fastSyncStartTime {start:?} (expected HH:MM)"))?;
+        parse_clock(stop)
+            .ok_or_else(|| format!("invalid fastSyncStopTime {stop:?} (expected HH:MM)"))?;
+        if start == stop {
```
williamhbaker: Does this work for `"09:00"` vs. `"9:00"`?

crates/runtime-next/src/leader/close_policy.rs
```diff
@@ -115,11 +123,23 @@ impl Policy {
...
+        // We want to extend while neither time-max nor usage-max is breached —
+        // but a schedule hold defers the time ceiling.
+        let policy_extend = !usage_above_max && (!time_above_max || holding);
         // We want to close if we're not extending, or time-min and usage-min is reached.
         let mut policy_close = !policy_extend || (usage_above_min && time_above_min);
 
+        // While holding for the schedule, only a usage ceiling (memory/disk
+        // safety) may close early; otherwise we wait for the commit instant.
+        if holding {
+            policy_close = usage_above_max;
```
williamhbaker: A requirement that hasn't been explicitly mentioned: We really want task spec updates to expediently cause the task to restart, even if it is amidst its scheduled holding period. That's maybe the single roughest edge of the connector-based version, since it artificially delays its `Acknowledged` message the runtime can't gracefully stop it on a spec update. So you get a situation where the spec is updated (maybe to update the sync schedule itself!), but it just hangs until the current schedule lets the ack drain. Even worse it might take 2 full cycles of this to complete for ~reasons.

There's a broader refactor I think should be looked at, and working the above into. The current implementation kind of short circuits / overrides the min/maxTxnDuration machinery, and I believe that mechanism would handle the spec change as we desire. I haven't looked at exactly what the alternative would be, but my sense is that if we try to ride closer to the path that min/maxTxnDuration already take we'd have an easier time acheiving this.

crates/runtime-next/src/leader/materialize/fsm.rs
```diff
@@ -1154,6 +1155,41 @@ pub struct TailDone {
...
+    task: &Task,
+    is_open: bool,
+    last_close: uuid::Clock,
+    now: uuid::Clock,
+) -> Option<Duration> {
+    let sched = task.sync_schedule.as_ref()?;
+
+    // Nothing to hold before a transaction is open; and the first transaction
+    // of a session (no prior close) is never held, mirroring the connector's
+    // "never delay the recovery commit".
```
williamhbaker: I think this comment is right, and that we shouldn't delay the recovery commit, but `last_close` is seeded from the recovered checkpoint so I don't think it will work as-is.

crates/runtime-next/src/leader/sync_schedule.rs
```diff
@@ -0,0 +1,485 @@
...
+        }
+    }
+    unreachable!("at least one day of the week must be active")
+}
+
+/// Midnight (UTC) of the calendar day containing `dt`, matching the connector's
+/// `atMidnight`.
+fn at_midnight(dt: chrono::DateTime<chrono::Utc>) -> chrono::DateTime<chrono::Utc> {
+    use chrono::{Datelike, TimeZone};
+    chrono::Utc
```
williamhbaker: I think this needs to be in the configured timezone? Using something other than `+00:00` in the tests may help catch it.

crates/validation/src/materialization.rs
```diff
@@ -1058,6 +1083,17 @@ fn temporary_group_by_migration(
         .collect()
 }
 
+/// Whether a connector endpoint config carries a top-level `syncSchedule` key,
+/// i.e. the legacy connector-side sync schedule. Detected without decryption:
+/// SOPS preserves object keys, so the key is present even when values are
+/// sealed.
+fn connector_config_has_sync_schedule(config: &models::RawValue) -> bool {
+    match serde_json::from_str::<serde_json::Value>(config.get()) {
+        Ok(serde_json::Value::Object(map)) => map.contains_key("syncSchedule"),
```
williamhbaker: I would double-check that it is actually possible to remove a sync schedule from a materialization config, for example via the UI. What I'm thinking of is a `syncSchedule: {}` where all the inputs are cleared, but it's still there as an empty map.

williamhbaker (commented):
Some initial higher level comments below

crates/runtime-next/src/leader/close_policy.rs
jgraettinger: My thinking at the time, when building closePolicy, was that we'd modulate the current values of minTxnDuration and maxTxnDuration based on the current schedule, but that these values would continue to be the lever -- we're just allowing them to dynamically very based on time-of-day instead of being fixed. This should work fine even if there's no source journal activity, because we re-poll at a minimum of ACTOR_TICK_INTERVAL (60s)

FYI task restarts are done by sending CloseNow into runtime-next, which short-circuits a long minTxnDuration (e.x. if we're in "slow mode" / 6h), so this should Just Work.

dyaffe:
Random question: Does this include the sync now API or functionality?  If so, do we have any idea what it will look like?

crates/runtime-next/src/leader/materialize/task.rs
dgreer-dev: Yeah I struggled with this.  The trick is we want `chrono-tz` to do the timezone validation, and putting that in `models` (where it would be most natural) pulls in that timezone database (not insignificant) into just about everything (including, like, `flow-web`s wasm bundle).

That seemed overkill to me.  I also considered feature-gating it and just turning on the feature for the minimal set, but that's a bit of TOML management.

Happy to put it in models, gated or not.  What do you think?

dgreer-dev:
> Random question: Does this include the sync now API or functionality? If so, do we have any idea what it will look like?

I'm planning that as the next ticket after this.  The vision is you use `flowctl` to do it.  It'll reach out to the control plane to get the right location in the data-plane to go to + a data-plane token, and then do an auth'd request directly to the reactor.  Don't quote me on that, but I think that's doable.

crates/runtime-next/src/leader/materialize/task.rs
dgreer-dev: 👍

crates/sources/tests/snapshots/schema_generation__catalog_schema_snapshot.snap
dgreer-dev: :+1:

crates/runtime-next/src/leader/materialize/sync_schedule.rs
dgreer-dev: Great call out

crates/models/src/sync_schedule.rs
dgreer-dev: Yep, it parses it as a `u32` so the leading zero shouldn't hurt.  I'll add a test to prove it so we don't have to assume though.

crates/runtime-next/src/leader/sync_schedule.rs
dgreer-dev: Yep you're right, this should be local midnight not UTC.  FWIW the existing connector logic does UTC, but yeah I agree that probably was in error.

crates/runtime-next/src/leader/materialize/sync_schedule.rs
dgreer-dev: 👍

crates/runtime-next/src/leader/materialize/fsm.rs
dgreer-dev: Good point -- it works for the first ever commit but that's it.  Fixing it.

crates/validation/src/materialization.rs
dgreer-dev: Yep, leaving `{}` is exactly what happens.  Fixing it so we treat that as not in conflict.

crates/runtime-next/src/leader/sync_schedule.rs
dgreer-dev: 👍

crates/sources/tests/snapshots/schema_generation__catalog_schema_snapshot.snap
dgreer-dev: Agreed, I'd like to get a value in there.  One sort of issue that was preventing me, the naming of all this stuff is a bit weird.  My initial thought was `slowSyncFrequency` but man that is a confusing phrase and it's killing me haha.

The current field names inherit two problems from the connector config: "frequency" for values that are actually intervals (30m is a period, not a rate), and a fast/slow vocabulary where the primary field (syncFrequency) doesn't carry the fastSync prefix its siblings do, while the "slow" regime is invisible and hardcoded. Since the model-level schedule is brand new with zero users, I'd like to fix the names now rather than extend the old scheme. 

Proposal:

```
syncSchedule:
  interval: 4h                 # commit cadence at all times not covered by `window`; required
  group: acme-warehouse        # optional: tasks sharing a group commit at the same instants
  window:                      # optional: a recurring local-time window with its own cadence
    interval: 15m              # commit cadence within the window; required
    start: "09:00"             # local time of day the window opens; required
    stop: "17:00"              # local time of day the window closes; required
    days: M-F                  # days of week the window applies; defaults to all days
    timezone: America/New_York # IANA name or +HH:MM offset; required
```

What do you think?  We'd still be able to mechanically turn old-connector-style in to this proposed style.

williamhbaker: My first impression is that I was a bit confused seeing "interval" at the top of the list, and trying to square that with my expectation that I'd be setting how fast it should go during the fast time, until I read the rest of the config and determined it actually meant the slow time, and the fast time was in "window".

In all honesty, that might just be because my brain is wired strangely, and I'm having a hard time coming up with anything I think is objectively better.

crates/models/src/sync_schedule.rs
williamhbaker: I guess what I'm wondering here is if the validation check works, since it is comparing strings and those would compare differently, even though they are the same.

dgreer-dev: Darn, yeah you're exactly right.  Fixed.

crates/runtime-next/src/leader/materialize/task.rs
williamhbaker: Feels like it is 100% necessary to have this be an in-your-face build time error if an incorrect timezone is put in, since if it doesn't error until it is running on the reactor it becomes a platform problem rather than a user input problem. So I'd probably opt for just putting it in the models crate. I did some really quick checks and the compile time is not nothing but I don't think it justifies the feature gating.

crates/runtime-next/src/leader/close_policy.rs
dgreer-dev: Yeah okay, this all makes sense to me.  I think it functions as written, but using the min/max txn numbers is a bit cleaner and I think will result in simpler code.  I'll get it in.

crates/runtime-next/src/leader/materialize/task.rs
dgreer-dev: 👍

crates/sources/tests/snapshots/schema_generation__catalog_schema_snapshot.snap
dgreer-dev: Resolved this, see the next PR comment for info

dgreer-dev:
Okay, this should address all the review feedback, plus the naming/shape discussion that moved to Slack, plus one addition that came out of that discussion: multiple windows ([reasoning here](https://estuaryworkspace.slack.com/archives/C03Q9A3GFEE/p1785300377684429?thread_ts=1785258821.760909&cid=C03Q9A3GFEE)).

Heads up: I rationalized the history back down to two commits (model/validation, then runtime enforcement) and force-pushed, so the inline threads will show as outdated. The commit messages narrate the design and the fixes in detail.

The model shape is now:

    syncSchedule:
      baseInterval: 4h              # required
      timezone: America/New_York    # required iff windows are set
      group: shared-warehouse       # optional jitter-coordination key
      windows:                      # optional, non-overlapping
        - interval: 15m
          start: "09:00"
          end: "17:00"
          days: [Mon, Tue, Wed, Thu, Fri]

Notes on the shape:

- `baseInterval` replaces both the hardcoded 4h slow interval (now fully configurable) and the implicit 30m default (now required).
- Any number of windows, each with its own interval and day mask. A window can be slower than the base (e.g. quiet hours), which is why the fast/slow naming is gone. Overlap is rejected at build time on the weekly wall-clock grid; adjacent windows are fine.
- `timezone` is schedule-level and shared by all windows. Per-window timezones would make the overlap check undecidable across DST (two zones shift relative to each other during the year), and a per-window override can be added later non-breaking if anyone ever needs it.
- `days` is a list of `chrono::Weekday` ("mon" / "Monday" parse fine, serializes "Mon"), replacing the `M-F` mini-dialect.
- Requiredness is structural (serde + JSON schema `required`) rather than custom validation.

Review items folded in:

- Timezone (including IANA names) validates at build time: models takes chrono-tz directly. flow-web's wasm build verified OK.
- Close policy rides min/maxTxnDuration as the levers: the open-duration band is an `evaluate()` input which the materialize FSM modulates per evaluation when a schedule is set. `schedule_hold` is gone and `evaluate()` has no schedule awareness. CloseNow punches through a hold the same way it punches through a long minTxnDuration.
- Recovery commits are never held (explicit session-start flag; `last_close` comes from the recovered checkpoint so it couldn't tell).
- The DST day-skip panic and UTC-vs-local midnight issues are fixed (local calendar scan), including the window-exit midnight case from the connector's `nextEndOfAlt` that the original port dropped. A year-long sweep over pathological zones (Santiago, Lord Howe) guards the calendar edges.
- Equal window times compare parsed values, so "9:00" vs "09:00" is rejected as the empty window it is.
- xxhash for the jitter seed, evaluator moved to `leader/materialize/`, malformed-config paths panic instead of silently degrading, and a cleared-out connector-side `syncSchedule: {}` (what the UI leaves behind on removal) no longer counts as a conflict.

crates/runtime-next/src/leader/materialize/sync_schedule.rs
williamhbaker: I'm realizing there is another case here, where if you have a window that is 1 hour long and starts in the hour that DST deletes, and only has 1 day configured on Sunday, you could also reach this. Okay...this is pretty absurd, so probably fine to ignore, but food for thought!

crates/runtime-next/src/leader/materialize/sync_schedule.rs
```diff
@@ -0,0 +1,773 @@
...
+/// (spring-forward gap) resolves forward past the one-hour gap; an ambiguous
+/// one (fall-back) takes the earlier instant.
+fn resolve_local<Tz: chrono::TimeZone>(
+    tz: &Tz,
+    naive: chrono::NaiveDateTime,
+) -> chrono::DateTime<chrono::Utc> {
+    match tz.from_local_datetime(&naive).earliest() {
+        Some(dt) => dt.with_timezone(&chrono::Utc),
+        None => tz
+            .from_local_datetime(&(naive + chrono::Duration::hours(1)))
```
williamhbaker: Here's a fun one...`Antarctica/Troll` actually jumps forward 2 hours. Also Lord Howe shifts 30 minutes. These are a bit out there, but I think there's a fairly simple refactoring here where we could instead bisect forward to find the next resolvable local time to make this more robust.

dgreer-dev: Yes, all this you say is correct, but I think I've got it sorted now.

Bisecting is right, but causes it's own issue that'll I'll just quote Claude on:

> One knock-on worth flagging, and it's your earlier "1-hour window inside the
deleted hour, Sunday only" case: bisection makes that one *reachable*. When
start and end both land in the same gap they now collapse onto one instant,
so the window is inactive on that date instead of inverted -- which means the
occurrence is lost, and if Sunday is the only enabled day, an 8-day scan
anchored on the Saturday before spans 03-07..03-14 and hits the
`unreachable!`. The fixed hour was accidentally masking it (02:15-02:45
became 03:15-03:45). So `next_start_of_window` now scans 15 days: a zone
transitions a few times a year, so two shots at each weekday is always
enough.

In short, I'm going to move to Troll Station and work to ban timezones.  We need to shift our marketing budget to norwegian research stations, I bet they have a ton of buggy software.

williamhbaker:
I didn't have time to finish the review, here are the areas I was still investigating, maybe you can look into these and see if there's anything there:

- I haven't been able to convince myself that the task will gracefully restart on a spec change if it is still in its minimum hold time. I'm thinking there's a pre-existing gap even with the `minTxnDuration` mechanism. I'd suggest running a manual test and verifying this works. There may be a pre-existing issue to fix, and if there is we should fix it now since this feature is sensitive to that working.
- I'm think there's an issue with the "group" mechanism with the grid anchored on `last_close`, once an interval has elapsed the target is already past → `unheld`. We aren't trying to guarantee spacing between commits, but want to find the next point on the grid that the next commit should land, so should this be anchored on `now`?

dgreer-dev:
> * I haven't been able to convince myself that the task will gracefully restart on a spec change if it is still in its minimum hold time. I'm thinking there's a pre-existing gap even with the `minTxnDuration` mechanism. I'd suggest running a manual test and verifying this works. There may be a pre-existing issue to fix, and if there is we should fix it now since this feature is sensitive to that working.

Yes, this was real (and pre-existing).  Added tests + fixed it.

> * I'm think there's an issue with the "group" mechanism with the grid anchored on `last_close`, once an interval has elapsed the target is already past → `unheld`. We aren't trying to guarantee spacing between commits, but want to find the next point on the grid that the next commit should land, so should this be anchored on `now`?

Yeah you are correct.  Fixed as suggested.

crates/runtime-next/src/leader/materialize/fsm.rs
```diff
@@ -1154,6 +1168,73 @@ pub struct TailDone {
...
+    is_open: bool,
+    session_start: bool,
+    open: uuid::Clock,
+    now: uuid::Clock,
+    open_age: Duration,
+) -> std::ops::Range<Duration> {
+    if task.sync_schedule.is_none() {
+        return task.close_policy.open_duration.clone();
+    }
+    let unheld = Duration::ZERO..Duration::MAX;
```
williamhbaker: With an interval of 0s, we'll always return `unheld` at line 1222. In this case I don't think we'd want `MAX` for the maximum transaction duration, and would be better off using `Duration::ZERO..task.close_policy.open_duration.end`, as the fallback to the min/maxTxnDuration. Using `MAX` means a source that is steadily getting newly appended data will just keep on growing until the combiner byte limit is hit, which is a non-ideal experience if you have a very busy source and are wanting the data to be as fresh as possible in the destination. So having some maxTxnDuration here is important I think.

crates/runtime-next/src/leader/materialize/sync_schedule.rs
```diff
@@ -0,0 +1,924 @@
...
+        .unwrap_or(base);
+
+    // Within the current regime, use its cadence. When the next fire would
+    // cross the nearest boundary AND the regime taking over there is faster,
+    // clamp to the boundary so the faster cadence starts on time; crossing
+    // into a slower regime uses the computed instant as-is. (A fire may
+    // overleap several regimes; each commit re-anchors, so evaluation
+    // converges one boundary per commit.)
+    let n = fire(after, this_period, jitter)?;
+    if n >= bound && next_period < this_period {
```
williamhbaker: This checks the regime at the nearest boundary only, so a fire that hops over a short slower window into a faster one is never clamped.

```yaml
syncSchedule:
  baseInterval: 15m                                    # daytime freshness
  timezone: "+00:00"
  windows:
    - { interval: 4h,  start: "22:00", end: "06:00" }  # overnight batching
    - { interval: 12h, start: "06:00", end: "06:30" }  # maintenance blackout
```

Anchored at 04:00, inside the overnight window: bound is 06:00, next_period is the blackout's 12h which isn't < 4h, so no clamp — and fire(04:00, 4h) gives 08:00. The hold runs 04:00 → 08:00, so the 15m daytime cadence starts 90 minutes late, every morning. Nothing re-anchors in between, since the hold is exactly what prevents the commit that would.

I think we can fix this with some kind of walk / loop over the intervals.

crates/runtime-next/src/leader/materialize/sync_schedule.rs
```diff
@@ -0,0 +1,924 @@
...
+
+    // Without windows, the base interval is effective 24/7.
+    if sched.windows.is_empty() {
+        return fire(after, base, jitter);
+    }
+
+    // A malformed schedule is rejected at build time (and re-checked at task
+    // startup), so reaching one here is an impossible state: fail loudly
+    // rather than silently degrading to the base interval.
+    let Some(zone) = sched.timezone.as_deref().and_then(Zone::parse) else {
```
williamhbaker: This recompiles the whole schedule on every call. I'm thinking we could pre-parse this and put it alongside `CompiledTriggers` on the `Task` itself. We could have a fallible constructor for that and clean up some of these `panic` assertions. I don't think there's a huge perf benefit, but it would get rid of a tiny bit of wasted work.

crates/runtime-next/src/leader/materialize/sync_schedule.rs
```diff
@@ -0,0 +1,924 @@
...
+
+/// A time of day (hour, minute) for a window, in the schedule's zone.
+#[derive(Clone, Copy)]
+struct ClockTime {
+    h: u32,
+    m: u32,
+}
+
+impl ClockTime {
+    fn parse(s: &str) -> Option<ClockTime> {
```
williamhbaker, line 261: This reimplements `models::sync_schedule::parse_clock` verbatim, and `active_days` (line 342) re-derives the day mask that `validate()` builds inline.

I think the "CompiledSchedule-on-Task" suggestion will clear this up though.

crates/runtime-next/src/leader/materialize/actor.rs
```diff
@@ -89,6 +89,17 @@ impl<P: crate::Publisher, L: crate::Logger> Actor<P, L> {
...
             "materialize Actor::serve started",
         );
+        // Surface that this task paces its commits to a sync schedule, once per
+        // leader session. The per-transaction cadence is observable via the
+        // transaction stats; this just makes the pacing visible in default logs.
+        if let Some(sync_schedule) = &self.task.sync_schedule {
+            service_kit::event!(
+                tracing::Level::INFO,
+                "leader",
+                sync_schedule = format!("{sync_schedule:?}"),
```
williamhbaker: `sync_schedule = service_kit::event::json(sync_schedule.clone())` here?

crates/models/src/sync_schedule.rs
```diff
@@ -0,0 +1,415 @@
...
+pub struct Window {
+    /// # Interval
+    /// Interval between transaction commits while the window is active.
+    /// A value of zero commits as fast as possible.
+    #[schemars(schema_with = "super::duration_schema")]
+    #[serde(with = "humantime_serde")]
+    pub interval: Duration,
+    /// # Start
+    /// Local time of day (`HH:MM`) at which the window opens.
+    pub start: String,
```
williamhbaker, line 53: Are we able to add schema `pattern` annotations for `start` and `end`?

crates/models/src/sync_schedule.rs
```diff
@@ -0,0 +1,415 @@
...
+/// A recurring local-time window and the commit interval effective within it.
+#[derive(Serialize, Deserialize, Clone, Debug, JsonSchema, PartialEq)]
+#[serde(deny_unknown_fields, rename_all = "camelCase")]
+pub struct Window {
+    /// # Interval
+    /// Interval between transaction commits while the window is active.
+    /// A value of zero commits as fast as possible.
+    #[schemars(schema_with = "super::duration_schema")]
+    #[serde(with = "humantime_serde")]
+    pub interval: Duration,
```
williamhbaker, line 49: This is getting schematized as `type: ["string", "null"]`, which seems like it should be just `type: ["string"]`.

crates/runtime-next/src/leader/materialize/sync_schedule.rs
```diff
@@ -0,0 +1,924 @@
+//! Evaluation of a materialization's [`models::SyncSchedule`]: the wall-clock
```
williamhbaker, line 1: It would be good to add this file and some description about the sync schedule to the crate README.

williamhbaker (commented):
A few more comments with what I was able to get through this afternoon. I'll take another pass on Monday and hopefully finish up then.

crates/runtime-next/src/leader/materialize/sync_schedule.rs
```diff
@@ -0,0 +1,924 @@
...
+//! instant at which a held-open transaction is next permitted to commit.
+//!
+//! This is a port of the connector's `schedule` package, generalized from one
```
williamhbaker: A minor thing in these comments...we probably shouldn't reference the connector's code long-term, since we'll be deleting that eventually. So I don't think the comments referring to the port or mirroring of that are going to hold up well.

williamhbaker (commented):
Just one final comment! This is looking good, let me know when you've addressed this final round of feedback and I'll take one last look.

crates/runtime-next/src/leader/materialize/sync_schedule.rs
dgreer-dev: Yeah this was a real oversight on my part.  Good call out.

crates/runtime-next/src/leader/materialize/sync_schedule.rs
dgreer-dev: Yeah pretty perf-neutral, but way cleaner.  Good call out.

crates/models/src/sync_schedule.rs
dgreer-dev: We can, in fact!

williamhbaker:
The latest commit got rid of the `group` field altogether, and in place computes jitter based off the connector image name. I think that's too coarse...maybe instead of bringing the `group` field back, we could base the jitter off the tenant name, using `models::tenant_from`? I think that would spread things out pretty well. It would only miss the cross-tenant coordination case which is something we can address later if it's really in demand.

dgreer-dev:
> The latest commit got rid of the `group` field altogether, and in place computes jitter based off the connector image name. I think that's too coarse...maybe instead of bringing the `group` field back, we could base the jitter off the tenant name, using `models::tenant_from`? I think that would spread things out pretty well. It would only miss the cross-tenant coordination case which is something we can address later if it's really in demand.

Yep, got this in.   Also collapsed down to two commits.

williamhbaker (approved):
LGTM, awesome!

PR #3271, opened 2026-07-28
docs: fix allow_existing_tables_for_new_bindings backfill claim, add binding migration steps

jwhartley:
**Description:**

Fixes an inaccurate caveat on the `allow_existing_tables_for_new_bindings` materialization feature flag, and adds a step-by-step for its most common real use case.

**Workflow steps:**

- Removed "Only Changes" backfill mode as a suggested alternative to `notBefore`. It's a capture-side setting (skips the initial table scan for future captures) and isn't retroactive, so it can't stop a materialization binding from replaying collection data that's already been captured. Confirmed against `sqlcapture/capture.go` and the shuffle/broker read-range code (`crates/shuffle`, `go/shuffle`, Gazette's `fragment/index.go`): a new binding's read range is governed entirely by `notBefore`, nothing about the capture's backfill mode reaches it.
- Added "Moving a binding to a new materialization" under the flag's section: pause writes, drain and disable the old binding, add the new binding with the flag plus `notBefore` set from the pause window, resume writes. This is the standard pattern for consolidating or splitting materializations without dropping or replaying existing destination data.
- Added a caveat clarifying the flag never truncates or drops the existing table (confirmed in `materialize-boilerplate/apply.go` + `materializer.go`: a new binding whose destination table already exists is reconciled additively, new columns added / widened, never recreated).
- Cross-linked from `backfilling-data.md`'s "Preserving existing data during a backfill" section.

- [Feature Flags](https://docs.estuary.dev/guides/advanced-usage/feature-flags/#allow_existing_tables_for_new_bindings)
- [Backfilling Data](https://docs.estuary.dev/guides/backfilling-data/#preserving-existing-data-during-a-backfill)

**Notes for reviewers:**

Several Kapa-indexed community threads repeat the same "Only Changes" suggestion in this exact context; flagging those separately for correction, not in scope for this PR.

### DIFF
--- a/site/docs/guides/advanced-usage/feature-flags.md
+++ b/site/docs/guides/advanced-usage/feature-flags.md
@@ -122,19 +122,7 @@ Allows materializations to write to tables that already exist in the destination
   - This flag alone does **not** prevent backfill of the source collection. A binding that's new to a materialization always starts reading its source collection from the beginning, the same as any other backfill. To avoid replaying data that's already in the existing table, also configure [`notBefore`](/reference/time-travel) on the binding.
   - This flag does not truncate or drop the existing table. A new binding whose destination table already exists is reconciled the same way an existing binding is when its schema changes: new columns are added and existing columns are widened as needed, but no data is removed.
 - **Applies to:** All SQL and warehouse materialization connectors (PostgreSQL, MySQL, Snowflake, BigQuery, Redshift, etc.)
-
-#### Moving a binding to a new materialization
-
-A common use for this flag is moving a binding from one materialization to another, for example when consolidating several materializations into one, or splitting a large one apart, without dropping or reprocessing the destination table's existing data.
-
-To do this without duplicating or losing data:
-
-1. Pause writes to the source (or disable the capture) and let the current materialization fully drain its backlog for that binding.
-2. Disable the binding on the original materialization.
-3. Add the binding to the new materialization:
-   - Set the `allow_existing_tables_for_new_bindings` feature flag under **Endpoint Config → Advanced** on the new materialization.
-   - Set [`notBefore`](/reference/time-travel) under **Binding → Advanced** on the new binding, to a timestamp within the pause window from step 1, or to just after the last `flow_published_at` already written to the destination table.
-4. Resume writes to the source (or re-enable the capture). New data flows through the new binding only; `notBefore` causes the runtime to skip re-reading everything published before it.
+- **Task guide:** [Moving a binding to a new materialization](/guides/move-binding-to-new-materialization)
 
 ### retain_existing_data_on_backfill
 
--- a/site/docs/guides/advanced-usage/feature-flags.md
+++ b/site/docs/guides/advanced-usage/feature-flags.md
@@ -119,8 +119,22 @@ Allows materializations to write to tables that already exist in the destination
 - **Caveats:**
   - Enabling this flag makes the connector load existing keys from the destination before merging, instead of skipping that lookup for keys it expects to be new. It is slightly slower but ensures merges and updates work correctly against rows that already exist in the table.
   - The connector cannot verify that existing table schemas are compatible.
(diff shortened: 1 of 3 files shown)


site/docs/guides/advanced-usage/feature-flags.md
```diff
@@ -119,9 +119,23 @@ Allows materializations to write to tables that already exist in the destination
...
   - The connector cannot verify that existing table schemas are compatible.
-  - This flag alone does **not** prevent backfill of the source collection. To avoid backfilling data into the existing table, also configure [`notBefore`](/reference/time-travel) or use "Only Changes" mode on the binding.
+  - This flag alone does **not** prevent backfill of the source collection. A binding that's new to a materialization always starts reading its source collection from the beginning, the same as any other backfill. To avoid replaying data that's already in the existing table, also configure [`notBefore`](/reference/time-travel) on the binding.
+  - This flag does not truncate or drop the existing table. A new binding whose destination table already exists is reconciled the same way an existing binding is when its schema changes: new columns are added and existing columns are widened as needed, but no data is removed.
 - **Applies to:** All SQL and warehouse materialization connectors (PostgreSQL, MySQL, Snowflake, BigQuery, Redshift, etc.)
 
+#### Moving a binding to a new materialization
```
aeluce: Perhaps slightly odd to nest this guide section under the feature flag page, which otherwise is more of a reference sheet of available flags. I don't have strong feelings about a better location for it, though. Unless it would fit under the "Preventing backfills" section on the main backfill page?

aeluce (dismissed):

site/docs/guides/advanced-usage/feature-flags.md
jwhartley: I am likely to add more tactical 'how to do <task>' information to the docs over time as regular + complex edge cases like this come up, perhaps a good time to decide on a new location for them? Then I can cross-link a reference here instead

jwhartley: Moved it out to its own page: `site/docs/guides/customize-dataflows/move-binding-to-new-materialization.md`, under the existing Customizing Data Flows sub-section. The feature flags page now just carries a **Task guide:** link on the flag, so it stays a reference sheet, and the backfill page links to the same page.

Following up on the Slack thread: task guides go under Customizing Data Flows rather than directly under Using Estuary or in Troubleshooting, since these are about changing task behavior rather than fixing it. I gave the page an explicit slug (`/guides/move-binding-to-new-materialization`) so the URL holds if you rearrange that section later.

While reviewing the moved content against the connector source I found a few things worth correcting, so the page now also covers:

- **The flag has to stay enabled.** The validation gate in `materialize-boilerplate/validate.go` only fires while `lastBinding == nil`, but `materializer.go` sets `DisableLoadOptimization` from the endpoint-level flags. Removing the flag after the move publishes cleanly and silently re-enables load-skipping against rows the binding never wrote, so merges become inserts with no error.
- **A drain verification step.** `notBefore` suppresses document append outright rather than deferring it, so an undrained tail is lost at cutover. There's now a `MAX(flow_published_at)` check to anchor the timestamp.
- **A note on incompatible column types**, which fail the publication and offer a backfill, recreating the table and defeating the point of the exercise.

One thing I did not change, since it predates this PR: the **Applies to** line on `allow_existing_tables_for_new_bindings` says "All SQL and warehouse materialization connectors". The flag is checked in `materialize-boilerplate`, not `materialize-sql`, so MongoDB, DynamoDB, and Elasticsearch support it too. That also contradicts the `always_drop_tables_on_backfill` entry just below, which does name them. Happy to fix it here or leave it for a separate pass, whichever you prefer.

jwhartley:

as-is. For this PR:

- `site/docs/guides/advanced-usage/feature-flags.md` becomes `content/guides/advanced-usage/feature-flags.md`
- `site/docs/guides/backfilling-data.md` becomes `content/guides/backfilling-data.md`
- `site/docs/guides/customize-dataflows/move-binding-to-new-materialization.md` becomes `content/guides/customize-dataflows/move-binding-to-new-materialization.md`

aeluce:
Addressed in https://github.com/estuary/docs/pull/54

PR #3272, opened 2026-07-28
docs: fix broken anchor, enforce anchor checking, document macOS build gotcha

jwhartley:
## What

1. Fixes the one broken anchor found site-wide: `firstwritewins-and-lastwritewins.md` linked `./#reduction-guarantees` (itself) instead of the reduction-strategies overview page, where that heading lives. (Resolved via merge with master, which had since fixed the same link independently using an absolute path — that version is what's in the final diff.)
2. Flips `onBrokenAnchors` from `warn` to `throw` in `docusaurus.config.js`, now that the only offender is fixed, so future broken anchors fail the build instead of shipping silently.
3. Documents a pre-existing `hubspot-real-time` redirect/page collision in `site/README.md`: the redirect's `from` path differs from the real `HubSpot-real-time` page only by case, so `npm run build` fails on macOS's case-insensitive filesystem with "the redirect plugin is not supposed to override existing files." Confirmed this does not affect production (Linux CI is case-sensitive), so it's a local-dev note, not a code change.

## Why

Found while investigating stale-anchor feedback on #3094 — a heading rename there broke an anchor link silently since `onBrokenAnchors` only warns. Verified locally that flipping it to `throw` catches that class of bug without any other backlog (confirmed via full `npm run build` before and after).

## Testing

- `npm run build` locally: only failure is the pre-existing hubspot-real-time macOS collision (unrelated, documented, not present in CI).
- Confirmed via grep this is the only `reduction-guarantees` broken anchor pattern in the docs tree.


aeluce (approved):
Agree that we should eventually fail builds with broken anchors. Would it make sense to start this practice with the separate docs repo? I believe we've held off previously so failing docs checks wouldn't hold up feature releases (though theoretically if we've squashed all the anchor issues for now… 🤞).

PR #3273, opened 2026-07-28
runtime-next: stop constructing per-iteration timers in actor loops

jgraettinger:
`runtime-next`'s actor event loops constructed a fresh `tokio::time::sleep` (or
`sleep_until`) inside their `select!` on *every* iteration, registering a timer
into — and on drop removing it from — the tokio runtime's single global timer
wheel (one wheel per runtime, not sharded per worker).

`wake_after` is also frequently `Duration::ZERO`, meaning the FSMs have
synchronous work queued and want to be stepped again immediately. That was not
free: `TimeSource::deadline_to_tick` adds `999_999ns` before truncating to
milliseconds, rounding every deadline up to the end of the current millisecond.
So the loop parked for up to 1ms per step, capping the input-free FSM sequences
taken at every transaction boundary (rotate → drain → persist → write-intents).

## Measurements

`flowctl raw preview-next` over `tests/soak/capture/flow.yaml` (reconfigured to
`rate: 0` so the connector emits flat-out rather than paced at 20 docs/s),
sampling `runtime_shard_capture_transactions` from `--debug-port /metrics` over a
15s window after a 6s warmup:

| | txn/s | docs/s |
|---|---|---|
| `master` | 1.0, 1.1 | 6382, 6179 |
| + commit 1 | 147.0 | 111052 |
| + commit 2 | 139.0, 141.7, 136.6 | 110412, 106883, 110617 |

**~17x on document throughput** (6.3k → 110k docs/s).

Read the transaction figure carefully. Captures default to `max_txn_duration` of
1s with no minimum (`crates/assemble/src/lib.rs`); on `master` the loop could not
reach a transaction's *usage* ceiling within that second, so every transaction
closed on the time limit — hence a flat 1.0 txn/s. Afterward they close on usage
in ~7ms. So 17x on docs/s is the honest headline; the ~130x on txn/s is that plus
release from the time cap.

Commit 1 delivers the entire measured win. **Commit 2 is unmeasured by this
probe, by construction**: the soak capture has no injected IAM credentials, so
`token_restart_at` is `None` and the `maybe_deadline(None)` it replaces was
already a free `std::future::pending()`. Its benefit applies only to tasks that
*do* carry injected credentials. The 147 → 139 delta between commits is inside
the run-to-run spread (136.6–147.0 at HEAD).

## Commit 1 — `sleep_unless_zero`, with a hoisted `Sleep`

Returns immediately on `ZERO`; otherwise resets a `Sleep` hoisted out of the
caller's loop, which lands on `TimerEntry::extend_expiration`'s lock-free CAS so
only a loop's first non-zero iteration takes the wheel mutex. Applied to the
three `wake_after` loops: `shard/capture`, `leader/materialize`, `leader/derive`.

Two deliberate choices, both documented at the definition:

- **Plain return on `ZERO`, not `yield_now()`.** The mpsc arms polled by every
  iteration consume tokio's cooperative budget, forcing a yield within ~128
  iterations regardless; bursting between those yields is wanted for cache and
  branch behavior.
- **Non-zero durations are never collapsed**, even below the 1ms tick. They are
  deadline-driven — `leader::close_policy` reports time remaining until its
  nearest threshold, shrinking to microseconds as it nears — so a sub-ms
  threshold would spin re-evaluating a deadline that has not passed.

## Commit 2 — hoist the IAM token-restart timers

Same technique for the `token_restart_at` arm in `shard/capture` and
`shard/materialize`, which is ungated and hot (once per FSM step; once per
inbound connector message, respectively). Absent injected credentials there is no
deadline, so the hoisted `Sleep` is armed at `far_future()` rather than keeping an
`Option` — which also supplies the disarm the arm now needs, since a fired `Sleep`
stays `Ready` and would otherwise win on every later iteration.
`token_restart_at` becomes write-once, and both copies of `maybe_deadline` lose
their only caller and are deleted.

## Notes

- Behavior-preserving throughout; no protocol, spec, or persisted-state changes.
- `shard/derive` and `shard/materialize`'s `wake_after`-less loops are untouched:
  they already `continue` past the `select!` for synchronous work and use a
  long-lived `Interval` for the idle tick, which is this same pattern.
- The cold `sleep` sites (`triggers.rs` webhook backoff, `container.rs` readiness
  timeout and image-pull retry) are left alone — none can receive `ZERO` and none
  is in a loop.
- 151 `runtime-next` unit tests pass; both commits build independently.
  `tests/split_e2e.rs` needs `etcd` and `gazette` on PATH and did not run
  locally; it drives its own transaction loop, not the actors'.

### DIFF
--- a/crates/runtime-next/src/leader/derive/actor.rs
+++ b/crates/runtime-next/src/leader/derive/actor.rs
@@ -95,6 +95,9 @@ impl<P: crate::Publisher, L: crate::Logger> Actor<P, L> {
         let mut stopping = false;
         // Transactions completed in this task session, for preview harness limits.
         let mut transactions_completed = 0usize;
+        // Timer for the lowest-priority arm; `sleep_unless_zero` resets it
+        // before each await, so this initial deadline is never observed.
+        let mut wake_sleep = std::pin::pin!(tokio::time::sleep(Duration::ZERO));
 
         while !matches!(head, fsm::Head::Stop) {
             loop_count += 1;
@@ -271,7 +274,7 @@ impl<P: crate::Publisher, L: crate::Logger> Actor<P, L> {
                 }
 
                 // Lowest priority.
-                _ = tokio::time::sleep(wake_after) => {}
+                _ = crate::sleep_unless_zero(wake_sleep.as_mut(), wake_after) => {}
             }
 
             if !wake_after.is_zero() {
--- a/crates/runtime-next/src/leader/materialize/actor.rs
+++ b/crates/runtime-next/src/leader/materialize/actor.rs
@@ -117,6 +117,9 @@ impl<P: crate::Publisher, L: crate::Logger> Actor<P, L> {
         let mut stopping = false;
         // Transactions completed in this task session, for preview harness limits.
         let mut transactions_completed = 0usize;
+        // Timer for the lowest-priority arm; `sleep_unless_zero` resets it
+        // before each await, so this initial deadline is never observed.
+        let mut wake_sleep = std::pin::pin!(tokio::time::sleep(Duration::ZERO));
 
         while !matches!(head, fsm::Head::Stop) {
             loop_count += 1;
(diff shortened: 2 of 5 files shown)


williamhbaker (approved):
LGTM

Issue #3246, opened 2026-07-23
runtime-v2: shuffle recovery can deadlock when binding priorities split cohorts under a bounded disk limit


jgraettinger:
## Summary

V2 shuffle recovery can enter a non-converging retry loop when a task mixes binding priorities (multiple read cohorts), an idempotent recovery transaction, and a bounded `estuary.dev/shuffle-disk-limit`. High-priority cohorts finish their portion of the recovery replay first and then tail past the hinted frontier into post-crash content that cannot be consumed until recovery completes. That content pins log segments without bound, eventually engages disk back-pressure globally, and stalls all reads — including the low-priority cohort that actually gates recovery completion. The causal-hint resolution timeout then fires, the attempt tears down with **zero persisted progress**, and the next attempt restarts from scratch against an even larger post-crash backlog.

Observed in production on a materialization with several hundred source journals, a handful of high-priority bindings, and a large default-priority cohort, recovering a hinted transaction of roughly 50 GB. Attempts at 2 GB and 10 GB disk limits looped indefinitely; raising the limit to 30 GB let the replay complete. The disk raise works because the pinned over-read is bounded by (downtime × live write rate of the high-priority journals) — but that bound grows with every failed attempt and is unknowable in advance, so it is provisioning roulette rather than a fix.

## Mechanisms involved

Each of these is individually by design; it is their interaction that deadlocks.

1. **Replay order is priority-first across cohorts, not global clock order.** Documents drain by `(priority DESC, adjusted_clock ASC)` (`crates/shuffle/src/slice/heap.rs`, `crates/shuffle/src/log/actor.rs`). A cohort is the set of bindings sharing a `(priority, read_delay)` tuple. Back-pressure propagates from the Log heap to Slice journal reads, enforcing priority system-wide.

2. **Recovered reads have no end bound.** `StartRead` is issued with `end_offset: 0` (`crates/shuffle/src/slice/actor.rs`), so a journal that finishes its hinted region keeps tailing to the live write head.

3. **A recovery consumption ceiling caps visibility at the hint.** `Frontier::resolve_hints` caps `last_commit` at `hinted_commit` (`crates/shuffle/src/frontier.rs`), and the client's scan only yields entries with `clock <= last_commit` (`crates/shuffle/src/log/reader/scan.rs`). Content past a producer's hint becomes a `Remainder`, which pins its log segments until recovery completes.

[...]

PR #3276, opened 2026-07-29
Local-stack QA quality of life: arm64 connector containers, child-process cleanup, readiness

jgraettinger:
Quality-of-life work that resolves sharp edges hit while doing QA against a local stack.

**arm64 connector containers.** The `docker --platform` changes are **required** for MacBook arm64 hosts running arm64 VMs under Lima with Rosetta. Without them, connector startup fails with a misleading `no such object` / "did it crash?" error whose real cause is that a platform-less `docker pull` fetches only the host variant under Docker's containerd image store. Also fixes a bogus-readiness race (docker's inline-pull chatter was accepted as the container's ready signal) and container-name collisions from truncated `SystemTime` nanos, which back-to-back reads return identical on ARM's 24MHz timer.

**Child-process cleanup.** `PR_SET_PDEATHSIG` so a hard-killed parent takes its children with it. `Child::drop` only covers paths that unwind; a SIGKILL or OOM kill left etcd and gazette clusters re-parented to init, serving and holding temp dirs indefinitely, and `local:stop` will never reap them since these are from tests and not systemd units.

**Local-stack readiness and paste-ability.** Nothing surfaces when a stack is actually useable: units are `Type=simple` and build in `ExecStartPre`, so both `is-active` and `flow-plane@<dp>.target` report active well before anything binds a port. Adds `local:stack-wait` (TCP probes, fails fast on a failed unit) which `local:stack` now blocks on. Also fixes "ready-to-paste" commands that were not paste-able, a unit listing that dropped the name of any unit with a problem, and documents that connector registration is a publish precondition that a present image does not imply.

**Verification.** QA'd end to end on an aarch64 host with the containerd image store: capture → TypeScript derivation → materialization all publishing and flowing, with each defect's mechanism reproduced independently first. `kill -9` orphan behavior confirmed against a negative control. Touched crates' test suites pass; `cargo fmt --all --check` clean.

_Branch name is a worktree artifact — this is unrelated to #3246._

### DIFF
--- a/crates/runtime/src/container.rs
+++ b/crates/runtime/src/container.rs
@@ -425,7 +425,12 @@ async fn docker_pull(image: &str) -> anyhow::Result<()> {
         let is_transient = err_str.contains("TLS handshake timeout")
             || err_str.contains("connection reset")
             || err_str.contains("i/o timeout")
-            || err_str.contains("unexpected EOF");
+            || err_str.contains("unexpected EOF")
+            // Docker's registry client gives up on a slow registry with
+            // `request canceled (Client.Timeout exceeded while awaiting
+            // headers)`, which reads as a client-side error but is really the
+            // registry being slow to answer.
+            || err_str.contains("Client.Timeout exceeded");
 
         if is_transient && attempt < MAX_RETRIES {
             tracing::warn!(
--- a/crates/runtime/src/container.rs
+++ b/crates/runtime/src/container.rs
@@ -12,6 +12,11 @@ use tokio::io::AsyncBufReadExt;
 // connectors.
 const CONNECTOR_INIT_PORT: u16 = 49092;
 
+// For now, we support only Linux amd64 connectors. `docker pull` must request
+// it explicitly: under Docker's containerd image store a platform-less pull
+// fetches only the host's variant, which `docker run --platform` cannot use.
+const CONNECTOR_PLATFORM: &str = "linux/amd64";
+
 const RUNTIME_PROTO_LABEL: &str = "FLOW_RUNTIME_PROTOCOL";
 const USAGE_RATE_LABEL: &str = "dev.estuary.usage-rate";
 const PORT_PUBLIC_LABEL_PREFIX: &str = "dev.estuary.port-public.";
@@ -35,9 +40,7 @@ pub async fn flow_runtime_protocol(image: &str) -> anyhow::Result<RuntimeProtoco
         docker_pull(image).await.context("pulling image")?;
     }
 
-    let inspect_output = docker_cmd(&["inspect", image])
-        .await
-        .context("inspecting image")?;
+    let inspect_output = inspect_image(image).await.context("inspecting image")?;
 
     let inspection = parse_image_inspection(&inspect_output)?;
     tracing::info!(
@@ -133,8 +136,7 @@ pub async fn start(
         connector_memory_limit(),
         "--cpus".to_string(),
         connector_cpu_limit(),
-        // For now, we support only Linux amd64 connectors.
-        "--platform=linux/amd64".to_string(),
+        format!("--platform={CONNECTOR_PLATFORM}"),
         // Attach labels that let us group connector resource usage under a few dimensions.
         format!("--label=image={}", image),
         format!("--label=task-name={}", task_name),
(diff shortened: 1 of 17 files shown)


jgraettinger:
Four commits added since this opened, in the same spirit as the rest: the `Dekaf Test` job was red on 6 of its last ~40 runs, for reasons unrelated to this branch.

**Group coordinator race.** `connect_to_group_coordinator` built the coordinator's broker URL before checking the sentinel that means "no coordinator yet" — empty host, port `-1`. Since 35bea4767f8 that construction is fallible, so `-1` failed as `invalid broker port` one line above the check meant to handle it, leaving that check unreachable. Every call site serves a live consumer-group API, and Dekaf has no way to return a retryable error mid-session — errors propagate out of `dispatch_request_frame` and close the connection — so an expected transient became a dropped session, which clients report as `BrokerTransportFailure`. A cluster with no `__consumer_offsets` auto-creates it *in response to this very request*, so the first consumer group of a cluster's lifetime reliably hits the window; in CI it landed on whichever test joined a group first, which is why the failing test varied run to run. Now decides the target before touching a URL, honors the broker's `error_code`, and polls with backoff while the coordinator loads.

**Test tolerance.** A shard reports PRIMARY before its connector container is necessarily running, so the proxy synthesizes a 503 before forwarding any request bytes — nothing was ingested, so the POST is safe to replay, and `inject_documents` retries on 503 and only 503. The test consumer likewise polls through the transient errors librdkafka raises while it is already reconnecting, as real consumers do, bounded by a deadline; bad credentials and unknown topics still fail fast. `docker_pull` adds `Client.Timeout exceeded`, which is Docker reporting a slow registry rather than a client fault.

**CI.** `dekaf-e2e` had inherited the default 30s/120s slow-timeout while its slowest tests run near 80s under load, so it gets its own budget. Stack-bringup failures produced no diagnostics at all; the workflow now dumps unit state and journals on failure.

**Verification.** The local Kafka unit's `--tmpfs` data dir makes the cold start reproducible on demand. Cold-restarted, 25/25 pass, with Kafka logging `__consumer_offsets` auto-creation 5ms before Dekaf logs `CoordinatorNotAvailable attempt=1` and 26ms before creation completes — one retry, and `invalid broker port` gone.

crates/async-process/src/lib.rs
```diff
@@ -156,6 +156,61 @@ impl Drop for Child {
...
+            }
+            // Close the race where the parent died between `fork` and the
+            // `prctl` above: the death signal was missed and is never
+            // re-delivered, so the orphan must notice on its own.
+            if libc::getppid() as u32 != parent {
+                libc::_exit(1);
+            }
+            Ok(())
+        });
+    }
```
mdibaiee: This scares me but I don't fully grok the safety mechanism

crates/async-process/src/lib.rs
```diff
@@ -156,6 +156,61 @@ impl Drop for Child {
     }
 }
 
+/// Spawn `cmd` as a [`Child`] which cannot outlive us.
+///
+/// Prefer this over `Command::spawn()` followed by `.into()`: it additionally
+/// applies [`kill_on_parent_death`], so a child is reaped even along the paths
+/// where [`Child::drop`] never runs.
+pub fn spawn(cmd: &mut Command) -> std::io::Result<Child> {
+    kill_on_parent_death(cmd);
```
williamhbaker: For the `spawn_blocking` caveat that is mentioned below... as I understand it, Tokio's multi-thread workers are themselves blocking-pool threads, so any `block_in_place` retires that worker 10s later and SIGKILLs whatever it forked. The agent has one of these [here](https://github.com/estuary/flow/blob/d69e11ebd744093d8113319c7c854ee854c19f03/crates/agent/src/logs.rs#L168), which seems like it would be a problem if that is hit?

mdibaiee (dismissed):

crates/runtime/src/container.rs
```diff
@@ -171,14 +173,14 @@ pub async fn start(
...
         .stdin(async_process::Stdio::null())
         .stdout(async_process::Stdio::null())
-        .stderr(async_process::Stdio::piped())
-        .spawn()
-        .context("failed to docker run the connector")?
-        .into();
+        .stderr(async_process::Stdio::piped());
+
+    let mut process =
+        async_process::spawn(&mut cmd).context("failed to docker run the connector")?;
```
williamhbaker: Does arming this here actually do anything? I think the child is the `docker run` client rather than the container, so SIGKILLing it leaves the container running, and sig-proxy can't forward an uncatchable signal. Since we talk gRPC straight to the container IP, wouldn't the only real effect be losing its stderr?

The same code is in `runtime-next/src/container.rs`. I think only the agent can actually hit it today, since it's the one process with a `block_in_place` (in its log handler) that can retire the thread that did the fork, and the agent imports `runtime` and not yet `runtime-next`.

williamhbaker (commented):
A couple of comments related to the async_process changes

jgraettinger:
I've dropped the async-process change from this branch, since it's controversial and the rest isn't.

The problem it's solving is a local development issue, not a production issue: if an agent is working on e2e tests using etcd / gazette, and manages to hit a live-lock while working on tests, the force-killed test runner leaves behind etcd and gazette processes which must be manually killed. Not a big issue in practice

williamhbaker (dismissed):
LGTM

mdibaiee (approved):

PR #3278, opened 2026-07-30
runtime-v2: fix shuffle recovery deadlock under split read cohorts

### LINKED ISSUE #3246: runtime-v2: shuffle recovery can deadlock when binding priorities split cohorts under a bounded disk limit

jgraettinger:
Fixes #3246 — V2 shuffle recovery could enter a non-converging retry loop when a
task mixes binding priorities (so reads split into multiple cohorts), replays an
idempotent-recovery transaction, and runs under a bounded
`estuary.dev/shuffle-disk-limit`. High-priority cohorts finished their hinted
region first and then tailed without bound into post-crash content that recovery
cannot consume. That over-read pinned log segments, engaged disk back-pressure
globally, and starved the low-priority cohort which actually gated recovery
completion — timing out with zero persisted progress, then retrying against a
larger backlog.

### The fix: recovery becomes its own leader session

A session which opens with unresolved causal hints now does the replay and
nothing else, then quiesces; post-recovery work runs as the next session, over
the same restart path every crash already uses.

- **Slice**: each read arms on its recovered unresolved-hint count and *parks*
  when that count reaches zero — dropping its in-hand batch tail and journal
  stream mid-batch, while retaining enough state that the resolving flush delta
  still reaches the Session. This corrects for disk limits *and* priority inversion:
  draining high-priority reads at the end of their recovery range means we're not
  burning disk on content we won't process, and it ensures lower-priority reads
  aren't starved.
- **Session**: a recovery session only starts reads for `(journal, binding)`
  pairs carrying an unresolved hint, and never promotes held-back progress, so
  it settles after exactly one checkpoint.
- **Leader**: closing an idempotent-replay transaction chains
  `Persist(commit) → Head::Stop`, with the replayed transaction's Acknowledge
  running as the next session's first act.

Parking is lossless by the Frontier invariant (below): the parked read's flush
delta reports a cut through the resolving close, so the next session resumes
there whole.

### Making the restart cheap

Naively, the recovery checkpoint's producer offsets sit at the floor the replay
resumed from, so the next session would re-read the whole hinted span. A
*ratchet* fixes that: while an unresolved checkpoint waits on hints, a delta
whose every commit and causal hint is already accounted for is adopted
wholesale, so the checkpoint carries the replay's true offsets and byte deltas
instead of a conservative bump. The next session re-reads only the novel tail.

### Supporting work

- **The Frontier invariant** is stated canonically on the `Frontier` type, with
  the shuffle README's Concepts section promoted to a real glossary (span, cut
  floor, open/closed, covers, accounts for) plus retired terms and why. Prose
  and renames only.
- **Cyclic hint resolution** drops its redundant reverse pass in favor of the
  stale-hint prune against `completed_clocks`, which is the stronger and more
  durable of the two mechanisms.

### Testing

Unit coverage over the ratchet and prune, updated `resume_from_checkpoint`
fixtures showing the two-session shape, and the scenario fuzz sweep now asserts
sweep-wide properties: recovery cuts never regress, and at least one replay
ratchets strictly forward.

Reviewing commit-by-commit is recommended; each message carries the detailed
rationale.


jgraettinger:
I had a QA agent verify the stacked change by:

* Instrumenting the soak test to insert a delay on the first Store, making idempotent recovery likely most of the time
* Chaos testing by removing materialization shard assignments to force recovery
* Variations: increasing the capture rate to engage back-pressure, and using differentiated binding priorities.

It observed the new park semantics, and that the post-recovery restart resumed at the tail, and couldn't break the change.

dgreer-dev (dismissed):
Makes sense to me, and the tests seem pretty strong.  LGTM.

Could we write an e2e to show the behavior?  Get a leader, `StartCommit`, bring it down before ack, and then start a new session that should just go to the correct frontier?  Not needed, but might be a useful check for the future?

jgraettinger:
Iirc there's a scenario fixture that uses e2e, added in the first change to show idempotent recovery and restart, and then tightened in the second commit snapshots to show a tighter resume. It's also covered by the fuzzer

dgreer-dev (approved):
Love it, LGTM

PR #3155, opened 2026-07-13
Update validation/publication to use snapshots

bbartman:
[Ticket 2781](https://github.com/estuary/flow/issues/2781)
Closes [#2781](https://github.com/estuary/flow/issues/2781)

**Description:**

### Motivation

Publications, discovers, and evolutions authorized each operation by running recursive `internal.user_roles()` queries inside Postgres. The agent already maintains an in-memory authorization `Snapshot` (user grants, role grants, data planes) used for task and journal authorization. This PR moves catalog authorization for those operations onto that same snapshot, replacing the recursive SQL with exact in-memory checks.

### Current design

- All authorization decisions in `publications::specs::resolve_live_specs`, discovers, and evolutions are made by `tables::UserGrant::is_authorized` and `tables::RoleGrant::is_authorized`, evaluated against a `Snapshot`.
- `fetch_live_specs` is now a plain fetch-by-catalog-name; the user filtering that previously happened in SQL is applied in memory afterwards (likewise `live_specs::get_live_specs` / `get_connected_live_specs`).
- Each operation resolves the snapshot watch **once** and threads the same `&Snapshot` through every phase of an attempt: publication initialize → resolve → build, and an entire discover including across the connector RPC.

### Why one Snapshot is pinned per operation

Every check within an attempt sees one consistent point-in-time view, so there are no torn decisions where one phase observes a grant another phase doesn't. It also makes the freshness contract well-defined: staleness is a property of *the* snapshot relative to the operation, not of whichever snapshot each phase happened to grab. Asserted by `test_publication_uses_one_snapshot_across_phases` and `test_discover_uses_one_snapshot_across_connector_rpc`.

### Freshness contract

A snapshot-based **denial** is authoritative only if the snapshot was taken after the operation was queued; otherwise the denial may just be a grant the snapshot hasn't observed yet.

- The reference instant is the queued row's `updated_at`, threaded as `DraftPublication::started_at`. It is durable across retry attempts, which is what lets retries converge (a per-attempt `now()` could never be overtaken by a snapshot).
- `Snapshot::taken_after(anchor)` is the single definition of "authoritative for this instant", shared with `envelope.rs` and task authorization, and allows 250ms of `TEMPORAL_SKEW`.
- Denial + authoritative snapshot ⇒ a real authorization error, reported to the user.
[...]

PR #3279, opened 2026-07-30
2781: Preemptively deploying states for Discovery and Publish

bbartman:
**Description:**

control-plane: carry discover/publication poll state for deploy compatibility

Define DiscoverState and PublicationState and wire them as the executors' automations::Executor::State, ahead of the stale-snapshot deferral changes which will actually read and write them. The automations framework decodes persisted task state with serde_json::from_str, and the prior `()` state only accepts JSON `null` — so this intermediate deploy is needed to keep in-flight tasks decodable across a deploy or rollback in either direction.

Nothing reads the state fields yet, hence #[allow(dead_code)].

We are doing this first in order to prevent a deserialization issue when we deploy the next set of changes [from here](https://github.com/estuary/flow/pull/3155)

**Workflow steps:**

**Notes for reviewers:**

### DIFF
--- a/crates/agent/src/discovers.rs
+++ b/crates/agent/src/discovers.rs
@@ -41,6 +41,22 @@ impl JobStatus {
 
 type ProcessResult = Result<tables::DraftCatalog, Vec<models::draft_error::Error>>;
 
+/// Poll state persisted to `internal.tasks` between polls, and therefore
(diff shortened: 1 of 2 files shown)


jshearer (approved):
LGTM -- we'll need to deploy this before https://github.com/estuary/flow/pull/3155 can land

PR #3280, opened 2026-07-30
Docs: Add Snowflake user type

aeluce:
**Description:**

Feedback from Snowflake on a recent blog included that Estuary's SQL setup script should include `TYPE = SERVICE` when creating the database user. No type would treat the user as `TYPE = PERSON` which would require MFA for the user.

This PR adds the service type to the SQL setup script in the connector reference pages and Snowflake tutorials. (Also removed password auth from the capture connector script and added some indentation on the materialization version.)

Snowflake connector references and tutorials

**Notes for reviewers:**

Thanks for reviewing!


danthelion (approved):
Thanks for catching this!

PR #3285, opened 2026-07-31
docs: rediscovery interval config setting for SQL capture connectors

Alex-Bair:
**Description:**

Documentation updates for https://github.com/estuary/connectors/pull/4962.


aeluce (approved):
Thanks! And good catch on adding it to the Customizing Data Flows page as well.

PR #3292, opened 2026-08-01
graphql: add a tenant reachability filter to serviceAccounts

GregorShear:
Adds `serviceAccounts(filter: { tenant: "acmeCo/" })`, narrowing a listing to the accounts an organization reaches through the role-grant graph.

- `tenant_reachable_prefixes` walks role grants from the caller's footholds within the tenant — their authorized prefixes clamped into its subtree — and keeps the prefixes reached with `QueryServiceAccounts`. Seeding from footholds rather than the tenant root means a branch-scoped caller witnesses only reach flowing from namespace they occupy; edges granted to sibling branches contribute nothing. For a caller whose foothold covers the tenant root, the walk is the tenant's full reach.
- The reachable set is intersected with the caller's own authorized prefixes. `intersect_prefixes` keeps the deeper prefix of each overlapping pair, so the result feeds `catalog_name ^@ ANY($1)` without admitting anything outside either set. The filter only ever removes rows the caller could already see.
- Naming a tenant requires at least one foothold — `QueryServiceAccounts` somewhere within (or covering) the tenant's namespace. Without a gate, filtering by an arbitrary tenant and observing which namespaces return rows would reveal its role-grant edges; the check is a function of the caller's own grants alone, so a denial reveals nothing about the tenant.

The filter lives on `serviceAccounts` only; promoting it into the shared `PrefixFilter` later is mechanical.

Unit tests cover the capability-scoped walk (including ancestor-edge inheritance, sibling-branch exclusion, and multi-seed union) and the intersection. An integration test asserts the filter keeps only the deeper `deepCo/team/` branch of a broader authorization, never widens past the caller's access, admits a branch-scoped caller filtering by the enclosing tenant, denies tenants the caller holds nothing in, and rejects malformed input.

Dashboard consumer: estuary/ui#2042, which depends on this deploying first.

### DIFF
--- a/crates/control-plane-api/src/server/public/graphql/authorized_prefixes.rs
+++ b/crates/control-plane-api/src/server/public/graphql/authorized_prefixes.rs
@@ -41,6 +41,66 @@ fn prune_covered(sorted: impl IntoIterator<Item = String>) -> Vec<String> {
     pruned
 }
 
+/// Returns the prefixes reachable through the role-grant graph from `seeds`
+/// with all `required_capabilities`. Capabilities arriving at the same prefix
+/// union across seeds and paths, matching the additive closure semantics of
+/// `reachable_prefixes`, and each seed is assumed to hold the requested
+/// capabilities on its own prefix. Results are sorted and pruned of covered
+/// children.
+///
+/// For the tenant filter, seeds are the caller's footholds within the named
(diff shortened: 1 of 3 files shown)


GregorShear:
closing in favor of scoped tokens (coming soon-ish)

PR #3294, opened 2026-08-01
materialize: consistency suite verifying exactly-once against a real runtime

mdibaiee:
**Description:**

A test suite that answers one question: when a materialization connector is crashed, stalled,
raced by a stale instance, or re-sharded at the worst possible moment, does the destination
still hold exactly the right data? Addresses estuary/connectors#4956 (steps 1 and 2).

It tests against a real Flow runtime, not a mock: each scenario publishes a real capture and
materialization to a local stack, injects a fault at a precise point in the protocol, and
verifies what the destination actually holds. Four pieces:

- A **shim** that poses as the connector. It sees every protocol message and injects faults —
  crash, stall, or a racing "zombie" instance — keyed on protocol events ("the 4th
  `Acknowledge`"). No changes to Flow and none to the connector under test.
- A **reference connector** over SQLite, implementing the four consistency classes real
  connectors use (modeled on postgres, databricks, and Snowpipe Streaming v2) with seven
  switchable defects.
- **Invariant checkers**: nothing lost, nothing duplicated, every value right — computed from
  an oracle each document carries, rather than asserted from snapshots.
- **15 scenarios**, each pairing a fault with a defect it must catch. Every scenario runs
  clean (must pass) and defective (must fail), so a checker that goes blind through
  refactoring becomes a test failure instead of a silent green.

**Workflow steps:**

```bash
mise run local:stack                       # once
mise run ci:consistency                    # whole suite against the reference connector —
                                           #   SQLite, so no credentials and no cloud spend
mise run ci:consistency --filter zombie    # one scenario
```

Any real connector can be the subject instead, named through `FLOW_CONSISTENCY_SUBJECT_*`
environment variables (binary, endpoint config, consistency class, `testctl`). Verified
against `materialize-databricks`, which passes every applicable scenario — including all
seven that split or join the task's shards.

The documentation is written to be **agent-drivable**: point an agent at
[`AGENT_README.md`](https://github.com/estuary/flow/blob/mdash/consistency-tests-impl-yhxx43xr/crates/materialize-consistency/AGENT_README.md) and it has everything needed to run the
suite against a connector — config decryption, required feature flags, which scenarios apply
to which class, and how to triage each kind of failure.

None user-facing. Internal:

- [`crates/materialize-consistency/README.md`](https://github.com/estuary/flow/blob/mdash/consistency-tests-impl-yhxx43xr/crates/materialize-consistency/README.md) — concise human overview: how it
  works, what each scenario checks, how to test a connector. **Start here.**
- [`crates/materialize-consistency/AGENT_README.md`](https://github.com/estuary/flow/blob/mdash/consistency-tests-impl-yhxx43xr/crates/materialize-consistency/AGENT_README.md) — the complete
  operating guide, for agents and thorough humans.
- `docs/materialize/consistency-testing.md` — the design record: decisions, rejected
  alternatives, and the runtime gap the suite measures.

**Notes for reviewers:**

- **Expected result is 15 of 16 passing.** `split-lands-on-prepared-transaction` is a
  deliberate red: it measures a known runtime gap — a transaction started under one shard
  split is not guaranteed to be replayed under that same split
  ([discussion 2581](https://github.com/estuary/flow/discussions/2581)) — which the
  counted-channel class cannot survive. It fails with a violation count when the gap lands,
  and as an *unexpected pass* otherwise, so it can never silently go green. Against
  `materialize-snowflake`'s Snowpipe Streaming v2 path it trips on every run.
- **Treat "a scenario failed against my connector" as a configuration question first** —
  the AGENT_README's "Reading a failure" section separates connector faults from environment
  and configuration ones.
- The suite is excluded from the default nextest profile (each scenario publishes tasks to a
  live stack) and runs on demand, not in CI. Reading the destination and cleanup go through
  `testctl` (estuary/connectors#4981, merged); `flowctl` and `activate` are untouched — shard
  surgery uses two `gazctl` scripts under `crates/materialize-consistency/scripts/`.
- Suggested reading order: `src/scenarios.rs` (what is verified and why), then
  `docs/materialize/consistency-testing.md` (the design reasoning), then `reference/mod.rs`
  (the class state machines).


crates/materialize-consistency/scripts/join-shards.py
jacobmarble, line 1: Should this instead be a format option for `gazctl shards list` ?

crates/materialize-consistency/scripts/shard-tools.sh
```diff
@@ -0,0 +1,78 @@
...
+[ $# -ge 2 ] || usage
+COMMAND=$1
+TASK=$2
+shift 2
+
+# Resolved rather than assumed: the suite spawns this with a minimal environment, so neither
+# tool is necessarily on PATH. Both are built into this stack's own GOBIN by
+# `mise run local:stack`, which is checked first; the GOPATH and ~/go/bin fallbacks are for a
+# manual invocation outside a stack.
+resolve() {
```
jacobmarble: This comment and function name don't tell me what is being resolved. From context, it seems to resolve a command line tool implemented in golang. Perhaps "Resolves a CLI implemented in Go, without PATH..." and `resolve_tool_path()` or similar?

crates/materialize-consistency/scripts/shard-tools.sh
```diff
@@ -0,0 +1,78 @@
...
+    exit 1
+}
+
+FLOWCTL=$(resolve FLOWCTL flowctl)
+GAZCTL=$(resolve GAZCTL gazctl)
+SELECTOR="estuary.dev/task-name=${TASK}"
+
+# Authorize gazctl against the task's data plane. `--name` resolves the data plane from
+# the catalog name, so the task must still exist; `--admin` is required because both
+# operations mutate.
```
jacobmarble: What are the two operations in "both operations mutate"? What do they mutate? Is this detail helpful to you?

```suggestion
# Authorize gazctl against the task's data plane. `--name` resolves the data plane from
# the catalog name, so the task must exist.
```

crates/materialize-consistency/src/bin/materialize-reference.rs
```diff
@@ -0,0 +1,92 @@
...
+enum Command {
+    /// Read every row of a materialized resource as newline-delimited JSON.
+    Read {
+        /// Path to the endpoint configuration, as JSON or YAML.
+        #[arg(long)]
+        config: String,
+        /// Path to the resource configuration, as JSON or YAML.
+        #[arg(long)]
+        resource: String,
+    },
```
jacobmarble, line 26: These comments are short but confusing. JSON only? JSON or YAML?

crates/materialize-consistency/src/bin/materialize-reference.rs
```diff
@@ -0,0 +1,92 @@
...
+    install_panic_hook();
+
+    match run() {
+        Ok(()) => std::process::ExitCode::SUCCESS,
+        Err(err) => {
+            // As one structured line, because that is what the reactor's log decoder can
+            // attribute: a plain `anyhow` chain is not discarded, but each of its lines is
+            // wrapped as a separate warning, so the cause arrives shredded across several
+            // entries with the level lost. One JSON object keeps the chain and the level
+            // together.
```
jacobmarble: This comment doesn't make sense to me. Maybe if it were split into more sentences?

crates/materialize-consistency/src/bin/materialize-reference.rs
```diff
@@ -0,0 +1,92 @@
...
+        // No subcommand: serve the materialization protocol on stdio.
+        None => reference::serve(connector_init::Codec::Json),
+    }
+}
+
+/// Load a config file as JSON or YAML.
+///
+/// YAML because that is how the connectors repository writes the endpoint configs its
+/// integration tests use, and because it subsumes JSON — so the same reader serves a
+/// harness passing a temporary JSON file and a person passing `config.local.yaml`.
```
jacobmarble: I don't understand the JSON and YAML comment or code. Does this function deserialize YAML contained inside JSON?

crates/materialize-consistency/src/reference/store.rs
```diff
@@ -0,0 +1,880 @@
...
+
+        // Switch to WAL only if the file is not already in it: changing journal mode needs
+        // a brief exclusive lock, and SQLite fails that outright rather than consulting the
+        // busy handler — so an unconditional `PRAGMA journal_mode = WAL` dies with
+        // "database is locked" whenever a sibling shard has the destination open. Journal
+        // mode is a durable property of the file, so the first opener sets it and everyone
+        // after reads `wal` and leaves it alone.
+        //
+        // Both statements go through `query_row` rather than the batch below because
+        // they *return* the mode, which `execute_batch` refuses.
```
jacobmarble: The comment could simply say "Switch journal mode to WAL if needed."

crates/materialize-consistency/src/reference/store.rs
```diff
@@ -0,0 +1,880 @@
...
+        Ok(Self { conn })
+    }
+
+    /// Begin a write transaction, taking the write lock up front.
+    ///
+    /// `BEGIN IMMEDIATE`, not the default `DEFERRED`: every transaction here reads and then
+    /// writes, and in WAL mode a deferred transaction that upgrades its read lock after
+    /// another connection has written returns SQLITE_BUSY_SNAPSHOT immediately, without
+    /// consulting the busy handler. Taking the write lock at `BEGIN` means contention waits
+    /// out the busy timeout instead, which is what a destination shared by two shards needs.
```
jacobmarble: How about "`BEGIN IMMEDIATE` takes write lock immediately so that contention waits out the busy timeout."

crates/materialize-consistency/src/reference/store.rs
```diff
@@ -0,0 +1,880 @@
...
+    ///
+    /// A connector does not need to consult its own pending work here, because the
+    /// protocol lets it wait instead: `LoadIterator::WaitForAcknowledged` blocks until the
+    /// previous transaction has been fully acknowledged, and only then may loads be issued,
+    /// at which point the destination already holds everything committed. This connector
+    /// processes requests in order and applies at `Acknowledge`, so it has that property
+    /// without an explicit wait. Work left pending by a previous session is applied at this
+    /// session's first `Acknowledge` — deliberately not at `Open`, which reclaims and
+    /// inspects nothing — and a load cannot precede that `Acknowledge`, so it still sees
+    /// applied state.
```
jacobmarble: Claude is explaining how the method is used, but I want to know what it does. Maybe the first sentence is enough?

crates/materialize-consistency/src/harness/mod.rs
```diff
@@ -0,0 +1,1460 @@
...
+    /// it already stored" — and every such cause implies a volume. Without a ceiling the
+    /// exemption also absorbs a subject that re-delivered the entire workload, which is a
+    /// different failure wearing the same justification.
+    ///
+    /// `None` where no count is meaningful. That is the case when the violation cannot be
+    /// observed at all — a real connector's destination is read back as a table, which returns
+    /// rows in no particular order, so no number of monotonicity violations says anything — and
+    /// when the checker emits a fixed number of violations however wrong the subject is, such
+    /// as one per account: a ceiling there would measure the workload's account count rather
+    /// than the subject.
```
jacobmarble: The second sentence in this comment has a long hyphenated parenthetical, and a phrase after colon. I can't understand it without a lot of effort. Could it be deleted?

crates/materialize-consistency/src/harness/mod.rs
```diff
@@ -0,0 +1,1460 @@
...
+    /// balance disagreeing with its own oracle" licenses an oracle disagreement *caused by
+    /// duplication*, so if no duplicate row appears anywhere in the run, there is an unexpected
+    /// divergence.
+    pub conditional_on: Option<Invariant>,
+    /// Classes this exemption applies to; `None` to apply to all classes. This allows us to
+    /// scope the exemption to the specific classes that really require it.
+    pub classes: Option<&'static [crate::reference::Class]>,
+}
+
+/// Marker in the error chain of a run that failed for a reason that says nothing about the subject.
```
jacobmarble: This is very difficult to parse:

Marker → in the error chain → of a run → that failed → for a reason → that says nothing → about the subject.
```suggestion
/// Marks a run that failed for reasons unrelated to the subject.
```

crates/materialize-consistency/src/harness/mod.rs
```diff
@@ -0,0 +1,1460 @@
...
+///
+/// The defective half of a scenario treats a failed run as the defect being caught, which is
+/// right for a defect that wedges the task (e.g. `ignore-key-range` leaves two shards fencing
+/// each other and neither can commit) but we should not treat environmental issues as defects: a
+/// gate that timed out during warmup, a split that never landed, a capture slow to deactivate, a
+/// collection read that could not settle.
```
jacobmarble: This is difficult to parse. Could it be omitted? It's just an enum, so the per-value comments be more helpful anyway.

crates/materialize-consistency/scripts/join-shards.py
mdibaiee: this is not a list operation, it modifies the shards topology, and data plane folks are not happy about including it in gazctl because joining of shards may soon be impossible, so this may very well be temporary.

crates/materialize-consistency/src/bin/materialize-reference.rs
mdibaiee: YAML is a superset of JSON. When we parse YAML, we automatically end up supporting JSON too, Claude was being explicit about this. I changed it to say YAML only and JSON is implicit in it.

jacobmarble: How did I not know that?!

crates/materialize-consistency/AGENT_README.md
```diff
@@ -0,0 +1,322 @@
...
+    if isinstance(n, list): return [strip(v) for v in n]
+    return n
+json.dump(strip(yaml.safe_load(sys.stdin)), sys.stdout)' > /tmp/subject-config.json
+```
+
+**A subject driven through a shard split also needs multi-shard operation enabled**, which for
+`materialize-databricks` means adding `scale_out` to `advanced.feature_flags` in that config —
+off by default, and the suite cannot know a given connector's flag names. Without it the connector
+refuses to open a partial-range shard at all and the task crash-loops — a failure that is the
+configuration's doing, not the connector's.
```
danielnelson: This paragraph is out of date now, should be able to remove it.

crates/materialize-consistency/src/invariants.rs
```diff
@@ -0,0 +1,989 @@
...
+/// A conflict — a path descending through a value that is already a scalar — leaves the
+/// scalar alone rather than discarding it. That only happens when a destination has both a
+/// column `x` and a column `x/y`, which no projection produces, and silently dropping
+/// data would be the worse failure of the two.
+fn insert_pointer(
+    object: &mut serde_json::Map<String, serde_json::Value>,
+    path: &str,
+    value: serde_json::Value,
+) {
+    let mut segments = path.split('/');
```
danielnelson: This was the bit that was tripping me up in BigQuery.  It replaces `/` with `_` in an identifier and this prevented the tests from from running with:
```
row does not describe an event: missing field `oracle`
```

My workaround was to also split on `_`, but I worry this might cause issues with fields containing `_` as part of the field name?
```
let mut segments = path.split(['/', '_']);
```

crates/materialize-consistency/src/invariants.rs
```diff
@@ -0,0 +1,989 @@
+//! What the destination must hold, and how we know.
+//!
+//! Every check here is an externally observable property of the destination,
+//! expressed in the workload's own vocabulary — conservation, oracle agreement,
+//! cardinality, monotonicity. None of them inspects connector internals: not
+//! staged file names, not checkpoint contents, not the shape of a connector state
+//! document. Those legitimately differ across the four connector classes, so a
+//! test that asserted on them would pass or fail for reasons unrelated to
+//! consistency and would obstruct the very refactors it should protect.
```
jacobmarble:
```suggestion
//! cardinality, monotonicity. None of them inspects connector internals; internals
//! differ across the four connector classes.
```

crates/materialize-consistency/src/invariants.rs
```diff
@@ -0,0 +1,989 @@
...
+//! read with `SELECT *` is no order at all — only the reference connector's own tables
+//! replay the order rows were appended in.
+//!
+//! So any check whose meaning depends on order must establish that order itself, from
+//! `seq`. Getting this wrong reports violations against a correct connector, which is the
+//! worst thing this suite can do.
+//!
+//! The monotonicity checks are the deliberate exceptions — `check_merged_delta`'s and
+//! `check_log`'s — because they are *about* arrival order rather than merely computed in it.
+//! Both are exempted for subjects whose destination cannot preserve it.
```
jacobmarble:
```suggestion
//! read with `SELECT *` is no order at all, so any check whose meaning depends on order
//! must establish that order itself, from `seq`.
//!
//! The monotonicity checks (`check_merged_delta` and `check_log`) are exceptions to the
//! ordering rule because they are *about* arrival order. Both are exempted for a connector
//! under test whose delivery order the harness cannot recover from the destination.
```

crates/materialize-consistency/src/invariants.rs
```diff
@@ -0,0 +1,989 @@
...
+    /// destination yields *columns*, and Flow names each column by the JSON pointer of
+    /// the field it projects, so the nested `oracle` object arrives flattened as
+    /// `oracle/seq` and `oracle/balance`. A materialized table is columns, and a standard
+    /// binding need not carry a root document at all (see the connectors'
+    /// `no_flow_document` option), so there is nothing to parse whole.
+    ///
+    /// Pointer-named columns are therefore folded back into the object they came from.
+    /// Unrelated columns — `flow_published_at`, the workload's `set/` and `transfer/`
+    /// fields — fold harmlessly and are then ignored, because nothing here is required to
+    /// understand every projection a connector chose to materialize.
```
jacobmarble:
```suggestion
    /// Two shapes arrive here, and both are legitimate:
    ///
    /// 1. **A whole document**, from a connector that stores documents as they are — the
    ///    reference one. The nested `oracle` is a nested object:
    ///    `{"id":7,"seq":3,"balanceDelta":-25,"oracle":{"seq":3,"balance":100}}`
    /// 2. **Columns**, from a SQL destination. Flow names each column by the JSON pointer of
    ///    the field it projects, so the same `oracle` arrives flattened into two columns:
    ///    `{"id":7,"seq":3,"balanceDelta":-25,"oracle/seq":3,"oracle/balance":100}`
    ///
    /// Shape 2 is not a fallback for shape 1: a materialized table *is* columns, and a
    /// standard binding need not carry a root document at all (see the connectors'
    /// `no_flow_document` option), so often there is nothing to parse whole.
    ///
    /// Both shapes are therefore reduced to shape 1 — pointer-named columns are folded back
    /// into the object they came from — and parsed once.
    ///
    /// A row carries more columns than the four fields [`Event`] declares. The connector adds its
    /// own — `flow_published_at`, `flow_document` — and the workload writes fields no
    /// invariant reads, such as `set/add` and `transfer/amount`. All of them fold harmlessly
    /// and are then ignored, because nothing here is required to understand every projection
    /// a connector chose to materialize.
```

crates/materialize-consistency/src/invariants.rs
```diff
@@ -0,0 +1,989 @@
...
+            .context("row does not describe an event")
+    }
+}
+
+/// Place `value` at the `/`-delimited `path` within `object`, creating objects as needed.
+///
+/// A conflict — a path descending through a value that is already a scalar — leaves the
+/// scalar alone rather than discarding it. That only happens when a destination has both a
+/// column `x` and a column `x/y`, which no projection produces, and silently dropping
+/// data would be the worse failure of the two.
```
jacobmarble:
```suggestion
/// A conflict is a path descending through a value that is already a scalar. It happens only
/// when a destination has both a column `x` and a column `x/y`, which no projection produces.
///
/// One of the two columns must lose, and this keeps the scalar. Descending would have to
/// replace `x` with an object, and a scalar `oracle` is the JSON-text form that
/// [`Event::from_row`] re-parses into the whole object — so it carries at least as much as the
/// `oracle/seq` being dropped, and possibly more.
```

crates/materialize-consistency/src/invariants.rs
```diff
@@ -0,0 +1,989 @@
...
+    /// Per-sequence document, so a delivered row can be compared against the one
+    /// the collection actually holds.
+    pub by_seq: BTreeMap<i64, Event>,
+}
+
+#[derive(Clone, Debug, Default)]
+pub struct Expectation {
+    pub accounts: BTreeMap<i64, Account>,
+    /// Documents whose `(id, seq)` the read surfaced more than once. `accounts` folds
+    /// duplicates into one, so this count is how a caller knows duplicates were present.
```
jacobmarble:
```suggestion
    /// How many times the read surfaced an `(id, seq)` it had already seen. A document
    /// surfaced three times counts twice, because both repeats are surplus.
    ///
    /// [`Account::seqs`] is a set, so a repeat fails its insert and changes nothing else —
    /// the balance is not added twice, and `by_seq` overwrites at the same key. This counter
    /// is therefore the only place a caller can learn that the read held duplicates.
```

crates/materialize-consistency/src/invariants.rs
```diff
@@ -0,0 +1,989 @@
...
+        check_standard(&b.merged_expected, standard, &mut violations);
+        check_standard_delta_agreement(standard, &b.merged_delta, &mut violations);
+    }
+
+    violations
+}
+
+/// The expectation itself must conserve. This is the baseline guard: a wiring
+/// problem that made the harness read an empty or torn collection shows up here
+/// as a failure rather than as a vacuous pass everywhere else.
```
jacobmarble:
```suggestion
/// The expectation itself must hold: it must name at least one account, and its accounts'
/// balances must sum to exactly zero. Every transfer is a matched pair of legs, so the
/// collection cannot sum to anything else.
///
/// This is the baseline guard. A wiring problem that made the harness read an empty or torn
/// collection shows up here as a failure, rather than as a vacuous pass everywhere else.
```

crates/materialize-consistency/src/invariants.rs
```diff
@@ -0,0 +1,989 @@
...
+/// Shared by both delta checkers, because they must score a regression identically: a run's
+/// suppression counts are read as evidence about which exemptions carry weight, and two scoring
+/// rules make that evidence incomparable.
+///
+/// `<` rather than `<=`, because a *repeated* seq is not a regression of order — it is a
+/// duplicate, and `NoDuplicates` owns that.
+///
+/// And the baseline advances to what was just delivered rather than holding a high-water mark,
+/// so a replay is one violation rather than one per row it replays: after 1..10 then 8, 9, 10,
+/// the 8 is the regression and the 9 and 10 following it are in order.
```
jacobmarble:
```suggestion
/// Walk each account's rows in the order the destination delivered them. Its `seq` must never
/// decrease along that walk.
///
/// Shared by both delta checkers, because they must count a regression identically. A run
/// counts the violations each exemption suppressed, and those counts are what the
/// `max_suppressed` ceilings are set from. `scenarios::REORDERING_CEILING` is 500 because
/// membership-change runs were each observed to suppress between 9 and 93 of these
/// violations. Both checkers feed that one ceiling, so if they counted a replay differently
/// the ceiling would be set from two different units.
```

crates/materialize-consistency/src/invariants.rs
```diff
@@ -0,0 +1,989 @@
...
+fn check_monotonic(rows: &[Event], binding: &str, out: &mut Vec<Violation>) {
+    let mut last_seq: BTreeMap<i64, i64> = BTreeMap::new();
+
+    for row in rows {
+        if let Some(previous) = last_seq.get(&row.id) {
+            if row.seq < *previous {
```
jacobmarble, line 455:
```suggestion
        if let Some(previous) = last_seq.get(&row.id) {
            // `<` rather than `<=`, because a *repeated* seq is not a regression
            // of order, it is a duplicate, and `NoDuplicates` owns that.
            if row.seq < *previous {
```

crates/materialize-consistency/src/invariants.rs
```diff
@@ -0,0 +1,989 @@
...
+                out.push(Violation {
+                    invariant: Invariant::Monotonicity,
+                    detail: format!(
+                        "account {}: {binding} delivered seq {} after {previous}",
+                        row.id, row.seq
+                    ),
+                });
+            }
+        }
+        last_seq.insert(row.id, row.seq);
```
jacobmarble, line 468:
```suggestion
        // The baseline advances to what was just delivered rather than holding a high-water
        // mark, so a replay is one violation rather than one per row it replays: after
        // 1..10 then 8, 9, 10, the 8 is the regression and the 9 and 10 after it are in order.
        last_seq.insert(row.id, row.seq);
```

crates/materialize-consistency/src/invariants.rs
```diff
@@ -0,0 +1,989 @@
...
+    }
+}
+
+/// The append-only collection materialized as deltas: the sharpest detector in
+/// the suite, because every document has a distinct key and so a duplicate
+/// delivery is an extra row rather than an invisible re-reduction.
```
jacobmarble:
```suggestion
/// Check the append-only collection, materialized with delta-updates.
///
/// That collection is keyed `[/id, /seq]`, so every document holds a distinct key and gets its
/// own row. A duplicate delivery is therefore a second row, which is counted directly. The
/// merged collection cannot do this. A duplicate there is summed into the reduced balance,
/// where only arithmetic reveals it.
///
/// Row counts settle loss, duplication and fabrication. The row's own fields settle integrity,
/// and arrival order settles monotonicity.
```

jacobmarble (commented):
I'm trying different ways to review lots of comments, so this pass is relatively short.

crates/materialize-consistency/src/invariants.rs
mdibaiee: The fields are known to us as we provide the collection to connectors. I think it makes sense to split based on however our connectors end up constructing nested field names. So far that is / and _

crates/materialize-consistency/AGENT_README.md
```diff
@@ -0,0 +1,315 @@
+# materialize-consistency: operating guide
```
danielnelson, line 1: I think you should try to merge this information into README.md.  This file contains a lot of that information again, but most of what it says would be useful to a human too.

danielnelson (dismissed):

crates/materialize-consistency/AGENT_README.md
mdibaiee: I have this in my TODO list and will address it. For now I want to merge this and will revisit the README after 👍🏽

jacobmarble (approved):

danielnelson (approved):

PR #3297, opened 2026-08-02
dekaf: opt-in `allow_empty_topics` for never-written collections

### LINKED ISSUE #3064: materialize-dekaf: empty (not-yet-written) collections return LeaderNotAvailable and block consumer-group rebalance for the whole subscription
## Summary

When a Dekaf materialization has one or more bindings whose collection has no journals yet (not-yet-written / empty), Dekaf returns `LeaderNotAvailable` for those topics. If a Kafka consumer group subscribes to a mix of populated and empty topics, the group rebalance never completes, so the consumer is never assigned any partitions and consumes nothing — including the populated topics in the same subscription.

## Impact

A single empty binding silently blocks the entire materialization's consumer. The task logs this repeatedly:

```
Collection binding exists but has no journals available
{collection_name: "<prefix>/public/<collection>", module: "dekaf::topology", task_name: "<prefix>/materialize-dekaf/dekaf-generic"}
```

Customer-visible symptom: the materialization shows 0 docs / 0 bytes and the consumer never makes progress, even though most bound collections have data. It also produces very high log volume — one warning per affected binding per metadata request, and a consumer hot-loops metadata trying to resolve a leader (tens of millions of log lines/day observed on one task).

## Reproduction

Confirmed with kcat against a subscription that mixes a populated topic with an empty one:

```
# populated topic alone -> works
kcat ... -X auto.offset.reset=earliest -G grp populated_topic
#   assigned: populated_topic [0]
#   Reached end of topic populated_topic [0] at offset N: exiting

# populated + empty topic -> never rebalances
kcat ... -X auto.offset.reset=earliest -G grp populated_topic empty_topic
#   % Waiting for group rebalance
#   (hangs, never finishes rebalance)
```

## Root cause

Introduced in estuary/flow#2566 (commit 58f0f0da54), which added the `CollectionStatus` enum (`Ready` / `NotFound` / `NotReady`) and maps `NotReady` (binding exists but journals not available) to `LeaderNotAvailable`, a retryable error, so consumers retry until journals appear.

Relevant code:
- `crates/dekaf/src/topology.rs` — `partitions.is_empty()` returns `CollectionStatus::not_ready()` and logs "Collection binding exists but has no journals available".
- `crates/dekaf/src/session.rs` — `metadata_all_topics` / `metadata_select_topics` map `NotReady` to `ResponseError::LeaderNotAvailable`.

A retryable error is reasonable for a transient collection reset, but for a collection that simply has no data yet it persists indefinitely, and a `LeaderNotAvailable` topic in a subscription prevents the whole consumer group from stabilizing.

## Proposed change

Present a not-yet-written collection as a valid empty topic: a single sentinel partition at offset 0, instead of `LeaderNotAvailable`. The group then stabilizes and empty topics sit idle at offset 0 until data arrives.

Design notes from triage:
- There's a recollection that empty collections may have presented as an empty partition at offset 0 in earlier behavior, with the current `LeaderNotAvailable` behavior arriving when collection-reset handling was made to flow through end-to-end. Confirm so the flag doesn't regress reset semantics.
- Gate behind a default-off advanced option / feature flag. Dekaf behavior changes are historically high-risk, so this should be opt-in, at least at first.
- Ownership: Dekaf currently has no formal owner and will likely land with the materializations team. Joseph Shearer is the subject-matter expert and should be looped in on design/review.

## Workaround

Write at least one document to each empty collection (for example, insert and delete a row in the source table) so journals are created; the topic then resolves normally. This is cumbersome when many tables are empty during a rolling migration.

## Priority / timing

Med. A hacky workaround exists. Targeting after the upcoming offsite.

mdibaiee:
Fixes #3064.

## Problem

A materialization binding whose collection has no journals is reported as `CollectionStatus::NotReady`, which maps to the retryable `LeaderNotAvailable`. That's right for a collection reset that's about to complete, but a collection which has simply never been written stays that way indefinitely.

The issue frames this as a metadata problem, but the sharper blocker is group management: `join_group` (`session.rs:1360-1376`) and `sync_group` (`session.rs:1573-1586`) fail the **entire request** with the first non-`Ready` topic's error code. So one empty binding makes every JoinGroup return `LeaderNotAvailable` and the group never forms — the consumer is assigned nothing and reads nothing, including from the populated topics in the same subscription. `offset_commit` is all-or-nothing in the same way; metadata / ListOffsets / Fetch / OffsetFetch are per-topic.

Reproduced against a local stack with a subscription mixing one populated and one never-written collection:

| subscription | assigned partitions | records read |
| --- | --- | --- |
| `topic_a` (populated) alone | 1 | 2 |
| `topic_a` + `topic_b` (empty) | 0 | 0 |

## Change

New endpoint option `allow_empty_topics`, **off by default**. When set, a collection with no journals is served as a valid topic with one sentinel partition sitting at offset 0:

- `topology.rs` — `Collection::empty_topic`, `partition_count()`, and `fetch_partition_offset` answering offset 0 for the sentinel (which also covers ListOffsets and OffsetForLeaderEpoch).
- `session.rs` — sentinel partition in metadata; an empty successful batch in Fetch; `partition_count()` when creating the upstream offsets topic; skip the journal-name lookup when logging commits.
- Group management needs no change: the collection is `Ready`, so JoinGroup/SyncGroup pass through and the group stabilizes.
- Side effect worth noting: the schema registry (`registry.rs:102`) also stops rejecting empty collections, since Avro schemas come from the spec rather than from journals.

Once the collection is first written, its real journals take over at the same leader epoch, so partition 0's committed offset stays valid.

This "single partition at offset 0" shape is what dekaf presented before 58f0f0da54 introduced `CollectionStatus` — the flag restores it without touching reset semantics on the default path. Dekaf behavior changes are historically high-risk, hence opt-in.

## Testing

`mise run ci:dekaf-e2e` — 27 passed (3 flaky-retried, all the pre-existing `SHARD_STOPPED` injection race, unrelated to this change). Two new e2e tests:

- `test_empty_binding_blocks_consumer_group` pins the default behavior (`LeaderNotAvailable`, nothing assigned, nothing consumed).
- `test_allow_empty_topics_unblocks_consumer_group` subscribes to one populated and **two** empty topics: it asserts metadata / ListOffsets / Fetch on each empty topic, then that all three partitions are assigned and both documents are consumed from the populated topic.

## Note for review

Separate pre-existing bug found while testing, left alone as out of scope: `KafkaApiClient::connect_to_group_coordinator` (`api_client.rs:494-500`) builds the coordinator URL *before* checking the `port == -1` sentinel, so a `FindCoordinator` response with port -1 (coordinator not yet elected) fails with "invalid broker port" and drops the client connection — the sentinel branch below it is unreachable. It's a one-line reorder if we want it.

cc @jshearer as SME on dekaf.

### DIFF
--- a/crates/dekaf-connector/src/lib.rs
+++ b/crates/dekaf-connector/src/lib.rs
@@ -46,6 +46,15 @@ pub struct DekafConfig {
     #[serde(default)]
     #[schemars(title = "Strict Topic Names")]
     pub strict_topic_names: bool,
+    /// Whether to expose bound collections which have never been written as
+    /// valid, empty topics holding a single partition at offset 0. Off by
+    /// default, in which case such a topic reports the retryable error
+    /// `LeaderNotAvailable` until its first document is written — which
+    /// prevents a consumer group from stabilizing across the whole
+    /// subscription, not just the empty topic.
+    #[serde(default)]
+    #[schemars(title = "Allow Empty Topics")]
+    pub allow_empty_topics: bool,
 }
 
 /// Configures a particular binding in a Dekaf-type materialization
--- a/crates/dekaf/src/lib.rs
+++ b/crates/dekaf/src/lib.rs
@@ -122,6 +122,13 @@ impl SessionAuthentication {
             SessionAuthentication::Redirect { config, .. } => config.deletions,
         }
     }
+
+    pub fn allow_empty_topics(&self) -> bool {
+        match self {
+            SessionAuthentication::Task(task_auth) => task_auth.config.allow_empty_topics,
+            SessionAuthentication::Redirect { config, .. } => config.allow_empty_topics,
+        }
+    }
 }
 
 impl TaskAuth {
--- a/crates/dekaf/src/session.rs
+++ b/crates/dekaf/src/session.rs
@@ -2,7 +2,8 @@ use super::{App, Collection, CollectionStatus, CollectionUnavailable, Read};
 use crate::{
     DekafError, KafkaApiClient, KafkaClientAuth, SessionAuthentication, TaskState,
     from_downstream_topic_name, from_upstream_topic_name, logging::propagate_task_forwarder,
-    read::BatchResult, to_downstream_topic_name, to_upstream_topic_name, topology::PartitionOffset,
+    read::BatchResult, to_downstream_topic_name, to_upstream_topic_name, topology,
+    topology::PartitionOffset,
 };
 use anyhow::{Context, bail};
 use bytes::{Buf, BufMut, Bytes, BytesMut};
@@ -896,6 +897,20 @@ impl Session {
                     }
                 }
 
+                // An empty topic has no journal to read; its sentinel partition is
+                // answered with an empty batch during the poll phase below.
+                if collection.empty_topic {
+                    metrics::counter!(
+                        "dekaf_fetch_requests",
+                        "topic_name" => key.0.to_string(),
+                        "partition_index" => key.1.to_string(),
+                        "task_name" => task_name.to_string(),
+                        "state" => "empty_topic"
+                    )
+                    .increment(1);
+                    continue;
+                }
+
                 let Some(partition) = collection
                     .partitions
                     .get(partition_request.partition as usize)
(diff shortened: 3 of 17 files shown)


jwhartley:
doc files no longer do.

at build time.

Suggested split:

- Open the doc half against `estuary/connectors`, where these paths map across:

  - `site/docs/reference/Connectors/materialization-connectors/Dekaf/` becomes `docs/reference/Connectors/materialization-connectors/Dekaf/`

All nine Dekaf pages move together under that one prefix, so it is a
single directory-level move.

PR #3298, opened 2026-08-03
docs: AWS IAM auth cross-account guidance, correct Identity Provider ARN wording

jwhartley:
Two fixes to the AWS IAM auth guide, both surfaced by a customer question about connecting to a resource in a second AWS account.

**1. The Identity Provider ARN wording was wrong.**

The trust-relationship section said the principal is "the ARN of the Identity Provider depending on the data plane (if you use a private deployment or BYOC, we will provide you with this value)". The IAM OIDC provider is a resource the user creates in their own AWS account, so its ARN is theirs, not ours. What we supply is the issuer, and the preceding section already explains how to find it in the dashboard. Reworded to point at the provider they just created and to copy the ARN from the console.

**2. No guidance on resources in another AWS account.**

`crates/iam-auth/src/providers/aws.rs` makes a single `sts:AssumeRoleWithWebIdentity` call with no chained `AssumeRole`. Combined with an IAM OIDC provider being account-scoped, that means both the provider and the role have to live in the account owning the resource. AWS documents the same fork for EKS IRSA ([cross-account access](https://docs.aws.amazon.com/eks/latest/userguide/cross-account-access.html)): create an IdP in the target account, or chain two `AssumeRole` calls. Only the first option is available to us, and the guide didn't say so.

The new section also notes why a resource-based policy on the target resource is not a substitute: connectors call account-level APIs during discovery, `dynamodb:ListTables` being the concrete case, which a resource policy on an individual resource cannot grant.

Also promotes a stray `#` mid-page to `##`. The anchor is derived from the heading text, so `#trust-relationship-in-role` is unchanged.

No code changes.

### DIFF
--- a/site/docs/guides/iam-auth/aws.md
+++ b/site/docs/guides/iam-auth/aws.md
@@ -43,12 +43,12 @@ For example, these are the issuer values for a few common public data planes:
 
 ![Add Identity Provider](../guide-images/aws-iam-1.png)
 
-# Trust Relationship in Role
+## Trust Relationship in Role
 
 Finally, return to the details page of your role, head to "Trust relationships" tab and add the following trust policy, replacing:
 
 * The OpenID URL with the correct issuer value for your data plane
-* The principal with the ARN of the Identity Provider depending on the data plane (if you use a private deployment or BYOC, we will provide you with this value) you use
+* The principal with the ARN of the Identity Provider you created in the previous step. This provider lives in your own AWS account, so copy its ARN from the IAM console rather than writing it by hand
 * The `:sub` condition with your tenant name so only tasks from your tenant are allowed to assume this role
 
 :::tip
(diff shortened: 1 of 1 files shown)


site/docs/guides/iam-auth/aws.md
```diff
@@ -77,3 +77,13 @@ Make sure the `:aud` and `:sub` values use the correct issuer. Minor discrepanci
     ]
 }
 ```
+
+## Resources in a Different AWS Account
+
+Estuary makes a single `sts:AssumeRoleWithWebIdentity` call. There is no second, chained `AssumeRole`, so a role in one account cannot be used as a stepping stone to a role in another.
```
danielnelson, line 83: Interesting, it seems it would be possible for the connector to make a chained `AssumeRole` based on the `AssumeRoleByWebIdentity` provided by the runtime.  Would this make cross account easier to work with?

danielnelson (approved):

site/docs/guides/iam-auth/aws.md
jwhartley: It would but I don't think it's come up before this customer. It was mentioned during design ([thread](https://estuaryworkspace.slack.com/archives/C03Q2NRFKDL/p1750441646671509?thread_ts=1750430746.521649&cid=C03Q2NRFKDL)) and I see [aws docs](https://docs.aws.amazon.com/eks/latest/userguide/cross-account-access.html) mention it.
I've submitted [connectors#5003](https://github.com/estuary/connectors/issues/5003) as low prio, we can update this doc if it gets implemented

PR #3302, opened 2026-08-03
control-plane-api: remove dead evolutions module

jshearer:
## Summary

Removes `crates/control-plane-api/src/evolutions/` and `crates/models/src/evolutions.rs`. Neither has had a caller since August 2025. This is deletion only, with no behavior change.

## What an evolution was

An evolution is a control-plane operation for applying a breaking collection change across the rest of a Data Flow. Collections are append-only with an enforced schema, so some edits (a key change, an incompatible field type) cannot apply in place. An evolution resolved that by one of two paths:

* Re-create the collection under a new name (`acmeCo/foo` becomes `acmeCo/foo_v2`), then rewrite every capture `target` and materialization `source` binding referencing the old name and increment their `backfill` counters.
* Reset it in place, keeping the name while discarding data and inferred schema, when `reset: true`.

It ran as an async job modeled on `discovers`: insert a row into the `evolutions` table with a `draft_id` and `collections`, the agent dequeued it, merged results into the draft, and optionally auto-published. Requests could scope the update to a specific list of materializations.

## Timeline

| Commit | Date | Change |
|---|---|---|
| `68a5359c430` | 2023-05-02 | Introduced. `evolutions` table (`18_evolutions.sql`), `agent-sql/src/evolutions.rs`, `agent/src/evolution.rs` |
| `3bc6ac97ddb` | 2023 | Evolutions increment the backfill counter |
| `e970b36d6bf` | 2023 | Can target a specific set of materializations |
| `30e668aca02` | 2024-06-12 | Controllers introduced. Materialization controllers begin handling incompatible schema changes via `onIncompatibleSchemaChange` |
| `24a67860a7c` | 2024-11-06 | Refactored to take and return `tables::DraftCatalog` rather than DB draft rows, so controllers could call it in memory |
| `5ac55598152` | 2025-01-29 | Handler becomes an automations executor. The commit notes there was no existing integration test coverage for interactive evolutions |
| `c089b54dc26` | 2025-08-19 | Removed |
| `42d0b8b80c1` | 2025-08-21 | Reintroduced as dead code |

## What happened

Two commits, two days apart.

`c089b54dc26` removed the feature deliberately. `onIncompatibleSchemaChange` handling had moved into the `validation` crate, so the premise no longer held:

> Also remove the "evolutions" handler, as it's no longer possible for publications to fail with "incompatible" collections that are resolved through a user-initiated control-plane automation.

That commit deleted `agent/src/evolution.rs` (411 lines), unregistered `agent::EvolutionExecutor` in `main.rs`, dropped the integration tests and snapshots, and stripped `incompatible_collections` and `requiresRecreation` from the publication `JobStatus`. It left `agent/src/evolution/executor.rs` orphaned on disk, 149 lines with no `mod` declaration referencing it.

`42d0b8b80c1` then reintroduced the logic. That commit was the `agent-sql` to `control-plane-api` migration, and history is linear with `c089b54dc26` as an ancestor, so it was rebased onto the removal. It deleted the orphaned `executor.rs` and added `control-plane-api/src/evolutions/mod.rs`, which is the pre-removal `agent/src/evolution.rs` with 52 lines of difference: the two `executor` wiring lines dropped, `next_name()` inlined from the agent's `lib.rs`, and `mod db;` re-exports added. The `agent-sql/src/evolutions.rs` half came across as `evolutions/db.rs`.

The module then survived every subsequent cleanup pass because it is declared `pub mod evolutions;` in a library crate, where rustc's `dead_code` lint does not fire on public items. It was carried through the Rust 2024 edition migration (`db602a6cf7d`), a repo-wide `cargo fmt` (`cb4e4426200`), and the `Envelope` authorization refactor (`e63cca1743b`).

## How this is handled today

Incompatible schema changes are resolved in the `validation` crate together with materialization controllers, driven by the `onIncompatibleSchemaChange` field on materialization specs, which offers `abort`, `backfill`, `disableBinding`, and `disableTask`. Collection resets are expressed as `CollectionDef::reset` on the model and propagate to dependents through `notify_dependents`. The concept is intact; what is gone is the discrete user-initiated job that used to sit in front of it.

## Loose ends not addressed here

* The `evolutions` table, its `create_evolution_task` trigger, and `internal.create_handler_task(id, 5)` all still exist. `automations::task_types::EVOLUTIONS` is still `TaskType(5)` and no executor registers it. The automations dequeue filters on `task_type = ANY($1)` built from registered executors, so a row inserted into `evolutions` enqueues an `internal.tasks` row that is never dequeued, leaving `job_status` at `{"type":"queued"}` indefinitely.
* Whether that path is reachable depends on the UI, which lives in a separate repo. As of an April 2026 reading, `src/api/evolutions.ts` exported `createEvolution()`, inserting into the table from `SchemaEvolution.tsx` and `IncompatibleCollections/CollectionsList.tsx`. This wants verifying before the table is retired, since dropping it while the insert path is live converts a hanging job into a hard error.
* `site/docs/concepts/advanced/evolutions.md:58` still instructs users to click Apply to trigger an evolution. The `onIncompatibleSchemaChange` sections of that page remain accurate. Revising the docs depends on the same UI question.
* `crates/agent/src/integration_tests/harness.rs:479` still carries a `delete from evolutions` cleanup CTE, left alone while the table exists.

A reasonable order for finishing: confirm and cut the UI insert path, drop the table and its triggers in a migration, retire `TaskType(5)` with a comment against reuse, then revise the docs.


bbartman (approved):
This looks good. It may cause some headaches for the large PR just FYI, because the PR that's going to follow that is going to have changes there but that shouldn't be a problem to fix.

GregorShear (approved):

PR #3303, opened 2026-08-03
Introduce an indirect form of built specifications

jgraettinger:
Establishes a second, "indirect" encoding of built specifications and the
Validate requests derived from them, and makes every component in the tree
able to read it. Behavior and performance are unchanged: the new form is
opt-in per task, and no task opts in yet.

## Why

A task with tens of thousands of bindings over a much smaller set of distinct
collections (a CDC capture mapping many schemas of one logical table into one
collection, say) inlines a full `CollectionSpec` into every binding. That cost
is paid in the built spec, in each connector request, at every hop inside the
data plane, and in persistence — and it grows with the *product* of binding
count and collection size.

In indirect form a spec-carrying message holds one shared
`linked_collections` table, and its bindings name entries by
`collection_index` instead of inlining `collection`. Cost then grows only
with the binding count.

## Shape of the change

* **Encoding is a per-message property.** Non-empty `linked_collections` means
  every binding of that message — active and inactive alike — uses
  `collection_index`; empty means every binding inlines its collection, exactly
  as today. Mixed forms are invalid, and Go's `Validate()` enforces this
  natively so the existing Unmarshal-then-Validate path covers both.
* **No inflate, no deflate on read.** A spec keeps whichever form it arrived
  in. Readers instead go through accessors (`proto_flow::linked` in Rust,
  `BindingCollection`/`TransformCollection` in Go) which resolve a binding to
  its collection over both forms, and also surface the collection's *identity*
  — the hook a later runtime-next change will key per-collection memoization
  on. The bulk of the diff is mechanical conversion of read sites to these
  accessors.
* **`crates/validation` is the only writer.** Builders now construct every
  built spec and Validate request in indirect form unconditionally; the
  `indirect-specs` shard flag gates one contained post-process that *inlines*
  the finished message when the flag is unset. So the default path round-trips
  intern → inline over every fixture, and unchanged snapshots are the proof of
  round-trip fidelity. When indirect form is universal, removal is a deletion:
  drop the argument and its branch, and what remains is the end-state builder.
* **The flag is one end-to-end assertion.** `shards.flags: {indirect-specs:
  "true"}` says the built spec, the Validate/Apply requests, every internal
  hop, *and the connector* all handle the indirect form. A connector that
  hasn't been updated simply doesn't work with a task that genuinely needs
  this — such tasks don't function today either. Only `crates/validation`
  reads the flag; nothing else branches on it.
* **The binding ceiling moves under the flag**, from 10,000 to just under
  `doc::combine`'s u16 limit, since indirect form is what makes that many
  bindings affordable. The runtime separately guards the format limit, so an
  over-large spec fails at publish rather than as silent corruption in a
  crash-looping shard. See that commit's message for why the two limits
  deliberately share no constant.

## Post-conditions

Indirected specs can be validated and built behind the flag, and every
component in this repo — data plane, both runtimes, shuffle, publisher, dekaf,
flowctl, and the in-repo `derive-sqlite` / `derive-typescript` /
`derive-python` connectors — reads them. Unflagged tasks are byte-identical to
today, so there's no `last_applied` churn and no rolling-deploy version skew.

Also included: a fix for a dead `LambdaConfigJson` check in Go's
`Transform.Validate()` that had never actually run.

## Follow-ons (not here)

* runtime-next memoization of binding-derived state (validators, extractors,
  projections) per distinct collection — the payoff the accessors' identity
  return exists for.
* Out-of-tree connector updates, and any eventual fleet-wide default.

Reviewing commit-by-commit is recommended; each is self-contained and inert
until the last two.


crates/validation/src/derivation.rs
```diff
@@ -251,11 +254,16 @@ async fn walk_derivation<C: Connectors>(
         shards,
     } = model;
 
-    if transforms_model.len() > crate::MAX_BINDINGS {
+    let indirect_specs = super::indirect_specs_flag(scope, &shards.flags, errors);
```
williamhbaker, line 257: There's a wrinkle with runtime-V1 derivations being frozen on an image tag that hasn't been rebuilt in quite a while, and they will not be able to handle the modified fields for indirect specs. We won't be able to enable this on any derivation that is not permanently on runtime V2. This is mostly just a callout since I can't see us being super eager to turn this on for derivations, but something to be aware of.

crates/models/src/shards.rs
```diff
@@ -102,6 +102,17 @@ pub struct ShardTemplate {
...
+/// The `indirect-specs` [`ShardTemplate::flags`] key, which asserts that the
+/// task's built spec, its Validate and Apply requests, every hop within the
+/// data plane, *and its connector* all handle indirected collections: a
+/// per-message `linked_collections` table which bindings name by index,
+/// instead of a full `CollectionSpec` inlined in every binding.
+///
+/// The assertion is end-to-end and its only valid value is `"true"`. A
+/// connector which hasn't been updated simply doesn't work with a task that
+/// truly needs this flag -- such tasks don't function today in any case.
+pub const INDIRECT_SPECS: &str = "indirect-specs";
```
williamhbaker, line 114: Another callout, we'll need to update the `flow-connector-init` (runtime/src/container.rs:23) at some point after this merges. I think that's need for `flowctl preview` and `flowctl catalog test`. Perhaps in the follow-up memoization change.

crates/validation/src/linked.rs
```diff
@@ -0,0 +1,318 @@
+//! The indirect-encoding writer: an interner over `flow::CollectionSpec`, plus a
+//! per-message step which installs a finished interner into its message.
+//!
+//! Builders in this crate always construct specs and Validate requests in indirect
+//! form, interning each binding's collection as the binding is built. The
+//! `indirect-specs` shard flag then gates a single post-process: when it's unset,
+//! `install_*` *inlines* the message instead -- re-inlining each binding's
```
williamhbaker, line 7: In the backwards-migration case, it seems like `last_{capture/derivation/materialization}` specs would retain their interned forms as this currently works. I can't think of a very compelling reason that we should try to close that gap, since if something is going to failed with interned specs they'll probably do so up front and not succeed a publication and then later have a problem with those `last_*` specs. So, just another callout that I thought was interesting.

williamhbaker (dismissed):
LGTM

Everything made sense. A few observations below, nothing that I think requires any change to this PR though.

crates/models/src/shards.rs
jgraettinger: Yes, taking a note of this.

jgraettinger:
Rebased, and additionally updated the soak test to use `indirect-specs: true` (except for the materialize-postgres task) and confirmed that's working.

williamhbaker (approved):

PR #3312, opened 2026-08-04
Reactor Mem Usage Profiling Results Cleanup

dgreer-dev:
## Description

A heap profile from a busy production reactor showed ~70% of its 1.9GB live Go
heap was retained task-spec data. Two commits, one per cause:

**runtime: disable grpc.EnableTracing in the reactor**

Gazette's diagnostics boilerplate enables grpc.EnableTracing, and x/net/trace
pins every logged message until its trace finishes. V2 session streams live for
days and never hit trace compaction, so the sent Task message (embedding the
full marshalled spec) stayed pinned: ~365MB. The reactor now turns tracing off,
trading /debug/requests for the memory. 

**runtime: V2 task terms hold raw spec bytes, not decoded specs**

Go reads almost nothing from a V2 task's decoded spec; it existed to produce
Task message bytes for Rust (~1.0GB retained). Terms now hold the raw built
spec encoding straight from the build DB and forward it verbatim, so V2 shard
startup no longer decodes or re-marshals specs at all. V1 apps are unchanged.
Note two edge behavior changes: Rust receives the control plane's encoding
verbatim (the old gogo re-encode dropped unknown fields), and a corrupt spec
blob now errors in Rust at session start rather than in Go at term init.

## Notes for reviewers

Verified on a local stack with runtime-v2 captures using wide synthetic
schemas, plus V2 materializations and derivations. Before: heap shows Marshal
bytes 100% retained under runOneSession and decoded specs under binding
Unmarshal, matching the production profile. After: both attributions are gone,
leaving only the raw bytes by design. Sessions commit continuously, heartbeat
stats publish, and spec-update term transitions drain and restart cleanly.
`mise run ci:gotest` passes. No protocol or Rust-side changes.

### DIFF
--- a/go/protocols/catalog/build_load.go
+++ b/go/protocols/catalog/build_load.go
@@ -115,6 +115,28 @@ func LoadMaterialization(db *sql.DB, name string) (*pf.MaterializationSpec, erro
 	return out, loadOneSpec(db, `SELECT spec FROM built_materializations WHERE materialization = ?;`, out, name)
 }
 
+// LoadCollectionBytes, LoadCaptureBytes, and LoadMaterializationBytes return
+// the raw protobuf encoding of a named built spec, without decoding it.
+// The V2 runtime forwards the built encoding verbatim to the Rust runtime,
+// which is its sole consumer: loading bytes avoids materializing (and then
+// retaining) the decoded spec, which is large for tasks with many bindings
+// or wide collection schemas.
+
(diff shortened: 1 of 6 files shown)


dgreer-dev:
Per Will B:
> i'd say we should probably avoid making changes to the Go runtime for relatively small gains like this, since we plan on moving increasingly more of that to Rust over time, and its not worth the risk of a change right now.

PR #3313, opened 2026-08-04
Docs: Webhook capture organization and fixes

aeluce:
**Description:**

Cleans up reference docs for webhook captures, including:
* Groups webhook variants together like we do for other connector variants to make them easier to find
* Adds newer endpoint properties that were missing from some variants and removes extraneous EndpointConfig/ResourceConfig rows in property tables
* Updates Jira variant now that webhooks support HMAC SHA256 signature verification
* Removes generic usage sections from variants to keep the reference focused on their specific systems
* Other minor typo fixes

Moved reference pages keep original slugs, so no additional redirects are needed.

Open to other suggestions if any of this doesn't sound like the right trajectory. Thanks!

Generic HTTP ingest (webhook) capture docs, as well as variants for Datadog, Intercom, Jira, and Twilio.

**Notes for reviewers:**

It looks like `source-http-ingest` does create Jira and Twilio variants, but they're not available through the dashboard. I don't believe they were added to the database; could someone on the capture side add those to the `connectors` and `connector_tags` tables?


Alex-Bair (approved):
LGTM, thanks!

I added rows into `connectors` and `connector_tags` tables for both `source-jira-ingest` and `source-twilio-ingest`. They're visible in the UI as options when creating capture connectors.

While I was in there, I also noticed that `source-datadog-ingest`'s row in the `connector_tags` table has a documentation link of https://go.estuary.dev/http-ingest instead of something more specialized like https://go.estuary.dev/datadog-ingest. Should there be separate short links for each variant of the webhook capture connector that point to different docs instead of all pointing to https://go.estuary.dev/http-ingest?

aeluce:
Thanks! And yep, it'd be preferable to have specialized short links since we have some distinct instructions for the different variants.

Alex-Bair:
Gotcha. I made the short links for those variants missing them, updated `connector_tags` with them, and I redirected all the http ingest short links to the new paths under `/http-ingest`.

PR #3316, opened 2026-08-04
docs: source-brevo rewrite

Alex-Bair:
**Description:**

Documentation updates for https://github.com/estuary/connectors/pull/4980.

### DIFF
--- a/site/docs/reference/Connectors/capture-connectors/brevo.md
+++ b/site/docs/reference/Connectors/capture-connectors/brevo.md
@@ -1,26 +1,39 @@
 ---
-description: Capture Brevo contacts data with Estuary's connector, including attributes and lists, using API key authentication and a configurable start date.
+description: Capture Brevo contact data with Estuary's connector, including attributes, lists, folders, segments, senders, and webhooks, using API key authentication.
 ---
 
 # Brevo
 
-This connector captures data from [Brevo's REST API](https://developers.brevo.com/reference).
+This connector captures data from [Brevo's REST API](https://developers.brevo.com/reference) into Estuary collections.
 
 ## Supported data resources
 
-The following data resources are supported through the Brevo APIs:
-
-* [Contacts](https://developers.brevo.com/reference/getcontacts-1)
-* [Contacts Attributes](https://developers.brevo.com/reference/getattributes-1)
-* [Contacts Lists](https://developers.brevo.com/reference/getlists-1)
+| Resource | Replication | API reference |
+|---|---|---|
+| `contacts` | Incremental | [Get all the contacts](https://developers.brevo.com/reference/get-contacts) |
+| `contacts_attributes` | Snapshot | [List all attributes](https://developers.brevo.com/reference/getattributes-1) |
+| `contacts_lists` | Snapshot | [Get all the lists](https://developers.brevo.com/reference/getlists-1) |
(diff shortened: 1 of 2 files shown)


site/docs/reference/Connectors/capture-connectors/brevo.md
```diff
@@ -1,27 +1,40 @@
 ---
-description: Capture Brevo contacts data with Estuary's connector, including attributes and lists, using API key authentication and a configurable start date.
+description: Capture Brevo contact data with Estuary's connector, including attributes, lists, folders, segments, senders, and webhooks, using API key authentication.
 ---
 
 # Brevo
 
-This connector captures data from [Brevo's REST API](https://developers.brevo.com/reference).
+This connector captures data from [Brevo's REST API](https://developers.brevo.com/reference) into Flow collections.
```
aeluce:
```suggestion
This connector captures data from [Brevo's REST API](https://developers.brevo.com/reference) into Estuary collections.
```
Or `Estuary's data collections`. Still attempting to use "Estuary" rather than "Flow" when talking about the product.

site/docs/reference/Connectors/capture-connectors/brevo.md
```diff
@@ -33,32 +46,36 @@ See [connectors](../../../concepts/connectors.md#using-connectors) to learn more
...
 ### Sample
 
 ```yaml
-
 captures:
   ${PREFIX}/${CAPTURE_NAME}:
     endpoint:
       connector:
-        image: ghcr.io/estuary/source-brevo:v1
+        image: ghcr.io/estuary/source-brevo:v2
```
aeluce, line 66: May want to update the version tag on the main capture connector README as well.

(Though as a side note, it may not make sense to keep using that README page the way we have, since it's very easy for it to fall out of date. I haven't landed on a certain solution yet. Maybe something automated. Or maybe breaking the list out into connector categories to help organize database, file, SaaS, etc connectors. Any suggestions are welcome!)

aeluce (dismissed):
Nice! Capture team has been keeping very busy lately. 😄 Just a couple nitpicky suggestions, but would be fine to merge as-is.

site/docs/reference/Connectors/capture-connectors/brevo.md
Alex-Bair: Thanks for catching this! I missed it. 🤦 Fixed.

site/docs/reference/Connectors/capture-connectors/brevo.md
Alex-Bair: Good point, I've updated the main capture connector README to reference the new `v2` image tag instead.

I don't particularly feel like having the package listed on that README page is particularly useful (or maybe it's just the tag I don't think is useful there?). Do users actually find that information useful in when they're reading that README? I think the package & tag are useful when trying to set up a capture via `flowctl`, but if I'm doing that, I'd rather go to the individual capture's page to learn how I should configure the task anyway. I guess, my opinion is that we shouldn't include the package on whatever index we land on for listing the connectors we support.

aeluce: Agreed. I wouldn't think one of the first things a user would need to know about a connector would be the package. I _think_ we now have example specs on all reference pages; I'd just want to double check that before removing to make sure we're providing the package name and current image tag somewhere for each connector.

aeluce (approved):

PR #3318, opened 2026-08-05
control-plane: data-plane selection at signup + colocated trial bucket (gated)

SeanWhelan:
**Description:**

New tenants can now pick a specific public data plane at signup. The `betaOnboard` directive claims gain an optional `requestedDataPlane` (full plane catalog name, e.g. `ops/dp/public/aws-us-east-1-c1`), validated server-side before it becomes the tenant's default data plane in their `storage_mappings`.

This also lands the backend half of colocated trial buckets (Phase 2 of estuary/sre#29). When the chosen plane is a public AWS plane and `COLOCATED_TRIAL_BUCKETS` is enabled, `provision_tenant` derives that plane's colocated S3 bucket instead of `gs://estuary-trial`, using the same naming formula as est-dry-dock#326 (`estuary-trial-{region}-{sha256(plane_name)[:8]}`, golden-vector-tested on both sides). The gate stays off in this PR. Flipping it is a separate follow-up once est-dry-dock#326 merges and real public planes have actually converged and gotten a bucket. Until then, and always for GCP/Azure planes, nothing changes: tenants still land on `gs://estuary-trial`.

Companion PR: estuary/ui#2045, which adds the picker itself. That one must merge and deploy after this one, not before (see notes below).

**Workflow steps:**

The onboarding UI shows a data-plane picker between the org name field and the survey (companion PR). Submitting without touching it still records an explicit default: `aws-us-east-1-c1`, or the first available plane if that one isn't present.

The submitted plane is untrusted client input. `provision_tenant` selects the planes a new tenant may use — public prefix, `not closed` — and rejects anything outside that set with `ProvisionError::PlaneNotSelectable`, which the directive turns into `invalidClaims` naming the rejected plane. If no claim is submitted at all, behavior is identical to today: `aws-us-east-1-c1` default, `gs://estuary-trial`.

To turn on colocated buckets later, set `COLOCATED_TRIAL_BUCKETS=1` (or `true`) on the agent's environment. No code change needed.

None. This is control-plane/agent-internal behavior and no user-facing docs cover the signup bucket-selection mechanics today.

**Notes for reviewers:**

Deploy order matters here. The `Claims` struct uses `deny_unknown_fields`, so this needs to merge and deploy before the UI PR ships. If an old agent gets `requestedDataPlane` from a new UI, signup fails with `invalidClaims` naming the unknown field. I checked this failure mode live, not just in theory.

`data_planes.closed` is the single source of truth for what a new tenant may be placed on, and the migration in this PR marks `gcp-us-central1-c1`/`c2` closed. I started out with a hardcoded exclusion list for those two, which left "selectable" defined in three places that could disagree — the list, the SQL building the tenant's `data_planes`, and a separate existence query in the agent — while `provision_tenant` ignored `closed` altogether. That combination meant closing a plane would hide it from the picker (`publicDataPlanes` already honors `closed`) but keep writing it into every new tenant's storage mapping. Now there's one query and one predicate, and retiring a plane is a data change rather than a code change plus a deploy. Existing tenants on those planes are unaffected; `closed` only governs new selection.

In `crates/control-plane-api/src/directives/beta_onboard.rs`, `provision_tenant` now computes plane ordering (`order_public_planes`) and storage specs (`storage_specs`) in Rust instead of a single hardcoded CTE. This is behavior-preserving for existing callers (`requested_data_plane: None`, `colocate_trial_bucket: false`), confirmed by the integration test's unchanged `AcmeTenant` snapshot.

`trial_bucket_name` parses the plane name itself and returns `Option<(bucket, region)>` rather than taking a region alongside the name, so a bucket can't be derived from a region that disagrees with its plane. `parse_data_plane_name` and `DataPlaneCloudProvider` also moved from `server::public::graphql::data_planes` to `crate::data_plane` — that logic was living in a GraphQL-presentation module despite being plain domain parsing the directive code needed too.

The golden vector is pinned as a unit test: `ops/dp/public/aws-us-east-1-c1` gives `estuary-trial-us-east-1-ccc98e22`. If this ever fails, the Rust and Python (est-dry-dock) derivations have drifted apart. Fix the drift, not the test.

262 tests across `agent` and `control-plane-api` pass, along with `cargo sqlx prepare --workspace --check`. Note these have to run under `cargo nextest` — `.config/nextest.toml` serializes the agent's DB tests via the `serial-db-tests` group, and under plain `cargo test` they run in parallel against one database, truncate each other's fixtures, and throw dozens of failures that look like real breakage but aren't.

Worth a look during review: the `beta_onboard` snapshot gains exactly one case in this PR — a closed public plane rejected with `invalidClaims` — and every pre-existing case is byte-identical. That diff is the evidence the refactor didn't change behavior.

I also live-tested this end to end on a local stack: a real signup through the actual onboarding form, gate on, plane matching the AWS naming convention, produced a `storage_mappings` row with the exact derived S3 bucket name. After applying the migration, `publicDataPlanes` drops the closed plane and the picker renders the rest.

Known follow-ups, deliberately not in this PR: `storage_specs` still builds its JSON by hand rather than through `models::StorageDef`, and `COLOCATED_TRIAL_BUCKETS` is read via an env var rather than a clap `Args` field. The second is why the colocated S3 path has unit coverage but no end-to-end coverage — worth closing before the gate is flipped on, since the harness could then exercise it.


crates/control-plane-api/src/directives/beta_onboard.rs
```diff
@@ -50,22 +72,109 @@ pub async fn tenant_exists(
     Ok(illegal.is_some() || exists.is_some())
 }
 
+/// The plane new tenants default to when signup carries no data-plane choice.
+pub const DEFAULT_PUBLIC_DATA_PLANE: &str = "ops/dp/public/aws-us-east-1-c1";
```
adrian-estuary, line 76: I thought `aws-us-east-1-c1` was to be closed and `us-east-1-c2` succeeded it. Should this be `c2`?

Also, how does this change relate to https://github.com/estuary/flow/pull/3402?

supabase/migrations/20260804120000_close_deprecated_gcp_data_planes.sql
adrian-estuary, line 1: Looks like a new migration was added to master since this PR was opened. Will need to bump the timestamp of this migration for it to get picked up.

crates/control-plane-api/src/directives/beta_onboard.rs
```diff
@@ -163,3 +262,126 @@ pub async fn provision_test_tenant(
     txn.commit().await.expect("commit tenant");
     user_id
 }
+
+#[cfg(test)]
+mod test {
+    // The golden vector is shared with est-dry-dock's Python implementation
+    // (est_dry_dock/models/__init__.py::trial_bucket_name). If this assertion
+    // ever fails, the two implementations have drifted and a tenant's storage
+    // would be misrouted — fix the drift, never the test.
```
adrian-estuary, line 271: How is this enforced? Is there a similar test / CI check within `est-dry-dock`?

crates/control-plane-api/src/directives/beta_onboard.rs
```diff
@@ -50,22 +72,109 @@ pub async fn tenant_exists(
...
         .filter(models::status::AlertType::is_default)
         .collect();
 
-    // Note that the gcp-us-central1-c1 (combustible-cronut) dataplane is excluded here
-    // because it's being deprecated and replaced.
+    // The set of planes a new tenant may be placed on. `closed` is the single
+    // source of truth for "retired from new selection" — the same flag
+    // `publicDataPlanes` honors, so the signup picker and provisioning cannot
+    // disagree, and retiring a plane is a data change rather than a deploy.
+    let public_planes: Vec<String> = sqlx::query_scalar!(
```
adrian-estuary, line 153: Seems like it's technically possible for `public_planes` to return empty here (though not likely in the real world). However, if it does return an empty Vec, it would get passed to the provisioning logic, which I _think_ would create storage mappings without any data planes, which I _think_ is an invalid state.  

The GQL layer checks for empty data planes in the `create_storgae_mapping` resolver, so this would be a divergence in behavior. 

So I think adding a defensive check for if `public_planes` is empty would be good. But definitely double check my thought process here.

adrian-estuary (changes requested):
Added a few comments, some of which are blocking. Mainly updating the migration.

SeanWhelan:
Thanks for the review. All four are addressed:

- Migration: bumped to 20260923120000 (73d470d).
- Empty `public_planes`: it now returns `NoSelectablePlanes` instead of writing a mapping with no planes. The agent treats it as an internal error (e7c3e86).
- Default plane: switched to `c2` (a18bbf0), which supersedes [#3402](https://github.com/estuary/flow/pull/3402). It's temporary, only covers signups from UI builds without the picker. Once [estuary/ui#2045](https://github.com/estuary/ui/issues/2045) ships, we can follow up to make `requestedDataPlane` required and remove the constant. The picker preselects the newest open plane in AWS us-east-1, so replacing a plane is just closing the old one.
- Golden vector: est-dry-dock's `test_trial_bucket_name_is_deterministic` asserts the same vector in its CI. Nothing checks across the two repos, so changing the algorithm and the expected value together in one repo would slip through. The formula shouldn't change so I think that's acceptable.

adrian-estuary (dismissed):
LGTM!

adrian-estuary (approved):

PR #3321, opened 2026-08-06
docs: use :::deprecated admonition for deprecated connector notices

jwhartley:
**Description:**

Registers a custom `deprecated` admonition type and switches the 16 deprecated connector pages to `:::deprecated`, so they can be distinguished from ordinary warnings.

**Workflow steps:**

No end-user workflow change. This gives Kapa's docs-search crawler a stable `theme-admonition-deprecated` CSS class to exclude deprecated connector pages from search results. Neither alternative works: Kapa can't match on page title text, and no URL/filename convention separates deprecated from current pages (Jira has 4 pages, only 2 deprecated). Excluding on `:::warning`'s class would drop the 35 unrelated warnings elsewhere on the site.

The component is ported verbatim from `estuary/docs`, which already implements it, keeping the two repos consistent.

None beyond the 16 connector pages. Reader-facing appearance is unchanged, since `Deprecated.js` reuses `:::warning`'s icon and `alert--warning` styling. The exception is `google-analytics.md`, previously the lone `:::danger` page, which now matches the other 15.

**Notes for reviewers:**

- Equivalent changes already merged in `estuary/connectors` and `estuary/docs` don't reach production, because `docs.estuary.dev` builds only from `site/**` in this repo.
- **Swizzle safety:** `Types.js` is purely additive over the bundled 3.9.2 copy (one import, one map entry) and preserves all 6 built-in types plus the 4 legacy aliases. Verified by diffing a full local build against unmodified `master`: **275 admonitions in both**, with `warning` 50 → 35, `danger` 1 → 0, `deprecated` 0 → 16, and `note`/`tip`/`info`/`caution`/`important` identical. Nothing hit Docusaurus's unknown-type fallback.
- **Pre-existing local build failure, unrelated to this PR:** `npm run build` also fails on unmodified `master` on case-insensitive filesystems, where the `hubspot-real-time` → `HubSpot-real-time` redirect collides. Linux CI is unaffected.
- **After merge:** the selector for Kapa to exclude is `.theme-admonition-deprecated`. Nothing needed in `estuary/docs`, whose deploy runs `git submodule update --remote` hourly and picks up the matching content automatically.

### DIFF
--- a/site/docs/reference/Connectors/capture-connectors/HubSpot.md
+++ b/site/docs/reference/Connectors/capture-connectors/HubSpot.md
@@ -3,7 +3,7 @@
 
 This connector captures data from a Hubspot account.
 
-:::warning
+:::deprecated
 This connector is deprecated. It is recommended that you use the new native [HubSpot connector](/reference/Connectors/capture-connectors/HubSpot-real-time) instead.
 :::
 
--- a/site/docs/reference/Connectors/capture-connectors/Salesforce/salesforce-historical-data.md
+++ b/site/docs/reference/Connectors/capture-connectors/Salesforce/salesforce-historical-data.md
@@ -6,7 +6,7 @@ It uses batch processing and is ideal for syncing your historical Salesforce dat
 [A separate connector is available for real-time Salesforce data capture](./salesforce-real-time.md).
 For help using both connectors in parallel, [contact your Estuary account manager](mailto:info@estuary.dev).
 
-:::warning
+:::deprecated
 This connector is deprecated. Consider using the new [native Salesforce connector](./salesforce-native.md) instead.
 :::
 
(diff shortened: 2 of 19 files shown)


aeluce (approved):
Approved, though docs served from the flow repo don't use Kapa search; is this mainly to let us to configure Kapa ahead of switching to docs and connectors repos? Changes already approved and added to future docs setup in https://github.com/estuary/docs/pull/11 and https://github.com/estuary/connectors/pull/4862.

There's also some newly deprecated connectors, like original-flavor SQS and Mailchimp captures.

jwhartley:
Yes, it's so we can configure + test ahead of switching. Noted on SQS and Mailchimp

PR #3325, opened 2026-08-07
docs: document adopting existing destination tables

jwhartley:
Attaching a materialization to a destination table that already exists was undocumented, so there was nothing to find for either case people actually hit: pre-creating a table to control DDL that Estuary does not manage, and migrating off another pipeline without dropping the destination.

## Changes

- **New guide**, [Migrate an Existing Pipeline to Estuary](site/docs/guides/migrate-to-estuary.md) at `/guides/migrate-to-estuary/`. Splits the empty-table case (one feature flag) from the populated-table case (also needs Exclude Flow Document and a cutover filter), since the trigger is rows existing rather than the table existing.
- **BigQuery reference**: splits Table Partitioning into "Partitioning a new table", the previously undocumented pre-create path, and "Repartitioning an existing table", the pre-existing procedure. Partitioning cannot be altered after creation, so pre-creating is the only way to avoid the copy-and-rename conversion, which matters when adding many tables.
- **feature-flags**: notes that `allow_existing_tables_for_new_bindings` is not sufficient once the table holds rows, and describes what the connector actually does to an adopted table (adds missing columns as nullable, drops `NOT NULL` on unselected columns, never recreates the table).
- **customize-materialization-fields**: gives Exclude Flow Document its migration motivation, not just storage savings.

## Two claims worth a second opinion

Both come from reading the source rather than from support threads, and both contradict guidance that has circulated internally:

- **`notBefore` filters on collection publish time**, not on when the row changed in the source (`crates/models/src/source.rs`). So a timestamp copied from the outgoing tool's last sync filters nothing once the capture backfills, and the whole backfill lands in the existing table. The guide recommends Only Changes mode, or `notBefore` set to the backfill's completion, instead.
- **Exclude Flow Document takes effect on an existing binding's next transaction**, so it does not require recreating the binding. The load template is selected from the endpoint config alone, per transaction (`materialize-bigquery/transactor.go`).

## Notes for review

- The page deliberately does not quote a specific error for the null-document failure, because the text differs by connector and BigQuery's is likely to change (estuary/connectors#5029). It may be too vague as a result.
- No verification or rollback section. Both were cut for length and are defensible additions.
- All internal links were checked against declared `slug:` frontmatter rather than file paths, since several targets (`time-travel`, `backfilling-data`, `materialization`) sit at directories that do not match their URLs.
- Sidebar is autogenerated, so the new page needs no registration.

### DIFF
--- a/site/docs/guides/advanced-usage/feature-flags.md
+++ b/site/docs/guides/advanced-usage/feature-flags.md
@@ -129,9 +129,11 @@ On the first publication after attaching, the connector reconciles the table aga
 
 - Columns in the field selection that the table does not have are added with `ALTER TABLE ... ADD COLUMN`, as nullable. This includes Estuary's own `flow_document`, `flow_published_at`, and `_meta/op` columns.
 - Columns the table has that are **not** in the field selection are altered to drop `NOT NULL`, if they had it. Keep this in mind when you pre-create a table from DDL you care about: a `NOT NULL` constraint on a column Estuary does not materialize will be relaxed.
-- The table itself is never dropped or recreated, so partitioning, clustering, and other table-level DDL survive.
+- The table itself is not dropped or recreated by this reconciliation, so partitioning, clustering, and other table-level DDL survive it.
 
 Because missing columns are added for you, you can pre-create a table with only the columns you need to control (the partition column and the collection key, for instance) and let the connector fill in the rest.
+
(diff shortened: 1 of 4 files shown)


jwhartley:
Scrapping this as generally we recommend backfilling in Estuary rather than trying to in-place migrate a dataflow

PR #3326, opened 2026-08-07
Runtime "Sync Now"

### LINKED ISSUE #3281: synchronous "sync now" API
# Sync Now: force an immediate commit of a materialization

## Changes at a glance

- New Flow capability bit (`1 << 21`), minted by `/authorize/user/task` at
  Write+.
- New verified gRPC service on the reactor front door: resolve shard zero,
  self-proxy to its primary, relay to the co-located sidecar. No semantics in
  Go.
- New unary-in/stream-out service in `runtime.proto`, served by the sidecar
  behind its own capability interceptor.
- Leader FSM: sync-now poke = `close_requested` + completion waiter, resolved
  at `Tail::Done`.
- New command: `flowctl raw sync-now --task ...`.
- **Open question**: restrict to schedule-held transactions, or a general
  "commit what you have" barrier?

## Motivation

The runtime sync schedule (#3229) holds a materialization's open transaction
(which continues combining new documents) until the next scheduled commit
instant. That's the right default for cost control, but it conflicts with a
common workflow: "materialize everything you have, then run my analytics." A
user with a nightly-paced warehouse materialization should be able to write

`flowctl raw sync-now --task acmeCo/foo/materialize-bar && run-analytics.sh`

and know that when the first command exits, everything the materialization had
read at the moment of the call is committed and queryable in the destination.

This ticket delivers the data-plane RPC plus the `flowctl raw sync-now`
command. This is a runtime-next (V2) feature only, materializations only: V1
tasks are entirely out of scope.

## Semantics

Sync-now is a one-shot poke delivered to the task's leader actor. Mechanically
it sets the FSM's existing `close_requested` input and registers a completion
waiter. `close_requested` bypasses policy *floors* (the sync-schedule hold,
configured minimum transaction duration; see the lifecycle-flag rules in
`crates/runtime-next/src/leader/close_policy.rs`) but never *structure*:
`tail_done` gating and unresolved-hint coherence still hold, so a caller
cannot force an incoherent or partial commit. The worst achievable outcome is
more, smaller transactions.

- Open transaction held by the schedule: the hold collapses and it proceeds to
  commit (`HELD_COLLAPSED`).
- Open transaction already closing: no-op (`ALREADY_CLOSING`).
- No open transaction: no-op, the task is already current (`IDLE`).
- Open transaction, no schedule hold in effect: see the open question below.
- Derivations/captures: not applicable (`NOT_APPLICABLE`); never an error for
  "nothing to do".
- Concurrent and repeated calls are explicitly allowed and must be safe: every
  call is an independent idempotent poke, and N concurrent calls become N
  waiters on the *same* commit, all resolving together. No rate limiting, no
  queuing, no coalescing machinery. A cron job, a retrying script, and an
  impatient human can all overlap without harm.

**Definition of done.** The call completes when the transaction that was open
at poke time is fully acknowledged - its contents are committed and visible in
the destination. Data arriving after the poke rides the next transaction and
is out of luck. Done does NOT mean "caught up to the capture's frontier": if
the task is behind on reading, sync-now commits what has been combined so far.
A "wait until caught up to clock X" semantic is a different feature and out of
scope.

## Open question: schedule override, or a new screwdriver in the toolbox?

What should sync-now do when the open transaction is *not* schedule-held -
it's just a big transaction, or one pinned by a configured minimum duration?

**Option A - restrict to schedules**: sync-now only collapses a
sync-schedule hold; anything else is a no-op with an explanatory outcome. The
feature stays exactly as motivated, and the blast radius is limited to tasks
that opted into scheduling.

**Option B - general commit barrier**: sync-now means "commit what you have
and tell me when it's queryable", regardless of why the transaction is open
(new outcome `CLOSE_REQUESTED`). 

## RPC shape: server-streaming with two messages

The RPC streams exactly two messages:

1. **Ack** (immediate): the outcome enum - `HELD_COLLAPSED | ALREADY_CLOSING |
   IDLE | NOT_APPLICABLE`, plus `CLOSE_REQUESTED` under Option B - and a
   snapshot of the open transaction's extents (sourced docs, bytes, open age),
   which the FSM already tracks. Immediate confirmation of what the call is
   waiting on.
2. **Done** (on completion): committed stats for the transaction. Sent when
   the leader FSM's `Tail` reaches `Done` for the target clock recorded at
   poke time.

A caller that doesn't want to wait just hangs up after the ack; the poke has
already landed server-side, so early hangup is harmless. This avoids two
server-side code paths for wait/no-wait.

## Dependency: connector-side ack schedules must not stack

**This is the one external blocker, and it needs a decision.** Eleven
warehouse-family connectors (snowflake, bigquery, databricks, redshift,
motherduck, clickhouse, fabric-warehouse, both icebergs, filesink) implement
the *legacy* sync schedule by delaying their `Acknowledged` response - and it
is **default-on**: an absent connector `syncSchedule` means a 30-minute fixed
ack delay (`defaultSyncFrequency` in
`connectors/go/materialize/schedule_config.go`). That delay lives inside the
connector, beyond the runtime's reach. Sync-now can collapse the runtime hold,
but the commit then sits in the connector's ack delay for up to 30 minutes:
the wait stays *correct* (it resolves on true completion), but the caller's
script blocks on a delay the runtime cannot collapse, which defeats the
feature.

The build-time `SyncScheduleConflict` check rejects a *configured* connector
schedule alongside a model schedule, but today:

- absent connector config + model schedule passes validation and silently
  stacks with the 30m default;
- the connector's documented "off" state (`syncFrequency: "0s"`) is counted as
  configured and *rejected* - so there is currently **no expressible config
  state that yields runtime-only pacing** on these connectors.

Proposal:

1. **Validator carve-out**: treat an all-zero connector schedule
   (`syncFrequency: "0s"`, no fast-sync fields) as unconfigured in
   `connector_config_has_sync_schedule`. Explicit `0s` becomes the sanctioned
   pairing and works on every connector image already shipped.
2. **Gate model `syncSchedule` on the V2 runtime flag** at validation. Today a
   V1 task can carry a model schedule that is silently inert; after the
   deferral change below it would be actively harmful (no pacing at all).
3. **Boilerplate deferral**: the connector receives the built
   `MaterializationSpec`, which now carries `sync_schedule_json`; change
   shared boilerplate so a spec-level schedule forces the connector ack
   schedule off (config and default both). One change covers all eleven
   connectors; V1 tasks are untouched. Ships on the normal image cadence;
   explicit `0s` is the bridge until a task's image has it.
4. Eventual removal of the connector-side machinery once conversion is
   complete and a census shows no remaining users; the config field stays as a
   tolerated no-op for stored-config compatibility.

If the open question resolves to Option B, this dependency has a second face:
an *unscheduled* warehouse task has no model schedule, so the deferral never
triggers and the connector's default 30m ack delay is in play - a sync-now
wait on such a task can legitimately block on it. Correct, never early, but
slow. The mitigation is the same explicit `syncFrequency: "0s"` (no conflict
arises without a model schedule); document it.

Sync-now itself builds no detection machinery for this: it's a documented
dependency, not a runtime check.

## Failure modes

- **No live leader session** (task disabled, V1 task, mid-restart, shard zero
  relocated): sidecar returns NOT_FOUND; the Go handler surfaces it; flowctl
  retries `Shard.List` once to catch a relocation, then errors with a hint
  that the task may not be running on the V2 runtime.
- **Session dies mid-wait** (shard reassignment, leader restart): parked
  waiters get an error; re-invoking is idempotent and re-establishes the
  contract ("all data as of the retry").

## flowctl

`flowctl raw sync-now --task acmeCo/foo/materialize-bar`, reusing the
`TaskSelector` and `user_task_authorization` plumbing from `raw list-shards`.
Waits for done by default (the script use case); a `--no-wait` flag hangs up
after the ack.

## Out of scope / follow-on

- A dashboard "Sync Now" affordance calling this same RPC. Note for that work:
  browsers can't speak raw gRPC, so the UI needs a grpc-web/CORS layer on the
  reactor path or a proxy; recorded here so the dashboard team knows the gap.
- Derivations (coherent under Option B; natural follow-on if wanted).
- Any V1-runtime behavior.
- "Caught up to frontier" wait semantics.
- Hold-suppression windows, "skip next hold" arming, rate limiting.

dgreer-dev:
`flowctl raw sync-now --task X` (and `POST /v1/task-control/sync-now`) forces an
immediate commit of a V2 materialization's open transaction and returns once that
transaction is committed and queryable in the destination. Closes #3281.

The workflow we want to enable is a user wanting to run analytics on the freshest
data available, regardless of any sync schedule (or other transaction-holding
reason).  Essentially we want to allow `flowctl raw sync-now ... && ./my-analyitcs.py`
to be a safe and useful command.  See below for specifics of the new command.

### Design points worth your attention

- **General commit barrier.** Sync-now means "commit what you have, tell me when
  it's queryable", whatever is holding the transaction (schedule, min-duration
  floor, or just size). Worst case is a smaller transaction. "Nothing to do" is
  never an error: an idle task Acks `IDLE`, a capture or derivation
  `NOT_APPLICABLE`, both exit 0.
- **One wait rule.** Every outcome but `IDLE`/`NOT_APPLICABLE` resolves at the
  target transaction's `Tail::Done`, so N concurrent callers park on one commit.
- **Server-side routing, no semantics in Go.** The front door resolves shard
  zero with the caller's claims, proxies to the primary's peer via `mayProxy`,
  and relays 1:1 to the co-located sidecar.
- **Two transports, one dialect.** `/v1/task-control/sync-now` exists as a second
  way to force the sync.  Calls the same code as flowctl, just an HTTP different path.
- **Validation arm** (first commit, independent of the runtime): an all-zero
  connector `syncSchedule` counts as unconfigured, so runtime-only pacing is
  expressible on every connector image already shipped; and a model-level
  `syncSchedule` without `enable-runtime-v2` is now rejected rather than
  silently inert.

### Verification

Rust e2e tests ensure transaction control works as advertised. Go tests stand up
two consumer test members against a stub sidecar, covering both transports
including the peer-proxied path. flowctl's response handling is covered against a
stub server over a real channel.

Also verified on a local stack against a schedule-held V2 materialization
(transcripts available). Highlights: three parallel `flowctl` pokes plus one
`curl -N` stream against one held transaction all exited 0 reporting identical
committed stats (one `HELD_COLLAPSED`, three `ALREADY_CLOSING` waiters);
destination row count equalled the pre-poke source count exactly on every run.

### Example Usage

```console
$ flowctl raw sync-now --task acmeCo/foo/materialize-bar
{"ack":{"outcome":"HELD_COLLAPSED","status":{"sourcedDocsTotal":"50","sourcedBytesTotal":"13691","openAgeMillis":"98573","headPhase":"Idle","tailPhase":"Done"}}}
{"done":{"committedDocsTotal":"50","committedBytesTotal":"13691","durationMillis":"99634"}}
```

Ack is immediate and carries the outcome plus a snapshot of the awaited
transaction. Done arrives when that transaction is committed and queryable in
the destination, with its committed stats. `durationMillis` is the
transaction's total open age, not the commit duration. Progress heartbeats
(~15s) are consumed for liveness but not printed unless `--progress`, since an
hour of them is a lot of lines unasked.

Any completed outcome exits 0, `IDLE` and `NOT_APPLICABLE` included - "nothing
to do" is success. Exit is non-zero for `NOT_FOUND` (not running here, or not
on the V2 runtime) and for `--timeout` elapsing.

Flags: `--progress` prints the heartbeats; `--no-wait` prints the Ack and hangs
up, which is harmless because the commit has already been forced; `--timeout
10m` bounds the whole invocation, which otherwise waits indefinitely and retries
leader restarts.

The same call over the REST transport, which is what the dashboard button will
use:

```console
$ curl -N -H "Authorization: Bearer $TOKEN" \
    -d '{"task":"acmeCo/foo/materialize-bar"}' \
    https://$REACTOR/v1/task-control/sync-now
{"result":{"ack":{"outcome":"HELD_COLLAPSED","status":{...}}}}
{"result":{"done":{"committedDocsTotal":"49",...}}}
```

### Commits
1. Types and basic verification
2. Wiring runtime-next
3. Go -> runtime pathway
4. e2e tests


crates/flowctl/src/raw/sync_now.rs
```diff
@@ -0,0 +1,489 @@
...
+use std::time::Duration;
+
+#[derive(Debug, clap::Args)]
+#[clap(rename_all = "kebab-case")]
+pub struct SyncNow {
+    #[clap(flatten)]
+    task: crate::ops::TaskSelector,
+    /// Also print the periodic progress heartbeats of the awaited transaction.
+    #[clap(long)]
+    progress: bool,
```
williamhbaker: I don't think the `progress`, `no_wait`, and `timeout` options are super well motivated right now. `progress` in particular I'll have more comments on later in the PR. `no_wait` and `timeout` just seem YAGNI.

crates/proto-flow/src/runtime.rs
```diff
@@ -1407,6 +1407,138 @@ pub mod derive {
...
+    /// For IDLE and NOT_APPLICABLE outcomes all fields are zero-valued.
+    /// Statistics are also zero-valued when the awaited transaction was
+    /// recovered from a prior leader session, as happens when a leader
+    /// restarts with a committed but un-acknowledged transaction: the
+    /// transaction's extents predate this session and are unknown to it.
+    #[derive(Clone, PartialEq, Eq, Hash, ::prost::Message)]
+    pub struct Status {
+        /// Number of source documents processed by the transaction so far.
+        #[prost(uint64, tag = "1")]
+        pub sourced_docs_total: u64,
```
williamhbaker: The doc / byte counts and open age are redundant with the recorded stats for the transaction. Rather than repeat them here in the sync now status, I think we'd defer to the stats journal. That is, the expectation is that if a user wants to know the transaction stats, they'd need to read the stats journal, rather than having explicit response fields here.

I'll also challenge `head_phase` and `tail_phase`: I don't think they are necessary as response fields. I'd suggest we keep this protocol as minimal as we can for now.

crates/proto-flow/src/runtime.rs
```diff
@@ -1407,6 +1407,138 @@ pub mod derive {
...
+        /// through full acknowledgement. This is NOT the duration of the
+        /// sync-now await, which is typically far shorter.
+        #[prost(uint64, tag = "3")]
+        pub open_age_millis: u64,
+    }
+    /// Outcome conveys what SyncNow did upon arrival at the task's leader,
+    /// and implies what the remainder of the stream awaits.
+    #[derive(Clone, Copy, Debug, PartialEq, Eq, Hash, PartialOrd, Ord, ::prost::Enumeration)]
+    #[repr(i32)]
+    pub enum Outcome {
```
williamhbaker: Similar to the above thoughts...I'd simplify this down. Do we really need an outcome enum? Very simplistically, I'd think along the lines of it either worked, or it didn't, rather than these granular descriptions.

crates/runtime-next/src/leader/materialize/actor.rs
```diff
@@ -37,6 +37,16 @@ pub struct Actor<P: crate::Publisher, L: crate::Logger> {
     shard_tx: Vec<mpsc::UnboundedSender<tonic::Result<proto::Materialize>>>,
     // Future for an in-flight stats flush, if any, yielding ACK intents.
     stats_write_fut: Option<BoxFuture<'static, tonic::Result<(P, BTreeMap<String, Bytes>)>>>,
+    // Clock of the last sync-now Progress heartbeat broadcast.
+    sync_now_heartbeat_at: uuid::Clock,
+    // Parked sync-now waiters, resolved as `tail_done_count` reaches their targets.
+    sync_now_waiters: Vec<SyncNowWaiter>,
```
williamhbaker: I know I've mentioned this, so there's probably a good reason not to do it, but I can't resist the urge to ask again: Rather than a list of all sync now callers that we accumulate, can we just keep track of the last one, and cancel all previous ones if a new caller arrives? That's a little less potential for problematic caller patterns to cause issues on our end.

crates/runtime-next/src/leader/materialize/actor.rs
```diff
@@ -45,6 +55,17 @@ pub struct Actor<P: crate::Publisher, L: crate::Logger> {
...
+/// A parked `TaskControl.SyncNow` caller, resolved when the actor's count of
+/// Tail::Done transitions reaches its target.
+struct SyncNowWaiter {
+    target: u64,
+    reply_tx: mpsc::UnboundedSender<tonic::Result<proto::SyncNowResponse>>,
+}
+
+/// Cadence of Progress heartbeats to parked sync-now waiters. Coarse: beats
+/// keep hour-long streams alive through LB idle timeouts and feed UI progress.
+const SYNC_NOW_HEARTBEAT: Duration = Duration::from_secs(15);
```
williamhbaker: There's mention of this keeping streams alive through idle timeouts. Is that the primary motivator? For the "is it doing anything" question, I'd defer to task logs, rather than this as a separate mechanism.

TBH I'd be a little surprised if we needed to send our own explicit heartbeat message to keep the connection alive, since things like TCP keepalive exist. A bit outside my area of expertise though, and I haven't thought about it much.

I do think the targeted, minimalist scope we want to maintain here is basically "you tell us that you want to the txn to commit, and we'll tell you when it's done". Any other nice user-facing features we should defer and endeavor to surface through the mechanisms we already have.

crates/validation/src/materialization.rs
```diff
@@ -695,6 +695,14 @@ async fn walk_materialization<C: Connectors>(
             if let Err(detail) = sync_schedule.validate() {
                 Error::SyncScheduleInvalid { detail }.push(scope, errors);
             }
+            // A model-level sync schedule is silently inert under the V1
+            // runtime, so reject it outright.
+            if crate::flag_value(&shards.flags, models::ENABLE_RUNTIME_V2) != Some("true") {
```
williamhbaker: I think it's fine to leave this as being silently inert under V1. We'll eventually be removing the runtime-v2 flag entirely and having every on V2, so I don't think we should add more than absolutely necessary in terms of these checks, since it will just be more than we have to remove.

crates/validation/src/materialization.rs
```diff
@@ -1168,15 +1176,32 @@ fn temporary_group_by_migration(
 /// An empty `syncSchedule` object, or one whose values are all null or empty
 /// strings, is NOT configured: the connector treats those identically to an
 /// absent key, and a UI removing the schedule may leave such a remnant behind.
+///
+/// A zero-duration `syncFrequency` with no other fields set is also NOT
```
williamhbaker: I'm not fully understanding this change. A connector-side syncFrequency of `0s` is fine to interpret as being configured. I don't think it's buying us anything to try to handle it otherwise?  It would be kind of confusing to have the connector config say "0s" and then have the runtime config say something else. It should just be blank in the connector config, or completely deleted.

go/runtime/task_control.go
```diff
@@ -0,0 +1,245 @@
...
+	sidecarEndpoint func(reactor pb.Endpoint) (string, error)
+}
+
+var _ pr.AuthTaskControlServer = &taskControl{}
+
+func (tc *taskControl) SyncNow(claims pb.Claims, req *pr.SyncNowRequest, stream pr.TaskControl_SyncNowServer) error {
+	return tc.syncNow(stream.Context(), claims, req, stream.Send)
+}
+
+// syncNow is the relay core shared by the gRPC and HTTP transports:
```
williamhbaker: I think we should have only the one transport for this, and that it should be the HTTP one. I don't think we need to keep a lower-level gRPC API available for `flowctl` to use, and should instead have `flowctl` use the HTTP path also.

williamhbaker (commented):
Some directional comments! I didn't get into the gritty details just yet since I am hoping things can be simplified quite a bit.

crates/flowctl/src/raw/sync_now.rs
dgreer-dev: Okay I will remove all of that

crates/proto-flow/src/runtime.rs
dgreer-dev: Will remove this too

crates/proto-flow/src/runtime.rs
dgreer-dev: Got it

go/runtime/task_control.go
dgreer-dev: Happy to do this either way, but `ListShards` does use GRPC fwiw

crates/runtime-next/src/leader/materialize/actor.rs
dgreer-dev: Heartbeats are pretty normal for long lived HTTP connections.   Javascript doesn't really let you control the low level socket and you're always at the mercy of proxies and things in the way that are pretty eager to drop dormant connections.  The SSE spec recommends sending a comment every 15 seconds for this reason.

Even in the gRPC case you're still left with an unknown number of proxies between the client and the server.  It seems like cheap insurance to me at negligible cost, but happy to defer to you on it.

crates/validation/src/materialization.rs
dgreer-dev: Yep makes sense, I'll get it in.

crates/validation/src/materialization.rs
dgreer-dev: I believe blank doesn't mean "off" here, it means the default 30m ack schedule
(defaultSyncFrequency in go/materialize/schedule_config.go; it kicks in once
the connector judges itself caught up and streaming). The only expressible
off state is an explicit syncFrequency: "0s", which is exactly the value the
conflict check used to reject. So without the carve-out there's no way to set
up runtime-only pacing on the ten connectors that use this boilerplate until
the connector-side deferral ships in their images.

You're right, the real fix is in the connectors not here.  Do we want to do that
instead?

crates/runtime-next/src/leader/materialize/actor.rs
dgreer-dev: If we cancel the previous ones I believe that will hang up the connection on them (and so they won't be told when it's done)

I don't think this can really grow all that large, it's one entry per live connection, and we prune it every 15-ish seconds.  If we're getting like, a million connections something else will exhaust before this list.

dgreer-dev:
Pushed a commit with the four resolved convo fixes.

- Dropped `--progress`, `--no-wait`, `--timeout`. flowctl now prints nothing and exits 0; the exit status is the whole contract.
- `SyncNowResponse` is down to three empty messages: `Ack`, `Heartbeat`, `Done`. Status and the outcome enum are gone, and so are Done's committed stats (same argument as Status).
- "Nothing to await" needs no label of its own: the leader acks and Done follows immediately, which is what IDLE and NOT_APPLICABLE already did.
- Dropped the validation error on `syncSchedule` without `enable-runtime-v2`.

crates/validation/src/materialization.rs
williamhbaker: Ah yes, I hadn't thought about the absent defaulting to 30m behavior that a lot of materializations have.

That does make it harder for a future deterministic migration strategy. Sometimes a materialization not having a schedule configured means it should actually be 30 minutes, and sometimes it means truly absent. Not all materialization connectors support a connector-side sync frequency configuration at all. 

It's possible that a 2 phase migration approach could be used, where the implicit 30-minute-if-absent connectors first make the configurations explicit, and then we can deterministically migrate them. I'm actually thinking it would be easier to just bake this into whatever migration process we use, where we could hard-code the list of connectors where absent = 30 minute default.

This also raises another question: We probably want that 30 minute default going forward, for the same reason it is in the connector side...we don't want users to not fill anything in there and then end up with really high warehouse bills because the materialization was running full-tilt. So I think that's something we need to think about separately.

dgreer-dev: For this I'm going to just remove the check so that you can have a runtime AND connector configs.  (Ideally connectors stop doing that)

go/runtime/task_control.go
```diff
@@ -0,0 +1,187 @@
...
+		return status.Error(codes.InvalidArgument, "task_name is required")
+	}
+	var shardZero, found = taskShardZero(tc.service.State, req.TaskName)
+	if !found {
+		return errTaskNotFound(req.TaskName)
+	}
+
+	// MayProxy because we never serve locally: the sidecar does, whether or not
+	// we're the primary.
+	var res, err = tc.service.Resolver.Resolve(consumer.ResolveArgs{
```
williamhbaker: This is kind of clunky for the routing we need...`Resolve` in particular does things we don't need here, like waiting for the shard store to finish recovery, the `MayProxy` workaround, and the `res.Done` handling. The one thing we do need from it is `res.Store` once we know we're the primary, but that decision itself
is pure keyspace state. I think a better model to follow would be in `buildLeaderfulJoin`.

A little more concretely, `taskShardZero` already locks the keyspace with `state.KS.Mu.RLock()`. Do all of the work inside that one lock:

1. Check the caller's claims against the spec's labels (previously `Resolve` was doing the claims check)
2. Call `primaryAssignment(state, shardZero)`. If not found, return Unavailable.
3. Compare it to the local member's ProcessSpec_ID.
4. If the primary is a peer, call `memberEndpoint(state, ...)` and dial that front door. We no longer need `MayProxy` or `ProxyHeader` here. That also lets `taskControlProxyMD`, `attachProxyHeader`, and `getProxyHeader` go away.
5. If the primary is us, call Resolve with MayProxy: false. We only want the `Store`.

crates/flowctl/src/raw/sync_now.rs
```diff
@@ -0,0 +1,546 @@
...
+        };
+    }
+}
+
+/// Build the HTTP client which calls the reactor front door.
+///
+/// `reqwest` verifies against webpki's bundled roots, and -- unlike tonic's
+/// `tls-native-roots` -- ignores `SSL_CERT_FILE`. Honor it explicitly, so that
+/// a data plane behind a private CA (such as a local stack) is reachable.
+fn new_http_client() -> anyhow::Result<reqwest::Client> {
```
williamhbaker: Could we just re-use `ctx.rest.http_client`? I get the thing about the local stack, but I think there's other workarounds for that. It's rather inconsistent to have this single command build its own special HTTP client because of the local stack thing, so I'd either follow the convention of it not working out of the box with a local stack (and I said, I'm pretty sure there's another workaround...our buddy Claude would certainly know), or fix it at a lower level. And for this I don't think I'd suggest the lower level fix just yet.

williamhbaker (commented):
Looking pretty good with the updates. I had a couple of pending comments that I think are still relevant, which will hopefully show up below this...

go/runtime/task_control.go
dgreer-dev: yeah this is a good call, and simplifies a bunch.

crates/flowctl/src/raw/sync_now.rs
dgreer-dev: yep that did sort of spiral out of control.  removed.

go/runtime/task_control.go
```diff
@@ -0,0 +1,186 @@
...
+// shards before deleting the old, so both can briefly appear here and we take
+// the older. Sync-now during a reset is meaningless either way: the task is
+// being backfilled from scratch.
+func taskShardZero(state *allocator.State, claims pb.Claims, taskName string) (*pc.ShardSpec, pb.Route, error) {
+	state.KS.Mu.RLock()
+	defer state.KS.Mu.RUnlock()
+
+	var spec *pc.ShardSpec
+	for _, taskType := range []string{"capture", "derivation", "materialize"} {
+		var prefix = allocator.ItemKey(state.KS, taskType+"/"+taskName+"/")
```
williamhbaker: `allocator.ItemKey` has an assertion that panics on certain characters...notably a space character, so if the API was called with a malformed tenant name containing a space, it could panic.

I think a safer alternative would be this: `var prefix = state.KS.Root + allocator.ItemsPrefix + taskType + "/" + taskName + "/"`

Alternatively, some sort of validation at the top of this function to ensure the provided task name is valid.

williamhbaker (commented):
Final comment!

go/runtime/task_control.go
dgreer-dev: yep, good call.  adding a fix + test

crates/flowctl/src/raw/sync_now.rs
williamhbaker: > but I think there's other workarounds for that

Looking at this again I think I'm probably wrong about this, and we'd need some kind of code change to make it work on a local stack. Fine for a follow-up though I think.

williamhbaker (dismissed):
LGTM

williamhbaker (approved):

PR #3330, opened 2026-08-09
runtime-next: build per-collection state once per Target / Source

jgraettinger:
Tasks routinely fan many bindings onto few collections: a capture of many database tables or topic streams writing one collection, or a materialization maintaining several views of one source. Everything the runtime derives from a collection — parsed and indexed schema validators, publisher targets (a journal client and partitions watch each), shuffle read state, inferred write-shapes — was built once *per binding*, so these tasks paid that cost many times over.

This stack makes bindings *reference* a per-collection struct which owns the derived state, built once:

- `doc::combine::Spec` gains a binding → validator indirection, so bindings of one collection share a validator while still combining per-binding.
- The publisher's `Binding` becomes `Target`, with an explicit task-binding → target mapping.
- Shuffle, derive, and materialize group bindings onto shared `Source`s; the capture leader groups them onto `Target`s.

Two semantic consequences ride along, both on the capture side: schema inference becomes per-collection (every binding widens its collection's one shape, instead of each inferring independently and blind to the others), and a `BackfillBegin` of a binding whose peers also write the same collection suppresses its truncation markers — a TRUNCATE of one source table among many must not truncate the shared collection.

The commit messages carry the fine-grain details of each layer.

### DIFF
--- a/crates/runtime-next/README.md
+++ b/crates/runtime-next/README.md
@@ -74,37 +74,67 @@ src/
 ├── leader/            # sidecar Leader service
 │   ├── service.rs       # gRPC entry, per-task Join rendezvous
 │   ├── join.rs          # protocol primitives for joining shards into a session
+│   ├── close_policy.rs  # when a transaction closes (min / max txn duration)
+│   ├── frontier_mapping.rs  # consumer.Checkpoint <-> shuffle::Frontier
 │   ├── shuffle.rs       # ShuffleSession / ShuffleSessionFactory traits + ShuffleServiceFactory
 │   │                     #   (journal-reading Session) impl; leader is monomorphized over the
 │   │                     #   factory (preview installs its own fixture replay from flowctl)
+│   ├── capture/
+│   │   ├── fsm.rs           # head/tail state machines for capture transactions
+│   │   └── task.rs          # Task + Binding + Target: the leader's data model
+│   ├── derive/
+│   │   ├── handler.rs       # gRPC stream handler, dispatches to startup/actor
+│   │   ├── startup.rs       # Recover / Open / Apply / Recovered phase
+│   │   ├── fsm.rs           # pipelined HeadFSM / TailFSM state machines
+│   │   ├── actor.rs         # event loop driving open / commit / acknowledge
+│   │   └── task.rs          # Task: the leader's data model
 │   └── materialize/
 │       ├── handler.rs       # gRPC stream handler, dispatches to startup/actor
 │       ├── startup.rs       # Recover / Open / Apply / Recovered phase
 │       ├── fsm.rs           # pipelined HeadFSM / TailFSM state machines
 │       ├── actor.rs         # event loop driving open / commit / acknowledge / trigger
-│       ├── frontier_mapping.rs  # consumer.Checkpoint <-> shuffle::Frontier
 │       ├── triggers.rs      # webhook trigger delivery
 │       ├── sync_schedule.rs # compiled sync-schedule evaluator (commit pacing)
-│       └── task.rs          # per-task state held by the leader actor
+│       └── task.rs          # Task: the leader's data model
 │
 └── shard/             # per-shard controller-facing service
     ├── service.rs       # gRPC entry, dispatches by task type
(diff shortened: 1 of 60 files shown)


jgraettinger:
## QA: local stack + soak test

Ran `tests/soak` end to end, then ramped fan-in bindings. **Headline: 30,000 capture fan-in bindings cost a very acceptable amount of memory.**

**Baseline:** full chain published clean — the documented inference-widening crash-loop, then recovery; all six at-rest invariants from `tests/soak/README.md` return 0/empty.

### Reactor / sidecar, ramping the capture to 30k bindings over 3 collections

| capture bindings | reactor RSS | sidecar RSS |
|---|---|---|
| 3 (baseline) | 371 MB | 53 MB |
| 3,000 | 446 MB | 55 MB |
| 30,000 | 699 MB | 56 MB |

Roughly **9.5 kB of reactor RSS per binding**, consistent across both the 300→3,000 and 3,000→30,000 steps. The sidecar stays flat, as it should — captures never join the shuffle leader.

These are directional: single-process aggregates hosting every shard/leader on the box, with ordinary drift between readings, and branch-only (no master arm on the stack).

### Fan-in vs fan-out, via `preview-next`

`flowctl raw preview-next` isolates one task's runtime in-process, so it does support an A/B against merge-base `b7eb642`. Capture fan-in peak RSS: 1,957 → 965 MB at 30,000 bindings, and 4,440 → 1,349 MB at the 65,436 ceiling — marginal cost **67 kB → 20 kB per binding**, improving with N as the fixed per-collection cost amortizes.

Fan-out (N bindings → N collections) is the control and barely moves — 850 → 727 MB at N=3,000 — which is right, since there's nothing to share. The within-branch spread says it best: at N=3,000 fan-in is 146 MB against fan-out's 727 MB, a **5× spread** that on the merge-base is only 3.3×. The spec side agrees independently: a live 30,000-binding capture builds to a 2,988 kB spec with 6 `linkedCollections`.

### Captures scale much further per binding than derive/materialize

Validator sharing is most of a capture binding; for derive/materialize it's rounding error against per-binding read and combine buffers, since each binding reads the shared journal independently. Taking `ledger` 2 → 150 bindings moved the **sidecar** 56 MB → 963 MB, ~6 MB/binding in the shuffle Session. So captures reach the 65,436 binding limit inside ~1.4 GB, while derive/materialize run out on a 15 GB box somewhere around 600–1,000.

### Correctness at scale

Ramped the live capture 3 → 30 → 300 → 3,000 → 30,000 fan-in bindings, verifying at each step (counts kept multiples of 3 so `id % N` never migrates an account across collections). Running ~3.5h at 30,000 capture plus 150 materialize fan-in bindings: zero ERRORs on all four tasks, all six invariants 0, 232k delta rows.

**Soak harness note:** adding materialize bindings emits a burst of real-looking `conservation violated` / `seq regression at sink` ERRORs that clear once they catch up — `materialize_soak` keys `self.standard` by `id` alone, so bindings backfilling at different offsets collide in one map. Not a runtime defect, but that store wants keying by `(binding, id)` if the soak should cover materialize fan-in.

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code)</sub>

crates/runtime-next/README.md
```diff
@@ -208,6 +238,51 @@ journal reads once they read through hinted spans. This prevents over-read
...
+
+The two key differently:
+
+- **Capture `Target`s key on `partition_template_name`** — journal identity — in
+  both spec forms. `Task::new` requires that bindings sharing a name also carry
+  equal `CollectionSpec` values.
+- **Materialize / derive / shuffle `Source`s key on the declared
+  `collection_index`** — value identity — never on collection name, because a
+  materialization's `group_by` can give two bindings of one named collection
+  differing read schemas. An inline-form binding carries no index and is its
```
williamhbaker, line 258: nit: I was a little confused by this line about "differing read schemas", but I believe it is referring to "differing key and projections", rather than the actual collection schema.

williamhbaker (approved):
LGTM

crates/runtime-next/README.md
jgraettinger: confirming your understanding, and yes this is poorly worded

PR #3332, opened 2026-08-10
docs: source-commercetools

Alex-Bair:
**Description:**

Documentation updates for https://github.com/estuary/connectors/pull/5032.

### DIFF
--- a/site/docs/reference/Connectors/capture-connectors/commercetools.md
+++ b/site/docs/reference/Connectors/capture-connectors/commercetools.md
@@ -1,0 +1,84 @@
+---
+description: Use the commercetools connector to sync orders, customers, and payments, using OAuth2 client credentials from a commercetools API Client.
+---
+
+# commercetools
+
+This connector captures data from [commercetools](https://docs.commercetools.com/api) into Estuary collections.
+
+## Supported data resources
+
+The following data resources are supported:
+
+| Resource | Replication Mode |
+|----------|------------------|
+| [orders](https://docs.commercetools.com/api/projects/orders) | Incremental |
+| [customers](https://docs.commercetools.com/api/projects/customers) | Incremental |
+| [payments](https://docs.commercetools.com/api/projects/payments) | Incremental |
+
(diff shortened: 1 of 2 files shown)


site/docs/reference/Connectors/capture-connectors/commercetools.md
```diff
@@ -0,0 +1,80 @@
...
+
+By default, each resource is mapped to an Estuary collection through a separate binding.
+
+
+## Prerequisites
+
+To set up the commercetools source connector, you'll need:
+
+* Your commercetools **Project key**, and the **region** hosting that Project. Both are visible in the Merchant Center under **Settings > Developer settings**, in the API URL shown for your Project.
+* A commercetools [API Client](https://docs.commercetools.com/api/authorization) with the `view_orders`, `view_customers`, and `view_payments` [scopes](https://docs.commercetools.com/api/scopes), created in the Merchant Center under **Settings > Developer settings > Create new API client**. Note the client ID and secret when the client is created.
```
aeluce:
```suggestion
* A commercetools [API Client](https://docs.commercetools.com/api/authorization) created in the Merchant Center under **Settings > Developer settings > Create new API client**, with the following [scopes](https://docs.commercetools.com/api/scopes):
   * `view_orders`
   * `view_customers`
   * `view_payments`

   Note the client ID and secret when the client is created.
```

This may be a lot of instruction condensed into one bullet point. Especially if we think we might add more resources and scopes later, we could pull out the scopes into its own sublist.

aeluce (dismissed):
Minor suggestion; not a blocker if you want to merge. Thanks!

site/docs/reference/Connectors/capture-connectors/commercetools.md
Alex-Bair: Good point, I've made that edit. I do think it's likely we'll add more resources & scopes later once this connector starts getting used, and having that sublist will definitely make it easier to read & add more scopes later. Thanks!!

aeluce (approved):

PR #3333, opened 2026-08-10
crates/models: resolve capture interval from the connector tag

nicolaslazo:
**Description:**

`connector_tags.default_capture_interval` has existed for a long time but nothing ever applied it.

`interval` becomes `Option<Duration>`, with unset meaning "resolve at build time". If the model does not provide its own `interval`, then the connector tag's takes precedence over the global fallback.

Build-time resolution costs propagation latency: a changed tag default reaches an existing capture only when that capture is next built. That happens automatically within twenty days via the periodic touch publication, and sooner if anything else republishes the task.

**Workflow steps:**

**Notes for reviewers:**

Tested locally using unit tests.

### DIFF
--- a/.sqlx/query-08e7942349b6f97210391b4fd3fdac705ddf26f92846737add3288ebaf71b0c1.json
+++ b/.sqlx/query-08e7942349b6f97210391b4fd3fdac705ddf26f92846737add3288ebaf71b0c1.json
@@ -1,0 +1,30 @@
+{
+  "db_name": "PostgreSQL",
+  "query": "\n        select\n            i.image as \"image!: String\",\n            ct.default_capture_interval as \"default_capture_interval: crate::Interval\"\n        from unnest($1::text[], $2::text[], $3::text[]) as i(image, image_name, image_tag)\n        join connectors c on c.image_name = i.image_name\n        join connector_tags ct on c.id = ct.connector_id and ct.image_tag = i.image_tag\n        ",
+  "describe": {
+    "columns": [
+      {
+        "ordinal": 0,
+        "name": "image!: String",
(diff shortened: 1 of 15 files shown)


jgraettinger:
What about resolving this from the Spec response, Open response, or an image label? For example we change validation semantics so that "unset" propagates down into the built spec, and then have the reactor choose the restart interval through a negotiation with the started connector.

It seems circuitous to have `connector_tags` control this, given it's a property of the connector, and I have uncertainty about how fit-for-purpose `connector_tags` will remain going forward.

jwhartley:
doc files no longer do.

at build time.

Suggested split:

nicolaslazo:
> What about resolving this from the Spec response, Open response, or an image label? For example we change validation semantics so that "unset" propagates down into the built spec, and then have the reactor choose the restart interval through a negotiation with the started connector.
> 
> It seems circuitous to have `connector_tags` control this, given it's a property of the connector, and I have uncertainty about how fit-for-purpose `connector_tags` will remain going forward.

Makes sense. I'd propose an expansion to `Open` responses as the proper long-term solution; but that entails a codebase-wide sweep to make every connector adopt the new protocol — on top of the risk of introducing protocol version mismatches into the pipeline. This is something I'd defer until we're deploying a larger feature that changes the wire format and the team overhead of monitoring is better justified.

Image labels are a viable alternative in the meantime. They are a stringly-typed format, so in the eyes of the control plane it is indifferent whether the label is unset, left deliberately empty, or skipped altogether — all cases fall back to the already defined 5-minute global timeout defined in the control plane. This flexibility will make them trivial to sunset, too.

PR #3334, opened 2026-08-10
runtime: remove shard recovery directory on Destroy

williamhbaker:
**Description:**

Gazette frees a shard's local state only through Store.Destroy. Flow's Stores did not remove the recovery-log playback directory. Thus each shard teardown left a directory on disk. Captures and materializations never removed it. Derivations removed it only for the SQLITE connector type.

Add taskBase.removeRecoveryDir. Call it from all six Destroy implementations, after the task service drops.

**Workflow steps:**

**Notes for reviewers:**

### DIFF
--- a/go/runtime/capture.go
+++ b/go/runtime/capture.go
@@ -406,6 +406,7 @@ func (c *captureApp) Destroy() {
 		_ = c.client.CloseSend()
 	}
 	c.taskBase.drop()
+	c.taskBase.removeRecoveryDir()
 	c.taskBase.opsCancel()
 }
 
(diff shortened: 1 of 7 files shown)


jgraettinger (approved):
LGTM

Gazette's RocksDB store clears it [here](https://github.com/gazette/core/blob/e54beb5c6e6452a94fe0c0bbe21a1695fa9fddd6/consumer/store-rocksdb/store_rocksdb.go#L187), but I gather we're not hitting that because we're instead wiring the rocksdb env and then handing that to Rust, so we're responsible for doing our own tear-down here.

PR #3339, opened 2026-08-11
runtime-next: absorb a shard Stop which outruns the session it addressed

### LINKED ISSUE #3289: runtime-v2: a publish can strand a shard's controller stream, leaving it FAILED until unassigned
A publish to **any runtime-v2 task with more than one shard** can leave some of
its shards permanently FAILED, flapping on a ~100s cycle. The destination stops
advancing, and whatever the publish asked for — including an Apply's DDL or a
backfill — never runs. `flowctl-go shards unassign` fixes it every time, within
~10s; nothing else does.

This is a runtime race. It is not specific to Snowflake, or to materializations,
and the connector is not involved.

## Cause

The Go shard controller and the Rust runtime share one long-lived gRPC stream.
Between sessions, Rust sits waiting for the next `Join`.

A publish ends the Rust session and cancels the Go controller's task term at the
same instant, so two messages cross in flight:

- Rust finishes its session, sends `Stopped`, and goes back to awaiting a `Join`.
- Go, seeing its term cancelled, sends `Stop` — asking Rust to end the session
  Rust has already ended on its own.

Rust receives a `Stop` when it will only accept a `Join`, treats that as a
protocol violation, and kills the stream:

```
runTransactions: readMessage: receiving Joined:
  Materialize protocol error (expected Join) from controller: {"stop":{}}
```

Go never notices: it reads the already-queued `Stopped` and returns a *clean*
shutdown, so nothing is logged for the poisoning. The stream is created once per
replica (`NewStore`) and reused by every later session — so every session after
this one fails its `Join`. That is exactly why `unassign` is the only remedy: it
destroys the replica, and the replacement opens a fresh stream.

Which of the two messages wins is Go's uniform-random `select` over two
simultaneously-ready cases — a coin flip per shard per publish, which is what the
observed 3-of-8, 1-of-8, 2-of-8, 1-of-8 incidence across four publishes looks
like.

## Fix

Absorb a `Stop` that arrives while awaiting a `Join`: the session it addressed is
already over, so the `Stop` is already satisfied, and its `Stopped` is already on
its way to the controller. One match arm in each of the three session loops
(materialize, derive, capture).

Go cannot fix this on its own — at the instant the controller decides to send,
whether the Rust session has ended is not knowable, so any check there is TOCTOU.

## Status

Ready on branch `jgm-runtime-v2-late-stop`: a regression test reproducing the
error string above through `Service::spawn_materialize`, then the fix. The
`runtime-next` suite passes. Not yet pushed.

## Done when

A publish activates without an operator unassigning shards, and an Apply or
backfill the publish carries takes effect on its own.

---

<details>
<summary>Two corrections to the original report, kept for the record</summary>

The report was filed against `materialize-snowflake` on the premise that **no
connector container started at all**, and that the shard error was `awaiting Etcd
revision <N>: context canceled`. Both were wrong, and correcting them is what led
to the cause.

In the observed window the task logged 54 `started connector container`, 54
`opening Snowflake`, and 328 completed commit cycles — the absence of connector
logs that motivated "this looks like the runtime" was not real. And all seven
`shardFailure` events carried one identical message, the `expected Join` error
above, at times matching the three publishes exactly.

Separately, `f4009a5ac3` (two-session idempotent recovery) has since made a shard
parked awaiting a `Join` a routine state rather than a publish-only one, so the
crossing no longer requires a publish to coincide with a session end.

</details>

jacobmarble:
Fixes #3289.

**Description:**

A publish to any runtime-v2 task with more than one shard can strand a shard's
controller stream, leaving it FAILED — and its Apply unrun — until an operator
unassigns it.

The controller and the Rust runtime share one long-lived gRPC stream, created
once per replica in `NewStore`. A publish ends the Rust session and cancels the
controller's task term at the same instant, so `Stopped` (ours) and `Stop`
(theirs) cross in flight: the session loop, which will only accept a `Join`,
fails the whole stream over the `Stop`. The controller reads the queued `Stopped`
and returns a *clean* shutdown, so nothing is logged — and every later session on
that replica then fails its `Join`. Which message wins is Go's random `select`
over two ready cases, so it's a coin flip per shard per publish.

Fix: absorb a `Stop` received while awaiting a `Join`. The session it addressed is
already over, so the `Stop` is satisfied and its `Stopped` is already on the way.
One match arm in each of the three session loops (materialize, derive, capture).
The controller can't fix this itself — whether the Rust session has ended isn't
knowable at the instant it decides to send.

**Workflow steps:**

None. Publishing a runtime-v2 task stops requiring a `shards unassign` to take
effect.

None user-facing. `crates/runtime-next/README.md` gains the invariant.

**Notes for reviewers:**

- First commit is red, second green: the test drives `Service::spawn_materialize`
  and reproduces the production error string byte-for-byte
  (`Materialize protocol error (expected Join) from controller: {"stop":{}}`).
- The three loops are fixed identically; the rationale comment lives on the
  materialize arm, the other two point at it.
- `f4009a5ac3` (two-session idempotent recovery) widened this: a shard parked
  awaiting a `Join` is now routine, so the crossing no longer needs a publish to
  coincide with a session end.
- `runtime-next` lib suite: 199/199. `tests/split_e2e.rs` needs a built gazette
  binary and wasn't run locally.

### DIFF
--- a/crates/runtime-next/README.md
+++ b/crates/runtime-next/README.md
@@ -181,6 +181,17 @@ documented inline in the proto.
 - **`shard/rocksdb.rs` is the single Persist application path.** Capture
   reuses it by synthesizing `Persist` messages locally rather than receiving
   them from a leader.
+- **A `Stop` may outrun the session it addresses.** The controller sends
+  `Stop` when it observes its task term cancelled, and it selects over that
+  cancellation and our `Stopped` concurrently — so the two cross on the wire
+  whenever a session ends at the same moment the term does. A publish does
+  exactly that to every shard at once, and an idempotent-replay session that
+  stops itself can coincide with one. Each session loop therefore absorbs a
+  `Stop` received while awaiting a `Join`: the session it addressed is over,
+  and its `Stopped` is already on its way to the controller.
+  Failing the stream instead strands the shard, since the stream is created
+  once per replica (`NewStore`) and reused by every later session — each of
+  which would then fail its `Join`, recoverable only by unassigning.
 
 ## Coexistence with `runtime`
 
--- a/crates/runtime-next/src/publish.rs
+++ b/crates/runtime-next/src/publish.rs
@@ -429,6 +429,30 @@ pub(crate) struct RecordingPublisher {
     pub stats: std::sync::Arc<std::sync::Mutex<Vec<ops::proto::Stats>>>,
 }
 
+/// Test [`PublisherFactory`] opening [`RecordingPublisher`]s, for tests which
+/// need to stand up a whole `Service` rather than an actor. Each `open` yields
+/// a fresh publisher, so this is for tests that never inspect what was
+/// published — those build a `RecordingPublisher` directly.
+#[cfg(test)]
+#[derive(Clone)]
+pub(crate) struct RecordingPublisherFactory;
+
+#[cfg(test)]
+impl PublisherFactory for RecordingPublisherFactory {
+    type Publisher = RecordingPublisher;
+
+    fn open(
+        &self,
+        _authz_subject: String,
+        _producer: uuid::Producer,
+        _stats_journal: &str,
+        _collection_specs: &[&proto_flow::flow::CollectionSpec],
+        _binding_targets: &[u32],
+    ) -> anyhow::Result<RecordingPublisher> {
+        Ok(RecordingPublisher::default())
+    }
+}
+
 #[cfg(test)]
 impl RecordingPublisher {
     pub fn take_docs(&self) -> Vec<(usize, String)> {
--- a/crates/runtime-next/src/shard/capture/handler.rs
+++ b/crates/runtime-next/src/shard/capture/handler.rs
@@ -169,6 +169,12 @@ where
             proto::Capture {
                 join: Some(join), ..
             } => join,
+
+            // A Stop addressed to a session which has already returned here;
+            // see the materialize session loop for how the controller comes to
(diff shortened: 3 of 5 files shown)


williamhbaker (approved):
LGTM

The issue as I'm understand it is that for multi-shard tasks, each shard has its own term, but there is one single session. A publication ends the term for each shard concurrently, which results in racing `stop` messages into the session, where the session will be ended by the first one that hits it and then others will arrive with no active session.

The conflict between the term being shard level and the session being task level is a bit unsatisfying but I'm not coming up with anything obviously better than what's proposed here.

jacobmarble:
> The issue as I'm understand it is that for multi-shard tasks, each shard has its own term, but there is one single session. A publication ends the term for each shard concurrently, which results in racing `stop` messages into the session, where the session will be ended by the first one that hits it and then others will arrive with no active session.
> 
> The conflict between the term being shard level and the session being task level is a bit unsatisfying but I'm not coming up with anything obviously better than what's proposed here.

You're right that the first `Stop` ends the session for every shard. But the late
Stops don't come back to that shared session — each shard has its own session
loop, and the late `Stop` lands there, in the gap where it's waiting for a `Join`.
Hence one arm per shard-side loop rather than something in the leader.

It also fits the numbers: racing into one session would fail 7 of 8 shards per
publish, and we saw 1–3. The coin flip is in each shard's own Go controller,
picking between sending `Stop` and exiting cleanly on the `Stopped` already
queued.

Issue #2781, opened 2026-03-17
Update validation/publication to use snapshots


jshearer:
Publications are doing work in the database (mainly authz?) that could be done significantly faster "clientside" using snapshots. 

The task is to figure out what exactly that looks like, and implement it.

What we need to do at a high level is switch from using the live in database query to using the snapshot in memory query that leverages the in graph for computations.

bbartman:
Made some progress on this after talking with @jgraettinger about the expected behavior. I'm working towards using the strangulation pattern to remove a piece at a time. I'm The first PR was to remove and replace references inside of discovery by using the snapshot instead of `internal.user_roles`.

I will update as I complete additional refactorings.

PR #3341, opened 2026-08-11
control-plane: Snapshot authorization model and retryable staleness error (1/7)

bbartman:
## What

Split 1 of 4 of #3155 (issue #2781): the foundation layer for moving catalog
authorization out of SQL and onto the in-memory authorization `Snapshot`.

- **`validation`**: adds `Error::AuthorizationSnapshotStale { catalog_name }` —
  the retryable signal that an authorization denial was evaluated against a
  Snapshot which may not yet reflect a committed grant — and
  `is_authz_snapshot_stale()` to classify it after `anyhow` wrapping.
- **`control-plane-api` Snapshot authorization model** (`server/snapshot.rs`):
  - `Authorization` enum (`Authorized` / `Denied` / `Stale`) with
    `ok_or_stale()`, which converts a provisional denial into the retryable
    error above.
  - `Snapshot::resolve_authorization()`, `user_authorization()`,
    `role_authorization()` — grant evaluation against the pinned Snapshot,
    classified by whether the Snapshot postdates the operation's freshness
    anchor (`taken_after`, allowing `TEMPORAL_SKEW`).
  - `Snapshot::spec_capabilities()` — in-process replacement for the
    SQL-computed `spec_capabilities` column, used to render "Available grants
    are:" error messages.
  - `STALE_RETRY_WAKE`, the re-poll interval used by later splits.
- **Rename** `App::snapshot` → `App::snapshot_watch`, ahead of follow-ups
  where executors pin per-poll `Snapshot` tokens distinct from the watch.

This PR is purely additive: nothing consumes the new surface yet.

## Tests

Unit tests in `server/snapshot.rs` (no DB required):
- `test_resolve_authorization` — the three-way classification, including the
  temporal-skew boundary.
- `test_spec_capabilities` — insta snapshots of capability rendering against
  `Snapshot::build_fixture`.
- `test_taken_after_allows_for_temporal_skew`.

## Stack

1. **→ this PR** — Snapshot authorization model + retryable staleness error
2. #3342 — move authorization in-process against a pinned Snapshot
3. #3343 — publications executor defers on stale snapshots
4. #3344 — discovers executor authorizes via Snapshot and defers on staleness

### DIFF
--- a/crates/control-plane-api/src/server/snapshot.rs
+++ b/crates/control-plane-api/src/server/snapshot.rs
@@ -73,6 +73,37 @@ pub struct SnapshotTask {
     pub data_plane_id: models::Id,
 }
 
+/// Outcome of an authorization check evaluated against a Snapshot,
+/// classified by `Snapshot::resolve_authorization`.
+#[must_use]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum Authorization {
+    /// The required grant exists in the Snapshot.
+    Authorized,
+    /// The grant is absent and the Snapshot is authoritative for the
+    /// operation's anchor: the denial is final.
+    Denied,
+    /// The grant is absent but the Snapshot predates the anchor: a grant
+    /// committed before the anchor may not be reflected yet, so the denial is
+    /// provisional and the operation should retry under a fresher Snapshot.
+    Stale,
+}
+
+impl Authorization {
+    /// Collapse to "is authorized?", surfacing a provisional denial as the
+    /// retryable `AuthorizationSnapshotStale` error which callers
+    /// (see `validation::is_authz_snapshot_stale`) convert into a retry.
+    pub fn ok_or_stale(self, catalog_name: &str) -> Result<bool, validation::Error> {
+        match self {
+            Authorization::Authorized => Ok(true),
+            Authorization::Denied => Ok(false),
+            Authorization::Stale => Err(validation::Error::AuthorizationSnapshotStale {
+                catalog_name: catalog_name.to_string(),
+            }),
+        }
+    }
+}
+
 // SnapshotMigration is the state of an underway data-plane migration.
 #[derive(Clone, Debug, serde::Serialize, serde::Deserialize)]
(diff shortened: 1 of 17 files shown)


bbartman:
Note that this combines what was previously two PRs https://github.com/estuary/flow/pull/3155 and https://github.com/estuary/flow/pull/3300

crates/control-plane-api/src/server/snapshot.rs
```diff
@@ -73,6 +73,37 @@ pub struct SnapshotTask {
     pub data_plane_id: models::Id,
 }
 
+/// Outcome of an authorization check evaluated against a Snapshot,
+/// classified by `Snapshot::resolve_authorization`.
+#[must_use]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum Authorization {
```
jgraettinger: What's the intended relationship between this type and the existing crate::AuthZResult ?

jwhartley:
doc files no longer do.

at build time.

Suggested split:

  - `site/docs/features/configuring-task-shards.md` becomes `content/features/configuring-task-shards.md`
  - `site/docs/getting-started/installation.md` becomes `content/getting-started/installation.md`
  - `site/docs/guides/customize-dataflows/materialization-sync-schedule.md` becomes `content/guides/customize-dataflows/materialization-sync-schedule.md`

Two notes on the non-doc `site/` files in this PR, because neither ports
cleanly:

- The `onBrokenAnchors: 'throw'` flip is a real improvement and it is still
  wanted. `estuary/docs` carries the identical `onBrokenAnchors: 'warn'` line
  and the same `TODO(johnny)` comment, so the anchor-checking gap you closed
  here is still open on the published site. Worth reopening against
  `estuary/docs` `docusaurus.config.js`. Expect it to surface real broken
  anchors on first run.
- The `site/README.md` macOS note is already solved differently in
  `estuary/docs`: its `docusaurus.config.js` detects Darwin and filters the
  case-colliding `hubspot-real-time` redirect out at build time, so the local
  build does not fail there. That note does not need to move.

PR #3342, opened 2026-08-11
control-plane, agent: move catalog authorization in-process against a pinned Snapshot (2/7)

bbartman:
## What

Split 2 of 4 of #3155 (issue #2781): every catalog authorization decision of
publications and discovers moves from SQL (`internal.user_roles()` /
computed `user_capability` + `spec_capabilities` columns) to in-process
evaluation against one `Snapshot` pinned per operation.

- **`live_specs`**: `fetch_live_specs` / `fetch_expanded_live_specs` lose
  their authorization parameters and computed-capability columns; the queries
  become plain spec fetches. `get_live_specs` / `get_connected_live_specs`
  gain a `&Snapshot` (and `started_at` anchor) and filter by
  `models::authz::CapabilitySet` in-process. A denial under a Snapshot that is
  stale for the operation's anchor raises the retryable
  `AuthorizationSnapshotStale` instead of silently dropping the spec;
  expansion (`get_connected_live_specs`) keeps denials as final silent
  omissions.
- **`publications`**: `resolve_live_specs` makes all four authorization
  decisions (drafter `SpecEdit`; spec read/write authorization to
  sources/targets; user `CatalogRead` on referenced specs) against the pinned
  Snapshot, with the same staleness policy. Data-plane-name authorization
  moves off `internal.user_roles()` onto `Snapshot::user_authorization`.
  `DraftPublication` gains `snapshot` and `started_at`, threaded through
  `Publisher::build` and the `Initialize` trait; `ExpandDraft`'s filter
  switches from legacy `admin` to `SpecEdit`
  (`filter_user_has_admin` → `filter_user_authz`).
- **`discovers`**: `Discover` pins `snapshot` / `started_at` across the
  whole operation (prechecks → connector RPC → merge); the live-capture
  fetch filters by `CatalogRead` via the Snapshot.
- **Executors and call sites** (`agent`): both executors pin one Snapshot per
  poll and thread it through; `main.rs` wires `snapshot_watch` into them.
  Controller/system paths (`controlplane.rs`, `create_data_plane`,
  `update_l2_reporting`) pass `started_at: None`, anchoring staleness to each
  spec's own `last_pub_id` — preserving today's semantics of evaluating
  against current-enough state.
- **Test harness**: the authorization Snapshot is now backed by a
  manually-driven watch with explicit `refresh_snapshot()` points (documented
  as the "Snapshot testing model"), and `truncate_tables` clears
  `storage_mappings` / `data_planes` so tests are deterministic next to a
  running local stack.

Behavior note (interim, until splits 3–4): a denial under a genuinely stale
Snapshot surfaces as a terminal task failure carrying the "please retry"
error, rather than deferring automatically. Splits 3 and 4 teach the two
executors to reschedule instead.

## Tests

- `live_specs::tests` (new, `#[sqlx::test]` on the new `authz_specs`
  fixture): unfiltered fetches never stale; authorized specs included;
  authoritative denials dropped; stale denials retryable; connected-spec
  expansion filtering.
- `publications::specs::resolve_tests` (new): each of the four in-process
  authorization branches, a fully-authorized positive control, the
  `started_at: None` / zero-`last_pub_id` terminal-denial fallback, and
  spec-to-spec authorization under `verify_user_authz: false`.
- Existing integration suites run under the new threading (mechanical
  signature updates in `locking_retries`, `source_captures`,
  `unknown_connectors`).

## Stack

1. #3341 — Snapshot authorization model + retryable staleness error
2. **→ this PR** — move authorization in-process against a pinned Snapshot
3. #3343 — publications executor defers on stale snapshots
4. #3344 — discovers executor authorizes via Snapshot and defers on staleness


crates/agent/src/controlplane.rs
```diff
@@ -655,6 +659,9 @@ impl<C: DiscoverConnectors + MakeConnectors> ControlPlane for PGControlPlane<C>
             logs_token,
             data_plane: data_plane.clone(),
             created_at,
+            snapshot,
+            // `filter_user_authz` is false, so no staleness anchor is consulted.
```
jgraettinger: nit: this is what I was getting in the last PR around choices of language: "staleness anchor" isn't previously defined or ubiquitous terminology in the control plane, but once the model gets the bit in its mouth for a concept it feels needs a name, it will start stamping it out everywhere without a firmer hand.

crates/agent/src/controlplane.rs
```diff
@@ -666,14 +673,22 @@ impl<C: DiscoverConnectors + MakeConnectors> ControlPlane for PGControlPlane<C>
...
+        let snapshot = self.snapshot_watch.token();
         let publication = DraftPublication {
             user_id: self.system_user_id,
             logs_token,
             draft,
             detail,
             dry_run: false,
             default_data_plane_name: default_data_plane,
-            // skip authz checks for controller-initiated publications
+            // Controllers construct a fresh publication per poll, so they have
```
jgraettinger: Controller tasks are backed by the `automations`-crate, which does have a `wake_at` timestamp that can be surfaced and plumbed through. This seems worth doing?

crates/control-plane-api/src/discovers/mod.rs
```diff
@@ -41,6 +41,19 @@ pub struct Discover {
...
+    /// operation: preflight checks, the connector RPC window, and the
+    /// post-RPC merge all consult this same instance, so one discover
+    /// observes exactly one authorization view regardless of refreshes
+    /// landing mid-flight.
+    pub snapshot: &'a crate::Snapshot,
+    /// Time at which the discover was queued, when the caller has a durable
+    /// one. Anchors authorization staleness of the merge's target collections:
+    /// a denial from a Snapshot older than this instant is provisional —
+    /// authority committed before queuing may be missing from it — and
+    /// reschedules the discover rather than silently dropping the live
```
jgraettinger: I don't think this retry is required, so long as target collections have an `expectLastPubId` (IIRC they do?). Discovery can use it's current view of AuthZ, and in the extremely rare case where it races a snapshot refresh that then lets the publication go through, they're protected from accidental overwrite.

crates/agent/src/discovers.rs
```diff
@@ -216,6 +226,12 @@ impl<C: DiscoverConnectors> DiscoverExecutor<C> {
             image_composed,
             data_plane,
             pool,
+            &snapshot,
+            // `None` anchors authorization staleness to each spec's own
```
jgraettinger: I've read this a couple times, and still am not following its meaning.

Using the queued discover row's `created_at` makes sense, but I'm not understanding why now() isn't the second-best value to use.

crates/control-plane-api/src/live_specs/mod.rs
```diff
@@ -1,39 +1,43 @@
...
 };
+use std::ops::Deref;
+use uuid::Uuid;
 
 /// Fetches live specs, returning them as a `tables::LiveCatalog`. Optionally
 /// filters the specs based on user capability. If `filter_capability` is
 /// `None`, then no filtering will be done.
+///
+/// `started_at` anchors the staleness check to the given time (request-relative).
+/// When `None`, staleness is anchored to each spec's publication time (spec-relative).
```
jgraettinger: This is cute, but how's a caller to know which path should be preferred, and why?
As a general rule, it's best to have only one mechanism and keep it well oiled.

PR #3346, opened 2026-08-11
runtime-next: log an auto-split only when one is started

dgreer-dev:
start_due_split logged "starting automatic journal split" as soon as a
due journal resolved to a splittable target, before split_partition
checked the journal's width against the MIN_PARTITION_WIDTH floor. A
journal already at the floor is due until its first outcome lands, so
that path emitted a start message for a split which never happened.

Move the message into split_partition, past the width check and just
ahead of the List / Apply RPCs which do the work.

Raise the matching at-floor message from debug to info. It fires at
most once per journal per assignment (finish_split then ignores the
journal terminally), and it is the only record of why a hot journal
stopped being considered for splits.


williamhbaker (approved):
LGTM

PR #3356, opened 2026-08-13
runtime-next: bound status trailers so connector errors aren't masked by ResourceExhausted

### LINKED ISSUE #3353: runtime_next: connector terminal errors are masked by ResourceExhausted / h2 protocol errors, which also occur independently
## Summary

When a materialization connector exits from a terminal error, the shard failure surfaced
to the user is frequently a generic transport error rather than the connector's own
message. Separately, the same transport error appears on shard failures where no
connector-side error exists at all, so it is not purely a masking artifact — something is
genuinely exhausting an HTTP/2 resource.

Both were observed on `AltopEnergy/MarketData/materialize-clickhouse` on 2026-08-11, and
the masking materially slowed down diagnosing a data-loss bug
(estuary/connectors#5044): anyone starting from `flowctl catalog status` saw a transport
error and had to know to go read the connector's own logs to find the real cause.

## 1. The connector's terminal error is masked

Two messages, in this order:

```
warn  runtime_next::shard::materialize::actor
      Materialize error (expected leader message) from leader
      Caused by: status: 'Some resource has been exhausted', self: "h2 protocol error: http2 error"

error runTransactions: readMessage: rpc error: code = ResourceExhausted desc = h2 protocol error: http2 error
```

The controller alert surfaced only the second:

```
"Observed 6 recent task shard failures, the latest at 2026-08-11 19:56:22 UTC. Last failure:
 runTransactions: readMessage: rpc error: code = ResourceExhausted desc = h2 protocol error: http2 error"
```

The actual cause, present in the task's logs at 18:54:16, was:

```
commit failed: moving stage to target marketdata_dartpricesieso: refusing to commit store
table: it contains 9 staged rows but 300699 rows were stored; not moving partitions so
staged rows are preserved for retry
```

That message is the one an operator needs. It names the table, the discrepancy, and the
fact that data was withheld rather than committed — and it never reached the shard's
failure status.

**Requested:** when a connector reports a terminal error and the runtime subsequently hits
a transport error on the same stream, propagate the connector's error as the shard failure
cause. The transport error is a consequence of the connector's exit, so it should not win
the race to become the reported failure.

## 2. The same error also occurs independently

On the same task, shard failures at **19:18:35** and **19:56:22** carry this error with
**no** connector-side error at those times — the connector's guard last fired at 19:17:25.
So this is not only a masking artifact; some HTTP/2 resource is genuinely being exhausted.

The task has **220 bindings**, which makes an oversized message or a flow-control limit a
plausible suspect (a `Loaded`/`Flushed` response, or a `StartedCommit` state update,
scaling with binding count). I did not confirm that and am not guessing further.

**Requested:** determine what exhausts the h2 resource on a high-binding-count
materialization, and whether the limit should be raised, the messages chunked, or the
condition surfaced as something more actionable than `ResourceExhausted`.

## Reproduction context

- Task: `AltopEnergy/MarketData/materialize-clickhouse`, runtime v2, 220 bindings.
- Window: 2026-08-11 18:54–19:56 UTC. The task's logs for that window contain both
  patterns, and `flowctl catalog status` shows the alert text quoted above.
- The connector-side data-loss bug that was being masked is
  estuary/connectors#5044; it is fixed in estuary/connectors#5053, but the masking is
  independent of that fix and will affect the next connector failure the same way.

jacobmarble:
**Description:**

Fixes #3353. When a materialization connector died, shards often reported
`ResourceExhausted: h2 protocol error: http2 error` instead of the connector's
actual error — which for a recent data-loss bug meant an operator had to know to
go read the connector's own logs to find the real cause.

Here's the mechanism. A gRPC error status travels entirely in an HTTP/2 trailer
(a header block sent after the response body). hyper caps how big a header block
can be at 16 KiB, which lets h2 spread one across at most 5 "continuation"
frames. Go past that and h2 assumes it's under attack: it kills the whole TCP
connection with an `ENHANCE_YOUR_CALM` error, which tonic reports as
`ResourceExhausted`. Every stream on that connection dies, and the status text
that was in flight is lost.

`connector-init` — which runs inside the connector container and reports the
connector's exit — built that status with no size limit at all: the connector's
last stderr line as the message, plus the whole encoded log again as binary
metadata. A connector that prints a big diagnostic as it dies therefore destroys
the very message explaining why it died.

Three changes:

1. **Bound the status before it's sent** (`connector-init`). Truncate the final
   log message, drop oversized structured fields, and bound the metadata copy.
   Budget is 2 KiB raw, so even worst-case encoding fits one 16 KiB frame with no
   continuation frame at all.
2. **Stop the leader from spreading one shard's failure to all of them.** When a
   session fails, the leader used to forward the failed peer's status object
   verbatim to every shard — so a single shard's transport error became every
   shard's reported cause, and an oversized trailer got relayed onto more
   connections. It now sends a formatted `leader session failed: …` message,
   which names the peer that actually failed and carries no metadata.
3. **Prefer the connector's error over the leader's echo.** When a shard's
   connector fails, the leader's broadcast comes back to that same shard. If the
   leader's copy lands first, the shard now checks whether its connector already
   has an error waiting and reports that instead.

Also bounded the two remaining copies of `anyhow_to_status` (in `runtime` and
`shuffle`) that had the same gap. The bounding helpers moved to `proto-grpc` so
all four crates share one copy instead of three drifting ones.

**Workflow steps:**

No user-facing workflow change. Shard failures in `flowctl catalog status` and in
task alerts should now name the connector's own error rather than a transport
error.

None.

**Notes for reviewers:**

- **The `connector-init` half ships in an image.** It needs a
  `flow-connector-init` release plus a bump of the inline pin before it takes
  effect in production. The runtime-side changes here protect tasks still running
  older images, since they stop the oversized trailer from being relayed onward.
- Nothing in the repo reads the `last-log-bin` metadata — only connector-init's
  own test. I bounded it rather than deleting it, per discussion.
- `prefer_connector_error` discards a ready connector *response* on the error
  path. The session is failing either way; the only question is which error
  describes why.
- I found four adjacent issues while tracing this and deliberately left them
  alone: an asymmetric h2 window setting in `gazette` (connection window raised,
  stream window at default, which disables adaptive windowing); the full built
  spec crossing five gRPC hops inline at session start; several clients still at
  tonic's 4 MB decode default; and vendored gazette's `MaxGRPCRecvSize` sitting at
  its 4 MiB default. Happy to open issues for these.
- Not verified end-to-end against a live stack. Unit tests cover the bounding
  arithmetic, the truncation, an end-to-end pass through `bidi` with a large
  stderr blob, and the connector-vs-leader error preference.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6494,6 +6494,8 @@ dependencies = [
 name = "proto-grpc"
 version = "0.0.0"
 dependencies = [
+ "anyhow",
+ "bytes",
  "labels",
  "proto-build",
  "proto-flow",
--- a/crates/connector-init/src/rpc.rs
+++ b/crates/connector-init/src/rpc.rs
@@ -123,6 +123,12 @@ where
         if !status.success() {
             tracing::error!(%status, "connector failed");
 
+            // Bound before the Log reaches the wire: this Status is the sole
+            // carrier of the connector's terminal error, and an oversized
+            // trailer destroys it outright. See MAX_TERMINAL_LOG_LEN.
+            let mut last_log = last_log;
+            bound_log(&mut last_log);
+
             let mut status = Status::unknown(&last_log.message);
             status.metadata_mut().insert_bin(
                 "last-log-bin",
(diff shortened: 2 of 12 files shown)


williamhbaker:
tagging @dgreer-dev since he had some some work recently along these lines also

dgreer-dev (approved):
LGTM on the fix.  The ticket also wanted a root cause so that is still outstanding, but maybe not relevant to the PR.

PR #3358, opened 2026-08-13
connector-init: demote broken-pipe stdin writes to debug

### LINKED ISSUE #3354: connector_init: demote 'i/o error writing to connector stdin' when the connector has already exited
## Summary

`connector_init::rpc` logs `i/o error writing to connector stdin` at `warn` when the
runtime writes to a connector that has already exited from a terminal error. It is pure
downstream noise — the write fails *because* the connector is already gone — but it is
logged at the same level as real problems and competes with the causal error for
attention.

Observed in a `materialize-clickhouse` task on 2026-08-11: three
occurrences, one per connector exit, each immediately following the connector's own
terminal error. While diagnosing a data-loss bug (estuary/connectors#5044) these were
three of the handful of `warn`/`error` lines in the window, and none of them said anything
about what actually went wrong.

## Requested

Suppress the message, or demote it to `debug`, when the connector has already reported a
fatal error or its process has already exited. A failed write to a dead process's stdin is
expected in that situation and carries no diagnostic value.

If the write can also fail for reasons unrelated to the connector having exited, keep
`warn` for that case and distinguish the two — the point is that the expected case should
not look like a problem.

## Context

Filed out of estuary/connectors#5044 (Part 3b), which tracked the log noise encountered
while diagnosing the incident. Related: the more consequential problem in that window was
the connector's terminal error being replaced by a transport error in the shard's failure
status, filed separately.

jacobmarble:
**Description:**

`flow-connector-init` pipes runtime requests into the connector's stdin. When a connector dies, the next write fails with a broken pipe — there's no reader left — and we logged that at `warn`. It's noise: the write failed *because* the connector was already gone, and it sat in the task's ops logs beside the connector's real error at the same level.

Broken pipes now log at `debug`. Other i/o errors keep their `warn` and original wording. A connector is allowed to not read its stdin at all, so this was never treated as an error — the function's comment already said so; now the level agrees. The sibling message was demoted the same way in a1bb93bff8.

Closes #3354.

**Workflow steps:**

No user-facing change. One fewer misleading `warn` when reading ops logs for a failed task.

None.

**Notes for reviewers:**

The issue asked to key on "the connector has already exited", but that isn't known at the write site — the exit future hasn't resolved, which is why we're still in the loop. `BrokenPipe` means the reader is gone either way.

### DIFF
--- a/crates/connector-init/src/rpc.rs
+++ b/crates/connector-init/src/rpc.rs
@@ -153,18 +153,14 @@ where
 
             // No message is ready. Should we flush?
             _ = async {}, if !buffer.is_empty() => {
-                if let Err(error) = stdin.write_all(&buffer).await {
-                    tracing::warn!(%error, "i/o error writing to connector stdin");
-                }
+                write_stdin(&mut stdin, &buffer).await;
                 buffer.clear();
                 continue;
             }
         };
 
         let Some(Ok(message)) = message else {
-            if let Err(error) = stdin.write_all(&buffer).await {
-                tracing::warn!(%error, "i/o error writing to connector stdin");
-            }
+            write_stdin(&mut stdin, &buffer).await;
             if let Some(Err(error)) = message {
                 tracing::error!(%error, "failed to read next runtime request");
             }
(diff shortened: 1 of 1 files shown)


williamhbaker (approved):
LGTM

PR #3359, opened 2026-08-14
dekaf: keep connection open on document validation error

### LINKED ISSUE #3338: dekaf: incompatible field schema closes session until spec refresh
**Reproduction Steps**:

1. Begin reading a dekaf topic with schema inference.
2. Write a document: `{"name": "alice", "value": 42}`
3. Write a document with an incompatible field schema: `{"name": "bob", "value": "howdy"}`

**Expected**:

Read documents as written due to schema widening:
```
{"name": "alice", "value": 42}
{"name": "bob", "value": "howdy"}
```

**Actual**:

Disconnected:
```
%6|1786407316.978|FAIL|rdkafka#consumer-1| [thrd:sasl_plaintext://127.0.0.1:10250/1]: sasl_plaintext://127.0.0.1:10250/1: Disconnected (after 21858ms in state UP)
%6|1786407317.015|FAIL|rdkafka#consumer-1| [thrd:sasl_plaintext://127.0.0.1:10250/1]: sasl_plaintext://127.0.0.1:10250/1: Disconnected (after 12ms in state UP, 1 identical error(s) suppressed)
% ERROR: Local: Broker transport failure: sasl_plaintext://127.0.0.1:10250/1: Disconnected (after 21858ms in state UP)
% ERROR: Local: All broker connections are down: 1/1 brokers are down: terminating
```

Server logs:
```
2026-08-11T00:15:17.014501Z  WARN dekaf_session{session_client_id="rdkafka" task_name="dbn/test/dekaf-generic"}:serve{idle_timeout=60s stale_read_timeout=60s tls_handshake_timeout=10s addr=[::1]:50252}: dekaf: error=Extracting field Extractor(
    Extractor {
        ptr: Pointer(
            [
                Property(
                    "value",
                ),
            ],
        ),
        policy: SerPolicy {
            str_truncate_after: 18446744073709551615,
            array_truncate_after: 18446744073709551615,
            nested_obj_truncate_after: 18446744073709551615,
        },
        default: Null,
        magic: None,
    },
), schema: Long
Caused by:
    at , value "howdy" does not conform to AVRO schema "long"
```

Cannot read again until `spec_ttl` timeout refreshes the collection spec.

danielnelson:
**Description:**

Collection schemas are refreshed every spec_ttl.  Currently if a document does not validate against its collection schema the session would log an error and close the connection.  This would interrupt all partitions on this connection and the client would need to reconnect.  If the error is corrected by the inferred schema you it should be updated on the next spec_ttl refresh, but if the client reconnects before then it would result in further disconnections.

With this change, when a partition fails to validate against the schema, it is placed into "cooldown".  LEADER_NOT_AVAILABLE will be reported for this partition.  This will indicate to the client that it should retry, but will not affect the ability for other partitions to make progress.  The partition remains in cooldown until the collection schema is updated where on the next fetch a new Read is created, or until a timeout (4m default, 2x the spec_ttl) where the connection will be closed with an error like before.

closes: #3338

**Workflow steps:**

No changes

None

**Notes for reviewers:**

I've tested this with rdkafka/kcat and tinybird

### DIFF
--- a/crates/dekaf/src/lib.rs
+++ b/crates/dekaf/src/lib.rs
@@ -346,7 +346,7 @@ impl App {
 /// Dispatch a read request `frame` of the current session, writing its response into `out`.
 /// `raw_sasl_auth` is the state of SASL "raw" mode authentication,
 /// and conditions the interpretation of request frames.
-#[tracing::instrument(level = "trace", err(level = "warn"), skip_all)]
+#[tracing::instrument(level = "trace", skip_all)]
 pub async fn dispatch_request_frame(
     session: &mut Session,
     raw_sasl_auth: &mut bool,
--- a/crates/dekaf/src/session.rs
+++ b/crates/dekaf/src/session.rs
@@ -32,8 +32,6 @@ struct PendingRead {
     // Last time this read was included in a Fetch request. Used to reap reads for
     // partitions the client has stopped fetching.
     last_accessed: std::time::Instant,
-    // Identifies the journal, used to look for entries in cooldown.
-    journal_name: String,
     // The Collection::schema_hash this read was started against, used to
     // detect if the read is still current.
     schema_hash: String,
@@ -125,8 +123,8 @@ pub struct Session {
     read_buffer_size: usize,
     // Total byte budget for a Fetch, divided evenly across partitions to cap per-partition reads.
     combined_partition_fetch_limit: usize,
-    // Journals currently cooling down after a schema-validation error, keyed by journal name.
-    cooldown: HashMap<String, CooldownEntry>,
+    // Partitions currently cooling down after a schema-validation error,
+    cooldown: HashMap<(TopicName, i32), CooldownEntry>,
     // How long a journal can stay in cooldown before this connection gives up and closes.
     schema_error_hard_fail_after: std::time::Duration,
 }
@@ -645,17 +643,11 @@ impl Session {
             tracing::info!(reaped, remaining = self.reads.len(), "reaped stale reads");
         }
 
-        // Drop cooldown entries for journals no client is fetching anymore.
-        // A cooldown otherwise only clears when a partition is fetched again
-        // with an updated schema hash, so an abandoned partition's entry
-        // would persist for the life of the connection.
-        let journals_in_use: std::collections::HashSet<&str> = self
-            .reads
-            .values()
-            .map(|pending| pending.journal_name.as_str())
-            .collect();
-        self.cooldown
-            .retain(|journal_name, _| journals_in_use.contains(journal_name.as_str()));
+        // Drop cooldown entries for (topic, partition) pairs no client is
+        // fetching anymore. A cooldown otherwise only clears when a partition
+        // is fetched again with an updated schema hash, so an abandoned
+        // partition's entry would persist for the life of the connection.
+        self.cooldown.retain(|key, _| self.reads.contains_key(key));
     }
 
     /// Fetch records from select "partitions" (journals) and "topics" (collections).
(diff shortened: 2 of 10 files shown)


crates/dekaf/src/lib.rs
```diff
@@ -346,7 +346,7 @@ impl App {
 /// Dispatch a read request `frame` of the current session, writing its response into `out`.
 /// `raw_sasl_auth` is the state of SASL "raw" mode authentication,
 /// and conditions the interpretation of request frames.
-#[tracing::instrument(level = "trace", err(level = "warn"), skip_all)]
+#[tracing::instrument(level = "trace", skip_all)]
```
mdibaiee, line 349: Why is this needed?

crates/dekaf/src/topology.rs
```diff
@@ -261,6 +262,17 @@ impl Collection {
...
+        // cheaply detect when a binding's effective schema has changed,
+        // without a network round trip (unlike `registered_schema_id`'s
+        // `avro_schema_md5`, which addresses the same schemas against the
+        // control plane's schema registry table).
+        let schema_hash = {
+            let key_json = serde_json::to_value(&key_schema).unwrap().to_string();
+            let value_json = serde_json::to_value(&value_schema).unwrap().to_string();
+            format!("{:x}", md5::compute(format!("{key_json}{value_json}")))
+        };
+
```
mdibaiee, line 275: AFAIK the `value_schema` here depends on field selections, which means if two bindings of the same collection have different field selections, we will have two schema hashes for the same collection.

In one iteration of the loop, one of those bindings sets the cooldown for the journal, the other one detects a change in hash and removes it, so there is an edge case where if multiple bindings of the same journal have different field selections, then they end up in this constant loop of setting a hash, then detecting a change in the hash, so on and so forth

crates/dekaf/src/session.rs
```diff
@@ -1103,7 +1187,56 @@ impl Session {
...
+                                first_failed_at: std::time::Instant::now(),
+                                schema_hash: pending.schema_hash.clone(),
+                            },
+                        );
+                        partition_responses.push(
+                            PartitionData::default()
+                                .with_partition_index(partition_request.partition)
+                                .with_error_code(ResponseError::LeaderNotAvailable.code()),
+                        );
+                        self.reads.remove(&key);
```
mdibaiee, line 1226: We remove the read here, but again in session.rs, we check that if there is no read for a partition, we clear it from the cooldown map... is this based on the expectation that the client / consumer keeps sending us reads so that in the session.rs check we will have a new read by the time that the reaper's 30s interval has triggered?

It seems like if we remove the read here, and the reaper's 30s interval triggers soon after, the consumer might not have sent us another read yet... so in session.rs we remove it from cooldown, and here we end up inserting it again..

This may be okay... but it means we can go beyond the 4 minute max time spent on cooldowns

crates/dekaf/src/lib.rs
danielnelson: This removes double logging of errors in the Session, they will be logged in main.rs in `serve()`.

crates/dekaf/src/session.rs
danielnelson: This sounds right, the reaper runs every 30s but removes only items 1m or older.  With kcat, it retries at 500ms so I didn't see this happen, but I could imagine it coming up with a different client.

This shouldn't be a correctness issue, it would log the schema error again and go back into cooldown but it wouldn't close the connection.  I think this behavior is okay, I don't think we want to close the connection in the case that the topic stops being read after an error occurs and I don't think we want to store any state around that it previously errored.

crates/dekaf/src/topology.rs
danielnelson: This seems like an issue, I'll look into it.

danielnelson: The CooldownEntry is now keyed on (topic, partition).

mdibaiee (approved):

PR #3364, opened 2026-08-17
control-plane-api: return stored storage mapping specs

GregorShear:
Stop stripping `collection-data/` from storage-mapping query results. 

The API automatically appends `collection-data/` to the storage bucket prefixes, and the intent was to hide this behavior from the user so as not to confuse them. We'll move that behavior to the UI instead of API since flowctl also depends on this query and needs the complete prefix.


jgraettinger (approved):
LGTM

PR #3373, opened 2026-08-18
control-plane, agent: discovers authorize live specs via Snapshot

bbartman:
## What

First slice of re-deriving #2781 as a strangler series (superseding the
#3341–#3344 stack): the discovers path moves its live-spec authorization off
SQL and onto the in-memory authorization `Snapshot`.

- **`live_specs`**: `get_live_specs` is deleted, replaced by:
  - `get_live_specs_filtered` — evaluates the user's capability to every
    requested name against a pinned `Snapshot`, before any database access,
    and silently omits the names they don't hold it to.
  - `get_live_specs_unfiltered` — the plain-fetch core, used by the
    auto-discover path and `PGControlPlane::get_live_specs` (controllers),
    byte-identical to the prior `None`-filter behavior.
  - `partition_by_authorization` — the pure decision function both the
    filtering wrapper and its unit tests are built on.
- **Discovers**: `Discover` carries an authorization `Snapshot` pinned once
  per executor poll, so decisions can't flip mid-operation (for example,
  during a long-running connector RPC). Both the live-capture precheck and
  the merge-phase collection fetch authorize via that Snapshot.

The Snapshot is **trusted as-is**: there is no staleness classification, no
early-refresh request, and no retryable error. A grant committed after the
Snapshot was taken becomes visible on the watch's own refresh cadence
(`MIN_REFRESH_INTERVAL` 20s / `MAX_REFRESH_INTERVAL` 5m).

## Behavior

Authorization semantics are **unchanged from master**: a spec the user cannot
read is treated as absent, exactly as the SQL-filtered fetch did, and any
authorization error surfaces when they try to publish. A filtered name is
indistinguishable from one that doesn't exist.

What changes is *where* the decision is made — in-process against the
Snapshot's grant graph rather than by `internal.user_roles()` in Postgres —
which is the latency win #2781 asks for.

One consequence to be aware of: a re-discover queued in the window between a
grant landing in Postgres and the next Snapshot refresh evaluates against the
older grants, so an existing capture the user just gained access to can be
re-drafted as new. Bounded by the refresh cadence above.

Data-plane authorization still evaluates `internal.user_roles()` in SQL and
is commented as the next slice.

## Tests

- `live_specs::test::test_partition_by_authorization` (DB-free, over
  `Snapshot::build_fixture`): the authorized/denied split, sorted and
  deduplicated, with unknown names and existing-but-unauthorized names
  partitioning identically.
- `snapshot::tests::test_taken_after_allows_for_temporal_skew`: first unit
  coverage of the `TEMPORAL_SKEW` boundary, which the API-layer callers of
  `taken_after` also depend on.
- `user_discovers::test_discover_filters_unauthorized_capture`: the precheck
  filter — the discover succeeds and drafts the unreadable capture as new.
- `user_discovers::test_discover_merge_filters_unauthorized_collection`: the
  merge-phase filter — an unreadable binding target is drafted fresh rather
  than merged against a live spec the user can't read.
- Test harness: the discovers executor authorizes against a manually-driven
  Snapshot watch that the harness refreshes before each discover poll, so
  grants created by a test are visible; `queue_user_discover` /
  `run_queued_discover` split the queue and poll steps.

### DIFF
--- a/crates/agent/src/discovers.rs
+++ b/crates/agent/src/discovers.rs
@@ -100,6 +100,9 @@ fn precheck_failed(status: JobStatus) -> (JobStatus, ProcessResult) {
 
 pub struct DiscoverExecutor<C: DiscoverConnectors> {
     pub handler: DiscoverHandler<C>,
+    /// Watch of the authorization Snapshot. Each poll pins one Snapshot for
+    /// the entire discover operation.
+    pub snapshot_watch: std::sync::Arc<dyn tokens::Watch<control_plane_api::Snapshot>>,
 }
 
 impl<C: DiscoverConnectors> automations::Executor for DiscoverExecutor<C> {
@@ -127,7 +130,15 @@ impl<C: DiscoverConnectors> automations::Executor for DiscoverExecutor<C> {
         let draft_id = row.draft_id;
         assert_eq!(row.id, task_id);
         let time_queued = chrono::Utc::now().signed_duration_since(row.updated_at);
-        let (status, result) = self.process(row, pool).await?;
+
+        // Pin one authorization Snapshot for the entire operation, so that
+        // authorization decisions cannot flip between discover phases.
+        let snapshot = self.snapshot_watch.token();
+        if let Err(status) = snapshot.result() {
+            anyhow::bail!("authorization snapshot is unavailable: {status}");
+        }
+
+        let (status, result) = self.process(row, snapshot, pool).await?;
         tracing::info!(id=%task_id, %time_queued, ?status, "finished");
         inbox.clear();
         Ok(DiscoverOutcome {
@@ -144,6 +155,7 @@ impl<C: DiscoverConnectors> DiscoverExecutor<C> {
     async fn process(
         &self,
         row: Row,
+        snapshot: std::sync::Arc<tokens::Refresh<control_plane_api::Snapshot>>,
         pool: &sqlx::PgPool,
     ) -> anyhow::Result<(JobStatus, ProcessResult)> {
         tracing::info!(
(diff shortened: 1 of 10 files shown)


crates/agent/src/discovers.rs
```diff
@@ -127,7 +130,15 @@ impl<C: DiscoverConnectors> automations::Executor for DiscoverExecutor<C> {
         let draft_id = row.draft_id;
         assert_eq!(row.id, task_id);
         let time_queued = chrono::Utc::now().signed_duration_since(row.updated_at);
-        let (status, result) = self.process(row, pool).await?;
+
+        // Pin one authorization Snapshot for the entire operation, so that
+        // authorization decisions cannot flip between discover phases.
+        let snapshot = self.snapshot_watch.token();
+        if let Err(status) = snapshot.result() {
```
jgraettinger: nit: snapshots are infallible -- if an error occurs, it doesn't surface as an error here, just a delayed refresh.

crates/control-plane-api/src/live_specs/mod.rs
```diff
@@ -10,27 +10,78 @@ pub use db::{
...
-pub async fn get_live_specs(
+/// Partitions the requested `names` by whether the user holds `capability`
+/// to them, evaluated against the authorization `snapshot`. Returns
+/// `(authorized, denied)`, each sorted and deduplicated.
+fn partition_by_authorization(
+    user_id: Uuid,
+    names: &[String],
+    capability: models::authz::CapabilitySet,
+    snapshot: &crate::Snapshot,
+) -> (Vec<String>, Vec<String>) {
```
jgraettinger: nit: can this be references into `names` to avoid the clones?

jgraettinger (dismissed):
LGTM % nits

crates/agent/src/discovers.rs
bbartman: Fixed.

crates/control-plane-api/src/live_specs/mod.rs
bbartman: Fixed.

jgraettinger (approved):
LGTM

PR #3374, opened 2026-08-18
ci: label PRs that ship with the Deploy agent-api workflow

GregorShear:
Labels a merged PR `pending:agent-api`, `pending:agent`, or `pending:flowctl`, and then remove the label on deploy.

This actually is just a label-reconciler script that runs on PR merge into master, when the deploy agent-api workflow is run, or it can be dispatched manually. A merged PR gets a deployment's `pending:xyz` label when it changed any file that gets built into that deployment's binary, and the version now running doesn't include that change yet.


skord (approved):
LGTM

PR #3381, opened 2026-08-18
docs: Explain 'Additional Backfill Filter' usage

willdonnelly:
**Description:**

**Workflow steps:**

**Notes for reviewers:**

### DIFF
--- a/site/docs/guides/backfilling-data.md
+++ b/site/docs/guides/backfilling-data.md
@@ -382,10 +382,116 @@ To configure this option:
 
 4. Fill out the "Minimum Backfill XID" or "Maximum Backfill XID" field with the `xmin` value you retrieved.
 
-5. Click "Backfill" and make sure that "incremental" is selected in the drop down.
+5. Click "Backfill" and make sure that "Incremental backfill (advanced)" is selected in the drop-down.
 
 6. Save and publish your changes.
 
 7. After the backfill completes, edit the capture again and clear the "Minimum Backfill XID" / "Maximum Backfill XID" field, then save and publish. The XID is only applied when a backfill is triggered, so it has no effect during normal CDC operation — but leaving it in place means that any future backfill (intentional or otherwise) will be silently limited to rows above/below the old transaction ID, potentially missing data. Clearing it ensures future backfills start from a clean state.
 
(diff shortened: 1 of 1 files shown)


aeluce (approved):
Is this ready for review? Wasn't sure with the draft state.

If it's ready, LGTM. Let me know if you were still planning any additions or changes, like if there were more edge cases to account for.

jwhartley:
Flagging this one specifically, because half of it is already live and the
other half is stranded here.

The connector-page half of your work merged as estuary/connectors `05b0cbce9`
and is on `docs.estuary.dev` now: 15 CDC pages (PostgreSQL x5, MySQL x3,
MariaDB x2, SQL Server x4, Oracle) carry a config-table row for
`/advanced/additional_backfill_filter` reading "Optional filter clause which
will be applied to all backfill queries for this binding. Contact Estuary
support for assistance before using this option."

The prose in this PR is the only place the option is actually explained: the
semantics, the indexing and random-sampling guidance, the `Without Primary Key`
pagination hazard, the `Precise` mode validation failure, and the warning that
the expression is used verbatim and a wrong one can leave rows silently missing.
None of that is anywhere on the site. I checked: `additional_backfill_filter`
appears in 15 published pages and in zero of them as prose.

So production currently tells users an expert-level, data-loss-capable option
exists and gives them no way to understand it. That makes this the highest
priority item in the doc migration backlog.

`estuary/flow`'s `site/` tree is no longer published, so this needs to move to
`estuary/docs`: `site/docs/guides/backfilling-data.md` becomes
`content/guides/backfilling-data.md`. The two anchors you reference,
`#dataflow-reset` and `#resource-configuration-backfill-modes`, both still
resolve in the new tree, so the section should port cleanly. Note that
estuary/docs#33 promoted the backfill modes to `###` headings since you
branched, so expect a small conflict around that area.

This is named as the first item of the open-PR work in the [migration
announcement](https://estuaryworkspace.slack.com/archives/C012S5U0WP2/p1787198059608649),
so it is not coming out of nowhere.

Happy to port it for you if that is easier, just say so. Otherwise it is yours.

willdonnelly:
Migrated to https://github.com/estuary/docs/pull/41, does that mean we abandon this one or are we trying to keep changes synchronized or what?

aeluce:
I believe we can merge this or close it out, either way. Docs should be served solely from the docs and connectors repos now, though James might be keeping docs in the flow repo around on the off chance we need a rollback. @jwhartley do you have a preference?

jwhartley:
Since it's already written we may as well merge it, just in case of a rollback.
In future please just do the docs/connectors (new) locations.

PR #3382, opened 2026-08-19
docs: name the single MotherDuck metadata table, document splitting it

jwhartley:
The endpoint `schema` description said "materialization metadata tables" plural. This connector
creates exactly one, `flow_checkpoints_v1`, and never sets `MetaSpecs`
([driver.go#L45](https://github.com/estuary/connectors/blob/d0178ae41/materialize-motherduck/driver.go#L45)).

Adds a short section on running two materializations against the same database. Sharing an endpoint
`database` and `schema` means sharing the checkpoints table, which can cause write contention. Moving
it on a live task points that task at an empty checkpoints table and replays every bound collection,
unless the rows are copied across first.

**Context:** [internal #integrations thread](https://estuaryworkspace.slack.com/archives/C03Q2NRFKDL/p1786647145357949),
where Will diagnosed the contention and gave the workaround this documents. Related:
estuary/connectors#5061 (closed `not_planned`).

This documents current behavior rather than a preferred end state. That thread favours an inherent
fix, along the lines of the Redshift table locking in
[materialize-redshift/driver.go#L851-L861](https://github.com/estuary/connectors/blob/5ce9e716a0090b868ab7e0cbbf39ae376c14d527/materialize-redshift/driver.go#L851-L861),
over documenting a config workaround. Nothing is tracked for that, so the docs gap is worth closing
in the meantime.

**Verified:** the site build renders the new section and its anchor resolves from the properties
table. The build's only failure is the pre-existing `hubspot-real-time` redirect conflict, which also
fails on a clean `master`.

### DIFF
--- a/site/docs/reference/Connectors/materialization-connectors/motherduck.md
+++ b/site/docs/reference/Connectors/materialization-connectors/motherduck.md
@@ -77,7 +77,7 @@ more of your Estuary collections to your desired tables in the database.
 |---------------------------------------|--------------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------|---------|------------------|
 | **`/token`**                          | MotherDuck Service Token | Service token for authenticating with MotherDuck.                                                                                                                | string  | Required         |
 | **`/database`**                       | Database                 | The database to materialize to.                                                                                                                                  | string  | Required         |
-| **`/schema`**                         | Database Schema          | Database schema for bound collection tables (unless overridden within the binding resource configuration) as well as associated materialization metadata tables. | string  | Required         |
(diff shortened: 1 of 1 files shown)


jacobmarble (approved):

jwhartley:
Docs have moved. Connector reference pages now live in `estuary/connectors`
under `docs/reference/Connectors/`, and `estuary/flow`'s `site/` tree is no
longer published.

This PR's changes need to move to `estuary/connectors`. The paths map
one-to-one: `site/docs/reference/Connectors/<page>` becomes
`docs/reference/Connectors/<page>`, so the patch usually applies as-is. For
this PR:

- `site/docs/reference/Connectors/materialization-connectors/motherduck.md` becomes `docs/reference/Connectors/materialization-connectors/motherduck.md`

is served from `estuary/docs` (which aggregates the connectors pages at build
time), so a merge to `master` no longer reaches

PR #3386, opened 2026-08-20
agent: discovers authorize data-planes via Snapshot

bbartman:
## What

Follow-up promised by #3373 (issue #2781): the discovers executor's data-plane
authorization moves off `internal.user_roles()` SQL and onto the authorization
`Snapshot` already pinned for the operation. After this change the discover
path evaluates no authorization decision in SQL.

- The inline query fetching the discover's data plane (gated on
  `internal.user_roles($2, 'read')`) is deleted, along with its `.sqlx` cache
  entry, in favor of `Snapshot::data_plane_by_catalog_name` +
  `tables::UserGrant::is_authorized` at `models::Capability::Read` — the same
  composition the GraphQL `dataPlanes` resolver uses for the identical
  decision.
- **Scope is deliberately minimal**: an unauthorized plane and a missing one
  remain conflated as `JobStatus::NoDataPlane`, exactly like the SQL this
  replaces. Distinguishing them (terminal `NotAuthorized` vs stale-snapshot
  retry/deferral) is left to following commits, as is publications' twin of
  this query — `resolve_live_specs`' `data_plane_names` authorization — which
  is annotated here as the remaining SQL duplicate to be strangled out.
- A `NoDataPlane` outcome now cancels the pinned Snapshot's `revoke` token
  (the `controlplane.rs` precedent), requesting an early background refresh so
  a plane or grant created after the Snapshot was taken is visible to a manual
  retry within roughly `MIN_REFRESH_INTERVAL` instead of up to
  `MAX_REFRESH_INTERVAL`.
- **Behavior change**: a data plane with no HMAC keys at all is excluded from
  the Snapshot by construction, so it now yields `NoDataPlane` instead of
  proceeding toward a discover that could never sign its way into the plane.

## Tests

Written first, red-green (each phase is its own commit):

- Characterization, green before and after the swap:
  `test_discover_unauthorized_data_plane` (plane exists outside the tenant's
  `ops/dp/public/` read grant), `test_discover_missing_data_plane`
  (authorized prefix, no such plane) — both pin `NoDataPlane`.
- New behavior, committed red and turned green by the swap:
  `test_discover_keyless_data_plane` (previously proceeded to Success via
  SQL), `test_discover_no_data_plane_requests_snapshot_refresh` (revoke token
  cancelled).
- Harness: `add_data_plane` (a second plane, so *denied* ≠ *missing*) and
  `queue_user_discover_in_plane` (caller-chosen `data_plane_name`).

All discover-related suites pass (`user_discovers`, `auto_discovers`,
`created_at`, `discovers::test`).


bbartman:
# Implementation plan: discovers data-plane authorization via Snapshot

Issue #2781 follow-up promised by #3373: replace the last `internal.user_roles()`
SQL evaluation in the discover path — the inline data-plane fetch in
`crates/agent/src/discovers.rs` — with an in-process check against the
authorization Snapshot already pinned for the operation.

## Design decisions

- **Scope**: mechanical data-source swap only. None of the retry/deferral
  machinery from the #3341–#3344 stack lands here; `DiscoverState` stays dead.
  Only discovery is affected — no other caller changes behavior.
- **Shape**: inline composition in `DiscoverExecutor::process`, mirroring the
  GraphQL `dataPlanes` resolver — `Snapshot::data_plane_by_catalog_name` for
  the lookup plus `tables::UserGrant::is_authorized` for the decision. No new
  shared helper yet.
- **Capability**: `models::Capability::Read` (legacy → Viewer bundle bits,
  all required), byte-identical to the GraphQL resolver, the
  `authorize_user_*` endpoints, and the #3344 stack's own choice. A dedicated
  fine-grained "use data plane" bit is a possible future sweep, out of scope.
- **Denied vs absent stay conflated** as `JobStatus::NoDataPlane`, exactly
  like today's SQL. Untangling them (terminal `NotAuthorized` vs stale-retry)
  is for following commits.
- **Revoke cancel**: on `NoDataPlane`, cancel the pinned Snapshot's `revoke`
  token (the `controlplane.rs` precedent) to request an early background
  refresh, narrowing the staleness window for a manual retry from up to 5
  minutes to roughly the 20s `MIN_REFRESH_INTERVAL`. Not retry behavior — the
  discover still fails terminally.
- **Accepted behavior change**: a data plane with no HMAC keys at all is
  absent from the Snapshot by construction (`try_fetch` filters them), so it
  now yields `NoDataPlane` instead of proceeding toward a doomed discover.
- **Strangle annotation**: one comment on the `data_plane_names`
  authorization branch of publications' `resolve_live_specs`
  (`crates/control-plane-api/src/publications/specs.rs`) marking it as the
  remaining SQL duplicate of this decision, to be strangled out in following
  commits. No other `internal.user_roles` caller is a duplicate of this
  decision; none are touched.

## Phases (one commit each)

### Phase 1 — tests first (red-green)

Harness additions:
- `add_data_plane(name, hmac_keys)` — a second plane, so *denied* is
  distinguishable from *missing* in tests.
- `queue_user_discover_in_plane(...)` — queue a discover against a
  caller-chosen `data_plane_name` (the existing helper hardcodes
  `ops/dp/public/test` and now delegates).

Four new integration tests in `user_discovers.rs`:

| Test | Scenario | Before swap | After swap |
|---|---|---|---|
| `test_discover_unauthorized_data_plane` | plane exists, outside tenant's `ops/dp/public/` read grant | green (`NoDataPlane`) | green |
| `test_discover_missing_data_plane` | authorized prefix, plane doesn't exist | green (`NoDataPlane`) | green |
| `test_discover_keyless_data_plane` | readable plane with no HMAC keys | **red** (SQL returns it; discover proceeds) | green (`NoDataPlane`) |
| `test_discover_no_data_plane_requests_snapshot_refresh` | `NoDataPlane` outcome cancels the Snapshot's revoke token | **red** (SQL path never touches revoke) | green |

The first two are characterization tests pinning today's behavior; the last
two assert the new behavior and are committed red with the failure shape
documented, then flipped green by Phase 2.

### Phase 2 — the swap

In `DiscoverExecutor::process`:
- Delete the inline `sqlx::query_as!` joining `internal.user_roles()` (and
  its orphaned `.sqlx` cache entry).
- Evaluate `tables::UserGrant::is_authorized(&snapshot.role_grants,
  &snapshot.user_grants, row.user_id, &row.data_plane_name,
  models::Capability::Read)` and resolve the plane via
  `snapshot.data_plane_by_catalog_name`, cloning the `tables::DataPlane` into
  the prepared `Discover`.
- Denied-or-absent → cancel `snapshot.revoke`, warn, return
  `JobStatus::NoDataPlane` (unchanged status).

All four Phase-1 tests green, plus the existing discover suites
(`user_discovers`, `auto_discovers`, `created_at`, `discovers::test`).

### Phase 3 — strangle annotation

Comment on the publications `resolve_live_specs` data-plane query naming it
the remaining SQL duplicate of this decision, to be strangled out in
following commits. No functional change.

bbartman:
Refactored so the pinned authorization Snapshot is resolved exactly once per discover: `Discover` now holds a `&Snapshot` instead of the `Arc<tokens::Refresh<Snapshot>>`, and the single `snapshot.result().unwrap()` happens in `DiscoverExecutor::poll` (user discovers) / `PGControlPlane::discover` (auto-discovers, which already resolved it for the data-plane lookup). `prepare_discover` and `build_merged_catalog` take the reference, so their redundant unwraps are gone. The callers own the Refresh for the duration of the awaited discover, so pinning semantics are unchanged. Commit: 7ae5c78bc98.

crates/agent/src/integration_tests/user_discovers.rs
```diff
@@ -392,6 +392,132 @@ async fn test_discover_filters_unauthorized_capture() {
...
+    assert!(
+        matches!(result.job_status, crate::discovers::JobStatus::NoDataPlane),
+        "expected NoDataPlane, got: {:?}",
+        result.job_status
+    );
+}
+
+#[tokio::test]
+async fn test_discover_unauthorized_data_plane() {
+    let mut harness = TestHarness::init("test_discover_unauthorized_data_plane").await;
```
jgraettinger: nit: could we collapse these into a single test with multiple steps / cases? The harness is identical and the added data planes don't seem to conflict.

The goal is reducing test SLOC, compacting understanding and reducing tokens used the next time we're in here. I'll typically ask an agent to do this kind of compaction (without reducing coverage) as a separate prompt, and it usually yields _increased_ coverage with fewer test SLOC.

crates/agent/src/discovers.rs
```diff
@@ -182,40 +183,28 @@ impl<C: DiscoverConnectors> DiscoverExecutor<C> {
...
-            WHERE data_plane_name = $1
-            AND EXISTS (
-                SELECT 1 FROM internal.user_roles($2, 'read') r
-                WHERE starts_with($1, r.role_prefix)
-            )
-            "#,
-            row.data_plane_name,
+        // Data-plane authorization is evaluated against the pinned Snapshot,
+        // like the live-spec authorization below. An unauthorized plane and a
+        // missing one are deliberately indistinguishable, matching the SQL
```
jgraettinger: nit: the "matching the SQL query this replaced" language is something Claude loves to do, and I'm frequently reminding it to frame prose as a porcelain description of how the system behaves _now_, and not to argue against or compare with behavior we no longer have -- this can only confuse a future reader.

crates/agent/src/discovers.rs
```diff
@@ -182,40 +183,28 @@ impl<C: DiscoverConnectors> DiscoverExecutor<C> {
...
+            &snapshot.user_grants,
             row.user_id,
-        )
-        .fetch_optional(pool)
-        .await
-        .context("fetching data-plane")?;
+            &row.data_plane_name,
+            models::Capability::Read,
+        );
+        let maybe_data_plane = if authorized {
```
jgraettinger: nit: this and the `if let Some(data_plane)` conditional below could be collapsed

crates/control-plane-api/src/discovers/mod.rs
```diff
@@ -43,10 +43,9 @@ pub struct Discover {
     pub created_at: String,
     /// Authorization Snapshot pinned for the entire discover, so that
     /// authorization decisions cannot flip mid-operation (for example, during
-    /// a long-running connector RPC). Snapshot refreshes are infallible --
-    /// a failed refresh only delays the next one -- so `snapshot.result()`
-    /// is always Ok once the watch has become ready.
-    pub snapshot: std::sync::Arc<tokens::Refresh<crate::Snapshot>>,
+    /// a long-running connector RPC). The caller holds the Snapshot's Refresh
```
jgraettinger: nit: "the caller holds the refresh alive" is already implied by it being a borrow-checked Rust reference / should be trimmed.

jgraettinger (dismissed):
LGTM % nits

jgraettinger (approved):
LGTM

crates/agent/src/discovers.rs
bbartman: fixed

crates/control-plane-api/src/discovers/mod.rs
bbartman: fixed

crates/agent/src/discovers.rs
bbartman: fixed

crates/agent/src/integration_tests/user_discovers.rs
bbartman: fixed

Issue #3366, opened 2026-08-17
First-Class Secrets in Estuary


jgraettinger:
# **Today's Baseline**

Estuary utilizes **sops** to protect connector endpoint configurations. Connector schemas
mark secrets with JSON schema annotations, and the config-encryption service encrypts these
specific secrets under an Estuary KMS.

* Integrity Protection: sops ensures that other fields (such as a DB address) are covered by
  HMAC to prevent unauthorized modifications.
* Access Requirements: A user must already possess all secrets to make changes to endpoint
  configs.
* OAuth connector-scoped client secrets are injected at encryption time — by the `oauth`
  edge function ([encrypt-config.ts](supabase/functions/oauth/encrypt-config.ts)), which
  substitutes values from the `connectors` table before calling config-encryption's
  encrypt route — alongside user access and refresh tokens, to form a fully-specified
  endpoint config.
* User-Managed Keys: Some users bring their own KMS, using sops themselves to encrypt
  configs and granting individual Estuary data-planes the capability to decrypt.

### **Overlay Mechanism**

Users frequently want to change portions of a config without needing to re-enter secrets.
We've introduced "overlays" for blessed JSON schema locations that are not protected by sops
HMAC.

### **ConfigUpdate Mechanism**

OAuth connectors call the config-encryption service with new credentials (after token
rotation) and emit a `configUpdate` log to the control plane which is published into the
task. This is our *only* facility for token rotation today.

## **Exfiltration Risks**

We must be wary of two kinds of disclosure attacks:

1. Privilege Escalation: An attacker with catalog access to read a specification (only)
   attempts to leverage that capability to exfiltrate the secret.
2. GitOps Exposure: A user employs GitOps to version control a wrapped endpoint config in a
   public repository where an attacker can see it.

As a baseline, we assume an attacker can obtain read access to a wrapped endpoint config.
Today decryption is tied only to the KMS, not the individual catalog name it configures, and
we must be wary of “cloning” attacks where an attacker lifts a wrapped spec to a name they
control. Our defense today relies on the attacker being unable to change parts of the config
that allow for disclosure (like changing a DB address to a malicious server) without
breaking the HMAC.

### **Mixed Authority**

OAuth client IDs and secrets are injected by the `oauth` edge function alongside user
access and refresh tokens. Due to this mixing we **cannot** grant users decryption of their endpoint
configs, even to users holding full Admin capability, because it would disclose
Estuary-managed secrets alongside.

# **Identified Problems**

[...]

Issue #2963, opened 2026-05-20
Support encryption of redactSalt in capture/derivation specs


jwhartley:
> Priority: Low (control-plane triage, 2026-06-18)
>
> Rarely used in practice. Worth doing as hardening, but not urgent.

---

## Context

Surfaced in [#2950 (review comment)](https://github.com/estuary/flow/pull/2950#discussion_r3261544099) while documenting the `redactSalt` override. The docs tell users to "treat the salt as sensitive" — but unlike endpoint configs, there's no automatic encryption for it today, so it ends up stored as plain text in the spec.

## Current state

`flowctl`'s `encrypt_configs` ([`crates/flowctl/src/draft/encrypt.rs`](https://github.com/estuary/flow/blob/master/crates/flowctl/src/draft/encrypt.rs)) only encrypts three things on publish:

1. Capture `endpoint.config`
2. Materialization `endpoint.config`
3. Materialization `triggers` (with HMAC-excluded fields handling)

Encryption is driven by JSON Schemas with `secret: true` annotations (the connector endpoint spec schema for endpoints, `triggers_schema()` for triggers).

The top-level `redactSalt` field on `CaptureDef` ([`crates/models/src/captures.rs:47`](https://github.com/estuary/flow/blob/master/crates/models/src/captures.rs#L47)) and `DerivationDef` ([`crates/models/src/derivation.rs:32`](https://github.com/estuary/flow/blob/master/crates/models/src/derivation.rs#L32)) is `Option<bytes::Bytes>` and is not walked by `encrypt_configs`. The runtime reads `redact_salt.to_vec()` directly off the protobuf `CaptureSpec`/`DerivationSpec` ([`crates/runtime/src/capture/task.rs:124`](https://github.com/estuary/flow/blob/master/crates/runtime/src/capture/task.rs#L124), [`crates/runtime/src/derive/task.rs:99`](https://github.com/estuary/flow/blob/master/crates/runtime/src/derive/task.rs#L99)), so there is no decryption boundary either.

Net: a user-supplied `redactSalt` is stored and shipped as plain bytes.

## Why it matters

If anyone obtains the stored salt plus a candidate plaintext value (email, phone, SSN, etc.), they can compute the corresponding hash directly. The doc's "treat the salt as sensitive" guidance is hard to live up to when the value sits as plain text in the spec.

## Proposed direction

[...]

PR #3389, opened 2026-08-21
agent: discovers require SpecEdit to the capture

bbartman:
## What

Next slice of the #2781 strangler series (following #3373 and #3386): the discovers executor now **proves** the user's authorization over the specs a discover touches, rather than only filtering. All decisions evaluate against the Snapshot pinned for the operation.

- **SpecEdit on the capture, hard-fail**: an unconditional precheck on `capture_name` runs ahead of data-plane resolution. A user without SpecEdit gets the new terminal `JobStatus::NotAuthorized` plus a single grant-based draft error, and the Snapshot's `revoke` token is cancelled to request an early background refresh (the `NoDataPlane` precedent).
- **Live-capture merge-base fetch filters at `SpecEdit | CatalogRead`**: the fetch also *discloses* the live model into the user's draft, and disclosure is CatalogRead's axis. Every real bundle conveying SpecEdit conveys CatalogRead too, so this is a no-op in practice — it keeps the fetch's read semantics locally correct rather than leaning on that bundling convention.
- **Collections keep the silent CatalogRead filter, verbatim**: an unauthorized binding-target collection remains indistinguishable from a nonexistent one (#3373's doctrine).
- The auto-discover path (`filter_user_authz: false`, system user) is untouched.

## Reviewer callout: behavior change for CatalogRead-only users

Previously a user lacking edit authority on the capture had it silently drafted as **brand-new** (filtered-as-absent), with the authorization error surfacing only at publish. Now the discover **fails up-front** with `NotAuthorized`. This is a deliberate security-posture change: a user who cannot SpecEdit the capture can never publish the result, so silent success was a UX trap, and the up-front check is cheaper than a doomed connector RPC.

This does **not** create an existence oracle: the check is a pure function of the grant graph and the requested name — its outcome is identical whether or not a spec exists at that name, which `test_discover_not_authorized_capture` pins by asserting byte-identical outcomes for an existing and a nonexistent unauthorized name. (The data-plane check, by contrast, must conflate unauthorized-with-missing because it is existence-conditional; this check never looks anything up.)

Also note the legacy-capability fallout, pinned by test: legacy `read` (Viewer) and `write` (Writer) grants convey CatalogRead but not SpecEdit and are now refused; `admin` conveys SpecEdit and succeeds. Nothing usable is lost — publications already require admin.

## Scope note: secrets in specs

CatalogRead remains the platform's spec-read boundary. A discover merge copies CatalogRead-visible live collection models (including any `derive` section) into the user's draft — the identical surface `flowctl catalog pull-specs` already grants, so this PR neither widens nor narrows it. Secret-in-spec exposure (unencrypted configs, plaintext `redactSalt`) is explicitly out of scope here and is owned by #3366 (first-class secrets) and #2963 (redactSalt encryption).

## No retry mechanism, by design

There is deliberately no stale-snapshot deferral or retry machinery, now or planned: an authorization failure is terminal, the `revoke` cancellation narrows the staleness window to roughly `MIN_REFRESH_INTERVAL`, and the error text asks the user to retry in a moment. `DiscoverState`, which was carried as dead state for a planned deferral mechanism (#3279), is accordingly emptied.

## UI

The dashboard will render `job_status.type == "notAuthorized"` as an unknown status until it learns the new variant; the draft error text is the user-facing message in the interim. The wire tag is pinned by a serde test.

## Tests

Written first, red-green (each phase is its own commit):

- `test_discover_not_authorized_capture`: terminal `NotAuthorized`, grant-based draft error, cancelled revoke token — asserted identically for existing and nonexistent unauthorized names (the no-oracle guard).
- `test_discover_capture_requires_spec_edit`: `read` → refused, `write` → refused, `admin` → success, via a bare user holding a single direct grant.
- `test_discover_not_authorized_wins_over_missing_plane`: the SpecEdit check precedes data-plane resolution.
- `test_job_status_not_authorized_serde`: pins the UI-facing wire tag.
- Removed `test_discover_filters_unauthorized_capture`, which pinned the prior filter semantics.
- Unchanged and still green: `test_discover_merge_filters_unauthorized_collection` (the collections filter), `test_user_discovers`, `test_discover_no_data_plane`, `created_at`, and the `auto_discovers` suite.

### DIFF
--- a/crates/agent/src/discovers.rs
+++ b/crates/agent/src/discovers.rs
@@ -193,7 +193,7 @@ impl<C: DiscoverConnectors> DiscoverExecutor<C> {
                 ),
                 catalog_name: row.capture_name.clone(),
                 detail: format!(
-                    "user is not authorized to edit specs under '{}'; if this access was granted recently, retry the discover in a moment",
+                    "user is not authorized to edit specs under '{}'; if this access was granted recently, please retry in a moment",
                     row.capture_name
                 ),
             }];
--- a/crates/agent/src/discovers.rs
+++ b/crates/agent/src/discovers.rs
@@ -27,6 +27,7 @@ pub enum JobStatus {
     },
     DeprecatedBackground,
     NoDataPlane,
+    NotAuthorized,
 }
 
 impl JobStatus {
@@ -380,6 +381,16 @@ mod test {
 
     use models::Id;
     use uuid::Uuid;
+
+    // `discovers.job_status` is a UI-facing contract: pin the wire tag.
+    #[test]
+    fn test_job_status_not_authorized_serde() {
+        let status = serde_json::to_value(super::JobStatus::NotAuthorized).unwrap();
+        assert_eq!(serde_json::json!({"type": "notAuthorized"}), status);
+
+        let round: super::JobStatus = serde_json::from_value(status).unwrap();
+        assert!(matches!(round, super::JobStatus::NotAuthorized));
+    }
 
     #[tokio::test]
     async fn test_prepare_discover() {
--- a/crates/agent/src/discovers.rs
+++ b/crates/agent/src/discovers.rs
@@ -98,9 +98,6 @@ impl<C: DiscoverConnectors> automations::Executor for DiscoverExecutor<C> {
 
     type Receive = serde_json::Value;
 
-    /// Polls are stateless and persist the JSON `null` that `None`
-    /// round-trips, keeping in-flight task state readable across a deploy
-    /// in either direction.
     type State = Option<DiscoverState>;
 
     type Outcome = DiscoverOutcome;
@@ -308,19 +305,18 @@ async fn prepare_discover<'a>(
     // embedded in its control-plane Id — is carried on the Discover request,
     // so that re-discovers resolve connector feature-flag defaults as the
     // running task does. It's empty for a task which doesn't exist yet.
-    // The executor has already required the user hold SpecEdit to the
-    // capture, but this fetch also *discloses* the live model into the
-    // user's draft, and disclosure is CatalogRead's axis. Bundles which
-    // convey SpecEdit convey CatalogRead too, so this is a no-op for
-    // real grants — but requiring both here keeps the fetch's read
-    // semantics locally correct rather than leaning on that bundling
-    // convention: a SpecEdit-only grant may discover, with the live
-    // capture treated as absent.
+    //
+    // process() already required the user hold SpecEdit to the capture, but
+    // this fetch *discloses* the live model into the user's draft, and
+    // disclosure is CatalogRead's axis — so the fetch filters on CatalogRead
+    // in its own right rather than leaning on the convention that bundles
+    // conveying SpecEdit convey CatalogRead too. Under a SpecEdit-only
+    // grant the discover proceeds with the live capture treated as absent.
     let name = &[capture_name.to_string()];
     let live = live_specs::get_live_specs_filtered(
         user_id,
         name,
-        models::authz::Capability::SpecEdit | models::authz::Capability::CatalogRead,
+        models::authz::Capability::CatalogRead,
         snapshot,
         pool,
     )
--- a/crates/agent/src/discovers.rs
+++ b/crates/agent/src/discovers.rs
@@ -98,9 +98,9 @@ impl<C: DiscoverConnectors> automations::Executor for DiscoverExecutor<C> {
 
     type Receive = serde_json::Value;
 
-    /// `None` round-trips as the JSON `null` that stateless polls have always
-    /// persisted, keeping in-flight tasks readable across a deploy in either
-    /// direction.
+    /// Polls are stateless and persist the JSON `null` that `None`
+    /// round-trips, keeping in-flight task state readable across a deploy
+    /// in either direction.
     type State = Option<DiscoverState>;
 
     type Outcome = DiscoverOutcome;
@@ -203,9 +203,9 @@ impl<C: DiscoverConnectors> DiscoverExecutor<C> {
             return Ok((JobStatus::NotAuthorized, Err(errors)));
         }
 
-        // Data-plane authorization is evaluated against the pinned Snapshot,
-        // like the live-spec authorization below. An unauthorized plane and a
-        // missing one are deliberately indistinguishable.
+        // Data-plane authorization: unlike the capture check above, resolving
+        // the plane consults what exists in the Snapshot, so an unauthorized
+        // plane and a missing one are deliberately indistinguishable.
         let Some(data_plane) = tables::UserGrant::is_authorized(
             &snapshot.role_grants,
             &snapshot.user_grants,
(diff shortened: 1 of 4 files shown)


bbartman:
## Plan of record

The negotiated plan this PR implements, with each decision and its rationale.

**Goal.** The discover flow proves the user's authorization over the specs it reads, against the pinned Snapshot: **SpecEdit** on the capture being discovered (the spec the discover validates and edits), and **CatalogRead** on all binding-target collections.

### Decisions

1. **Secret-exposure posture**: CatalogRead remains the platform's spec-read boundary. A collection spec can embed a connector config (derivations: `derive.using.connector.config`) and `redact_salt`, and the merge copies CatalogRead-visible live models whole into the user's draft — but this is the identical surface `pull-specs` already grants. Out of scope here; owned by #3366 / #2963.
2. **Filter or fail**: hard-fail on the capture's SpecEdit check (prefix-based, leaks nothing, runs before the connector RPC, and a user who can't SpecEdit the capture can never publish the result). Collections keep the silent CatalogRead filter — binding targets are only fully known after the connector responds, and an existing capture can legitimately carry bindings into prefixes the discovering user can't read; a hard failure there would also be an existence oracle.
3. **Failure shape**: new `JobStatus::NotAuthorized` variant plus one grant-based draft error, so the UI shows something actionable before it learns the new status type. Guardrails: error wording speaks only to grants (never "exists but not accessible"), and the check is unconditional — never nested inside an "if a live spec exists" branch. Both are pinned by tests.
4. **Precheck mechanics**: SpecEdit check runs first, before the data-plane check (most fundamental gate, most actionable signal); `NotAuthorized` cancels `snapshot.revoke` like `NoDataPlane` does; the live-capture fetch filters at `SpecEdit | CatalogRead` — the hard check gates the *operation*, the fetch filter carries the *disclosure* semantics, so a hypothetical SpecEdit-without-CatalogRead grant may discover but has the live model treated as absent.
5. **Staleness**: no deferral or retry mechanism, now or ever. `revoke.cancel()` narrows the staleness window; the error text asks the user to retry in a moment. `DiscoverState` is emptied accordingly (its `awaiting_snapshot_after` member existed only for the now-cancelled deferral plan).
6. **Untouched**: the merge-phase collections filter (verbatim), and the auto-discover / controller path (`filter_user_authz: false`, system user).

### Test plan

- `test_discover_not_authorized_capture` — terminal status, grant-based error, cancelled revoke token; **identical outcomes for an existing and a nonexistent unauthorized name** (a future refactor that makes the check existence-conditional turns this red).
- `test_discover_capture_requires_spec_edit` — legacy grant loop: `read` and `write` refused (CatalogRead/Writer without SpecEdit), `admin` succeeds.
- `test_discover_not_authorized_wins_over_missing_plane` — pins check ordering.
- `test_job_status_not_authorized_serde` — pins the `{"type":"notAuthorized"}` wire tag.
- Negative space that must stay green: `test_discover_merge_filters_unauthorized_collection`, `test_user_discovers`, `test_discover_no_data_plane`, `created_at`, `auto_discovers`, `test_prepare_discover`.

### Commit shape

Red-green, per the #3386 convention: commit 1 adds the `NotAuthorized` variant and all tests (verified red — the grant-loop test fails with `NoDataPlane` where `NotAuthorized` is expected); commit 2 adds the precheck, fetch change, and `DiscoverState` cleanup (all 14 discover-related tests green).

### Follow-ups (outside this PR)

- estuary/ui: teach the dashboard the `notAuthorized` job status.
- #3366 / #2963 own the secrets-in-specs exposure this PR deliberately does not change.

crates/agent/src/discovers.rs
```diff
@@ -110,9 +98,9 @@ impl<C: DiscoverConnectors> automations::Executor for DiscoverExecutor<C> {
 
     type Receive = serde_json::Value;
 
-    /// `None` — the common, never-deferred case — round-trips as the JSON
-    /// `null` that stateless polls have always persisted, keeping in-flight
-    /// tasks readable across a deploy in either direction.
+    /// Polls are stateless and persist the JSON `null` that `None`
+    /// round-trips, keeping in-flight task state readable across a deploy
+    /// in either direction.
```
jgraettinger: nit: I'm not understanding what this comment means ("deploy in either direction"?)

crates/agent/src/discovers.rs
```diff
@@ -183,9 +171,41 @@ impl<C: DiscoverConnectors> DiscoverExecutor<C> {
...
+                    tables::synthetic_scope(models::CatalogType::Capture, &row.capture_name)
+                        .to_string(),
+                ),
+                catalog_name: row.capture_name.clone(),
+                detail: format!(
+                    "user is not authorized to edit specs under '{}'; if this access was granted recently, retry the discover in a moment",
+                    row.capture_name
+                ),
+            }];
+            return Ok((JobStatus::NotAuthorized, Err(errors)));
```
jgraettinger, line 200: Please QA that UI behavior is reasonable on a local stack; I'm not sure if there are UI changes that must happen for a new JobStatus, or if this "Just Works" or not.

crates/agent/src/discovers.rs
```diff
@@ -288,14 +308,19 @@ async fn prepare_discover<'a>(
     // embedded in its control-plane Id — is carried on the Discover request,
     // so that re-discovers resolve connector feature-flag defaults as the
     // running task does. It's empty for a task which doesn't exist yet.
-    // Filter to only specs the user holds CatalogRead to: a capture they
-    // can't read is treated as absent, and any error surfaces when they
-    // try to publish.
+    // The executor has already required the user hold SpecEdit to the
```
jgraettinger: It's inappropriate to make this decision here. I think this should just be `CatalogRead`. Yes, it'd be weird to have SpecEdit but not CatalogRead, and it's not a thing today, but there's nothing logically inconsistent about that positioning.

jgraettinger (commented):
Some comments.

I would prefer that fetching live collections is also a hard error if `CatalogRead` isn't met, instead of silently filtering, but am happy with tackling SpecEdit for the capture alone if that's an easier change to ship.

crates/agent/src/discovers.rs
bbartman: Sorry this was pulled from a previous change, and it looks like it did the thing where it grabs the wrong context for this comment. Basically, this state exists in order to support the original PR. which is why it seems to imply an either direction, those directions being either we use it or remove it, because I cleaned up the unused parameter here, but I didn't want to remove the empty object. The empty object is for future proofing, if this ever gets a state. I'm going to remove the comment for now.

crates/agent/src/discovers.rs
bbartman: I may be missing something here. Is this because it's enforced higher up the call chain? Also I fixed this and made it catalog read.

crates/agent/src/discovers.rs
bbartman: image
This is what I'm seeing, I'm not 100% sure if this is correct but I a local stack to recreate the not authorized error, but I'm not sure if this what it should look like or not.

I will say that the displayed error message here doesn't make all that much sense because the message because currently I'm creating a capture. I'm going to change it to remove the reference to discovery, and just make it say please try again.

jgraettinger (approved):

PR #3393, opened 2026-08-25
control-plane-api: publications check spec-to-spec authorization via Snapshot

bbartman:
Continues the #2781 strangulation: with user-capability and data-plane authorization already Snapshot-evaluated (#3387), this migrates the last SQL-computed authorization in the publications path — the **spec-to-spec** `reads_from`/`writes_to` checks — onto the pinned Snapshot, and makes `fetch_live_specs` a fully pure data fetch.

Stacked on #3387 (`bmb/2781-publications-snapshot-authz`).

## What changes

- `Snapshot` gains `is_role_authorized` (wrapping `tables::RoleGrant::is_authorized`, mirroring `is_user_authorized`) and `spec_capabilities` (the directly-held grants of a spec's name, for error rendering).
- `resolve_live_specs` evaluates spec-to-spec authorization of drafted specs against the Snapshot instead of the `spec_capabilities` column that `fetch_live_specs` computed from a `role_grants` subquery. Error messages are rendered identically.
- `fetch_live_specs` drops the `fetch_spec_capabilities` flag, the `role_grants` subquery, and the `LiveSpec.spec_capabilities` field (plus the placeholder column in `fetch_expanded_live_specs`), completing the fetch/decide split: fetches are pure data reads; every authorization decision happens in-process against the pinned Snapshot.

## Deliberate semantic changes

1. **Single-hop → grant-graph walk.** The SQL column matched direct `role_grants` rows only. `Snapshot::is_role_authorized` walks the full grant graph — the *same walk which authorizes running tasks* (`authorize_task`). A spec authorized through a chain (e.g. `dogs/` admin of `middle/`, `middle/` write to `cats/`) now verifies at publication exactly as it would run, and bundle-carrying grants count the way the runtime counts them. Strictly more permissive; mirrors the accepted #2848 upward-walk divergence for user capability, and closes a gap where publication could reject a draft whose running task the runtime would happily authorize.
2. **Denials request an early Snapshot refresh**, consistent with every other Snapshot-evaluated denial that fails a publication: the needed role grant may postdate the Snapshot, and cancelling the revoke token narrows the staleness window for a retry.
3. **Freshness.** Spec-to-spec checks now read the pinned Snapshot rather than live SQL — the same accepted staleness story as the rest of #3387, now also covering system publications (which always evaluate spec-to-spec even though they skip user checks).

Because of (3), the test harness's control plane moves off its DB-backed watch — whose init-time Snapshot predates test fixtures, and whose source cool-off prevents a re-take within a test — and onto the shared manually-driven watch, which `TestControlPlane::publish` re-takes and pins exactly as the executor poll paths do.

## Test plan

- New red-green coverage: the previously untested `writes_to` denial leg (with grant rendering and revoke-cancellation), and the grant-chain publication — pinned as denied under single-hop SQL in the first commit, flipped to authorized by the migration.
- New unit tests for `Snapshot::spec_capabilities` and `Snapshot::is_role_authorized` (legacy capability comparisons through bundle bits).
- Existing pins updated: the spec-to-spec no-refresh assertion flips to refresh-requested; `test_publication_system_user_data_plane_filter` observes cancellation on the Snapshot `publish` pins; `test_injected_ops_collections_are_not_locked` re-pins after creating the role grant its build needs.
- Full `agent` (107) and `control-plane-api` (165) suites pass.

Remaining `user_roles`/RLS estate after this PR (unchanged): storage-mappings directive, `create_data_plane` + `update_l2_reporting` `ops/` checks, and the PostgREST `_ext`-view estate.


bbartman:
## Approved plan (tracking)

Migration of the spec-to-spec authorization leg (issue #2781), stacked on #3387. Decisions approved 2026-08-25: land as a **stacked PR** (not appended to #3387), and spec-to-spec denials **cancel the revoke token** per the every-Snapshot-denial-cancels rule.

1. ✅ `agent: pin spec-to-spec write authorization and its single-hop scope` — red baseline: the untested `writes_to` denial (grant rendering + no-refresh pin), and the grant chain (`dogs/` admin of `middle/`, `middle/` write to `cats/`) pinned as **denied** under the SQL direct-row match.
2. ✅ `control-plane-api: add role-grant walk and grant rendering to Snapshot` — `Snapshot::is_role_authorized` (mirrors `is_user_authorized`) and `Snapshot::spec_capabilities` (direct grants of a spec's name, error rendering only), with unit tests. Unused until the migration.
3. ✅ `control-plane-api: publications check spec-to-spec authorization via Snapshot` — `resolve_live_specs` evaluates `reads_from`/`writes_to` via the Snapshot; denials `request_refresh()`; red pins flip (chain now authorizes; refresh now requested). Includes the harness consequence: `TestControlPlane` moves onto the shared manually-driven watch, re-taken and pinned per publish, since system publications always evaluate spec-to-spec.
4. ✅ `control-plane-api: fetch_live_specs becomes a pure data fetch` — drop the `fetch_spec_capabilities` flag, `role_grants` subquery, `LiveSpec.spec_capabilities`, and the expansion placeholder column; regenerate `.sqlx`.

**Scope limits:** no staleness contract (`ok_or_stale` / retryable stale errors — deferred with the rest of #3387's follow-ups); user checks stay on legacy Admin/Read (no new-bundle capabilities); remaining `user_roles` estate (storage-mappings directive, `create_data_plane`, `update_l2_reporting`, PostgREST `_ext` views) untouched.

crates/control-plane-api/src/server/snapshot.rs
```diff
@@ -121,6 +121,38 @@ impl Snapshot {
         )
     }
 
+    /// Returns whether the role or spec named `subject_role_or_name` holds
+    /// `capability` to `object_role_or_name`, evaluated against this
+    /// Snapshot's role grants. Unlike the SQL `spec_capabilities` match it
```
jgraettinger: nit: I'd strike the "Unlike the SQL..." sentence, this is historical rationalization / differentiation that's irrelevant going forward and doesn't belong in the comment prose.

crates/control-plane-api/src/server/snapshot.rs
```diff
@@ -121,6 +121,38 @@ impl Snapshot {
...
+            capability,
+        )
+    }
+
+    /// Returns the "spec capabilities" of a spec named `catalog_name`: the
+    /// role grants whose `subject_role` is a prefix of the name — the grants
+    /// the spec holds by virtue of its own name. Only the directly-held
+    /// grants are rendered, not the transitive reach of the walk, so this is
+    /// only for error reporting to improve authorization error messages.
+    pub fn spec_capabilities(&self, catalog_name: &str) -> Vec<tables::RoleGrant> {
```
jgraettinger: nit: I'm worried about routine sprawl in Snapshot.

This feels like it ought to be a free function that _takes_ a &Snapshot, not a behavior of one.

At the moment there's one function calling it: make it a nested function or closure of that routine body?

jgraettinger (dismissed):
LGTM % nits

crates/control-plane-api/src/server/snapshot.rs
bbartman: Fixed.

crates/control-plane-api/src/server/snapshot.rs
bbartman: Fixed. refactored to a closure, I had claude check for other possible places that this could be used and there wasn't any, so I figured that it could just be a closure for now.

jgraettinger (approved):

PR #3400, opened 2026-08-25
runtime: captures use the V2 runtime by default

williamhbaker:
**Description:**

A capture shard now selects V2 unless its labels set `enable-runtime-v2` to "false". That flag is the only way to keep a capture on V1. Derivations and materializations do not change. They still need an explicit "true".

The restart errors no longer say the flag is "set" or "unset", because those words no longer describe the check. Each error now names the runtime the shard selects, and the runtime it runs.

**Workflow steps:**

**Notes for reviewers:**

Agentic Q/A completed on a local stack


jgraettinger (approved):
LGTM

PR #3403, opened 2026-08-26
control-plane-api: Add public/private filter option to DataPlanes query

### LINKED ISSUE #2065: Switch data plane table queries over to GQL
Serve the admin data planes table (`src/components/tables/DataPlanes/`) from the GQL `dataPlanes` query instead of PostgREST.

Keep the same UX, and add pagination.

adrian-estuary:
**Description:**
Related to https://github.com/estuary/ui/issues/2065

Adding a new `public` boolean filter to allow filtering the `DataPlanesQuery` results to just public or private data planes. This is required to support the public/private filter control on the admin "Data Planes" table while supporting cursor-based pagination. This is part of the Supabase -> GQL migration.

**UI control needing this functionality**
Screenshot 2026-08-26 at 1 54 00 AM

**GQL query showing the filter results**
Screenshot 2026-08-26 at 1 59 53 AM

**Workflow steps:**

The filter will be used by the data planes table in the admin UI, as shown above.

No documentation changed.

**Notes for reviewers:**

### DIFF
--- a/crates/control-plane-api/src/server/public/graphql/data_planes.rs
+++ b/crates/control-plane-api/src/server/public/graphql/data_planes.rs
@@ -1250,6 +1250,102 @@ mod tests {
         );
     }
 
+    // The `filter.public.eq` argument narrows the listing by the public /
+    // private distinction, which is derived from the plane's name prefix
+    // rather than a column. The fixture's only private plane is otherwise
+    // excluded from snapshots for having no HMAC keys, so give it keys and
+    // grant alice read on `ops/dp/private/` to make it visible.
+    #[sqlx::test(
+        migrations = "../../supabase/migrations",
+        fixtures(path = "../../../fixtures", scripts("data_planes", "alice"))
+    )]
+    async fn test_graphql_data_planes_public_filter(pool: sqlx::PgPool) {
+        let _guard = test_server::init();
+
+        let private_dp = "ops/dp/private/defunct/az-westeurope-c1";
+        let public_dps = vec![
+            "ops/dp/public/aws-us-west-2-c1".to_string(),
+            "ops/dp/public/gcp-us-central1-c2".to_string(),
+        ];
+
+        // give test user read access to the private plane
+        sqlx::query("UPDATE data_planes SET hmac_keys = '{c2VjcmV0}' WHERE data_plane_name = $1")
+            .bind(private_dp)
+            .execute(&pool)
+            .await
+            .unwrap();
+        sqlx::query(
+            "INSERT INTO role_grants (subject_role, object_role, capability)
+             VALUES ('aliceCo/', 'ops/dp/private/', 'read')",
+        )
+        .execute(&pool)
+        .await
+        .unwrap();
+
+        let server =
+            test_server::TestServer::start(pool.clone(), test_server::snapshot(pool, false).await)
+                .await;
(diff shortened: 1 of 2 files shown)


crates/control-plane-api/src/server/public/graphql/data_planes.rs
```diff
@@ -16,6 +16,9 @@ pub struct DataPlanesFilter {
     pub id: Option<filters::IdFilter>,
     /// Filter on the `closed` flag.
     pub closed: Option<filters::BoolFilter>,
+    /// Filter on the `public` flag, which is derived from the data plane's
+    /// name prefix (`ops/dp/public/` vs `ops/dp/private/`).
+    pub public: Option<filters::BoolFilter>,
```
adrian-estuary, line 21: Kept the same naming convention as `closed` (i.e. `public` vs `isPublic`).

crates/control-plane-api/src/server/public/graphql/data_planes.rs
```diff
@@ -1239,6 +1250,102 @@ mod tests {
...
+            .execute(&pool)
+            .await
+            .unwrap();
+        sqlx::query(
+            "INSERT INTO role_grants (subject_role, object_role, capability)
+             VALUES ('aliceCo/', 'ops/dp/private/', 'read')",
+        )
+        .execute(&pool)
+        .await
+        .unwrap();
```
adrian-estuary: Needed this to give the test user read access to the private plane. I didn't want to update the fixtures because that might break other tests. Let me know if there is a better way to do this. 🙏

crates/control-plane-api/src/server/public/graphql/data_planes.rs
```diff
@@ -1239,6 +1250,102 @@ mod tests {
...
+
+        // give test user read access to the private plane
+        sqlx::query("UPDATE data_planes SET hmac_keys = '{c2VjcmV0}' WHERE data_plane_name = $1")
+            .bind(private_dp)
+            .execute(&pool)
+            .await
+            .unwrap();
+        sqlx::query(
+            "INSERT INTO role_grants (subject_role, object_role, capability)
+             VALUES ('aliceCo/', 'ops/dp/private/', 'read')",
```
GregorShear: If you wanted to be picky about it we would typically never give a tenant access to the entire private prefix. in practice `aliceCo/` would have a role grant to `ops/dp/private/aliceCo/`

```suggestion
             VALUES ('aliceCo/', 'ops/dp/private/defunct/', 'read')",
```

GregorShear (dismissed):

crates/control-plane-api/src/server/public/graphql/data_planes.rs
GregorShear: actually it might be even better to import an existing private dataplane for aliceco

Add `private_links.sql` to the fixture list and test against `ops/dp/private/aliceCo/aws-us-east-1-c1`.

adrian-estuary: Updated! Thanks for the suggestion, @GregorShear!

crates/control-plane-api/src/server/public/graphql/data_planes.rs
adrian-estuary: Updated to instead include the `private_links` fixture, giving the test user access to a specific private data plane.

GregorShear (approved):

PR #3408, opened 2026-08-26
flowctl: bound preview-next fixture memory and disk

### LINKED ISSUE #3399: flowctl: preview-next fixture writer bypasses shuffle-log disk backpressure, growing memory and disk unbounded
## Summary

`flowctl raw preview-next --fixture` grows unbounded in **both memory and disk** where `flowctl preview` (V1) stays flat, because the fixture's log writer is uncoupled from consumer progress. Against a connector paced at 20 MB/s, flowctl reached **21.7 GB RSS** on work where V1 peaked at **0.34 GB**, and wrote ~4 GiB of fixture log per 5 GiB round.

#3391 bounded the fixture's *blocks* (fixing the `block::encode` assert). It did not address backpressure, which is the subject of this issue.

## Root cause

The live V2 path and the fixture path write the shuffle log through different entry points, and the fixture's bypasses the disk-backlog gate.

`LogActor` gates every append (`crates/shuffle/src/log/actor.rs:101-102`):

```rust
let may_buffer =
    !self.append_heap.is_empty() && !self.block.is_full() && !disk_back_pressure;
```

`disk_back_pressure` trips once sealed segments exceed `shuffle_disk_limit_bytes` (default `134217728` — 128 MiB, `crates/labels/src/shard.rs:203`) and releases at half that, with hysteresis. Sealed segments are reclaimed only as the **consumer** unlinks them while reading, so in production the writer is throttled by consumer progress and the log backlog stays pinned near 128 MiB.

The fixture calls `Writer::append_block` directly (`crates/flowctl/src/raw/preview_next/fixture.rs:683` and `:709`) and consults **neither** gate. It parses and writes as fast as it can drain the FIFO, with no coupling to the connector at all.

V1 `flowctl preview` has no shuffle log — it feeds the connector from the FIFO directly, so a slow connector backpressures the generator through the pipe. Hence 0.34 GB.

## What this is *not*

Not the fixture's own buffer. `FIXTURE_BLOCK_ENTRIES = 32 * 1024` and `FIXTURE_BLOCK_BYTES = 64 * 1024 * 1024` (`fixture.rs:76-77`) bound it to 64 MiB per shard, and #3391's comment states the byte threshold exists precisely so peak memory tracks block size rather than transaction size. That part works.

The residue is downstream: `preview-next` runs the fixture writer *and* the entire V2 runtime — leader, shard actors, log readers — in one process, and the unthrottled writer lets the read side run arbitrarily far ahead of a slow connector. `SealedSegment` is only a path plus a size (`crates/shuffle/src/log/writer/sealed.rs:8`), so the `sealed` Vec is negligible; the disk backlog is real files, and the RSS is the runtime's in-flight state above them.

## Evidence

materialize-snowflake, 5 GiB transactions, delta updates, 50 GiB scenarios:

| Path | docs | flowctl peak RSS | fixture log on disk |
|---|---|---|---|
| V1 `preview` | 1 KB | 0.34 GB | n/a |
| V1 `preview` | 1 MB | 0.49 GB | n/a |
| V2 `preview-next`, 1 shard | 1 MB | 9.5 GB | ~4 GiB per 5 GiB round |
| V2 `preview-next`, 1 shard | 1 KB | 21.7 GB | ~4 GiB per 5 GiB round |
| V2 `preview-next`, 4 shards, 200 GiB | 1 MB | 15.9 GB | ~160 GiB total |

Identical work, same FIFO, same generator, same connector drain rate — only the reader differs.

The growth curve confirms an unthrottled reader. On a 32 GB host, flowctl held 13.1 GB at t+60s while the connector — paced at 20 MB/s — could only have consumed ~1.2 GB in that window, then continued: 17.2, 19.7, 26.2, 28.2 GB at 30-second intervals until death. V1 under an equally slow consumer (26 MiB/s) stays flat at 0.34 GB.

## Proposed fix

Give the fixture writer a backlog gate equivalent to `LogActor`'s: track sealed-segment bytes, stop appending above `shuffle_disk_limit_bytes`, resume at half. This should bound the disk growth directly and, by recoupling the writer to consumer progress, likely bound the memory growth too.

## Downstream impact (benchmarking)

- 200 GiB scenarios need ~160 GiB of disk. Runs died with `writing fixture log block: No space left on device` on the default 100 GB boot disk.
- 1 KB-document arms at 5 GiB transactions need 64 GB RAM. On 32 GB, the Snowpipe SDK refuses to write once **system** memory passes 90%, raising `MemoryThresholdExceeded` served as **HTTP 429** — while the sidecar's own process held only 109 MB (16%). This reads as a Snowflake rate limit and is the opposite of the truth; genuine rate-limit 429s are `PER_CHANNEL_UNCOMPRESSED_BYTES_RATE_LIMIT_EXCEEDED` only.

## Open questions for whoever picks this up

1. Which allocation actually holds the 21.7 GB. Unmeasured — a heap profile would settle it, and would confirm whether the backlog gate alone is sufficient or only necessary.
2. Why 1 KB documents cost 2–3x what 1 MB documents cost for the same total bytes. Per-document `BlockMeta` plus packed key plus archived-node overhead is the obvious candidate, but this is observed, not explained.

Source citations verified against `196b09cd06`.

jacobmarble:
**Description:**

Closes #3399

`flowctl raw preview-next --fixture` runs a task locally against a file (or FIFO) of documents instead of live collection data. It grew without limit in both memory and disk — 21.7 GB of RAM on benchmark runs, and ~160 GiB of scratch files for a 200 GiB scenario — where V1 `flowctl preview` stays flat on the same work.

There were two independent causes:

1. **Memory.** The fixture feeder collected every document of a transaction into a list of parsed JSON values, and wrote them out only when it hit the `{"commit": true}` marker. So peak memory tracked *transaction* size: about 2x the source bytes, and 4x for small documents, because each parsed value carries fixed overhead. It now writes each document as it parses it, so it holds at most one block per shard.

2. **Disk.** The feeder wrote log segments as fast as its producer fed it, whether or not the consumer had read them. In production `LogActor` stops appending once unread segments pass a byte limit, and resumes as the consumer reads and deletes them; the feeder consulted no such limit. It now does the same, against a new 512 MiB budget.

The issue proposed the disk limit as the fix for both. It isn't — holding transaction size fixed and starving the connector to 20 MB/s tripled disk use and left peak memory untouched.

**Workflow steps:**

Nothing changes in how you use it: same flags, same output. The Store requests the connector receives are byte-identical to before.

Two things you may notice. A slow connector now paces the fixture's producer rather than letting it race ahead. And scratch log files are LZ4-compressed as they age, which this path never did.

None.

**Notes for reviewers:**

- Two commits, best read in order: the first adds two tests that **fail**, the second makes them pass.
- The pause only happens between transactions. A transaction's blocks sit past the last read barrier the consumer was given, so it can't read them until that transaction commits — pausing part-way through would wait on something that can never happen. The bound is therefore the limit plus one transaction.
- 512 MiB rather than the 2 GiB data-plane default preview inherits today: that default is per-shard, so `--shards 4` would reserve 8 GiB of a local disk.
- `write_transaction` survives as a thin wrapper over the new pieces, so its other caller is unchanged.
- Measured before/after (1 KB documents, one shard): one 2 GiB transaction, 3728 → 483 MiB peak RSS; 8 MiB transactions with the connector at 20 MB/s, 1011 → 486 MiB of log on disk. 1 GiB against a fast connector takes 10.5s either way.

### DIFF
--- a/crates/flowctl/src/raw/preview_next/fixture.rs
+++ b/crates/flowctl/src/raw/preview_next/fixture.rs
@@ -1125,6 +1125,273 @@ mod test {
         feeder.await.unwrap().unwrap();
     }
 
+    /// A Task with one materialization binding on `acmeCo/events`, keyed on
+    /// `/id`: fixture documents for that collection are routed and written, so
+    /// tests can observe what the feeder actually puts on disk.
+    fn one_binding_task() -> shuffle::proto::Task {
+        let collection = proto_flow::flow::CollectionSpec {
+            name: "acmeCo/events".to_string(),
+            key: vec!["/id".to_string()],
+            uuid_ptr: "/_meta/uuid".to_string(),
+            write_schema_json: r#"{
+                "type": "object",
+                "required": ["id"],
+                "properties": {"id": {"type": "integer"}, "pad": {"type": "string"}}
+            }"#
+            .into(),
+            partition_template: Some(proto_gazette::broker::JournalSpec {
+                name: "acmeCo/events".to_string(),
+                ..Default::default()
+            }),
+            ..Default::default()
+        };
+
+        shuffle::proto::Task {
+            task: Some(shuffle::proto::task::Task::Materialization(
+                proto_flow::flow::MaterializationSpec {
+                    name: "acmeCo/sink".to_string(),
+                    bindings: vec![proto_flow::flow::materialization_spec::Binding {
+                        collection: Some(collection),
+                        partition_selector: Some(Default::default()),
+                        ..Default::default()
+                    }],
+                    ..Default::default()
+                },
+            )),
+        }
+    }
+
+    /// Total bytes of log segment files living under `dir`. Nothing unlinks
+    /// them in these tests (no consumer reads), so this is the writer's
+    /// standing on-disk backlog.
+    fn live_segment_bytes(dir: &std::path::Path) -> u64 {
+        let mut total = 0;
+        for entry in std::fs::read_dir(dir).unwrap() {
+            let entry = entry.unwrap();
+            if entry.path().extension().is_some_and(|e| e == "flog") {
+                total += entry.metadata().unwrap().len();
+            }
+        }
+        total
+    }
+
+    /// One fixture line: a document of roughly `pad` bytes of padding.
+    ///
+    /// The padding is incompressible. A gated writer will reclaim disk by
+    /// LZ4-compressing sealed segments as well as by the consumer unlinking
+    /// them, so compressible padding would let compression alone drain the
+    /// backlog — and let `streaming_writer_respects_the_shuffle_disk_limit`
+    /// pass without the gate ever engaging.
+    fn fixture_doc(id: usize, pad: usize) -> String {
+        const ALPHABET: &[u8; 64] =
+            b"ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+-";
+
+        // A plain LCG: deterministic across runs, and cheap enough to generate
+        // hundreds of MiB of padding.
+        let mut rng = id as u64 | 1;
+        let padding: String = (0..pad)
+            .map(|_| {
+                rng = rng
+                    .wrapping_mul(6364136223846793005)
+                    .wrapping_add(1442695040888963407);
+                ALPHABET[(rng >> 58) as usize & 63] as char
+            })
+            .collect();
+
+        format!("[\"acmeCo/events\",{{\"id\":{id},\"pad\":\"{padding}\"}}]\n")
+    }
+
+    /// The fixture writer must not accumulate an unbounded on-disk backlog.
+    ///
+    /// `LogActor` gates every append on `shuffle_disk_limit_bytes` worth of
+    /// sealed segments (`crates/shuffle/src/log/actor.rs`), which are reclaimed
+    /// only as the consumer unlinks them while reading — so in production the
+    /// writer is throttled by consumer progress. The fixture calls
+    /// `Writer::append_block` directly and consults no such gate, so with no
+    /// consumer it writes the whole fixture to disk.
+    ///
+    /// Fed through a FIFO, a gated writer stops draining the pipe and
+    /// back-pressures to this test's own writes; that is success.
+    ///
+    /// The asserted bound is the gate's true guarantee, which has three slack
+    /// terms past the limit itself:
+    ///
+    /// - one segment: the backlog is accounted in whole sealed segments, so
+    ///   engagement overshoots by up to the segment roll threshold;
+    /// - one segment: `live_segment_bytes` also counts the active (unsealed)
+    ///   segment, which the backlog accounting never sees;
+    /// - one transaction: the gate is taken only at transaction boundaries —
+    ///   a transaction's blocks lie beyond the last relayed frontier's
+    ///   `flushed_lsn`, so the consumer cannot read, and so cannot reclaim,
+    ///   them until that transaction commits.
+    ///
+    /// The limit inherited by preview is the data-plane-wide default (2 GiB,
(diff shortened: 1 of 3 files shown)


jacobmarble:
## Measurements

All from a local harness: a minimal `local:` Python materialization that discards every Store (rate-limitable), plus a generator whose transaction size and total volume are set independently. "before" is `88629ced90`, "after" is this branch. Single shard, 1 KB documents, unless stated.

### The headline pair

| scenario | metric | before | after |
|---|---|---|---|
| 2 GiB in one transaction | peak RSS | 3728 MiB | **483 MiB** |
| 2 GiB in 8 MiB transactions | peak RSS | 256 MiB | 253 MiB |
| 1 GiB in 8 MiB txns, connector at 20 MB/s | peak log on disk | 1011 MiB | **486 MiB** |
| 1 GiB in 8 MiB txns, connector unpaced | peak log on disk | 316 MiB | 289 MiB |

### Why the disk limit alone would not have fixed memory

Transaction size held at 8 MiB, 1 GiB total, only the connector's drain rate varied. Disk triples; memory does not move.

| connector | peak log on disk | peak RSS |
|---|---|---|
| unpaced | 316 MiB | 272 MiB |
| 20 MB/s | 1011 MiB | 272 MiB |

### What memory actually tracked (before)

Total volume held at 2 GiB, transaction size varied — peak RSS scales with the transaction, not the run.

| transaction size | peak RSS | ratio to source bytes |
|---|---|---|
| 128 MiB | 463 MiB | 3.6x |
| 256 MiB | 830 MiB | 3.2x |
| 512 MiB | 1274 MiB | 2.5x |
| 1 GiB | 2025 MiB | 2.0x |
| 2 GiB | 3861 MiB | 1.9x |

And per document, at a fixed 1 GiB transaction — smaller documents cost more per byte, which is the fixed overhead of each parsed JSON value. This answers open question 2 on #3399.

| document size | peak RSS | ratio to source bytes |
|---|---|---|
| 256 B | 4099 MiB | 4.0x |
| 1 KB | 2025 MiB | 2.0x |
| 64 KB | 1476 MiB | 1.4x |
| 1 MB | 1249 MiB | 1.2x |

### Cost and equivalence

| check | result |
|---|---|
| 1 GiB, fast connector, wall clock | 10.4 / 10.5s before, 10.5 / 10.7s after |
| Store requests reaching the connector | byte-identical at 1, 2 and 4 shards |

The equivalence check ran a fixture with a multi-block transaction, an empty transaction, and a trailing transaction with no commit marker, and compared a hash of every Store request the connector received.

jacobmarble:
Benchmarked `bb3bf9b5df` end to end: materialize-snowflake, `--shards 4`, 40 transactions of 5 GiB, 1 MB documents, `c3-highmem-8` / 400 GB pd-ssd. Full 200 GiB, exit 0, no panics.

| | `#3391` only | this PR | |
|---|---|---|---|
| Peak flowctl RSS | 15.85 GB | **5.86 GB** | 2.7x less |
| Disk consumed | ~138 GB | **10.8 GB** | 12.8x less |
| Steady MiB/s | 55.838 | **65.039** | +16.5% |
| Steady MB/s | 58.55 | **68.20** | +16.5% |
| Wall | 3,601 s | **3,097 s** | −504 s |
| Rate-limit 429s | 0 | 0 | — |
| Pacer utilisation | 73.2% | 85.2% | — |

Memory and disk are bounded across all 40 transactions, not just early ones. It also got **faster**, which I did not expect from a scheduling change — bounding the backlog seems to stop the feeder thrashing disk it never needed.

The earlier `disk_backlog_bytes underflow` is gone; `bb3bf9b5df` fixed it. Per-transaction spread also tightened (40.95–69.51 MiB/s, median 66.12, against 29.66–69.41 before).

jgraettinger (approved):

PR #3410, opened 2026-08-26
flowctl: preview-next no longer hangs when its fixture feeder dies

### LINKED ISSUE #3409: flowctl: preview-next hangs forever when its fixture feeder dies
### What happens

When `flowctl raw preview-next --fixture` loses its fixture feeder, the process does not exit. It blocks forever waiting for documents that can never arrive.

The feeder runs as a tokio task. A panic there unwinds that worker without aborting the process, and closes the read end of the fixture FIFO. The generator on the other end then dies with `BrokenPipeError`. `preview-next` itself stays alive, logging `waiting for documents round=N` on a five-minute cadence, at ~1% CPU, indefinitely. There is no timeout and no non-zero exit.

Observed most recently on a 200 GiB benchmark: the run hung for **2h12m** before I noticed and killed it. Unattended — CI, a scheduled benchmark, an agent-driven run — it would hang until something external reaped it.

```
thread 'tokio-runtime-worker' (23078) panicked at
  crates/flowctl/src/raw/preview_next/fixture.rs:658:10:
disk_backlog_bytes underflow
...
2026-08-26T18:28:33.937170Z  INFO ops: waiting for documents fields={"round":3}
2026-08-26T18:28:33.977577Z  INFO ops: waiting for documents fields={"round":3}
```

### This is not specific to one panic

The same hang is documented in #3391's own description, from an entirely different cause — the `block::encode` 65,536-entry assert:

> flowctl panicked during the first transaction. That killed the reader of the fixture pipe, so the generator died with a broken pipe and the run hung.

So it has now cost real time at least twice, from two unrelated defects. Any future feeder failure produces it again. The underlying panics get fixed one at a time; this is the property that turns each of them from a fast, legible crash into a silent multi-hour stall.

### Expected

A dead fixture feeder should terminate `preview-next` with a non-zero exit and a diagnostic naming the cause. Options, roughly in order of preference:

1. Have the feeder task report its termination to the supervising task, and fail the run when it ends before EOF. This is the honest fix: an error propagates instead of a liveness property being inferred.
2. Detect the closed FIFO write end and treat premature EOF as an error rather than as "no documents yet".
3. As a backstop, bound `waiting for documents` — if no document arrives within some generous interval and no writer holds the pipe, exit non-zero.

A panic-to-abort policy for the feeder task would also work, though it converts a diagnosable error into a hard abort and loses the surrounding context.

### Repro

Any panic inside the fixture feeder while a generator writes to the FIFO. The most direct is the one in #3408: run a `--shards 4` fixture workload of several 5 GiB transactions with high-entropy 1 MB documents, which over-credits the disk backlog counter until it underflows.

### Notes

Found while benchmarking `materialize-snowflake` Snowpipe Streaming v2 (estuary/connectors#4828). Filed separately from #3408 at review time, since the panic there is one instance rather than the cause.

jacobmarble:
**Description:**

Fixes #3409.

`flowctl raw preview-next --fixture` can read its fixture from a FIFO or stdin. A background task — the "feeder" — reads lines off the pipe and hands each committed transaction to the consumer. If that feeder panics, the run used to hang forever instead of failing.

Two things combined to hide the death:

1. A panic in a tokio task unwinds only that task. The process keeps running, and the feeder's cleanup code — the part that signals "no more transactions" — is skipped.
2. The consumer waits on a channel for the next transaction. `services::Run` holds its own copy of the sending end, so the channel never closes. Waiting on a dead feeder looks exactly like waiting on a slow one.

So the run sat at ~1% CPU with no timeout and no exit. It cost 2h12m on one benchmark, and the same hang has now happened from two unrelated panics.

The fix supervises the feeder instead of inferring that it is alive. The feeder holds its log writers open until the run ends, so it outlives the session loop on every ordinary path — including a bad fixture line, which still hands over a transaction boundary first. So if the feeder finishes early, it died.

We now run the session loop and the feeder's task handle side by side. Whichever ends first decides what happens. Feeder first: stop the session, let it drain, and fail the run naming the panic. Session loop first: release the writers and join the feeder, which surfaces a stream error that would otherwise look like a clean stop.

**Workflow steps:**

No user-facing change when things go well. When a fixture feeder panics, `preview-next` now exits non-zero with `fixture feeder panic: ...` instead of hanging.

None.

**Notes for reviewers:**

- Three commits. The first adds the test and is **red**; the second makes it green; the third reworks the fix after review (see the comment below). Check out the first commit alone if you want to watch it hang.
- No fixture input can make the feeder panic today, so the test spawns a panicking task in its place and checks that the supervisor fails the run.

### DIFF
--- a/crates/flowctl/src/raw/preview_next/fixture.rs
+++ b/crates/flowctl/src/raw/preview_next/fixture.rs
@@ -1125,6 +1125,88 @@ mod test {
         feeder.await.unwrap().unwrap();
     }
 
+    /// A feeder that dies before EOF must fail the run, not stall it.
+    ///
+    /// A panic inside the feeder unwinds only its tokio worker. It sends no
+    /// `Boundary`, never cancels `eof_stop`, and never releases `hold` — while
+    /// `services::Run` still retains a `frontier_tx` clone, so the channel stays
+    /// open. The consumer therefore cannot distinguish a dead feeder from a slow
+    /// producer, and parks on `recv_checkpoint` until something external reaps
+    /// the process. Aborting the task reproduces exactly that state.
+    #[cfg(unix)]
+    #[tokio::test]
+    async fn test_streaming_feeder_death_is_not_a_hang() {
+        use tokio::io::AsyncWriteExt;
+
+        let tmp = tempfile::tempdir().unwrap();
+        let path = tmp.path().join("fifo");
+        assert!(
+            std::process::Command::new("mkfifo")
+                .arg(&path)
+                .status()
+                .unwrap()
+                .success()
+        );
+
+        let (opener, frontier_tx) = fixture_opener();
+        // `services::Run` holds a sender for the life of the run, so a dead
+        // feeder does not close the channel.
+        let _run_tx = frontier_tx.clone();
+
+        let eof_stop = tokio_util::sync::CancellationToken::new();
+        let hold = tokio_util::sync::CancellationToken::new();
+
+        let (_dir, feeder) = start_streaming(
+            &empty_task(),
+            Some(path.clone()),
+            tmp.path(),
+            1,
+            frontier_tx,
+            eof_stop.clone(),
+            hold.clone(),
+        )
+        .unwrap();
+
+        // The producer holds the pipe open for the whole test, as a benchmark
+        // generator does: the feeder's death is the only thing that changes.
+        let mut pipe = tokio::fs::OpenOptions::new()
+            .write(true)
+            .open(&path)
+            .await
+            .unwrap();
+        pipe.write_all(b"[\"a/coll\", {\"k\": 1}]\n{\"commit\": true}\n")
+            .await
+            .unwrap();
+        pipe.flush().await.unwrap();
+
+        let mut src = opener
+            .open(Default::default(), Vec::new(), shuffle::Frontier::default())
+            .await
+            .unwrap();
+        src.request_checkpoint();
+        src.recv_checkpoint().await.unwrap();
+
+        feeder.abort();
+        let _ = feeder.await;
+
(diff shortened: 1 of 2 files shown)


jacobmarble:
Good call — taken in 8fb7a14266.

On `try_join_all` specifically: it does not fit, for a concrete reason. It waits for *every* future on the success path, and the feeder never finishes on that path — it holds its shuffle-log writers open until the run ends, because the consumer unlinks segment files as it reads them. So it would deadlock.

You can remove that hold by hoisting the writers into the caller and letting the feeder borrow them. But a borrowing future is not `'static`, so it cannot be spawned at all. That means the feeder and the consumer share one task, and stop running in parallel. `write_transaction` is synchronous and heavy — parse, allocate, archive, validate, hash, per document — and this path exists mainly to drive throughput benchmarks. Wrong trade.

So I took the middle: join the handle, keep the task.

The invariant that makes it clean is the one above — the feeder outlives the session loop on every ordinary path, a malformed fixture line included, since that still hands over a `Boundary` first. So the feeder completing early *means* it died. That is now selected on directly:

- Feeder ends first: cancel the session stop, drain the loop, report its panic or error.
- Session loop ends first: release the keepalive and join, as `finish_fixtures` did.

This subsumes `finish_fixtures`, which only ever joined the handle after a session loop that a dead feeder never let return. The drop guard is gone. It also catches a feeder that returns early *without* panicking, which the guard could not tell apart from a graceful end.

One cost worth flagging: the regression test moved. It used to drive a real FIFO in `fixture.rs` and assert on the stop token. Detection now lives in `mod.rs`, so the test drives the supervisor with a panicking task instead. It covers the supervision honestly, but no longer exercises the real feeder end to end.

jgraettinger (approved):

PR #3411, opened 2026-08-26
Docs: Point OpenMetrics API endpoint at api.estuary.dev

skord:
The OpenMetrics page gave customers the bare Cloud Run hostname
(`agent-api-1084703453822.us-central1.run.app`) as their scrape endpoint. Traffic to
that hostname bypasses the Cloud Armor WAF, which only inspects requests arriving
through the load balancer at `api.estuary.dev`.

Swaps all four occurrences to `api.estuary.dev`: the stated endpoint (line 15),
the Prometheus `static_configs.targets` (46), the Datadog `openmetrics_endpoint` (71),
and the `curl` example (109). No other `run.app` references remain under `site/`.

Both hostnames return `401` unauthenticated on `/api/v1/metrics/acmeCo/`, so
`api.estuary.dev` routes to the same service on the same path.

This is the `site/` twin of estuary/docs#49, which made the identical change to
`content/features/openmetrics-api.md`. The two files were byte-identical beforehand.

Refs estuary/security#776

### DIFF
--- a/site/docs/features/openmetrics-api.md
+++ b/site/docs/features/openmetrics-api.md
@@ -12,7 +12,7 @@ Integrating the API with monitoring platforms like Prometheus or Datadog also al
 
 The OpenMetrics API consists of one main endpoint:
 
-```https://agent-api-1084703453822.us-central1.run.app/api/v1/metrics/{prefix}/```
+```https://api.estuary.dev/api/v1/metrics/{prefix}/```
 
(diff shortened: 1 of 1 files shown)


skord:
I've updated this here even with the deprecation so anyone who refers to this in other parts of this repository doesn't get incorrect information.

aeluce (approved):

PR #3412, opened 2026-08-26
runtime-next: yield from the materialize actor's scan loop

dgreer-dev:
The shard's Actor::serve drives Scanning and Draining with a `continue` that never awaits, and a frontier scan of an append-only binding emits no C:Load at all (load optimization skips every new key), so a large transaction ran the entire scan as one synchronous loop on the task's single-worker tokio runtime.

The hyper h2 connection to connector-init lives on that same runtime with a 5s keep-alive ping and hyper's default 20s timeout. hyper checks the ping timer before reading pending frames, so once the scan outlasted 20s the connection was torn down on the actor's next yield and every open stream failed with "h2 protocol error: error reading a body from connection", surfaced as "Materialize error (expected connector response) from connector" roughly 30s into the transaction.

Yield every 64 loop iterations (one shuffle-log block each), as the capture and derive drain loops already do, and set an explicit 60s keep-alive timeout on the connector-init channel so a shorter stall is not misread as a dead peer.

### DIFF
--- a/crates/runtime-next/src/container.rs
+++ b/crates/runtime-next/src/container.rs
@@ -278,6 +278,10 @@ pub async fn start<L: crate::Logger>(
         .expect("formatting endpoint address")
         .connect_timeout(std::time::Duration::from_secs(5))
         .http2_keep_alive_interval(std::time::Duration::from_secs(5))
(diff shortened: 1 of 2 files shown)


jgraettinger (approved):
LGTM

PR #3415, opened 2026-08-27
Housekeeping and prep for the catalog-test harness

jgraettinger:
**Description:**

Stack 1 of 5, replacing the V1 catalog-test machinery with a Rust harness. This one clears the decks: three commits delete dead scaffolding, two are small correctness fixes that the rest of the stack depends on.

**Read these two:**

* `examples: shuffle derivations on the key their reductions require` — `shuffle: any` compiles to an empty shuffle key, which both runtimes resolve to the *source* collection key. A derivation with an order-dependent reduction must shuffle on a key that pins each of its output keys to a single shard, which the source key does not in these three cases. `citi-bike/stations` flakes ~35% of runs under multi-shard catalog tests; `last-seen` and `stock/daily-stats` carry the same defect but can't expose it today.
* `doc: compare numbers of differing representation under an epsilon` — `doc::diff` compared floats under a magnitude-scaled epsilon only when *both* sides were floats. An integer against a float fell through to exact comparison, which orders `0` before `0.0` — a distinction JSON Schema does not draw. Two integers still compare exactly.

**Skim these three:**

* `mise: pin usage 6.4.0` — routine pin refresh, three majors behind. No task's `#USAGE` spec changes.
* `plans: retire plans/ directory` — point-in-time design docs for completed or superseded work, plus the two READMEs that cited them.
* `tests: retire the dead shell end-to-end suite` — nothing in `.github/` references `tests/run-all.sh`, so these catalogs, row fixtures, and the SSH-forwarding docker-compose scaffolding have been dead weight. **The sshforwarding fixtures carried committed SSH private keys.** `flowctl-go api await` existed solely to serve `run-end-to-end.sh` and goes with it.

**Workflow steps:**

No user-facing change.

None.

**Notes for reviewers:**

`ghcr.io/estuary/source-test:dev` remains referenced by `go/bindings/testdata/build.flow.yaml` and `crates/shuffle/tests/*.flow.yaml`, which are unaffected by removing `tests/source-test/`.

The examples change lands here, ahead of the stack, because PR 3 switches CI's `ci:catalog-test` to the multi-shard runtime-next runner — landing the spec fix first keeps every rung green. The one thing to watch on this PR is that the examples still pass under V1 `flowctl-go test`, which is what CI runs at this point in the stack.

---

1. #3415 — Housekeeping and prep ← **you are here**


williamhbaker (approved):
LGTM

PR #3416, opened 2026-08-27
runtime-local: extract the local task-drive layer from flowctl

jgraettinger:
**Description:**

Stack 2 of 5. `flowctl raw preview-next` and the forthcoming catalog-test runner need the same thing — run a runtime-next task locally with synthetic shards, no Gazette broker, no etcd, no Go consumer — and differ only in three runtime-next host seams. This moves that machinery into a new `runtime-local` crate rather than let a second copy grow beside it, migrating `flowctl` onto it in the same commit so no divergent copy ever exists.

Four commits:

* `tests: move preview snapshots to tests/preview` — re-points the `flowctl preview` snapshot tests at `raw preview-next`. The baseline the extraction is checked against.
* `labels: hoist even shard splitting into labels::shard::even_splits` — four places computed the same even split of the `u32` key space. `runtime-local` is the fifth caller, so the helper lands first.
* `runtime-next: shard handlers own their RocksDB directory` — a shard now owns and removes its own storage, which is what makes a dropped local run safe, plus the protocol changes that follow from it.
* `runtime-local: extract the local task-drive layer from flowctl` — the move itself.

**Workflow steps:**

`flowctl raw preview-next` behaves identically. `poetry run pytest tests/preview` runs the snapshot suite — it is not wired into CI, so it's a developer-run baseline; CI coverage for this stack arrives in #3417.

None. `crates/runtime-local/README.md` is new; `tests/README.md` becomes a roadmap over `preview/` and `soak/`, with the detailed instructions moving to `tests/preview/README.md`.

**Notes for reviewers:**

⚠️ **The RocksDB-ownership commit does not build on its own.** `flowctl raw preview-next` is broken against the new `SessionLoop` until the commit after it, which extracts that layer and adopts all of it. Teaching a soon-to-be-replaced copy semantics it will never use wasn't worth it, so the two are separable stories but not separable builds. Review them as a pair; the PR tip builds.

Ownership is the change with consequences outside this stack: a controller can no longer open a shard's RocksDB by path once its stream ends, so `--initial-state` and `--output-state` move onto the protocol, `Stopped` loses `Copy`, and Go's `removeRecoveryDir` becomes a no-op wherever a descriptor was sent.

The extraction is otherwise behavior-preserving. Three things worth a close look:

* **The two generalized seams** — `Controls` over its publisher and logger factories, and `build_shuffle` replacing `fixture: bool`. The latter is what keeps journal *authorization* in flowctl: reading live journals needs a logged-in user token, and only flowctl has one.
* **Clock policy stays with the caller**, because preview must keep reproducing the legacy harness's `3600 * ordinal + index` stamping while the test runner paces clocks against its own synthetic time.
* **A real defect fixed in passing** — `Run` documented that its server task aborts on drop, but a plain `JoinHandle` *detaches* rather than aborts, so the tonic server outlived every `Run`.

---

2. #3416 — runtime-local: extract the local task-drive layer ← **you are here**


crates/runtime-local/src/segments.rs
```diff
@@ -0,0 +1,739 @@
...
+
+    let flushed_lsn: Vec<u64> = writers.iter().map(|w| w.last_lsn.as_u64()).collect();
+    shuffle::Frontier::new(journals, flushed_lsn).context("building fixture checkpoint frontier")
+}
+
+/// Write one whole binding-tagged transaction and return its checkpoint
+/// frontier: a [`TxnState::new`] -> [`push_binding`] -> [`finish_txn`] sequence
+/// over `items`, each of which carries the binding it feeds and its clock.
+#[allow(clippy::too_many_arguments)]
+pub fn write_transaction_for_bindings(
```
jacobmarble, line 573: This must be called by a subsequent change in the stack, I can't find any call to it in this change.

jacobmarble (dismissed):

jgraettinger:
@jacobmarble PTAL

jacobmarble (approved):

PR #3417, opened 2026-08-27
catalog-tests: the catalog test harness, and `flowctl raw test`

jgraettinger:
**Description:**

Stack 3 of 5, and the core of it: a new `catalog-tests` crate, layered above `runtime-local` (#3416) and `runtime-next`. Plus the one protocol addition it needs, and the CLI that drives it.

Five commits:

* `proto: add derive Reset / ResetDone messages` — the mechanism the harness uses to isolate test cases from one another.
* `runtime-next: forward Reset to derive connectors` — handled shard-locally; the leader is not involved.
* `derive-sqlite: report no checkpoint for an ephemeral database` — a `:memory:` database is not remote-authoritative, and saying so is what lets the harness run derive-sqlite multi-shard.
* `catalog-tests: the catalog test harness` — the crate.
* `flowctl: add the raw test subcommand` — and CI's `ci:catalog-test` task now uses it instead of `flowctl-go test`.

**Workflow steps:**

```
flowctl raw test --source examples/flow.yaml
```

A line per test, keyed by test name. A failure names the source path and JSON pointer of the step it failed at and indents the rendered diff beneath; a case the run never reached says so. Then a tally. Exits non-zero if any test failed.

None user-facing. `crates/catalog-tests/README.md` is new and carries the design.

**Notes for reviewers:**

---

3. #3417 — catalog-tests: the harness, and `flowctl raw test` ← **you are here**


crates/catalog-tests/src/scheduler.rs
```diff
@@ -0,0 +1,336 @@
...
+/// On failure the graph is still quiesced before returning: a case which fails
+/// partway leaves reads pending (and possibly delayed into the future), and
+/// those must not leak into the next case — where they would feed a just-reset
+/// connector and write inside that case's verify window. The step failure is
+/// the error returned; a failure while quiescing is appended to it.
+pub async fn run_test_case<D: Driver>(
+    graph: &mut Graph,
+    driver: &mut D,
+    test: &TestSpec,
+) -> anyhow::Result<String> {
```
williamhbaker, line 55: nit: The comment says that the scope of the last step reached is returned and used for error reporting, but as I understand it the error reporting part happens on the error path, and the string value from the `Ok` result is never used... in `crates/catalog-tests/src/run.rs:192` it is discarded. It seems like `LiveDriver::last_scope` is actually doing the tracking of error scopes.

williamhbaker (approved):
LGTM

crates/catalog-tests/src/scheduler.rs
jgraettinger: Thanks, I'll address in a next PR.

PR #3418, opened 2026-08-27
control-plane-api: run publication tests on catalog-tests

jgraettinger:
**Description:**

Stack 4 of 5, one commit. Publication tests now call `catalog_tests::run_tests` in-process, against the `build::Output` the publication already produced.

**Workflow steps:**

No change to how a user runs tests — they still publish a draft and see test results in publication logs. What changes is where those results come from and how precisely failures are reported.

None.

**Notes for reviewers:**

Most of the surface is deletion. One behavior changes:

* **Failures anchor precisely.** Each failing case becomes a `tables::Error` scoped to the failing *step's* source URL and JSON pointer, where before a whole run collapsed into one error at `flow://publication/test/api/test` with "View logs for details". A case the run never reached is reported as an error too, so a publication never silently understates its test coverage.

---

4. #3418 — control-plane-api: run publication tests on catalog-tests ← **you are here**

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2130,6 +2130,7 @@ dependencies = [
  "billing-types",
  "build",
  "bytes",
+ "catalog-tests",
  "chrono",
  "clap",
  "colored_json",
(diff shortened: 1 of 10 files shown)


bbartman (commented):
I believe this looks OK. I don't see any problems with it.

bbartman (approved):
Sorry I missed a button.

PR #3419, opened 2026-08-27
Remove the V1 catalog-test machinery

jgraettinger:
**Description:**

Stack 5 of 5, and pure deletion. Nothing drives the V1 catalog-test machinery any more: CI runs on `catalog-tests` as of #3417, and publication tests as of #3418.

Two commits, kept separate so ~2,950 lines of proto regen don't bury the hand-written Go deletions:

* `go: remove the V1 catalog-test machinery`
* `proto: remove the Testing RPC service`

**Workflow steps:**

`flowctl-go test`, `flowctl-go api test`, and `flowctl-go temp-data-plane` no longer exist. Use `flowctl raw test --source ...` (#3417).

None.

**Notes for reviewers:**

Read the first commit; the second is almost entirely regenerated code.

**One flag worth a second look.** `FlowConsumerConfig.Flow.TestAPIs` / `--flow.test-apis` is removed along with the `pf.RegisterTestingServer` block it gated. It also gated something non-obvious: control-plane construction was *skipped* when it was set, because a temp data plane has no control plane to talk to. Only temp-data-plane ever passed the flag, so for every real consumer that term was already true and dropping it changes nothing.

The `Testing` service's three RPCs are all superseded — ResetState by `derive.Request.Reset`, AdvanceTime by a counter inside `catalog-tests`, and Ingest by a combine-by-key and an append to an in-memory collection store. `derive.Request.Reset` itself stays in the connector protocol: it is the mechanism the V2 flow now drives.

---

5. #3419 — Remove the V1 catalog-test machinery ← **you are here**


williamhbaker (dismissed):
LGTM

williamhbaker (approved):

Issue #3376, opened 2026-08-18
Capability-masked tokens


GregorShear:
Add a new grant type to `POST /api/v1/auth/token` that mints an access token limited to a selected set of capabilities. The new token identifies the requesting user (`sub`, `email`) just like a normal token, but also carries a new optional `capability_mask` claim on `ControlClaims`: a flat list of `models::authz::Capability` bits representing what the user has selected for an agent to use on their behalf. Tokens without this claim keep full authority, so existing tokens are unaffected.

## Request

```json
{
  "grant_type": "capability_token",
  "capability_mask": ["CatalogRead", "JournalRead"],
  "upgrade_token": "<optional previously-minted reduced token>"
}
```

The caller authenticates with a normal bearer access token. An empty mask is valid and mints an identity-only token.

## Rules

1. The new token's mask is the list in the request. If an `upgrade_token` is included, its mask is added to the list.
2. The `upgrade_token` must have a valid signature and must belong to the same user who is making the request. It may be expired — presenting an expired reduced token alongside a valid credential re-mints it with a fresh expiry, without re-running user approval.

The mask may list capabilities the user doesn't hold — those bits are inert, since enforcement intersects with real grants at use time. 

## Enforcement

When a reduced token is used, authorization filters out any capability that is not in the mask while walking the user's grants. So the token holder can only do things that both the user's grants and the mask allow. Tokens without the claim skip the filter. The grant walk lives in `tables::UserGrant::is_authorized`; handlers get verified claims from `Envelope::claims()`.

## Open questions

- The agent should only ever hold one of these masked tokens, but how will we prevent the agent from calling this token exchange endpoint and widening its own mask?

bbartman:
## Implementation plan (revised 2026-08-25: enforcement reworked around an axum `Authority` extractor; endpoint census and PR grouping added)

[...]

PR #3422, opened 2026-08-27
control-plane-api, tables: legacy decisions migrate to effective capability bits

### LINKED ISSUE #3232: storageMappings query fails for bundle-only grants (legacy-capability mismatch)
## Summary

The `storageMappings` GraphQL query fails an entire request when the caller holds a **bundle-only** grant — one whose legacy `capability` is `None` and whose capability comes solely from `bundles`.

## Cause

The resolver authorizes and displays capability through two functions that disagree about bundles:

- [`authorized_prefixes`](crates/control-plane-api/src/server/public/graphql/authorized_prefixes.rs) selects prefixes with `bits.is_superset(min_bits)`, where the bits come from `effective_bits(legacy, bundles)`. A bundle-only grant **qualifies**, so its rows are fetched.
- Per row, the resolver calls [`tables::UserGrant::get_user_capability`](crates/tables/src/behaviors.rs#L174), which maps each node to its `.legacy` capability, drops `None`, and takes the max. For a bundle-only grant it returns `None`, and the resolver's `.ok_or_else(...)` at [storage_mappings.rs](crates/control-plane-api/src/server/public/graphql/storage_mappings.rs) converts that into a hard error: `missing capability for catalog prefix '…'`.

A row readable only through a bundle-only grant therefore aborts the whole query instead of being returned.

## Blast radius

Deprecating the `by` filter (#3200) makes this much easier to trigger. In unfiltered mode, a single bundle-only grant anywhere in the caller's grant graph pulls matching rows into the result set and fails their entire listing — not just the affected row.

## Scope

The failing combination (`authorized_prefixes` + a `get_user_capability` that hard-errors on `None`) is currently unique to `storageMappings`. The root-cause function `get_user_capability` — legacy-only, `None`-dropping — is also consumed at [server/mod.rs:153](crates/control-plane-api/src/server/mod.rs#L153); that path should be reviewed for the same divergence. A proper fix reports bundle-aware capability rather than the legacy enum, so it likely spans more than this one resolver.

## Fix options

- Expose `capabilityBits` (bundle-aware) and stop treating an absent legacy capability as an error.
- Or, minimally: since `authorized_prefixes` already established read access to the row's prefix, degrade gracefully instead of failing the request when the legacy capability is `None`.

## Gate

Only reachable once bundle-only **user** grants can actually exist. If they cannot yet, this is a latent bug to fix before that lands.

_Filed from #3200 review (@jshearer)._

bbartman:
Task 4 of #3376 (stacked on #3406): every remaining request-path authorization and visibility decision that read the legacy `read`/`write`/`admin` label now keys on mask-attenuated effective capability bits. The legacy label is reporting metadata only, per decision 3 of the [implementation plan](https://github.com/estuary/flow/issues/3376#issuecomment-5342195183).

Closes #3232.

## What changes

- **`tables::UserGrant::get_user_authorization`** is a new single-name walk reduction returning both halves of a name's authority: effective bits accumulated additively across covering nodes (the decision input), and the max un-attenuated legacy label (compatibility metadata, `None` under bundles-only coverage). It supersedes `get_user_capability`, which is removed; the label semantics are pinned by the converted tests.
- **`LiveSpecRef`**: the six gated sub-resolvers (`liveSpec`, `activeAlerts`, `alertHistory`, `status`, `lastPublication`, `publicationHistory`) and the `writtenBy`/`readBy` minimum-capability pagination filter (previously an `Ord` comparison on the legacy enum) decide on effective bits. `userCapability` is null iff the bits fall short of Viewer; otherwise it reports the literal legacy column, with `none` standing in for bundles-only coverage.
- **`storageMappings`** no longer errors the whole query when a row's covering grants carry no legacy label (the #3232 bug); the row serves with `userCapability: none`.
- **`dataPlanes`** no longer panics on the same shape (`.expect("capability guaranteed by pre-filter")` trusted the legacy label where the pre-filter guarantees bits).
- **Vocabulary normalization**: `alertConfigs` and `inviteLinks` spell their `filtered_authorized_prefixes` thresholds as `CapabilityBundle::Viewer` / `CapabilityBundle::Admin` — bit-identical to the legacy enums they replace.
- The hardcoded `Admin` literals in `createStorageMapping`/`updateStorageMapping` responses stay: the field is informational, the mutations require admin-level effective bits, and no consumer reads it (the only `userCapability` consumer anywhere is the dashboard's auth-roles store on `PrefixRef`, untouched here).

## Behavior changes

Thresholds are unchanged everywhere — only the decision input moves. The one deliberate behavior change beyond mask-correctness: **bundles-only grants** (`capability = 'none'`, non-empty `bundles`) now work through `liveSpecs`, `storageMappings`, and `dataPlanes` listings, consistent with the walk-based checks that already admit them. `PrefixRef` already reports these as `none`; `LiveSpecRef`/`StorageMapping`/`DataPlane` now do the same.

Explicitly excluded per the plan: `live_specs/db.rs`'s SQL-computed `user_capability` and `get_connected_live_specs` (async-executor-only; task 9 territory), and per-endpoint re-vocabulary (e.g. `CreateInviteLink` for invite links).

## Tests

- `tables`: the combined reduction (bundles-only, masked attenuation, un-attenuated label, additive multi-path composition), plus the converted legacy-label pins.
- `liveSpecs`: a bundles-only viewer lists and reads specs (labeled `none`); `writtenBy`/`readBy` admit bundles-only-covered names and withhold unauthorized names; a sub-Viewer mask is refused at the root with the structured missing-capabilities body, so legacy labels are never a side-channel around the mask.
- `storageMappings` / `dataPlanes`: bundles-only listings serve with `none` where they previously errored/panicked.
- Unmasked parity: all existing snapshots unchanged.

## Note

The 2026-08-26 Daily Sync recorded an open follow-up ("Discuss Legacy Bits", Brian with Dave and Greg) on exactly this gate-on-bits question. The written record — #3232, #2936's design comment defining `capability: none` for bundles-only grants, and the `PrefixRef` doc — all point the way this PR implements it; that follow-up is the venue to veto before merge.

### DIFF
--- a/crates/control-plane-api/src/server/mod.rs
+++ b/crates/control-plane-api/src/server/mod.rs
@@ -145,13 +145,15 @@ where
     Ok((None, ()))
 }
 
-/// Looks up the user's authorization grants for each item in
-/// `prefixes_or_names`, and calls the provided `attach` function with each
-/// item and its capability. The `Some` results are returned in a vec.
+/// Looks up the user's authorization for each item in `prefixes_or_names`,
+/// and calls the provided `attach` function with each item, its effective
+/// capability bits, and its legacy capability label. The `Some` results are
+/// returned in a vec.
 ///
-/// Grant reachability is evaluated under the bearer's capability `mask`;
-/// the legacy capability of a reached grant passes through un-attenuated
-/// (see `tables::UserGrant::get_user_capability`).
+/// The bits are mask-attenuated and are the only authorization decision
+/// input; the legacy label is un-attenuated compatibility metadata, `None`
+/// when coverage comes entirely from `bundles`-column grants (see
+/// `tables::UserGrant::get_user_authorization`).
 pub fn attach_user_capabilities<I, F, T>(
     snapshot: &Snapshot,
     claims: &crate::ControlClaims,
@@ -161,19 +163,19 @@ pub fn attach_user_capabilities<I, F, T>(
 ) -> Vec<T>
 where
     I: IntoIterator<Item = String>,
-    F: FnMut(String, Option<models::Capability>) -> Option<T>,
+    F: FnMut(String, models::authz::CapabilitySet, Option<models::Capability>) -> Option<T>,
 {
     prefixes_or_names
         .into_iter()
         .flat_map(|prefix| {
-            let capability = tables::UserGrant::get_user_capability(
+            let (bits, legacy) = tables::UserGrant::get_user_authorization(
                 &snapshot.role_grants,
                 &snapshot.user_grants,
                 claims.sub,
                 &prefix,
                 mask,
             );
-            attach(prefix, capability)
+            attach(prefix, bits, legacy)
         })
         .collect()
 }
--- a/crates/control-plane-api/src/server/mod.rs
+++ b/crates/control-plane-api/src/server/mod.rs
@@ -145,15 +145,9 @@ where
     Ok((None, ()))
 }
 
-/// Looks up the user's authorization for each item in `prefixes_or_names`,
-/// and calls the provided `attach` function with each item, its effective
-/// capability bits, and its legacy capability label. The `Some` results are
-/// returned in a vec.
-///
-/// The bits are mask-attenuated and are the only authorization decision
-/// input; the legacy label is un-attenuated compatibility metadata, `None`
-/// when coverage comes entirely from `bundles`-column grants (see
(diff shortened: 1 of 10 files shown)


bbartman:
I need someone with more knowledge of the change from `models::Capability::Read,` to `           models::authz::CapabilityBundle::Viewer,` because I don't know if these are the same thing or not. Claude seems to think they are.

Issue #3372, opened 2026-08-18
First-Class Secrets — Phase 1 Implementation


jgraettinger:
Implements Phase 1 of [issue #3366](https://github.com/estuary/flow/issues/3366)
(First-Class Secrets in Estuary). Read the issue first: it holds the motivation,
threat model, authorization rules, and phasing. This plan records the concrete
implementation decisions reached for Phase 1 and sequences them into commits.

**Phase 1 scope**: the `secrets` table; capability bits and GQL surface;
control-plane authorize endpoints; config-encryption `/secret/encrypt` and
`/secret/decrypt`; runtime resolution (both runtimes — see Commit 6); publication
validation; the `flowctl secret` surface with implicit decryption in
`flowctl raw preview-next`; local-stack support and an E2E test catalog.

**Development process**: separate commits, each passing tests, developed in
sequence. Squash required fixes into their originating commit as QA proceeds.
PR presentation (single vs. stack) is decided later.

## Decisions & rationale

Settled decisions a fresh session must not relitigate. Rationale is compressed;
the issue has the long form.

### Data model

- `internal.secrets` table (no PostgREST exposure, no RLS — all access flows
  through control-plane-api):

  ```sql
  create table internal.secrets (
    catalog_name  public.catalog_name primary key,
    id            public.flowid unique not null default internal.id_generator(),
    document      json not null
  );
  create index secrets_catalog_name_spgist
    on internal.secrets using spgist ((catalog_name::text));
  ```

- **`id` is the lifecycle identity**, in the spirit of `flowid` elsewhere:
  each set of a secret is a distinct entity, and `flowid`s are time-ordered so
  it also answers "which is newer?". There is **no `updated_at` column** and no
  version counter. Naming follows codebase idiom: the column is bare `id`
  (like `live_specs.id`); code and APIs outside the table refer to it as
  `secret_id` (like `live_spec_id`).
- **`json`, not `jsonb`**: sops MAC verification is sensitive to document
  traversal order; `jsonb` normalizes key order.
- **No trigger on `secrets` in Phase 1.** The GQL mutation is the sole writer
  and enforces all write semantics in its SQL. (A silent-skip replay guard
  trigger arrives in Phase 2 with the `setSecret` reporting pipeline, on a
  separate `secret_updates` delta table — not this table.)
- **The wrapped document is opaque text end-to-end**: transported as
  `models::RawValue` (in GQL and REST response types alike), stored via
[...]

PR #3423, opened 2026-08-27
First-class secrets: the model, control plane, and config-encryption foundations

jgraettinger:
Commits 1–5 of [#3372](https://github.com/estuary/flow/issues/3372), the Phase 1
implementation plan for first-class secrets ([#3366](https://github.com/estuary/flow/issues/3366)).
That issue holds the motivation, threat model, authorization rules, and the full
commit sequence; read it first. Nothing here is user-reachable yet.

- **models, protocols** — the task `secrets` stanza, threaded into built specs and
  Validate/Discover requests. Inert.
- **control-plane-api** — the `internal.secrets` table, the `ViewSecret` /
  `EditSecret` / `DecryptSecret` capabilities, and the `setSecret` / `deleteSecret` /
  `secrets` GraphQL surface.
- **control-plane-api** — the `/authorize/task/decrypt-secret` and
  `/authorize/user/decrypt-secret` endpoints.
- **config-encryption** — `/secret/encrypt` and `/secret/decrypt`, wrapping secrets
  under a second keychain whose decrypt grant is held by this service alone. Its
  deploy workflow is configured for that keychain in the same commit; the new
  arguments are required, so the deploy must land with it.
- **tokens, flow-client-next, unseal** — the client machinery which resolves a
  `secrets` stanza against those routes.

A `mise` commit at the base of the stack declares task prerequisites and fixes the
generated-artifact checks. It is independent of the rest and predates it only
because the secrets work is what surfaced the gaps.

Remaining Phase 1 work — runtime resolution, publication validation, `flowctl
secret`, and the E2E catalog — follows in later PRs.

Review commit-by-commit: each message carries the rationale for its layer.


crates/config-encryption/src/secrets.rs
```diff
@@ -0,0 +1,762 @@
...
+/// consumes a value the caller already holds -- and it is deliberately
+/// separable from setting the secret, which is where authority is enforced.
+#[axum::debug_handler(state=Arc<App>)]
+#[tracing::instrument(skip(app, body), err(Debug, level = tracing::Level::WARN))]
+pub async fn encrypt_secret(
+    axum::extract::State(app): axum::extract::State<Arc<App>>,
+    axum::extract::Query(EncryptQuery { name }): axum::extract::Query<EncryptQuery>,
+    body: axum::body::Bytes,
+) -> Result<axum::Json<Box<serde_json::value::RawValue>>, Error> {
+    validate_name(&name)?;
```
bbartman, line 82: If you have to do this a lot I highly recommend refactoring it into an extractor that just returns the Encrypted Query. It makes the code easier to read and maintain. If you want an example reach out.

If necessary we could also make this into an axum library for reuse by other services.

crates/config-encryption/src/secrets.rs
```diff
@@ -0,0 +1,762 @@
...
+/// ignored, but fails closed as a `NameMismatch`.
+#[axum::debug_handler(state=Arc<App>)]
+#[tracing::instrument(skip(app, headers), err(Debug, level = tracing::Level::WARN))]
+pub async fn decrypt_secret(
+    axum::extract::State(app): axum::extract::State<Arc<App>>,
+    axum::extract::Query(DecryptQuery { name, started }): axum::extract::Query<DecryptQuery>,
+    headers: axum::http::HeaderMap,
+) -> Result<axum::response::Response, Error> {
+    validate_name(&name)?;
+    let token = bearer_token(&headers)?;
```
bbartman, line 110: I recommend creating an extractor for this as well. It makes it super reusable inside of axum and changes things into declarative, in terms of the arguments that the function takes.

crates/tokens/src/fetch.rs
```diff
@@ -0,0 +1,138 @@
...
+    S: Source,
+{
+    let started = crate::now();
+
+    loop {
+        let retry_after = match source.refresh(started).await? {
+            Ok((token, _valid_for, _revoke)) => return Ok(token),
+            Err(retry_after) => retry_after,
+        };
+        tokio::time::sleep(retry_after.to_std().unwrap_or_default()).await;
```
adrian-estuary, line 29: Can `retry_after` ever be `<= 0`? If so, wouldn't this loop without delay, since `unwrap_or_default` would return `0` in that case? 

Should there be a minimum delay imposed here or a guard to break the loop if `retry_after` is zero / too small?

crates/config-encryption/src/secrets.rs
```diff
@@ -0,0 +1,762 @@
...
+    /// owns both its format and the fact that it's required there.
+    #[serde(default)]
+    pub started: Option<String>,
+}
+
+/// Wrap a secret value into its sops document.
+///
+/// This route is unauthenticated. Wrapping discloses nothing -- it only
+/// consumes a value the caller already holds -- and it is deliberately
+/// separable from setting the secret, which is where authority is enforced.
```
bbartman, line 74: Is there ever a chance that that this endpoint could be hammered by another service or external services? If there is we need a bearer token. 

One other thing to consider here is, if there is a chance that someone could decode our encryption keys by encrypting multiple documents successively then that should also be a concern here. Although I don't know enough about the encryption framework behind the scenes to figure out if that is or isn't the case. I just wanted to call that out.

crates/control-plane-api/src/server/authorize_task_secret.rs
```diff
@@ -0,0 +1,802 @@
...
+        .unwrap_or_default();
+
+    let name = labels::expect_one(unverified.claims().sel.include(), labels::SECRET_NAME)
+        .map_err(|err| tonic::Status::invalid_argument(err.to_string()))?;
+    let name = models::Name::new(name);
+
+    // `err` renders as ": {name} doesn't match pattern ...", restating the name.
+    if let Err(err) = validator::Validate::validate(&name) {
+        return Err(tonic::Status::invalid_argument(format!("invalid secret name{err}")).into());
+    }
```
bbartman, line 67: I would consider refactoring the authorization into an axum Extractor. It will simplify the code. Although I'm not sure how much re-use we would get out of the refactor.

bbartman: I would also include evaluate_authorization but that might be overkill.

crates/config-encryption/src/lib.rs
```diff
@@ -197,15 +215,60 @@ pub async fn encrypt_config(
...
+pub fn cors_layer() -> tower_http::cors::CorsLayer {
+    tower_http::cors::CorsLayer::new()
+        .allow_origin(tower_http::cors::Any)
+        .allow_methods([axum::http::Method::POST])
+        .allow_headers([
+            axum::http::header::AUTHORIZATION,
+            axum::http::header::CONTENT_TYPE,
+            axum::http::header::ACCEPT,
+        ])
+}
```
bbartman, line 248: We should refactor this into an axum utility library. We have a similar function to support something similar inside of the control-plane-api I believe. Note this may be out of scope for this task.

crates/tokens/src/fetch.rs
```diff
@@ -0,0 +1,138 @@
...
+    }
```
bbartman, line 30: This needs the ability to time out eventually, after either some amount of time, or some number of retries. It's possible that we fail to fetch indefinitely for one reason or another, and we would never know.

crates/unseal/src/secrets.rs
```diff
@@ -0,0 +1,382 @@
...
+        anyhow::bail!(
+            "configuration location '{pointer}' is not a JSON pointer: it must be empty, or begin with '/'"
+        );
+    };
+
+    Ok(rest
+        .split('/')
+        .map(|token| token.replace("~1", "/").replace("~0", "~"))
+        .collect())
+}
```
bbartman, line 126: The reason is listed here but I'm lacking some context. Why aren't we using JSONPointer and managing the attribute vs array index differently? I'm missing context, so I'm sure there's a reason for doing this, I just want some context is all.

bbartman (commented):
OK. So I skimmed over most of this did a lot of reading, going commit by commit as suggested in the review description. I tried to focus on the parts of the code surrounding the Axum and graphQL related stuff. 

I have some general cleanup stuff around Axum that I recommend doing to improve readability, and re-usability, over all this seems fine. Nothing glaring stands out to me but you should get some additional approvals from people with more experience with other parts of the system to give it a once over, as I'm still new and still learning the system.

adrian-estuary (commented):
Added a small comment/question. Will defer to Brian for approval, though, as he has much more context on this change.

crates/config-encryption/src/secrets.rs
jgraettinger: This is N==1, the only planned route taking EncryptQuery. Extraction adds indirection for the reader, which is a real cost to be aware of and manage.

crates/config-encryption/src/secrets.rs
jgraettinger: This is N==1 as well, the only route of this service taking a bearer token, and we're doing so in an unauthenticated "forwarding" capacity only. Better to be very explicit about's happening.

(As contrast, we have dedicated extractors in control-plane-api for parsing / verifying / plumbing control plane access tokens)

crates/config-encryption/src/secrets.rs
jgraettinger: Note this is repeating / migrating the status quo -- the endpoint config encryption route of this same service.

By design, this service deliberately _doesn't_ take bearer tokens, because doing so brings in a host of AuthZ responsibility and widens the scope of this service -- scope we're keeping as shrink-wrapped and tight as possible. Thus the architecture of delegating the AuthN/AuthZ check to the control-plane API in the decryption route.

But to answer your question, no real concern about "hammering". It's behind a scale-out cloud run service and KMS key service, it's embarrassingly parallel, and it gives an attacker no benefit I'm aware of.

Replay attacks are mitigated by the architecture of `sops` and KMS infrastructure. Please read up on these. tl;dr is sops picks a cryptographic random nonce that's protected by KMS, and uses that _nonce_ (with AES) to encrypt the rest of the document. KMS also injects its own IV.

crates/control-plane-api/src/server/authorize_task_secret.rs
jgraettinger: I'm not following -- evaluate_authorization is the pre-existing refactor of the re-usable aspects of the AuthZ flow. Everything else here is bespoke to exactly this route. What value are you getting at for refactoring into an Extractor? To my eyes this adds only indirection, a negative.

crates/config-encryption/src/lib.rs
jgraettinger: I disagree -- this is a precise, concise, and explicit statement of an important policy that readers of this service need to know. Refactoring it out a) assumes that policy here and in the control plane are and will always be the same (not a given -- they differ today already), and b) introduces a new dependency coupling in a service designed to have as few as possible.

[Rob Pike on "A little copying is better than a little dependency"](https://www.youtube.com/watch?v=PAAkCSZUG1c&t=568s)

crates/tokens/src/fetch.rs
jgraettinger: As it's a drop-safe future, the idiomatic Rust pattern is to let the *caller* decide timeout policy. `test_caller_bounds_retries` in this PR shows this pattern, using the `tokio::time::timeout` combinator.

crates/config-encryption/src/secrets.rs
bbartman: OK if it's only used once it doesn't matter.

crates/unseal/src/secrets.rs
jgraettinger: From "The secrets Stanza" of the referenced design https://github.com/estuary/flow/issues/3366

> Pointer tokens always address object properties: a token like 2 is the
object key "2", never an array index, and arrays are atomic values, changed by targeting
their parent key. The empty pointer merges at the config root and requires an object-shaped
secret value.

Unpacking: a number is ambiguous (contextually depends on the parent, and if the parent itself doesn't exist, it's truly ambiguous). We have to pick a non-ambiguous handling. The design goes with "it's a property" in all cases, treating arrays as atomic values that must be entirely over-written. Easiest to explain, most-useful in application.

crates/config-encryption/src/secrets.rs
bbartman: OK. If it's only done once it doesn't matter.

crates/config-encryption/src/secrets.rs
bbartman: OK makes sense. Thanks for the context.

crates/config-encryption/src/lib.rs
bbartman: I didn't know about the limited module dependency requirement, that makes sense.

crates/tokens/src/fetch.rs
jgraettinger: It absolutely can be == 0, and that's an intended pattern and use case within `tokens` crate architecture today. It basically means "immediately retry this request over the network". As an example of how this can happen: a Snapshot doesn't authorize a requested access, but the snapshot is stale, and the calculated estimate of the next-refresh time for it is ~now or even in the past. The server sends a retry_after of zero meaning "this should be ready imminently". Note there's still a natural cool-off of the network round trip time.

crates/control-plane-api/src/server/authorize_task_secret.rs
bbartman: So the indirection isn't a negative it creates reusable units with understood behaviors that allow readers to skip sections of the code to make it easier to understand at a high level. I should add that I'm coming from a world where authentication and authorization was the only thing that we did so it was a massive part of every single function, and repeating it in every single function actually made the functions more noisy and visually annoying to slog through, as well as easy to mess up. So any time I see authz code I want to move it into a declaration style, outside of the body of the thing the endpoint is responsible for. My point is I want to hide it because it gets in the way of the main responsibility of the function. So I may end up being too noisy on this, but from my perspective will always be towards the clean code strategy. Hopefully that give you some perspective on why I keep recommending these kinds of refactorings.

crates/tokens/src/fetch.rs
bbartman: Ok then disregard.

crates/unseal/src/secrets.rs
bbartman: OK great. That makes sense. Thank you.

crates/tokens/src/fetch.rs
jgraettinger: If you're curious, here's where we calculate that policy today for the bulk of Snapshot AuthZ requests hitting control plane APIs (including for secrets): https://github.com/estuary/flow/blob/f4956afae9cd7871442d72efe82101723b6198f2/crates/control-plane-api/src/envelope.rs#L152-L162

IIRC there are other use cases of `tokens` that use literal-zero because the request itself blocks.

jgraettinger:
Thanks for reviewing! I'll hold off on rebasing until approved, and then rebase + re-stamp approval.

crates/tokens/src/fetch.rs
adrian-estuary: Thanks for the context!

jgraettinger:
PTAL (just rebased, no other changes)

bbartman (approved):
Looks good to me

PR #3428, opened 2026-08-27
doc: bound allocator growth when draining a key across many segments

### LINKED ISSUE #3413: combiner drain memory excursion for huge, low cardinality docs

williamhbaker:
**Description:**

Each reduction allocates its output from the drainer's bump allocator, which cannot reclaim the previous output, and the allocator was replaced only after a key was yielded. A key repeated across many spill segments grew it without bound within one drain step.

Once the allocator passes `BUMP_THRESHOLD` mid-key, archive the intermediate reduction into an owned buffer, replace the allocator, and continue reducing from the archived document.

Split up into 2 commits, with the first being refactoring to support the second, which contains the new logic.

Q/A testing on a local stack. The expected peak memory reduction during drain was observed, reading from a collection of huge documents with hot keys.

Closes https://github.com/estuary/flow/issues/3413

**Workflow steps:**

**Notes for reviewers:**

### DIFF
--- a/crates/doc/src/combine/spill.rs
+++ b/crates/doc/src/combine/spill.rs
@@ -423,65 +423,69 @@ impl<F: io::Read + io::Seek> SpillDrainer<F> {
         let key = self.spec.keys[binding].as_ref();
         let validator = &mut self.spec.validators[validator_index];
 
-        // `reduced` root which is updated as reductions occur.
-        let mut reduced: Option<HeapNode<'_>> = None;
+        let reduced: Option<HeapNode<'_>> = loop {
+            // `reduced` root which is updated as reductions occur.
+            let mut reduced: Option<HeapNode<'_>> = None;
 
-        // Attempt to reduce additional entries.
-        while let Some(cmp::Reverse(next)) = self.heap.peek() {
-            if meta.0 != next.head.meta.0
-                || !Extractor::compare_key(key, root.get(), next.head.root.get()).is_eq()
-            {
-                self.in_group = false;
-                break;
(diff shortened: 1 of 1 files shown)


crates/doc/src/combine/spill.rs
```diff
@@ -1167,6 +1189,213 @@ mod test {
...
+    /// archives the intermediate result and replaces it. The reservation is never
+    /// written, so it costs address space rather than resident memory.
+    fn exhaust_alloc<F: io::Read + io::Seek>(drainer: &SpillDrainer<F>) {
+        let layout = std::alloc::Layout::from_size_align(BUMP_THRESHOLD + 1, 8).unwrap();
+        drainer.alloc.alloc_layout(layout);
+        assert!(bump_mem_used(&drainer.alloc) > BUMP_THRESHOLD);
+    }
+
+    #[test]
+    fn test_drain_archives_intermediate_reductions() {
```
jgraettinger, line 1202: nit: seems like this could be collapsed into the existing `ArbitraryValue` and / or merge patch fuzz test?

jgraettinger (approved):
LGTM, thanks!

crates/doc/src/combine/spill.rs
williamhbaker: Yeah I think it could be merged into `crates/doc/tests/spill_merge_fuzz.rs` at least, but we'd need to do something like make `BUMP_THRESHOLD` injectable. Probably a good idea as a follow-up / next time we're in here.

PR #3429, opened 2026-08-27
shuffle: never adopt an expanding compression of a sealed segment

jacobmarble:
**Description:**

Shuffle spills its read-ahead data to disk in "segments". When a segment is
sealed (no more writes), we try to shrink it: rewrite every block with LZ4 and
keep the result. LZ4 does not always shrink — on data that is already
incompressible it grows, and each block adds a little header. We measured a
2 MiB segment of random payloads coming back **8159 bytes larger**.

That broke the disk accounting. The owner of a segment subtracts its size from
a backlog counter once, at seal time, and expects the credits it gets back over
the segment's life to add up to exactly that. When the file grew instead:

1. the compression credit clamped to zero, hiding the growth,
2. `sealed.size` was overwritten with the *bigger* number, and
3. the final unlink credit gave back more than was ever borrowed.

After enough grown segments the counter goes below zero and panics with
`disk_backlog_bytes underflow`. A 200 GiB incompressible `preview-next
--fixture` run died that way four transactions in. `LogActor` has the same
`checked_sub().expect(..)`, so production shuffle carried the same latent bug.

Two changes:

- `try_compress` now keeps the original file whenever the rewrite would not be
  smaller, and reports the original size. We still save disk where compression
  actually helps.
- `serve()` also clamps the new size to the size that was debited. That keeps
  the give-back-what-you-borrowed identity true even if the first guard ever
  regresses.

**Workflow steps:**

No user-facing change. Shuffle and `flowctl preview-next` no longer panic on
incompressible spill data.

None.

**Notes for reviewers:**

- The clamp in `serve()` is belt-and-braces on top of the fix in
  `try_compress`. It is cheap and makes the accounting safe by construction, so
  it seemed worth keeping both.
- `test_try_compress_round_trip` already covers the shrinking path;
  `test_try_compress_keeps_original_when_expanding` is new and pins the
  growing one.

### DIFF
--- a/crates/shuffle/src/log/writer/sealed.rs
+++ b/crates/shuffle/src/log/writer/sealed.rs
@@ -78,11 +78,21 @@ impl SealedSegment {
                     match result {
                         Ok(new_size) => {
                             // A segment's lifetime credits must total exactly the
-                            // `size` its owner debited when it was sealed. The min
-                            // guards that identity: `try_compress` never adopts a
-                            // larger file, so this is defensive only.
-                            let new_size = new_size.min(sealed.size);
-                            let reclaimed = sealed.size - new_size;
+                            // `size` its owner debited when it was sealed, and
+                            // `try_compress` never adopts a file larger than the one
+                            // it read. A failed subtraction therefore means that
+                            // guard regressed, so fail here rather than silently
+                            // drift the owner's backlog measure.
+                            let Some(reclaimed) = sealed.size.checked_sub(new_size) else {
+                                return Some((
+                                    Err(anyhow::anyhow!(
+                                        "compression expanded sealed segment {:?} from {} to {new_size} bytes",
+                                        sealed.path,
+                                        sealed.size,
+                                    )),
+                                    State::Done,
+                                ));
+                            };
                             sealed.size = new_size;
                             Some((Ok(reclaimed), State::Compressed(sealed)))
                         }
--- a/crates/shuffle/src/log/writer/sealed.rs
+++ b/crates/shuffle/src/log/writer/sealed.rs
@@ -33,7 +33,9 @@ impl SealedSegment {
     /// 1. If the reader unlinks the file before compression, yields the original
     ///    size and terminates.
     /// 2. If the file survives past `COMPRESS_AFTER`, yields bytes saved by
-    ///    compression (`old_size - new_size`), then continues polling.
+    ///    compression (`old_size - new_size`), then continues polling. When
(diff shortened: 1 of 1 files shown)


crates/shuffle/src/log/writer/sealed.rs
```diff
@@ -75,7 +77,12 @@ impl SealedSegment {
 
                     match result {
                         Ok(new_size) => {
-                            let reclaimed = sealed.size.saturating_sub(new_size);
+                            // A segment's lifetime credits must total exactly the
+                            // `size` its owner debited when it was sealed. The min
+                            // guards that identity: `try_compress` never adopts a
+                            // larger file, so this is defensive only.
+                            let new_size = new_size.min(sealed.size);
+                            let reclaimed = sealed.size - new_size;
```
mdibaiee: I am guessing the previous code is what led to the finding that the size was growing after compression. I think this current code seems to hide an issue like that. Perhaps we can do even a `checked_sub` so that if there is a regression and we end up expanding again, it is a loud error for us to fix

jacobmarble: Thanks, integrated.

mdibaiee (approved):
LGTM pending CI passes

mdibaiee (dismissed):
Actually since this touches shuffle machinery and not only flowctl preview, let's make sure @estuary/data-plane team reviews this. I'm not qualified enough to review this machinery in detail.

crates/shuffle/src/log/writer/sealed.rs
```diff
@@ -196,10 +220,15 @@ fn try_compress(path: &std::path::Path) -> anyhow::Result<u64> {
             .context("segment file overflows 4GB after payload read")?;
     }
 
-    // Disarm so Segment::drop doesn't unlink the file we're about to replace.
+    // Disarm so Segment::drop doesn't unlink the original: we keep it below,
+    // or atomically replace it.
     segment.disarm();
     drop(segment);
 
+    if new_size >= original_size {
```
jgraettinger: instead of discarding here on a whole-file basis, it'd be better to tackle it block-by-block above: test if `compressed` is shorter than `raw`, and if it's not, discard `compressed` and re-write `raw` with its `raw_len` and `lz4_len` of zero.

jacobmarble: Done.

jgraettinger (approved):
LGTM, thanks!

PR #3430, opened 2026-08-27
shuffle: discharge causal hints made unreachable by a byte gap

### LINKED ISSUE #3414: jagged reads at the pruned fragment horizon render hints unresolvable

williamhbaker:
**Description:**

A causal hint names an ACK that a session must read before its checkpoint resolves. A byte gap can remove that ACK: retention pruned it, `begin_mod_time` skipped it, or it was never written. No read can observe it. The session stalls until `CAUSAL_HINT_RESOLUTION_TIMEOUT`, restarts from the same checkpoint, and repeats.

This PR adds a **binding gap floor**: a clock below which a binding's causal hints are unreachable or irrelevant.

- A read raises the floor when the broker resolves its start offset to a larger one.
- The session discharges hints below the floor. Producer commits are not affected, because a byte gap says nothing about whether a producer committed.
- Floors travel in `Frontier.binding_gap_floors`, ratchet in the shuffle session, and persist as `GF:{state_key}` rows in shard zero. Each leader actor seeds its floors from the recovered committed Frontier.

Of note are the cases intentionally NOT covered by this code:
- Remote authoritative derivations, where the checkpoint commits but the subsequent recover log does not, and resumption from a crash follows the recovered checkpoint. This would fail to persist any read byte gap for the transaction. It would take quite a combination of events for this to happen, and long term we are planning to move away from the remote authoritative derivation model entirely.
- Fragments pruned in the middle of a collection, rather than from the tail. Hints might still be unresolvable if this happens, but also the consistency of the collection is almost certainly compromised anyway.
- Fragments pruned after a hinted materialization checkpoint is persisted, but before an idempotent replay of it. Similarly, a `notBefore` addition racing an idempotent replay of a hinted checkpoint. Both of these could leave hints unreachable. The pruned fragment case has a rationale similar to the above point. The `notBefore` addition seems quite obscure and easy to remedy through user action.

See individual commit messages for more detail.

Completed Q/A with a local stack: Drove the stack through various transitions and confirmed that the hint discharge and its floors are durable and operating as expected.

Closes https://github.com/estuary/flow/issues/3414

**Workflow steps:**

**Notes for reviewers:**


crates/shuffle/src/slice/read.rs
```diff
@@ -116,6 +124,38 @@ impl ReadState {
...
+    /// zero may instead have found the journal's head pruned, which
+    /// `not_before` says nothing about, so it also samples its first document
+    /// plus a margin (`sample_gap_floor`). Pruning past a checkpoint, or
+    /// mid-read, raises nothing; the hints it strands are left to the other
+    /// authorities.
+    pub fn note_start_gap(&mut self, offset: i64, not_before: uuid::Clock) {
+        self.gap_floor = self.gap_floor.max(not_before);
+        if offset == 0 {
+            self.pending_gap = true;
+        }
```
dgreer-dev: Is this an issue when the gap is a `begin_mod_time` cut rather than pruning? With a recent `notBefore`, the probe fast-forwards past every existing fragment, so `offset == 0` sets `pending_gap` and `sample_gap_floor` then samples a clock of about now, putting the floor an hour *ahead* of the clocks the read is about to consume.

crates/shuffle/src/lib.rs
```diff
@@ -98,6 +98,23 @@ pub use service::{DEFAULT_SHUFFLE_DISK_LIMIT_BYTES, Service};
 pub const PRODUCER_STALENESS_HORIZON: std::time::Duration =
     std::time::Duration::from_secs(48 * 60 * 60);
 
+/// Margin added to the clock of the first document consumed by a read from
+/// offset zero after a byte gap, raising the binding's gap floor: the clock
+/// below which its causal hints are discharged as unreachable
+/// ([`frontier::Completed::is_gap_stale`]).
+///
+/// Such a read may have found the journal's head pruned, which `notBefore` says
```
jgraettinger: nit: perhaps do a de-duplication / don't repeat yourself pass over the comment prose?

crates/shuffle/src/slice/read.rs
jgraettinger: More broadly, I don't think we want `begin_mod_time` cuts at all within this mechanism.

`begin_mod_time` is used to filter appends, but doesn't affect ACK propagation or causal hint resolution. It just suppresses the _content_ of those transactions. I think we can totally ignore that begin_mod_time is a thing in this implementation, and focus just on offset jumps / gaps, and that it would still work because `begin_mod_time` sometimes causes such a gap (and sometimes doesn't, in which case no gap floor needs to be recorded).

crates/shuffle/src/session/state.rs
```diff
@@ -329,6 +329,12 @@ impl CheckpointPipeline {
         let mut completed = crate::frontier::Completed::new(binding_cohorts);
         completed.update(resume_checkpoint);
 
+        // An idempotent-recovery session must read through its hinted frontier
+        // exactly, so it does not observe gap floors.
+        if !recovery_session {
+            completed.advance_gap_floors(&resume_checkpoint.binding_gap_floors);
```
jgraettinger: I'm not sure I'm following why this is required.

Also, what happens if idempotent replay content has been entirely pruned due to a gap?

crates/shuffle/src/slice/read.rs
williamhbaker: I had been trying (but not completely accomplishing, per Dave's comment) to avoid effectively turning off causal hints for the 1-hr of clock time following something like a collection truncation, where a `begin_mod_time` would regularly be applied. The additional fudge factor is needed to account for clocks not being guaranteed to be in ascending order in a journal, but that requires multiple producers and significantly different clocks amongst them.

Practically I think the extra buffer could be more like 10 minutes and be fine for actually doing what it needs to do, and then it would be less disruptive as well when applied to gaps caused by `begin_mod_time`. So I'll plan on doing that: Ignore `begin_mod_time` and tighten the extra padding from 1 hr to 10 minutes.

crates/shuffle/src/session/state.rs
williamhbaker: Agreed on this check not being required, and it would have prevented the ratchet from advancing when hints were discharged by a gap floor. I'll remove that.

Related to this and your question about idempotent replay content that has been entirely pruned...there's a similar check in `on_progressed` that I had over-applied here. I think that one is still necessary though, otherwise a recovery transaction could potentially discharge its own projected hints if the `last_commit`s happen to fall near the gap.

This means that if idempotent replay content gets pruned between when the frontier is persisted and when it is replayed there is still a potential for a hint resolution stall. I've given that scenario quite a bit of thought but haven't come up with a solution I like very much.

crates/shuffle/src/lib.rs
williamhbaker: Done, made a few cleanups/improvements.

williamhbaker:
Force-pushed changes based off review feedback, and to take care of a merge conflict.

crates/shuffle/src/session/state.rs
```diff
@@ -471,6 +472,15 @@ impl CheckpointPipeline {
             "received Progressed response",
         );
 
+        // A floor cannot tell an ACK the gap removed from one just past it, so
+        // the session which must replay its hinted frontier exactly acts on no
```
jgraettinger, line 476: Trying to think through this case makes my head hurt 🫠.

Per your comment before:
> This means that if idempotent replay content gets pruned between when the frontier is persisted and when it is replayed there is still a potential for a hint resolution stall. I've given that scenario quite a bit of thought but haven't come up with a solution I like very much.

That makes sense, if the producer itself is gone and doesn't re-appear -- and we're in "backfill the binding" territory, so that's fine.

What happens if the hinted commit producer is still active, but the hinted commit has been pruned? In that case we may silently fold it's next ACK into the idempotent txn ... I think? As that's the first ACK that clears the hinted commit.

I'm not sure what, if anything, we do about this (nor am I confident I have the scenario right). I suppose the biggest actual concern is that it's a silent rather than noisy failure.

jgraettinger (approved):
contingent LGTM, though a concern below. I'd like to spend a bit more time thinking and talking about this, on the possibly-unfounded aspiration that there's a simpler approach

jgraettinger (dismissed):
Enthusiastic LGTM this time.

I spent some cycles thinking about alternatives, and they all reduced into the same functional architecture as is here. I can't see a way around having a fudge factor, and as soon as there's a fudge, it implies the persistence requirement.

williamhbaker:
Rebased to address a merge conflict, and pushed on more commit ([d6c6367](https://github.com/estuary/flow/pull/3430/commits/d6c63671892806acea071da00a5ea24dd5a13655)) that addressed an additional gap uncovered by a final round of testing: The leader needs to be able to discharge hints that the session discharged because of an observed gap, if it had previously seen them in a peeked frontier. So it does that now, with a little bit of extra code.

jgraettinger (approved):
Still LGTM

PR #3431, opened 2026-08-27
Labeler fixes

GregorShear:
Operability fixes for the label reconciler. The `fetch-depth: 0` checkout downloads the repository's multi-GB blob history on every master push and dominates the run (the first post-merge run spent 20+ minutes cloning); `filter: blob:none` keeps the full commit graph — `git log <deployed>..HEAD` is unchanged — while fetching file contents only for the checked-out tree. Reconciliation now runs as one matrix job per target (`agent-api` / `agent` / `flowctl`, `fail-fast: false`) so a failing target is visible by job name and doesn't affect the others — e.g. until the `OPS_READ_TOKEN` secret is configured, only the `agent` job fails; the script's `--only` flag selects the target, and running it without the flag still reconciles everything for local use.

### DIFF
--- a/.github/scripts/reconcile_deploy_labels.py
+++ b/.github/scripts/reconcile_deploy_labels.py
@@ -247,30 +247,42 @@ def main() -> None:
     agent_dirs = closure(repo_root, "agent")
     flowctl_dirs = closure(repo_root, "flowctl")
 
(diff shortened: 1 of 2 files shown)


skord (approved):
Lgtm

PR #3433, opened 2026-08-28
control-plane-api, agent: capability-token validity is configurable

bbartman:
Follow-up to #3424 (stacked on #3427): the validity window of `capability_token` mints becomes a deploy-time setting instead of a hardcoded constant.

## What this does

- `App` gains a `capability_token_validity` field, fed by a new agent setting `--capability-token-validity` / `CAPABILITY_TOKEN_VALIDITY` (humantime syntax, e.g. `1h`, `30m`), following the existing pattern of the controller duration settings (`CONTROLLER_DEQUEUE_INTERVAL`, `CONTROLLER_HEARTBEAT_TIMEOUT`).
- `sign_capability_token` takes the validity as a parameter — it stays pure; no env reads at request time.
- **The default is `1h`**, matching the SQL `generate_access_token` mint, so an unconfigured deployment (production today, local stacks, test harnesses) mints exactly the window it always has. No deploy changes are required anywhere; overriding is one line in `deploy-agent-api.yaml`'s `env_vars` or a local stack's agent env file.

## Deliberate trade-offs (noted on #3376)

- An override diverges from the SQL mint's hardcoded `interval '1 hour'` — claim-*set* parity between the two mints is unaffected, but the windows differ when configured. The flag docs say so.
- A longer window widens the exposure of a leaked masked token; the one-hour bound is part of the "one approval = one short-lived token" design, and the flag docs carry that warning.

## Tests

`test_validity_is_configurable` pins a non-default (90s) window at the signing seam; the end-to-end mint test continues to exercise the harness-configured one-hour default.

### DIFF
--- a/crates/agent/src/integration_tests/harness.rs
+++ b/crates/agent/src/integration_tests/harness.rs
@@ -1790,6 +1790,8 @@ impl TestHarness {
             self.publisher.clone(),
             snapshot_watch.clone(),
             None,
+            // The production default of the --capability-token-validity setting.
+            std::time::Duration::from_secs(3600),
         ));
 
         self.control_plane_app = Some(app);
--- a/crates/agent/src/main.rs
+++ b/crates/agent/src/main.rs
@@ -95,6 +95,17 @@ struct Args {
     )]
     #[arg(value_parser = humantime::parse_duration)]
     heartbeat_timeout: std::time::Duration,
+    /// Validity window of access tokens minted by the `capability_token`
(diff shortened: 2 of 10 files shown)


bbartman:
Folded into #3424 (task 5's PR) rather than landing as a separate stacked PR: the configurable validity is a refinement of the mint itself, so it lives with it. The relocated commit also consolidates the default into a shared DEFAULT_CAPABILITY_TOKEN_VALIDITY constant with a weld test pinning the clap default string to it. #3427 is rebased on the updated base.

PR #3435, opened 2026-08-28
control-plane-api: create_data_plane gates ops admin via Snapshot

bbartman:
Part 1 of a 3-PR stack strangling the last reachable Rust call sites of `internal.user_roles()`, continuing the Snapshot-authorization work of #2781. Stacked on the publications-spec-authz branch.

## Plan

After the publications migration, exactly three reachable Rust paths still authorize via `internal.user_roles()` SQL:

1. **`/admin/create-data-plane`** — this PR.
2. **`/admin/update-l2-reporting`** — next PR, an identical swap.
3. **The storage-mappings directive** (`user_has_admin_capability`) — final PR, which must thread the agent's `snapshot_watch` into `DirectiveHandler`.

Everything else that references `user_roles`/`auth_roles` is SQL-land reachable only via PostgREST/RLS (`live_specs_ext`, `combined_grants_ext`, `gateway_auth_token`, …) and is out of scope for this stack. In particular, the #2848 workaround `ensure_private_data_plane_grants` stays: its removal is gated on the RLS-side migration, not on these three.

## This change

- Swaps `create_data_plane`'s SQL gate for `evaluate_names_authorization(snapshot, claims, Admin, ["ops/"])`, evaluated through `Envelope::authorization_outcome`. Denials now follow the standard authorization-retry protocol (early Snapshot refresh + 307 retry when the Snapshot may be stale) instead of failing terminally, matching the `authorize_*` endpoints. This is an HTTP-envelope mechanism; GraphQL resolvers and the (upcoming) directive handler instead take one-shot pinned-Snapshot checks.
- The denial message becomes the crate-standard "…is not authorized to access prefix or name 'ops/' with required capability admin". (An earlier revision introduced a dedicated `evaluate_ops_admin` helper with a bespoke message; review flagged it as a Middle Man over `evaluate_names_authorization` and it was folded away.)
- Adds shared test fixtures `make_grants` / `snapshot_of_grants` to `test_server`, deduplicating the grant-table construction previously repeated in the `authorized_prefixes` tests.

### Deliberate semantic change

`internal.user_roles()` walks the grant graph only *downward* from roles the user holds. The Snapshot walk (`tables::UserGrant::is_authorized`) also reaches grants held by *ancestor* subject roles — the same gap tracked by #2848. An admin chain into `ops/` through an ancestor subject (e.g. admin of `estuary/support/` with a `(estuary/ → ops/, admin)` grant) was wrongly denied by the SQL gate and is authorized here. The named test `test_evaluate_ops_admin_ancestor_subject_chain` pins this.

## Tests

- Unit tests in `server::tests` pin the ops-gate semantics: direct `ops/` admin, unrelated-tenant admin denied (with the exact denial message), `read` ≠ admin, the ancestor-subject chain allowed, and the same chain denied without admin capability.
- Endpoint tests cover the denial paths only (no bearer → 401, non-ops-admin → terminal 403 with the expected message). An authorized request proceeds into data-plane provisioning, which authorization tests have no business exercising — so allowed cases live in the unit tests.

## Review

An `xhigh` two-axis review (standards + spec) ran against this PR; its findings — a named-test granularity nit, the Middle Man helper, and the duplicated test fixture — are addressed in the two follow-up commits.

### DIFF
--- a/crates/control-plane-api/src/server/mod.rs
+++ b/crates/control-plane-api/src/server/mod.rs
@@ -502,12 +502,19 @@ mod tests {
         // A read grant to `ops/` is not admin.
         let snapshot = snapshot_of_grants(&[(user, "ops/", Read)], &[]);
         assert!(evaluate_ops_admin(&snapshot, &claims(user)).is_err());
+    }
 
-        // An admin chain reached through an *ancestor* subject role:
-        // the user admins `estuary/support/`, and `estuary/` (its ancestor)
-        // is granted admin of `ops/`. The replaced `internal.user_roles()`
-        // gate walked only downward and wrongly denied this; the Snapshot
-        // walk authorizes it. This pins the deliberate semantic change.
+    /// Pins the deliberate semantic change from the replaced
+    /// `internal.user_roles()` gate: an admin chain reached through an
+    /// *ancestor* subject role. The user admins `estuary/support/`, and
+    /// `estuary/` (its ancestor) is granted admin of `ops/`. The SQL gate
+    /// walked only downward from held roles and wrongly denied this; the
+    /// Snapshot walk authorizes it.
+    #[test]
+    fn test_evaluate_ops_admin_ancestor_subject_chain() {
+        use models::Capability::{Admin, Read};
+        let user = uuid::Uuid::from_bytes([0x11; 16]);
+
         let snapshot = snapshot_of_grants(
             &[(user, "estuary/support/", Admin)],
             &[("estuary/", "ops/", Admin)],
--- a/crates/control-plane-api/src/server/mod.rs
+++ b/crates/control-plane-api/src/server/mod.rs
@@ -136,6 +136,31 @@ where
     Ok((None, ()))
 }
 
+/// Evaluate whether the user identified by `claims` is an admin of the
+/// `ops/` role, which gates the `/admin/*` endpoints.
+/// Return a policy_result shape which fits Envelope::authorization_outcome.
+///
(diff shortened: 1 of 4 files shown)


crates/control-plane-api/src/server/mod.rs
```diff
@@ -407,3 +407,89 @@ const fn map_capability_to_gazette(capability: models::Capability) -> u32 {
...
+            status.message(),
+            "user@example.test is not authorized to access prefix or name 'ops/' with required capability admin"
+        );
+
+        // A read grant to `ops/` is not admin.
+        let snapshot = snapshot_of_grants(&[(user, "ops/", Read)], &[]);
+        assert!(evaluate(&snapshot).is_err());
+    }
+
+    /// Pins the deliberate semantic change from the `internal.user_roles()`
```
jgraettinger: nit: this test feels like overkill -- testing functionality tested elsewhere -- and we should avoid prose which references the historical archeology of how the codebase evolved into it's current shape (prose should reflect the current shape, not argue against or reference shapes we no longer have)

jgraettinger (dismissed):
LGTM % nit

crates/control-plane-api/src/server/mod.rs
bbartman: I tell claude not to do that literally every single time, and no matter how many times I tell it it still does it.

jgraettinger: I tackle it as part of dedicated simplification and review tasks, under the umbrella of agentic and human review done after the initial solution is worked up.

jgraettinger (approved):

PR #3436, opened 2026-08-28
control-plane-api: update_l2_reporting gates ops admin via Snapshot

bbartman:
Part 2 of the 3-PR stack strangling the last reachable Rust call sites of `internal.user_roles()`. Stacked on #3435.

## Plan

With #3435 having migrated `/admin/create-data-plane` and established the shared gate (`evaluate_names_authorization(snapshot, claims, Admin, ["ops/"])` through `Envelope::authorization_outcome`), this PR applies the identical swap to `/admin/update-l2-reporting` — the second of the three remaining call sites. The third (the storage-mappings directive) follows in the final PR, which threads the agent's `snapshot_watch` into `DirectiveHandler`.

## This change

- `update_l2_reporting` drops its `internal.user_roles()` SQL gate for the shared Snapshot policy. Denials gain the standard authorization-retry protocol (early Snapshot refresh + 307 retry on possible staleness), and the ancestor-subject grant walk applies — the deliberate semantic change pinned by `test_evaluate_ops_admin_ancestor_subject_chain` in #3435.
- This removes the **last caller** of the exact query text `select role_prefix from internal.user_roles($1, 'admin') where role_prefix = 'ops/'`, so `cargo sqlx prepare` drops its orphaned offline-cache entry. Two `live_specs` query-metadata files also regenerated with updated column nullability against current migrations (deliberate: full-prepare was chosen over hand-deleting the orphan; CI's `sqlx-check` arbitrates).

## Tests

Denial-path endpoint tests mirror create-data-plane's (no bearer → 401, non-ops-admin → terminal 403 with the standard message). With the gate's policy shared and unit-tested in #3435, per-endpoint *wiring* is the remaining regression surface — which is exactly what these pin. Allowed cases (including the ancestor chain) stay in the shared gate's unit tests; no success-path endpoint test, since an authorized request proceeds into the L2 template publication, out of scope for authorization tests.


jgraettinger (approved):
LGTM

PR #3437, opened 2026-08-28
control-plane-api: Add tenant filter to dataPlanes query

adrian-estuary:
**Description:**

Related to https://github.com/estuary/ui/issues/2065

Adding a new `tenant` string filter to allow filtering the DataPlanesQuery results to just data planes owned by that tenant. This is required to migrate the DataPlanesTable to GQL, since the current supabase query does this filtering.

**Filter by tenant `acmeCo/`**
Screenshot 2026-08-28 at 11 13 39 AM

**Filter by tenant `umbrellaCo/`**
Screenshot 2026-08-28 at 11 14 47 AM

**Includes public data planes by default**
Screenshot 2026-08-28 at 11 17 49 AM

**Workflow steps:**

The filter will be used by the data planes table in the admin UI to filter planes to the selected tenant.

No documentation changed.

**Notes for reviewers:**

### DIFF
--- a/crates/control-plane-api/src/fixtures/bob_co.sql
+++ b/crates/control-plane-api/src/fixtures/bob_co.sql
@@ -1,0 +1,92 @@
+-- The `bobCo/` tenant: two users and three private data planes.
+--
+-- It exists so tests can exercise queries that must distinguish one
+-- tenant's data planes from another's.
+--
+--   * alice and bob both read every plane bobCo/ can reach. Their tenant-level
+--     `admin` grant carries `Delegate`, so authorization chains through the
+--     `role_grants` edges below and out to each plane.
+--   * only bob manages bobCo's private planes. The `manage_data_plane` bundle
+--     rides on a `user_grants` row addressed to bob, rather than on the
+--     tenant-wide role edge, precisely so it does not reach alice.
+do $$
+declare
+  alice_uid uuid := '11111111-1111-1111-1111-111111111111';
+  bob_uid uuid := '22222222-2222-2222-2222-222222222222';
+begin
+
+  insert into public.tenants (id, tenant) values
+    (internal.id_generator(), 'bobCo/')
+  ;
+
+  -- `hmac_keys` is populated directly (rather than via `encrypted_hmac_keys`)
+  -- so the snapshot loader takes these rows as live without exercising the
+  -- SOPS decrypt path, which `data_planes.sql` already covers. Every column
+  -- that isn't a plane's identity is derived from its slug: fixtures assert
+  -- over names and capabilities, and uniform addresses keep the per-plane
+  -- rows to a line each.
+  insert into public.data_planes (
+    id,
+    data_plane_name,
+    data_plane_fqdn,
+    hmac_keys,
+    broker_address,
+    reactor_address,
+    ops_logs_name,
+    ops_stats_name,
+    ops_l1_events_name,
+    ops_l1_inferred_name,
+    ops_l1_stats_name,
+    ops_l2_events_transform,
+    ops_l2_inferred_transform,
+    ops_l2_stats_transform,
+    enable_l2,
+    cidr_blocks
+  )
+  select
+    dp.id::flowid,
+    dp.name::catalog_name,
+    'dp.' || dp.slug,
+    '{c2VjcmV0,b3RoZXI=}'::text[],
+    'broker.dp.' || dp.slug,
+    'reactor.dp.' || dp.slug,
+    ('ops/tasks/' || dp.slug || '/logs')::catalog_name,
+    ('ops/tasks/' || dp.slug || '/stats')::catalog_name,
+    ('ops/rollups/L1/' || dp.slug || '/events')::catalog_name,
+    ('ops/rollups/L1/' || dp.slug || '/inferred')::catalog_name,
+    ('ops/rollups/L1/' || dp.slug || '/stats')::catalog_name,
+    'from.dp.' || dp.slug,
+    'from.dp.' || dp.slug,
+    'from.dp.' || dp.slug,
+    false,
+    '{10.20.0.0/16}'::cidr[]
+  from (values
+    ('663333333333', 'ops/dp/private/bobCo/aws-us-east-1-c1',    'private/bobCo-one'),
+    ('664444444444', 'ops/dp/private/bobCo/gcp-us-central1-c1',  'private/bobCo-two'),
+    ('665555555555', 'ops/dp/private/bobCo/az-eastus-c1',        'private/bobCo-three')
+  ) as dp(id, name, slug)
+  ;
+
+  insert into public.user_grants (user_id, object_role, capability) values
+    (alice_uid, 'bobCo/', 'admin'),
+    (bob_uid, 'bobCo/', 'admin')
+  ;
+
+  -- The tenant reads its own private planes and the shared public ones. `read`
+  -- conveys the `Viewer` bundle, which includes `ViewDataPlanePrivateNetworking`
+  -- but stops short of modifying it.
+  insert into public.role_grants (subject_role, object_role, capability) values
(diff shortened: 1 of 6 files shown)


crates/control-plane-api/src/fixtures/bob_co.sql
adrian-estuary, line 1: Added a couple of fixtures to represent a single user (`bob`) having admin access to two tenants (`bobCo` and `bobCo2`) so that I could test the new `tenant` filter for data planes.

crates/control-plane-api/src/server/public/graphql/data_planes.rs
```diff
@@ -487,11 +491,26 @@ impl DataPlanesQuery {
...
         let public_eq = public.and_then(|f| f.eq);
 
+        // Tenant filter must end with a slash so that it captures the full tenant path segment in the
+        // private data plane names. This prevents the tenant filter from matching two tenants with the same
+        // starting characters (e.g. keeps "foo" from matching both "foo-bar" and "foo-baz").
+        if tenant.as_ref().is_some_and(|t| !t.ends_with("/")) {
+            return Err(async_graphql::Error::new(format!(
+                "tenant filter must end with a slash"
+            )));
+        }
```
adrian-estuary: @GregorShear - Thoughts on this check? Could also just have it ignore invalid tenants and just return public data planes.

GregorShear: any reason not to append that `/` if not already present?

adrian-estuary: @GregorShear - Nope, we can definitely do that. Will update.

crates/control-plane-api/src/server/public/graphql/data_planes.rs
```diff
@@ -431,32 +435,26 @@ fn paginate_by_name<T>(
...
-                    .map(|b| name(t) < b.as_str())
-                    .unwrap_or(true)
-            })
-            .collect();
+    let total = sorted.len();
 
-        let total = filtered.len();
-        let skip = total.saturating_sub(limit);
-        let rows: Vec<T> = filtered.into_iter().skip(skip).collect();
-        (rows, skip > 0, before.is_some())
```
adrian-estuary, line 449: @GregorShear - Found a small bug in the cursor pagination implementation for data planes. Basically if an empty or invalid cursor was passed in, since it was determining the `hasPreviousPage` based on the existence of the passed in `after` cursor, it could return `true` for has previous even though the actual results don't have previous page. Added some extra test cases to validate and tested the behavior locally.

crates/control-plane-api/src/server/public/graphql/data_planes.rs
adrian-estuary: Updated!

bbartman:
This looks good, you get my approval. Just finishing reviewing things from Greg and you should be fine.

bbartman (commented):
This looks good from a backend perspective.

crates/control-plane-api/src/server/public/graphql/data_planes.rs
GregorShear: huh cool. good catch. do other resolvers have the same bug?

crates/control-plane-api/src/server/public/graphql/data_planes.rs
```diff
@@ -14,6 +14,10 @@ pub struct DataPlanesFilter {
     /// Match data planes by id, such as a `LiveSpec`'s `dataPlaneId`.
     /// Unknown or unauthorized ids are silently dropped.
     pub id: Option<filters::IdFilter>,
+    /// Match data planes by tenant. Only data planes whose name prefix matches the tenant are returned.
+    /// All public data planes are considered to match all tenants. Unknown tenants are ignored and only
+    /// public data planes are returned.
+    pub tenant: Option<String>,
```
GregorShear: I acknowledge this isn't codified anywhere yet, but i'd like for anything that appears in the `filters` param to be defined in the `filters::` module, so for `tenant` filtering I'd add a new `StringFilter` like:

```rust
pub struct StringFilter {
    pub eq: Option<String>,
}
```

and that will potentially be extended with `starts_with`, `in`, etc...

crates/control-plane-api/src/server/public/graphql/data_planes.rs
GregorShear: [low conviction] maybe my comment about `StringFilter` above makes the trailing slash unnecessary?

crates/control-plane-api/src/server/public/graphql/data_planes.rs
GregorShear: I also acknowledge that this pattern commits us to supporting filters in consumers that didn't necessarily want them, but it feels worthwhile if it makes the API consistent and predictable. Feel free to challenge this...

adrian-estuary: No challenge. Makes sense to me. Will update.

crates/control-plane-api/src/server/public/graphql/data_planes.rs
adrian-estuary: @GregorShear - I already updated this, but it appears that the ([tenant query](https://github.com/estuary/flow/blob/master/crates/control-plane-api/src/server/public/graphql/tenant.rs#L38-L41)) validates the name as a prefix and errors if the passed in value is invalid, similar to my original approach. 

For consistency's sake, I'm thinking of reverting to the original logic (though will use the prefix validation instead). Let me know your thoughts.

Screenshot 2026-09-01 at 9 29 48 AM

crates/control-plane-api/src/server/public/graphql/data_planes.rs
adrian-estuary: Yea, seems so. Other resolvers (storage mappings, connectors, etc...) seem to roll their own pagination, but the code looks pretty similar and appears to have the same bug. 

Tested w/ the storage mappings and with an empty `after` cursor, `hasPreviousPage` is unexpectedly true.
Screenshot 2026-09-01 at 9 38 31 AM

Added a ticket to validate and fix. Want to keep this PR's scope small.
https://github.com/estuary/flow/issues/3449

crates/control-plane-api/src/server/public/graphql/data_planes.rs
adrian-estuary: Updated!

crates/control-plane-api/src/server/public/graphql/data_planes.rs
adrian-estuary: Cool, updated to use the validate function instead of normalizing.

crates/control-plane-api/src/server/public/graphql/data_planes.rs
```diff
@@ -431,32 +437,26 @@ fn paginate_by_name<T>(
...
-            .take(limit)
-            .collect();
-
+        // Forward: the window starts just past the cursor.
+        let start = match after.as_ref() {
+            Some(a) => sorted.partition_point(|t| name(t) <= a.as_str()),
+            None => 0,
+        };
+        let rows: Vec<T> = sorted.into_iter().skip(start).take(limit).collect();
         let has_next = rows.len() == limit;
```
GregorShear:
```suggestion
        let lookahead = limit + 1;
        let rows: Vec<T> = sorted.into_iter().skip(start).take(lookahead).collect();
        let has_next = rows.len() == lookahead;
```
right?

GregorShear: eh, i botched the suggestion, but you get the idea.

adrian-estuary: Updated!

GregorShear (approved):

PR #3439, opened 2026-08-28
shuffle: remap materialization key routing to couple shard read rates

### LINKED ISSUE #3421: runtime-v2 back-pressure equalization

dgreer-dev:
Fixes #3421.

## Problem

Shards own solid key-hash ranges. A Slice reading journals aligned with its
shard's range appends only to its own Log, so nothing ties its read rate to
the other shards'. When one shard of a multi-shard task lands on a busy
machine, it reads slower than the rest, its log falls behind in wall-clock,
and each transaction mixes documents of widely differing age. We saw a ~2x
rate divergence between two materialization shards, with transaction size
spiking as a result. Correctness was fine; the transactions were just
inconsistent in what "now" meant.

## Change

Key routing becomes "mostly solid". Within route_to_shards, a small fixed
fraction of keys (4/256 = 1.56%, selected by the key hash's low byte) route
by a remapped hash (its 16-bit halves swapped) which lands them uniformly
across the key space rather than in their own range. Every Slice then
appends a trickle of documents into every Log. Because a Log drains its
merge heap in (priority, clock) order, a lagging Log holds a fast Slice's
remapped Append behind its own older documents, and the Slice is
head-of-line blocked on that send, which back-pressures its journal reads
toward the slowest shard.

Notes on the mechanics:

- The remap is entirely internal to route_to_shards. The key hash that
  connectors and log entries observe is always the true hash of the packed
  key.
- It is a pure function of the key hash, so every Slice agrees on where a
  key lands, across sessions and topologies.
- It composes with the existing 2D routing rather than replacing it:
  r_clock filtering and the keyed broadcast across r_clock splits still
  apply to remapped keys.
- Swapping the halves puts the true hash's second byte at the top of the
  routed hash, so the (biased) selecting low byte doesn't bias which range
  a remapped key lands in.
- The math is built up in a block of REMAP_* consts; changing the fraction
  is a one-const edit.

This applies to all shuffled reads: materializations, derivations, and
ad-hoc collection reads. Materializations hold no shard-local per-key state
(the destination is the state, and transactions Load it), and there are no
scaled stateful derivations, so re-homing a key is safe everywhere. Splits
are safe for the same reason a split is safe today: a remapped key's routed
hash is fixed, so a split moves it exactly as it would any key whose hash
falls in the split range.

## Observability

A second commit makes the effect measurable in prod:
shuffle_slice_last_source_published_at_time_seconds{shard_id, cohort} is
the unix time of the last document a Slice dequeued from its ready-read
heap, set at the point of dequeue. `time() - gauge` is a shard's read lag,
and max - min across sibling shards is their skew, which is also the
measure of how tightly the remapped routing actually couples shards. It is
labeled by cohort because clocks are only comparable within one: bindings
of differing priority or read delay legitimately diverge. The gauge holds
absolute time per Prometheus practice, so it never goes stale and needs no
periodic refresh, and the per-document cost is one atomic store against a
pre-created per-cohort gauge handle.

## Testing

- agent assisted verification that keys are landing in correct places
- route_to_shards tests: a remapped key routes by its remapped hash while
  its non-remapped neighbor routes home; the remap composes with r_clock
  topologies (keyed broadcast and filtered routing).
- remap_key_hash tests: identity outside the threshold; the swap keeps the
  biased selecting byte out of the routed hash's top byte.
- cargo test -p shuffle (unit, scenarios), -p runtime-next, and -p flowctl
  preview_next all pass with no snapshot churn.

### DIFF
--- a/crates/shuffle/README.md
+++ b/crates/shuffle/README.md
@@ -345,6 +345,18 @@ computes its hash, and routes to target Log shard(s) using
 `filter_r_clocks` additionally filters by the rotated clock value,
 distributing reads across shards in the r_clock dimension.
 
+Key routing is deliberately "mostly solid": within `route_to_shards()`, a
+small fixed fraction of keys (`routing::REMAP_THRESHOLD / REMAP_BUCKETS`,
+selected by the key hash's low bits) route by a remapped hash — the hash's
+16-bit halves swapped — landing them across the key space rather than in
+their own range. This puts a
+trickle of every Slice's documents into every Log, so a lagging Log's
+clock-ordered merge (§8) back-pressures a Slice that has raced ahead,
+bounding how far shards' wall-clocks can diverge within a transaction. The
+remap is a pure function of the key hash, so every Slice agrees on it, and
+it is internal to routing: the key hash that connectors and log entries
+observe is always the true hash of the packed key.
+
 The document, its packed key, metadata, and journal context are sent
 as an `Append` message to each target Log. Journal names are
 delta-encoded across consecutive sends to minimize wire overhead.
--- a/crates/shuffle/README.md
+++ b/crates/shuffle/README.md
@@ -319,6 +319,14 @@ Before processing the heap top, the Slice gates on wall-clock time: if
 sleeps until the clock catches up. This is how read delays impose
 cross-transform ordering guarantees.
 
+Each dequeue also records the document's clock on the
+`shuffle_slice_last_source_published_at_time_seconds` gauge, labeled by
+cohort: how far this shard's shuffled read has progressed in source
+published-at time. `time() - gauge` is the shard's read lag, and max - min
+across sibling shards is their skew — the measure of how tightly remapped
+routing (§7) actually couples shards. Clocks are only comparable within a
+cohort: bindings of differing priority or read delay legitimately diverge.
+
 ### 6. Document Sequencing
 
 The top document is sequenced against per-producer state using
--- a/crates/shuffle/src/slice/actor.rs
+++ b/crates/shuffle/src/slice/actor.rs
@@ -849,6 +849,10 @@ impl SliceActor {
             // Copy so the `binding` borrow can end below, freeing &mut self for re-borrow.
             let (cohort, read_delay) = (binding.cohort, binding.read_delay);
 
+            let (clock_seconds, clock_nanos) = clock.to_unix();
+            self.metrics.last_source_published_at[cohort as usize]
+                .set(clock_seconds as f64 + clock_nanos as f64 / 1_000_000_000.0);
+
             if sequenced.is_commit {
                 if flags.is_ack() {
                     // This ACK is (binding, journal)-scoped: it commits only
--- a/crates/shuffle/src/slice/handler.rs
+++ b/crates/shuffle/src/slice/handler.rs
@@ -68,9 +68,15 @@ where
     handler.set_field("token", serde_json::to_string(&authz.claims()).unwrap());
     handler.set_phase("opening");
 
-    let metrics = super::Metrics::new(shard_id);
     let task = task.context("Open must include task")?;
     let (bindings, sources, validators) = crate::Binding::from_task(&task)?;
+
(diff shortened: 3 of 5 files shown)


dgreer-dev:
I haven't been able to get a smoking gun red test of this locally, but I was able to observe fuzzed keys going to each shard in approximately the right proportions.

crates/shuffle/src/binding.rs
```diff
@@ -29,6 +29,15 @@ pub struct Binding {
     ///
     /// This field is only true for read-only derivation transforms.
     pub filter_r_clocks: bool,
+    /// Should a small fraction of keys be routed away from the shard owning
```
jgraettinger: I think we strike this, and make it universally true.

crates/shuffle/src/binding.rs
```diff
@@ -414,6 +426,19 @@ impl Binding {
     pub fn state_key(&self) -> &str {
         self.journal_read_suffix.rsplit("/").next().unwrap()
     }
+
+    /// Hash of a document's packed shuffle key, as routed to shards by
+    /// `slice::routing::route_to_shards`. This is the single definition of a
+    /// binding's key routing, shared by the Slice and by `flowctl preview`'s
+    /// fixture feeder so the two cannot disagree on where a key lands.
+    pub fn routing_key_hash(&self, packed_key: &[u8]) -> u32 {
```
jgraettinger: I don't think we should change the hash -- note that connectors _observe_ the hash -- but should instead contain this as a detail within `route_to_shards`  (for example, it does an extra rotation of the key hash so that low bits dominate)

crates/shuffle/src/session/actor.rs
```diff
@@ -90,6 +107,10 @@ impl SessionActor {
                 // and detects stalled causal hint resolution.
                 _ = ticker.tick() => {
                     self.checkpoint.on_tick()?;
+                    // The exporter evicts series idle for 10m, and a stalled
```
jgraettinger: nit: the session never stalled; I don't think the 10m expiry is a particular concern for a clock-skew metric.

jgraettinger (commented):
* First commit looks great!
  * Should we consider using a different word than "fuzz", given it's usage as term-of-art (including in this codebase) for fuzz testing? "remap" or "mix" ?

* I'm not sure the session gauge is the best placement, and prometheus best-practice is to use absolute unix time rather than relative time deltas.
   * Unix time is absolute, and is trivially converted into a time relative to either "now" or "sample time" by a PromQL querier.
   * Instead, what about instrumenting a gauge in the slice actor's try_log_request_tx ? The cohort and clock are readily available there, and it's the point of dequeue. Something like `shuffle_slice_last_source_published_at_time_seconds` following the precedent of our [OpenMetrics API](https://docs.estuary.dev/reference/openmetrics-api/#available-metrics)  (its [implementation](https://github.com/estuary/flow/blob/master/crates/control-plane-api/src/server/public/open_metrics.rs#L407))?

dgreer-dev:
>   * Should we consider using a different word than "fuzz", given it's usage as term-of-art (including in this codebase) for fuzz testing? "remap" or "mix" ?

Yeah that's smart.

> * I'm not sure the session gauge is the best placement, and prometheus best-practice is to use absolute unix time rather than relative time deltas.
>   
>   * Unix time is absolute, and is trivially converted into a time relative to either "now" or "sample time" by a PromQL querier.
>   * Instead, what about instrumenting a gauge in the slice actor's try_log_request_tx ? The cohort and clock are readily available there, and it's the point of dequeue. Something like `shuffle_slice_last_source_published_at_time_seconds` following the precedent of our [OpenMetrics API](https://docs.estuary.dev/reference/openmetrics-api/#available-metrics)  (its [implementation](https://github.com/estuary/flow/blob/master/crates/control-plane-api/src/server/public/open_metrics.rs#L407))?

Yeah actually that composes pretty cleanly.

New push in with those changes up!

jgraettinger (approved):
LGTM!

PR #3441, opened 2026-08-28
go/flow/keyhash: split the packed-key hash into its own package

jacobmarble:
**Description:**

Some connectors recompute Flow's packed-key hash to match shard routing. They can't import `go/flow` for it — that pulls in the gazette consumer stack — so they kept copies. estuary/connectors#5149 consolidated those copies.

This moves `PackedKeyHash_HH64` and its key into `go/flow/keyhash`, so there's one canonical home to point at. The golden-vector test comes along; it no longer needs etcd and cgo to run. Callers in `go/flow` and `go/shuffle` updated. No behavior change.

**Workflow steps:**

No user-facing change.

None.

**Notes for reviewers:**

- Kept the stuttery `PackedKeyHash_HH64` name so it matches the connectors package.
- Follow-up: connectors could import this and delete its copy.

### DIFF
--- a/crates/doc/src/extractor.rs
+++ b/crates/doc/src/extractor.rs
@@ -197,7 +197,8 @@ impl Extractor {
     /// The hash is the top 32 bits of a HighwayHash over the tuple's bytes
     /// using a fixed public (non-cryptographic) key.
     ///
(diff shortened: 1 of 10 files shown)


jgraettinger (approved):
LGTM

PR #3442, opened 2026-08-28
flowctl: `preview` runs on the V2 runtime, remove V1 preview

jgraettinger:
Replaces the V1-runtime-backed `flowctl preview` with the runtime-next + shuffle
implementation developed as `flowctl raw preview-next`, and deletes V1 preview
along with the runtime harness it was the sole consumer of.

`raw preview-next` is retained as a hidden, deprecated alias onto the same
`clap::Args` struct, so existing callers — estuary/connectors CI, the in-repo
pytest suite — keep working untouched.

## Commits

1. **`flowctl: preview runs on the V2 runtime`** — re-point `Command::Preview`,
   hide + deprecate the `raw preview-next` alias, delete
   `flowctl/src/preview/{mod,journal_reader}.rs` (−733).
2. **`runtime: remove the now-dead V1 preview harness`** — delete
   `runtime/src/harness/` and the private `exchange` combinator (−1131). Split
   out so the first commit *proves* the harness has no remaining callers and
   this one reviews as a pure deletion.

## Behavior changes

The flag surface is a strict superset (`--shards` and `--debug-port` are
additive; nothing is removed). Three behaviors do change:

* **Derivation stdout gains `["collection/name",<doc>]` framing**, which
  captures and materializations already used. Stripping the wrapper reproduces
  the old output exactly. Nothing downstream parses it. The three `site/`
  derivation tutorials show bare documents and are now stale — but `site/` no
  longer serves docs.estuary.dev (90e826189d8), so that correction belongs in
  `estuary/docs` and is deliberately not made here.
* **`--delay` with `--fixture` is now an error.** A fixture's transaction
  boundaries are fixed by its own commit markers, so a raised minimum
  transaction duration has nothing to batch.
* **Unset `--sessions` no longer equals `--sessions -1` for captures.** Unset
  holds an EOF'd session through the capture's poll interval as production does;
  `-1` stops at EOF (f259b8f6bf6).

## Why the module isn't moved to `preview/`

`flowctl preview` is backed by a module still named `raw/preview_next/`. That is
deliberate. #3416 renames five of that module's files out to a new
`crates/runtime-local` crate and rewrites the rest, so moving now means
rename-vs-rename conflicts on every one — and would land the wrong layout
regardless, since after that split the code deserving the name `preview/` is only
the flowctl-specific residue.

`git merge-tree` against the rebased #3415–#3419 stack reports **zero conflicts**
in either direction. The only file both touch is `raw/mod.rs` (this PR edits the
`PreviewNext` arm; #3417 adds a `Test` arm beside it), and the merged result was
checked by hand to be semantically correct.

A `TODO(johnny)` on `raw::Command::PreviewNext` records the deferred move, the
in-repo callers that must migrate first (`tests/preview/*.py` and the
`tests/soak/**` comments, both introduced by #3416), and the eventual removal of
the alias.

## Verification

* Both crates build clean under `mise` with no unused-code warnings — the
  compiler is the proof the harness is dead.
* `flowctl preview --help` carries the original doc text with the V2 flag set.
* `raw --help` no longer lists `preview-next`; `raw preview-next --help` still
  dispatches and reports itself deprecated.

Note that the `tests/*.py` snapshot suite could not verify anything: it is not
wired into CI, and has been broken since 85253a198cd swapped the fixture reader
without converting the checked-in fixtures. The framing claim above was instead
verified directly against the V1 and V2 snapshot artifacts.


williamhbaker (dismissed):
LGTM

williamhbaker (approved):

PR #3447, opened 2026-08-31
validation: stamp created_at on a task's first build

### LINKED ISSUE #3425: Some tasks lack built_spec.createdAt
All tasks should have a createdAt stamped in them.

dgreer-dev:
Closes #3425

This change stamps the first build from the publication Id's date, which is
minted moments before the control-plane Id. Once stamped, the date is
immutable: later builds carry it forward verbatim rather than re-deriving it,
so a connector's feature-flag resolution can never change under a running
task. Builds of pre-existing specs which lack the field stamp it from the
control-plane Id, as before.

Notes:

- `flowctl` local builds of never-published specs carry a far-future
  `createdAt` (~2090) from the sentinel pub Id. Behaviorally identical to
  "new"; published tasks carry their real date forward.


mdibaiee (dismissed):

williamhbaker (approved):
LGTM

PR #3448, opened 2026-08-31
go/bindings: restamp TestBuildCatalog snapshots with created_at

jgraettinger:
**Description:**

`TestBuildCatalog`'s Go snapshots are stale, and every open PR is failing `Platform Test` because of it.

#3447 (`validation: stamp created_at on a task's first build`, 35b7229532b) made a first build stamp `created_at` from its publication Id where it previously left the field empty. `TestBuildCatalog`'s fixtures are all first builds, so their capture and materialization specs now carry `created_at:"2090-05-21"` — the date embedded in the test's fixed `BuildId: "7777777777777777"`. That PR updated the Rust `insta` snapshots but not these `cupaloy` ones.

**Workflow steps:**

```
go test -tags libsqlite3 -run TestBuildCatalog ./go/bindings/
```

None.

**Notes for reviewers:**

Four snapshots move — the task specs. Collections and tests don't carry the field, and their snapshots are untouched. Each diff is exactly one inserted `created_at:"2090-05-21"` token and nothing else. The value is deterministic: it derives from the test's hardcoded `BuildId`, not from wall clock.


williamhbaker (approved):
LGTM

PR #3454, opened 2026-09-01
ci,local: one package set for local-stack binaries, and warm CI caches

jgraettinger:
The commit message carries the full reasoning and measurements. Notes for
review, and for after merge:

- **The cache benefit does not appear on this PR.** No master-scoped entry
  exists for `dekaf-test`, `platform-test` or `generated-artifacts-check`
  today. The first master run after merge writes them; pull requests opened
  after that are the ones that restore.

- **The first build on any checkout is a full rebuild.** Naming a different
  package set changes every dependency's `-C metadata` hash, so folding
  `data-plane-controller` into the shared set invalidates what a warm target
  directory already holds. One time, locally and in CI.

- **`local:supabase --no-block` and the `save-if`/`push` pairing** each have a
  matched half elsewhere in the file. Removing the `push` trigger without
  removing `save-if` writes no cache at all.

Validated on a local stack: 13/13 services accepting connections after 52s,
26/26 Dekaf e2e tests passing in 124s. CI on this PR exercises the reordered
Dekaf job, still on a cold cache.

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -11,7 +11,7 @@ capped-shard-reconfiguration = { max-threads = 3 }
 [profile.default]
 slow-timeout = { period = "30s", terminate-after = 4 }
 # Exclude e2e tests from Dekaf by default because they require special setup.
-# See mise/tasks/ci/dekaf-e2e for details. The materialization consistency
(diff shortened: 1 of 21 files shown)


williamhbaker (approved):
LGTM

PR #3456, opened 2026-09-02
shuffle: gate the startup heap drain on a Barrier from the Session

### LINKED ISSUE #3457: coordinate initial reads with journal list completion

williamhbaker:
**Description:**

At startup, a Slice could drain a document too early. The Session had not yet routed all of the journals from the initial listing. One of these journals could hold a document with an earlier adjusted clock or a higher binding priority. The
Slice must drain that document first.

Each binding has one listing owner. The listing owner now sends a `ListingSnapshotComplete` message for its binding. The Session waits for one of these messages from each binding. Then the Session puts one `InitialReadsStarted` message in the request queue for each Slice. This message comes after the initial `StartRead` messages. A Slice keeps its drain gate closed until the `InitialReadsStarted` message arrives.

See the commit messages for the details.

Closes https://github.com/estuary/flow/issues/3457

**Workflow steps:**

**Notes for reviewers:**

Tested against a local stack, on a derivation whose transforms have descending priorities, so a premature drain shows up as a priority inversion. With an 8s delay injected into the top-priority binding's listing, the drain order is `[alpha, beta, gamma]` with the Barrier and `[beta, gamma, alpha]` without it. Also checked: bindings with no journals still sync (empty partition selector, and a collection never written to), broker outage during startup, crash recovery, and journals that appear after the Barrier.


dgreer-dev (dismissed):
LGTM

crates/shuffle/src/session/actor.rs
```diff
@@ -16,8 +16,12 @@ pub struct SessionActor {
     pub session_response_tx: mpsc::UnboundedSender<tonic::Result<shuffle::SessionResponse>>,
     /// Per-shard channels for sending SliceRequest messages.
     pub slice_request_tx: Vec<mpsc::Sender<shuffle::SliceRequest>>,
-    /// FIFO queue of per-shard requests to be transmitted to their Slice channel.
+    /// FIFO queue of per-shard requests, with InitialReadsStarted after StartReads.
     pub slice_requests: std::collections::VecDeque<(usize, shuffle::SliceRequest)>,
+    /// Whether each binding's initial listing snapshot has completed.
+    pub listing_snapshot_complete: Vec<bool>,
```
jgraettinger: nit: I'm not sure this is worth the per-binding bool, as opposed to just keeping a counter of completed or remaining ones (it's just used for finer-grain error reporting of what can fundamentally only be an implementation error, right?). We can still detect if we see more than `num_bindings`.

jgraettinger (dismissed):
LGTM % nit, elegant!

crates/shuffle/src/session/actor.rs
williamhbaker: Yep a counter works, that's a little cleaner. Made that update.

jgraettinger (approved):
still LGTM

PR #3460, opened 2026-09-02
shuffle, runtime-next: send panic statuses to peers

### LINKED ISSUE #3461: better error messages for panics

williamhbaker:
**Description:**

Panics in spawned shuffle and runtime-next handlers currently drop the response sender, causing peers to see a successful end-of-stream instead of an error. This adds `proto_grpc::catch_panic` to return bounded gRPC statuses for handler errors and panics, and applies it to all spawned shuffle, leader, and shard handlers. Panic locations and enabled backtraces remain in the local process logs.

Closes https://github.com/estuary/flow/issues/3461

**Workflow steps:**

**Notes for reviewers:**

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6497,6 +6497,7 @@ version = "0.0.0"
 dependencies = [
  "anyhow",
  "bytes",
+ "futures",
  "labels",
  "proto-build",
  "proto-flow",
--- a/crates/proto-grpc/Cargo.toml
+++ b/crates/proto-grpc/Cargo.toml
@@ -16,6 +16,7 @@ tokens = { path = "../tokens" }
 
 anyhow = { workspace = true }
 bytes = { workspace = true }
+futures = { workspace = true }
 tokio = { workspace = true }
 tonic = { workspace = true }
 tonic-prost = { workspace = true }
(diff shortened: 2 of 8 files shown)


dgreer-dev (approved):
LGTM, but a pair of nits:

Maybe emit a `tracing::error!` so that the error makes it into the tracing
subsystem and not just stderr?

Also the error code ends up being `Unknown`, which is already crowded:
`Error::is_transient` in `crates/gazette/src/lib.rs` has two arms matching
`Code::Unknown` plus a magic message substring to recover semantics the code
didn't carry. Tonic has `Internal` which seems to map to a `panic!` pretty well?

williamhbaker:
Thanks, I played around with both of those a bit. On the `tracing::error!`: the status `catch_panic` returns already reaches the task ops logs (Go unwraps it and surfaces it as shard failed) so an extra tracing line would _mostly_ duplicate it. On the code: `Unknown` is a convention for user-facing messages (see `crates/runtime-next/src/lib.rs:162-178`)...`status_to_anyhow` unwraps it so context accumulates across hops, vs. a relayed `Internal` that would drop it from the final message.

PR #3463, opened 2026-09-04
product name: use "Estuary" instead of "Estuary Flow"

jwhartley:
The product is "Estuary", not "Estuary Flow" ([context](https://estuaryworkspace.slack.com/archives/C012S5U0WP2/p1788452319856409)).

Two of these print in `--help`, via clap's `#[command(about, ...)]` on the `flowctl` and `dekaf` `Cli` structs. The rest are READMEs and comments, which agents read and imitate when drafting documentation.

Left alone as identifiers, not the product name: `LICENSE-BSL`, `pyproject.toml`, the `estuary-flow-poc` bucket in fixtures and `seed.sql`, and the `estuary-flow-web` tarball in CI.

Nothing under `site/` is touched, since that tree is deprecated and `estuary/docs` is the live source. #2964 edits `site/docs/getting-started/getting-started.md` for the same rename and may want redirecting there.

Companion PR: estuary/connectors#5191

### DIFF
--- a/crates/dekaf/src/main.rs
+++ b/crates/dekaf/src/main.rs
@@ -24,7 +24,7 @@ use tokio::io::{AsyncRead, AsyncWrite, AsyncWriteExt};
 use tokio::net::TcpStream;
 use url::Url;
 
-/// A Kafka-compatible proxy for reading Estuary Flow collections.
+/// A Kafka-compatible proxy for reading Estuary collections.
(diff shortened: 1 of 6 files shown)


aeluce (approved):
I hadn't been bothering with more internal-facing READMEs and documentation but good catch that some of the comments get printed out in the command line.

PR #3465, opened 2026-09-08
Derivation connector cleanups, user-defined environment variables, and shared gRPC helpers

jgraettinger:
A grab-bag of derivation-connector and runtime-next cleanups, which set up for upcoming V2 connector proxy work. See individual commit messages for details:

- **proto-grpc**: centralize channel dialing, status conversions, protocol verification, and shared buffer / message-size constants; drop the duplicate implementations in `gazette`, `shuffle`, and `runtime-next`.
- **runtime-next**: remove the never-used unary Spec / Discover / Validate requests (and their `log_level` fields) from the V2 protocol. Field numbers are reserved, not reused, since the protocol can skew across a rolling deploy.
- **derive-sqlite**: take the VFS URI as an explicit `connector()` argument rather than decoding it from a runtime-internal protobuf extension.
- **derive-python / derive-typescript**: fix line-buffer handling between unary requests; stop emitting unset `Response` variants as explicit nulls; add a user-defined `environment` map, with validation policy shared via the new `connector-environment` crate.
- **derive-typescript**: bump Deno to 2.9.6.


jgraettinger:
QA'd on a local stack, including upgrading the V1 ops derivations in-place to the built derive-typescript image.

crates/models/src/derive_typescript.rs
```diff
@@ -13,6 +14,9 @@ pub struct DeriveUsingTypescript {
     /// interface.
     #[schemars(schema_with = "DeriveUsingTypescript::module_schema")]
     pub module: super::RawValue,
+    /// # Environment variables made available to the TypeScript module.
+    #[serde(default, skip_serializing_if = "BTreeMap::is_empty")]
+    pub environment: BTreeMap<String, String>,
```
dgreer-dev: `environment` on a V1 derivation lands on the frozen `:dev` image, which has no
`deny_unknown_fields`, so it gets silently dropped: publish succeeds and the var
never shows up. Do we care, or is V2 close enough that gating this on
`enable-runtime-v2` isn't worth it?

crates/proto-grpc/src/status.rs
```diff
@@ -68,6 +68,90 @@ pub fn anyhow_to_status(err: anyhow::Error) -> tonic::Status {
...
+            None => Ok(()),
+            Some(value) => Err(self.fail_msg(value)),
+        }
+    }
+
+    #[must_use]
+    #[cold]
+    pub fn fail_msg<T: serde::Serialize>(&self, msg: T) -> anyhow::Error {
+        let mut rendered = serde_json::to_string(&msg).unwrap();
+        rendered.truncate(4096);
```
dgreer-dev: Extremely unlikely, but `truncate(4096)` panics if byte 4096 lands mid-UTF8-character. `bound_message` above already backs off to a char boundary - worth reusing?

dgreer-dev (dismissed):
LGTM (give or take a pretty unlikely panic)

williamhbaker:
Not sure if it's relevant here, but a reminder about this, since the PR description made me think of it: https://estuaryworkspace.slack.com/archives/C03Q9A3GFEE/p1781845124452399

jgraettinger:
> Not sure if it's relevant here, but a reminder about this, since the PR description made me think of it: https://estuaryworkspace.slack.com/archives/C03Q9A3GFEE/p1781845124452399

Yep, noted. The intention is to add validation checks that require V2 if either `secrets` or `environment` is populated.

The purpose of this PR is to inertly land connector-side changes which currently block final end-to-end testing of secrets (because that testing wants to use `:stable` tags, which can only be built by merging + CI).

crates/proto-grpc/src/status.rs
jgraettinger: 🤦 Good catch, thank you! Updated.

dgreer-dev (dismissed):
LGTM

dgreer-dev (approved):
LGTM

PR #3469, opened 2026-09-09
proto-flow: gather connector Request/Response fields into a `oneof kind`

jgraettinger:
Gathers the capture/derive/materialize `Request` and `Response` sub-message fields
into a prost `oneof kind`, and boxes the two largest variants — prost embeds each
field by value, so these structs were 3-4 KB apiece. The `.proto` files are
unchanged apart from comments; the commit messages have the detail.

Wire and JSON encodings are identical across the refactor: the twelve per-variant
regression snapshots added in the first commit are byte-for-byte unchanged, which
is the acceptance criterion.

QA'd on a local stack: the V2 soak chain ran clean across ~9k accounts and a mid-run
reactor restart, with all six invariants at zero and schema inference converging as
documented. The derive-python/derive-typescript `environment` feature was also
exercised on both `:stable` and locally-built `:local` connectors.


williamhbaker (approved):
LGTM, neat!

PR #3470, opened 2026-09-09
runtime-next: make per-shard tokio worker threads configurable

jgraettinger:
Each assigned shard builds its own `TaskService` and tokio runtime, which until now was pinned to a single worker thread. This reads the count from an optional `FLOW_RUNTIME_WORKER_THREADS` instead, so a shard that turns out to be async-executor bound can be given more workers without a rebuild.

The default stays at one: the count multiplies by a reactor's shard count, so a densely packed reactor would otherwise spawn thousands of mostly-idle workers. Malformed and zero values fall back to that default silently.

V2 only — `crates/runtime/src/task_service.rs` and the sidecar's runtime are untouched.


williamhbaker (approved):
LGTM

PR #3478, opened 2026-09-10
go.mod: bump gazette to latest

williamhbaker:
**Description:**

Brings in the changes from https://github.com/gazette/core/pull/495.

They surface a dead recovery-log recorder through the `set_sync(true)` RocksDB write, which runtime-next relies on because it bypasses the consumer framework's `StartCommit`. A fenced recorder now fails that write instead of retrying REGISTER_MISMATCH forever, and a cancelled one fails it instead of reporting a checkpoint as durable which never committed to the log.

Tested on a local stack: Ran the soak test and synthetically fenced the live shard repeatedly. Previously it would hang forever, but now crashes and restarts, and the soak test invariants are maintained. The motivating shard context cancellation case was similarly tested, although with a modified `flowctl-go` binary to get the timing of the cancellation just right, which is not an easy thing to simulate with the reactor code unmodified.

**Workflow steps:**

**Notes for reviewers:**


jgraettinger (approved):
LGTM, thanks!

PR #3480, opened 2026-09-10
data-plane-controller: publish private link DNS details after PulumiUp1

### LINKED ISSUE #3477: data-plane-controller: write private link DNS details after PulumiUp1 instead of end-of-converge
## What to build

A private link's endpoint is created during `PulumiUp1`, and est-dry-dock's `link_results` export carries its final AWS-assigned DNS names at that moment. The controller does not write them to `internal.data_plane_private_links` until the end of the Ansible stage, roughly ten minutes later (`PulumiUp1`, a 300s DNS propagation wait, the Ansible run, then the write drained on the next poll during `PulumiUp2`).

For that whole window the row reads `status = 'pending'` with `details = null`, while the endpoint already exists and its DNS name is final. Support is usually adding a link in response to a customer who is waiting on that DNS name to configure their own side, and the only way to get it early is to read the raw DPC logs for the `pulumi-up-one` diff. That is a poor interface for a value the control plane already has.

Write each link's `details` as soon as `PulumiUp1` completes, so the row carries the DNS names within about a minute of the endpoint being created.

## Why this looks safe

Nothing after `PulumiUp1` affects a private link:

- Ansible has no private-link, VPC-endpoint, or `vpce` references at all. It configures servers; the endpoint is a VPC-level resource with no host-side configuration.
- `PulumiUp2` is scoped to scaling: it updates DNS records for resources that have now started and destroys scaled-down resources that have now stopped. A link insert scales nothing.
- Both `pulumi up` invocations run the same program, so `PulumiUp2` re-evaluates the endpoint and finds no diff.
- A link that can fail has already failed by `PulumiUp1`: `PrivateLinkError` is caught per link and produces a `failed` result with its error.
- The 300s `AwaitDNS1` wait is one `DNS_TTL_ACTUAL`, waiting out the 300s TTL on records est-dry-dock itself publishes for replaced and scaling-down hosts. The endpoint's `vpce-*.vpce.amazonaws.com` name is published by AWS at creation and is not one of those records, so the wait does not gate it.

One edge case: adding the first link to a plane with no bastion and no allowed CIDRs flips `uses_private_network`, which changes network topology and may pull in resources Ansible then configures. Worth confirming this does not need different treatment from the steady-state case.

## Sketch

The pieces this needs mostly exist. `pinned_links` is already pinned at exactly `PulumiUp1`, with a comment noting that later polls retain those pins "for attributing its exports", so the generation guard works unchanged. `write_private_link_statuses` already takes pinned links plus results, already guards on generation, and is idempotent, so the existing end-of-Ansible write can stay and simply overwrite with identical content.

That leaves: run `pulumi stack output` in `on_pulumi_up_1` (the same call already exists in `on_ansible`), stage the parsed `link_results` in a new `Outcome` field drained the way `publish_exports` is, and pass it to the existing write. `ControlExports` already tolerates a missing `link_results`, so a partial stack output at this stage parses.

No schema change is needed. `pending` is already a legal value of the `status` check constraint, and `details` is a plain nullable `jsonb` column independent of it.

## Open questions

These are design calls for whoever owns the private-links model, not implementation details:

1. **Should `status` also become `provisioned` at `PulumiUp1`?** The evidence above says `PulumiUp1` genuinely is the moment a link is provisioned, which would make an early `provisioned` accurate. But the current design deliberately means "a completed converge observed this", and leaving `status` at `pending` preserves that invariant while still delivering the DNS names early. Leaving it `pending` is the smaller change and the safer default; confirming which is wanted should come before implementation.

2. **What should `observed_at` do on an early write?** The current UPDATE sets it to `now()` unconditionally alongside `status`, `details` and `error`. If `status` stays `pending`, stamping `observed_at` early contradicts "a converge observed this", so the early path likely needs to leave it null.

## Acceptance criteria

- [ ] After a link's endpoint is created in `PulumiUp1`, its row's `details` carries the same `dns_entries` and `service_name` the end-of-converge write would have produced.
- [ ] The generation guard still holds: a link whose `config` is edited mid-converge is not stamped by the converge that pinned the earlier generation.
- [ ] The existing end-of-Ansible write still runs and is a no-op in content for links already written at `PulumiUp1`.
- [ ] A link that fails during `PulumiUp1` still records `failed` with its error.
- [ ] The behaviour of `status` and `observed_at` on the early write matches whichever answer the open questions above settle on.

## Blocked by

None (can start immediately), though the two open questions should be answered before implementation.

jwhartley:
Closes #3477.

## What changed

`on_pulumi_up_1` now reads the stack output and stages its `link_results`, which the outcome writes into `internal.data_plane_private_links` at the end of that poll. Each link's row carries its `dns_entries` and `service_name` within about a minute of the endpoint being created, instead of after the 300s DNS wait and the Ansible run.

`status` and `observed_at` are deliberately left alone on the early write. `provisioned` means "a completed converge observed this", and publishing the DNS names early does not change that, so a successful link stays `pending` with a null `observed_at` until the existing end-of-Ansible write settles it. A `failed` result is the exception: a `PrivateLinkError` is caught per link by the Pulumi program and no later stage revisits it, so the failure records in full immediately rather than sitting hidden for ten minutes. This is the `converge_complete` flag on `write_private_link_statuses`, and it is the only behavioural branch in the SQL.

The duplicated `pulumi stack output` invocation is factored into a `stack_output` helper, now called from both `on_pulumi_up_1` and `on_ansible`.

## Why this is safe

Nothing between `PulumiUp1` and the end of the converge affects a private link:

- Ansible has no private-link, VPC-endpoint, or `vpce` references. The endpoint is a VPC-level resource with no host-side configuration.
- `PulumiUp2` is scoped to scaling, and both `pulumi up` invocations run the same program, so it re-evaluates the endpoint and finds no diff.
- The 300s `AwaitDNS1` wait covers records est-dry-dock publishes itself for replaced and scaling-down hosts. The endpoint's `vpce-*.vpce.amazonaws.com` name is published by AWS at creation and is not one of those records.

The generation guard needed no change. `pinned_links` is already pinned at exactly `PulumiUp1`, and `write_private_link_statuses` already guards on it and is idempotent, so the end-of-Ansible write overwrites with identical `details`. No schema change: `pending` is already legal and `details` is an independent nullable `jsonb` column.

## How it was verified

One new sqlx test against the real migrations, alongside the existing `write_private_link_statuses_records_and_guards_generation`:

- `write_private_link_statuses_early_publishes_details_only` — the early write publishes `dns_entries` while `status` stays `pending` and `observed_at` stays null; a failed link records status, error and `observed_at` immediately; the subsequent end-of-converge write then flips the successful link to `provisioned` with the same details.

The generation guard is not re-tested on the early path: the guard is a `JOIN` on generation that this change does not touch, and the pre-existing test already covers it.

Full crate suite passes (14 unit + 4 integration). The three command-trace snapshots each gain one `pulumi stack output` entry at `PulumiUp1`; the diffs are purely additive, with nothing removed or reordered. `.sqlx` regenerated: one query replaced, no other churn.

Note: `cargo clippy` fails in `crates/models` on a pre-existing issue (`crates/models/src/status/collection.rs:52`, non-semver `since` field), present on a clean `master` checkout and unrelated to this change.


jwhartley:
FYI @jshearer

jshearer:
Hey @jwhartley! I took a look at this, and had a thought:

Links are already marked `provisioned` before the converge completes. The final write happens before `PulumiUp2` and `AwaitDNS2`. As you note, nothing after `PulumiUp1` changes the link results: `AwaitDNS1` just sleeps, and Ansible doesn’t handle links. The final write applies the same results staged at `PulumiUp1`, ten minutes later.

Can we just write the full row at `PulumiUp1` using `write_private_link_statuses` unchanged? That gives us the same end state without the flag, the `CASE` expressions, the `failed` exemption, or the duplicate private-link write.

On AWS, for example, `provisioned` already doesn’t mean “connectable.” The provider’s create waiter treats `pendingAcceptance` as a terminal state, and we don’t set `auto_accept`, so `provisioned` can be reported before the customer accepts the connection request. Reporting it earlier doesn’t change that. Making `provisioned` mean “accepted and reachable” would require a connection-state check, which is a separate change involving est-dry-dock and DPC.

jwhartley:
@jshearer that sounds much simpler if there's nothing else that's acting on it - updated, thanks

PR #3481, opened 2026-09-10
data-plane-controller: new task type for testing separate builds

qaoj:
**Description:**

Adds a new task type to the executor in automations, adds a configurable option on the DPC to make use of it. 

This allows us to run a separate deployment of the data-plane-controller against the task types that have been assigned to this type manually.

Changes the deployment behavior of DPC, no longer deploys either deployment automatically, now requires a manual fire of the workflow.

**Workflow steps:**

In our internal.tasks table, you can now set something that was previously of `task_type` 1 to `task_type` 13 and it will run under the other DPC deployment.

N/A

**Notes for reviewers:**

Verified this locally by running 2 copies of DPC against local supabase, one with the dev arg and one without. Confirmed they only operated on their respective task types.

### DIFF
--- a/.github/workflows/deploy-data-plane-controller-dev.yaml
+++ b/.github/workflows/deploy-data-plane-controller-dev.yaml
@@ -1,10 +1,23 @@
-name: Deploy data-plane-controller
+name: Deploy data-plane-controller (dev)
 
-# Deployed manually rather than on every master build, so a candidate image can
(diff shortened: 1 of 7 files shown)


mwillman-estuary (approved):
Looks good to me, but I'm no Rust expert.  I don't see any logical errors.

PR #3482, opened 2026-09-10
control-plane-api: consolidate the authorization gate onto Envelope

bbartman:
**Description:**

Preparatory refactor for #3376 (capability-masked tokens). Consolidates the user-authorization gate that REST handlers and GraphQL resolvers use onto `crate::Envelope`, so the mask can later be threaded through one place instead of many.

Before this change the same evaluate-then-`authorization_outcome` dance existed in three shapes: a GraphQL-local `verify_authorization(env, prefix, capability)` free function in `graphql/mod.rs`, a pure `evaluate_names_authorization(snapshot, claims, capability, names)` free function in `server/mod.rs`, and nine call sites (alerts, alert_configs, live_spec_refs, open_metrics, status, update_l2_reporting, create_data_plane, authorize_user_secret, storage_mappings) that inlined it by hand. All of them now go through inherent methods on `Envelope`:

- `verify_authorization(prefix, capability)` for the ~30 single-name gates.
- `verify_authorization_iter(names, capability)` for the two multi-name gates (`/api/v1/catalog/status` and `liveSpecs`).
- `verify_storage_mapping_authorization(prefix, data_planes)` for the three storage-mapping mutations, replacing that module's local `evaluate_authorization`/`check_authorization` pair. Both halves of that policy (user Admin on the prefix, prefix Read on each data plane) still evaluate as one `AuthZResult` through one `authorization_outcome`.
- `pub(crate) evaluate_names_authorization(names, capability)` as the shared pure policy half.

The methods return `crate::ApiError`, so REST handlers use them directly and GraphQL resolvers convert through the existing `From<ApiError> for async_graphql::Error`. Denial messages and status codes are unchanged. The old free functions and the `pub(crate)` re-export in `lib.rs` are deleted.

The `Envelope` wrappers route their policy result through a small shared `user_authorization_outcome` helper, which lets an `Unauthenticated` result escape as a terminal error before `authorization_outcome` runs. This preserves prior behavior. The old gates evaluated `env.claims()?` before the policy, so a missing bearer failed fast with 401; moving the claims lookup inside the evaluator would otherwise have sent anonymous callers down the 307 retry path and revoked the Snapshot. The dispatch deliberately lives in the user-policy wrappers and not in `authorization_outcome` itself: the task-token policies (`authorize_task`, `authorize_task_secret`, `authorize_dekaf`) report a Snapshot-dependent `Unauthenticated` when no known data-plane key validates the token, because the issuing plane may be newer than the Snapshot, and those must stay provisional. `test_task_decrypt_secret_retries_stale_snapshot` pins that.

**Workflow steps:**

No user-facing change. Every authenticated request that was accepted or denied before is accepted or denied identically. Unauthenticated requests still receive 401 on REST. On GraphQL the unauthenticated error now renders through `ApiError` like every other authorization error (`Unauthenticated: <message>` with the `ApiError` as source) rather than through `tonic::Status`'s `Display`; the code and behavior are the same.

None. Internal control-plane-api refactor with no API or schema change.

**Notes for reviewers:**

- The `server/mod.rs` unit test `test_evaluate_ops_admin_gate` moved to `envelope.rs` with its three cases and exact denial message intact. `envelope.rs` also gains an `Envelope` test constructor (JWT sign-then-verify for `Verified` claims, a fixed Snapshot watch, a lazy pool) and five further tests: admit on grant, deny as `PermissionDenied`, multi-name denial names the offending member, unauthenticated escapes the provisional path without revoking the Snapshot, and a real denial after the Snapshot still yields `AuthZRetry`. The unauthenticated test was confirmed to fail with the dispatch removed.
- Coordination with the open #3376 stack: #3406 (task 3c) edits the free `evaluate_names_authorization` in `server/mod.rs` and its test module, the GraphQL `verify_authorization`, and storage-mappings' `check_authorization`, all of which this PR deletes or moves. Whichever lands second will need to rebase the mask pre-check into the new `Envelope` methods.
- Intentionally not touched: the pre-existing typo in the storage-mapping denial message (`is not an authorized as an Admin`). Fixing it is a visible message change and belongs in its own commit.
- Verified with `cargo check -p control-plane-api --tests` (clean), `cargo doc --document-private-items` (no unresolved intra-doc links), and `cargo nextest run -p control-plane-api envelope::tests storage_mappings` (all passing).


crates/control-plane-api/src/server/public/graphql/mod.rs
```diff
@@ -74,30 +74,6 @@ fn may_access(
     ))
 }
```
GregorShear: any reason not to move this to the envelope, too?

crates/control-plane-api/src/envelope.rs
```diff
@@ -172,6 +174,165 @@ impl Envelope {
...
+        Ok((None, ()))
+    }
+
+    /// Errors unless the current user may bind `data_plane_names` into a storage
+    /// mapping under `catalog_prefix`. That requires two grants evaluated as one
+    /// policy against a single Snapshot: the user holds Admin on the prefix, and
+    /// the prefix itself holds Read on every named data plane. Both halves pass
+    /// through one authorization outcome so a stale denial of either yields a
+    /// single refresh-and-retry decision.
+    pub async fn verify_storage_mapping_authorization(
```
GregorShear, line 312: I question whether the storage mapping helpers should be in the envelope - are there generic pieces we could define here that would compose back in `graphql/storage_mappings.rs`?

crates/control-plane-api/src/envelope.rs
```diff
@@ -172,6 +174,165 @@ impl Envelope {
...
+        prefix: &str,
+        capability: impl Into<models::authz::CapabilitySet> + std::fmt::Display + Copy,
+    ) -> Result<(), crate::ApiError> {
+        let policy_result = self.evaluate_names_authorization([prefix], capability);
+        self.user_authorization_outcome(policy_result).await
+    }
+
+    /// This does the same thing as verify authorization just for a list of
+    /// prefixes instead of one.
+    pub async fn verify_authorization_iter<Iter, S>(
```
GregorShear, line 226: would love to take this opportunity to refine the menu of authz functions we make available. Is this a good list?
```
verify_authorization
verify_authorization_iter
user_authorization_outcome
evaluate_names_authorization
may_access
```

GregorShear: FWIW i've struggled with and punted on this a couple times before. Would be worth chatting about in slack on monday

GregorShear:
Thanks for pulling this PR together - feels way easier to reason about and we're on the right track

crates/control-plane-api/src/envelope.rs
bbartman: Ok sounds, good. To be fair that is the list but here are other functions with the same name as those in storage_mappings in multiple locations that I didn't touch

GregorShear: Right that's my point. Here's the current list - is it good or are there changes we should make?

IMO those function names could all mean the same thing.

bbartman: There were a lot of them and they weren't in a path that used the envelope. But yeah I understand. Let's chat Monday.

bbartman: For context, here are the other `evaluate_authorization` are declared:
1) `crates/control-plane-api/src/server/authorize_dekaf.rs` 
2) `crates/control-plane-api/src/server/authorize_task_secret.rs`
3) `crates/control-plane-api/src/server/authorize_task.rs`
4) `crates/control-plane-api/src/server/authorize_user_task.rs`
5) `crates/control-plane-api/src/server/authorize_user_prefix.rs`
6) `crates/control-plane-api/src/server/authorize_user_task.rs`

For the refactoring I explicitly focused on graphQL and stayed away from the other endpoints. If you would like I can take a stab at doing a refactoring for the other endpoint's `evaluated_authorization` functions in a separate PR.

crates/control-plane-api/src/server/public/graphql/mod.rs
bbartman: No, it was an oversight. I fixed and pushed the change.

crates/control-plane-api/src/envelope.rs
bbartman: OK so after some refactoring I was able to refactor some of the code between, 3 of the other uses of evaluate_authorization. It's not much. But when I looked into how the other three instances are used there really isn't a good way to refactor those into something more easily read.

I will say that it should be possible to improve responses on an non-graphQL endpoint if we were to enable us to create a specially formatted response type that could masquerade as an error but get reformatted as part of the response if it's returned as an error message. This would improve readability for most functions that use `env.authorization_outcome` which is everyone. I'm not sure if we want to go that route or not because while it would maintain the behavior I'm not sure what it would look like for something like graphQL. I'll look into it more.

GregorShear: Looking into this more I think we want to go even deeper/narrower with the change, so that `tables::UserGrant::is_authorized` gets a new optional capability_mask parameter, and the envelope wraps that with a helper function. Then all of the existing callers of `is_authorized` either call the envelope function if it's in the context of a request and the token/claims/capability_mask is present, or if it's not in the context of a request ([here for example](https://github.com/estuary/flow/blob/9c1e6e6b0b494593a326bf1db54ffca5bf2b437a/crates/agent/src/discovers.rs#L177)) then we continue to call `is_authorized` directly with None capability_mask)

i think this means we can leave whatever logic there is around authz results and retrying after snapshot refresh - all stuff i think we could improve on but happy to leave it out of scope here.

bbartman: This is a straight refactoring. The mask addition change is done later in the PR stack.

bbartman: I should add that it's not optional, the option is basically a masked token or a unrestricted token mask. Or all capabilities, the mask doesn't add capabilities it only enables ones that the user could already do.

PR #3483, opened 2026-09-10
docs: correct default read access to ops logs and stats

midianyoseph:
The authorization page said a tenant's prefix is granted read access to its logs under a sub-prefix of `ops/`, with `ops/acmeCo/` as the example. Logs and stats are provisioned per data plane, not per tenant, so that grant doesn't exist.

Drops the claim and adds an "Access to logs and stats" subsection: public planes write to shared `ops/tasks/public/<plane>/` collections with no default read grant (use `flowctl logs --task` / `flowctl raw stats --task`), private planes get a read grant on `ops/tasks/private/<tenant>/`.


jonwihl (approved):
LGTM

PR #3488, opened 2026-09-11
control-plane-api: thread the capability mask through GraphQL authorization

bbartman:
Part of #3376. Stacked on #3486 (masked user grant walk).

Threads the bearer's `CapabilityMask` from the `Envelope` into every authorization walk reachable from the GraphQL endpoint. Until now every call site above `tables` pinned `ALL_CAPABILITIES`.

## Scope

- **In**: `Envelope::evaluate_names_authorization` and `evaluate_storage_mapping_authorization` (behind `verify_authorization`), `attach_user_capabilities`, `authorized_prefixes` / `filtered_authorized_prefixes`, `may_access`, and the `prefixes`, `storageMappings`, `dataPlanes` resolvers. Through the shared free functions this also masks the list queries for live spec refs, live specs, alert configs, invite links, secrets, and service accounts.
- **Accepted spillover**: REST callers of `verify_authorization` (`create_data_plane`, `update_l2_reporting`, `authorize_user_secret`, `open_metrics`) are masked too, since the shared policy fn reads the mask.
- **Out, still pinned**: executor paths (`live_specs::partition_by_authorization`, `Snapshot::user_capability` / `is_user_authorized`, agent discovers) and the REST `/authorize/user/*` and `status` handlers. The `Snapshot` wrappers are documented as evaluating unmasked.

## Behaviour

- **Gates** (the two `Envelope` policy fns) run a `masked_shortfall` pre-check before the walk. If the mask withholds any required bit the request is denied naming those bits in claim vocabulary, in `Capability` declaration order, with no user identity:

  ```
  token does not enable capabilities [SpecEdit, Delegate] required to access prefix or name 'acmeCo/'
  ```

  The shortfall is a function of the token alone and never consults grants, so it discloses nothing about the bearer's authorization. An empty shortfall defers to the walk, whose denial is unchanged. The Status flows through the ordinary `authorization_outcome`, so it lands terminally only against a current Snapshot and takes the provisional 307 retry otherwise.
- **Filters** (everything else) narrow silently.
- `userCapability` still reports the literal legacy column of a reached grant. The mask governs reachability, not what a reached grant reports; attenuated reporting belongs to the legacy-decision migration.
- `Envelope::capability_mask()` borrows rather than consuming the Envelope.

## Commits

1. `Envelope` gates: mask into both policy fns, `masked_shortfall`, `&self` accessor, shared grant-walk test fixtures in `test_server`. Compiles and passes on its own.
2. GraphQL wiring: explicit `mask` parameter after the user identity on the free fns, resolver sites, `Snapshot` doc notes, `make_masked_access_token`, DB-backed tests.

## Tests

- `envelope.rs`: both gates honor the mask; shortfall message; `Delegate` confinement; no grant disclosure; provisional retry still taken when started after the Snapshot.
- `authorized_prefixes.rs` and `server/mod.rs`: raw-table tests under a `Viewer` mask; existing tests pass `ALL_CAPABILITIES` explicitly.
- DB-backed GraphQL via `test_server`: `prefixes` capabilities attenuate and role-grant prefixes vanish without `Delegate`; `dataPlanes` hidden without `Delegate`; `storageMappings` narrows; `tenant` denied through `verify_authorization` with the shortfall message.
- `may_access` has no GraphQL-level test: every field it gates needs `ViewDataPlanePrivateNetworking`, which the `Read` pre-filter already requires, so no mask can reach a gated field yet fail the gate.

## Not in this PR

Typed definitive denial that bypasses the retry path with a structured 403 body, and the legacy-decision migration.


bbartman:

PR #3490, opened 2026-09-14
connector: a V2 connector service, served in-process and over the network

jgraettinger:
Introduces `connector.Connector`: one protocol which owns every part of running
a Flow connector — extracting and unsealing its endpoint configuration,
injecting IAM credentials, dispatching to a docker image / local subprocess /
in-process connector, pumping its logs, and tearing it down in the right order.
It's served two ways from a single `Service`: in-process without a wire hop, and
over gRPC on each `TaskService`'s Unix domain socket and, through a Go
pass-through proxy, on the reactor's public address.

Every *internal* caller is moved onto it: `runtime-next`, the local runtimes,
catalog tests, and flowctl all route connectors through a `Router` rather than
starting them directly.

**No control-plane component dials the network service yet.** The V1
`ProxyConnectors` path is retained and is still what the agent uses, so a
reactor built from this branch serves today's control plane unchanged. The new
surface answers only bearers minted by the reactor itself.

This is therefore a deployment unit on its own: roll it out to reactors first,
then deploy the components which dial it.

Individual commit messages carry the details.

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -17,13 +17,28 @@ slow-timeout = { period = "30s", terminate-after = 4 }
 default-filter = 'not binary_id(/dekaf::e2e/) and not binary_id(/materialize-consistency::scenarios/)'
 
 [profile.dekaf-e2e]
-# Migration tests excluded due to 8-minute timeout.
+# Migration tests excluded due to 8-minute timeout; run them with the
+# dekaf-e2e-migration profile below (`ci:dekaf-e2e-run --migration`).
 default-filter = 'binary_id(/dekaf::e2e/) and not test(/migration/)'
 # These tests publish specs, wait for shards, and consume through Dekaf, so
 # they're routinely slower than the default 30s/120s. Under CI load the slowest
 # have been observed at ~80s, which is close enough to the inherited 120s
 # terminate-after to become its own source of flakiness.
 slow-timeout = { period = "60s", terminate-after = 5 }
+
+# The migration scenarios the dekaf-e2e profile excludes, run on demand rather
+# than in CI.
+#
+# A migrated shard cannot go primary on the target plane until the source plane
+# has persisted its recovery-log fragments (up to 3 minutes, on a Gazette race)
+# and the target's fragment refresh has picked them up (5 minutes by default).
+# Minutes of waiting is the expected case, not a hang -- see
+# SHARD_PRIMARY_TIMEOUT in crates/dekaf/tests/e2e/migration.rs. The kill
+# threshold clears that 8-minute in-test bound with headroom, so a genuinely
+# stuck test still dies rather than running to the harness timeout.
+[profile.dekaf-e2e-migration]
+default-filter = 'binary_id(/dekaf::e2e/) and test(/migration/)'
+slow-timeout = { period = "120s", terminate-after = 10 }
 
 [profile.consistency]
 default-filter = 'binary_id(/materialize-consistency::scenarios/)'
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -25,6 +25,9 @@ Use regular `cargo` and `go` tools to build crates. Run tests via the `mise`
 tasks below, which declare their own prerequisites (submodules, Supabase,
 gazette, BigTable).
 
+Stop a local stack before running `control-plane-api` crate tests (some tests
+race the stack over DB tables), and at the completion of a QA run.
+
 ```bash
 # Rust tests. NOT `cargo test`, which ignores the `serial-db-tests` group in
 # .config/nextest.toml and lets DB-backed tests truncate each other's rows.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6572,6 +6572,7 @@ dependencies = [
  "anyhow",
  "bytes",
  "futures",
+ "insta",
  "labels",
  "proto-build",
  "proto-flow",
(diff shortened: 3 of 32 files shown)


crates/connector/src/protocol.rs
```diff
@@ -0,0 +1,431 @@
...
+        .transpose()?
+        .flatten()
+    {
+        let tokens = iam_config.generate_tokens(&ctx.task_name).await?;
+        *initial_config_slot = tokens.inject_into(initial_config_slot)?.to_string().into();
+
+        iam_token_restart_at = Some(proto_flow::as_timestamp(crate::token_restart_deadline(
+            std::time::SystemTime::now(),
+            tokens.expires_at(),
+        )));
```
dgreer-dev, line 215: Total nit, but we should `zeroize(tokens)` here

jgraettinger: discussed over slack

crates/connector/src/serve.rs
```diff
@@ -0,0 +1,318 @@
...
+                    }
+                }
+                Some(Err(status)) => break Err(crate::status_to_anyhow(status)),
+                None => break Ok(()),
+            },
+        }
+    };
+
+    // Killing an image connector closes stderr, allowing log draining to finish.
+    std::mem::drop(guard);
```
dgreer-dev, line 256: If the in-process client drops both channels while the connector ignores stdin EOF and never responds, this pump parks on connector_rx.next() and the Guard is never dropped, so the container outlives the session (master killed it on stream drop). Would a response_tx.closed() arm in the select, or moving the Guard back into the response stream, close that?

Concretely: catalog-tests Timeouts::start firing on a DerivationSession::start whose image connector is stuck before Open and not reading stdin drops both ends and leaves the container running; hanging_connector.py only hides this because it exits on stdin EOF.

go/runtime/flow_consumer.go
```diff
@@ -348,6 +379,33 @@ func (f *FlowConsumer) InitApplication(args runconsumer.InitArgs) error {
...
+			Plane:            config.Plane(),
+			Process:          f.localProcessSpec(),
+		},
+		func(log ops.Log) { ops.LogToLogrus(log, "connector-proxy") },
+	)
+	if err != nil {
+		return fmt.Errorf("creating V2 connector proxy task service: %w", err)
+	}
+	pcn.RegisterConnectorServer(args.Server.GRPCServer,
+		&connectorProxyV2{conn: f.proxySvcV2.Conn()})
```
dgreer-dev, line 398: The V1 path takes a --proxy-runtimes slot (default 2) before starting a proxied connector; V2 has no equivalent here or in crates/connector, so once the agent dials there can be a burst of discovers?  Do we need the cap or is dropping it intended?  I think this could be a decent number of containers that start?

go/protocols/connector/connector.proto
```diff
@@ -0,0 +1,96 @@
...
+/// Response is one message of a Connector RPC's response stream.
+//
+// The stream is zero or more `log`, then `started` carrying the connector's
+// Spec, then a mix of `log` and protocol responses of the request stream's
+// type until EOF.
+//
+// A failure before the connector is running ends the stream with a Status and
+// no `started`. After protocol responses end, stderr is drained before EOF or
+// a terminal Status, subject to a teardown timeout. A terminal Status is the
+// stream's last item.
```
dgreer-dev: I don't see the mentioned teardown timeout.  I'm not sure we need one, this might just be a doc thing?

go/runtime/flow_consumer.go
jgraettinger: Dropping it is intended. These need to run somewhere, and we can either balance them across the fleet (DNS round-robin works very well), or run them all in the single agent pod as we do today. But the semaphore is incompatible with catalog-tests workloads that need to start all derivations simultaneously before any test can run.

crates/connector/src/serve.rs
jgraettinger: It gets even messier, because what we're actually killing on drop isn't the image container. It's just the `podman run` that starts it and proxies its stdio, and SIGKILL-ing that leaves the container running merrily along. It does intercept SIGINT and bounces that to the daemon, but that's no help on drop.

An important defense is that `flow-connector-init`, injected into the container, is the init process on the other end of `response_tx`. On input drop it kills the subordinate connector itself to drain stdout, and it additionally will show itself out after 10s of no inbound connection from the reactor. Without this, past experience was we'd have hung containers all over the place.

On the specific question here, looking with fresh eyes I'm thinking:
* Make the response_tx.send(...) best effort, ignoring error
* Add an arm on response_tx.closed() with its former error ("Connector client dropped ...")

go/protocols/connector/connector.proto
jgraettinger: I'll remove the reference to a teardown timeout

jgraettinger:
PTAL

crates/connector/src/serve.rs
dgreer-dev: Yeah I see the change, that looks right to me.  Maybe the `closed()` arm would be better above c`onnector_rx.next()` if the connector keeps emitting responses?  With `biased` I think there'd be a delay actually getting down to the closed.

(Not sure it really matters, it'll get hit eventually)

dgreer-dev (approved):
LGTM.  It'll be nice when we can fully deprecate the runtime v1 stuff...

crates/connector/src/serve.rs
jgraettinger: I had considered the ordering here, and did it this way for performance reasons: if there's data ready, then avoid polling closure and _then_ polling data in a hot loop. A send into a closed channel is immediately ready (doesn't await), and it will briefly spin before draining the input buffer and hitting the close arm.

dgreer-dev: Good point, that's the much more common and important case

PR #3491, opened 2026-09-14
control-plane-api: definitive capability-mask denials bypass the retry path

bbartman:
Part of #3376 (stacked on #3488): a capability-mask shortfall becomes a **definitive** denial with a structured `403` body, bypassing the Snapshot refresh-and-retry path.

## What this does

- `AuthZResult<Ok>` now errs with `AuthZError { Retriable(tonic::Status), Definitive(Forbidden) }`. `authorization_outcome` returns a `Definitive` denial immediately as the new `ApiError::Forbidden`; the `Retriable` path is unchanged.
- `Forbidden { error, message, missing_capabilities }` renders as a `403` JSON body on REST and as `extensions.error` / `extensions.missing_capabilities` on GraphQL, so "you need capability X" is machine-readable identically on both surfaces.
- `Forbidden::required_covered` replaces `masked_shortfall` in both Envelope gates and inside `user_is_authorized_for`, so all three `/authorize/user/*` mints deny definitively. The message keeps this chain's wording (names the withheld bits in claim vocabulary and the target; never the user's grants).
- `createRefreshToken` refuses masked bearers with `unmasked_token_required` before the service-account lookup; revocation stays open to masked bearers.

## Decisions

- **Typed error over smuggling.** The definitive/retriable distinction is an enum variant, not a Status metadata marker, so the compiler enforces that every `ApiError` consumer handles `Forbidden` and `authorization_outcome` cannot forget the bypass. Policy functions pay `.into()` at their Status returns (35 sites); reviewed and kept over a details-payload alternative.
- **`Forbidden` lives in `server/error.rs`** beside `ApiError` / `AuthZRetry`; there is no `Authority` extractor on this chain.
- **No `/admin/*` guards.** `create-data-plane` and `update-l2-reporting` already authorize through the masked walk on `ops/` Admin, so a masked bearer is denied by the same gate, now as the structured 403 (pinned by the create-data-plane test). The old RequireUnmasked rationale (SQL `user_roles` couldn't bind the mask) no longer applies.
- **`missing_capabilities` is `Vec<String>`**: `Capability` has no `name()` here; `Display` gives the same PascalCase spelling.
- `hide_denial_as_not_found` (service accounts) still rewrites only walk denials. A `Forbidden` passes through unchanged: it discloses nothing about grants.

## Tests

202 tests, all green. New: the 403 body and predicates (`error.rs`); definitive-vs-retriable routing in both Snapshot geometries and the inverted `test_mask_shortfall_after_snapshot_is_definitive` (`envelope.rs`); `createRefreshToken` masked refusal + open revocation (`refresh_tokens.rs`); masked bearer on `/admin/create-data-plane` gets `missing_capabilities`. The tenant GraphQL snapshot was accepted with the new shape (extensions present, `PermissionDenied:` prefix gone).

Next: `bmb/3376-capability-token-mint` stacks on this branch and ports the `capability_token` grant from #3424.


bbartman:

PR #3492, opened 2026-09-14
control-plane-api: capability_token grant mints masked access tokens

bbartman:
Part of #3376 (stacked on #3491): `POST /api/v1/auth/token` gains the `capability_token` grant, the first production path that mints masked tokens. This ports #3424 onto the restart chain.

```json
{"grant_type": "capability_token", "capability_mask": ["CatalogRead", "Viewer"]}
```

An unmasked, human bearer mints an access token that copies the caller's verified identity claims and carries `capability_mask` as a claim. Enforcement (the masked grant walk, landed earlier in this chain) intersects the mask with the user's live grants at use time. There is no `upgrade_token` parameter, per Gregor's revision on the issue: a client wanting more capabilities requests the full list in a fresh mint.

## Decisions implemented

- **Copy-through claim set.** `sub`, `role`, `aud`, and `email` (when present) are copied verbatim; fresh `iat`; `exp = iat + validity`. Aside from `email` and `capability_mask` this is exactly the SQL `generate_access_token` claim set, so the two mint paths cannot drift and a masked token remains a functional Supabase/PostgREST token.
- **The mask is stamped verbatim**: bundle names, unknown names, duplicates, empty list (identity-only token). Enforcement decides what a name enables; unrecognized names can never widen.
- **Rust-side signing** with the control plane's HMAC key (`authorize_dekaf` precedent). The SQL function is untouched.
- **Configurable validity**: agent `--capability-token-validity` / `CAPABILITY_TOKEN_VALIDITY` (humantime) feeds `App::capability_token_validity` into the pure signer. Default `DEFAULT_CAPABILITY_TOKEN_VALIDITY` = 1h, welded to the clap default string by a test in `agent`. Longer windows widen leaked-token exposure; the flag docs say so.
- **Masked callers are refused** with the structured `unmasked_token_required` 403 (`Forbidden::require_unmasked`, from #3491): a reduced token cannot widen or re-mint itself. This answers the issue's open question.
- **Service-account callers are refused** with `service_account_forbidden` (decided 2026-08-24 as v1 behavior). The lookup moves to `grants::is_service_account`, shared with the refresh-token mutations, and is paid only at the mint. The mask refusal runs first, without a database round-trip.
- **The handler extracts `Envelope`** (this chain has no `Authority` extractor), so a presented bearer is verified regardless of grant: a broken bearer now 401s even for `refresh_token`, which previously ignored the Authorization header. Bearer-less `refresh_token` requests are byte-identical. No known client sends a bearer here.

## Tests

`test_capability_token_mint` drives the real router end-to-end: copy-through claim set and one-hour expiry (presenting the minted token back as a bearer proves the signature through production Envelope verification); verbatim mask stamping; the identity-only empty mask, still refused as a mint caller ("masked" is claim presence, not value); minted names driving real enforcement at `/api/v1/catalog/status` (Viewer mask reaches the walk, identity mask gets the definitive `missing_capabilities` 403 from #3491); scoped-role, email-less copy-through; missing-bearer 401; masked-caller and service-account-caller 403 bodies (insta-pinned); mask refusal preceding the service-account lookup; invalid-bearer 401 on the `refresh_token` grant; 422 for structurally invalid requests. `test_validity_is_configurable` pins a non-default window at the signing seam.

204 control-plane-api tests green; the agent weld test passes.

## End-to-end suite (second commit)

`server/masked_token_suite/` holds the no-amplification and compatibility suite: every test mints through the real `capability_token` grant and drives the minted token through real routes over a DB-backed snapshot, so it exercises the composed mint-plus-enforcement pipeline that the mint test and the per-surface enforcement tests each cover only half of. The suite reads a dedicated `masked_suite` fixture (dana: direct admin grant, a `Delegate`-gated role edge, an `Assume`-gated role edge; erin: a single `read` grant; `otherCo/` unreachable), with per-edge rationale at the top of the fixture.

- **`walk.rs`** — `test_no_amplification_sweep` runs an 11-mask family (unmasked, empty, unknown-only, single bundles, `Viewer` ± `Delegate`/`Assume`, `Admin`, every bundle) across the four path classes at `/authorize/user/collection`, asserts masked-allowed ⇒ unmasked-allowed for every cell, asserts the all-bundles mask is byte-identical to no mask, and pins the full grid as a snapshot. `test_unheld_bits_are_inert` shows an `Admin` mask still gets exactly the `read` the live grant holds at the role-edge hop. `test_legacy_metadata_cannot_authorize` shows a ref the mask leaves unreachable presents with `userCapability` and gated fields null through GraphQL, while a reached ref keeps its literal legacy label.
- **`lifecycle.rs`** — `test_grants_change_under_a_live_token` mints once for erin, then upgrades and later deletes her grant, pinning that each change is invisible until the next snapshot refresh and takes effect at it with no re-mint (via the new `test_server::RefreshableSnapshot`).
- **`guards_and_compat.rs`** — `test_credential_loop` runs the unmasked refresh-token loop end to end (create via GraphQL, exchange at `/api/v1/auth/token`, full authority from the SQL-minted token in both access-token and bearer-credential form), then revokes the credential with an identity-only minted token and confirms the exchange is dead. The SQL exchange runs under a new `jwt_sign_polyfill` fixture supplying the HS256 `sign()` and secret a `sqlx::test` database lacks.

Refusals already pinned elsewhere (a minted token cannot re-mint, service accounts are refused at the mint, a masked bearer cannot `createRefreshToken`, `/admin` routes deny through the masked walk on `ops/` Admin) are deliberately not repeated. 209 control-plane-api tests green from a clean build.


bbartman:

PR #3494, opened 2026-09-15
Allow negative read priority on derivation transforms and materialization bindings

### LINKED ISSUE #2682: Allow negative read priority on derivation/materialization bindings
## Problem

When binding priorities are all the same (the default of `0`), the platform levels reads across bindings and journals so that it proceeds in ascending wall-clock (publication time) order. This means that when a new binding with a large historical backlog is introduced — for example, adding a new L1 rollup into an L2 derivation — all of that historical data is ordered before any current data from other bindings, because its publication time is earlier.

This caused a real issue with the catalog-stats L2 derivation: introducing a new test data plane's L1 binding caused the L2 derivation to stall on current data while it worked through the historical backlog, because the leveled reads prioritized the older publication times.

[Read priority](https://docs.estuary.dev/concepts/derivations/#read-priority) exists for exactly this use case — marking a binding as lower priority so the derivation prefers other bindings first and works through the backlog second. However, using it today requires raising the priority of **all other** bindings, because priority is a non-negative integer and the default is `0`. You can't express "this one binding is lower priority than normal."

## Proposal

Allow negative values for the `priority` field on derivation transforms and materialization bindings. This way, a user (or operator) can set `priority: -1` on a single binding to express that it is lower-priority than bindings using the default (`0`), without needing to touch every other binding.

This should be backwards-compatible with all current usages, since existing values `> 0` continue to be higher priority and the default of `0` is unchanged.

## Expected behavior

- `priority: -1` (or any negative value) is accepted in derivation transform and materialization binding configs
- Negative-priority bindings are read after all default (`0`) and positive-priority bindings
- Existing behavior for `priority >= 0` is unchanged

jgraettinger:
**Description:**

Widens `priority` from `uint32` to `i32` on derivation transforms and materialization bindings, so a single binding can be demoted below the default priority of zero.

Reads of equal priority are leveled by publication time, so a binding introduced with a large historical backlog orders ahead of every other binding's current data. Adding a new data plane's L1 binding to the catalog-stats L2 derivation stalled it on current data for exactly this reason. Read priority is the lever for this, but with a `uint32` and a default of zero, demoting one binding meant promoting every other one.

Nothing about the ordering changes. Both read heaps (Rust `crates/shuffle`, Go `go/shuffle`) already order by `(priority DESC, adjusted_clock ASC)`, so negative values sort after the default without a comparator change.

Closes #2682.

**Workflow steps:**

Set `priority: -1` (or any negative value) on a transform or binding to have it read only when bindings of default or greater priority have no ready documents:

```yaml
materializations:
  acmeCo/example/db-views:
    bindings:
      - source: acmeCo/current-data
        resource: { table: current }
      - source: acmeCo/huge-backlog
        resource: { table: backlog }
        priority: -1
```

Previously this required raising every *other* binding to `priority: 1`. Existing specs are unaffected: `priority >= 0` behaves exactly as before.

- `site/docs/concepts/derivations.md` — [Read priority](https://docs.estuary.dev/concepts/derivations/#read-priority) and the transform spec comment.
- `site/docs/concepts/materialization/materialization.md` — [Processing order](https://docs.estuary.dev/concepts/materialization/#processing-order) and the binding spec comment.

Both previously documented `priority` as `integer >= 0`; both now describe negative priority and the backlog case that motivates it.

**Notes for reviewers:**

The one judgment call worth checking is `int32` vs `sint32` in the `.proto`. `int32` leaves the encoding of non-negative values byte-identical to the prior `uint32` encoding, so a reactor on either side of this change reads the other's built specs. `sint32` would zig-zag every value and break that. Negative values cost a ten-byte sign-extended varint.

The existing proto regression snapshots are unchanged, which is the compatibility evidence. New tests in both codecs (`go/protocols/flow/priority_test.go`, `test_negative_priority_round_trips` in `crates/proto-flow/tests/regression.rs`) assert the exact bytes, so a later switch to `sint32` would fail loudly.

Other things to look at:

- The JSON schema loses `"minimum": 0` and reports `format: int32`.
- `crates/validation/tests/model.yaml` gains a `priority: -1` transform, so the golden snapshot shows the value surviving into the built spec.
- Heap tests in all three implementations give the de-prioritized entry the earliest clock of all and assert it still pops last.


jgraettinger:
Ran this through a QA pass on a local stack. The only finding is that V1 `:dev` derivations cannot support it.

jgraettinger:
To state for the record, we'll need to release this widely before negative priorities are safe to start using in the dashboard, etc.

dgreer-dev (approved):
LGTM

PR #3495, opened 2026-09-15
control-plane-api, tables: legacy decisions migrate to effective capability bits

### LINKED ISSUE #3232: storageMappings query fails for bundle-only grants (legacy-capability mismatch)

bbartman:
Part of #3376 (stacked on #3492): request-path access decisions in the GraphQL API key on mask-attenuated **effective capability bits**, never on the legacy `capability` column. The legacy label becomes reporting metadata only. This ports the old #3422 onto the restart chain and **closes #3232**.

## Why

Two functions disagreed about bundles-only grants. `authorized_prefixes` admits a prefix on effective bits, so a grant whose authority comes entirely from the `bundles` column qualifies; the per-row `get_user_capability` then reported that grant's legacy column, `None`, and each consumer treated the absent label as an error. `storageMappings` failed the whole query ("missing capability for catalog prefix"), `dataPlanes` panicked on an `.expect`, and `LiveSpecRef` presented every such spec as inaccessible. A user holding a bundles-only `viewer` grant could not list what the pre-filter had already authorized.

## Commits

1. **`tables`: `get_user_authorization` pairs effective bits with the legacy label.** A single walk reduction returns `UserAuthorization { bits, legacy }`: bits accumulate additively across covering nodes and are mask-attenuated like every walk emission; `legacy` is the max un-attenuated label among covering nodes that carry one, `None` under bundles-only coverage. `legacy_label()` owns the reporting rule that `none` stands in for bundles-only coverage. `get_user_capability` is deleted; `Snapshot::user_capability` (the unmasked executor-path wrapper from #3387) keeps its contract as a wrapper over the legacy half.
2. **`control-plane-api`: GraphQL access decisions key on effective bits.** `attach_user_capabilities` hands closures the full `UserAuthorization`. `LiveSpecRef.userCapability` is null (and every gated field null) when the bits fall short of the Viewer bundle, otherwise the literal legacy label. The writtenBy/readBy filter is a `CapabilitySet` superset check. `storageMappings` and `dataPlanes` report `none` for bundles-only rows instead of erroring or panicking. The `userCapability` docstring states the semantics and the SDL is regenerated.
3. **`control-plane-api`: thresholds speak the bundle vocabulary.** `alertConfigs`, `inviteLinks` and the `dataPlanes` pre-filter spell their minimums as `CapabilityBundle::Viewer` / `Admin`, bit-identical to the enums they replace. The storage-mapping mutation responses' `Admin` literals stay as informational metadata, now commented.

## Decisions

- **Threshold unchanged, input migrates.** Every gate keeps today's semantic threshold (legacy Read ⇒ Viewer bundle bits; legacy Admin ⇒ Admin bundle bits) but evaluates effective bits, never the label.
- **Gate fails ⇒ ref presents inaccessible**: `userCapability: null` and all gated fields null, exactly as an unmasked user sees a ref they can't access. No grant disclosure to masked bearers.
- **Literal legacy label, `none` for bundles-only.** A derived floor (mapping bits back to the nearest legacy enum) was rejected: the team is demoting legacy, `PrefixRef` already reports this way, and a floor would invent a value no grant row holds. A bundles-only viewer therefore sees `userCapability: none` beside a fully populated spec. The 8/26 Daily Sync "Discuss Legacy Bits" item covers this question; the written record (#3232, the PrefixRef doc from #2936) backs the choice.
- **No new refusal at the `liveSpecs` root.** It already refuses a sub-Viewer mask for explicit names and prefixes through the shared policy's typed definitive denial (#3491); a test pins that, so a reached grant's legacy label is never a side channel around the mask.
- **Executor paths untouched.** Publications and discovers still consult the legacy label unmasked, so a bundles-only grant still cannot publish. That is the same deferred class as discovers mask propagation and gets its own ticket.

## Tests

Each defect has a test written and run red against the prior code, landing with its fix in commit 2: `storage_mappings_list_under_a_bundles_only_grant` (#3232's error), `test_graphql_data_planes_bundles_only_grant` (the `.expect` panic), `test_graphql_live_specs_bundles_only_grant` (null refs), and `test_graphql_written_by_read_by_filter_on_effective_bits` (withheld names, plus the un-filtered writesTo path showing an accessible bundles-only referent beside a null ref). `test_graphql_live_specs_mask_below_viewer_is_refused` pins the root refusal for both inputs. `tables` gains `test_get_user_authorization` and `test_get_user_authorization_multi_path_composition`; the two legacy-label tests are renamed onto the combined reduction. `mask_gates_reachability_but_keeps_literal_legacy` now pins both halves.

261 control-plane-api and tables tests green from a clean build; `cargo check -p agent --tests` clean.


bbartman:

PR #3499, opened 2026-09-15
network: pin auth-redirect orig_url to the request host

skord:
The `/auth-redirect` handler is dispatched ahead of `verifyAuthorization` in `serveConnHTTP`, so it is reachable with caller-supplied parameters. Its `orig_url` was validated only for non-emptiness before being passed to `http.Redirect`, so a task domain could be used to redirect to an unrelated host, and the auth cookie set just above was written before that redirect.

This requires `orig_url` to parse and to name the same host as the request, and moves the check ahead of `SetCookie` so a rejected request leaves no cookie. The normal flow is unchanged: `startAuthRedirect` already builds the value as `"https://" + req.Host + req.URL.Path`.

Adds `go/network/auth_test.go` covering the accepted same-host case, foreign and scheme-relative hosts, a host-relative path, an unparseable URL, and the two missing-parameter paths.

Resolves the open `go/unvalidated-url-redirection` code scanning alert.


jgraettinger (approved):
LGTM

PR #3500, opened 2026-09-15
workflows: declare minimal GITHUB_TOKEN permissions

skord:
The org and repo default workflow permission is `write`, so both of these workflows ran with a write-all `GITHUB_TOKEN`. Adding a `permissions:` block drops everything not named.

`control-ci-cd` authenticates to Supabase with `SUPABASE_ACCESS_TOKEN` and uses `GITHUB_TOKEN` only for the checkout, so `contents: read`. `flowctl-release` uploads the built binaries to the published release via `actions/upload-release-asset`, so `contents: write`.

Resolves both open `actions/missing-workflow-permissions` code scanning alerts.


williamhbaker (approved):
LGTM

PR #3501, opened 2026-09-16
Control-plane: Refactored is_authorized into a member function of the envelope

bbartman:
**Description:**

This is a refactor only PR only one change in it, that change is wrapping the UserRole::is_authorized function that is reachable from the graphQL endpoints into a single location. This is being done so that we can more add support the
capability_mask later on.

**Workflow steps:**

This is a refactor only change no workflows are changed as a result of this change.

N/A.

**Notes for reviewers:**

I have a single callout here and that is, should we rename is_authorized? should it be named something else, like `is_user_authorized`, then it would match the snapshots function of the same name.

Another question, should we also refactor `tables::UserGrant::attach_user_capabilities`. I ask because I was digging into the part where do the `reachable_nodes` through `get_user_capability` refactoring and saw that this was in the path, and is doing kinda a similar thing to `user_is_authorized` so I was wondering if refactoring it now would be prudent.


crates/control-plane-api/src/envelope.rs
```diff
@@ -83,6 +83,24 @@ impl Envelope {
         self.refresh.result().expect("Snapshot refresh never fails")
     }
 
+    /// Determine if a user is authorized with a capability on a specific
+    /// prefix.
+    pub fn is_authorized(
+        &self,
+        user_id: uuid::Uuid,
```
GregorShear, line 90: No need to pass in `user_id` - the envelope already has the claims and can read the subject.

and in that light your suggestion to rename to `user_is_authorized` makes sense

bbartman: OK. The problem that I have with accessing the claims, is that it changes the return type because the `claims()` function can return an error, and if you want `is_authorized` to be able to return an error then that's fine, however, I'm concerned about some of the context that it's used within, in that they may result in errors surfacing out of the order they previously did. I'm checking into this further, because I tried to do this before and it cause some issues in another PR that wasn't reviewed, and instead I decided against it.  For this reason specifically:

```The real risk is not order but routing. In two sites an Unauthenticated status would start flowing through authorization_outcome, which would turn an immediate 401 into a 307 retry.``` This is from claude.

A better approach would be to validate the claims when constructing the Envelope so it doesn't return an error message, unless there some reason not to do it? that way we wouldn't have to error when accessing the claims. Again, unless there's a reason not to do that. Hope this makes sense.

I'll push the rename changes in a sec.

**Comment by jshearer** [comment]
My original request was to route GraphQL authorization through `Envelope::is_authorized()`. The first PR (this one) would make that change without changing behavior, then a follow-up PR would read the token's capability mask and pass it to `UserGrant::is_authorized()`. This would give us one place where a token's mask enters an authorization check.

Like we talked about on the call yesterday, that covers only one of three consumers of `UserGrant::reachable_nodes` though, and my suggestion was to start there and follow the threads until you solve all of its direct and indirect call-sites.

As I was reviewing this work, I'm realizing that many callers are inside pure policy functions, rather than directly inside resolvers: `evaluate_names_authorization`, `check_authorization`, `attach_user_capabilities`, `authorized_prefixes`, etc. They receive a Snapshot or grant tables, plus claims or a user ID. Passing an `Envelope`, or even a `ControlClaims` into those functions adds undesirable coupling to request state which then also must be mocked at test time.

**I propose** making the subject of a particular authorization check an explicit value: `models::authz::Subject`. Today it's just the user's ID, but my core observation is that the intent of this work is to extend it to also contain more fields: now the capability mask, soon one or more catalog prefixes, and who knows what else down the line.

```rust
// models/src/authz.rs

pub struct Subject {
    pub user_id: uuid::Uuid,
    // PR 2: pub mask: Option<CapabilitySet>,
}

impl Subject {
    /// Uses the user's grants without additional caller restrictions.
    /// This still requires grants; it does not bypass authorization.
    /// For callers acting on a `user_id` stored on a row.
    pub fn unrestricted(user_id: uuid::Uuid) -> Self
}

impl ControlClaims {
    pub fn subject(&self) -> authz::Subject   // PR 2: reads self.capability_mask
}
```

For this PR, I propose the following changes:
1. Add `pub struct models::authz::Subject` with only `pub user_id: uuid::Uuid`.
2. Replace subject UUID arguments with `Subject` throughout authorization and graph-traversal call chains.
3. Add `ControlClaims::subject() -> Subject` as the single adapter from credential claims to authorization context.
4. Use `claims.subject()` in request paths where authorization calls currently receive `claims.sub`.
5. Use `authz::Subject::unrestricted(user_id)` in background paths that intentionally exercise the user's full authority.

`evaluate_names_authorization`, `check_authorization`, and `attach_user_capabilities` keep their original snapshot and claims arguments. Their calls into `UserGrant` call and pass the new `claims.subject()`. Helpers that receive a subject UUID instead receive `Subject`.

Both `ControlClaims` and `Subject` belong to `models`, so the adapter stays within that crate and we don't have to touch `Envelope` at all. The `tables` crate receives a subject without depending on credentials or request state.

PR 2 then adds the optional mask to `Subject`. The adapter copies the mask from the claims, and the grant traversal layer enforces it. *Existing authorization callers continue to pass the subject without further changes.*

PR #3502, opened 2026-09-16
Introduce `models::authz::Subject` to encapsulate authorization context beyond just user identity

### LINKED ISSUE #3376: Capability-masked tokens

jshearer:
This PR introduces `models::authz::Subject` and replaces user UUID arguments throughout the authorization call graph, with no authorization behavior change.

The central observation is that the subject of an authorization check includes both the user’s identity and the restrictions on the authority the caller can exercise. Today, `Subject` contains only `user_id`. Follow-up work can add a capability mask, and later catalog-prefix constraints, to the same value.

This gives us a clear separation:

- `ControlClaims::subject()` is the single adapter from credentials to authorization context.
- Background operations use `Subject::unrestricted(user_id)` to exercise the user’s full granted authority.
- Grant traversal receives a subject without depending on credentials or request state. Existing helpers retain their snapshot and claims arguments; `Envelope` needs no changes.

`Subject` passes through all three consumers of `UserGrant::reachable_nodes`: `is_authorized`, `get_user_capability`, and `reachable_prefixes`. The next PR can then add the capability mask to `Subject`, populate it in the claims adapter, and enforce it during traversal without changing authorization callers. A third PR will add the token endpoint support.


bbartman (commented):
This looks good.

bbartman (dismissed):
Forgot to approve.

bbartman (approved):
Re approving for simple change.

GregorShear (approved):

PR #3503, opened 2026-09-16
network: drop the bearer token from authorization errors

skord:
## Problem

When a caller presents a valid but non-matching task authorization, `verifyAuthorization` embedded the caller's entire Bearer JWT in its error message:

```go
return fmt.Errorf("invalid authorization for task prefix %s (%s)", shardIDPrefix, bearer)
```

That error reaches the client two ways in `go/network/frontend.go`: as the 403 response body, and, for browser requests, as the `err` query parameter of the `307` redirect to the dashboard's `/data-plane-auth-req`. The second is the one that matters, since a short-lived data-plane token then persists in browser history, dashboard access logs, and any telemetry that records URLs.

The token is the caller's own, so this is an exposure hygiene weakness rather than a cross-user compromise path.

## Fix

Drop `bearer` from the message. The task prefix carries the diagnostic value on its own.

This is the only site in the package that interpolated a credential into an error, so removing it at the source closes both sinks. The remaining paths are already token-free: the `missing authorization` case, and the gazette verifier errors from `auth.verifyWithKeys`, which report only how verification failed.

I left the `err` parameter carrying raw error text rather than switching to a stable error code. With the token gone, the remaining text is what the dashboard already displays, and changing it would touch that contract.

## Testing

`go build`, `go vet`, `gofmt -l`, and `go test ./go/network/` pass.


williamhbaker (approved):
LGTM

PR #3509, opened 2026-09-17
CapabilityMask Support

bbartman:
**Description:**

Here are the features that have been added:

1) Created the `CapabilityMask` type.
2) Update the `CapabilityBundle` type to support parsing and printing of names. This is to support parsing of the capabilities provided as part of the claim's `capability_mask` field.
3) Add the `CapabilityMask` to the `Subject` type.
3) Update the related function to support the new capability mask.

**Workflow steps:**

N/A

N/A

**Notes for reviewers:**

The `capability_mask` parsing works as follows, for signed tokens, we ignore unknown capabilities. This is to prevent issue when receiving a newly minted token on an old services during a slow or Blue-Green role out. Basically, instead of returning an error if we encounter a CapabilityBundle name that we are unaware of then we simply ignore it. We should only enforce Erroring on unknown capabilities during token minting.


bbartman:
@strix-security Ready for rereview.

bbartman:
@strix-security Ready for rereview

bbartman:
@strix-security Ready for rereview.

crates/models/src/authz.rs
```diff
@@ -68,7 +73,36 @@ impl std::fmt::Display for Capability {
...
+    Eq,
+    PartialOrd,
+    Ord,
+    Hash,
+    Serialize,
+    Deserialize,
+    strum::EnumString,
+    strum::IntoStaticStr,
+)]
+#[strum(serialize_all = "snake_case")]
```
jshearer: Is there any reason we need Strum's parsing and naming machinery vs `CapabilityBundle`'s existing Serde implementation? Bundle names already have an established serialized representation; capability masks should consume that same representation unless there's a good reason otherwise.

crates/models/src/authz.rs
```diff
@@ -68,7 +73,36 @@ impl std::fmt::Display for Capability {
...
+/// 1) These bundles are stored within the grant part of the database, and are
+/// used along side of the capability column to enable graph traversals. Beyond
+/// storage these things are expanded into capability bits in order to support
+/// a more fine grain access control between nodes of the role graph.
+/// 2) To provide the users with an interface for selecting capabilities that
+/// can be used as part of the capability mask feature.
+///
+/// This means that there are two representations of this, while there isn't
+/// really another way to do this currently, we are leveraging strum to provide
+/// access to parsing of the enum in pascal case and serde to provide parsing of
```
jshearer: These docs still describe external names as PascalCase, but the final commit makes them snake_case. The `name()` doc below also still promises PascalCase. A caller following this guidance would supply `"Viewer"`, which the parser silently ignores.

Please replace the two-representation explanation with the current contract: bundles have stable snake_case names and expand into internal capability bits. The `ControlClaims::capability_mask` field should document the wire rules where callers encounter them: omitted/null means no mask, an empty array means an empty mask, recognized bundles are unioned, and unknown names are ignored during claim interpretation. The paragraph asserting there is no other way to represent this can go.

crates/models/src/authz.rs
```diff
@@ -171,6 +205,11 @@ impl CapabilityBundle {
             Self::Assume => Assume.into(),
         }
     }
+
+    /// This returns the pascal cased name of the enum variant.
+    pub fn name(self) -> &'static str {
```
jshearer: `name()` and `is_empty()`, appear to have no callers, and the new `From<CapabilityBundle>` conversion, which duplicates `capabilities()` is used only by its test.

crates/tables/src/behaviors.rs
```diff
@@ -1717,4 +1741,368 @@ mod test {
             CapabilityBundle::Viewer.capabilities(),
         );
     }
+
+    fn assert_reachable_masked(
+        role_grants: &RoleGrants,
+        user_grants: &UserGrants,
+        subject: &authz::Subject,
```
jshearer: This duplicates `assert_reachable` above: collect nodes, convert names, sort, deduplicate, convert expectations, and compare. The meaningful difference is accepting a `Subject` instead of a UUID. Please update the existing helper to take `&Subject` and use it for both masked and unrestricted cases.

The shared `masked_walk_scenario` is useful. The repeated cases built on it could use a small local table, while keeping the distinct parent-prefix, `Assume`, and legacy-reporting checks. We can reduce the test machinery without losing those separate behavioral boundaries.

crates/models/src/authorizations.rs
```diff
@@ -20,12 +22,22 @@ pub struct ControlClaims {
...
+    pub capability_mask: Option<Vec<String>>,
 }
 
 impl ControlClaims {
     /// Converts these claims into the authorization subject used by grant evaluation.
     pub fn subject(&self) -> crate::authz::Subject {
-        crate::authz::Subject { user_id: self.sub }
+        crate::authz::Subject {
+            user_id: self.sub,
+            capability_mask: authz::CapabilityMask::from_claims(self.capability_mask.as_ref()),
```
jshearer: Let's keep claim interpretation in `ControlClaims::subject()`. The claim is a list of bundle names, which is the token's wire format, so the conversion belongs with `ControlClaims` rather than in `authz`.

crates/tables/src/behaviors.rs
```diff
@@ -1717,4 +1741,368 @@ mod test {
...
+        let subject_no_mask = authz::Subject {
+            user_id,
+            capability_mask: None,
+        };
+        // Unmasked: every reached prefix surfaces. supportCo/ emits its bare
+        // Assume bit, so it is non-empty and never a candidate for the skip;
+        // that Assume bit is what carries the walk on to daveCo/, which
+        // arrives with full Admin bits.
+        let unmasked = UserGrant::reachable_prefixes(&role_grants, &user_grants, &subject_no_mask);
+        assert_eq!(
```
jshearer: The unmasked half of this test doesn't exercise preservation of an empty-capability node: every prefix in this scenario has nonempty capabilities when unmasked. The test would therefore still pass if `reachable_prefixes` accidentally dropped the `subject.capability_mask.is_some()` condition and always filtered empty nodes.

Please add one otherwise unrelated direct grant with legacy capability `None` and no bundles to this scenario, and assert that its prefix survives unmasked but is omitted under the mask. That directly protects the compatibility behavior of the condition.

crates/tables/src/behaviors.rs
```diff
@@ -296,7 +316,12 @@ fn next_neighbors<'a>(
         .flatten()
         .map(move |g| g.to_node_ref(delegatable));
 
-    p1.chain(p2).chain(p3)
+    p1.chain(p2).chain(p3).map(move |mut node| {
```
jshearer: Small simplification: apply the optional mask to the existing `delegatable` set once, after computing it, and let the existing `to_node_ref(delegatable)` calls perform their intersection. Each path already computes `edge_bits & delegatable`, and `(edge_bits & delegatable) & mask` is equivalent to `edge_bits & (delegatable & mask)`.

This removes the extra node-mutation stage and keeps capability intersection in the existing construction path. Leave the terminal-node decision based on the current node's capabilities as it is.

crates/models/src/authz.rs
```diff
@@ -187,3 +226,166 @@ impl From<super::Capability> for CapabilitySet {
...
+        // A mask without `Delegate` confines a token to direct grants even
+        // when the underlying grant would delegate.
+        assert!(editor.contains(Capability::Delegate));
+        assert!(
+            !CapabilityMask::new(viewer)
+                .apply(editor)
+                .contains(Capability::Delegate)
+        );
+
+        insta::assert_debug_snapshot!(names(CapabilityMask::new(editor).apply(viewer)), @r###"
```
jshearer: If we remove the newtype and Strum parser, their dedicated `mask_apply_intersects` and `from_name_is_strict` tests should go too. Preserve the relevant claim-input cases in the `ControlClaims::subject()` tests and the intersection behavior in the traversal tests.

The new `legacy_capability_parity` test can also go: it restates the unchanged legacy mapping and the added `From` conversion. Existing legacy-grant traversal tests exercise that mapping through actual behavior.

crates/tables/src/behaviors.rs
```diff
@@ -1717,4 +1741,368 @@ mod test {
...
+            ));
+        }
+        assert!(UserGrant::reachable_prefixes(&role_grants, &user_grants, &subject).is_empty());
+    }
+
+    #[test]
+    fn test_masked_walk_without_delegate_is_direct_only() {
+        use Capability::*;
+
+        // The mask strips Delegate from the direct Editor grant, so the
```
jshearer: Could we trim the comments that walk through the expected nodes? Here the test name and expected list already explain the direct-only result; the five-line walkthrough makes readers work through the same information twice. Keep short explanations of non-obvious contracts, such as why legacy values remain available.

The empty-mask test's comment also needs a correction: only the direct-grant nodes are emitted with empty capabilities; downstream nodes aren't reached.

crates/tables/src/behaviors.rs
```diff
@@ -220,11 +235,16 @@ impl super::UserGrant {
 // up to 2^N per prefix where N is the number of capability bits. If deep
 // grant graphs cause latency, replace bfs_reach with a manual BFS that keys
 // visited state on object_role alone and prunes dominated capability subsets.
+//
+// The capability mask what is reachable by attenuating the capabilities of a
```
jshearer: Could this paragraph explain why the mask is applied inside the walk? For example:

> Mask each emitted node before it is explored, so subsequent traversal uses its restricted capabilities.

That explains the consequence of the operation's placement more clearly than restating intersection.

crates/models/src/authz.rs
```diff
@@ -68,7 +73,36 @@ impl std::fmt::Display for Capability {
...
+/// 2) To provide the users with an interface for selecting capabilities that
+/// can be used as part of the capability mask feature.
+///
+/// This means that there are two representations of this, while there isn't
+/// really another way to do this currently, we are leveraging strum to provide
+/// access to parsing of the enum in pascal case and serde to provide parsing of
+/// this in snake case. The snake case lives within the database, while the
+/// external representation is accepted by the users. The only overlap these
+/// features have is during part of `Display`/`Debug` when the strum
+/// representation is printed.
```
jshearer: Keep the bundle documentation focused on its domain role, and keep the decision that grants and tokens name bundles rather than bits. The claim-specific rules belong on `ControlClaims::capability_mask`.

```suggestion
/// A stable, named bundle of capability bits. Grants and token capability
/// masks name bundles rather than bits, so bits may be added, split, or
/// merged without changing stored grants or minted tokens.
```

crates/tables/src/behaviors.rs
```diff
@@ -1717,4 +1741,368 @@ mod test {
...
+            .collect();
+        nodes.sort_by(|a, b| a.0.cmp(&b.0).then(a.1.as_u32().cmp(&b.1.as_u32())));
+        nodes.dedup();
+
+        let expected: Vec<(String, EnumSet<Capability>)> = expected
+            .into_iter()
+            .map(|(prefix, caps)| (prefix.to_string(), caps))
+            .collect();
+
+        assert_eq!(nodes, expected);
```
jshearer: Could we reuse `assert_reachable` by changing it to accept `&Subject`? This helper duplicates its collection, sorting, deduplication, and comparison logic.

crates/tables/src/behaviors.rs
```diff
@@ -1717,4 +1741,368 @@ mod test {
...
+        let masked = UserGrant::reachable_prefixes(&role_grants, &user_grants, &subject);
+        assert_eq!(
+            masked.keys().collect::<Vec<_>>(),
+            vec![&"acmeCo/", &"bobCo/shared/", &"carolCo/upstream/",],
+        );
+
+        let subject_no_mask = authz::Subject {
+            user_id,
+            capability_mask: None,
+        };
```
jshearer: Use the existing constructor to state the unmasked case directly:

```suggestion
        let subject_no_mask = Subject::unrestricted(user_id);
```

crates/tables/src/behaviors.rs
```diff
@@ -1717,4 +1741,368 @@ mod test {
...
+
+    #[test]
+    fn test_masked_walk_without_delegate_is_direct_only() {
+        use Capability::*;
+
+        // The mask strips Delegate from the direct Editor grant, so the
+        // walk terminates there: bobCo/ and carolCo/ are unreachable even
+        // though the underlying grants reach them. The supportCo/ grant's
+        // Assume bit is likewise stripped, so that node surfaces empty and
+        // daveCo/ is unreachable too.
```
jshearer: This walkthrough restates the expected list. Keep the one mechanism this test protects, that masking away Delegate and Assume makes the direct grants terminal:

```suggestion
        // Masking away Delegate and Assume makes the direct grants terminal, so
        // nothing beyond them is reached even though the underlying grants delegate.
```

crates/tables/src/behaviors.rs
```diff
@@ -220,11 +235,16 @@ impl super::UserGrant {
 // up to 2^N per prefix where N is the number of capability bits. If deep
 // grant graphs cause latency, replace bfs_reach with a manual BFS that keys
 // visited state on object_role alone and prunes dominated capability subsets.
+//
+// The capability mask what is reachable by attenuating the capabilities of a
+// node against those of the mask. So the mask removes capabilities from nodes
+// that are not inside of the mask.
```
jshearer: Explain why the mask is applied inside the walk rather than to its output:

```suggestion
// The mask bounds what any edge may convey, including what Assume restores,
// so a node's restricted capabilities also decide its own later expansion.
```

crates/models/src/authorizations.rs
```diff
@@ -20,12 +22,22 @@ pub struct ControlClaims {
     // Authorized user email, if known.
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub email: Option<String>,
+    // This is the set of capabilities that the claims are limited to. The
+    // capabilities within this mask do not add privilege, but instead just
+    // enable a users privilege if they are set, and disabled if not. If no mask
+    // is present, the token is not a "capability masked token", and is treated
+    // as usual.
```
jshearer: Document the claim's input rules at the field, using the existing bundle vocabulary:

```suggestion
    // Capability bundles limiting which granted capabilities the token may use.
    // Omitted or null means no mask; an empty list means identity only.
    // Bundles are unioned, and unknown names are ignored so that tokens minted
    // by a newer service still verify. Only capability bits are masked; a
    // grant's legacy `capability` value is not attenuated.
```

crates/tables/src/behaviors.rs
```diff
@@ -1717,4 +1741,368 @@ mod test {
...
+                ("daveCo/", CapabilityBundle::Admin.capabilities()),
+                ("supportCo/", EnumSet::from(Assume)),
+            ],
+        );
+    }
+
+    #[test]
+    fn test_masked_walk_empty_mask_denies_all() {
+        // An identity-only token: every node is emitted fully attenuated,
+        // so no authorization can succeed anywhere and no prefix surfaces.
```
jshearer: With an empty mask, only the direct-grant nodes are emitted; downstream nodes are not reached:

```suggestion
        // An empty mask leaves only direct-grant nodes, with no capabilities;
        // the walk stops there and no prefix is listed.
```

crates/tables/src/behaviors.rs
```diff
@@ -162,10 +168,19 @@ impl super::UserGrant {
             &'a str,
             (authz::CapabilitySet, models::Capability),
         > = Default::default();
+
         for node in Self::reachable_nodes(role_grants, user_grants, subject) {
+            // Attenuation is only enabled when we have a capability mask set.
+            // We do this here because the get_user_capability doesn't filter
+            // these out so we have to do this here instead of inside of
+            // `reachable_nodes`.
```
jshearer: Explain both why filtering happens here and why the mask-presence check is retained:

```suggestion
            // Keep empty nodes in the walk for legacy capability reporting.
            // Omit them from prefix listings only for masked subjects, so a
            // masked token does not learn of grants it cannot use, while
            // existing unmasked listings are preserved.
```

crates/models/src/authz.rs
```diff
@@ -187,3 +226,166 @@ impl From<super::Capability> for CapabilitySet {
...
+    fn from(bundle: CapabilityBundle) -> Self {
+        bundle.capabilities()
+    }
+}
+
+/// This capability mask type. this is a set of capabilities associated with a
+/// token. This is a new type so we don't confuse a capability mask with
+/// a set of required capabilities, as they would both be a capability set.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub struct CapabilityMask(CapabilitySet);
```
jshearer: What does this newtype buy us over `Option<CapabilitySet>` on `Subject`? `new()` accepts any set, `apply()` is just intersection at one call site, and `Option` already distinguishes absent from empty. 

Any reason not to use the existing `CapabilitySet`?

crates/tables/src/behaviors.rs
```diff
@@ -130,16 +130,22 @@ impl super::UserGrant {
         user_grants: &'a [super::UserGrant],
         subject: &authz::Subject,
     ) -> impl Iterator<Item = super::NodeRef<'a>> + 'a {
-        // Copy out what the walk needs so the returned iterator borrows only
-        // the grant tables, not `subject`.
         let user_id = subject.user_id;
```
jshearer, line 135: Why was this deleted? The second copy down below makes the lifetime constraint it explains more important, not less: without it the two `let` bindings look like needless copies, and folding them into the closure would not compile.

```suggestion
        // Copy out what the walk needs so the returned iterator borrows only
        // the grant tables, not `subject`.
        let user_id = subject.user_id;
```

crates/tables/src/behaviors.rs
jshearer: Please remove this additional node-mutation stage and apply the optional mask to the existing `delegatable` set once, after computing it. Every `to_node_ref(delegatable)` path already intersects the edge's capabilities with that set, so the existing construction path can perform the full operation: `(edge_bits & delegatable) & mask` equals `edge_bits & (delegatable & mask)`.

This gives capability intersection one construction path without adding another transformation to each emitted node. Leave the terminal-node decision based on the current node's capabilities as it is.

crates/control-plane-api/src/server/public/graphql/authorized_prefixes.rs
```diff
@@ -286,12 +286,8 @@ mod tests {
             },
         ]);
         let rg = tables::RoleGrants::new();
-
-        let reachable = tables::UserGrant::reachable_prefixes(
-            &rg,
-            &ug,
-            &models::authz::Subject::unrestricted(ALICE),
-        );
+        let subject = models::authz::Subject::unrestricted(ALICE);
```
jshearer: Please revert this hoist and the one at line 326, along with the matching one in `behaviors.rs` at line 1348. They change no behavior in tests this PR does not otherwise touch, and the original calls were already rustfmt output.

crates/tables/src/behaviors.rs
```diff
@@ -1717,4 +1741,368 @@ mod test {
...
+
+        let subject_no_mask = authz::Subject {
+            user_id,
+            capability_mask: None,
+        };
+        // Unmasked: every reached prefix surfaces. supportCo/ emits its bare
+        // Assume bit, so it is non-empty and never a candidate for the skip;
+        // that Assume bit is what carries the walk on to daveCo/, which
+        // arrives with full Admin bits.
+        let unmasked = UserGrant::reachable_prefixes(&role_grants, &user_grants, &subject_no_mask);
```
jshearer, line 1952: This unmasked half does not protect the `is_some()` condition at line 177: no node in the fixture is empty when unmasked, so the test still passes if empty prefixes are filtered unconditionally. Add one unrelated direct grant with legacy capability `None` and no bundles, locally so the shared fixture's other tests keep their expectations, and assert its prefix survives unmasked and is omitted under the mask.

jshearer (changes requested):
Got some changes here mostly around some clean-ups and test behavior, as well as keeping comments correct and appropriately focused on the layer of the domain they're attached to.

crates/models/src/authz.rs
bbartman: To prevent parameter confusion. So the functions that take the parameter also accept something that could be accepted. Now that it's down to only a single place I believe removing it could be prudent, since it's kept inside `Subject` type.

crates/models/src/authz.rs
bbartman: Yes there is. So the this is a consequence of wanting to handle the case during a blue-green deployment that we can parse the names 1 at a time, and not issue an error message in the event that we don't recognize one of the names because a new bundle was issued from a newer server. The enforcement that the a CapabilityBundle must exist should only be enforced within the minting endpoint, where they are signed. 

The reasoning for using strum and not `serde` in this case is that the serde functions (at least for Json) require things to be in quotes, and if we parse them as a `Vec<String>` inside of the control claims, we have to re-add quotation marks in order for them to be parsable. If there's another way that we should be doing this inside of our Repo, I didn't see it. I did however see us using strum for things.

jshearer: Agreed on handling names individually and ignoring unknown ones. You’re right that `serde_json::from_str` would require JSON quotes, but Serde’s `StrDeserializer` works directly with the already-decoded string:

```rust
use serde::{de::value, Deserialize};

mask.iter()
    .filter_map(|name| {
        CapabilityBundle::deserialize(
            value::StrDeserializer::<value::Error>::new(name),
        )
        .ok()
    })
    .map(|bundle| bundle.capabilities())
    .fold(CapabilitySet::empty(), |set, bits| set | bits)
```

This keeps the current behavior, including ignoring unknown names, while reusing the enum’s existing name mapping. I verified it against the PR’s actual types, including mixed known/unknown entries.

bbartman: OK sounds fine then.

bbartman: Fixed. The implementation has moved into `ControlClaims`.

crates/models/src/authz.rs
bbartman: Fixed.

crates/models/src/authorizations.rs
bbartman: OK. The implementation was moved. The tests for it were also moved minus the one you indicated we didn't need below.

crates/models/src/authz.rs
bbartman: All of the extra stuff has been removed, the tests minus the `legacy_capability_parity` test have been migrated into the `authorization.rs` file with the new implementation for parsing of a capability mask.

crates/models/src/authz.rs
bbartman: OK this has been refactored and removed.

crates/tables/src/behaviors.rs
bbartman: OK this has been fixed.

crates/tables/src/behaviors.rs
bbartman: I added the tests case. So some of this is what I was wondering about when it comes to `get_user_capability` In the current implementation we can return a node with no capabilities, but that also is able to return a node with legacy capabilities, which is where I'm a bit confused.

We can talk about this outside of this PR.

bbartman:
@strix-security Ready for rereview.

crates/models/src/authorizations.rs
```diff
@@ -34,6 +46,34 @@ impl ControlClaims {
...
+    /// A `None` claim means the token has no mask and the subject keeps every
+    /// granted capability.
+    ///
+    /// An empty list yields an empty set, so the token is restricted from
+    /// anything that requires a capability.
+    ///
+    /// Unrecognized bundle names are dropped rather than rejected. A token
+    /// minted by a newer server may name bundles this version does not know,
+    /// and it should still verify with the capabilities we do understand.
+    fn parse_capability_mask(&self) -> Option<authz::CapabilitySet> {
```
jshearer: Could we keep this calculation local to `subject()` rather than introducing `parse_capability_mask()`? The helper has one production caller, its documentation repeats the field contract, and testing it directly lets the tests bypass both claim deserialization and the actual `ControlClaims` to `Subject` conversion.

A local `capability_mask` calculation before constructing `Subject` keeps the conversion readable. Then a compact table can deserialize JSON claims and assert on `subject()` for omitted, null, empty, recognized, mixed known/unknown, and `team_admin` cases, including preservation of `sub`. The casing and whitespace cases in `parse_capability_mask_is_strict_about_names` test Serde's exact matching rather than our code, so they can go, for the same reason as `from_name_is_strict`.

crates/tables/src/behaviors.rs
```diff
@@ -1717,4 +1741,396 @@ mod test {
...
+        ));
+    }
+
+    #[test]
+    fn test_masked_walk_get_user_capability() {
+        use Capability::*;
+
+        // Legacy-capability grants, so nodes carry a legacy value the
+        // mask must NOT attenuate: it's compatibility metadata, not an
+        // authorization decision.
```
jshearer: The mask intentionally leaves the legacy value unchanged here, but describing it as "metadata, not an authorization decision" is misleading while existing callers still use `get_user_capability()` for authorization. For example, publications require a legacy `Admin` value for each drafted spec, and `Read` for each spec it references:

https://github.com/estuary/flow/blob/447e5884467d67034c4ac8936fd278f0c8f22247/crates/control-plane-api/src/publications/specs.rs#L791-L796

https://github.com/estuary/flow/blob/447e5884467d67034c4ac8936fd278f0c8f22247/crates/control-plane-api/src/publications/specs.rs#L861-L864

I acknowledge that we'd like to clean these up asap, but could we state only the contract this test establishes?

```suggestion
        // Legacy capability values pass through unchanged for reached nodes;
        // capability masks attenuate only capability bits.
```

crates/tables/src/behaviors.rs
```diff
@@ -1717,4 +1741,396 @@ mod test {
...
+        };
+        let masked = UserGrant::reachable_prefixes(&role_grants, &user_grants, &subject);
+        assert_eq!(
+            masked.keys().collect::<Vec<_>>(),
+            vec![&"acmeCo/", &"bobCo/shared/", &"carolCo/upstream/"],
+        );
+    }
+
+    #[test]
+    fn test_masked_walk_parent_prefix_pickup() {
```
jshearer, line 1966: This test looks redundant: parent-prefix discovery is covered by `test_inherited_capabilities` and `test_data_plane_user_visibility`, and masking is applied once after all three edge streams are chained, so there is no parent-prefix-specific masking path for it to protect. The neighboring tests already cover masked Delegate and terminal behavior.

crates/tables/src/behaviors.rs
```diff
@@ -1717,4 +1741,396 @@ mod test {
...
+            &role_grants,
+            &user_grants,
+            &subject,
+            "daveCo/thing",
+            Capability::SpecEdit,
+        ));
+    }
+
+    #[test]
+    fn test_masked_walk_prefixes_omit_fully_attenuated() {
```
jshearer, line 1925: Now that `test_masked_walk_prefixes_keep_empty_when_unmasked` exists, this test is subsumed by it: the masked assertion is identical, and the unmasked assertion is a subset of the one there (which also lists `orphanCo/`). Could we drop this one? I'd meant for `orphanCo/` to be added to this test rather than a new one, but either way we only need one of them.

jshearer (commented):
Thanks for addressing that feedback. I've got a couple more changes, fairly small:

* Test the claim conversion through `subject()` on deserialized JSON, and fold `parse_capability_mask()` into it.
* Drop two tests that other tests already cover.
* Fix the legacy-capability comment so it doesn't say the value is not used for authorization.

Could you also make sure you squash before merging? A few of the commit messages (e.g. "Addressed some of the comments from an AI code review") shouldn't land in the history.

crates/models/src/authorizations.rs
bbartman: Fixed.

crates/tables/src/behaviors.rs
bbartman: Fixed.

crates/tables/src/behaviors.rs
bbartman: Fixed.

crates/tables/src/behaviors.rs
bbartman: Fixed.

bbartman:
@strix-security Ready for rereview.

jshearer (approved):
One edge case that came up in my review of this just now is that refresh tokens will allow you to bypass capability-scoped access tokens:

* The GraphQL `createRefreshToken` mutation only requires being logged in
* Exchanging a refresh token goes through the SQL `generate_access_token`, which never adds a mask, so you get back a full-access token.
* The fix is either to reject that mutation when a mask is present, or to store the mask on `refresh_tokens` and have the exchange copy it into the new token.

I'm inclined towards the former for now (forbidding masked tokens from calling `createRefreshToken` at all), but let's do that as a fast-follow rather than in this PR.

Other than that, I did a final exercise of this in a local stack and it looks like it all works as designed, so LGTM!

PR #3510, opened 2026-09-17
unseal: strip encrypted_suffix in Rust rather than shelling out to jq

dgreer-dev:
## Description

Removing the sops `encrypted_suffix` from decrypted keys was done by spawning `jq`
over the decrypted document. This replaces that with a short recursive walk over
the parsed document, which does the same key rewrite in-process.

Upside is one fewer process spawn per connector start, and `jq` is no longer a
runtime dependency of the reactor, dekaf, or control-plane-agent images. It was
only installed there for this one call, so it comes out of those Dockerfiles.

## Notes for reviewers

The old comment justified jq as avoiding the plaintext in our heap, but the
function already returns the plaintext as a `RawValue` in our heap and every
caller parses it, so this only adds one transient copy.

The jq removal from `control-plane-agent.Dockerfile` is safe because the agent
image contained nothing else that used it.


williamhbaker (approved):
LGTM

PR #3511, opened 2026-09-17
control-plane: 2973 big table stats - implement GCP BigTable client

adrian-estuary:
**Description:**
Adding a new `gcp` crate to contain all of the Bigtable client logic, including decoding the stream protocol (which is the bulk of the logic). 

**Workflow steps:**

This crate will be used in upcoming changes to read catalog stats from Bigtable. Intended usages is as follows:

```rust
let client = gcp::bigtable::client::new(...);

let table_name = "projects/SOME_PROJECT/instances/SOME_INSTANCE/tables/catalog_stats_hourly";
let row_set = get_rows_for_range(names, start, end);
let row_filter = filters::and(vec![
    filters::cells_per_column(1),
    filters::columns("f", &["flow_document"]),
]);

let req = ReadRowsRequest {
    table_name: table_name,
    rows: Some(row_set),
    filter: Some(row_filter),
    ..Default::default()
};

let rows = self.client.read_rows(req).await?; // Stream<Row>
let stats = rows.into(); // Stream<CatalogStats>
```

No external docs changed. Just added a README to the crate.

**Notes for reviewers:**

The decoder is the bulk of the change and is fairly complex to follow through. High-level it's a state machine with 3 states. There are a ton of comments in that file to help explain the different transitions.


Cargo.toml
```diff
@@ -268,6 +269,7 @@ assert_cmd = "2"
 criterion = "0"
 glob = "0"
 insta = { version = "1", features = ["redactions", "json", "yaml"] }
+mockall = "0"
```
adrian-estuary, line 272: used to generate mocks for easier unit testing

Cargo.toml
```diff
@@ -87,7 +87,8 @@ futures-core = "0"
 futures-util = "0"
 fxhash = "0"
 gcp_auth = "0.12"
-googleapis-tonic-google-bigtable-v2 = "0.36"
+googleapis-tonic-google-bigtable-v2 = "0.39"
+googleapis-tonic-google-bigtable-admin-v2 = "0.45"
```
adrian-estuary, line 91: will use this `bigtable-admin-v2` dep shortly for test utilities to easily seed data for integration tests

PR #3512, opened 2026-09-17
deps: upgrade sqlx to 0.9 and rusqlite to 0.39 (SQLite 3.51)

jgraettinger:
Dependency upgrades, with the follow-on changes each required.

## sqlx 0.8.6 => 0.9.0

- The `runtime-tokio-native-tls` feature is split into `runtime-tokio` and
  `tls-native-tls`.
- `Encode::encode_by_ref` drops the lifetime parameter of `ArgumentBuffer`,
  and non-literal SQL passed to `query()` now needs `AssertSqlSafe`.
- `ipnetwork` 0.20 => 0.21 no longer enables `serde` by default, so
  `data-plane-controller` declares it directly with the feature on.
- MSRV moves to 1.94. The pinned `sqlx-cli` in the mise tasks and the CI cache
  key move to 0.9.0.
- The `.sqlx/` prepared-query cache is regenerated in its own commit. sqlx 0.9
  records a column `origin` and describes with a generic plan, which changes
  inferred nullability for three columns that already carry explicit `!`/`?`
  overrides. No Rust types change.

## rusqlite 0.32.1 => 0.39.0

`libsqlite3-sys` is a `links` crate, so a single version is shared by the whole
workspace, including the SQLite that Go links through `libbindings.a`. sqlx 0.9
raises its `libsqlite3-sys` ceiling, which is what allows this bump.

- Bundled SQLite 3.46.0 => 3.51.3.
- `u64`/`usize` conversions are now gated behind rusqlite's `fallible_uint`
  feature; enabled to keep `ToSql`/`FromSql` as they were.
- rusqlite requires `time` >= 0.3.47, which deprecates the runtime
  `format_description::parse`; `avro` uses the compile-time macro instead.
- `hmac` and `sha2` are pinned to their current minor versions, which otherwise
  drift when the lockfile is re-resolved.

## json: hand-rolled `format: time` validator

RFC 3339 `full-time` is a fixed-width grammar, so it's now checked directly
rather than through `time`'s format-description machinery. This decouples the
validator from `time`'s parsing behavior, which shifts again in 0.3.47, and is
about 20x faster. Leap seconds (`23:59:60` at UTC) are now accepted, so the
previously-allowlisted cases of the official test suite are enabled and the
suite runs clean.

## derive-sqlite

While on a current SQLite, some tidying of the connection setup:

- A connection authorizer is installed after the bootstrap script runs, so it
  covers migrations and transform lambdas but not our own setup.
  `ATTACH`/`DETACH` and `load_extension()` are denied, and pragmas are restricted
  to an allowlist. This is backed by an attached-database limit of zero and
  SQLite's defensive mode.
- `BEGIN`/`COMMIT`/`ROLLBACK` from user SQL are denied, since the crate holds
  its own exclusive transaction between commits. `SAVEPOINT` remains allowed.
- A rejected statement is now reported with its text, via the `Error::Prepare`
  variant that `Lambda::new` already had available.

## Testing & Performance

`mise run ci:platform-test` and `ci:catalog-test`, including the Go test suite, which
exercises the recorded-log VFS against the newly bundled SQLite.

An agent additionally ran a sweep over the hand-rolled time validator in comparison
to the `time` crate, and confirmed all mismatches are expected (leap-year handling).

The hand-rolled time validator is about 10x faster than the prior implementation,
~34ns vs ~320ns per iteration under a throw-away critereon benchmark harness.


dgreer-dev (commented):
LGTM

dgreer-dev (approved):
LGTM (but with approve checked)

PR #3514, opened 2026-09-18
Control plane: cut over to the V2 connector service

jgraettinger:
When this PR deploys, the control plane stops using the V1 connector proxy and
uses the V2 `connector.Connector` service in its place.

Validate and Discover move over with behavior equivalent to V1. Beyond that,
connector `Spec` requests (connector tag discovery) and the derivation
lifecycles of publication catalog tests also run remotely, on data planes. The
control plane makes no further local connector invocations: the agent no longer
talks to a Docker daemon, and user connector images no longer run on
control-plane hosts. The agent's `--connector-network` flag is gone.

## Connector interface

The three streaming validation traits (capture / derive / materialize) collapse
into a single object-safe unary callable — `validation::Connectors`, one
`connector::Request` in and one `(Started, Kind)` response out — which is the
entire interface validation requires of the outside world. `Started` now carries the
connector's `Spec` and container metadata, so validation no longer opens a
separate Spec stream ahead of each Validate. The same callable serves discovers.

`ControlPlaneConnectorFactory` mints these callables per operation, bound to the
platform log sink and to the authorization `Snapshot`, which resolves a data
plane ID into a reactor route. `build::local` takes a connector router as a
parameter, so `flowctl` and the local test harnesses supply their own.

## Publication catalog tests

Catalog tests previously ran their derivations' connectors inside the agent
process, against the agent's Docker daemon. `CatalogTestConnectorRouter` now
resolves the owning data plane of every enabled built derivation and proxies that
derivation's connector to the plane's reactors, one route per plane. The test
harness itself still runs in the agent; only the connectors move.

This is setting up for first-class secrets, which will require proof of data-plane
residency for all connector starts that access secrets.

## Connector tags

Tag jobs ask a public data plane to run the image and report its protocol and
spec schemas. The protocol is inferred from existing tags of the same
image where possible, and probed across capture and materialization otherwise.

## Data planes resolve by ID

`tables::DataPlanes` existed so `validation` could forward a plane's details
for dialing. The `Snapshot` already holds this, so the table is removed and
validation deals only in `models::Id`. Storage mappings carry data-plane IDs
rather than names, resolved against the Snapshot up front (or against PostgREST,
in `flowctl`).

## Removals

- `runtime.ConnectorProxy`, its V1 capture / derive / materialize forwarding
  server, and its Rust client. `--proxy-runtimes` remains as an ignored
  compatibility flag; Go and Rust protocol bindings are regenerated.
- The V1 runtime's in-process Dekaf path. `dekaf_connector::connector()` now
  takes an already-resolved inner configuration and no longer parses the
  `{variant, config}` wrapper or decrypts a sops document of its own
  (removing a wart in the connector service, and setting up for secrets).
- `tables::CatalogResolver`, which had two implementations and no dynamic or
  generic use, in favor of direct calls.


crates/agent/src/connector_tags.rs
```diff
@@ -128,39 +127,88 @@ impl TagExecutor {
...
+                    .map(|spec| (ConnectorProtocol::Capture, spec))
+            }
+            Some(ConnectorProtocol::Materialization) => {
+                spec_materialization(&image_composed, connectors.as_ref(), data_plane_id)
+                    .await
+                    .map(|spec| (ConnectorProtocol::Materialization, spec))
             }
+            // Probe both protocols, taking the first that succeeds.
+            None => match spec_capture(&image_composed, connectors.as_ref(), data_plane_id).await {
+                Ok(spec) => Ok((ConnectorProtocol::Capture, spec)),
```
jshearer, line 196: “Probe both protocols, taking the first that succeeds” doesn't appear to reliably distinguish captures from materializations: capture and materialization `Spec` messages are wire-compatible enough for a materialization connector to satisfy the capture probe.

I had an agent exercise this e2e to confirm and `materialize-postgres` gets coded as `protocol=capture` when introduced to a fresh db.

<details>
<summary>Agent's output</summary>
# PR 3514 connector-tag database proof

Result: **the exact-binary run persisted `protocol = capture`, not
`materialization`**, while persisting `job_status = {"type":"success"}`.

The job began with:

```text
protocol=NULL
job_status={"type":"queued"}
```

After the real connector-tag executor completed and committed its outcome:

```text
image_name  | ghcr.io/estuary/materialize-postgres
image_tag   | :local
protocol    | capture
job_status  | {"type":"success"}
```

The image had `FLOW_RUNTIME_PROTOCOL=materialize` and contained the verified `materialize-postgres` binary. The exercised path was:

`TagExecutor → ControlPlaneConnectorFactory → authenticated gRPC connector service → Docker → flow-connector-init → materialize-postgres → committed PostgreSQL row`

The cause is now clear: a new tag tries the capture `Spec` first. The materialization response is protobuf-compatible with the capture response shape, so that probe succeeds. The materialization fallback never runs, and the executor writes `capture`.

The deployed reactor was represented by an authenticated loopback connector service, and we invoked the executor’s real `poll` and committed `Outcome::apply` directly rather than dequeuing through the automation server. Neither substitution changes protocol detection or persistence. The pending review remains untouched.

The proof was run from the isolated worktree
`/tmp/flow-pr3514-cancellation` at exact HEAD
`a47980f319409e3f86b8242b0d3e30fad5b906c0`. The pending review checkout
`/home/joseph/estuary/flow-pr3514` was read only and remained clean at the
same HEAD.

## Path exercised

The harness invoked the actual `agent::TagExecutor::poll`, followed by the
actual returned `TagOutcome::apply` in a committed PostgreSQL transaction.
The connector call used
`control_plane_api::connectors::ControlPlaneConnectorFactory`, a snapshot
loaded from PostgreSQL with `PgSnapshotSource`, an authenticated loopback
gRPC route, `connector::Service` with `Plane::Local`, the PR-head
`flow-connector-init`, and the Docker image
`ghcr.io/estuary/materialize-postgres:local`.

The image used in the decisive run contains
`/tmp/pr3514-materialize-postgres` byte-for-byte. Both hash to
`c7d69eb7653a64addef043e1b052db8d03cae7e0a4a566eb20940acd0fe42277`.
Its OCI label is `FLOW_RUNTIME_PROTOCOL=materialize`.

## Why the capture probe wins

At this HEAD, a NULL-protocol row tries `spec_capture` first and accepts its
first success. The connector service establishes its typed protocol from the
caller's outer request, so the capture probe opens the image through the
Capture RPC and wraps the returned Spec as Capture. The materialization Spec
wire payload is compatible enough with the capture Spec shape to decode and
pass the executor's response checks. The resulting capture-shaped Spec has
empty `resource_path_pointers`, matching the executor log in
`run-output.txt`. This path does not consult the image's
`FLOW_RUNTIME_PROTOCOL=materialize` label, and the fallback materialization
probe is never reached.

</details>

crates/control-plane-api/src/connectors.rs
```diff
@@ -0,0 +1,104 @@
...
+            let (data_plane_name, route) = resolved?;
+            let logger = move |log: &proto_flow::ops::Log| log_handler.log(log);
+
+            proto_grpc::connector::unary(
+                &route,
+                &logger,
+                request,
+                *CONNECTOR_TIMEOUT,
+                *CONNECTOR_TIMEOUT,
+            )
```
jshearer, line 68: This looks like a resource leak: a connector invocation started for validation or discovery can remain running indefinitely on the data plane after the originating request times out or is cancelled, if it stops producing responses and logs. Retrying can then start another invocation while the first is still alive.

I had an agent test this locally against the actual connector service over gRPC, using a silent subprocess connector. The caller timed out after one second, but the connector handler was still waiting for startup 120 seconds later. It only cleared when the test subprocess exited.

crates/control-plane-api/src/publications/specs.rs
```diff
@@ -913,88 +912,77 @@ pub async fn resolve_live_specs(
...
         };
+
+        // Map data-plane names to IDs, accruing mapping errors as we go.
+        // An unauthorized and missing plane have the same diagnostic.
+        let mut data_plane_ids = Vec::with_capacity(store.data_planes.len());
+        for name in &store.data_planes {
+            let data_plane = snapshot
+                .is_user_authorized(subject, name, Capability::Read)
+                .then(|| snapshot.data_plane_by_catalog_name(name))
+                .flatten();
```
jshearer, line 935: Does this end up requiring access to planes from unrelated mappings? Say I have admin on `acmeCo/team-a/` and can use plane A, while `acmeCo/team-b/` maps to private plane B that I can't access.

Reading this, publishing only `team-a/orders` would then fail on plane B before validation picks the mapping for team A. Am I following that right? Could we resolve only the partition/recovery mappings that apply to the draft?

crates/agent/src/connector_tags.rs
```diff
@@ -128,39 +127,88 @@ impl TagExecutor {
...
+                    plane.data_plane_name.starts_with("ops/dp/public/") && plane.can_sign()
+                })
+                .map(|plane| plane.control_id)
+            {
+                break data_plane_id;
             }
+
+            snapshot.request_refresh();
+            tracing::info!("awaiting a Snapshot with a ready public data-plane");
+            _ = refresh.expired().await;
```
jshearer, line 149: This appears to hold the automations slot open while waiting. If no public plane becomes ready, it looks like enough queued tags could occupy all the slots. The existing retry path seems to leave the tag queued and release the slot, any reason not to exit with an error so that we yield the slot back and retry later?

jshearer (commented):
I (and Fable+Astra) think I found a couple of bugs here. Two of them are related to `connector_tags` which, because that table is on the chopping block, IMO lowers their perceived severity. That being said, we'll _probably_ introduce at least one new materialization connector before we nuke `connector_tags`, so the fix is still worth it.

crates/agent/src/connector_tags.rs
jgraettinger: Yes, I'm aware of this one. There's a fast follow behind this PR which adds a guard in the data-plane. I deemed the window acceptable.

crates/agent/src/connector_tags.rs
jgraettinger: Yes: it means local stack startup races, loses, and awaits a 5 minute `automations` heartbeat backoff.

crates/control-plane-api/src/connectors.rs
jgraettinger: Dropping the response channel causes immediate teardown ([here](https://github.com/estuary/flow/blob/master/crates/connector/src/serve.rs#L176) during startup, and [here](https://github.com/estuary/flow/blob/master/crates/connector/src/serve.rs#L251) post-Open).

So, a client that goes away (HTTP/2 stream reset) is torn down. A client that's still patiently awaiting a connector, including graceful shutdown, is allowed to continue running in the data-plane. That's intended - let the client control timeout and retry policy.

Would you please be more specific about exactly what was reproduced?

crates/control-plane-api/src/publications/specs.rs
jgraettinger: Yes, in theory. In practice, data-planes are always granted to `acmeCo/`, never `acmeCo/team-a/`, so this scenario doesn't exist.

As a meta-comment, this section of code and the storage mapping concern certainly needs new stewardship. There are a number of interrelated concerns in how `data_plane_ids` are used. Tactically, removing `recovery/` storage mappings would potentially open up a refactor that removes them from tables::StorageMapping: one of their most-important current use cases is pinning as fact that a deprecated `recovery/` mapping exactly equals its non-`recovery/` mapping, to ensure their removal remains possible. After that, it mostly is used for admission checks and mapping novel specs into a data-plane, and it's not at all clear that `validation` is the right home for this.

crates/control-plane-api/src/connectors.rs
jshearer: When the control-plane `unary()` operation times out, it returns an error and drops its local response receiver. However, the [`EndpointRouter`'s forwarding task](https://github.com/estuary/flow/blob/a47980f319409e3f86b8242b0d3e30fad5b906c0/crates/proto-grpc/src/connector/router.rs#L95-L121) remains blocked solely on `tonic_rx.next()`, waiting for the next message from the connector.

If the connector sends another message, the forwarding task attempts delivery, discovers the receiver is gone, and drops the remote stream, all good. But if the connector stays silent (if it's hung waiting on a network response or something -- granted, any scenario I can think of here involves a connector bug), the forwarding task keeps that stream open, so the data-plane service never receives the disconnect that would trigger cleanup.

So concretely, don't we also need to select between `tonic_rx.next()` to handle the next connector msg _and_ `response_tx.closed()` to correctly propagate client-side (i.e control plane side) premature closures to the data-plane for cleanup?

crates/agent/src/connector_tags.rs
jshearer: Hm.. okay. I suppose the premise of "no public data plane available" isn't meaningful in production, fair point.

crates/control-plane-api/src/publications/specs.rs
jshearer: > Yes, in theory. In practice, data-planes are always granted to acmeCo/, never acmeCo/team-a/, so this scenario doesn't exist.

Mhmm... okay my bad, I was remembering the way we create storage mappings targeting nested prefixes in order to route tasks deployed to `acmeCo/dev` to one plane and `acmeCo/prod` to another. This is distinct from the referenced area above.

Buuuut, what about a storage mapping that names a no-longer-extant plane? Today, that only fails publications of specs directly into that nonexistent data-plane, whereas with this change, the existence of such a broken storage mapping anywhere under a tenant will break publications across the tenant, including things like auto-discovers.

That's not necessarily a bug, but according to this query, we have on the order of 50 such storage mappings, so I believe we'll want to clean them up before this goes live.

```sql
SELECT storage_mappings.catalog_prefix, dp.name
FROM storage_mappings, json_array_elements_text(storage_mappings.spec->'data_planes') dp(name)
WHERE NOT EXISTS (SELECT 1 FROM data_planes d WHERE d.data_plane_name = dp.name);
```

crates/agent/src/connector_tags.rs
jgraettinger: I've since pulled it into this PR.

crates/control-plane-api/src/connectors.rs
jgraettinger: Good catch! Yes, I missed that the client could hang. Fixed

crates/control-plane-api/src/publications/specs.rs
jgraettinger: I went back and did a bigger refactor of how specs are placed into data-planes, moving it out of `validation` altogether. `tables::StorageMapping::data_plane_ids` is gone, and placement is determined by `control_plane_api::publications::placements` before the build. Only data-plane names which are actually used get resolved and authorized. I'm pretty happy with how this came out.

jgraettinger:
PTAL. I had to rebase due to a conflict on master, but only added commits that should be individually coherent (starting with `control-plane-api: remove unused publications/db_complete.rs`)

crates/control-plane-api/src/publications/specs.rs
```diff
@@ -912,89 +911,67 @@ pub async fn resolve_live_specs(
...
+        };
+        let prefix = models::Prefix::new(prefix);
+
+        mappings.insert_row(&prefix, id, spec.stores, recovery.stores);
+        mapping_planes.insert(prefix, spec.data_planes);
+    }
 
-    Ok(live)
+    if let Some(orphan) = recoveries.keys().next() {
+        anyhow::bail!("storage mapping recovery/{orphan} has no {orphan} mapping");
```
jshearer: nit: any user can reach this by drafting a spec named `recovery/...` and get an error string containing (likely) a different tenant than their own.

Had an agent reproduce this with the real `resolve_live_specs` against a DB holding two correctly paired tenants, `aliceCo/` and `bobCo/`, as a user whose only grant is admin of `bobCo/`:

```
bobCo/ admin drafts ["bobCo/anvils"]
  => Ok: live.errors = [], storage mappings = ["bobCo/"]
bobCo/ admin drafts ["bobCo/anvils", "recovery/foo"]
  => Err: storage mapping recovery/aliceCo/ has no aliceCo/ mapping
```

jshearer (dismissed):
Thanks for addressing the feedback! :) LGTM % one non-blocking comment about error messages potentially leaking unrelated tenant names.

crates/control-plane-api/src/publications/specs.rs
jgraettinger: Amended to push the join down into the DB (making non-twin'd rows invisible), filtering `recovery/`

jgraettinger:
Okay, think I'm finally done here. Last batch of updates:
* Preserve a hard error if a mapping omits its recovery/ twin. Basically we're back at our current operational posture: a stray recovery/ mapping won't break anything (though it should be cleaned up), and a missing recovery/ mapping is an error.
* Remove Debug from Manual because it holds HMAC keys (only used on local development stacks, but still good to remove)
* Connector tags skips closed public planes (also skips planes without keys, which would have filtered cronut anyway, but better captures intent / defense in depth)

PTAL

jshearer (approved):
LGTM!

PR #3515, opened 2026-09-18
supabase: oauth edge function validation

Alex-Bair:
**Description:**

The OAuth edge function did not validate where a rendered URL pointed before using it. This PR adds that validation, along with some nearby tidying.

This PR's scope includes:

- Adding a `deno test` step to the control-plane workflow, and running the workflow on pull requests that touch the functions.
- Returning the error response when a connector or connector-tag lookup fails, instead of building it and discarding it. A missing row fell through to destructure null and surfaced as a 500. This affected `access-token.ts` and both lookups in `encrypt-config.ts`, where the second site also reported the wrong error.
- Logging only the origin and path of a failed token exchange.
- Adding `renderValidatedTemplateUrl`, which renders a template twice — once normally, once with caller values replaced by a sentinel — and compares the two hosts to derive what each template allows. Templates that leave the whole host to the caller, such as Salesforce `my_domain`, are checked against a per-provider host-suffix allowlist and fail closed for providers without an entry.
- Extracting the request handler out of `index.ts` so it is reachable from a test without binding a port.
- Routing both the authorize and token-exchange paths through the validated renderer, and answering a refusal with a 400.

See individual commits for more details.

**Notes for reviewers:**

Best reviewed commit-by-commit.


Alex-Bair:
@strix-security re-review please. I only removed a couple comments since you last reviewed & found no security issues, so you should still be happy.

supabase/functions/oauth/access-token.ts
```diff
@@ -28,25 +28,28 @@ export async function accessToken(req: Record<string, any>) {
...
     }
 
     const { oauth2_spec, oauth2_client_id, oauth2_injected_values, oauth2_client_secret } = data as OauthSettings;
 
-    const url = compileTemplate(
+    const url = renderValidatedTemplateUrl(
         oauth2_spec.accessTokenUrlTemplate,
-        {
-            redirect_uri: redirect_uri ?? "https://dashboard.estuary.dev/oauth",
+        oauth2_spec.provider,
```
skord, line 43: The validated template comes from `data`, which is `connector_config` straight off the request body when the caller supplies it (line 21). So the caller supplies the template *and* the provider, not just the values interpolated into them. A template with no interpolation at all has no sentinel in the reference render, lands in Case A, and is compared only against itself.

Confirmed through `handleRequest`:

```json
{"operation":"access-token","connector_config":{"oauth2_spec":{
  "provider":"x",
  "accessTokenUrlTemplate":"https://169.254.169.254/latest/meta-data/iam/security-credentials/",
  "accessTokenResponseMap":{"stolen":"/access_token"}}}}
```

Returns 200 with the metadata response mapped back into the reply. `flowctl raw oauth` already drives this path (crates/flowctl/src/raw/oauth.rs:197,232), so it is reachable by any authenticated user.

The mechanism here validates caller-supplied *values* against a trusted template. When the template itself is caller-supplied there is nothing left to validate against, so this path needs a different answer: either the allowlist applies to the rendered host unconditionally when `connector_config` was supplied, or `connector_config` stops being accepted for the token exchange.

supabase/functions/_shared/helpers.ts
```diff
@@ -64,6 +64,115 @@ export const generateUniqueRandomKey = (size: number = 40) => {
...
+// Replaces every non-empty string in a render context with SENTINEL. Rendering
+// a template with the result shows where caller values land in the URL.
+const withSentinels = (v: any): any => {
+    if (Array.isArray(v)) return v.map(withSentinels);
+    if (v !== null && typeof v === "object") {
+        return Object.fromEntries(
+            Object.entries(v).map(([k, inner]) => [k, withSentinels(inner)]),
+        );
+    }
+    if (typeof v === "string" && v.length > 0) return SENTINEL;
```
skord: `withSentinels` only substitutes strings, so a JSON number or boolean renders identically in both passes. The reference host then contains no sentinel, `marker` is -1, and Case A's exact-match compares the URL against itself and passes.

Against the real Salesforce template, `{"config":{"my_domain":2852039166}}` renders `https://2852039166/services/oauth2/token`. WHATWG `URL` parses an all-numeric host as IPv4, so `actual.hostname` is `169.254.169.254` and `HOSTNAME_RE` never sees it. Confirmed on this branch. `2130706433` gives `127.0.0.1`, `167772161` gives `10.0.0.1`, and `true` renders host `true`.

Sentinel every non-null scalar, not just non-empty strings.

supabase/functions/_shared/helpers.ts
```diff
@@ -64,6 +64,115 @@ export const generateUniqueRandomKey = (size: number = 40) => {
...
+};
+
+// Host suffixes each provider may reach. A provider needs an entry only when
+// its template leaves the whole host to the caller. Providers not listed here
+// fail closed.
+export const ALLOWED_HOST_SUFFIXES: Record<string, string[]> = {
+    salesforce: [".my.salesforce.com"],
+};
+
+const SENTINEL = "xsentinelx";
```
skord: `SENTINEL` is a fixed constant visible in this source, so a caller who supplies the literal `xsentinelx` can place a marker where the template has none and move Case detection to a weaker branch.

With template `https://{{{config.a}}}.{{{config.b}}}/oauth` and an unlisted provider (which should fail closed in Case C), `{a:"attacker.example", b:"xsentinelx"}` gives reference host `xsentinelx.xsentinelx`. `marker` is 0, `suffix` becomes `"xsentinelx"`, and `attacker.example.xsentinelx` ends with `.xsentinelx`, so Case B allows it. Confirmed on this branch. The same works through the `...fromCaller(params)` spread on access-token.ts:51, where any extra request key becomes a template variable.

Generate it per call, for example `crypto.getRandomValues` hex, so it cannot be guessed or supplied.

supabase/functions/_shared/helpers.ts
```diff
@@ -64,6 +64,115 @@ export const generateUniqueRandomKey = (size: number = 40) => {
...
+
+    let actual: URL, reference: URL;
+    try {
+        actual = new URL(url);
+        reference = new URL(skeleton);
+    } catch {
+        throw new OAuthUrlRefused("OAuth URL template did not render a valid URL");
+    }
+
+    if (actual.protocol !== reference.protocol) {
```
skord, line 141: `protocol`, `port` and `host` are compared against the reference, but `username` and `password` are not, so everything before an `@` in a caller-supplied host segment is invisible to every check below.

Template `https://{{{config.subdomain}}}.example.com` with `subdomain: "a@b.example.com"` renders `https://a@b.example.com.example.com/oauth/tokens` and is allowed. Confirmed on this branch. The `.example.com` the template meant to fix has become userinfo on a request that carries the platform client secret, and the caller controls a Basic credential on it.

`if (actual.username || actual.password) throw new OAuthUrlRefused(...)`, or compare both fields against the reference like the others.

supabase/functions/_shared/helpers.ts
```diff
@@ -64,6 +64,115 @@ export const generateUniqueRandomKey = (size: number = 40) => {
...
+    }
+    if (actual.port !== reference.port) {
+        throw new OAuthUrlRefused("OAuth URL port changed");
+    }
+    if (!HOSTNAME_RE.test(actual.hostname)) {
+        throw new OAuthUrlRefused(`OAuth URL host ${actual.host} is malformed`);
+    }
+
+    const labels = reference.hostname.split(".");
+    const marker = labels.indexOf(SENTINEL);
```
skord: `labels.indexOf(SENTINEL)` only matches a label that is *exactly* the sentinel, so a template that interpolates a caller value within a label is misclassified as Case A rather than caught.

`https://api-{{{config.sub}}}.example.com` gives reference host `api-xsentinelx.example.com`, `indexOf` returns -1, and Case A then requires the rendered host to equal that literal. Every real request is refused:

```
OAuth URL host api-acme.example.com does not match api-xsentinelx.example.com
```

Same for `https://{{{config.sub}}}-api.example.com`. This is an availability break rather than a hole: any connector whose OAuth template uses that shape starts returning 400 the moment this deploys. Worth auditing the published `connector_tags` for that pattern before merge.

At minimum the `marker === -1` branch should assert `!reference.hostname.includes(SENTINEL)` and refuse explicitly rather than falling through to exact match, so this fails loudly instead of being silently classified wrong.

skord (dismissed):
Ran the validator on this branch under Deno rather than reading it, and the two-render sentinel approach does not close the hole it targets. Two bypasses reach the token exchange with the platform `client_secret` attached, both confirmed against a Salesforce-shaped template. Five blocking comments inline. I have a handful of smaller findings (two paths that now 500 instead of 400, an unanchored `endsWith` in the allowlist, test and CI-pinning gaps) which I'll hold until the structure settles.

supabase/functions/oauth/access-token.ts
Alex-Bair: I considered this out of scope for the issue I'm currently assigned and am working on (822). I assumed this would get addressed by a separate PR by whoever picks up that other issue (823). Are you suggesting I should expand the scope of this PR to address that other issue as well?

supabase/functions/_shared/helpers.ts
jshearer: This one looks legit 👍

skord (approved):
LGTM

supabase/functions/_shared/helpers.ts
Alex-Bair: Thanks, fixed by sentineling every scalar. Falsy values become `""` so `is_sandbox: false` still picks the right host.

supabase/functions/_shared/helpers.ts
Alex-Bair: Thanks, fixed by drawing the sentinel per render.

supabase/functions/_shared/helpers.ts
Alex-Bair: Thanks, fixed by refusing any username or password on the rendered URL.

supabase/functions/_shared/helpers.ts
Alex-Bair: Thanks, fixed by refusing these shapes explicitly. No template in the connectors repo uses either.

PR #3516, opened 2026-09-18
supabase: require an authenticated user for inline oauth specs

skord:
## Problem

`oauth/index.ts` dispatches on the request body without establishing who the caller is. Most operations only name a `connector_id` and read the OAuth spec from the `connectors` table, but both `auth-url` and `access-token` also accept an optional `connector_config` that stands in for that row:

```ts
if (connector_config) {
    data = connector_config;
} else {
    // ... load from the connectors table
}
```

That path exists for `flowctl raw oauth`, where a developer is iterating on a connector's OAuth spec before it is in the table. Its callers hold real user tokens. But the function never required one, so the anonymous key that every request already carries was enough to take it.

## Fix

Gate `connector_config` on an authenticated user in the dispatcher, before any handler runs, and answer an anonymous request with a 401. This follows what `billing/index.ts` already does: build a client with the caller's `Authorization` header and require `auth.getUser()` to return a user.

Requests that omit `connector_config` are unchanged, which is every request the dashboard makes.

## Testing

`deno check --node-modules-dir=auto supabase/functions/oauth/index.ts` is clean for the changed file. It reports one pre-existing `TS2322` in `access-token.ts:34` (the nullable `output_data` at the `// TODO - check for empty data`), which #3515 fixes.

`crates/flowctl/src/raw/oauth.rs:189,255` is the only sender of `connector_config`, and it already attaches the user's access token with `bearer_auth` alongside the anon `apikey` header, so the `flowctl` workflow is unaffected.

## Note on #3515

#3515 moves this dispatcher into a new `oauth/handler.ts` and reduces `index.ts` to `serve(handleRequest)`. Whichever lands second takes a small conflict: the gate moves to the top of `handleRequest`. The two changes are independent, and neither subsumes the other.


skord:
Withdrawing this as @Alex-Bair is going to take this on

PR #3517, opened 2026-09-18
oauth: require a user when callers supply a `connector_config`

Alex-Bair:
**Description:**

The OAuth edge function accepts a caller-supplied `connector_config` in place of a row from the connectors table, and nothing checked who was asking. This PR requires an authenticated user for that path.

This PR's scope includes:

- Gating `connector_config` on an authenticated user in the dispatcher. When it is present, the caller's `Authorization` header must resolve to a user via `auth.getUser()`, and anything else gets a 401. Requests that name a connector are unchanged. This is the same check `billing/index.ts` already performs, and `flowctl raw oauth`, the one sender of `connector_config`, already attaches a user token.
- Turning off session handling on the per-request client that check uses. Outside a browser, supabase-js starts a token-refresh interval per client that is never cleared. The new tests surfaced the leak.

See individual commits for more details.

**Notes for reviewers:**

This picks up where #3516 left off, moved to `handler.ts` after #3515 landed.


skord (approved):
LGTM from my perspective

PR #3518, opened 2026-09-18
runtime-next: stop a session whose shuffle remainders pin too much disk

### LINKED ISSUE #3498: shuffle disk buffer deadlock detection
Detect a deadlock due to a full shuffle disk buffer, and error.

williamhbaker:
**Description:**

A remainder holds its log segment file open, and the Log reclaims disk only once that file is unlinked. Remainders which pin half of a shard's disk limit therefore hold the Log's back-pressure engaged for as long as the session runs. A shard now counts the segments its remainders pin and sends one Stop when that count is too high, so the next session starts without them.

**Workflow steps:**

**Notes for reviewers:**


crates/runtime-next/src/shard/mod.rs
```diff
@@ -39,6 +39,41 @@ pub(crate) async fn stopped_message(
...
+/// backlog is reclaimed as the shard consumes it, so the Log's backlog is
+/// roughly the pinned bytes. Back-pressure engages at the full disk limit, so
+/// half of it leaves the session a margin to stop within. Without the clamp, a
+/// limit under two segment thresholds would stop on any remainder at all.
+pub(crate) fn max_pinned_segments(labeling: &ops::proto::ShardLabeling) -> usize {
+    let limit_bytes = match labeling.shuffle_disk_limit_bytes {
+        0 => shuffle::DEFAULT_SHUFFLE_DISK_LIMIT_BYTES,
+        limit => limit,
+    };
+    ((limit_bytes / 2 / shuffle::log::writer::DEFAULT_SEGMENT_THRESHOLD) as usize).max(1)
```
jgraettinger, line 54: nit: not for this PR, but perhaps we rename to `writer::SEGMENT_THRESHOLD`. I don't believe we allow it to be configured. If it could float, that of course makes this calculation sensitive to having the same value on both sides, but we can cross that bridge later.

jgraettinger (approved):
LGTM

